PR #3842, opened 2022-02-17
Temp list variable documentation

sanderr:
# Description

- Made an attempt to clear up confusing nature of `TempListVariable` naming and semantics. @wouterdb could you check for correctness? @arnaudsjs could you check for clarity?
- Introduced ABC for result variables associated with a relation attribute. @arnaudsjs I added this because I believed it might help with determining which `DelayedResultVariable` instances are relevant for #3098, because your remark yesterday got me thinking that this wasn't very clear in the current naming and typing. If you don't think this adds value I can revert this part.

loosely related to #3098

# Self Check:

Strike through any lines that are not applicable (`~~line~~`) then check the box

- [ ] Attached issue to pull request
- [ ] Changelog entry
- [ ] Type annotations are present
- [ ] Code is clear and sufficiently documented
- [ ] No (preventable) type errors (check using make mypy or make mypy-diff)
- [ ] Sufficient test cases (reproduces the bug/tests the requested feature)
- [ ] Correct, in line with design
- [ ] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )

# Reviewer Checklist:

- [ ] Sufficient test cases (reproduces the bug/tests the requested feature)
- [ ] Code is clear and sufficiently documented
- [ ] Correct, in line with design

PR #3860, opened 2022-02-21
Add support for type precedence hints

### LINKED ISSUE #3098: Add support for type precedence hints
- [x] Add the ability to load hints from project.yaml
- [x] The hints should be of the form `<at:type>.<ar:relation> before <bt:type><br:relation>` 
- [x] Ensure the compile can not freeze any collection of `<bt:type><br:relation>` if there is an unfrozen `<at:type>.<ar:relation>`
- [ ] Estimate compiler overhead
- [x] Considered experimental 
- [x] Log back hints 
- [x] Document the usage with a big fat warning

arnaudsjs:
# Description

Add support to pass type hints to the compiler that indicate which entity relationships should be frozen first. 

closes #3098 

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (#3945)

# Reviewer Checklist:


changelogs/unreleased/add-type-precedence-rules.yml
```diff
@@ -0,0 +1,7 @@
+---
+description: Add support to pass type precedence hints to the compiler
+issue-nr: 3098
+change-type: minor
+destination-branches: [master, iso5]
+sections:
+  feature: "{{description}}"
```
arnaudsjs, line 7: Should I mention this in the changelog or not?

changelogs/unreleased/add-type-precedence-rules.yml
```diff
@@ -0,0 +1,7 @@
+---
+description: Add support to pass type precedence hints to the compiler
+issue-nr: 3098
+change-type: minor
+destination-branches: [master, iso5]
```
arnaudsjs, line 5: Should this also go into ISO4?

src/inmanta/execute/scheduler.py
```diff
@@ -430,3 +470,220 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
...
+            return []
+        work: Set[TypePrecedenceGraphNode] = set(self.root_nodes)
+        result: List[EntityRelationship] = []
+
+        def get_next_ready_item_in_work() -> TypePrecedenceGraphNode:
+            assert work
+            for current_node in work:
+                if all(dep.entity_relationship in result for dep in current_node.dependencies):
+                    return current_node
+            raise CycleInTypeHintsError("Cycle in type hints")
```
arnaudsjs, line 667: Should I print the set of rules that cause the cycle. I assume that's not required, since the set of type will should be rather limited at all times. Correct?

src/inmanta/module.py
```diff
@@ -1120,8 +1168,21 @@ class ProjectMetadata(Metadata, MetadataFieldRequires):
       project.yml.
     :param freeze_operator: (Optional) This key determines the comparison operator used by the freeze command.
       Valid values are [==, ~=, >=]. *Default is '~='*
+    :param type_hints: [EXPERIMENTAL FEATURE] A list of type hints that indicate the order in which the compiler
+                       should freeze list. A type hint should be specified using the following format:
+                       `<first-type>.<relation-name> before <then-type>.<relation-name>`. With this rule in place,
+                       the compiler will first freeze `first-type.relation-name` and only then `then-type.relation-name`.
     """
 
+    _re_identifier = r"[a-zA-Z_][a-zA-Z_0-9-]*"
```
arnaudsjs, line 1177: Please pay specific attention to this during the review. I am not entirely sure whether this is correct.

src/inmanta/execute/scheduler.py
```diff
@@ -273,6 +279,41 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
         # first evaluate all definitions, this should be done in one iteration
         self.define_types(compiler, statements, blocks)
 
+        # Validate the type hints to the defined types
+        valid_type_hints = []
+        if self.type_hints:
+            LOGGER.warning(
+                "[EXPERIMENTAL FEATURE] Using type hints defined in the project.yml file to determine list freeze order."
+            )
+            for hint in self.type_hints:
```
wouterdb, line 288: this should go into a separate method I think

src/inmanta/execute/scheduler.py
```diff
@@ -430,3 +470,220 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
...
+    in the following order:
+
+    * First return the DelayedResultVariables, which are not an instance of
+      TempListVariable and that do not have an order constraint.
+    * Then return DelayedResultVariables with order constraint.
+      They are returned in an order that is valid with respect to the
+      constraints.
+    * Finally, all TempListVariables are returned.
+
+    TempListVariables is subclass of DelayedResultVariables that is not
```
wouterdb, line 511: Why last and not first?

wouterdb (changes requested):
1. what is the performance overhead and how does it scale with additional type hints?
2. if I understand correctly, this mechanism is imprecise because it only takes into account the waitqueue and not the zerowaiters and not things that have waiting providers?

I would have expected a reference counting mechanism on the `RelationAttribute` object itself. Where the 'RelationAttribute' objects are linked according to the type precedence before actual evaluation starts, such that `EntityRelationship` don't have to be repeatedly constructed and matched, but the actual `RelationShip` can just verify if the 'unfrozen_instances_count' of any of its predecessors is >0. (But my idea may create a larger overhead, because it is used on every relation, even if it has no hints)

src/inmanta/execute/scheduler.py
```diff
@@ -430,3 +470,220 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
             raise RuntimeException(stmt.expression, "not all statements executed %s" % all_statements)
 
         return True
+
+
+@dataclass(frozen=True)
+class EntityRelationship:
```
wouterdb, line 476: Why have this instead of using the actual 'RelationAttribute' ?

I expect this object will be created a gazillion times?

src/inmanta/execute/scheduler.py
```diff
@@ -430,3 +470,220 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
...
+            return self._get_next_constraint_variable()
+        except IndexError:
+            # Empty
+            pass
+        return self._tmp_list_variables.popleft()
+
+    def _get_next_constraint_variable(self) -> DelayedResultVariable[Any]:
+        if not self._constraint_variables:
+            raise IndexError()
+        while len(self._freeze_order_working_list) > 0:
```
wouterdb, line 579: isn't using length slow compared to truthiness? (this is on the compiler critical path)

wouterdb (commented):
The overhead when not using hint actually looks quite good, I think (small test)

src/inmanta/execute/scheduler.py
sanderr: https://github.com/inmanta/inmanta-core/pull/3842/files#diff-e7d9b287c2e82ef2aa1f4b76c981aca767736791144db695c7770cb13e8cfac0R407

src/inmanta/execute/scheduler.py
```diff
@@ -430,3 +470,220 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
...
+    """
+    A queue for DelayedResultVariables that is prioritized based on the
+    type hints passed to the Compiler. This queue will return elements
+    in the following order:
+
+    * First return the DelayedResultVariables, which are not an instance of
+      TempListVariable and that do not have an order constraint.
+    * Then return DelayedResultVariables with order constraint.
+      They are returned in an order that is valid with respect to the
+      constraints.
```
sanderr, line 508: I wonder if this could cause any issues. It is definitely the simplest and most intuitive approach, but it also means that simply adding a type hint that defines an order between two relations shifts both relations towards the end of the freezing order, which effectively orders them with respect to all other relations as well. I don't see any immediate issues with this myself but I think we need to carefully consider all consequences.
(@wouterdb)

sanderr (commented):
Partial review only at the moment.

sanderr (commented):
I think we also need this in `find_wait_cycles`. Perhaps construct a new prioritized queue from all drvs in that method? It could be an expensive operation but I believe `find_wait_cycles` is very much an edge case and not on the critical path.
@wouterdb do you agree we need it in that method as well? Do you agree a small performance loss there is not really a concern?

src/inmanta/execute/scheduler.py
wouterdb: I don't think that answers the question. 

If the hinting is 100% accurate, they should come out first, because if they are safe to freeze, they should be frozen?

src/inmanta/execute/scheduler.py
```diff
@@ -66,9 +69,57 @@ class Scheduler(object):
...
+                if not isinstance(current_type, Entity):
+                    raise InvalidTypeHintException(f"A type hint was defined for non-entity type {current_type}")
+                assert isinstance(current_type, Entity)  # Make mypy happy
+                attributes_of_entity = current_type.attributes
+                if relationship_name not in attributes_of_entity:
+                    raise InvalidTypeHintException(
+                        f"A type hint was defined for {entity_type_name}.{relationship_name}, "
+                        f"but entity {entity_type_name} doesn't have an attribute {relationship_name}.",
+                    )
+                if not isinstance(attributes_of_entity[relationship_name], RelationAttribute):
```
wouterdb, line 116: I would use this code to attach the type hint to the relation, to remove lookup all over the place, which should further improve performance.

code like `self.queues.has_type_hint(self.type)` would become `self.attribute.hints`

src/inmanta/execute/scheduler.py
wouterdb: I also think this can be removed, if '_get_valid_type_hints' place the actual relation in the type hint

src/inmanta/execute/scheduler.py
wouterdb: agreed, but I have no further insights either.

src/inmanta/module.py
wouterdb: it should match what is in the lexer, I think we don't allow `-` anymore

wouterdb (changes requested):

sanderr (changes requested):
I just realized I find "type hint" to be a rather confusing name. Something like "precendence hint" or "type ordering hint" would make more sense in my opinion. But I realize I might be a bit late with this remark as it might be a hassle to update all references at this time. I'm fine with leaving it as is since it's a highly experimental feature that should only be used as a temporary solution by someone who knows what they're about.

changelogs/unreleased/add-type-precedence-rules.yml
wouterdb: unfortunately too big for ISO4 . I'm afraid.

src/inmanta/execute/runtime.py
```diff
@@ -527,6 +537,18 @@ def get_value(self) -> "Instance":
...
+    def get_progress_potential(self) -> int:
+        """How many are actually waiting for us"""
+        result = len(self.waiters)
+
+        from inmanta.ast.attribute import RelationAttribute
+
+        if isinstance(self.attribute, RelationAttribute):
+            # Ensure that relationships with a type hint cannot end up in the zerowaiters queue
+            # of the scheduler. We know the order in which those types can be frozen safely.
+            result += int(self.attribute.has_type_hint())
```
wouterdb, line 549: I would prefer to add `has_type_hint` to all `Attribute` types and have it return False by default. (avoid instanceof)

src/inmanta/execute/scheduler.py
```diff
@@ -430,3 +490,207 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
...
+        """
+        return (
+            len(self._unconstraint_variables)
+            + sum(len(v) for v in self._constraint_variables.values())
+            + len(self._tmp_list_variables)
+        )
+
+    def append(self, drv: DelayedResultVariable[object], dont_reset_working_list: bool = False) -> None:
+        """
+        Append on the right side of the queue.
```
wouterdb, line 556: It would be good to document why there is `dont_reset_working_list`

wouterdb (changes requested):
We need to make a ticket for the end-user doucmentation part

src/inmanta/execute/scheduler.py
```diff
@@ -647,76 +618,76 @@ class TypePrecedenceGraph:
...
-            self.add_precedence_rule(first_type, then_type)
+        for first_attribute in relation_attributes_with_type_hints:
+            for then_attribute in first_attribute.type_hints:
+                self.add_precedence_rule(first_attribute, then_attribute)
 
-    def add_precedence_rule(self, first_type: EntityRelationship, then_type: EntityRelationship) -> None:
+    def add_precedence_rule(self, first_attribute: RelationAttribute, then_attribute: RelationAttribute) -> None:
         """
-        Add a rule that `first_type` should be frozen before `then_type`.
+        Add a rule that `` should be frozen before `then_attribute`.
```
sanderr, line 634:
```suggestion
        Add a rule that `first_attribute` should be frozen before `then_attribute`.
```

sanderr (changes requested):

wouterdb (approved):

sanderr (changes requested):
This [discussion](https://github.com/inmanta/inmanta-core/pull/3860#discussion_r815955029) remains open. The rest looks good to me.

src/inmanta/execute/scheduler.py
```diff
@@ -430,3 +488,238 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
...
+        """
+        if entity_relationship not in self.type_to_node:
+            node = TypePrecedenceGraphNode(entity_relationship)
+            self.type_to_node[entity_relationship] = node
+        else:
+            node = self.type_to_node[entity_relationship]
+
+        if attach_to_root:
+            self.root_nodes.add(node)
+        return node
```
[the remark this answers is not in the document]
arnaudsjs: Fixed

sanderr (approved):

src/inmanta/ast/attribute.py
```diff
@@ -163,14 +163,18 @@ class RelationAttribute(Attribute):
     """
 
     def __init__(self, entity: "Entity", value_type: "Type", name: str, location: Location) -> None:
+        """
+        :param freeze_dependents: Contains the set of RelationAttributes that can only be frozen
```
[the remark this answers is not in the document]
arnaudsjs: Yes. Indeed. It should be `ivar`

PR #3861, opened 2022-02-21
Issue/3803 disallow multiline strings with single quotes

### LINKED ISSUE #3803: disallow multiline strings with single quotes
The following is currently valid syntax:
```
std::print(
    "test
    string"
)
```
It should not be. This is technically a breaking change, so the branches to apply this on should be carefully considered.

Hugo-Inmanta:
# Description

Trying to write a single quoted string over multiple lines now raises a syntax error.

closes #3803 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3803-disallow-multiline-strings-with-single-quotes.yml
+++ b/changelogs/unreleased/3803-disallow-multiline-strings-with-single-quotes.yml
@@ -1,6 +1,7 @@
-description: Trying to write a single quoted string over multiple lines now raises a syntax error.
+description: Writing a string over multiple lines is now only supported for strings within triple quotes.
 change-type: major
 destination-branches: [master, iso5]
 sections: {
-  bugfix: "{{description}}"
+  bugfix: "{{description}}",
+  deprecation-note: "{{description}} This was previously allowed for strings within single quotes due to a bug."
 }
--- a/changelogs/unreleased/3803-disallow-multiline-strings-with-single-quotes.yml
+++ b/changelogs/unreleased/3803-disallow-multiline-strings-with-single-quotes.yml
@@ -1,0 +1,6 @@
+description: Trying to write a single quoted string over multiple lines now raises a syntax error.
+change-type: major
+destination-branches: [master, iso5]
+sections: {
+  bugfix: "{{description}}"
+}
--- a/changelogs/unreleased/3803-disallow-multiline-strings-with-single-quotes.yml
+++ b/changelogs/unreleased/3803-disallow-multiline-strings-with-single-quotes.yml
@@ -1,5 +1,5 @@
 description: Writing a string over multiple lines is now only supported for strings within triple quotes.
-change-type: major
+change-type: minor
 destination-branches: [master, iso5]
 sections: {
   bugfix: "{{description}}",
--- a/src/inmanta/parser/plyInmantaLex.py
+++ b/src/inmanta/parser/plyInmantaLex.py
@@ -65,7 +65,7 @@ tokens = ["INT", "FLOAT", "ID", "CID", "SEP", "STRING", "MLS", "MLS_END", "CMP_O
 
 
 def t_RSTRING(t: lex.LexToken) -> lex.LexToken:  # noqa: N802
-    r"r(\"([^\\\"]|\\.)*\")|r(\'([^\\\']|\\.)*\')"
+    r"r(\"([^\\\"\n\r]|\\.)*\")|r(\'([^\\\'\n\r]|\\.)*\')"
     t.value = t.value[2:-1]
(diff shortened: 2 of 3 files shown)


Hugo-Inmanta:
Improvement suggestion:
We could improve the error message. Currently this snippet :
```
std::print(
  r"test
  string"
  )

```
Would raise:
`Syntax error: Illegal character '"'`
, which corresponds to the opening quote of the string. This message isn't indicative of the real problem (That multilines strings aren't allowed within single quotes).

changelogs/unreleased/3803-disallow-multiline-strings-with-single-quotes.yml
```diff
@@ -0,0 +1,6 @@
+description: Trying to write a single quoted string over multiple lines now raises a syntax error.
```
sanderr, line 1: I find "single quoted" a bit of an ambiguous wording if we put this in the changelog but on the other hand I wouldn't know how to say this otherwise so no objections.

changelogs/unreleased/3803-disallow-multiline-strings-with-single-quotes.yml
```diff
@@ -0,0 +1,6 @@
...
+change-type: major
+destination-branches: [master, iso5]
+sections: {
+  bugfix: "{{description}}"
```
sanderr, line 5:
```suggestion
  bugfix: "{{description}}"
  deprecation-note: "{{description}} This was previously allowed due to a bug."
```
I think we want this in both sections.

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -65,7 +65,7 @@
 
 
 def t_RSTRING(t: lex.LexToken) -> lex.LexToken:  # noqa: N802
-    r"r(\"([^\\\"]|\\.)*\")|r(\'([^\\\']|\\.)*\')"
+    r"r(\"([^\\\"\n\r]|\\.)*\")|r(\'([^\\\'\n\r]|\\.)*\')"
```
wouterdb, line 68: why `\r`? at this point there should be no '\r`

tests/compiler/test_strings.py
```diff
@@ -50,3 +50,88 @@ def test_rstring(snippetcompiler):
...
+
+def test_no_multiline_in_single_quoted_string_double_apostrophe(snippetcompiler):
+    snippetcompiler.setup_for_error(
+        """
+std::print(
+  "test
+  string"
+)
+        """,
+        "Syntax error: Illegal character '\"' ({dir}/main.cf:3:3)",
```
sanderr, line 63: You're right that this is very unintuitive (many of our syntax errors are). A possible improvement could be to raise the error in the parser rather than the lexer. It's only a start but I think I'm in favor of taking that approach. @wouterdb wdyt?

sanderr (changes requested):
Add a deprecation note. My other comments are just open discussions.

changelogs/unreleased/3803-disallow-multiline-strings-with-single-quotes.yml
Hugo-Inmanta: How about :
```suggestion
description: Writing a string over multiple lines is now only supported for strings within triple quotes.
```

sanderr: I like it!

src/inmanta/parser/plyInmantaLex.py
sanderr: Though it doesn't hurt to have it here either and the fact that there should be no `\r` is an implementation detail more than an invariant.

wouterdb: it is rather important, because the line numbering code also only does '\n'

tests/compiler/test_strings.py
wouterdb: It would improve the error reporting in this case, it might make it worse in case you really have a missing '\"'
(i.e. what would this do)

```
entity Test:
   string blah = "the value
end

[rest of model]

first_string_after_that = "a"
```

or worse

```
entity Test:
   string blah = "the value
end

[rest of model]
entity Test2:
   string blah = the value"
end

```

wouterdb (commented):
see examples I added

tests/compiler/test_strings.py
sanderr: Fair point. To improve that we would need to shift quite a bit more to the parser. Let's leave it as is then?

sanderr (approved):

wouterdb (approved):

PR #3878, opened 2022-02-22
Improved SSL documentation

wouterdb:
# Description
Updated SSL documentation in response to a support ticket.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/administrators/auth.rst
+++ b/docs/administrators/auth.rst
@@ -19,6 +19,7 @@ For the autostarted agents and compiler to work, either add the CA cert to the t
 set :inmanta.config:option:`server.ssl-ca-cert-file` to the truststore.
 
 .. code-block:: ini
+
     [server]
     # The ssl certificate used by the server
     ssl_cert_file=/etc/inmanta/server.crt
--- a/docs/administrators/auth.rst
+++ b/docs/administrators/auth.rst
@@ -5,14 +5,37 @@ Setting up authentication
 
 This guide explains how to enable ssl and setup authentication.
 
-SSL
----
 SSL is not strictly required for authentication but higly recommended. Inmanta uses bearer tokens
 for authorizing users and services. These tokens should be kept private and are visible in plain-text in the request headers
 without SSL.
 
+
+SSL: server side
+----------------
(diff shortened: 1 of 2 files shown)


docs/administrators/auth.rst
```diff
@@ -5,15 +5,38 @@ Setting up authentication
...
+    # The ssl certificate used by the server
+    ssl_cert_file=/etc/inmanta/server.crt
+    # The private key used by the server, associated with the certificate
+    ssl_key_file=/etc/inmanta/server.key.open
+
+    # The certificate chain that the compiler and agents should use to validate the server certificate
+    ssl_ca_cert_file=/etc/inmanta/server.chain
+    # The address at which the compiler and agent should connect
+    # Must correspond to hostname the ssl certificate is bound to
+    server_address=localhost
```
arnaudsjs, line 32: An empty line is required between the directive and the body.

```suggestion
.. code-block:: ini

    [server]
    # The ssl certificate used by the server
    ssl_cert_file=/etc/inmanta/server.crt
    # The private key used by the server, associated with the certificate
    ssl_key_file=/etc/inmanta/server.key.open

    # The certificate chain that the compiler and agents should use to validate the server certificate
    ssl_ca_cert_file=/etc/inmanta/server.chain
    # The address at which the compiler and agent should connect
    # Must correspond to hostname the ssl certificate is bound to
    server_address=localhost
```

docs/administrators/auth.rst
```diff
@@ -5,15 +5,38 @@ Setting up authentication
...
+
+
+
+SSL: remote agents and remote compiler
```
arnaudsjs, line 36: I find this title a bit strange. It suggests that there is something like a local agent or a local compiler, which is never the case.

```suggestion
SSL: agents and compiler
```

arnaudsjs (changes requested):

docs/administrators/auth.rst
wouterdb: what about autostarted agent and compiler-on-the-server?

arnaudsjs (approved):

PR #3881, opened 2022-02-23
Add not equals status filter to resource list endpoint

### LINKED ISSUE #3845: Add support to add a not to the resource state in the v2 resource list
Currently only inclusive filters can be used. However, to hide orphans by default it makes a lot more sense to use "!orphans'

andraskvr:
# Description

closes #3845 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -86,6 +86,8 @@ class QueryType(str, enum.Enum):
     IS_NOT_NULL = enum.auto()
     CONTAINS_PARTIAL = enum.auto()
     RANGE = enum.auto()
+    NOT_CONTAINS = enum.auto()
+    COMBINED = enum.auto()
 
 
 class RangeOperator(enum.Enum):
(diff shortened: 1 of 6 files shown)


src/inmanta/data/__init__.py
```diff
@@ -1394,6 +1412,37 @@ def get_contains_filter(
...
+    ) -> Tuple[str, List[object]]:
+        """
+        Returns a tuple of a PostgresQL statement and any query arguments to filter on values that are contained in a given
+        collection.
+        """
+        cls.validate_field_name(name)
+        filter_statement: str
+        values: List[object]
+        (filter_statement, values) = cls._combine_filter_statements(
+            (cls.get_filter_for_query_type(query_type, name, val, index + i, col_name_prefix))
```
wouterdb, line 1427: if I understand correctly, this assumes that every filter has exactly one value. I see no reason why this assumption would hold. Nothing of this nature is documented and many of the filters have no value, in the future we may add some with multiple values.

This method is an accident waiting to happen.

arnaudsjs (changes requested):

src/inmanta/server/validate_filter.py
```diff
@@ -203,6 +203,42 @@ def to_query_type(self) -> Optional[Tuple[QueryType, object]]:
...
+    field: Optional[Dict[QueryType, List[ReleasedResourceState]]]
+
+    @validator("field", pre=True)
+    @classmethod
+    def parse_field(cls, v: object) -> Optional[Dict[QueryType, ReleasedResourceState]]:
+        if v is None:
+            return None
+        if isinstance(v, list):
+            status_contains_filters = [status_filter for status_filter in v if not status_filter.startswith("!")]
+            status_not_contains_filters = [status_filter[1:] for status_filter in v if status_filter.startswith("!")]
```
wouterdb, line 220: This is part of the input validation logic. What happens if we pass in something that is not indexable (an integer e.g.)?  or has no startswith method?

wouterdb (changes requested):

src/inmanta/server/validate_filter.py
```diff
@@ -203,6 +203,42 @@ def to_query_type(self) -> Optional[Tuple[QueryType, object]]:
         return None
 
 
+class CombinedContainsFilterResourceState(Filter):
+    """Represents a valid ReleasedResourceState constraint,
+    which handles the filters as contains or not contains filters based on their values"""
+
+    # Pydantic doesn't support Generic models on python 3.6
```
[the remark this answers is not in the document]
andraskvr: The comment was left over from the previous version, I removed it (since this one has a custom validator anyway). 
But you're right that this isn't a blocker anymore in general so I'll check if that works (and if it's simpler) and create a separate ticket for it

arnaudsjs (approved):

src/inmanta/data/__init__.py
```diff
@@ -1287,20 +1295,19 @@ def _get_composed_filter(
         values = []
         index_count = max(1, offset)
         for key, value in query.items():
-            (filter_statement, value) = cls._get_filter(key, value, index_count, col_name_prefix=col_name_prefix)
+            name = cls._add_column_name_prefix_if_needed(key, col_name_prefix)
```
wouterdb, line 1298: Why is the validate not here?

src/inmanta/data/__init__.py
wouterdb: The typing/naming/documentation (`# The value describes a combination of other query types`)  suggest that this method combines filters, with no specific constraint on what kind of filters. Within that definition, it is broken, because it can not e.g. combine an is_null filter safely/correctly).

wouterdb (approved):

PR #3897, opened 2022-02-23
issue3826/ Wrong repo path leads to unclear error

### LINKED ISSUE #3826: Wrong repo path leads to unclear error
When you include more than one `{}` in your repo path you will get an unclear error:

```bash
DEBUG    inmanta.module:module.py:286 could not clone repo
  File "/home/bart/dev/bics_ce/env/lib64/python3.6/site-packages/inmanta/module.py", line 279, in clone
    url = self.baseurl.format(name)
IndexError: tuple index out of range
```

This can happen for example when you forget the the repo separator is ` ` and not `:`

The indexerror can only happen in such a case and should be transformed in a clear error message.

- use replace '{}' instead of 'format'
- allow only one occurrence of '{}'

FloLey:
# Description

When including more than one {} in a repo path we get an unclear error.
This commit logs a more readable error message.

closes #3826 

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3826-error-msg-wrong-repo.yml
+++ b/changelogs/unreleased/3826-error-msg-wrong-repo.yml
@@ -1,0 +1,3 @@
+description: An error message is now shown if a wrong repo path is used
+change-type: patch
+destination-branches: [master, iso4, iso5]
--- a/changelogs/unreleased/3826-error-msg-wrong-repo.yml
+++ b/changelogs/unreleased/3826-error-msg-wrong-repo.yml
@@ -1,3 +1,4 @@
 description: An error message is now shown if a wrong repo path is used
 change-type: patch
 destination-branches: [master, iso4, iso5]
+sections: { bugfix: "{{description}}" }
--- a/tests/moduletool/test_repos.py
+++ b/tests/moduletool/test_repos.py
@@ -16,6 +16,7 @@
     Contact: code@inmanta.com
 """
 import os
+import shutil
 
 from inmanta.module import InvalidMetadata, LocalFileRepo, RemoteRepo, gitprovider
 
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/3826-error-msg-wrong-repo.yml
```diff
@@ -0,0 +1,3 @@
+description: An error message is now shown if a wrong repo path is used
+change-type: patch
+destination-branches: [master, iso4, iso5]
```
arnaudsjs, line 3: I think we should mention this change in the `bugfix` section of the changelog.

tests/moduletool/test_repos.py
```diff
@@ -43,6 +46,26 @@ def test_remote_repo_good(git_modules_dir, modules_repo):
...
+    shutil.rmtree(os.path.join(coroot, "test-repository"))
+
+
+def test_remote_repo_good2(git_modules_dir, modules_repo):
+    repo = RemoteRepo("https://github.com/rmccue/{}")
+    coroot = os.path.join(git_modules_dir, "clone_remote_good")
+    result = repo.clone("test-repository", coroot)
+    assert result
+    assert os.path.exists(os.path.join(coroot, "test-repository", "README"))
+    shutil.rmtree(os.path.join(coroot, "test-repository"))
```
arnaudsjs, line 58: It's not a good idea to write content into the directory provided by the `git_modules_dir` fixture. This fixture is a session-scoped fixture. This means that the changes made by you test case will be passed onto the next test case. Adding `shutil.rmtree()` at the end of the test case is not sufficient, as this line will be skipped when the test case raises an exception. This way you get the situation where one failing test case causes several other to fail as well. This makes the debugging of a failing test case a lot more difficult.

In the above-mentioned test cases you only need a temporary directory to perform to clone a Git repository into. In that case it's better to use the `tmpdir` fixture. This is a function scoped fixture that provides an empty temporary directory. This fixture will automatically remove all the content in the tmpdir at the end of each test case. So there is not need to do an `shutil.rmtree()` at the end of the test case.

arnaudsjs (changes requested):

arnaudsjs (approved):

sanderr (approved):

PR #3898, opened 2022-02-23
use generics for typing fixes

sanderr:
# Description

How about this? I'm not familiar with and have not looked into the broader semantics of this code so let me know if something does not make sense. Should I include @andraskvr?

Mypy diff is empty.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/data/paging.py
+++ b/src/inmanta/data/paging.py
@@ -68,11 +68,14 @@ class VersionedQueryIdentifier(QueryIdentifier):
     version: int
 
 
-class PagingCountsProvider(ABC):
+QI = TypeVar("QI", bound=QueryIdentifier)
+
+
+class PagingCountsProvider(ABC, Generic[QI]):
     @abstractmethod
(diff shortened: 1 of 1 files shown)


src/inmanta/data/paging.py
```diff
@@ -114,21 +117,20 @@ async def count_items_for_paging(
...
 
-class ResourceHistoryPagingCountsProvider(PagingCountsProvider):
+class ResourceHistoryPagingCountsProvider(PagingCountsProvider[ResourceQueryIdentifier]):
     def __init__(self, data_class: Type[Resource]) -> None:
         self.data_class = data_class
 
     async def count_items_for_paging(
         self,
-        query_identifier: QueryIdentifier,
+        query_identifier: ResourceQueryIdentifier,
```
bartv, line 126: This is the trick that I missed to make it work.

bartv (approved):

PR #3899, opened 2022-02-24
[FIX] remove failing test 

Hugo-Inmanta:
# Description

Test cast was failing now that multiline is only supported for triple quoted strings

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


sanderr (approved):

changelogs/fix-obsolete-test-case.yml
```diff
@@ -0,0 +1,6 @@
+description: Remove obsolete test case that was failing now that multiline strings are only supported for triple qutoes
+change-type: patch
+destination-branches: [master]
+sections: {
+  bugfix: "{{description}}"
+}
```
sanderr, line 6:
```suggestion
```
I don't think this deserves an entry in the changelog.

sanderr (changes requested):

sanderr (approved):

PR #3909, opened 2022-02-25
Update container usage doc

edvgui:
# Description

Fix naming of containers in doc.

### DIFF
--- a/docs/install/install-server-with-docker/install-server.inc
+++ b/docs/install/install-server-with-docker/install-server.inc
@@ -5,10 +5,10 @@ Use docker pull to get the desired image:
 
 .. code-block:: sh
 
-    docker pull ghcr.io/inmanta/oss:2022
+    docker pull ghcr.io/inmanta/orchestrator:2022
 
 
-This command will pull the latest version of the Inmanta OSS Orchestrator image.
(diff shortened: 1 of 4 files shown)


docs/install/install-server-with-docker/install-server.inc
```diff
@@ -21,7 +21,7 @@ Here is a minimalist docker-compose file content that can be used to deploy the
     version: '3'
     services:
         postgres:
-            container_name: iso_postgres
+            container_name: oss_postgres
```
sanderr, line 24: I might be wrong but I believe this would require a change at line 41 as well.

sanderr (changes requested):
I'm not that familiar with how this would be used. Does it make sense to use something as generic as `oss_server` that doesn't mention "inmanta" at all? (ISO is Inmanta Service Orchestrator while OSS is just Open Source Software).

docs/install/install-server-with-docker/install-server.inc
edvgui: Ah indeed, and the default config of the server as well (which is in the inmanta repo and also requires a rebuild of all the existing container images)

sanderr (approved):

PR #3913, opened 2022-02-25
Issue/3829 unclear error when compiler venv has incompatible python version

### LINKED ISSUE #3829: Unclear error when compiler venv has incompatible Python version
The compiler currently expects the compiler venv to have the same Python version as the active process. I think this is a valid expectation. When this is violated however it does not give a clear error message. Instead a hard to diagnose `FileNotFound` error is raised.

Steps to reproduce (from a 3.9 environment with `inmanta-core>=6` installed):
```console
(inmanta-core-39) sander@bedevere:~/downloads/venvtest$ python3.6 -m venv .env
(inmanta-core-39) sander@bedevere:~/downloads/venvtest$ python -c 'from inmanta.env import VirtualEnv; VirtualEnv(".env").init_env()'
CREATING
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/env.py", line 702, in init_env
    self._write_pth_file()
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/env.py", line 765, in _write_pth_file
    with open(self._path_pth_file, "w", encoding="utf-8") as fd:
FileNotFoundError: [Errno 2] No such file or directory: '.env/lib/python3.9/site-packages/inmanta-inherit-from-parent-venv.pth'
```

I believe a guard should be put in place: the `init_env` code should check if `self.site_packages_dir` exists. If it doesn't, this means the venv hosts a different Python version than the running process, in which case an exception should be raised.

Hugo-Inmanta:
# Description
Raise an exception when the compiler venv has a different python version from the running process.

closes #3829 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
+++ b/changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
@@ -1,7 +1,7 @@
 description: An exception is now raised when there is a mismatch between the python versions of the compiler venv and the active process
 issue-nr: 3829
 change-type: patch
-destination-branches: [master]
+destination-branches: [iso4, iso5, master]
 sections: {
   bugfix: "{{description}}"
 }
--- a/changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
+++ b/changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
@@ -1,4 +1,5 @@
 description: An exception is now raised when there is a mismatch between the python versions of the compiler venv and the active process
+issue-nr: 3829
 change-type: patch
 destination-branches: [master]
 sections: {
--- a/changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
+++ b/changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
@@ -1,4 +1,4 @@
-description: An exception is now raised when there is a mismatch between the python versions of the compiler venv and the active process
+description: An exception is now raised when there is a mismatch between the python version of the compiler venv and the python version of the active process
 issue-nr: 3829
 change-type: patch
 destination-branches: [iso4, iso5, master]
--- a/changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
+++ b/changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
@@ -1,0 +1,6 @@
+description: An exception is now raised when there is a mismatch between the python versions of the compiler venv and the active process
+change-type: patch
+destination-branches: [master]
+sections: {
+  bugfix: "{{description}}"
+}
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -842,8 +842,8 @@ class VenvCreationFailedError(Exception):
         super().__init__(msg)
         self.msg = msg
 
+
 class VenvActivationFailedError(Exception):
     def __init__(self, msg: str) -> None:
         super().__init__(msg)
         self.msg = msg
-
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -706,7 +706,9 @@ class VirtualEnv(ActiveEnv):
                     )
             else:
                 # get version as a (major, minor) tuple for the venv and the running process
-                venv_python_version = subprocess.check_output([self.python_path, "--version"]).decode("utf-8").strip().split()[1]
+                venv_python_version = (
+                    subprocess.check_output([self.python_path, "--version"]).decode("utf-8").strip().split()[1]
+                )
                 venv_python_version = tuple(map(int, venv_python_version.split(".")))[:2]
 
                 running_process_python_version = sys.version_info[:2]
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -699,7 +699,7 @@ class VirtualEnv(ActiveEnv):
             if not os.path.exists(self.site_packages_dir):
                 # the venv hosts a different python version than the running process
                 raise VenvCreationFailedError(
-                    msg=f"Unable to create new virtualenv at {self.env_path} due to python version mismatch"
+                    msg=f"Unable to use virtualenv at {self.env_path} due to python version mismatch"
                 )
 
             # Venv was created using an older version of Inmanta -> Update pip binary and set sitecustomize.py file
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -705,7 +705,7 @@ class VirtualEnv(ActiveEnv):
                     )
             else:
                 python_subprocess = subprocess.Popen([self.python_path, "--version"], stdout=subprocess.PIPE, encoding="utf-8")
-                venv_python_version = tuple(map(int,python_subprocess.stdout.readline().strip().split()[1].split(".")))
+                venv_python_version = tuple(map(int, python_subprocess.stdout.readline().strip().split()[1].split(".")))
                 LOGGER.debug(f"virtual env python version: {venv_python_version}")
                 running_process_python_version = sys.version_info[:3]
                 LOGGER.debug(f"running_process_python_version: {running_process_python_version}")
@@ -714,7 +714,6 @@ class VirtualEnv(ActiveEnv):
                     raise VenvActivationFailedError(
                         msg=f"Unable to use virtualenv at {self.env_path} due to python version mismatch"
                     )
-
 
             # Venv was created using an older version of Inmanta -> Update pip binary and set sitecustomize.py file
             self._write_pip_binary()
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -701,19 +701,19 @@ class VirtualEnv(ActiveEnv):
                 # on UNIX based systems, the python version is in the path to the site packages dir:
                 if not os.path.exists(self.site_packages_dir):
                     raise VenvActivationFailedError(
-                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version"
+                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version "
                         "is different from the Python version of this process."
                     )
             else:
                 # get version as a (major, minor) tuple for the venv and the running process
-                venv_python_version = subprocess.check_output([self.python_path, "--version"]).decode("utf-8").split()[1]
+                venv_python_version = subprocess.check_output([self.python_path, "--version"]).decode("utf-8").strip().split()[1]
                 venv_python_version = tuple(map(int, venv_python_version.split(".")))[:2]
 
                 running_process_python_version = sys.version_info[:2]
 
                 if venv_python_version != running_process_python_version:
                     raise VenvActivationFailedError(
-                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version"
+                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version "
                         "is different from the Python version of this process."
                     )
 
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -675,10 +675,6 @@ class VirtualEnv(ActiveEnv):
         """
         self._parent_python = sys.executable
 
-        if not os.path.exists(self.site_packages_dir):
-            # the venv hosts a different python version than the running process
-            raise VenvCreationFailedError(msg=f"Unable to create new virtualenv at {self.env_path}")
-
         # check if the virtual env exists
         if not os.path.exists(self.python_path):
             # venv requires some care when the .env folder already exists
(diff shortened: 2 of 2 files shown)


changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
```diff
@@ -0,0 +1,6 @@
+description: An exception is now raised when there is a mismatch between the python versions of the compiler venv and the active process
+change-type: patch
+destination-branches: [master]
```
arnaudsjs, line 3: There is something I don't understand here. Currently only ISO4 uses a compiler venv. ISO5+ execute compiles using the process venv. As such I would assume we only need a patch for ISO4. But the ticket talks about `inmanta-core>=6` (which is ISO5) and this PR provides a solution for ISO6. Am I missing something here?

changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
```diff
@@ -0,0 +1,6 @@
+description: An exception is now raised when there is a mismatch between the python versions of the compiler venv and the active process
```
arnaudsjs, line 1: Issue number is missing from the changelog entry

arnaudsjs (changes requested):

changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
sanderr: `inmanta-core>=6` still provides the option to pass another compiler venv (used by pytest-inmanta among others). This ticket is about that. I admit it could have been more clear.

sanderr, line 3:
```suggestion
destination-branches: [master, iso5]
```

arnaudsjs: Conclusion: This change has to go into all lines of development (master, iso5 and iso4).

src/inmanta/env.py
```diff
@@ -696,6 +696,12 @@ def init_env(self) -> None:
                 raise VenvCreationFailedError(msg=f"Unable to create new virtualenv at {self.env_path}")
             LOGGER.debug("Created a new virtualenv at %s", self.env_path)
         elif not os.path.exists(self._path_pth_file):
+            if not os.path.exists(self.site_packages_dir):
+                # the venv hosts a different python version than the running process
+                raise VenvCreationFailedError(
+                    msg=f"Unable to create new virtualenv at {self.env_path} due to python version mismatch"
```
sanderr, line 702:
```suggestion
                    msg=f"Unable to use virtualenv at {self.env_path} due to python version mismatch"
```
It doesn't necessarily get created. If it already exists (which is the case here), we just activate it.

sanderr (changes requested):

changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
Hugo-Inmanta: Should it go to iso4 as well as suggested above by @arnaudsjs ?

sanderr: I think so (see my response there)

changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
```diff
@@ -0,0 +1,7 @@
...
```
arnaudsjs, line 1:
```suggestion
description: An exception is now raised when there is a mismatch between the python version of the compiler venv and the active process
```

src/inmanta/env.py
```diff
@@ -696,6 +696,12 @@ def init_env(self) -> None:
                 raise VenvCreationFailedError(msg=f"Unable to create new virtualenv at {self.env_path}")
             LOGGER.debug("Created a new virtualenv at %s", self.env_path)
         elif not os.path.exists(self._path_pth_file):
+            if not os.path.exists(self.site_packages_dir):
+                # the venv hosts a different python version than the running process
+                raise VenvActivationFailedError(
+                    msg=f"Unable to use virtualenv at {self.env_path} due to python version mismatch"
+                )
```
arnaudsjs, line 703: This check only works for Linux operating systems, because the python version is present in the path to the `site-packages` directory. On windows this check would not work (See: https://github.com/inmanta/inmanta-core/blob/6316212832ee0c295cc98a055f35944d10908268/src/inmanta/env.py#L225), but I didn't check this on a Windows machine. I think we should implement this check in a platform independent way.

arnaudsjs (changes requested):

changelogs/unreleased/3829-unclear-error-when-compiler-venv-has-incompatible-python-version.yml
Hugo-Inmanta: Sounds weird to me without the final "s" :thinking:

arnaudsjs: A certain venv can only have one python version at a certain point in time, never multiple. As such, it doesn't make sense to use plural here, right?

Hugo-Inmanta: As I understand it there is one version for the compiler venv and one version for the active process, so 2 in total hence the plural.

sanderr (changes requested):
Requesting changes so you can ping me via rerequest review when the final changes are in.

src/inmanta/env.py
Hugo-Inmanta: After a discussion with @arnaudsjs we came up with these 2 solutions to check out the python version on a Windows machine:
- Parsing the `pyenv.cfg` file (very fragile solution because no guarantees on the format of the file)
- Forking a new process and executing `<path-to-python-binary> --version` (Slow operation and potential impact on the runtime of the test cases)
@wouterdb  @sanderr what do you think ?

sanderr: How about checking for the fully qualified python binary? I'm not 100% sure but I think it should always be present (e.g. in a Python 3.9 venv there should always be a `bin/python3.9`.

Hugo-Inmanta: Ok so on windows it seems the equivalent of the `ENV/bin` folder is the `ENV\Scripts` folder.
Here is its content:
```bash
Directory of C:\Users\hugo\tmp\test_env_310\Scripts

02/03/2022  10:08    <DIR>          .
02/03/2022  10:08    <DIR>          ..
02/03/2022  10:08             2.086 activate
02/03/2022  10:08             1.015 activate.bat
02/03/2022  10:08            20.571 Activate.ps1
02/03/2022  10:08               393 deactivate.bat
02/03/2022  10:08           106.360 pip.exe
02/03/2022  10:08           106.360 pip3.10.exe
02/03/2022  10:08           106.360 pip3.exe
02/03/2022  10:08           263.056 python.exe
02/03/2022  10:08           251.792 pythonw.exe
               9 File(s)        857.993 bytes
               2 Dir(s)   8.019.701.760 bytes free
```
               
Do you think it's ok/safe to use `pip3.10.exe` to infer the version of python ?

There are also lots of files in `\Lib\site-packages\pip\*\__pycache__` containing the python version in their names, but then again I don't know how robust it is to use them as such.

For example `__init__.cpython-310.pyc` in `C:\Users\hugo\tmp\test_env_310\Lib\site-packages\pip\_internal\cli\__pycache__`

sanderr: Ah, shame. I don't think that would be a safe check because you can create a venv without pip (and I think the compiler might even actually do so). So back to `python --version` I guess. I just had a chat with @wouterdb and he proposed to try the performant check (the one in your current implementation) and if that fails fall back to the `--version` one. This way the tests wouldn't suffer either.

src/inmanta/env.py
```diff
@@ -696,11 +696,24 @@ def init_env(self) -> None:
...
             LOGGER.debug("Created a new virtualenv at %s", self.env_path)
         elif not os.path.exists(self._path_pth_file):
-            if not os.path.exists(self.site_packages_dir):
-                # the venv hosts a different python version than the running process
-                raise VenvActivationFailedError(
-                    msg=f"Unable to use virtualenv at {self.env_path} due to python version mismatch"
-                )
+            # make sure the venv hosts the same python version as the running process
+            if sys.platform != "win32":
+                # on linux based systems, the python version is in the name of the executable:
```
sanderr, line 701:
```suggestion
                # on UNIX based systems, the python version is in the path to the site packages dir:
```

src/inmanta/env.py
```diff
@@ -696,11 +696,24 @@ def init_env(self) -> None:
...
+                if not os.path.exists(self.site_packages_dir):
+                    raise VenvActivationFailedError(
+                        msg=f"Unable to use virtualenv at {self.env_path} due to python version mismatch"
+                    )
+            else:
+                python_subprocess = subprocess.Popen([self.python_path, "--version"], stdout=subprocess.PIPE, encoding="utf-8")
```
sanderr, line 707: Any reason to not use `subprocess.check_output()`? It would make this easier to read imo.

src/inmanta/env.py
```diff
@@ -696,11 +696,24 @@ def init_env(self) -> None:
...
+                venv_python_version = tuple(map(int, python_subprocess.stdout.readline().strip().split()[1].split(".")))
+                LOGGER.debug(f"virtual env python version: {venv_python_version}")
```
sanderr, line 709: I'm not sure about the value these logs add: either they're a match in which case I think it doesn't add anything or it's not a match and we can just incorporate this information in the error message.

sanderr (changes requested):
As a general remark I think we only need to compare up to the minor version. A 3.10.2 process should be compatible with a 3.10.1 venv, but not with a 3.9.2 one.

sanderr (approved):

src/inmanta/env.py
```diff
@@ -696,6 +696,25 @@ def init_env(self) -> None:
...
+            else:
+                # get version as a (major, minor) tuple for the venv and the running process
+                venv_python_version = subprocess.check_output([self.python_path, "--version"]).decode("utf-8").split()[1]
+                venv_python_version = tuple(map(int, venv_python_version.split(".")))[:2]
+
+                running_process_python_version = sys.version_info[:2]
+
+                if venv_python_version != running_process_python_version:
+                    raise VenvActivationFailedError(
+                        msg=f"Unable to use virtualenv at {self.env_path} due to python version mismatch ({'.'.join(map(str,venv_python_version))} vs {'.'.join(map(str,running_process_python_version))})"  # NOQA E501
```
arnaudsjs, line 715: The `# NOQA E501` should be removed here. This means that the line is too long, so we just need to spread the string across multiple lines.

arnaudsjs (changes requested):

src/inmanta/env.py
```diff
@@ -696,6 +696,27 @@ def init_env(self) -> None:
                 raise VenvCreationFailedError(msg=f"Unable to create new virtualenv at {self.env_path}")
             LOGGER.debug("Created a new virtualenv at %s", self.env_path)
         elif not os.path.exists(self._path_pth_file):
+            # make sure the venv hosts the same python version as the running process
+            if sys.platform != "win32":
+                # on UNIX based systems, the python version is in the path to the site packages dir:
+                if not os.path.exists(self.site_packages_dir):
+                    raise VenvActivationFailedError(
+                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version"
```
arnaudsjs, line 704:
```suggestion
                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version "
```

src/inmanta/env.py
```diff
@@ -696,6 +696,27 @@ def init_env(self) -> None:
...
+            else:
+                # get version as a (major, minor) tuple for the venv and the running process
+                venv_python_version = subprocess.check_output([self.python_path, "--version"]).decode("utf-8").split()[1]
+                venv_python_version = tuple(map(int, venv_python_version.split(".")))[:2]
+
+                running_process_python_version = sys.version_info[:2]
+
+                if venv_python_version != running_process_python_version:
+                    raise VenvActivationFailedError(
+                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version"
```
arnaudsjs, line 716:
```suggestion
                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version "
```

src/inmanta/env.py
```diff
@@ -696,6 +696,27 @@ def init_env(self) -> None:
...
+            if sys.platform != "win32":
+                # on UNIX based systems, the python version is in the path to the site packages dir:
+                if not os.path.exists(self.site_packages_dir):
+                    raise VenvActivationFailedError(
+                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version"
+                        "is different from the Python version of this process."
+                    )
+            else:
+                # get version as a (major, minor) tuple for the venv and the running process
+                venv_python_version = subprocess.check_output([self.python_path, "--version"]).decode("utf-8").split()[1]
```
arnaudsjs, line 709: Just to make sure we don't take newline characters into account in the comparison.
```suggestion
                venv_python_version = subprocess.check_output([self.python_path, "--version"]).decode("utf-8").strip().split()[1]
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #3917, opened 2022-02-28
Issue/3851 update classifiers in setup py

### LINKED ISSUE #3851: Update classifiers in setup.py
The classifiers in setup.py still list python 3.5 on master, while 3.8 and 3.9 are not mentioned at all:
https://github.com/inmanta/inmanta-core/blob/b364df85f4e90b5fde983c1721f02205e7090edb/setup.py#L65

Hugo-Inmanta:
# Description

Update python versions in the setup.py classifiers

closes #3851 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3851-update-classifiers-in-setup-py.yml
+++ b/changelogs/unreleased/3851-update-classifiers-in-setup-py.yml
@@ -1,6 +1,6 @@
 change-type: patch
 issue-nr: 3851
 description: Update python versions in the setup.py classifiers
-destination-branches: [master]
-sections:
-    upgrade-note: "{{description}}"
+destination-branches: [master, iso5]
+sections: {}
+
--- a/changelogs/unreleased/3851-update-classifiers-in-setup-py.yml
+++ b/changelogs/unreleased/3851-update-classifiers-in-setup-py.yml
@@ -1,0 +1,6 @@
+change-type: patch
+issue-nr: 3851
+description: Update python versions in the setup.py classifiers
+destination-branches: [master]
+sections:
+    upgrade-note: "{{description}}"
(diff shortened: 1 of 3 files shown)


Hugo-Inmanta:
Should I remove `python 3.5` from the classifiers as well ?

andraskvr:
> Should I remove `python 3.5` from the classifiers as well ?

@Hugo-Inmanta Yes, I think so, if you check line 48, we already require python 3.6 at least

changelogs/unreleased/3851-update-classifiers-in-setup-py.yml
```diff
@@ -0,0 +1,6 @@
+change-type: patch
+issue-nr: 3851
+description: Update python versions in the setup.py classifiers
+destination-branches: [master]
```
edvgui, line 4: Shouldn't this go into iso5 as well?

changelogs/unreleased/3851-update-classifiers-in-setup-py.yml
```diff
@@ -0,0 +1,6 @@
...
+sections:
+    upgrade-note: "{{description}}"
```
edvgui, line 6: Should this really go in the changelog?  (Is this useful information to the user of the package?)

edvgui (changes requested):

changelogs/unreleased/3851-update-classifiers-in-setup-py.yml
Hugo-Inmanta: I guess not, I'll remove it

changelogs/unreleased/3851-update-classifiers-in-setup-py.yml
Hugo-Inmanta: I wonder, is there a way for me to know towards which branches to push the changes ?

edvgui: I guess the documentation must mention somewhere that iso5 supports python 3.9 and iso4 python 3.6, but I could not point you to where that might be, so I can definitely not blame you for not knowing it :wink:

edvgui (approved):

PR #3929, opened 2022-03-01
Improve parser typing

bartv:
# Description

I have been collecting typing improvements. This PR contains those related to the parser. It does a number of this:

1. It removes many unused methods from the stubs so that there are no useless warnings on the stubs anymore
2. It adds typing to those that we use
3. It already contains a number of fixes in the parser/compiler code
4. I added a bit of an ugly hack to correct type what comes out of the parser by making the stubs import Statement

There are quite some warnings left related to str vs locatable string but I am not familiar enough with this part of the code.

Anyone can contribute improvements if they have any but this is not high priority.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


src/inmanta/ast/statements/__init__.py
```diff
@@ -208,9 +208,12 @@ def _add_to_dataflow_graph(self, graph: Optional[DataflowGraph]) -> None:
 
 
 class Literal(ExpressionStatement):
+    lexpos: Optional[int]
```
sanderr, line 211: Why not declare the type in the constructor? I believe this technically declares it as a class attribute as well though in practice it makes little difference of course.

sanderr (approved):

src/inmanta/ast/statements/__init__.py
bartv: no idea. I fixed it.

Adding this is correct, no?

sanderr: Adding what is correct?

bartv: lexpos :)

sanderr: Oh, for some reason I hadn't realized you'd added this, I thought you'd only added the typing. I'll double check.

wouterdb (approved):

src/inmanta/ast/statements/__init__.py
sanderr: Yeah, looks like it is set by the parser. I find it a bit strange but it certainly corresponds with the actual current type.

PR #3936, opened 2022-03-02
Handle resources with agents down in dryrun diff

### LINKED ISSUE #3905: Attach a warning to the current state compare when agents are not started.
When a dryrun is requested, it will only get results for resource of which the agent is actually running. The goal of this ticket is to perform a check if there are any that are not started. If that is the case, the list of agents should be added to the dry run record in the database.

When a dryrun is fetched, the diff status of these resource should be set to 'no agent' and be included in the total count.

andraskvr:
# Description

closes #3905 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


wouterdb (approved):
nice

PR #3939, opened 2022-03-03
[WIP] remove Shift/reduce conflicts

FloLey:
# Description
This Commit removes all the warnings of shift/reduce conflicts by enforcing a shift.
By default if no precedence rules are found, a shift is performed and a warning issued.
By adding the precedence rules to perform a shift when there are conflict the same behavior
should be kept without the warnings. 

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/parser/plyInmantaParser.py
+++ b/src/inmanta/parser/plyInmantaParser.py
@@ -69,7 +69,7 @@ precedence = (
     ("left", "IN"),
     ("left", "RELATION_DEF", "TYPEDEF_INNER", "OPERAND_LIST", "EMPTY"),
     ("left", "CID", "ID"),
-    ("left", "MLS"),
+    ("right", "MLS"),
     ("left", "MLS_END"),
 )
 
--- a/src/inmanta/parser/plyInmantaParser.py
+++ b/src/inmanta/parser/plyInmantaParser.py
@@ -67,7 +67,7 @@ precedence = (
     ("left", "CMP_OP"),
     ("nonassoc", "NOT"),
     ("left", "IN"),
-    ("left", "RELATION_DEF", "TYPEDEF_INNER", "EMPTY"),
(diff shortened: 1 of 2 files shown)


wouterdb (approved):
I have little expertise to add here, this is ages ago for me, if the parser is still working, well done

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -67,8 +67,10 @@
     ("left", "CMP_OP"),
     ("nonassoc", "NOT"),
     ("left", "IN"),
-    ("right", "MLS"),
-    ("right", "MLS_END"),
+    ("left", "RELATION_DEF", "TYPEDEF_INNER", "OPERAND_LIST", "EMPTY"),
+    ("left", "CID", "ID"),
+    ("left", "MLS"),
```
sanderr, line 72:
```suggestion
    ("right", "MLS"),
```
We produce `mls` from end to start, which I believe means right associativity.

sanderr (changes requested):
Very nice! I'm not sure how I feel about this heavy guidance/forcing with the precendence rules. That said, I definitely like the result so I guess we'll just have to wait and see if this turns out to complicate things in the long run. I would keep this out of iso4 though.
I've requested on change related to associativity. I'm not 100% sure about this so let me know if you disagree. I've tried it and it didn't introduce any new conflicts at least. The other associativity rules I'm not sure about either, it feels a bit weird to declare associativity for non-operator type rules, but the Yacc docs don't go into details on conventions / best practices so I don't really know. I'm fine with how you declared them now and we can always change them later if they turn out to produce unintended behavior.

FloLey:
@sanderr could you specially verify my changelog file, it feels empty but It's neither a bugfix nor a new feature. So I don't know what to add.

sanderr (approved):
Change entry looks good to me!

PR #3942, opened 2022-03-03
Fix bug that makes the inmanta deploy command fail

### LINKED ISSUE #3927: Inmanta deploy command fails when the database config options are set to non-default values
**Problem description:**

The `inmanta deploy` command sets up a PostgreSQL server and an Inmanta server to deploy a model. The PostgreSQL server set up by this command expects the Inmanta server to connect to it with the `database.username` config option set to `postgres`. This is the default value for that option. The `inmanta deploy` command fails when any of the inmanta configuration files sets the `database.username` config option to a different value then `postgres`. The same issue exists for the `database.host` config option.

**Solution:**

Force the above-mentioned config options to a specific value in https://github.com/inmanta/inmanta-core/blob/728fe7de6894dabc7c5a218cb1cb6c91b80ecc8f/src/inmanta/deploy.py#L100

timebox to 2 h
**Workaround:**

Add a `.inmanta` file with the following content to the directory where the `inmanta deploy` command is executed:

```
[database]
host=localhost
username=postgres
```

**Additional notes:**

It has to be investigated whether a similar issue exists for other configuration options.

arnaudsjs:
# Description

This PR fixes a bug that makes the `inmanta deploy` command fail when the `database` and `server` sections of the inmanta configuration files contain non-default values.

This PR doesn't resolve all the issues of its kind. The following problems can still occor:

* An Inmanta configuration file enables SSL/TLS by setting the `ssl-cert-file` and `ssl-key-file` options
* The `*_rest_transport` section defines the token config optio

The above-mentioned issues require the capability to unset a config option via a configuration file. That feature is currently not supported.

closes #3927

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3927-ensure-config-isolation-inmanta-deploy-cmd.yml
+++ b/changelogs/unreleased/3927-ensure-config-isolation-inmanta-deploy-cmd.yml
@@ -1,0 +1,7 @@
+---
+description: Fixed bug that makes the `inmanta deploy` command fail when the database and server sections of the inmanta configuration files contain non-default values.
+issue-nr: 3927
+change-type: patch
+destination-branches: [master, iso5, iso4]
+sections:
+  bugfix: "{{description}}"
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -416,8 +416,7 @@ def deploy(options: argparse.Namespace) -> None:
     run = deploy_module.Deploy(options)
     try:
         if not run.setup():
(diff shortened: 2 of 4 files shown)


wouterdb (approved):

sanderr (approved):
I'm wondering if we should create a ticket to have this command completely ignore the system config files?

arnaudsjs:
I created a ticket for the remaining problems:  #3948

PR #3949, opened 2022-03-04
Simplify MLS in compiler

FloLey:
# Description

Previously the Lexer would produces 2 tokens to identify multi line strings. 
First matching on `"""` and create `MLS`, then matching on `.*"""` and produce `MLS_END`. 
The parser did then need to rebuild the multi line string using grammar rules.
whit this commit the MLS token will match the entire multi line string directly 
and removes the burden of using grammar rules. 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/parser/plyInmantaLex.py
+++ b/src/inmanta/parser/plyInmantaLex.py
@@ -131,7 +131,7 @@ def t_JCOMMENT(t: lex.LexToken) -> None:  # noqa: N802
 
 
 def t_MLS(t: lex.LexToken) -> lex.LexToken:
-    r'["]{3}.*["]{3}'
+    r'"{3}([\s\S]*?)"{3}'
     t.value = bytes(t.value[3:-3], "utf-8").decode("unicode_escape")
     lexer = t.lexer
 
--- a/src/inmanta/parser/plyInmantaLex.py
+++ b/src/inmanta/parser/plyInmantaLex.py
@@ -24,8 +24,6 @@ from inmanta.ast.constraint.expression import Regex
 from inmanta.ast.variables import Reference
 from inmanta.parser import ParserException
 
-states = (("mls", "exclusive"),)
-
 keyworldlist = [
     "typedef",
     "as",
@@ -60,7 +58,7 @@ literals = [":", "[", "]", "(", ")", "=", ",", ".", "{", "}", "?", "*"]
 reserved = {k: k.upper() for k in keyworldlist}
 
 # List of token names.   This is always required
-tokens = ["INT", "FLOAT", "ID", "CID", "SEP", "STRING", "MLS", "MLS_END", "CMP_OP", "REGEX", "REL", "PEQ", "RSTRING"] + sorted(
+tokens = ["INT", "FLOAT", "ID", "CID", "SEP", "STRING", "MLS", "CMP_OP", "REGEX", "REL", "PEQ", "RSTRING"] + sorted(
     list(reserved.values())
 )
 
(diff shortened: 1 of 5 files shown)


src/inmanta/parser/plyInmantaLex.py
```diff
@@ -132,46 +130,22 @@ def t_JCOMMENT(t: lex.LexToken) -> None:  # noqa: N802
...
-
-def t_mls_end(t: lex.LexToken) -> lex.LexToken:
-    r'.*["]{3}'
-    t.lexer.begin("INITIAL")
-    t.type = "MLS_END"
-    value = t.value[:-3]
-
+def t_MLS(t: lex.LexToken) -> lex.LexToken:
+    r'"{3}([\s\S]*?)"{3}'
+    t.value = bytes(t.value[3:-3], "utf-8").decode("unicode_escape")
```
sanderr, line 135: Why add unicode escape decoding? This actually changes behavior.

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -132,46 +130,22 @@ def t_JCOMMENT(t: lex.LexToken) -> None:  # noqa: N802
...
-
-
-def t_mls_end(t: lex.LexToken) -> lex.LexToken:
-    r'.*["]{3}'
-    t.lexer.begin("INITIAL")
-    t.type = "MLS_END"
-    value = t.value[:-3]
-
+def t_MLS(t: lex.LexToken) -> lex.LexToken:
+    r'"{3}([\s\S]*?)"{3}'
```
sanderr, line 134: What's the difference between `[\s\S]` and `.`? It looks like you addressed the concern @Hugo-Inmanta raised but could you add a test case that explicitly tests this?

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -132,46 +130,22 @@ def t_JCOMMENT(t: lex.LexToken) -> None:  # noqa: N802
...
+    t.value = bytes(t.value[3:-3], "utf-8").decode("unicode_escape")
     lexer = t.lexer
 
     end = lexer.lexpos - lexer.linestart + 1
     (s, e) = lexer.lexmatch.span()
     start = end - (e - s)
 
     t.value = LocatableString(
-        value, Range(lexer.inmfile, lexer.lineno, start, lexer.lineno, end), lexer.lexpos, lexer.namespace
+        t.value, Range(lexer.inmfile, lexer.lineno, start, lexer.lineno, end), lexer.lexpos, lexer.namespace
```
sanderr, line 143: The lineno isn't accurate anymore.

sanderr (changes requested):

wouterdb (changes requested):
I would like to have a testcase to see if line numbers after a MLS are still correct

src/inmanta/parser/plyInmantaLex.py
FloLey: I did this  for consistency as it was also done like this for strings. I will remove it.

src/inmanta/parser/plyInmantaLex.py
FloLey: apparently `.` doesn't match newlines and `[\s\S]` does. I'll add a test

sanderr: I see, thanks.

sanderr (approved):

wouterdb (approved):

Issue #2793, opened 2021-03-15
implement safer promises for conditionals


sanderr:
# Context
Currently, the compiler has no way of reasoning about statements in a conditional body block until they get emitted. In rare cases this could lead to the compiler freezing relations that should get additional values in a conditional body. It might be worthwhile to have these relations depend a priori on the wrapping condition by having the compiler track a promise from the conditional statement on the relation. These additional promises make the compiler behavior more strict than it is now (so this could be considered a major change) in that some models that were previously valid could now lead to circular promises. However, all models where this would be the case are believed to be bad practice implementations anyway (more about this below).

# Implementation suggestion
Add a `ConditionalPromise(IPromise)` which implements `set_value` to just resolve the promise without setting an actual value. The conditional statement could hold a promise to each relation variable that is a left-hand side in any of its body blocks. As soon as the conditional statement gets executed (a branch is selected), the promise is resolved. At this point new statements are emitted that might add new promises.

This implementation would result in an "overpromise" (safe promise, strict promise, potential promise) semantics. When a relation might get a value from a conditional statement, we reason as if it will get a value from it.

# Breaking change
This safe promising would introduce promise cycles in models where the AST requirements subgraph for a disjunction (an `if-else` for example) is circular. Where the current compiler would allow this as long as the dynamic promises don't conflict (the dynamic requirements form a tree, no loops), these additional static promises would always cause promise loops.

Formally, this applies to any model that has two conditions `condition_a` and `condition_b` with bodies `body_{a,b}{1,2}` for `true` and `false` respectively that have `condition_a` (or a requirement of it) as a LHS in `body_b1` or `body_b2` together with `condition_b` (or a requirement of it) in `body_a1` or `body_a2`.

The theory at this point is that this would affect only these types of models (1) and that all such models model an inherently circular logic that is either invalid or can be simplified to an equivalent, non-circular logic (2). Further investigation of edge cases or a general proof of this statement is required before implementing this ticket.

An example of a (nonsense) model that has this property is the following:
```
entity A:
end

A.others [0:] -- A

implement A using std::none

x = A()
y = A()

if std::count(x.others) > 0:
        y.others += x
end

if std::count(y.others) > 0:
        x.others += y
end
```

# Generalization
This could probably be extended to any statement that has nested blocks, such as implementations (see #3026), for loops, ...

This ticket was created as conclusion of the [this discussion on slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1615803648035900).

PR #3950, opened 2022-03-04
Issue/2793 eager promise tracking

sanderr:
# Description

Implements eager promise tracking as designed in inmanta/inmanta-telco#96.

Part of #2793

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


src/inmanta/ast/constraint/expression.py
```diff
@@ -77,7 +77,6 @@ def __init__(self, attr: Optional[Reference], name: LocatableString) -> None:
     def requires_emit(self, resolver: Resolver, queue: QueueScheduler) -> Dict[object, ResultVariable]:
         # introduce temp variable to contain the eventual result of this stmt
         temp = ResultVariable()
-        temp.set_provider(self)
         # construct waiter
         resumer = IsDefinedReferenceHelper(temp, self.attr, self.name)
```
wouterdb, line 81: shouldn't the `isDefinedReferenceHelper` also put a promise on temp, to replace the `set_provider`

src/inmanta/ast/constraint/expression.py
```diff
@@ -219,8 +218,6 @@ def requires_emit(self, resolver: Resolver, queue: QueueScheduler) -> Dict[objec
         temp: ResultVariable = ResultVariable()
         temp.set_type(Bool())
 
-        temp.set_provider(self)
-
         # wait for the lhs
         HangUnit(queue, resolver, self.children[0].requires_emit(resolver, queue), temp, self)
```
wouterdb, line 222: shouldn't the `HangUnit` also put a promise on temp, to replace the set_provider

src/inmanta/ast/statements/call.py
```diff
@@ -263,7 +263,6 @@ class FunctionUnit(Waiter):
     def __init__(self, queue_scheduler, resolver, result: ResultVariable, requires, function: FunctionCall) -> None:
         Waiter.__init__(self, queue_scheduler)
         self.result = result
-        result.set_provider(self)
         # requires is used to track all required RV's
```
wouterdb, line 266: shouldn't this also become a promisse?

src/inmanta/execute/runtime.py
```diff
@@ -69,15 +69,33 @@ def receive_result(self, value: T, location: Location) -> None:
...
+
+
+class ISetPromise(IPromise, Generic[T]):
+    """
+    A promise to the owner to set a value.
+    """
+
     __slots__ = ()
 
+    # TODO: can recur be removed? If not, document! Might be required only for list promise
```
wouterdb, line 89: recur is for relations, to set the forward relation, then reverse relation and then stop.

src/inmanta/ast/statements/call.py
sanderr: In practice for each of your comments we're dealing with plain `ResultVariable`, which don't do anything with the promises they get (`self.provider` was only used for `OptionVariable`). In addition these variables get passed around to methods that expect a `ResultVariable` rather than an `IPromise`.

wouterdb: Ok, I see, we may have to document this detail (that even if it has the capability to hand out promises, it is not used and not usefull?)

sanderr:
So this is clearly still a draft but I think the proposed structure and PoC should be clear. At this point I'm mainly looking for feedback on this structure and whether a simpler approach is possible.

src/inmanta/ast/statements/__init__.py
```diff
@@ -149,15 +234,103 @@ def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.NodeReference:
...
 class RawResumer(ExpressionStatement):
+    """
+    Resume on a set of requirement variables when they become ready (i.e. they are complete).
+    """
+
     def resume(self, requires: Dict[object, ResultVariable], resolver: Resolver, queue: QueueScheduler) -> None:
         pass
 
 
+class VariableReferenceHook(RawResumer):
```
wouterdb, line 254: This rename is step forward. If you feel like you could name these things better than 'Resumer' and 'Hangunit', feel free to rename, as I don't consider these names to be very good.

src/inmanta/ast/statements/__init__.py
```diff
@@ -149,15 +234,103 @@ def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.NodeReference:
...
+    def __init__(
+        self,
+        instance: Optional["Reference"],
+        name: str,
+        variable_resumer: "VariableResumer",
+    ) -> None:
+        super().__init__()
+        self.instance: Optional["Reference"] = instance
+        self.name: str = name
+        self.variable_resumer: "VariableResumer" = variable_resumer
```
wouterdb, line 269: Why this Resumer in a Resumer structure? 

In my mind, a resumer is a kind of synthetic statement that introduces a temporary result variable to `break` a large expression into sub-expression that can be scheduled independently. The  temporary result variable acts a s a Future that ties the producer and consumer together.  This introduces a different mode of chaining resumers directly?

src/inmanta/ast/variables.py
```diff
@@ -101,111 +101,87 @@ def __repr__(self) -> str:
...
+    def __init__(self, reader: VariableReader) -> None:
+        self.reader: VariableReader = reader
 
-    def get_location(self) -> Location:
-        return self.location
+    def resume(self, requires: Dict[object, ResultVariable], resolver: Resolver, queue_scheduler: QueueScheduler) -> None:
+        self.reader.write_target(requires[self.reader])
 
+    # TODO: does this ever get called? Can we get rid of it?
+    def execute(self, requires: Dict[object, object], resolver: Resolver, queue: QueueScheduler) -> T:
```
wouterdb, line 157: The execute (and in fact the whole resume mechanism) exists to avoid the split between `VariableReader`  and `VariableReadResumer` .

`resume` is merely a second stage execute, avoiding instantiation.

src/inmanta/ast/statements/__init__.py
```diff
@@ -149,15 +234,103 @@ def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.NodeReference:
...
+
+    def schedule(self, resolver: Resolver, queue_scheduler: QueueScheduler) -> None:
```
wouterdb, line 271: Moving scheduling out of the constructor (as is done in the ExecutionUnit side of things) is an improvement.

src/inmanta/execute/runtime.py
```diff
@@ -233,15 +317,37 @@ class DelayedResultVariable(ResultVariable[T]):
         (a queue variable can be dequeued by the scheduler when a provider is added)
     """
 
-    __slots__ = ("queued", "queues", "listeners")
+    __slots__ = ("queued", "queues", "listeners", "promises", "done_promises")
 
     def __init__(self, queue: "QueueScheduler", value: Optional[T] = None) -> None:
         ResultVariable.__init__(self, value)
+        self.promises: List[IPromise] = []
```
wouterdb, line 324: There variables will have to be set to None upon Freeze, to prevent a major memory leak I think.

src/inmanta/ast/statements/assign.py
```diff
@@ -204,9 +249,32 @@ def _add_to_dataflow_graph(self, graph: typing.Optional[DataflowGraph]) -> None:
...
+        if root:
+            return []
+        # TODO: handle explicit namespace references, are they always in scope (think not)? If so, add to parent docstring.
+        #       If not, how to deal with them?
+        # TODO: this check is a temporary hack, clean it up!
+        if "::" in self.instance.name or self.instance.name.split(".", 1)[0] not in in_scope:
+            return []
+
+        promise: ConditionalPromise = ConditionalPromise(provider=self)
+        hook: VariableReferenceHook = VariableReferenceHook(
```
wouterdb, line 264: This has the exact same structure as the first stage of execution.

Would it be more interesting to 'convert' promises when picking them? instead of re-emitting? 
It would also eliminate the problem of in-accurate tracking because of a race between removal of a promise through picking and re-emit of the non-conditional equivalent?

src/inmanta/ast/statements/generator.py
```diff
@@ -243,18 +248,37 @@ def __init__(self, condition: ExpressionStatement, if_branch: BasicBlock, else_b
...
         self.condition.normalize()
         self.if_branch.normalize()
         self.else_branch.normalize()
 
+    def emit_progression_promises(
+        self, resolver: Resolver, queue: QueueScheduler, *, in_scope: FrozenSet[str], root: bool = False
+    ) -> Sequence[ConditionalPromiseABC]:
+        if_block: ConditionalPromiseBlock
+        else_block: ConditionalPromiseBlock
+        if_block, else_block = (
```
wouterdb, line 272: would there be any benefit in contracting cognitional promisses that exist on both side into a non-conditional? I guess not?

src/inmanta/ast/blocks.py
```diff
@@ -80,7 +89,40 @@ def normalize(self) -> None:
...
+
+        Expected to be called after normalization and before emit. May be called multiple times with different scopes.
+
+        :param resolver: Resolver for the parent context promises should be acquired on.
+        :param in_scope: Set of variables that are considered in scope. Promises will only be acquired for these variables or
+            their attributes. This allows to emit promises only relative to a certain parent context, excluding sibling and/or
+            intermediate parent (shadowed) declarations.
+        :param root: If true, this block is considered the root reference for this nested promise emit and promises are
+            acquired on this block's variables (`in_scope` may be empty). If false, this block's variables are considered out
+            of scope in addiiton to the ones declared in the out_of_scope parameter.
```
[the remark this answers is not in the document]
sanderr: Forgot to update this docstring apparently, I changed it to `in_scope`.

src/inmanta/ast/blocks.py
```diff
@@ -80,7 +89,40 @@ def normalize(self) -> None:
...
+        for s in self.__stmts:
+            try:
+                sub_promises.extend(s.emit_progression_promises(resolver, queue, in_scope=in_scope_block, root=root))
+            except RuntimeException as e:
+                e.set_statement(s)
+                raise e
+        return ConditionalPromiseBlock(sub_promises=sub_promises)
+
     def emit(self, resolver: Resolver, queue: QueueScheduler) -> None:
+        self.emit_progression_promises(resolver, queue, root=True)
```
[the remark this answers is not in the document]
sanderr: Related to [your question about the significance of being root](https://github.com/inmanta/inmanta-core/pull/3950#discussion_r833258608). I'll respond here. When a new block is emitted, we want to emit promises for all its conditional subblocks. That's the purpose of this line, and by setting `root=True` we indicate that we don't want conditional promises for statements in this block but only for statements in blocks below it. I don't really like the current approach here but so far I haven't been able to come up with a better one.

src/inmanta/ast/statements/__init__.py
sanderr: If I understand your question correctly you're asking why I refactored this from an ABC class with a couple of subclasses to a concrete resumer class with an ABC `VariableResumer`? The main reason is to make abstraction of whatever is done with the `ResultVariable` object once it's obtained. I found it made more sense to split this conceptually in a waiter for the resultvariable and an action on that resultvariable, than the mix that was currently in place, which involved two execution phases for the same object (first wait for the instance, then cache the variable and continue execution for the variable).

In my mind this composition makes for a clear separation of responsibilities: the one object waits for an instance and fetches its attribute, then delegates to the other object that knows what to do with that variable.

Does my explanation make sense? Do you think I should revert this?

Side note: the `VariableResumer` is a bit different from the other `Resumer` classes (I didn't even make it inherit from them): it doesn't take requires, just a resultvariable. I started out with "action" rather than resumer but I didn't like it as much. Perhaps yet another name could be useful? It's just a class that works on or with a resultvariable in some way.

src/inmanta/ast/variables.py
sanderr: I need to dig into this a bit more, it still feels strange to me and I'm not sure I've handled this correctly at the moment.

src/inmanta/ast/statements/generator.py
sanderr: I think it would be a negligible improvement but I find it hard to say for sure.

src/inmanta/ast/statements/__init__.py
sanderr: Other side note: my idea was that attribute assignment could be refactored at some point to use these same abstractions, because conceptually it just does the same: wait for the instance, fetch the attribute, assign to it.

sanderr:
This PR is ready for review again. I implemented eager promises as designed in inmanta/inmanta-telco#96 and fixed most todo's. I've got a couple of questions still:
1. Do we need to something to the core docs? I personally don't think so because this should be relatively transparent to the end user. For internal purposes I think it should suffice to finalize inmanta/inmanta-telco#96 and move it to the internal docs.
2. The current implementation has some limitations. Two minor ones are mentioned in the design PR (bidirectional relations' rhs and variable shadowing). Another one I only discovered while implementing. (See the two remaining TODO's in the code, I won't expand on the cause here because the TODO notes explain it pretty in-depth iirc). The consequence of this latest limitation is that promises within an implementation block will only be acquired when the block emits. For conditional promises (with `when`) this means that the condition needs to be evaluated before promises within the block can be taken into account. For non-conditional promises, the limitation is less pronounced: promises to global variables' attributes can only be acquired when the constructor is executed, not before. So, three known limitations, of which this latest one is the most impactful in my opinion. Do we consider all three acceptable for now, or do we want to make an attempt to address any of them?

The mypy-diff is given below. There are quite some new type errors, but none of them significant. Most of them stem from the limited typing capability of `ResultVariable`.
```diff
- src/inmanta/ast/__init__.py:286: error: Attribute "visible_namespaces" already defined on line 283  [no-redef]
+ src/inmanta/ast/__init__.py:292: error: Attribute "visible_namespaces" already defined on line 289  [no-redef]
- src/inmanta/execute/runtime.py:306: error: Argument 1 to "Promise" has incompatible type "BaseListVariable"; expected "ListVariable"  [arg-type]
- src/inmanta/execute/runtime.py:772: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/execute/runtime.py:1156: error: "ResultVariable[Any]" has no attribute "promises"  [attr-defined]
+ src/inmanta/ast/statements/__init__.py:40: error: Cannot find implementation or library stub for module named "inmanta.ast.assign"  [import]
+ src/inmanta/ast/statements/__init__.py:176: error: Incompatible types in assignment (expression has type "object", variable has type "Sequence[EagerPromise]")  [assignment]
+ src/inmanta/ast/statements/__init__.py:200: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/__init__.py:209: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/__init__.py:245: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/__init__.py:293: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/__init__.py:363: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/__init__.py:388: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/__init__.py:415: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/assign.py:107: error: Argument 3 to "ExecutionUnit" has incompatible type "ListLiteral"; expected "ResultVariable[object]"  [arg-type]
+ src/inmanta/ast/statements/assign.py:231: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/assign.py:294: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
- src/inmanta/ast/statements/assign.py:314: error: Argument 2 to "__init__" of "AssignStatement" has incompatible type "None"; expected "Reference"  [arg-type]
- src/inmanta/ast/variables.py:58: error: Incompatible return value type (got "Dict[str, Union[Type, ResultVariable[Any]]]", expected "Dict[object, ResultVariable[Any]]")  [return-value]
- src/inmanta/ast/variables.py:61: error: Missing type parameters for generic type "ResultCollector"  [type-arg]
- src/inmanta/ast/variables.py:64: error: Item "Type" of "Union[Type, ResultVariable[Any]]" has no attribute "listener"  [union-attr]
- src/inmanta/ast/variables.py:66: error: Incompatible return value type (got "Dict[str, Union[Type, ResultVariable[Any]]]", expected "Dict[object, ResultVariable[Any]]")  [return-value]
- src/inmanta/ast/variables.py:85: error: Argument 1 to "Reference" has incompatible type "str"; expected "LocatableString"  [arg-type]
+ src/inmanta/ast/variables.py:69: error: Dict entry 0 has incompatible type "str": "Union[Type, ResultVariable[Any]]"; expected "object": "ResultVariable[Any]"  [dict-item]
- src/inmanta/ast/variables.py:116: error: Missing type parameters for generic type "ResultCollector"  [type-arg]
+ src/inmanta/ast/variables.py:76: error: Incompatible types in assignment (expression has type "Union[Type, ResultVariable[Any]]", variable has type "ResultVariable[Any]")  [assignment]
- src/inmanta/ast/variables.py:138: error: Incompatible return value type (got "Union[Type, ResultVariable[Any]]", expected "ResultVariable[Any]")  [return-value]
+ src/inmanta/ast/variables.py:122: error: Argument 1 to "Reference" has incompatible type "str"; expected "LocatableString"  [arg-type]
+ src/inmanta/ast/variables.py:164: error: Incompatible return value type (got "object", expected "T")  [return-value]
+ src/inmanta/ast/variables.py:180: error: Argument 1 to "write_target" of "VariableReader" has incompatible type "ResultVariable[T]"; expected "ResultVariable[object]"  [arg-type]
- src/inmanta/ast/variables.py:174: error: Returning Any from function declared to return "T"  [no-any-return]
- src/inmanta/ast/variables.py:182: error: Returning Any from function declared to return "Location"  [no-any-return]
- src/inmanta/ast/variables.py:264: error: Missing type parameters for generic type "ResultCollector"  [type-arg]
+ src/inmanta/ast/variables.py:269: error: Missing type parameters for generic type "VariableReader"  [type-arg]
- src/inmanta/ast/statements/generator.py:93: error: Argument 2 of "execute" is incompatible with supertype "ExpressionStatement"; supertype defines the argument type as "Resolver"  [override]
+ src/inmanta/ast/statements/generator.py:101: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/generator.py:109: error: Argument 2 of "execute" is incompatible with supertype "RequiresEmitStatement"; supertype defines the argument type as "Resolver"  [override]
- src/inmanta/ast/statements/generator.py:194: error: Need type annotation for "target"  [var-annotated]
- src/inmanta/ast/statements/generator.py:196: error: Argument 5 to "ExecutionUnit" has incompatible type "For"; expected "ExpressionStatement"  [arg-type]
+ src/inmanta/ast/statements/generator.py:213: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/generator.py:382: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/statements/generator.py:533: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
- src/inmanta/ast/statements/generator.py:580: error: Argument 1 to "copy_location" of "Statement" has incompatible type "Instance"; expected "Locatable"  [arg-type]
+ src/inmanta/ast/statements/generator.py:624: error: Argument 1 to "copy_location" of "Locatable" has incompatible type "Instance"; expected "Locatable"  [arg-type]
+ src/inmanta/ast/statements/call.py:94: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
- src/inmanta/ast/statements/call.py:266: error: Argument 1 to "set_provider" of "ResultVariable" has incompatible type "FunctionUnit"; expected "Statement"  [arg-type]
+ src/inmanta/ast/constraint/expression.py:87: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/ast/constraint/expression.py:260: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
+ src/inmanta/compiler/help/explainer.py:142: error: Item "None" of "Optional[Reference]" has no attribute "pretty_print"  [union-attr]
```

src/inmanta/ast/statements/__init__.py
```diff
@@ -149,54 +193,258 @@ def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.NodeReference:
...
+    def schedule(self, resolver: Resolver, queue: QueueScheduler) -> Waiter:
+        """
+        Schedules this instance for execution. Waits for the variable's requirements before resuming.
+        """
+        return RawUnit(
+            queue,
+            resolver,
+            # no need for gradual execution here because this class represents an attribute reference on self.instance,
+            # which is not allowed on multi variables (the only kind of variables that would benefit from gradual execution)
+            self.instance.requires_emit(resolver, queue) if self.instance is not None else {},
```
wouterdb, line 241: If I understand correctly, if the instance is None, the resume method can be called directly here? (there are no requires)?

sanderr: Indeed. My assumption was that the impact of creating the unit anyway would be very limited and I found it simplified the logic (better stick to the semantics of the resume method). But I have no strong attachment to this approach so if you prefer I can call it directly.

src/inmanta/ast/statements/__init__.py
```diff
@@ -149,54 +193,258 @@ def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.NodeReference:
...
+        self._waiter = None
+
+    def fulfill(self) -> None:
+        """
+        If a promise was already acquired, fulfills it, otherwise cancels the waiter so no new promise is acquired when the
+        variable becomes available.
+        """
+        if self._waiter is not None:
+            # waiter is still waiting, remove it
+            self._waiter.queue.remove_from_all(self._waiter)
```
wouterdb, line 381: I wonder if this is a good idea, performance wise. 

it also doesn't 'cancel' the waiter in any way. 
(remove_from_all is iirc pure bookkeeping)

src/inmanta/ast/statements/__init__.py
```diff
@@ -149,54 +193,258 @@ def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.NodeReference:
...
     def requires(self) -> List[str]:
         return [req for v in self.children for req in v.requires()]
 
     def requires_emit(self, resolver: Resolver, queue: QueueScheduler) -> Dict[object, ResultVariable]:
-        return {rk: rv for i in self.children for (rk, rv) in i.requires_emit(resolver, queue).items()}
+        parent_req: Mapping[object, ResultVariable] = super().requires_emit(resolver, queue)
+        own_req: Mapping[object, ResultVariable] = {
+            rk: rv for i in self.children for (rk, rv) in i.requires_emit(resolver, queue).items()
+        }
+        return {**parent_req, **own_req}
```
wouterdb, line 420: This is bang on the critical path, I don't know if this affect performance, but it might.

src/inmanta/ast/statements/generator.py
```diff
@@ -79,13 +83,25 @@ def __init__(self, instance_type: "Entity", implements: "Implement") -> None:
...
+        injected_variables: Set[str] = {"self"}.union(self.type.get_all_attribute_names())
+        self._own_eager_promises = [
+            # implementations live in the namespace's context rather than the constructor's context so for promises that cross
+            # the boundary we translate references so that they are resolved correctly in any context wrapping the constructor.
+            dataclasses.replace(promise, instance=promise.instance.fully_qualified())
+            for implementation in self.implements.implementations
+            for promise in implementation.statements.get_eager_promises()
+            if promise.get_root_variable() not in injected_variables
+        ]
+        # TODO: temporary override because implementation blocks have not normalized yet at this point, meaning the above is
```
wouterdb, line 95: this I don't get.

wouterdb (commented):
1. major question for performance: how efficient is is `{**dict}`

src/inmanta/execute/runtime.py
```diff
@@ -233,15 +306,32 @@ class DelayedResultVariable(ResultVariable[T]):
...
+        self.done_promises: Set[IPromise] = set()
```
wouterdb, line 314: I also don't know the performance impact of Set Vs List.

wouterdb (approved):

wouterdb:
1. I think internal docs is the best idea
2. I would defer the known issue to a next iteration
3. it looks nice

Performance is now the big question

src/inmanta/ast/statements/__init__.py
sanderr: Didn't seem to have a noticable performance impact but I removed it anyway because it makes it simpler to understand.

src/inmanta/ast/statements/__init__.py
sanderr: Seemingly negligible performance impact (still double checking), so I'm leaving this as is unless you have objections.

src/inmanta/ast/statements/generator.py
sanderr: This is about the normalization dependency loop we discussed yesterday. The list comprehension above is accurate if everything below it has already normalized but with the current normalization order it hasn't. Still need to look further into the options we talked about yesterday.

sanderr:
Apart from the performance improvements, 3 small commits since you reviewed: bf225e6, 7c60e68f, 8113eb54

wouterdb (approved):

PR #3953, opened 2022-03-04
Issue/3885 to from int

### LINKED ISSUE #3885: Fix from and to level in const
The loglevel enum has two methods that are wrong. They are written as static methods but defined as instance methods.

It seems that are not used. We should either fix them (add @staticmethod) or delete them:

https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/const.py#L174

remove these methods.

wouterdb:
# Description

removed from_int and from_string 

closes #3885 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/3885_loglevel_from_x.yml
+++ b/changelogs/unreleased/3885_loglevel_from_x.yml
@@ -1,4 +1,4 @@
 description: Remove loglevel from_int and from_string
 issue-nr: 3885
 change-type: patch
-destination-branches: [master, iso5]
+destination-branches: [master, iso5, iso4]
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/3885_loglevel_from_x.yml
```diff
@@ -0,0 +1,4 @@
+description: Remove loglevel from_int and from_string
+issue-nr: 3885
+change-type: patch
+destination-branches: [master, iso5]
```
arnaudsjs, line 4: This issue is also present on `iso4`

arnaudsjs (approved):

changelogs/unreleased/3885_loglevel_from_x.yml
wouterdb: ok! lt's hit it hard!

wouterdb:
```suggestion
destination-branches: [master, iso5, iso4]
```

PR #3959, opened 2022-03-07
Treat expressions as statements

### LINKED ISSUE #3367: Queries can not be used without assigning their value
It seems that using a query without assigning it to a variable raises a `ParserException`, though I would have expected this to be semantically correct.

The model below will raise a parsing error on the last line:
```
import test

test::Test(
    testt="a",
    tests=test::Test(testt="b"),
)

# This is valid
a = test::Test[testt="a"]

# This is not
test::Test[testt="a"]
```
```
inmanta.parser.ParserException: Syntax error at token [ (./main.cf:10:12)
```

Here is a project which allows to reproduce the error: [model.zip](https://github.com/inmanta/inmanta-core/files/7388926/model.zip)

FloLey:
# Description

This commit makes it so that all expressions are now treated as statements.
Some changes where made to MLS as well as it can now occur as a "statement string" or a multi-line comment,
where previously a standalone MLS was always a comment. To be a string it needed to be part of  an assignment.
If the MLS is at the top of a file it will be considered as a comment, otherwise as a Literal.

closes #3367

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1263,8 +1263,6 @@ mls
     )
     assert len(statements) == 1
     mls = statements[0]
-    import pudb
-    pu.db
     assert isinstance(mls, LocatableString)
 
     assert mls.lnr == 2
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1355,17 +1355,6 @@ a=|
         )
 
 
-def test_error_on_relation():
-    with pytest.raises(ParserException) as e:
-        parse_code(
-            """
-Host.provider [1] -- Provider test"""
-        )
-    assert e.value.location.file == "test"
-    assert e.value.location.lnr == 3
-    assert e.value.location.start_char == 2
-
-
 def test_doc_string_on_new_relation():
     statements = parse_code(
         """
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1874,17 +1874,6 @@ def test_1766_empty_model_multiple_newline():
     assert len(statements) == 0
 
 
-def test_1707_out_of_place_regex():
-    with pytest.raises(ParserException) as pytest_e:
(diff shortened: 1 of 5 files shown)


tests/test_parser.py
```diff
@@ -1875,17 +1859,6 @@ def test_1766_empty_model_multiple_newline():
     assert len(statements) == 0
 
 
-def test_1707_out_of_place_regex():
```
FloLey, line 1878: This is allowed now. As a regex is now a statement.

tests/test_parser.py
```diff
@@ -1361,17 +1356,6 @@ def test_bad_2():
         )
 
 
-def test_error_on_relation():
```
FloLey, line 1364: As everything can be written on a single line, this is now allowed: "test" is a reference to a variable.
I think it would be safer to enforce 1 statement per line:
```
"""
Host.provider [1] -- Provider
test
"""
```

FloLey:
I feel like I am missing some tests but I don't now which.

tests/test_parser.py
sanderr: While (as I said before) I do agree in a sense, it's not that simple. We still want to support
```
MyEntity(
    x=1
)
```
I don't think it's feasible to implement something like this now.

wouterdb (approved):

sanderr (approved):

sanderr:
About the tests, I don't immediately see what else you could add. This is a pretty big change to the parser but we can't go and verify that all possible parse paths are tested. I think just have to have confidence that they all have tests already. If not, I'm sure the advanced testing we've got planned for the end of the sprint will reveal any issues.

sanderr:
(in which case we can improve the testing)

PR #3965, opened 2022-03-07
Improve error message regarding index attribute missing in constructor

### LINKED ISSUE #3902: Improve compiler error message
When an attribute or relation that is part of the constructor the following message is given for example:

```
attributes police are part of an index and should be set in the constructor. (reported in Construct(pm::Rate) /module/model/mpls/ciscoxr.cf:77))
```

However this message could be improved by doing the following:
1. In this case it is a relation but the message says attributes.
2. It only implicitly mentions the type
3. The import rewrites the name so it does not point directly to the type or the index

A potential improved message could be:

**Invalid Constructor call: missing relation 'police'. The relation ciscoxr_services::policy_map::Rate.police is part of an index.**

For an attribute it should say **attribute**

arnaudsjs:
# Description

* Improve the compiler error message that is given when an index attribute is missing in the constructor call.
* Fix the typing of the `Entity.get_attribute()` method.

closes #3902 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -399,7 +399,7 @@ class IndexAttributeMissingInConstructorException(TypingException):
             attribute_kind = "relation" if isinstance(attribute, RelationAttribute) else "attribute"
             exc_message += (
                 f"\n\t* Missing {attribute_kind} '{attribute.name}'. "
-                f"The relation {entity.get_full_name()}.{attribute.name} is part of an index."
+                f"The {attribute_kind} {entity.get_full_name()}.{attribute.name} is part of an index."
             )
         return exc_message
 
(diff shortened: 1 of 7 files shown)


wouterdb (approved):

src/inmanta/ast/statements/generator.py
```diff
@@ -378,6 +380,30 @@ def resume(self, requires: Dict[object, ResultVariable], resolver: Resolver, que
...
+
+    def _get_error_message(self, entity: "Entity", unset_attributes: List[str]) -> str:
+        exc_message = "Invalid Constructor call:"
+        for attribute_name in unset_attributes:
+            attribute: Optional[Attribute] = entity.get_attribute(attribute_name)
+            assert attribute is not None  # Make mypy happy
+            attribute_kind = "relation" if isinstance(attribute, RelationAttribute) else "attribute"
+            exc_message += (
+                f"\n\t* Missing {attribute_kind} '{attribute.name}'. "
+                f"The relation {entity.get_full_name()}.{attribute.name} is part of an index."
```
sanderr, line 402:
```suggestion
                f"The {attribute_kind} {entity.get_full_name()}.{attribute.name} is part of an index."
```

sanderr (changes requested):

sanderr (approved):

PR #3971, opened 2022-03-08
Add documentation regarding the relation precedence policy

### LINKED ISSUE #3945: Add end-user documentation regarding type precedence hints
Add documentation for type precedence hints added in #3098 

- Add this to the `troubleshooting` section
- Mention clearly that this is an experimental feature

arnaudsjs:
# Description

Add documentation regarding the relation precedence policy feature to the troubleshooting documentation

closes #3945  

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3945-docs-relation-precedence-policy.yml
+++ b/changelogs/unreleased/3945-docs-relation-precedence-policy.yml
@@ -3,5 +3,3 @@ description: Add documentation regarding the relation precedence policy feature 
 issue-nr: 3945
 change-type: minor
 destination-branches: [master, iso5]
-sections:
-  feature: "{{description}}"
--- a/changelogs/unreleased/3945-docs-relation-precedence-policy.yml
+++ b/changelogs/unreleased/3945-docs-relation-precedence-policy.yml
@@ -1,0 +1,7 @@
+---
+description: Add documentation regarding the relation precedence policy feature to the troubleshooting documentation
+issue-nr: 3945
+change-type: minor
+destination-branches: [master, iso5]
+sections:
+  feature: "{{description}}"
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -325,12 +325,12 @@ Example: Consider the following ``project.yml`` file.
     repo: https://github.com/inmanta/
     description: A quickstart project that installs a drupal website.
     relation_precedence_policy:
-      - "A::Submod_A.relation before B::Submod_B.other_relation"
+      - "A::EntityA.relation before B::EntityB.other_relation"
 
 
 The last two lines of this file define the relation precedence policy of the project. The policy contains one rule
-saying that the relationship ``relation`` of entity ``A::Submod_A`` should be considered completely populated before
-the relation ``other_relation`` of entity ``B::Submod_B`` can be considered complete.
+saying that the relationship ``relation`` of entity ``A::EntityA`` should be considered completely populated before
+the relation ``other_relation`` of entity ``B::EntityB`` can be considered complete.
 
 Each rule in a relation precedence policy should have the following syntax:
 
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -307,7 +307,9 @@ Relationship precedence policy
 
 .. warning::
 
-    The relationship precedence policy feature is currently experimental
+    The inmanta compiler is very good at determining in which order it should evaluate the orchestration model. Unfortunately in very complex models it might not be able to do this. In that case you can give the compiler some instruction by providing it with relationship precedence rules.
+
+    This is a very powerful tool because you can override all the intelligence in the compiler. This means that if you provide the correct rule it will fix the compilation. If you provide a wrong rule it can make this even worse. However, it can never make the orchestrator compile incorrect results.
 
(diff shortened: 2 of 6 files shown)


wouterdb (commented):
Perhaps also make the explainer link to the documentation?

changelogs/unreleased/3945-docs-relation-precedence-policy.yml
```diff
@@ -0,0 +1,7 @@
+---
+description: Add documentation regarding the relation precedence policy feature to the troubleshooting documentation
+issue-nr: 3945
+change-type: minor
+destination-branches: [master, iso5]
+sections:
+  feature: "{{description}}"
```
sanderr, line 7: Does this require a change entry? The feature itself was just added so the documentation would be assumed to be part of that.

docs/troubleshooting.rst
```diff
@@ -277,6 +277,127 @@ the log level of the command to the DEBUG level and the ``-X`` option shows stac
 
    $ inmanta -vvv export -X
 
+
+Compilation fails
+=================
+
+In rare situations, the compiler might fail with a ``List modified after freeze`` or an
+``Optional variable accessed that has no value`` error, even thought the model is syntactically correct. The following
```
sanderr, line 285:
```suggestion
``Optional variable accessed that has no value`` error, even though the model is syntactically correct. The following
```

docs/troubleshooting.rst
```diff
@@ -277,6 +277,127 @@ the log level of the command to the DEBUG level and the ``-X`` option shows stac
...
+.. code-block:: text
+    :linenos:
+
+    name: quickstart
+    modulepath: libs
+    downloadpath: libs
+    repo: https://github.com/inmanta/
+    description: A quickstart project that installs a drupal website.
+    relation_precedence_policy:
+      - "A::Submod_A.relation before B::Submod_B.other_relation"
```
sanderr, line 328: These names do not seem to follow our conventions (I'm not even sure if they're correct syntax), which I would try to avoid in our docs. Also, the names `Submod_{A,B}` seem to indicate they're modules while they're actually entities.

docs/troubleshooting.rst
```diff
@@ -277,6 +277,127 @@ the log level of the command to the DEBUG level and the ``-X`` option shows stac
...
+  complete too soon.
+
+The following sections provide information on how this issue can be resolved.
+
+Relationship precedence policy
+------------------------------
+
+.. warning::
+
+    The relationship precedence policy feature is currently experimental
```
sanderr, line 310: @wouterdb do we still consider it bad practice to rely on this feature as anything other than a temporary workaround? Is it always better and feasible to solve the issue by removing ambiguity in the model?

sanderr (changes requested):

src/inmanta/compiler/help/templates/modified_after_freeze.j2
```diff
@@ -20,9 +20,10 @@ This can mean one of two things:
 The procedure to solve this is the following:
 
 1. Ensure the model is correct by checking that the problematic assignment at {{location | underline}} is not conditional on the value it assigns.
-2. Report a bug to the inmanta issue tracker at https://github.com/inmanta/inmanta/issues or directly contact inmanta. This is a priority issue to us, so you will be helped rapidly and by reporting the problem, we can fix it properly.
-3. {%if reverse %}[applies]{%else%}[does not apply here]{%endif%} If the exception is on the reverse relation, try to give a hint by explicitly using the problematic relation{%if reverse_example%}: {{reverse_example | underline}}{%endif%}.
-4. Simplify the model by relying less on `is defined` but use a boolean instead.
+2. Set a relation precedence policy on the Inmanta project (See: https://docs.inmanta.com/community/latest/troubleshooting.html#compilation-fails).
```
sanderr, line 23: @wouterdb same question here. Up till now I understood that this was more something we could offer as an emergency resolution on a support request. Do we want the user to take this step themselves?

sanderr (commented):
Still unsure about the intended usage (see comments), looks good otherwise.

docs/troubleshooting.rst
wouterdb: I consider it a last resort for now.

src/inmanta/compiler/help/templates/modified_after_freeze.j2
wouterdb: I think so, it may make things easier for us. 

I think this is what @bartv  also preferred

wouterdb (approved):

sanderr (approved):

docs/troubleshooting.rst
bartv: I would indeed give a warning that this can easily make things worse. Something like:

```
The inmanta compiler is very good at determining in which order it should evaluate the orchestration model. Unfortunately in very complex models it might not be able to do this. In that case you can give the compiler some instruction by providing it with relationship precedence rules.

This is a very powerful tool because you can override all the intelligence in the compiler. This means that if you provide the correct rule it will fix the compilation. If you provide a wrong rule it can make this even worse. However, it can never make the orchestrator compile incorrect results.
```

bartv (approved):

PR #3973, opened 2022-03-08
Issue/unclear error when compiler venv has incompatible python version iso4

### LINKED ISSUE #3829: Unclear error when compiler venv has incompatible Python version

Hugo-Inmanta:
# Description

Specific PR to handle the case for ISO4

Closes #3829 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -92,7 +92,8 @@ class VirtualEnv(object):
         if os.path.isdir(self.env_path) and os.listdir(self.env_path):
             # make sure the venv hosts the same python version as the running process
             if sys.platform.startswith("linux"):
-                # On linux distributions we can check the versions match because the env's version is in the site-packages dir's path
+                # On linux distributions we can check the versions match because
+                # the env's version is in the site-packages dir's path
 
                 if not os.path.exists(self.site_packages):
                     raise VenvActivationFailedError(


wouterdb (approved):

src/inmanta/env.py
```diff
@@ -355,3 +370,15 @@ def _get_installed_packages(cls, python_interpreter: str) -> Dict[str, str]:
             LOGGER.debug("%s: %s", cmd, output.decode())
 
         return {r["name"]: r["version"] for r in json.loads(output.decode())}
+
+
+class VenvCreationFailedError(Exception):
+    def __init__(self, msg: str) -> None:
+        super().__init__(msg)
+        self.msg = msg
```
arnaudsjs, line 378: This exception is not used anywhere? Was this change also merged on master.

src/inmanta/env.py
```diff
@@ -101,6 +101,21 @@ def _init_env(self) -> bool:
                 return False
             LOGGER.debug("Created a new virtualenv at %s", self.env_path)
 
+        else:
```
arnaudsjs, line 104: These changes do not correspond with the changes in https://github.com/inmanta/inmanta-core/commit/4e889fe930c811dc5951d3cce5d020dcb45ce6f4. Why?

src/inmanta/env.py
```diff
@@ -101,6 +101,21 @@ def _init_env(self) -> bool:
...
+            # make sure the venv hosts the same python version as the running process
+            if sys.platform != "win32":
+                # get version as a (major, minor) tuple for the venv and the running process
+                venv_python_version = subprocess.check_output([python_bin, "--version"]).decode("utf-8").strip().split()[1]
```
arnaudsjs, line 108: In https://github.com/inmanta/inmanta-core/commit/4e889fe930c811dc5951d3cce5d020dcb45ce6f4 we only invoke `python --version` on windows systems. Here we do it the other way around and no check is done on Linux systems. Why?

arnaudsjs (changes requested):

src/inmanta/env.py
Hugo-Inmanta: That's exactly what my assumption was :+1:

src/inmanta/env.py
Hugo-Inmanta: There is no `_path_pth_file` logic in iso4, should I have added it ? I don't truly understand what is going on here, I think I need some help to finish this ticket.

src/inmanta/env.py
```diff
@@ -101,6 +101,36 @@ def _init_env(self) -> bool:
                 return False
             LOGGER.debug("Created a new virtualenv at %s", self.env_path)
 
+        else:
+            # make sure the venv hosts the same python version as the running process
+
+            if sys.platform != "win32":
+                # On UNIX distribs we can check the versions match because the env's version is in the site-packages dir's path
+                binpath = os.path.abspath(os.path.join(self.env_path, "bin"))
+                base = os.path.dirname(binpath)
```
arnaudsjs, line 110: These two lines seem to be redundant. I think we can replace it with:

```
base = os.path.abspath(self.env_path)
```

src/inmanta/env.py
```diff
@@ -101,6 +101,36 @@ def _init_env(self) -> bool:
                 return False
             LOGGER.debug("Created a new virtualenv at %s", self.env_path)
 
+        else:
+            # make sure the venv hosts the same python version as the running process
+
+            if sys.platform != "win32":
```
arnaudsjs, line 107: I don't think this is correct. We should perform the `python --version` check on any platform that is not Linux and the faster check on Linux systems. This implementation does it the other way around. It performs the fast check on any platform that is not a windows, but we don't know whether the fast check will work on any non-windows system.

I think the same bug exists on master and iso5, so we will have to patch it there as well.

src/inmanta/env.py
```diff
@@ -101,6 +101,36 @@ def _init_env(self) -> bool:
...
+                # On UNIX distribs we can check the versions match because the env's version is in the site-packages dir's path
+                binpath = os.path.abspath(os.path.join(self.env_path, "bin"))
+                base = os.path.dirname(binpath)
+                site_packages = os.path.join(
+                    base, "lib", "python%s" % ".".join(str(digit) for digit in sys.version_info[:2]), "site-packages"
+                )
```
arnaudsjs, line 113: I think it would make sense to set a `self.site_packages` variable in the constructor. Now we are calculating this value in two different location of the same class.

arnaudsjs (changes requested):

src/inmanta/env.py
Hugo-Inmanta: Created a PR here: https://github.com/inmanta/inmanta-core/pull/3992 for master and iso5

arnaudsjs (approved):

src/inmanta/env.py
```diff
@@ -61,6 +61,22 @@ def __init__(self, env_path: str) -> None:
         self._parent_python: Optional[str] = None
         self._packages_installed_in_parent_env: Optional[Dict[str, str]] = None
 
+        python_name = os.path.basename(sys.executable)
+
+        if sys.platform == "win32":
+            self.binpath = os.path.abspath(os.path.join(self.env_path, "Scripts"))
+            self.base = os.path.dirname(self.binpath)
+            self.site_packages = os.path.join(self.base, "Lib", "site-packages")
+            self.python_bin = os.path.join(self.env_path, "Scripts", python_name)
```
sanderr, line 70: These can even be moved outside of the conditions by using `self.binpath`.

src/inmanta/env.py
```diff
@@ -73,22 +89,34 @@ def _init_env(self) -> bool:
...
-
         # check if the virtual env exists
-        if sys.platform == "win32":
-            python_bin = os.path.join(self.env_path, "Scripts", python_name)
-        else:
-            python_bin = os.path.join(self.env_path, "bin", python_name)
+        if os.path.isdir(self.env_path) and os.listdir(self.env_path):
+            # make sure the venv hosts the same python version as the running process
+            if sys.platform.startswith("linux"):
+                # On linux distribs we can check the versions match because the env's version is in the site-packages dir's path
```
sanderr, line 96: "distribs", is this a typo or a French abbreviation? Just curious, the meaning is clear.

sanderr (approved):

src/inmanta/env.py
Hugo-Inmanta: Oh yes, I guess that's French :D I'll change to `distributions`

PR #3976, opened 2022-03-09
Fix bug that makes every inmanta warning end with an empty line

### LINKED ISSUE #3951: newlines in warnings
Some warnings produce a newline after the warning as seen during the demo (for exemple the HyphenDeprecationWarning)

arnaudsjs:
# Description

Fix bug that makes every inmanta warning end with an empty line.

closes #3951 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3951-remove-newline-in-warnings.yml
+++ b/changelogs/unreleased/3951-remove-newline-in-warnings.yml
@@ -1,0 +1,7 @@
+---
+description: Fix bug that makes every inmanta warning end with an empty line.
+issue-nr: 3951
+change-type: patch
+destination-branches: [master, iso5, iso4]
+sections:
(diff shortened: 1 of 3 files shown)


sanderr (changes requested):
I think `if file is not None` we need to add the newline. Otherwise we'll just concatenate all warnings in a row.

arnaudsjs:
> I think `if file is not None` we need to add the newline. Otherwise we'll just concatenate all warnings in a row.

Nice catch! When I look at this more closely, I don't see a code path where `file is not None` would evaluate to True. And if it would evaluate to True. Shouldn't we raise an exception when `file is not None` evaluates to True? I don't think it's a good idea to have dead code in our code base.

PR #3992, opened 2022-03-11
[FIX] explicitely check for linux distributions instead of 'non-windows'

Hugo-Inmanta:
# Description

 Explicitely check for linux distributions instead of 'non-windows'

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/fix-platform-check.yml
+++ b/changelogs/unreleased/fix-platform-check.yml
@@ -1,0 +1,5 @@
+---
+description: explicitely check for linux distribution instead of checking for "non-windows"
+change-type: patch
+destination-branches : [master, iso5]
+sections: {}
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -696,16 +696,17 @@ class VirtualEnv(ActiveEnv):
                 raise VenvCreationFailedError(msg=f"Unable to create new virtualenv at {self.env_path}")
             LOGGER.debug("Created a new virtualenv at %s", self.env_path)
         elif not os.path.exists(self._path_pth_file):
-            # make sure the venv hosts the same python version as the running process
-            if sys.platform != "win32":
-                # on UNIX based systems, the python version is in the path to the site packages dir:
+            # Make sure the venv hosts the same python version as the running process
+            if sys.platform.startswith("linux"):
+                # On linux based systems, the python version is in the path to the site packages dir:
                 if not os.path.exists(self.site_packages_dir):
                     raise VenvActivationFailedError(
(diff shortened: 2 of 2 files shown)


changelogs/unreleased/fix-platform-check.yml
```diff
@@ -0,0 +1,5 @@
+---
+description: explicitely check for linux distribution instead of checking for "non-windows"
```
arnaudsjs, line 2:
```suggestion
description: explicitly check for linux distribution instead of checking for "non-windows"
```

arnaudsjs (changes requested):

arnaudsjs (approved):

sanderr (commented):
Why is this required? I was going to comment the opposite on the other PR but I assume there's a reason for this?

src/inmanta/env.py
```diff
@@ -719,6 +701,24 @@ def init_env(self) -> None:
                         "is different from the Python version of this process."
                     )
 
+            raise VenvCreationFailedError(
+                msg=f"Unable to create new virtualenv at {self.env_path} (environment already exists)"
+            )
```
sanderr, line 706: I don't think this is right. If the environment already exists we should just continue.

src/inmanta/env.py
```diff
@@ -676,36 +676,18 @@ def init_env(self) -> None:
...
-            except Exception:
-                raise VenvCreationFailedError(msg=f"Unable to create new virtualenv at {self.env_path}")
-            LOGGER.debug("Created a new virtualenv at %s", self.env_path)
-        elif not os.path.exists(self._path_pth_file):
-            # make sure the venv hosts the same python version as the running process
-            if sys.platform != "win32":
-                # on UNIX based systems, the python version is in the path to the site packages dir:
+        if os.path.isdir(self.env_path) and os.listdir(self.env_path):
+            # Make sure the venv hosts the same python version as the running process
+            if sys.platform.startswith("linux"):
```
sanderr, line 681: Could you explain why this needed to be turned around?

sanderr (changes requested):

src/inmanta/env.py
Hugo-Inmanta: I just noticed this mistake

src/inmanta/env.py
Hugo-Inmanta: This was in response to this comment : https://github.com/inmanta/inmanta-core/pull/3973#discussion_r824479544

arnaudsjs (approved):
I have not new remarks next to the ones mentioned my Sander.

sanderr (approved):

PR #4008, opened 2022-03-16
Create a notification when an exporting compile fails

### LINKED ISSUE #3983: Create notification for failed exporting compile
When a compile fails that exports resources, a notification should be created by the compilation service.

This notification should say:
- Title: Compilation failed
- Message: An exporting compile has failed
- Severity: Error
- URI: a link to the compile report

andraskvr:
# Description

closes #3983 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -45,10 +45,9 @@ from inmanta.protocol import encode_token, methods, methods_v2
 from inmanta.protocol.common import ReturnValue
 from inmanta.protocol.exceptions import BadRequest, NotFound
 from inmanta.protocol.return_value_meta import ReturnValueWithMeta
-from inmanta.server import SLICE_COMPILER, SLICE_DATABASE, SLICE_NOTIFICATION, SLICE_TRANSPORT
+from inmanta.server import SLICE_COMPILER, SLICE_DATABASE, SLICE_TRANSPORT
 from inmanta.server import config as opt
 from inmanta.server.protocol import ServerSlice
-from inmanta.server.services.notificationservice import NotificationService
 from inmanta.server.validate_filter import CompileReportFilterValidator, InvalidFilter
 from inmanta.types import Apireturn, ArgumentTypes, JsonType, Warnings
 from inmanta.util import ensure_directory_exist
@@ -470,7 +469,6 @@ class CompilerService(ServerSlice):
     """
 
     _env_folder: str
-    _notification_service: NotificationService
 
     def __init__(self) -> None:
         super(CompilerService, self).__init__(SLICE_COMPILER)
@@ -485,7 +483,7 @@ class CompilerService(ServerSlice):
         self.listeners.append(listener)
 
     def get_dependencies(self) -> List[str]:
-        return [SLICE_DATABASE, SLICE_NOTIFICATION]
+        return [SLICE_DATABASE]
 
     def get_depended_by(self) -> List[str]:
         return [SLICE_TRANSPORT]
@@ -495,7 +493,6 @@ class CompilerService(ServerSlice):
         state_dir: str = opt.state_dir.get()
         server_state_dir = ensure_directory_exist(state_dir, "server")
         self._env_folder = ensure_directory_exist(server_state_dir, "environments")
-        self._notification_service = cast(NotificationService, server.get_slice(SLICE_NOTIFICATION))
 
     async def start(self) -> None:
         await super(CompilerService, self).start()
@@ -600,14 +597,6 @@ class CompilerService(ServerSlice):
         await asyncio.gather(*[notify(listener) for listener in self.listeners])
         await compile.update_fields(handled=True)
         LOGGER.log(const.LOG_LEVEL_TRACE, "listeners notified for %s", compile.id)
-        if compile.success is False and compile.do_export:
-            await self._notification_service.notify(
-                compile.environment,
-                title="Compilation failed",
-                message="An exporting compile has failed",
-                severity=const.NotificationSeverity.error,
-                uri=f"/api/v2/compilereport/{compile.id}",
-            )
 
     async def _recover(self) -> None:
         """Restart runs after server restart"""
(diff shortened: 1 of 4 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -597,6 +600,14 @@ async def notify(listener: CompileStateListener) -> None:
         await asyncio.gather(*[notify(listener) for listener in self.listeners])
         await compile.update_fields(handled=True)
         LOGGER.log(const.LOG_LEVEL_TRACE, "listeners notified for %s", compile.id)
+        if compile.success is False and compile.do_export:
```
arnaudsjs, line 603:
```suggestion
        if not compile.success and compile.do_export:
```

tests/server/test_compilerservice.py
```diff
@@ -51,6 +52,10 @@
 @pytest.fixture
 async def compilerservice(server_config, init_dataclasses_and_load_schema):
     server = Server()
+    notification_service = NotificationService()
+    await notification_service.prestart(server)
+    await notification_service.start()
+    server.add_slice(notification_service)
```
arnaudsjs, line 58: Shouldn't we also stop the notification_service?

tests/server/test_compilerservice.py
```diff
@@ -184,6 +189,10 @@ def get_runner(self, remote_id: uuid.UUID) -> HangRunner:
 
     # manual setup of server
     server = Server()
+    notification_service = NotificationService()
+    await notification_service.prestart(server)
+    await notification_service.start()
```
arnaudsjs, line 194: Same here.

tests/server/test_compilerservice.py
```diff
@@ -1013,3 +1022,34 @@ async def test_compileservice_api(client, environment):
...
+    compilerservice = server.get_slice(SLICE_COMPILER)
+    env = await data.Environment.get_by_id(uuid.UUID(environment))
+
+    result = await client.list_notifications(env.id)
+    assert result.code == 200
+    assert len(result.result["data"]) == 0
+
+    compile_id, _ = await compilerservice.request_recompile(env, False, True, uuid.uuid4())
+
+    assert await compilerservice.is_compiling(env.id) == 200
```
arnaudsjs, line 1037: This line introduces a race condition when the compile is fast. I would remove it.

tests/server/test_compilerservice.py
```diff
@@ -1013,3 +1022,34 @@ async def test_compileservice_api(client, environment):
...
+
+    async def compile_done():
```
arnaudsjs, line 1039:
```suggestion
    async def compile_done() -> bool:
```

tests/server/test_compilerservice.py
```diff
@@ -1013,3 +1022,34 @@ async def test_compileservice_api(client, environment):
...
+        res = await compilerservice.is_compiling(env.id)
+        return res == 204
+
+    await retry_limited(compile_done, 10)
+
+    async def notification_logged():
```
arnaudsjs, line 1045:
```suggestion
    async def notification_logged() -> bool:
```

tests/server/test_compilerservice.py
```diff
@@ -1013,3 +1022,34 @@ async def test_compileservice_api(client, environment):
...
+async def test_notification_on_failed_exporting_compile(server, client, environment: str) -> None:
+    compilerservice = server.get_slice(SLICE_COMPILER)
+    env = await data.Environment.get_by_id(uuid.UUID(environment))
+
+    result = await client.list_notifications(env.id)
+    assert result.code == 200
+    assert len(result.result["data"]) == 0
+
+    compile_id, _ = await compilerservice.request_recompile(env, False, True, uuid.uuid4())
+
```
arnaudsjs, line 1036: Improve readability:

```suggestion
    compile_id, _ = await compilerservice.request_recompile(env, force_update=False, do_export=True, remote_id=uuid.uuid4())

```

tests/server/test_compilerservice.py
```diff
@@ -1013,3 +1022,34 @@ async def test_compileservice_api(client, environment):
...
+    assert await compilerservice.is_compiling(env.id) == 200
+
+    async def compile_done():
+        res = await compilerservice.is_compiling(env.id)
+        return res == 204
+
+    await retry_limited(compile_done, 10)
```
arnaudsjs, line 1043:
```suggestion
    await retry_limited(compile_done, timeout=10)
```

tests/server/test_compilerservice.py
```diff
@@ -1013,3 +1022,34 @@ async def test_compileservice_api(client, environment):
...
+
+    async def notification_logged():
+        result = await client.list_notifications(env.id)
+        assert result.code == 200
+        return len(result.result["data"]) > 0
+
+    await retry_limited(notification_logged, 10)
```
arnaudsjs, line 1050:
```suggestion
    await retry_limited(notification_logged, timeout=10)
```

src/inmanta/server/services/compilerservice.py
arnaudsjs, line 603:
```suggestion
        if not compile.success and compile.do_export:
```

src/inmanta/server/services/compilerservice.py
```diff
@@ -597,6 +600,14 @@ async def notify(listener: CompileStateListener) -> None:
...
+            await self._notification_service.notify(
```
arnaudsjs, line 604: Why did we add this code in the `_notify_listeners()` method? It breaks a bit the cohesion of the method in the sense that this call has nothing to do with listeners. But maybe there is a good reason for this. Is it because this method is always invoked, even when the server restarts?

arnaudsjs (changes requested):

src/inmanta/server/services/compilerservice.py
arnaudsjs: If we did it for the above-mentioned reason, we should execute `await compile.update_fields(handled=True)` after sending the notification.

wouterdb: I agree that this is not the right place, but that instead it should be a listener that this slice registers to itself or the notification service register to this slice.

wouterdb (changes requested):

wouterdb (approved):

arnaudsjs (approved):

PR #4015, opened 2022-03-17
Add notification cleanup

### LINKED ISSUE #3982: Add notification cleanup
The goal of this ticket is to create a background task that regularly cleans notifications from the database.

- [ ] Add an environment settings for the amount of days to keep notifications. By default set it to 365 days
- [ ] Add a cleanup task to the delete all messages that are older than the number in the settings. Run the task every hour and run it at startup. The task has to fetch the new config every time.

andraskvr:
# Description

closes #3982 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -1784,6 +1784,7 @@ SERVER_COMPILE = "server_compile"
 RESOURCE_ACTION_LOGS_RETENTION = "resource_action_logs_retention"
 PURGE_ON_DELETE = "purge_on_delete"
 PROTECTED_ENVIRONMENT = "protected_environment"
+NOTIFICATION_RETENTION = "notification_retention"
 
 
 class Setting(object):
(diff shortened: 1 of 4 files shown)


wouterdb (approved):

src/inmanta/data/__init__.py
```diff
@@ -5522,6 +5530,16 @@ async def list_notifications(
         ]
         return dtos
 
+    @classmethod
+    async def clean_up_notifications(cls) -> None:
+        environments = await Environment.get_list()
+        for env in environments:
+            time_to_retain_logs = await env.get(NOTIFICATION_RETENTION)
+            keep_notifications_until = datetime.datetime.now().astimezone() - datetime.timedelta(days=time_to_retain_logs)
+            LOGGER.debug(f"Cleaning up notifications in environment {env.name} that are older than {keep_notifications_until}")
```
arnaudsjs, line 5539: The job to cleanup old compile reports also logs with info level. To keep things consistent, we should do the same here.
```suggestion
            LOGGER.info(f"Cleaning up notifications in environment %s that are older than %s", env.name, keep_notifications_until)
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #4024, opened 2022-03-18
Fix order of stages in compile report details

andraskvr:
# Description

closes inmanta/web-console#3082

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/compile-report-stage-order.yml
+++ b/changelogs/unreleased/compile-report-stage-order.yml
@@ -3,3 +3,5 @@ issue-nr: 3082
 issue-repo: web-console
 change-type: patch
 destination-branches: [master, iso5, iso4]
+sections:
+    bugfix: "{{description}}"
--- a/changelogs/unreleased/compile-report-stage-order.yml
+++ b/changelogs/unreleased/compile-report-stage-order.yml
@@ -1,0 +1,5 @@
+description: Fix order of stages in compile report details
+issue-nr: 3082
+issue-repo: web-console
+change-type: patch
+destination-branches: [master, iso5, iso4]
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -3322,7 +3322,7 @@ class Compile(BaseDocument):
                     {cls.table_name()} comp
                     INNER JOIN compiledetails cd ON cd.substitute_compile_id = comp.id
                     LEFT JOIN public.report rep on comp.id = rep.compile
-        ) SELECT * FROM compiledetails;
+        ) SELECT * FROM compiledetails ORDER BY report_started ASC;
         """
         values = [cls._get_value(environment), cls._get_value(id)]
         result = await cls.select_query(query, values, no_obj=True)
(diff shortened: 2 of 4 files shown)


wouterdb (approved):

changelogs/unreleased/compile-report-stage-order.yml
```diff
@@ -0,0 +1,5 @@
+description: Fix order of stages in compile report details
```
arnaudsjs, line 1: Shouldn't we add this to the `bugfix` section?

tests/server/test_compile_details.py
```diff
@@ -86,6 +85,10 @@ async def env_with_compiles(client, environment):
     return environment, ids, compile_requested_timestamps
 
 
+def parse_datetime_to_utc(time: str) -> datetime.datetime:
+    return datetime.datetime.strptime(time, "%Y-%m-%dT%H:%M:%S.%f").replace(tzinfo=datetime.timezone.utc)
```
arnaudsjs, line 89: This is a handy method that will be useful in other places in the future. Let's anticipate on that and put it into the util package.

src/inmanta/data/__init__.py
```diff
@@ -3322,7 +3322,7 @@ async def get_compile_details(cls, environment: uuid.UUID, id: uuid.UUID) -> Opt
                     {cls.table_name()} comp
                     INNER JOIN compiledetails cd ON cd.substitute_compile_id = comp.id
                     LEFT JOIN public.report rep on comp.id = rep.compile
-        ) SELECT * FROM compiledetails;
+        ) SELECT * FROM compiledetails ORDER BY report_started ASC;
```
arnaudsjs, line 3325: Did you do run an `EXPLAIN ANALYZE` on this query? This sort operation might have a rather big performance impact.

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
andraskvr: After your question I did :)
The query only deals with one specific compile (or a few, if there were substitutions), so it has a quite limited scope. The sorting doesn't seem to affect the performance much:
Without sorting: https://explain.dalibo.com/plan/no6#plan
With sorting: https://explain.dalibo.com/plan/poA

arnaudsjs: Indeed. The query plan indicates that the cost of the sorting is high. But in practice the number of records returns by the query should be very small.

arnaudsjs: Does the performance improve if we add an index on `Report(compile, report_started ASC)`?

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
andraskvr: It does not, and I think it's because we are sorting the records of the CTE. 
(For building the CTE, the `report_compile_index` is used).
On the other hand, the number of records we're dealing with here should be small indeed.

PR #4033, opened 2022-03-21
Fix race condition in test case test_compile_details

arnaudsjs:
# Description

This PR fixes a race condition in the `test_compile_details()` test case. The background task that removes old compile reports was removing the compile reports created by the `env_with_compiles` fixture, causing the test case to fail with the following exception:

```
_____________________________ test_compile_details _____________________________

server = <inmanta.server.protocol.Server object at 0x7f8b8c0eed60>
client = <inmanta.protocol.endpoints.Client object at 0x7f8b726bda00>
env_with_compiles = ('5565bfc6-df9d-4517-914c-5b352d6b4565', [UUID('f36a91ab-3df5-42a7-942b-2f1b78a1e716'), UUID('73a916f6-8d60-4f88-ba52-... 0), datetime.datetime(2021, 9, 9, 11, 1), datetime.datetime(2021, 9, 9, 11, 2), datetime.datetime(2021, 9, 9, 11, 3)])

    async def test_compile_details(server, client, env_with_compiles):
        environment, ids, compile_requested_timestamps = env_with_compiles
    
        # A compile that has no substitute_compile_id, and has reports
        result = await client.compile_details(environment, ids[0])
>       assert result.code == 200
E       assert 404 == 200
E         +404
E         -200
```

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/fix-race-condition-in-test-compile-details.yml
+++ b/changelogs/unreleased/fix-race-condition-in-test-compile-details.yml
@@ -1,0 +1,4 @@
+---
+description: Fix race condition in the `test_compile_details()` test case. The background task that removes old compile reports was removing the compile reports created by the `env_with_compiles` fixture.
+change-type: patch
+destination-branches: [master, iso5]
--- a/changelogs/unreleased/fix-race-condition-in-test-compile-details.yml
+++ b/changelogs/unreleased/fix-race-condition-in-test-compile-details.yml
@@ -1,4 +1,4 @@
 ---
 description: Fix race condition in the `test_compile_details()` test case. The background task that removes old compile reports was removing the compile reports created by the `env_with_compiles` fixture.
 change-type: patch
-destination-branches: [master, iso5]
+destination-branches: [master, iso5, iso4]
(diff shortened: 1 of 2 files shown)


sanderr (approved):

arnaudsjs:
> I don't understand the change and how it fixes the race condition, could you explain?

The fixture `env_with_compiles` used to create Compile records in the database with a timestamp in `2021-09-09`. By default, the server runs a background task that regularly cleans up old compiles and their attached compile reports (See: https://github.com/inmanta/inmanta-core/blob/8c5328a440e4576d9e0fc591573b32184f81e987/src/inmanta/server/services/compilerservice.py#L502). The default retention time is seven days and the tests do not modify this setting. As such, there exists a race where the cleanup services deletes the Compile reports generated by `env_with_compiles` in the middle of the test case, causing the test case to fail with the above-mentioned exception.

This fix makes sure that the timestamp of the compile records are never older than 7 says. As such they cannot be cleaned up by the cleanup service. I will also add a comment in the fixture that explains this.

FloLey (approved):
Thanks! now I understand :D

PR #4060, opened 2022-03-25
[FIX] set up dependabot for iso5

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/inmanta-service-orchestrator/issues/274
# Self Check:

- [x] Changelog entry
~~- [ ] No (preventable) type errors~~

# Reviewer Checklist:


sanderr:
I would be careful with "partially closes". I think GitHub searches for "closes <issue-rference>" when you close a PR and closes any issues it finds like that. I don't think it's smart enough to interpret the "partially". I usually go for "part of <issue-reference>" in cases like this.

.github/dependabot.yml
```diff
@@ -1,11 +1,26 @@
...
+  open-pull-requests-limit: 10
+  allow:
+    # Allow both direct and indirect updates for all packages
+    - dependency-type: "all"
+  ignore:
+    - dependency-name: "inmanta-core"
+    - dependency-name: "inmanta-license"
+    - dependency-name: "inmanta-lsm"
+    - dependency-name: "inmanta-support"
+    - dependency-name: "inmanta-ui"
```
sanderr, line 26: This one is my bad. I only realized now that we actually want to minimize the repos where we apply this on because the difference between iso5 and master means that we would need to diverge branches when we do a major release. As such, we only want to apply this on repos that actually depend on other components. This one doesn't (all other repos I listed do).
```suggestion
```
So I would keep just the change to set up dependabot for iso5, and perhaps update the title and change entry description to match it.

Sorry once more for the miscommunication.

sanderr (changes requested):

.github/dependabot.yml
Hugo-Inmanta: No worries :ok_hand:

sanderr (approved):

PR #4102, opened 2022-04-05
[FEAT] Add environment setting to set the number of stored versions a…

### LINKED ISSUE #3505: Increase the number of stored versions in the database
Currently we only keep the 10 latest versions. We should increase this number.

- Make this an environment setting, defaulting to 
- deprecate the option in the config file 

Open questions:

- time based
- performance of e.g. resource view

Hugo-Inmanta:
…nd deprecate old config option

# Description
- Add an environment setting to set the number of stored version
- Deprecate the option in the config file
 
closes #3505 
# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3505-increase-the-number-of-stored-versions-in-the-database.yml
+++ b/changelogs/unreleased/3505-increase-the-number-of-stored-versions-in-the-database.yml
@@ -4,5 +4,5 @@ change-type: patch
 destination-branches: [iso4, iso5, master]
 sections: {
   feature: "{{description}}",
-  deprecation-note: "The server_version_to_keep option in the server configuration file is now deprecated in favor of the environment setting"
+  deprecation-note: "The available-versions-to-keep option in the server configuration file is now deprecated in favor of the environment setting"
 }
--- a/changelogs/unreleased/3505-increase-the-number-of-stored-versions-in-the-database.yml
+++ b/changelogs/unreleased/3505-increase-the-number-of-stored-versions-in-the-database.yml
@@ -1,7 +1,7 @@
 description: Add environment setting to set the number of stored versions.
 issue-nr: 3505
 change-type: patch
-destination-branches: [iso4, iso5, master]
+destination-branches: [iso5, master]
 sections: {
   feature: "{{description}}",
   deprecation-note: "The available-versions-to-keep option in the server configuration file is now deprecated in favor of the environment setting"
--- a/changelogs/unreleased/3505-increase-the-number-of-stored-versions-in-the-database.yml
+++ b/changelogs/unreleased/3505-increase-the-number-of-stored-versions-in-the-database.yml
@@ -1,0 +1,8 @@
+description: Add environment setting to set the number of stored versions.
+issue-nr: 3505
+change-type: patch
+destination-branches: [iso4, iso5, master]
+sections: {
+  feature: "{{description}}",
+  deprecation-note: "The server_version_to_keep option in the server configuration file is now deprecated in favor of the environment setting"
+}
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -181,6 +181,46 @@ async def test_autostart_batched(client, server, environment):
     assert len(sessionendpoint._sessions) == 1
 
 
+async def test_environment_setting_n_versions(client, server):
+    """
+    This test creates two environments with different values for the environment
+    setting AVAILABLE_VERSIONS_TO_KEEP and checks that the number of versions is
+    in line with the desired setting
+    """
+
+    env_1_expected_value = 30
+    env_2_expected_value = 100
+
+    # Create project
+    result = await client.create_project("env-test")
+    assert result.code == 200, result.result
+    project_id = result.result["project"]["id"]
+
+    # Create environment 1
+    # AVAILABLE_VERSIONS_TO_KEEP is set to 30
+    result = await client.create_environment(project_id=project_id, name="env_1")
+    env_1_id = result.result["environment"]["id"]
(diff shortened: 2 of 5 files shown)


changelogs/unreleased/3505-increase-the-number-of-stored-versions-in-the-database.yml
```diff
@@ -0,0 +1,8 @@
+description: Add environment setting to set the number of stored versions.
+issue-nr: 3505
+change-type: patch
+destination-branches: [iso4, iso5, master]
+sections: {
+  feature: "{{description}}",
+  deprecation-note: "The server_version_to_keep option in the server configuration file is now deprecated in favor of the environment setting"
```
andraskvr, line 7: `available-versions-to-keep` option?

andraskvr (changes requested):
I think it would be good to add a test case for this. 
For example with two environments that have different values for this setting, add some empty versions, and check the number of versions after cleanup to assert that the different settings were respected

wouterdb (changes requested):

tests/test_server.py
```diff
@@ -181,6 +181,46 @@ async def test_autostart_batched(client, server, environment):
...
 
 
+async def test_environment_setting_n_versions(client, server):
+    """
+    This test creates two environments with different values for the environment
+    setting AVAILABLE_VERSIONS_TO_KEEP and checks that the number of versions is
+    in line with the desired setting
+    """
+
+    env_1_expected_value = 30
```
wouterdb, line 191: I would use lower number here, to make the testcase faster.

Hugo-Inmanta: It's quite long currently indeed (~7s)
I will override the default value of 100 as well for `env_2` after checking that it is set correctly

tests/test_server.py
```diff
@@ -181,6 +181,46 @@ async def test_autostart_batched(client, server, environment):
...
+    # Check values
+    result = await client.get_setting(tid=env_1_id, id=data.AVAILABLE_VERSIONS_TO_KEEP)
+    assert result.code == 200
+    assert result.result["value"] == env_1_expected_value
+
+    result = await client.get_setting(tid=env_2_id, id=data.AVAILABLE_VERSIONS_TO_KEEP)
+    assert result.code == 200
+    assert result.result["value"] == env_2_expected_value
+
+    await check_n_versions_kept(client, server, project_id, env_1_id, max(env_1_expected_value, env_2_expected_value))
```
wouterdb, line 220: Does this even work?  The cleanup is run periodically, so how do you know it has already run?

and 2 -> you don't create more than expected, so this is kind of trivial (i.e. the test will pass, even if the cleanup never happens)

tests/test_server.py
```diff
@@ -202,8 +246,12 @@ async def test_version_removal(client, server):
         assert res.code == 200
         result = await client.get_project(id=project_id)
 
+        result = await client.get_setting(tid=env_id, id=data.AVAILABLE_VERSIONS_TO_KEEP)
```
wouterdb, line 249: I would pass this in as a parameter to the method. 

Now, if this method always returns a 100, the test would still succeed. 
(i.e. the assertion depends on what you are testing and it can be easily avoided)

wouterdb (changes requested):

tests/test_server.py
wouterdb, line 191: Why factor out these values into constants/variables if changing them will just break the testcase (i.e. the values are constants here, hardcoded elsewhere)

tests/test_server.py
```diff
@@ -202,8 +246,12 @@ async def test_version_removal(client, server):
...
+        assert result.code == 200
+        n_versions_to_keep = result.result["value"]
+
         versions = await client.list_versions(tid=env_id)
-        assert versions.result["count"] <= opt.server_version_to_keep.get() + 1
+        assert versions.result["count"] <= n_versions_to_keep + 1
```
Hugo-Inmanta, line 254:
```suggestion
        assert versions.result["count"] == n_versions_to_keep
```
Make it so we check that we have exactly the expected number of versions

tests/test_server.py
```diff
@@ -181,6 +181,54 @@ async def test_autostart_batched(client, server, environment):
...
+    """
+
+    # Create project
+    result = await client.create_project("env-test")
+    assert result.code == 200
+    project_id = result.result["project"]["id"]
+
+    # Create environment
+    result = await client.create_environment(project_id=project_id, name="env_1")
+    env_1_id = result.result["environment"]["id"]
```
andraskvr, line 206: I think it would be useful to check the different setting values with multiple environments in the same test case, because the global config option changed to an environment specific setting. If it was still a global option the test could still pass.

andraskvr (changes requested):

tests/test_server.py
Hugo-Inmanta: I see ! Good point

tests/test_server.py
```diff
@@ -181,6 +181,54 @@ async def test_autostart_batched(client, server, environment):
...
+    versions = await client.list_versions(tid=env_1_id)
+    assert versions.result["count"] == n_versions_to_create
+
+    await server.get_slice(SLICE_ORCHESTRATION)._purge_versions()
+
+    versions = await client.list_versions(tid=env_1_id)
+    assert versions.result["count"] == min(n_versions_to_keep, n_versions_to_create)
+
+
 async def test_version_removal(client, server):
```
wouterdb, line 232: I think this whole testcase is superseded by the new one and can be removed

wouterdb (changes requested):

wouterdb (approved):

andraskvr (approved):

PR #4106, opened 2022-04-06
Performance fix for yang modules

wouterdb:
# Description

- Improve performance of duplicate checking
- remove dead code 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/improve-typing-performance.yml
+++ b/changelogs/unreleased/improve-typing-performance.yml
@@ -1,0 +1,5 @@
+description: Improve performance of duplicate detection in compiler typing
+change-type: patch
+destination-branches:
+  - iso5
+  - master
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -49,8 +49,8 @@ if TYPE_CHECKING:
     from inmanta.ast.statements.define import DefineAttribute, DefineImport  # noqa: F401
     from inmanta.execute.runtime import ExecutionContext, ResultVariable  # noqa: F401
 
-
 import inmanta.ast.attribute
+
 
 class EntityLike(NamedType):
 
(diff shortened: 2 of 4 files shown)


src/inmanta/ast/entity.py
```diff
@@ -274,39 +288,6 @@ def get_attribute(self, name: str) -> Optional["Attribute"]:
                     return attr
         return None
 
-    @memoize
-    def __get_related(self) -> "Set[Entity]":
-        # down
```
sanderr, line 279: In the new check we don't seem to go down anymore. Then again, I see no reason at all why we would want to do so. It was a conscious decision to drop this?

sanderr (changes requested):

src/inmanta/ast/entity.py
wouterdb: The reason we had to go up and down is that the tree is still being built up.

If the tree is complete, and we check for every relation, we only have to check up. Because then the lower one will find the higher one. 

If the tree is incomplete, we have to check both up and down, because if we are the upper one and we have been added last, we can not just check up. 

(Or am I wrong?)

sanderr: I see. I think you're right.

sanderr (approved):
Depending on performance we could additionally add a cache on `get_attribute`, but we then need to be careful of the functools cache memory leak issue.

changelogs/unreleased/improve-typing-performance.yml
```diff
@@ -0,0 +1,5 @@
+description: Improve performance of duplicate detection in compiler typing
+change-type: patch
```
edvgui, line 2: If this is a patch this should also go into iso4 no?

wouterdb: It is an improvement, so I would not make the effort to put it in iso4.

iso4 should only receive the things that it really needs.

edvgui: Then this should be made a minor change? My point was that mergetool will (I think) reject the PR as it is now

PR #4109, opened 2022-04-07
First attempt at venv adoption

wouterdb:
# Description

First draft of adopting the current venv.

For reviewers:
1. do I activate in the right place? Now I also do this if the server starts. What is a better place?
2. do I need to add a testcase, if so, how will it work?

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


sanderr (commented):
Core concept looks good to me (as in it looks in line with the current proposal on slack). I think we might get some undefined behavior when the venv inmanta is installed in is the active one though. We might want to check for `VIRTUAL_ENV ~= sys.executable`.

Haven't checked your questions in detail yet.

PR #4114, opened 2022-04-08
Type cast docs

sanderr:
# Description

Added documentation for primitive type casts to the language reference

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -136,6 +136,10 @@ The basic primitive types are ``string``, ``number``, ``int`` or ``bool``. These
     assert = bool(1.2) == true
     assert = bool(0) == false
     assert = bool(null) == false
+    assert = bool("x") == true
+    # like in Python, only empty strings are considered false
+    assert = bool("false") == true
+    assert = bool("") == false
     assert = string(true) == "true"
 
 Constrained primitive types can be derived from the basic primitive type with a typedef statement.
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -135,6 +135,7 @@ The basic primitive types are ``string``, ``number``, ``int`` or ``bool``. These
     assert = number(true) == 1
     assert = bool(1.2) == true
     assert = bool(0) == false
+    assert = bool(null) == false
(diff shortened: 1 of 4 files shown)


docs/language.rst
```diff
@@ -125,7 +125,17 @@ Literal values can be assigned to variables
 Primitive types
 ==============================
 
-The basic primitive types are ``string``, ``number``, ``int`` or ``bool``.
+The basic primitive types are ``string``, ``number``, ``int`` or ``bool``. These basic types also support type casts:
+
+.. code-block:: inmanta
+
+    assert = true
```
sanderr, line 132: Is this too meta for the docs? What is most readable?
- assignments without assertion: `x = int("1")`?
- type cast expressions without assignment: `int("1")?
- something else?

edvgui: It is probably a bit hard to get the meaning of this assert statement here for newbies, but I think it doesn't degrade the utility of the code snippet, which is mainly (I think) showing of the syntax.  (I actually never did it like that in the past, I like it a lot)
I would keep it like that

docs/language.rst
```diff
@@ -125,7 +125,17 @@ Literal values can be assigned to variables
...
+    assert = int("1") == 1
+    assert = number("1.2") == 1.2
+    assert = number(true) == 1
+    assert = bool(1.2) == true
+    assert = bool(0) == false
```
edvgui, line 137: How about bool conversion of a string? I would like `bool("false") == false` but if it works like python this won't be the case

edvgui (commented):
Very efficient

wouterdb (approved):

docs/language.rst
edvgui: (I just tested it and indeed it does work like in python)

sanderr: It indeed just delegates casting to the corresponding Python types.

edvgui (approved):

wouterdb (approved):

Issue #4130, opened 2022-04-12
v1tov2 does not specify packages sufficiently strict


sanderr:
The v1 to v2 module conversion (used for `inmanta module v1tov2` but also for `inmanta module build/install` on a v1 module) uses `find_namespace:` but does not specify which namespace packages are in scope. As a result, the resulting wheel includes any directories with some python files beneath them as namespace packages. This impacts the on-commit v2 module builds. For example the v2 lsm module ships the `docs` and `tests` packages.

A potential solution would be for core to include the following in the generated `setup.cfg`:
```
[options.packages.find]
include = inmanta_plugins
```

This should also go into the cookiecutter

PR #4147, opened 2022-04-19
Issue/4130 v1tov2 does not specify packages sufficiently strict

Hugo-Inmanta:
# Description

Explicitely limit namespace packages to `inmanta_plugins` 

Part of  #4130 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/4130-v1tov2-does-not-specify-packages-sufficiently-strict.yml
+++ b/changelogs/unreleased/4130-v1tov2-does-not-specify-packages-sufficiently-strict.yml
@@ -1,6 +1,7 @@
-description: Limit included namespace packages to inmanta_plugins
+description: Limit included namespace packages to inmanta_plugins for v1tov2 module conversion.
 issue-nr: 4130
 change-type: patch
 destination-branches: [iso5, master]
 sections: {
+    bugfix: "{{description}}"
 }
--- a/changelogs/unreleased/4130-v1tov2-does-not-specify-packages-sufficiently-strict.yml
+++ b/changelogs/unreleased/4130-v1tov2-does-not-specify-packages-sufficiently-strict.yml
@@ -1,0 +1,6 @@
+description: Limit included namespace packages to inmanta_plugins
+issue-nr: 4130
+change-type: patch
+destination-branches: [iso5, master]
+sections: {
+}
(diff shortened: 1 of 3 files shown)


src/inmanta/moduletool.py
```diff
@@ -1132,5 +1132,6 @@ def get_setup_cfg(self, config_in: Optional[configparser.ConfigParser] = None) -
         config["options"]["zip_safe"] = "False"
         config["options"]["include_package_data"] = "True"
         config["options"]["packages"] = "find_namespace:"
+        config["options.packages.find"]["include"] = "inmanta_plugins"
```
sanderr, line 1135: You're still missing a `config.add_section("options.packages.find")`, see line 1121.

changelogs/unreleased/4130-v1tov2-does-not-specify-packages-sufficiently-strict.yml
```diff
@@ -0,0 +1,6 @@
+description: Limit included namespace packages to inmanta_plugins
+issue-nr: 4130
+change-type: patch
+destination-branches: [iso5, master]
+sections: {
```
sanderr, line 5:
```suggestion
sections: {
  bugfix: "{{description}}"
```

This is a potentially user visible issue so I would include a bugfix note.

changelogs/unreleased/4130-v1tov2-does-not-specify-packages-sufficiently-strict.yml
```diff
@@ -0,0 +1,6 @@
+description: Limit included namespace packages to inmanta_plugins
```
sanderr, line 1:
```suggestion
description: Limit included namespace packages to inmanta_plugins for v1tov2 module conversion.
```
Expanded this a bit because it will be included in the user-visible changelog.

sanderr (changes requested):

src/inmanta/moduletool.py
arnaudsjs, line 1135: Otherwise submodules are not packaged.
```suggestion
        config["options.packages.find"]["include"] = "inmanta_plugins*"
```

arnaudsjs (changes requested):

src/inmanta/moduletool.py
arnaudsjs: I would also add a test to verify that submodules are packaged as well. If this test doesn't exist yet.

sanderr: Have you tested this? I seem to recall I tried this first and it didn't work (but I could be wrong). I haven't tested submodules so it's very much possible that my suggestion doesn't work there.

arnaudsjs: Yes, I did.

arnaudsjs (approved):

sanderr (approved):

PR #4172, opened 2022-04-27
[FIX] explicitly use bash instead of sh

### LINKED ISSUE #3832: make mypy-diff errors
I have trouble running `make mypy-diff` as /`usr/bin/sh` points to `dash` and not to `bash`.
I currently run it by changing `sh` in the script by `bash`.

Hugo-Inmanta:
# Description

Explicitly use `bash` instead of relying on `sh` symlinking to it and therefore causing an error for shells where the `[[` keyword is not supported

closes #3832 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3832-make-mypy-diff-error-double-bracket-not-found.yml
+++ b/changelogs/unreleased/3832-make-mypy-diff-error-double-bracket-not-found.yml
@@ -1,6 +1,6 @@
 description: Explicitly use bash instead of assuming sh symlinking to it to
     prevent errors for shells where the [[ keyword is not supported
 change-type: patch
-destination-branches: [master, iso5]
+destination-branches: [master, iso4, iso5]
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/3832-make-mypy-diff-error-double-bracket-not-found.yml
```diff
@@ -0,0 +1,6 @@
+description: Explicitly use bash instead of assuming sh symlinking to it to
+    prevent errors for shells where the [[ keyword is not supported
+change-type: patch
+destination-branches: [master, iso5]
```
sanderr, line 4: I think this can safely go into iso4 as well.
```suggestion
destination-branches: [master, iso5, iso4]
```

sanderr (approved):

PR #4179, opened 2022-04-28
[FEAT] move experimental feature to permanent and set default caching…

### LINKED ISSUE #3838: Enable caching of CF files
In https://github.com/inmanta/inmanta-core/pull/2580/files we introduced experimental support for caching of compiled CF files.

- [x] For ISO6 and OSS, we should make this a default feature, that is on by default, but can be disabled by an option. 
- [x] Verify if this works with modules V2 for both editable and non-editable install:
    - if it causes crashes, resolve the crash
    - if it doesn't crash and doesn't cache, make a ticket to resolve this.  (we should base the solution on what python does in this case)

Hugo-Inmanta:
… to true

# Description
- Set default caching of compiled files to true
- Add the `--no-cache` option to disable caching

closes #3838 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -366,8 +366,8 @@ def test_startup_failure(tmpdir, postgres_db, database_name):
     ) in stdout
     assert code == 4
 
-
-def test_compiler_exception_output(snippetcompiler):
+@pytest.mark.parametrize("cache_cf_files", [True, False])
+def test_compiler_exception_output(snippetcompiler, cache_cf_files):
     snippetcompiler.setup_for_snippet(
         """
 entity Test:
@@ -379,12 +379,13 @@ implement Test using std::none
 o = Test(attr="1234")
         """
     )
+    cwd = snippetcompiler.project_dir if cache_cf_files else "."
 
     output = (
-        """Could not set attribute `attr` on instance `__config__::Test (instantiated at ./main.cf:8)` """
-        """(reported in Construct(Test) (./main.cf:8))
+        f"""Could not set attribute `attr` on instance `__config__::Test (instantiated at {cwd}/main.cf:8)` """
+        f"""(reported in Construct(Test) ({cwd}/main.cf:8))
 caused by:
-  Invalid value '1234', expected Number (reported in Construct(Test) (./main.cf:8))
+  Invalid value '1234', expected Number (reported in Construct(Test) ({cwd}/main.cf:8))
 """
     )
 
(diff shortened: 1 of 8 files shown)


tests/test_module_loader.py
```diff
@@ -45,7 +44,7 @@
 @pytest.mark.parametrize_any("editable_install", [True, False])
 def test_v2_module_loading(editable_install: bool, tmpdir: py.path.local, snippetcompiler, capsys, modules_v2_dir: str) -> None:
     # Work around caching problem in venv
-    feature_compiler_cache.set("False")
+    # feature_compiler_cache.set("False")
```
Hugo-Inmanta, line 47: This can probably go.
Before extending this test case I tried with it on and off and didn't see any difference.
Now that the test case actually checks that the caching takes place, we cannot set it to `False` anymore though.

tests/test_module_loader.py
```diff
@@ -78,7 +86,7 @@ def test_v1_and_v2_module_installed_simultaneously(
        * The V2 module is loaded and not the V1 module.
     """
     # Work around caching problem in venv
-    feature_compiler_cache.set("False")
+    # feature_compiler_cache.set("False")
```
Hugo-Inmanta, line 89: Same as above, no difference here when it's set to `False` or `True`.

tests/test_app.py
```diff
@@ -367,7 +367,8 @@ def test_startup_failure(tmpdir, postgres_db, database_name):
     assert code == 4
 
 
-def test_compiler_exception_output(snippetcompiler):
+@pytest.mark.parametrize("cache_cf_files", [True, False])
+def test_compiler_exception_output(snippetcompiler, cache_cf_files):
```
wouterdb, line 371: I would add a comment to indicate this test case is used to also test the caching and why there is no assertions related to that.  (basic smoke test for argument parsing)

tests/test_module_loader.py
wouterdb: remove!

tests/test_module_loader.py
wouterdb: remove!

tests/test_module_loader.py
arnaudsjs: I don't really remember. Let's remove it.

sanderr (approved):

wouterdb (approved):

Issue #3076, opened 2021-06-29
Fix memory leak introduced by usage rlu_cache


arnaudsjs:
The `src/inmanta/module.py` file uses the `lru_cache` decorator. This introduces a memory leak (see: https://bugs.python.org/issue44310)

PR #4187, opened 2022-04-29
Issue/3076 fix memory leak introduced by using rlu cache

Hugo-Inmanta:
# Description

**Problem:** 
lru-cache keeps strong references to cached items.
**Solution:**
Use a dictionary as a cache in each instance so that the memory is freed when this instance gets garbage collected.

Part of #3076 [ISO5, master]

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3076-fix-memory-leak-introduced-by-using-rlu-cache.yml
+++ b/changelogs/unreleased/3076-fix-memory-leak-introduced-by-using-rlu-cache.yml
@@ -1,0 +1,6 @@
+description: Fix memory leak caused by lru-cache keeping strong references to cached items
+change-type: patch
+destination-branches: [master, iso5]
+sections: {
+  bugfix: "{{description}}"
+}
--- a/src/inmanta/module.py
+++ b/src/inmanta/module.py
@@ -1471,7 +1471,7 @@ class Project(ModuleLike[ProjectMetadata], ModuleLikeWithYmlMetadataFile):
         self.project_path = path
         self.main_file = main_file
 
-        self._ast_cache = None  # Cache for expensive method calls
+        self._ast_cache: Optional[Tuple[List[Statement], BasicBlock]] = None  # Cache for expensive method calls
         self._metadata.modulepath = [os.path.abspath(os.path.join(path, x)) for x in self._metadata.modulepath]
         self.module_source: ModuleV2Source = ModuleV2Source(
             [repo.url for repo in self._metadata.repo if repo.type == ModuleRepoType.package]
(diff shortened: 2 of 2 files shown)


Hugo-Inmanta (commented):
Should I add a test ?

src/inmanta/module.py
```diff
@@ -2252,11 +2256,14 @@ def ensure_versioned(self) -> None:
         if not os.path.exists(os.path.join(self.path, ".git")):
             LOGGER.warning("Module %s is not version controlled, we recommend you do this as soon as possible.", self.name)
 
-    @lru_cache()
     def get_ast(self, name: str) -> Tuple[List[Statement], BasicBlock]:
         if self._project is None:
             raise ValueError("Can only get module's AST in the context of a project.")
 
+        # Check local cache
+        if hit := self._ast_cache.get(name, None) is not None:
```
wouterdb, line 2264: 1. I wouldn't use the walrus operator here
2. this is wrong, I think, as it caches the result of the comparison, this would make all compiler tests go up in smoke. 

(I think this clearly make the point against using it?)

wouterdb (changes requested):

src/inmanta/module.py
Hugo-Inmanta: This is only the un-caching part, the caching part happens L2283

Hugo-Inmanta: I think I misinterpreted your comment, but all the compiler tests are indeed going up into smoke. I will change it !

wouterdb (approved):

changelogs/unreleased/3076-fix-memory-leak-introduced-by-using-rlu-cache.yml
```diff
@@ -0,0 +1,6 @@
+description: Fix memory leak caused by lru-cache keeping strong references to cached items
+change-type: patch
+destination-branches: [master, iso5]
```
arnaudsjs, line 3: This bug also exists on iso4. Are you planning to create a separate PR for ISO4? That could be useful since a lot of changes were done to the module.py file between ISO4 and ISO5 and doing everything in one PR might cause a big merge conflict. I am not sure what is best.

arnaudsjs (changes requested):

changelogs/unreleased/3076-fix-memory-leak-introduced-by-using-rlu-cache.yml
Hugo-Inmanta: I will create a separate PR for ISO4 then.

arnaudsjs (approved):

PR #4203, opened 2022-05-04
Add inmanta-cli environment recompile command

### LINKED ISSUE #4052: Add environment recompile to inmanta-cli
1. add the ability to trigger a recompile via inmanta-cli
2. add a '--update' option

`inmanta-cli environment recompile --update`

andraskvr:
# Description

closes #4052 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/main.py
+++ b/src/inmanta/main.py
@@ -416,17 +416,17 @@ def environment_delete(client: Client, environment: str) -> None:
 @environment.command(name="recompile")
 @click.argument("environment")
 @click.option(
-    "--update", "-u", is_flag=True, required=False, default=False, show_default=True, help="Update the model before recompiling"
+    "--update", "-u", is_flag=True, required=False, default=False, show_default=True, help="Update the model and its dependencies before recompiling"
 )
 @click.pass_obj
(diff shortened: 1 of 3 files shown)


wouterdb (approved):

src/inmanta/main.py
```diff
@@ -413,6 +413,34 @@ def environment_delete(client: Client, environment: str) -> None:
...
+        "notify_change",
+        arguments=dict(
+            id=env_id,
+            update=update,
+            metadata={
+                "type": "cli",
+                "message": "Compile triggered from the cli",
+            },
+        ),
+    )
```
arnaudsjs, line 439: This API endpoint could return warning. I am specifically talking about the following warning: `Skipping compile because server compile not enabled for this environment.`. If we would end up in that scenario, this command would exist with status code 0 and it will log that the update was triggered successfully. That's not good with respect to user experience. I think we should at least log these warnings to the console.

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #4212, opened 2022-05-06
performance improvements for eager promises

sanderr:
# Improved performance for eager promise tracking.
Performance improvements by order of significance (copied from [comment on #3950](https://github.com/inmanta/inmanta-core/pull/3950#issuecomment-1117338668)):
1. big impact (+- 15% time, 10% mem): instead of passing `ResultVariable([])` through the requires dict, when there are no promises, don't pass anything.
2. smallish impact (+- 4% time, 2.5% mem): don't use full `ResultVariable` for passing promises, instead use simple object with minimal variable-like functionality.
3. smallish impact (+- 0% time, 2% mem): `__slots__` for all AST nodes
4. smallish impact (+- 1.5-3% time, 0% mem): `dct.update(other)` instead of `{**dct, **other}`
5. smallish impact (+- 1% time, 4% mem): clean up `DelayedResultVariable.promises` and `DelayedResultVariable.done_promises` on freeze.

All percentages are based on the compilerscaling test for 10000 elines and are relative to the run that is exactly the same except for the mentioned change. i.e. a 5% time impact means the changes resulted in a run time of 95% of whatever it was before.

All improvements together result in a 12.6% increase in compilation time and a 7.5% increase in memory usage relative to master.

# Other changes
- Use simple `ResultVariable`-like object (from 2) for `For`'s gradual helper as well. Applied last so it didn't affect performance metrics stated above. Negligible impact but cleaner.


src/inmanta/execute/runtime.py
```diff
@@ -310,24 +347,31 @@ class DelayedResultVariable(ResultVariable[T]):
...
-        self.done_promises: Set[IPromise] = set()
+        self.promises: Optional[List[IPromise]] = []
+        self.done_promises: Optional[Set[IPromise]] = set()
         self.queued = False
         self.queues = queue
         if self.can_get():
             self.queue()
 
     def get_promise(self, provider: "Statement") -> ISetPromise[T]:
         promise: ISetPromise[T] = SetPromise(self, provider)
```
wouterdb, line 358: Would it matter, that in case `self.promises is None` you always return the same dummy. 

(i.e. construct an no-op `ISetPromise` on the module level an always return that same, single instance)

src/inmanta/execute/runtime.py
```diff
@@ -310,24 +347,31 @@ class DelayedResultVariable(ResultVariable[T]):
...
-        self.promises.append(promise)
+        if self.promises is not None:
+            # only track the promise if this variable has not been frozen yet.
+            self.promises.append(promise)
         return promise
 
     def get_progression_promise(self, provider: "Statement") -> ProgressionPromise:
+        if self.promises is None:
+            return None
```
wouterdb, line 366: I guess this method is not typed correctly?

tests/compiler/test_slots.py
```diff
@@ -90,8 +95,63 @@ def test_slots_rt():
 
 
 def test_slots_ast():
-    assert_slotted(Location("", 0))
-    assert_slotted(Range("", 0, 0, 0, 0))
+    """
+    Verify that all AST nodes below RequiresEmitStatement and all location objects use slots.
+    """
+    for ast_node_cls in chain(get_all_subclasses(RequiresEmitStatement), get_all_subclasses(Location)):
```
wouterdb, line 101: nice!

wouterdb (approved):

src/inmanta/execute/runtime.py
sanderr: Indeed, the parent has the correct type but I forgot to change this one.

src/inmanta/execute/runtime.py
sanderr: I'll try but I don't think so. I think this is a pretty rare path: the result variable is frozen before even the name pointing to it could be dereferenced.

tests/compiler/test_slots.py
sanderr: It took some puzzling but I think it was worth it.

src/inmanta/execute/runtime.py
sanderr: No measurable impact.

wouterdb (approved):

PR #4224, opened 2022-05-10
Issue/2582 different escaping rules for single and triple quotes

### LINKED ISSUE #2582: Different escaping rules for single and triple quotes
The following piece of code:

```
std::print("""trippel hello\nworld""")
std::print("single hello\nworld")
```

Will print:
```
trippel hello\nworld
single hello
world
```

The expected output should be:
```
trippel hello
world
single hello
world

### PR DESCRIPTION
**Pull request by Hugo-Inmanta**
# Description

Use the same escaping rules for multi-line strings and for regular strings.

closes #2582 

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/2582-different-escaping-rules-for-single-and-triple-quotes.yml
+++ b/changelogs/unreleased/2582-different-escaping-rules-for-single-and-triple-quotes.yml
@@ -1,0 +1,5 @@
+---
+description: Use the same escaping rules for multi-line strings and for regular strings.
+issue-nr: 2582
+change-type: patch
+destination-branches: [iso5, master]
--- a/tests/compiler/test_strings.py
+++ b/tests/compiler/test_strings.py
@@ -135,3 +135,16 @@ std::print(test_string_11)
 '''
     )
     compiler.do_compile()
+
+
+def test_escaping_rules_single_vs_triple_quotes(snippetcompiler):
+    snippetcompiler.setup_for_snippet(
+        r'''
+test_string_1 = """trippel hello\nworld"""
+test_string_2 = "single hello\nworld"
+std::print(test_string_1)
+std::print(test_string_2)
+'''
+    )
+    compiler.do_compile()
+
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: changelogs/unreleased/2582-different-escaping-rules-for-single-and-triple-quotes.yml [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+description: Use the same escaping rules for multi-line strings and for regular strings.
+issue-nr: 2582
+change-type: patch
+destination-branches: [iso5, master]
```
arnaudsjs, line 5: I guess this issue also exists in ISO4?

tests/compiler/test_strings.py
```diff
@@ -135,3 +135,16 @@ def test_allowed_line_breaks(snippetcompiler):
...
+    snippetcompiler.setup_for_snippet(
+        r'''
+test_string_1 = """trippel hello\nworld"""
+test_string_2 = "single hello\nworld"
+std::print(test_string_1)
+std::print(test_string_2)
+'''
+    )
+    compiler.do_compile()
+
```
arnaudsjs, line 150: This test case doesn't contain an assertion?

arnaudsjs (changes requested):

tests/compiler/test_strings.py
Hugo-Inmanta: Yes, this is just a visual test to check that the problem mentioned in the [issue](https://github.com/inmanta/inmanta-core/issues/2582) is solved. I will mention it in the docstring and the name of the test.

arnaudsjs: Like that the test case is not really useful. Nobody is ever going to have a look at this output after this change is merge and we will never notice if a change breaks this functionality. Next to that, it's not really difficult to add an assertion, so let's add it.

Hugo-Inmanta: Of course! I was missing the point :+1:

changelogs/unreleased/2582-different-escaping-rules-for-single-and-triple-quotes.yml
```diff
@@ -0,0 +1,5 @@
+---
+description: Use the same escaping rules for multi-line strings and for regular strings.
+issue-nr: 2582
+change-type: patch
+destination-branches: [iso4, iso5, master]
```
wouterdb, line 5: definitely not iso4

sanderr (approved):

arnaudsjs (approved):

PR #4243, opened 2022-05-16
Fix issue where updating the modules always spawns a pip sub-process

### LINKED ISSUE #4055: Compiler performance has decreased
The runtime of the tests of the lsm module has increased. After investigation it turned out that the `project.compile()` method of the `project` fixture of `pytest-inmanta` project, has become substantially slower. The assumption is that this performance degradation is related to the introduction of module V2.

arnaudsjs:
# Description

The `are_installed()` method in `env.py`, that checks whether a requirement is installed in the active Python environment, didn't take into account the markers set on a requirement. As such, the method would always evaluate to `False` when a marker was set that wasn't applicable for the active environment. This caused degraded performance on all test suites that ran using python >=3.7 and used the `project` fixture, because the requirements.txt file of the std module has the following requirement set:

```
dataclasses~=0.7;python_version<'3.7'
```

Each time the `are_installed()` method evaluates to False, a pip sub-process is spawned. This slowed down the tests a lot. This patch increases the runtime of the `lsm` test suite by 2.2x

closes #4055 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/4055-fix-bug-compiler-performance-decreased.yml
+++ b/changelogs/unreleased/4055-fix-bug-compiler-performance-decreased.yml
@@ -1,0 +1,7 @@
+---
+description: Fix bug where `inmanta project install` and `inmanta project update` always invokes pip, even when all dependencies are already met.
+issue-nr: 4055
+change-type: patch
+destination-branches: [master, iso5]
+sections:
+  bugfix: "{{description}}"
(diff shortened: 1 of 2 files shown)


sanderr (approved):
Nice catch! I'm curious to see the impact.

PR #4275, opened 2022-05-23
[FIX] bump stylesheet version to 4.11.1

### LINKED ISSUE #3770: Update the version of the swagger UI served by the Inmanta server
The swagger UI served by the `/docs` endpoint of the Inmanta server is quite old. The documentation hosted on our webserver, already has the new, more user friendly version of the UI (See: https://docs.inmanta.com/community/dev/reference/openapi.html#http://). This ticket should make sure that the Inmanta server serves the new version of the swagger UI.

Hugo-Inmanta:
# Description

bump stylesheet version to 4.11.1

closes #3770 
# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/protocol/openapi/converter.py
+++ b/src/inmanta/protocol/openapi/converter.py
@@ -124,7 +124,7 @@ class OpenApiConverter:
         <meta charset="UTF-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <script src="//unpkg.com/swagger-ui-dist@^3.25.0/swagger-ui-bundle.js"></script>
-        <link rel="stylesheet" href="//unpkg.com/swagger-ui-dist@^3.25.0/swagger-ui.css" />
+        <link rel="stylesheet" href="//unpkg.com/swagger-ui-dist@^4.11.1/swagger-ui.css" />
         <title>Inmanta Service Orchestrator API</title>
     </head>
     <body>


Hugo-Inmanta:
I haven't tested if this version bump behaves correctly. I don't know how/if it should be tested manually ?

src/inmanta/protocol/openapi/converter.py
```diff
@@ -124,7 +124,7 @@ def get_swagger_html(self, openapi_spec: str) -> str:
         <meta charset="UTF-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <script src="//unpkg.com/swagger-ui-dist@^3.25.0/swagger-ui-bundle.js"></script>
-        <link rel="stylesheet" href="//unpkg.com/swagger-ui-dist@^3.25.0/swagger-ui.css" />
+        <link rel="stylesheet" href="//unpkg.com/swagger-ui-dist@^4.11.1/swagger-ui.css" />
```
andraskvr, line 127: The `script` and the `stylesheet` version should match.
I think there is also a `latest` tag that you can use, but a specific version is also fine.

If you check the page from the `/docs` endpoint and it looks ok, you don't have to do any special testing.

Hugo-Inmanta: > If you check the page from the /docs endpoint and it looks ok, you don't have to do any special testing.

I can only do this after it is merged in and the nightly build for documentation has run, right ?

andraskvr: Not exactly. This is served by an endpoint on the server, so you only need a server (even one started during the tests is fine) with your patch applied to it.
I checked it, and it's fine (looks almost exactly like the previous version)

andraskvr (approved):

PR #4276, opened 2022-05-23
issue 3383: verify requirements not broken after module install

### LINKED ISSUE #3383: environment checks after module install do not suffice
The checks performed on the active Python environment after installing a new v2 module seem to be insufficient. This ticket describes a PoC scenario but further investigation is required:
- pinpoint the root cause
- which of the steps in the example scenario produce undesired results, regardless of whether they contribute to the root cause
- could other scenarios present similar issues?

# Example
The example below is not minimal. It illustrates an interaction between two module installs and core itself. I believe similar issues can arise between only modules as well as between a single module and core.

Consider two modules `a` and `b` with `requests<2.26` and `idna>=3` as their dependencies respectively (full modules zip attached [here](https://app.zenhub.com/files/67439560/f91f9fea-ab8c-461e-ba68-834d0269acd6/download)). Steps to reproduce:
1. create a new venv
2. install latest core -> core depends on `cookiecutter`, which depends on `requests`, latest version (2.26) of which depends on `idna<4`. Environment now contains `requests==2.26.0, idna==3.3`. `pip check` shows no issues.
3. `inmanta module install inmanta-module-a` -> dependency `requests<2.26` downgrades to `requests==2.25.1`, which constrains `idna<3`, so that is downgraded as well. Environment now contains `requests==2.25.1, idna==2.10`. `pip check` shows no issues.
4. `inmanta module install inmanta-module-b` -> dependency `idna>=3` upgrades `idna` again, without touching `requests`. Environment now contains `requests==2.25.1, idna==3.3`. `pip check` shows `requests 2.25.1 has requirement idna<3,>=2.5, but you have idna 3.3.`. Core is broken as a result.

# Additional questions
- Can a module's requirements currently result in installation of a package that conflicts directly with a constraint of core?
- To what extent should we protect/check transitive dependencies of previously installed modules? `inmanta.env.ActiveEnv.check` currently only checks their direct dependencies for conflicts.
- To what extent should we protect/check direct/transitive dependencies of core?

FloLey:
# Description

After installing a module verify the whole tree below modules, as well as the inmanta product and extension packages.
If any of those checks fail raise an exception.
For any other failure in the check output a warning.

closes #3383 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


sanderr (changes requested):

changelogs/unreleased/4209-module-install-pip-check.yml
```diff
@@ -0,0 +1,5 @@
+---
+description: run "pip check" after module installations to ensure no requirements have been broken.
+issue-nr: 4209
```
arnaudsjs, line 3: This issue number isn't correct. And the one in the PR references itself.
```suggestion
issue-nr: 3383
```

arnaudsjs (changes requested):

tests/test_module_loader.py
```diff
@@ -605,3 +605,68 @@ def test_project_requirements_dont_overwrite_core_requirements_index(
...
+    snippetcompiler_clean,
+    modules_v2_dir: str,
+    tmpdir: py.path.local,
+) -> None:
+    """
+    Show an error message when installing a module that breaks the dependencies
+    of another one. minimalv2module depends on y~=1.0.0 which requires x~=1.0.0.
+    after the install of minimalv2module we try to install minimalv2module2 which
+    requires x~=2.0.0. The y~=1.0.0 requirement is now broken as python package x
+    has now version 2.0.0 and y needs 1.0.0.
```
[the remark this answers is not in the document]
FloLey: @sanderr  So first install a V1 module with y~=1.0.0 and than a V2 module with y~=2.0.0? is that what you mean?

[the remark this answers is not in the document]
sanderr: I mistyped. I meant v1 module which depends on `x~=1.0.0` (without `y`) and keep the v2 module as is. But that's equivalent to what you describe.

src/inmanta/module.py
```diff
@@ -1988,10 +1990,13 @@ def verify_python_requires(self) -> None:
         """
         Verifies no incompatibilities exist within the Python environment with respect to installed module v2 requirements.
         """
-        if not env.ActiveEnv.check(in_scope=re.compile(f"{ModuleV2.PKG_NAME_PREFIX}.*")):
-            raise CompilerException(
-                "Not all installed modules are compatible: requirements conflicts were found. Please resolve any conflicts"
-                " before attempting another compile. Run `pip check` to check for any incompatibilities."
+        constraints: List[Requirement] = [Requirement.parse(item) for item in self.collect_python_requirements()]
+        constraints_for_marker = filter(lambda c: not (c.marker and not c.marker.evaluate()), constraints)
```
sanderr, line 1994: Wouldn't it be better to check for markers in `ActiveEnv.check` where `constraint_violations` is built?

tests/test_module_loader.py
```diff
@@ -605,3 +605,127 @@ def test_project_requirements_dont_overwrite_core_requirements_index(
...
+    snippetcompiler_clean,
+    modules_dir: str,
+    modules_v2_dir: str,
+    tmpdir: py.path.local,
+) -> None:
+    """
+    Show an error message when installing a module that breaks the dependencies
+    of another one. modulev1 depends on y~=1.0.0.
+    after the install of modulev1 we try to install minimalv2module2 which
+    requires y~=2.0.0. those 2 requirements conflict which each other.
```
sanderr, line 686: My assumption that this didn't raise an exception before was right?

sanderr (changes requested):
Nice! One minor change request.

src/inmanta/module.py
FloLey: Why would it be better?

sanderr: Because this is actually a bug in the `check` implementation rather than something specific to this verification.

arnaudsjs: I find this a easier to read:
```suggestion
        constraints_for_marker = filter(lambda c: not c.marker or c.marker.evaluate(), constraints)
```

src/inmanta/module.py
```diff
@@ -1988,10 +1990,13 @@ def verify_python_requires(self) -> None:
...
+        if not env.ActiveEnv.check(constraints=constraints_for_marker):
+            raise env.ConflictingRequirements(
+                "Module dependency resolution conflict: a module dependency constraint "
+                "was violated by another module. This most likely indicates an incompatibility between "
+                "two or more of the installed modules. You can get more details on the issue with 'pip check'."
```
arnaudsjs, line 1998: The `pip check` suggestion will only work for V2 module dependencies. I see the `check()` method writes warning log lines with more information about the conflict. Wouldn't it be better to refer to those. Or even better:

* Make the `check()` method raise the `env.ConflictingRequirements` exception and add an argument to the exception that contains the details regarding the violations.
* Catch the `env.ConflictingRequirements` here and turn it into a `CompilerException` with the above-mentioned message that also includes the specific details of the conflicts that exist.

Wdyt?

arnaudsjs (changes requested):

src/inmanta/env.py
```diff
@@ -605,9 +605,10 @@ def check(cls, in_scope: Optional[Pattern[str]] = None, constraints: Optional[Li
         constraint_violations: List[Tuple[Requirement, Optional[version.Version]]] = [
             (constraint, installed_versions.get(constraint.key, None))
             for constraint in all_constraints
-            if constraint.key not in installed_versions or str(installed_versions[constraint.key]) not in constraint
+            if (constraint.key not in installed_versions or str(installed_versions[constraint.key]) not in constraint)
+            and not (constraint.marker and not constraint.marker.evaluate())
```
sanderr, line 609: FYI, you can just chain ifs, which can help keep it readable without needing a bunch of parentheses and indentation:
```
l = [
    x
    for x in range(10)
    if x % 2 == 0
    if x > 5
]
assert l == [6, 8]
```

src/inmanta/module.py
sanderr: I like the suggestion. An `inmanta venv check` command or something could be nice when the user is trying to fix the issue. But depending on if/when we implement #4279 that might not be worth the effort.

FloLey: I can try to do that. The more info in the exception the better I guess? should I keep the logged warning or removed it?

arnaudsjs: I don't think it's useful to write the same information to two different streams. I would remove the log line if we incorporate that information into the error message.

FloLey:
Like this? @arnaudsjs

src/inmanta/env.py
```diff
@@ -59,7 +59,22 @@ class PackageNotFound(Exception):
...
+    def get_msg(self) -> str:
+        if self.msg:
+            return self.msg
+        msg: str = "Module dependency resolution conflict: a module dependency constraint \
+was violated by another module. This most likely indicates an incompatibility between \
+two or more of the installed modules. You can get more details on the issue with 'pip check'.\n"
+        if self.conflicts is not None:
+            for constraint, v in self.conflicts:
+                msg += "Incompatibility between constraint %s and installed version %s\n" % (constraint, v)
+        return msg
```
arnaudsjs, line 77: I wouldn't reference the `pip check` command anymore, as it might not detect the version conflict depending on whether the conflict involves a V1 module or not.

```suggestion
        msg: str = "Module dependency resolution conflict: a module dependency constraint \
was violated by another module. This most likely indicates an incompatibility between \
two or more of the installed modules."
        if self.conflicts is not None:
            for constraint, v in self.conflicts:
                msg += "\n\t* Incompatibility between constraint %s and installed version %s" % (constraint, v)
        return msg
```

arnaudsjs (changes requested):

src/inmanta/env.py
```diff
@@ -608,10 +617,11 @@ def check(cls, in_scope: Optional[Pattern[str]] = None, constraints: Optional[Li
...
             if not constraint.marker or constraint.marker.evaluate()
         ]
-        # constraints_for_marker = filter(lambda c: not (c.marker and not c.marker.evaluate()), constraints)
-        for constraint, v in constraint_violations:
-            LOGGER.warning("Incompatibility between constraint %s and installed version %s", constraint, v)
-        return len(constraint_violations) == 0
+        if len(constraint_violations) != 0:
+            msg: str = "Module dependency resolution conflict: a module dependency constraint \
+was violated by another module. This most likely indicates an incompatibility between \
+two or more of the installed modules."
```
sanderr, line 623: This is not in its place here. `ActiveEnv.check` does not know or care about modules, just the environment. This could be added higher up where the exception is caught.

src/inmanta/module.py
```diff
@@ -1991,12 +1991,10 @@ def verify_python_requires(self) -> None:
...
-        if not env.ActiveEnv.check(constraints=constraints):
-            raise env.ConflictingRequirements(
-                "Module dependency resolution conflict: a module dependency constraint "
-                "was violated by another module. This most likely indicates an incompatibility between "
-                "two or more of the installed modules. You can get more details on the issue with 'pip check'."
-            )
+        try:
+            env.ActiveEnv.check(constraints=constraints)
+        except env.ConflictingRequirements as e:
+            raise CompilerException(e.get_msg())
```
sanderr, line 1997: Why reraise this as a plain `CompilerException`? Is this left-over from before the inheritance change?

sanderr (changes requested):

src/inmanta/env.py
FloLey: What should I put as message here then? As now it is not optional anymore? :p

sanderr: The same goes for me.

src/inmanta/env.py
```diff
@@ -58,8 +59,19 @@ class PackageNotFound(Exception):
...
-    pass
+class ConflictingRequirements(CompilerException):
+    def __init__(
+        self, msg: Optional[str] = None, conflicts: Optional[List[Tuple[Requirement, Optional[version.Version]]]] = None
+    ):
+        self.conflicts = conflicts
+        self.msg = msg
+
+    def get_msg(self) -> str:
+        message: str = self.msg if self.msg else ""
```
sanderr, line 70:
```suggestion
        message: str = self.msg if self.msg else "Conflicting requirements:"
```
Otherwise the `\n\t` might look weird.

sanderr (approved):

arnaudsjs (changes requested):

sanderr (changes requested):

src/inmanta/env.py
```diff
@@ -596,37 +608,54 @@ def _install_from_list(
...
+
+        inmanta_requirements: Set[str] = set([requirement.key for requirement in cls._get_requirements_on_inmanta_package()])
+
+        full_strict_scope: Set[str] = set()
+        all_constraints: Set[Requirement] = set(constraints if constraints is not None else [])
+
+        for dist_info in pkg_resources.working_set:
+            requires = [requirement for requirement in dist_info.requires()]
+            package = Requirement.parse(f"{dist_info.key}=={dist_info.version}")
+            if strict_scope.fullmatch(dist_info.key) or dist_info.key in inmanta_requirements:
```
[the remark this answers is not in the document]
FloLey: from the test I did `dist_info.requires()`seems to get all the requirements of itself and of its children and of the children of it's children, ... . I didn't find the doc of this function really useful. 
But I might have concluded that to fast. I'll look at it again.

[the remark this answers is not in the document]
sanderr: I believe so. I assumed so at first from the fact that the previous implementation already used `.requires()` but turned out not to offer sufficient protection. I just tested it to confirm it now:
```python
>>> pkg_resources.working_set.find(pkg_resources.Requirement("inmanta")).requires()
[Requirement.parse('inmanta-core==6.0.2'), Requirement.parse('inmanta-ui==3.0.0')]
```

tests/test_env.py
```diff
@@ -426,21 +433,22 @@ def check_log(version: Optional[version.Version]) -> None:
             rec.message for rec in caplog.records
         }
 
-    assert env.ActiveEnv.check(in_scope)
+    env.ActiveEnv.check(in_scope)
 
     caplog.clear()
-    assert not env.ActiveEnv.check(in_scope, constraints)
+    env.ActiveEnv.check(in_scope, constraints)
```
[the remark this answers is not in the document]
FloLey: indeed, I will fix this

src/inmanta/env.py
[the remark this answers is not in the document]
FloLey: so you mean it is not recursive enough right?

tests/test_env.py
```diff
@@ -398,8 +398,14 @@ def test_active_env_check_basic(
...
             caplog.clear()
-            assert env.ActiveEnv.check(in_scope) == expect[0]
-            if not expect[0]:
+            if expect[0]:
+                env.ActiveEnv.check(in_scope)
+            else:
+                with pytest.raises(env.ConflictingRequirements):
+                    env.ActiveEnv.check(in_scope)
+            if expect[1] == "":
+                assert caplog.text == ""
```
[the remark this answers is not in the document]
FloLey: I'm not sure I understand :p I really want to verify the caplog is empty as nothing should be logged. Verifying that something is not present in the logs is not the same

src/inmanta/env.py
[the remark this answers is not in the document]
sanderr: Indeed.

diff.diff
```diff
@@ -0,0 +1,122 @@
+diff --git a/src/inmanta/env.py b/src/inmanta/env.py
```
sanderr, line 1: Don't forget to delete this file.

tests/test_env.py
```diff
@@ -395,12 +395,31 @@ def test_active_env_check_basic(
...
     def assert_all_checks(expect_test: Tuple[bool, str] = (True, ""), expect_nonext: Tuple[bool, str] = (True, "")) -> None:
+        """
+        verify what the check methode for 2 different scopes: for an existing package and a non existing one.
+
+        param: expect_test: tuple with as first value a bool and as seconde value a string. the bool is true if the execution
+        will not raise an error, false if it will raise an error. the second argument is the warning message that can be find
+        in the logs.
+        param: expect_nonext: tuple with as first value a bool and as seconde value a string. the bool is true if the execution
+        will not raise an error, false if it will raise an error. the second argument is the warning message that can be find
+        in the logs.
```
arnaudsjs, line 409:
```suggestion
        verify what the check method for 2 different scopes: for an existing package and a non existing one.

        param: expect_test: Tuple with as first value a bool and as second value a string. The bool is true if the execution
        will not raise an error, false if it will raise an error. The second argument is the warning message that can be find
        in the logs.
        param: expect_nonext: Tuple with as first value a bool and as second value a string. The bool is true if the execution
        will not raise an error, false if it will raise an error. The second argument is the warning message that can be find
        in the logs.
```

src/inmanta/env.py
```diff
@@ -86,6 +88,42 @@ def get_packages_in_working_set(cls) -> Dict[str, version.Version]:
...
+            return reduce(_get_tree_recursive_single, dists, acc)
+
+        def _get_tree_recursive_single(acc: abc.Set[str], dist: str) -> abc.Set[str]:
+            if dist in acc:
+                return acc
+
+            if dist not in installed_distributions:
+                return acc | {dist}
+
+            # recurse on direct dependencies if they have not yet been recursed on in a previous iteration
```
sanderr, line 119:
```suggestion
            # recurse on direct dependencies
```
The second part is no longer relevant: we recurse anyway, it just happens to be a noop if it has already been processed. My bad.

src/inmanta/env.py
```diff
@@ -617,47 +655,54 @@ def check(
...
 
         if len(constraint_violations_strict) != 0:
             raise ConflictingRequirements("Conflicting requirements:", constraint_violations_strict)
 
         for constraint, v in constraint_violations:
             LOGGER.warning("Incompatibility between constraint %s and installed version %s", constraint, v)
 
+    def _get_all_requires(dist_info: Distribution):
+        return
+
```
sanderr, line 705: This looks like something you experimented with and forgot to remove.

tests/test_env.py
```diff
@@ -38,14 +38,14 @@
...
-        allow_module_level=True,
-    )
+# if "inmanta-core" in env.process_env.get_installed_packages(only_editable=True):
+#     pytest.skip(
+#         "The tests in this module will fail if it runs against inmanta-core installed in editable mode, "
+#         "because the build tag on the development branch is set to .dev0 by default. The inmanta package protection feature "
+#         "would make pip install a non-editable version of the same package. But no version with build tag .dev0 exists "
+#         "on the python package repository.",
+#         allow_module_level=True,
+#     )
```
sanderr, line 48: Don't forget to comment this out again when you're done.

sanderr (changes requested):
Nice, looks about ready to me. Just some cleanup left.

sanderr (approved):

arnaudsjs (changes requested):

tests/test_env.py
```diff
@@ -409,38 +428,35 @@ def assert_all_checks(expect_test: Tuple[bool, str] = (True, ""), expect_nonext:
...
-    assert not env.ActiveEnv.check(in_scope, constraints)
-    check_log(None)
+    with pytest.raises(env.ConflictingRequirements):
+        env.ActiveEnv.check(in_scope, constraints)
 
     caplog.clear()
     create_install_package("test-package-one", version.Version("1.0.0"), [])
-    assert env.ActiveEnv.check(in_scope, constraints)
+    env.ActiveEnv.check(in_scope, constraints)
+    assert "Incompatibility between constraint" not in "caplog.tex"
```
arnaudsjs, line 453:
```suggestion
    assert "Incompatibility between constraint" not in caplog.text
```

src/inmanta/env.py
```diff
@@ -596,37 +649,59 @@ def _install_from_list(
...
+            if (c.key not in installed_versions or str(installed_versions[c.key]) not in c) and (
+                not c.marker or (c.marker and c.marker.evaluate())
+            ):
+                if c.key in full_strict_scope:
+                    constraint_violations_strict.add((c, installed_versions.get(c.key, None)))
+                else:
+                    constraint_violations.add((c, installed_versions.get(c.key, None)))
+
+        if len(constraint_violations_strict) != 0:
+            raise ConflictingRequirements("Conflicting requirements:", constraint_violations_strict)
```
arnaudsjs, line 701: In the log message below `v` can be None as well.

arnaudsjs (changes requested):

tests/test_env.py
FloLey: I should re-read my changes. sorry for this

arnaudsjs (approved):

PR #4280, opened 2022-05-23
security fixes for tests/test_env.py

sanderr:
# Description

While looking through this code I noticed there were some minor security issues here. We verify for some packages that they fail to install. If someone were to upload malicious packages with these names to PyPi, they would get loaded on any host running these tests.

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -90,7 +90,6 @@ def test_install_fails(tmpdir, caplog, monkeypatch):
     compose = env.PipCommandBuilder.compose_install_command
 
     def mock_compose(*args, **kwargs):
-        raise Exception("WORKS!")
         if "index_urls" in kwargs or len(args) < 5:
             return compose(*args, **{**kwargs, "index_urls": []})
         else:
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -95,10 +95,11 @@ def test_install_fails(tmpdir, caplog, monkeypatch):
             return compose(*args, **{**kwargs, "index_urls": []})
         else:
             return compose(*args[:5], [], *args[5:], **kwargs)
+
     monkeypatch.setattr(env.PipCommandBuilder, "compose_install_command", mock_compose)
 
     with pytest.raises(Exception):
-        venv.install_from_list([package_name], )
+        venv.install_from_list([package_name])
 
     log_sequence = LogSequence(caplog)
     log_sequence.contains("inmanta.env", logging.INFO, f"requirements:\n{package_name}")
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -93,7 +93,7 @@ def test_install_fails(tmpdir, caplog, monkeypatch):
         if "index_urls" in kwargs or len(args) < 5:
             return compose(*args, **{**kwargs, "index_urls": []})
         else:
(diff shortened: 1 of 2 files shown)


tests/test_env.py
```diff
@@ -79,13 +79,25 @@ def test_basic_install(tmpdir):
...
-def test_install_fails(tmpdir, caplog):
+def test_install_fails(tmpdir, caplog, monkeypatch):
     venv = env.VirtualEnv(tmpdir)
     venv.use_virtual_env()
     caplog.clear()
     caplog.set_level(logging.INFO)
     package_name = "non-existing-pkg-inmanta"
 
+    # monkeypatch pip install to set --no-index for security reasons (anyone could publish this package to PyPi)
+    compose = env.PipCommandBuilder.compose_install_command
```
arnaudsjs, line 90: This code can be removed.

tests/test_env.py
```diff
@@ -79,13 +79,25 @@ def test_basic_install(tmpdir):
...
+
+    def mock_compose(*args, **kwargs):
+        raise Exception("WORKS!")
+        if "index_urls" in kwargs or len(args) < 5:
+            return compose(*args, **{**kwargs, "index_urls": []})
+        else:
+            return compose(*args[:5], [], *args[5:], **kwargs)
```
arnaudsjs, line 97: This code as well

arnaudsjs (changes requested):

tests/test_env.py
```diff
@@ -79,13 +79,25 @@ def test_basic_install(tmpdir):
...
     venv.use_virtual_env()
     caplog.clear()
     caplog.set_level(logging.INFO)
     package_name = "non-existing-pkg-inmanta"
 
+    # monkeypatch pip install to set --no-index for security reasons (anyone could publish this package to PyPi)
+    compose = env.PipCommandBuilder.compose_install_command
+
+    def mock_compose(*args, **kwargs):
+        raise Exception("WORKS!")
```
sanderr, line 93:
```suggestion
```

arnaudsjs (changes requested):

tests/test_env.py
```diff
@@ -79,13 +79,24 @@ def test_basic_install(tmpdir):
...
     package_name = "non-existing-pkg-inmanta"
 
+    # monkeypatch pip install to set --no-index for security reasons (anyone could publish this package to PyPi)
+    compose = env.PipCommandBuilder.compose_install_command
+
+    def mock_compose(*args, **kwargs):
+        if "index_urls" in kwargs or len(args) < 5:
+            return compose(*args, **{**kwargs, "index_urls": []})
+        else:
+            return compose(*args[:5], [], *args[5:], **kwargs)
```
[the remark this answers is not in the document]
sanderr: After testing it turns out you're right but I must confess I don't get it. I made two mistakes
1. I forgot to exclude the existing index url argument, this part I get
2. I got the count wrong, this I don't get. Why does `cls` not show up in args? Because `compose_mock` is not a classmethod?

Anyway, it will definitely do, I'm just a bit confused.

arnaudsjs (approved):

PR #4296, opened 2022-05-25
Add protocol varkw support

bartv:
# Description

This PR adds support for **kwargs to collect all "unknown" attributes. This is required for certain APIs where the service attributes are merged at the top-level.

This feature is a bit dangerous because it disables lots of the validation. To limit the risk the following is done:

1. You need to explicitly set varkw in the decorator
2. A **kwargs variable is only allowed when varkw is set to true
3. The type of varkw always needs to be object (you could differ in the handler still) to force the user to validate the data

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/protocol/common.py
+++ b/src/inmanta/protocol/common.py
@@ -381,6 +381,7 @@ class MethodProperties(object):
         typed: bool = False,
         envelope_key: str = const.ENVELOPE_KEY,
         strict_typing: bool = True,
+        varkw: bool = False,
     ) -> None:
         """
         Decorator to identify a method as a RPC call. The arguments of the decorator are used by each transport to build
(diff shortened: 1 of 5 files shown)


wouterdb (approved):

src/inmanta/protocol/common.py
```diff
@@ -507,6 +517,26 @@ def _validate_function_types(self, typed: bool) -> None:
...
+            # check if a varkw is added to the method and has the type object
+            if full_spec.varkw is None:
+                raise InvalidMethodDefinition(
+                    f"varkw is set to true in the annotation but there is no ** variable in method {self.function}. "
+                    "Add `**kwargs: object` for example."
+                )
+
+            self._varkw_name = full_spec.varkw
+
+            # if the variable is there, it needs to have the type object. All other specialisations need to be validated
```
sanderr, line 530: Nice!

sanderr (approved):

PR #4316, opened 2022-06-02
 Update the inmanta export command for resourceSet

FloLey:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/export.py
+++ b/src/inmanta/export.py
@@ -486,6 +486,7 @@ class Exporter(object):
             tid=tid,
             version=version,
             resources=resources,
+            resource_set={"key": "value"},
             unknowns=unknown_parameters,
             resource_state=self._resource_state,
             version_info=version_info,
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -23,7 +23,11 @@ import pytest
 from inmanta import config, const
 from inmanta.ast import ExternalException
 from inmanta.const import ResourceState
+from inmanta.data import Resource
+from inmanta.env import LocalPackagePath
 from inmanta.export import DependencyCycleException
+from inmanta.module import InmantaModuleRequirement
+from utils import module_from_template
 
 
 def test_id_mapping_export(snippetcompiler):
@@ -412,3 +416,49 @@ exp::Test3(
         e.value.format_trace()
         == "Failed to get attribute 'real_name' for export on 'exp::Test3'\ncaused by:\nKeyError: 'tom'\n"
     )
+
+
+async def test_resource_set(snippetcompiler_clean, modules_v2_dir: str, tmpdir, environment) -> None:
+    """
+    test that the resourceSet is exported correctly
+    """
+    init_cf = """
+entity Res extends std::Resource:
+    string name
+end
+
+implement Res using std::none
+
+a = Res(name="test123")
+std::ResourceSet(name="test", resources=[a])
+"""
+    init_py = """
+from inmanta.resources import (
+    Resource,
+    resource,
+)
+@resource("minimalv2module::Res", agent="name", id_attribute="name")
+class Res(Resource):
+    fields = ("name",)
+"""
+    module_name: str = "minimalv2module"
+    module_path: str = str(tmpdir.join(module_name))
+
+    module_from_template(
+        os.path.join(modules_v2_dir, module_name),
+        module_path,
+        new_requirements=[InmantaModuleRequirement.parse("std")],
+        new_content_init_cf=init_cf,
+        new_content_init_py=init_py,
+    )
+    snippetcompiler_clean.setup_for_snippet(
+        """
+import minimalv2module
+        """,
+        install_v2_modules=[LocalPackagePath(module_path, False)],
+        project_requires=[InmantaModuleRequirement.parse("minimalv2module"), InmantaModuleRequirement.parse("std")],
+    )
+
+    await snippetcompiler_clean.do_export()
+    test = await Resource.get_list(environment=environment)
+    print(test)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1108,8 +1108,8 @@ class SnippetCompilationTest(KeepOnFail):
             dirs.extend(add_to_module_path)
         return f"[{', '.join(dirs)}]"
 
-    def do_export(self, include_status=False, do_raise=True):
-        return self._do_export(deploy=False, include_status=include_status, do_raise=do_raise)
+    async def do_export(self, include_status=False, do_raise=True):
+        return await off_main_thread(self._do_export(deploy=False, include_status=include_status, do_raise=do_raise))
 
     def get_exported_json(self) -> JsonType:
         with open(os.path.join(self.project_dir, "dump.json")) as fh:


src/inmanta/export.py
```diff
@@ -438,7 +441,32 @@ def deploy_code(self, conn: protocol.SyncClient, tid: uuid.UUID, version: Option
...
+        resource_sets: Optional[Dict[str, Optional[str]]] = {} if resource_set_instances else None
+        for resource_set_instance in resource_set_instances:
+            name = resource_set_instance.get_attribute("name").value
+            resources_in_set = resource_set_instance.get_attribute("resources")
+            for resource_in_set in resources_in_set.value:
+                resource_id = str(
+                    Resource.create_from_model(self, str(resource_in_set.type), DynamicProxy.return_value(resource_in_set))
+                )
+                if resource_id in resource_sets:
+                    raise CompilerException("resource '%s' can not be part of multiple ResourceSets" % resource_id)
```
FloLey, line 458: I'm not sure this should be a compilerException?

src/inmanta/export.py
```diff
@@ -438,7 +440,34 @@ def deploy_code(self, conn: protocol.SyncClient, tid: uuid.UUID, version: Option
...
+        if not resource_set_instances:
+            return None
+        resource_sets: Dict[str, Optional[str]] = {}
+        for resource_set_instance in resource_set_instances:
+            name: str = resource_set_instance.get_attribute("name").value
+            resources_in_set: ResultVariable = resource_set_instance.get_attribute("resources")
+            for resource_in_set in resources_in_set.value:
+                resource_id: str = str(
+                    Resource.create_from_model(self, str(resource_in_set.type), DynamicProxy.return_value(resource_in_set))
+                )
```
arnaudsjs, line 457: The fact that this becomes a resource_id and not a resource_version_id is very implicit. I would make this more explicit
```suggestion
                resource_id: str = 
                    Resource.create_from_model(self, str(resource_in_set.type), DynamicProxy.return_value(resource_in_set))
                ).id.resource_str()
```

tests/test_export.py
```diff
@@ -412,3 +416,113 @@ def test_3787_key_error_export(snippetcompiler):
...
+        module_path,
+        new_requirements=[InmantaModuleRequirement.parse("std")],
+        new_content_init_cf=init_cf,
+        new_content_init_py=init_py,
+    )
+    snippetcompiler_clean.setup_for_snippet(
+        """
+import minimalv2module
+        """,
+        install_v2_modules=[LocalPackagePath(module_path, False)],
```
arnaudsjs, line 463:
```suggestion
        install_v2_modules=[LocalPackagePath(module_path, editable=False)],
```

tests/test_export.py
```diff
@@ -412,3 +416,113 @@ def test_3787_key_error_export(snippetcompiler):
...
```
arnaudsjs, line 520:
```suggestion
        install_v2_modules=[LocalPackagePath(module_path, editable=False)],
```

arnaudsjs (changes requested):

src/inmanta/export.py
```diff
@@ -438,7 +440,34 @@ def deploy_code(self, conn: protocol.SyncClient, tid: uuid.UUID, version: Option
...
+        """
+        if not resource_set_instances:
+            return None
+        resource_sets: Dict[str, Optional[str]] = {}
+        for resource_set_instance in resource_set_instances:
+            name: str = resource_set_instance.get_attribute("name").value
+            resources_in_set: ResultVariable = resource_set_instance.get_attribute("resources")
+            for resource_in_set in resources_in_set.value:
+                resource_id: str = str(
+                    Resource.create_from_model(self, str(resource_in_set.type), DynamicProxy.return_value(resource_in_set))
```
wouterdb, line 456: This is very inefficient. It re-runs the complete resources creation process again, to drop everything except the ID. 
The resource creation process can be very expensive. 

I think it is better to use an approch similar to `convert_requires`, where the dict from instance to resource is used. 

This may require some refactoring

src/inmanta/export.py
wouterdb: I think this is OK

wouterdb (changes requested):

src/inmanta/export.py
```diff
@@ -351,6 +350,9 @@ def run(
         self._validate_graph()
 
         resources = self.resources_to_list()
+        resource_sets: Optional[List[Instance]] = (
```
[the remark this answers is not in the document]
FloLey: indeed, I'll change that. But in the 
```
        result = conn.put_version(
            tid=tid,
            version=version,
            resources=resources,
            resource_sets=resource_sets,
            unknowns=unknown_parameters,
            resource_state=self._resource_state,
            version_info=version_info,
            compiler_version=get_compiler_version(),
        )

```
resource_sets should still be set to `None` if there are no resource_sets right?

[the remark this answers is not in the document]
arnaudsjs: No. Both `None` and `{}` indicate that there are not resource sets. Maybe we should change the default value of `resource_sets` argument in the API endpoint from None to `{}`. That way there is no room for confusion anymore.

src/inmanta/export.py
```diff
@@ -438,7 +441,30 @@ def deploy_code(self, conn: protocol.SyncClient, tid: uuid.UUID, version: Option
...
+        """
+        return a dictonary with as keys resource_ids and as values the name of the resource_set
+        the resource belongs to. return None if no resource_set is defined.
+        """
+        resource_sets: Dict[str, Optional[str]] = {}
+        for resource_set_instance in resource_set_instances:
+            name: str = resource_set_instance.get_attribute("name").get_value()
+            resources_in_set: List[Instance] = resource_set_instance.get_attribute("resources").get_value()
+            for resource_in_set in resources_in_set:
+                resource_id: str = self.resource_mapping[resource_in_set].id.resource_str()
```
wouterdb, line 454: This can fail. 

It is possible for an entity to be of type 'std::Resource' but not be an actual resource that is exported (i.e. the python half is missing).

If such a resource is in the resource set, this will fail. 

for example

```
std::ResourceSet(name="resource_set_1", resources=[std::Resource()])

implement std::Resource using std::none
```

I would add a testcase to this effect.


I'm unsure how to handle this, I think a warning might be sufficient? 
(some resource are also sometimes a resources, sometimes they are not. A resource can throw a specific exception to be discarded. Not failing hard on this may allow more flexible use of such resources, ....) 

@arnaudsjs wdyt?

wouterdb (changes requested):

src/inmanta/export.py
arnaudsjs: Yes. I think a warning would be sufficient.

src/inmanta/export.py
```diff
@@ -438,7 +441,30 @@ def deploy_code(self, conn: protocol.SyncClient, tid: uuid.UUID, version: Option
 
         upload_code(conn, tid, version, code_manager)
 
-    def commit_resources(self, version: int, resources: List[Dict[str, str]], metadata: Dict[str, str], model: Dict) -> None:
+    def get_resource_sets(self, resource_set_instances: List[Instance]) -> Dict[str, Optional[str]]:
+        """
+        return a dictonary with as keys resource_ids and as values the name of the resource_set
+        the resource belongs to. return None if no resource_set is defined.
+        """
+        resource_sets: Dict[str, Optional[str]] = {}
```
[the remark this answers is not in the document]
FloLey: you mean `assert self.resource_mapping is not None`, right? Like suggested in the previous comment

src/inmanta/export.py
```diff
@@ -189,8 +190,36 @@ def _load_resources(self, types: Dict[str, Entity]) -> None:
...
                         )
 
+        self._load_resource_sets(types, resource_mapping)
         Resource.convert_requires(resource_mapping, ignored_set)
 
+    def _load_resource_sets(self, types: Dict[str, Entity], resource_mapping: Dict["Instance", "Resource"]) -> None:
+        """
+        load the resource_sets in a dict with as keys resource_ids and as values the name of the resource_set
+        the resource belongs to.
+        This method should only be called at the end of self._load_resources()
```
wouterdb, line 200: I would explain the reason.

```suggestion
        This method should only be called after all resources have been extracted from the model.
```

src/inmanta/export.py
```diff
@@ -189,8 +190,36 @@ def _load_resources(self, types: Dict[str, Entity]) -> None:
...
+        """
+        resource_sets: Dict[str, Optional[str]] = {}
+        resource_set_instances: List["Instance"] = (
+            types["std::ResourceSet"].get_all_instances() if types and "std::ResourceSet" in types else []
+        )
+        assert resource_mapping is not None
```
wouterdb, line 206: I would move this to the start of the method. 

Perhaps also add an assert that types is not None

tests/test_export.py
```diff
@@ -412,3 +417,155 @@ def test_3787_key_error_export(snippetcompiler):
...
+    test that an error is raised if a resource that is not exported is in a resource_set
+    """
+    init_cf = """
+std::ResourceSet(name="resource_set_1", resources=[std::Resource()])
+implement std::Resource using std::none
+"""
+    module_name: str = "minimalv2module"
+    module_path: str = str(tmpdir.join(module_name))
+
+    module_from_template(
```
wouterdb, line 544: Why go through the effort of setting up a module, if this whole module can just go into the main file (i.e. `setup_for_snippet` )

I think module_from_template is very slow and I see no reason to use it here. 

I think, if you drop all the modules v2 stuff, this will no longer be a slow test?

Also, why use `snippetcompiler_clean` instead of `snippetcompiler`? This clean version is a lot slower as well.

(sorry I didn't see this the first time around). 


(this applies to all testcases)

wouterdb (changes requested):

tests/test_export.py
FloLey: @wouterdb Do I understand correctly that only the part about `snippetcompiler_clean` applies to all the test cases as for the first two the whole module isn't contained into the main file ?

wouterdb (approved):

tests/test_export.py
```diff
@@ -412,3 +417,133 @@ def test_3787_key_error_export(snippetcompiler):
...
+        await snippetcompiler.do_export_and_deploy()
+    assert (
+        e.value.format_trace()
+        == "resource 'modulev1::Res[the_resource_a,name=the_resource_a]' can not be part of multiple ResourceSets"
+    )
+
+
+async def test_resource_not_exported(snippetcompiler, caplog, environment) -> None:
+    """
+    test that an error is raised if a resource that is not exported is in a resource_set
```
arnaudsjs, line 530:
```suggestion
    test that a warning is logged if a resource that is not exported is in a resource_set
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #4324, opened 2022-06-07
[FEAT] override inmanta_state_dir fixture

Hugo-Inmanta:
# Description

Add support for changing the Inmanta state directory to a writable location by the current user.

part of [#189](https://github.com/inmanta/pytest-inmanta/issues/189)

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -76,6 +76,7 @@ import sys
 import tempfile
 import time
 import traceback
+import typing
 import uuid
 import venv
 from configparser import ConfigParser
@@ -126,6 +127,15 @@ if PYTEST_PLUGIN_MODE:
 else:
     import utils
     from db.common import PGRestore
+
+if typing.TYPE_CHECKING:
+    # Local type stub for mypy that works with both pytest < 7 and pytest >=7
+    # https://docs.pytest.org/en/7.1.x/_modules/_pytest/legacypath.html#TempdirFactory
+
+    class TempdirFactory:
+        def mktemp(self, path: str) -> py.path.local:
+            ...
+
 
 logger = logging.getLogger(__name__)
 
@@ -659,12 +669,13 @@ def log_state_tcp_ports(request, log_file):
 
 
 @pytest.fixture(scope="function")
-async def server_config(event_loop, inmanta_config, postgres_db, database_name, clean_reset, unused_tcp_port_factory):
+async def server_config(
+    event_loop, inmanta_config, postgres_db, database_name, clean_reset, unused_tcp_port_factory, inmanta_state_dir
+):
     reset_metrics()
 
-    state_dir = tempfile.mkdtemp()
-
     port = str(unused_tcp_port_factory())
+    state_dir = str(inmanta_state_dir)
 
     config.Config.set("database", "name", database_name)
     config.Config.set("database", "host", "localhost")
@@ -686,7 +697,6 @@ async def server_config(event_loop, inmanta_config, postgres_db, database_name, 
     config.Config.set("server", "auto-recompile-wait", "0")
     config.Config.set("agent", "agent-repair-interval", "0")
     yield config
-    shutil.rmtree(state_dir)
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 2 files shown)


wouterdb:
@Hugo-Inmanta Why is this here? This requires a bit more explanation?

Hugo-Inmanta:
> @Hugo-Inmanta Why is this here? This requires a bit more explanation?

This is because of this discussion: https://github.com/inmanta/pytest-inmanta/pull/297#discussion_r887758091 and a talk with @arnaudsjs. As I understood it, we want to override the fixture in inmanta-core so that other fixtures that need a writable state dir use the same one.

arnaudsjs:
> > @Hugo-Inmanta Why is this here? This requires a bit more explanation?
> 
> This is because of this discussion: [inmanta/pytest-inmanta#297 (comment)](https://github.com/inmanta/pytest-inmanta/pull/297#discussion_r887758091) and a talk with @arnaudsjs. As I understood it, we want to override the fixture in inmanta-core so that other fixtures that need a writable state dir use the same one.

Indeed. I thought this was the intention of your comment on https://github.com/inmanta/pytest-inmanta/pull/297#discussion_r887758091. But maybe I misunderstood.

tests/conftest.py
```diff
@@ -659,12 +669,13 @@ def map_data_line(line):
...
-async def server_config(event_loop, inmanta_config, postgres_db, database_name, clean_reset, unused_tcp_port_factory):
+async def server_config(
+    event_loop, inmanta_config, postgres_db, database_name, clean_reset, unused_tcp_port_factory, inmanta_state_dir
+):
     reset_metrics()
 
-    state_dir = tempfile.mkdtemp()
-
     port = str(unused_tcp_port_factory())
+    state_dir = str(inmanta_state_dir)
```
arnaudsjs, line 678:
```suggestion
    state_dir = inmanta_state_dir
```

tests/conftest.py
```diff
@@ -659,12 +669,13 @@ def map_data_line(line):
 
 
 @pytest.fixture(scope="function")
-async def server_config(event_loop, inmanta_config, postgres_db, database_name, clean_reset, unused_tcp_port_factory):
+async def server_config(
+    event_loop, inmanta_config, postgres_db, database_name, clean_reset, unused_tcp_port_factory, inmanta_state_dir
```
arnaudsjs, line 673:
```suggestion
    event_loop, inmanta_config, postgres_db, database_name, clean_reset, unused_tcp_port_factory, inmanta_state_dir: str
```

tests/conftest.py
```diff
@@ -1509,3 +1519,14 @@ def guard_testing_venv():
...
+
+@pytest.fixture(scope="function")
+def inmanta_state_dir(tmpdir_factory: "TempdirFactory") -> Iterator[str]:
+    """
+    This fixture overrides the inmanta_state_dir fixture defined in pytest-inmanta so
+    that fixtures that need to set up and write to a state directory use the same one.
+    """
+    inmanta_state_dir = tmpdir_factory.mktemp("inmanta_state_dir")
+    yield str(inmanta_state_dir)
+    inmanta_state_dir.remove(ignore_errors=True)
```
arnaudsjs, line 1532: Why ignore errors?

arnaudsjs (changes requested):

wouterdb (changes requested):
I would not do this here, because then we can only use it in the LSM module after this has been released (i.e. it binds lsm to this version of `pytest-inmanta-extensions`). This will open the gates of dependency hell.

We already have an entry point in the server fixture where you can put this initialization: server_pre_start.

https://github.com/inmanta/inmanta-lsm/blob/master/tests/conftest.py#L54

I would put this code there, so we don't bind `pytest-inmanta-extensions` to lsm.

tests/conftest.py
Hugo-Inmanta: After pondering for a while, I don't have a satisfying answer to this. I found very little documentation about this (https://py.readthedocs.io/en/latest/path.html#py._path.local.LocalPath.remove).
I think we want to make sure that the directory is cleaned up correctly when it's no longer used but I don't know how to go about it. 
I can only think of 1 error that could occur during deletion:
- access control rights -> the current process is not allowed to remove a subdir created by another process

arnaudsjs: True. And if that happens, something is clearly wrong. So there is no reason to hide it.

PR #4327, opened 2022-06-07
improve testcase stability

wouterdb:
# Description

Fix for race in testcase 

# Self Check:

- [X] Changelog entry

### DIFF
--- a/tests/agent_server/test_server_agent.py
+++ b/tests/agent_server/test_server_agent.py
@@ -1501,9 +1501,11 @@ async def test_autostart_mapping_update_uri(server, client, environment, async_f
     async_finalizer(a.stop)
 
     # Wait until agent is up
+    async def agent_in_db():
+        return len(await data.AgentInstance.get_list()) == 1
+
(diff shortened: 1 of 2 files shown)


tests/agent_server/test_server_agent.py
```diff
@@ -1501,9 +1501,11 @@ async def test_autostart_mapping_update_uri(server, client, environment, async_f
     async_finalizer(a.stop)
 
     # Wait until agent is up
+    async def agent_in_db() -> None:
```
arnaudsjs, line 1504:
```suggestion
    async def agent_in_db() -> bool:
```

arnaudsjs (changes requested):

tests/agent_server/test_server_agent.py
wouterdb: oh my, feels like monday morning. 

Thanks for catching this.

arnaudsjs (approved):

PR #4348, opened 2022-06-09
issue/4271-endpoint-partial-compiles

### LINKED ISSUE #4271: Add endpoint for partial compiles
This issue should add an extra API endpoint to the server that is invoked to store the resource of an partial compile. This new endpoint is equivalent to the `put_version` endpoint with the following two changes:

* The `resources` argument only contains the resources that resulted from the partial compile.
* The endpoint gets an extra argument that indicates which resource sets should be deleted from the model.

Tasks:

- [x] Add new `put_partial` endpoint
- [x] Add logic to compose the new version of the model from the resources in the previous version of the model and the resources in the partial compile (merge algorithm)
- [x] Validate constraints on the model generated from the partial compile (e.g. the dependency graph of the resulting model should be closed, resources cannot migrate to another resource_set using partial compile, etc.)

FloLey:
# Description

This commit adds an extra API endpoint to the server that is invoked to store the resource of an partial compile. This new endpoint is equivalent to the put_version endpoint with the following two changes:

The resources argument only contains the resources that resulted from the partial compile.
The endpoint gets an extra argument that indicates which resource sets should be deleted from the model.

closes #4271 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +426,138 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+                or Id.parse_id(resource_old["id"]).resource_str() == Id.parse_id(resource_partial["id"]).resource_str()
+            )
+            if resource_old is None:
+                result[resource_partial["id"]] = resource_partial
+            else:
+                if resource_set_partial != resource_set_old:
+                    raise BadRequest(
+                        f"A partial compile cannot migrate a resource({resource_partial['id']}) to another resource set"
+                    )
+                if resource_set_partial == None and resource_partial != resource_old:  # todo:: is the != comparison enough?
```
FloLey, line 553: see todo

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +426,138 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+                    raise BadRequest(
+                        f"Resource ({resource_partial['id']}) without a resource set cannot be updated via a partial compile"
+                    )
+                else:
+                    result[resource_partial["id"]] = resource_partial
+        return list(result.values())
```
FloLey, line 559: I think I could just use an array from the start. Should I change this?

src/inmanta/protocol/methods_v2.py
```diff
@@ -19,17 +19,56 @@
...
+    :param unknowns: A list of unknown parameters that caused the model to be incomplete
+    :param version_info: Module version information
+    :param compiler_version: version of the compiler, if not provided, this call will return an error
+    :param resource_sets: a dictionary describing which resource belongs to which resource set
+    :param removed_resource_sets: a list of resource_sets that should be deleted from the model
+    :param **kwargs: bypass the type checking for resources that result from the partial compile
+    """
+
+
+# TODO: create ticket to clean kwargs and type resources
```
FloLey, line 70: reminder for myself

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +427,151 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+            result: Set[str] = set()
+            for res in paired_resources:
+                resource_set = res[2]
+                if resource_set:
+                    result.add(resource_set)
+            return result
+
+        async def get_old_resources() -> List[tuple[Dict[str, Any], Optional[str]]]:
+            old_data = await data.Resource.get_list()
+            result: List[tuple[Dict[str, Any], Optional[str]]] = []
```
arnaudsjs, line 527: This is quite a complicated type definition. The problem with this is that it doesn't really explain what is inside the data structure. In other words, the code becomes hard to read. It would be good to introduce a new class (or a dataclass) here.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +427,151 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+            for res in old_data:
+                resource: Dict[str, Any] = {
```
arnaudsjs, line 529:
```suggestion
                resource: Dict[str, ResourceVersionIdStr] = {
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +427,151 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+        self,
+        env: data.Environment,
+        version: int,
+        resources: list[Any],
+        resource_state: Dict[ResourceIdStr, ResourceState] = {},
+        unknowns: List[Dict[str, PrimitiveTypes]] = [],
+        version_info: Optional[model.ModelVersionInfo] = None,
+        compiler_version: Optional[str] = None,
+        resource_sets: Dict[ResourceIdStr, Optional[str]] = {},
+        removed_resource_sets: List[str] = [],
```
arnaudsjs, line 457: I don't think the API layer verifies the `ResourceIdStr` type, because it's in reality just a alias for the str type (https://github.com/inmanta/inmanta-core/blob/6d247b20c7fd876febb1d01cff54ad312adc1303/src/inmanta/data/model.py#L181), but I could be wrong. Can you add a test to verify what happens when we provide a non resource id value as a key of one of those dictionaries?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +427,151 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+    ) -> List[Any]:
+        """
+        :param partial_updates: a list of dictonaries that results from a partial compile.
+        :param resource_sets: a dictionary mapping each resource (using its id) with a resource_set if it is in one
+        :param removed_resource_sets: The names of the resource sets removed in this partial compile.
+        """
+
+        def pair_resources_partial_update_to_old_version(
+            old_resources: List[tuple[Dict[str, Any], Optional[str]]], partial_updates: List[Dict[str, Any]]
+        ) -> List[tuple[Dict[str, Any], Optional[Dict[str, Any]], Optional[str], Optional[str]]]:
```
arnaudsjs, line 480: Same here. This type annotations are very hard to interpret.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +427,151 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+            1. the resource as returned by the partial compile
+            2. the resource as it is in the old version
+            3. the resource_set the resource belongs to in the partial compile
+            4. the resource_set the resource belongs to in the previous version
+
+            :param old_resources: a list of tuples containing the resource in the previous version of the model
+            and its resource_set.
+            :param partial_updates: The list of resources part of the partial compile.
+            """
+            paired_resources: List[tuple[Dict[str, Any], Optional[Dict[str, Any]], Optional[str], Optional[str]]] = []
```
arnaudsjs, line 492: Same here.

tests/agent_server/test_resource_sets.py
```diff
@@ -93,3 +93,430 @@ async def test_resource_sets_via_put_version(server, client, environment, client
     resource_sets_from_db = {resource.resource_id: resource.resource_set for resource in resource_list}
     expected_resource_sets = {**resource_sets, "test::Resource[agent1,key=key4]": None}
     assert resource_sets_from_db == expected_resource_sets
+
+
+async def test_put_partial_replace_resource_set(server, client, environment, clienthelper):
+    """
+    When an partial compile updates a certain resource set, the entire resource set
```
arnaudsjs, line 100:
```suggestion
    When a partial compile updates a certain resource set, the entire resource set
```

tests/agent_server/test_resource_sets.py
```diff
@@ -93,3 +93,430 @@ async def test_resource_sets_via_put_version(server, client, environment, client
     resource_sets_from_db = {resource.resource_id: resource.resource_set for resource in resource_list}
     expected_resource_sets = {**resource_sets, "test::Resource[agent1,key=key4]": None}
     assert resource_sets_from_db == expected_resource_sets
+
```
arnaudsjs, line 96: I would be nice to have a test that verifies one complex scenario. A test that starts with:
* resources in several different resource sets
* resources in the shared set.

The partial update does:
* An update of a subset of the resources sets
* An addition of shared resources

src/inmanta/server/services/orchestrationservice.py
sanderr: Or you could just change the return type to `collections.abc.Sequence`, I believe `dict.values()` is compatible with that.

sanderr: Oh no, `.values()` appears to be some sort of generator:
```python
>>> dct = {"a": 1, "b": 2}
>>> l = dct.values()
>>> l
dict_values([1, 2])
>>> del dct["a"]
>>> l
dict_values([2])
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +427,151 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+        version: int,
+        resources: list[Any],
+        resource_state: Dict[ResourceIdStr, ResourceState] = {},
+        unknowns: List[Dict[str, PrimitiveTypes]] = [],
+        version_info: Optional[model.ModelVersionInfo] = None,
+        compiler_version: Optional[str] = None,
+        resource_sets: Dict[ResourceIdStr, Optional[str]] = {},
+        removed_resource_sets: List[str] = [],
+    ) -> None:
+        pydantic.parse_obj_as(List[Dict[str, Any]], resources)
```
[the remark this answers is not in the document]
FloLey: I added an extra testcase

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: I wouldn't replace "ResourceIdStr" with "str". Mentioning it as "ResourceIdStr" has the advantage that it's clear to the reader that this should be a resource id. The only think that is missing here is the type validation. That still need to be added somehow. 

I would leave the test in. It would be a shame to throw away the effort we did in writing it. Next to that it performs an important check, so it's definitely useful.

tests/agent_server/test_resource_sets.py
arnaudsjs: I would even further extend it. Let's also add a new resource set and remove one. I don't think there is already a test that verifies these scenarios, is it?

tests/agent_server/test_resource_sets.py
```diff
@@ -93,3 +93,723 @@ async def test_resource_sets_via_put_version(server, client, environment, client
...
+    """
+    version = await clienthelper.get_version()
+    resources = [
+        {
+            "key": "key1",
+            "value": "value1",
+            "id": "test::Resource[agent1,key=key1],v=%d" % version,
+            "send_event": False,
+            "purged": False,
+            "requires": ["test::Resource[agent1,key=key2],v=%d" % version, "test::Resource[agent1,key=key3],v=%d" % version],
```
arnaudsjs, line 428: The requires field never contains the version of the model

tests/agent_server/test_resource_sets.py
```diff
@@ -93,3 +93,723 @@ async def test_resource_sets_via_put_version(server, client, environment, client
...
+            "test::Resource[agent1,key=key2]": "set-b",
+        },
+    )
+
+    assert result.code == 400
+    assert result.result["message"] == (
+        "Invalid request: Type validation failed for resources in put_"
+        "patrial.excepted an argument of type List[Dict[str, Any] but "
+        "received ['key1', 'key2]']"
+    )
```
arnaudsjs, line 732: Mentioning the used API endpoint is not very useful here. The user knows which endpoint he/she was invoking.
```suggestion
        "Invalid request: Type validation failed for resources argument. Expected an argument of type List[Dict[str, Any]] but "
        "received ['key1', 'key2]'"
    )
```

tests/agent_server/test_resource_sets.py
```diff
@@ -93,3 +93,723 @@ async def test_resource_sets_via_put_version(server, client, environment, client
...
+
+
+async def test_put_partial_ResourceIdStr(server, client, environment, clienthelper):
```
arnaudsjs, line 735: It would surprise me if this method name passes the flake8 check.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +427,151 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+            """
+            result: Set[str] = set()
+            for res in paired_resources:
+                resource_set = res[2]
+                if resource_set:
+                    result.add(resource_set)
+            return result
+
+        async def get_old_resources() -> List[tuple[Dict[str, Any], Optional[str]]]:
+            old_data = await data.Resource.get_list()
```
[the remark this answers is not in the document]
FloLey: I fixed the issue but I don't know how to write this test. could you help?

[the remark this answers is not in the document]
arnaudsjs: I would do the following:
* Create two environment and in each environment store (`put_version()`) the same resources with the same resource set (So both environment have the same content)
* Then do a partial compile (`put_partial()`) on one environment that updates the resource set
* Verify that the resources have been updates in the environment on which the partial compile was executed and the other one hasn't changed at all.

sanderr (commented):
Intermediate comments.

FloLey:
@arnaudsjs I tried to take all your comments into account, I hope I didn't miss some.

src/inmanta/protocol/methods_v2.py
FloLey: I'll remove them before merging, once I've created the ticket :p

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +561,63 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+        version: int,
+        resources: List[JsonType],
+        resource_state: Dict[ResourceIdStr, const.ResourceState],
+        unknowns: List[Dict[str, PrimitiveTypes]],
+        version_info: JsonType,
+        compiler_version: Optional[str] = None,
+        resource_sets: Optional[Dict[ResourceIdStr, Optional[str]]] = None,
+    ) -> Apireturn:
+        return await self._put_version(
+            env, version, resources, resource_state, unknowns, version_info, compiler_version, resource_sets
```
arnaudsjs, line 577: `_put_version` expects a `version_info` argument of type `Optional[model.ModelVersionInfo]`, but this method provides a `JsonType`.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,133 @@
...
+
+    async def merge_partial_with_old(self) -> List[Any]:
+
+        old_resources: Dict[str, ResourceWithResourceSet] = await self._get_old_resources()
+        paired_resources: List[PairedResource] = self._pair_resources_partial_update_to_old_version(
+            old_resources, self.partial_updates
+        )
+        updated_resource_sets: Set[str] = set(res.new_resource_set for res in paired_resources if res.new_resource_set)
+
+        def copy_with_incremented_version(resource: Dict[str, Any]) -> Dict[str, Any]:
```
arnaudsjs, line 152: This implementation builds on top of the assumption that the version associated with each resource in the resources argument of the put_partial endpoint, matches the version argument of the put_partial endpoint. But I don't think we currently have a check that validates whether that assumption is satisfied. Do you add such a check as well?

arnaudsjs (changes requested):

tests/agent_server/test_resource_sets.py
```diff
@@ -93,3 +94,915 @@ async def test_resource_sets_via_put_version(server, client, environment, client
...
+        version_info=None,
+        compiler_version=get_compiler_version(),
+        resource_sets={
+            "test::Resource[agent1,key=key1]": "set-a",
+            "test::Resource[agent1,key=key2]": "set-a",
+            "test::Resource[agent1,key=key3]": "set-b",
+            "test::Resource[agent1,key=key4]": "set-b",
+            "test::Resource[agent1,key=key91]": "set-f",
+            "test::Resource[agent1,key=key92]": "set-f",
+        },
```
arnaudsjs, line 739: The `resource_sets` argument should only contains entries for the resource part of the partial compile.
```suggestion
        resource_sets={
            "test::Resource[agent1,key=key1]": "set-a",
            "test::Resource[agent1,key=key2]": "set-a",
            "test::Resource[agent1,key=key91]": "set-f",
            "test::Resource[agent1,key=key92]": "set-f",
        },
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,133 @@
...
+        old_resource_set: Optional[str],
+    ) -> None:
+        self.new_resource = new_resource
+        self.old_resource = old_resource
+        self.new_resource_set = new_resource_set
+        self.old_resource_set = old_resource_set
+
+    def same_resource(self):
+        new_resource = set(self.new_resource).difference("id")
+        old_resource = set(self.old_resource).difference("id")
```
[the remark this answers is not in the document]
FloLey: because the id contains the version number and it will always be different.

tests/agent_server/test_resource_sets.py
```diff
@@ -93,3 +94,915 @@ async def test_resource_sets_via_put_version(server, client, environment, client
...
+        compiler_version=get_compiler_version(),
+        resource_sets={
+            "test::Resource[agent1,key=key1]": "set-a",
+            "test::Resource[agent1,key=key2]": "set-a",
+        },
+    )
+
+    assert result.code == 400, result.result
+    assert result.result["message"] == (
+        "Invalid request: Resource test::Resource[agent1,key=key1] was found in the " "resource_sets but not in the resources"
```
arnaudsjs, line 222:
```suggestion
        "Invalid request: Resource test::Resource[agent1,key=key1] was found in the resource_sets but not in the resources"
```

tests/agent_server/test_resource_sets.py
```diff
@@ -93,3 +94,915 @@ async def test_resource_sets_via_put_version(server, client, environment, client
...
+    assert result.result["message"] == ("Invalid request: Invalid id for resource hello")
+
+
+async def test_put_partial_different_env(server, client):
+    """
+    verify that put_partial won't modify other env.
+    """
+
+    result = await client.create_project("env-test-1")
+    project_id = result.result["project"]["id"]
```
arnaudsjs, line 922:
```suggestion
    result = await client.create_project("env-test-1")
    assert result.code == 200
    project_id = result.result["project"]["id"]
```

arnaudsjs (changes requested):

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,133 @@
...
+        self,
+        new_resource: Dict[str, Any],
+        old_resource: Optional[Dict[str, Any]],
+        new_resource_set: Optional[str],
+        old_resource_set: Optional[str],
+    ) -> None:
+        self.new_resource = new_resource
+        self.old_resource = old_resource
+        self.new_resource_set = new_resource_set
+        self.old_resource_set = old_resource_set
```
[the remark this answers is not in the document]
FloLey: Good question, i'll change it. I think I added ResourceWithResourceSet after I created PairedResource and I didn't think of changing it.

src/inmanta/server/services/orchestrationservice.py
FloLey: so what should I do? I thought I should not change the stable API?

arnaudsjs: Yes. You should not change the signature of a release endpoint in a breaking way or alter it's behavior, but anything else can be done. I think the easiest way forwards here is ass follows:

* Change the type of the `version_info` argument of the `_put_version()` endpoint to `Optional[JsonType]`
* In the handler of the `put_partial()` endpoint: Convert the `version_info` instance to a dictionary (a JsonType) before passing it to the `self._put_version()` method. This can be done using `version_info.dict()`

That way the type annotations are correct. Does that make sense.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -419,6 +427,158 @@ def safe_get(input: JsonType, key: str, default: object) -> object:
...
+        return await self._put_version(
+            env, version, resources, resource_state, unknowns, version_info, compiler_version, resource_sets
+        )
+
+    @handle(methods_v2.put_partial, env="tid")
+    async def put_partial(
+        self,
+        env: data.Environment,
+        version: int,
+        resources: list[Any],
```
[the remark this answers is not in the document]
bartv: If I understand correctly this contains resources which we cannot really type correctly because we would need support for JsonType (Dict[str, Any]). So there are two options: allowing this in protocol or using kwargs.

In the case of kwargs you indeed need to validate everything yourself. The way I did that was creating a pydantic object for all the arguments and creating that (with the option that additional attributes are not allowed)

This would indeed mean that you add `**kwargs: object` also to the handler and than do `a = custom_pydantic_class(**kwargs); a.resources`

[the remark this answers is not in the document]
sanderr: But when the field names are known ahead of time, why would we use `**kwargs` rather than `myfield: object`? You still need to validate it yourself but we at least still have a simple signature.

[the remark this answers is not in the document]
bartv: If object works I would use that, but iirc protocol also does not allow that.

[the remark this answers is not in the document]
sanderr: I see. I think it would be beneficial to add support (@FloLey not in this PR, don't worry). It's not perfect because it circumvents type checks but it does so in a similar manner as `**kwargs` while it has some advantages over the latter.

What about the second part of my comment? As I understood it protocol should check the type signature of kwargs fields in the method handlers and raise an exception if they're not typed as `object`. Did I misunderstand that, or is this check just not working as intended?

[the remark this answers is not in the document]
bartv: Protocol validates what does in, it does not check what the handlers itself define. There is "trust" that the handlers are correct.

arnaudsjs (changes requested):

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+                    merged_resources[new_resource.resource["id"]] = new_resource.resource
+        return list(merged_resources.values())
+
+    def _merge_resource_sets(self, old_resource_sets: Dict[ResourceIdStr, Optional[str]]) -> Dict[ResourceIdStr, Optional[str]]:
+        assert self.paired_resources is not None
+        paired_resources: List[PairedResource] = self.paired_resources
+        updated_resource_sets: Set[str] = set(
+            res.new_resource.resource_set for res in paired_resources if res.new_resource.resource_set
+        )
+        to_keep: List[str] = list(updated_resource_sets) + self.removed_resource_sets
```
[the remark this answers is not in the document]
FloLey: my naming was really bad. The 'to_keep' variable should not be called this way. It is used to track the resource_set that we don't want to copy into the next version: the removed_resource_sets as they are removed and the updated_resource_sets as they need to be overwritten.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+        new_resource_sets: Dict[ResourceIdStr, Optional[str]] = {k: v for k, v in old_resource_sets.items() if v not in to_keep}
```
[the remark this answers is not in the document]
FloLey: at that point it does indeed but the next call
`new_resource_sets.update(self.resource_sets)
`
adds the updates to new_resource_sets. Should I name this otherwise? (`update` returns nothing)

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+        new_resource_sets.update(self.resource_sets)
+        return new_resource_sets
```
arnaudsjs, line 208:
```suggestion
        unchanged_resource_sets: Dict[ResourceIdStr, Optional[str]] = {k: v for k, v in old_resource_sets.items() if v not in to_keep}
        return {**unchanged_resource_sets, **self.resource_sets}
```

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
arnaudsjs: I added a suggestion in a new comment to improve the naming.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+            return resource
+
+        to_keep: List[Dict[str, Any]] = [
+            copy_with_incremented_version(r.resource)
+            for r in list(old_resources.values())
+            if r.resource_set not in self.removed_resource_sets
+            and (r.is_shared_resource() or r.resource_set not in updated_resource_sets)
+        ]
+
+        merged_resources: Dict[ResourceIdStr, Dict[str, Any]] = {r["id"]: r for r in to_keep}
```
sanderr, line 169:
```suggestion
        merged_resources: Dict[ResourceIdStr, Dict[str, object]] = {r["id"]: r for r in to_keep}
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+
+        for paired_resource in paired_resources:
+            new_resource = paired_resource.new_resource
+            old_resource = paired_resource.old_resource
+            assert new_resource is not None
```
sanderr, line 174: Why this assert? It is not `Optional`.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+            assert (
+                old_resource is None
+                or old_resource is not None
+                and Id.parse_id(old_resource.resource["id"]).resource_str()
```
sanderr, line 178:
```suggestion
                or Id.parse_id(old_resource.resource["id"]).resource_str()
```
A bit simpler

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+                == Id.parse_id(new_resource.resource["id"]).resource_str()
+            )
+            if paired_resource.is_new_resource():
+                merged_resources[new_resource.resource["id"]] = new_resource.resource
+            else:
+                if paired_resource.resource_changed_resource_set():
+                    raise BadRequest(
+                        f"A partial compile cannot migrate a resource({new_resource.resource['id']}) " "to another resource set"
```
sanderr:
```suggestion
                        f"A partial compile cannot migrate a resource ({new_resource.resource['id']}) to another resource set"
```

src/inmanta/server/services/orchestrationservice.py
sanderr: A set would be more efficient for the lookup on the next line, and I guess the `update_resource_sets` can become arbitrarily big, so I think this is relevant.

sanderr (changes requested):
The logic looks good to me, nice! I still need to review the tests but I expect few remarks there.

src/inmanta/data/model.py
```diff
@@ -285,7 +285,7 @@ class ModelVersionInfo(BaseModel):
 class Resource(BaseModel):
     environment: uuid.UUID
     model: int
-    resource_id: ResourceVersionIdStr
+    resource_id: ResourceIdStr
```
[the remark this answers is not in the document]
FloLey: I think so :p I was surprised as `resource_version_id` is also present in the `Resource` class.

src/inmanta/protocol/methods_v2.py
```diff
@@ -24,12 +24,51 @@
...
+
+    :param tid: The id of the environment
+    :param version: The version of the configuration model
+    :param resource_state: A dictionary with the initial const.ResourceState per resource id
+    :param unknowns: A list of unknown parameters that caused the model to be incomplete
+    :param version_info: Model version information
+    :param compiler_version: version of the compiler, if not provided, this call will return an error
+    :param resource_sets: a dictionary describing which resources belong to which resource set
+    :param removed_resource_sets: a list of resource_sets that should be deleted from the model
+    :param **kwargs: support value 'resources' of type 'List[Dict[str, Any]'
```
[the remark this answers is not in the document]
FloLey: well it's temporary until I create the ticket and the ticket is handled.

src/inmanta/resources.py
```diff
@@ -532,6 +532,9 @@ def set_version(self, version: int) -> None:
 
         self._version = version
 
+    def increment_version(self) -> None:
```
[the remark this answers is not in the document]
FloLey: I see what you mean. I'll make a new Id instance with the incremented id.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+            pair: PairedResource = PairedResource(ResourceWithResourceSet(partial_update, resource_set), None)
+            if key in old_resources:
+                pair.old_resource = ResourceWithResourceSet(old_resources[key].resource, old_resources[key].resource_set)
+            paired_resources.append(pair)
+        self.paired_resources = paired_resources
+
+    async def _get_old_resources(
+        self,
+    ) -> Tuple[Dict[ResourceIdStr, ResourceWithResourceSet], Dict[ResourceIdStr, Optional[str]]]:
+        old_data = await data.Resource.get_resources_in_latest_version(environment=self.env.id)
```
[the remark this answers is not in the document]
FloLey: Should I do something about this now or create a ticket?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+        returns a list of paired resources
+
+        :param old_resources: a list of tuples containing the resource in the previous version of the model
+        and its resource_set.
+        :param partial_updates: The list of resources part of the partial compile.
+        """
+        partial_updates: List[Dict[str, Any]] = self.partial_updates
+        paired_resources: List[PairedResource] = []
+        for partial_update in partial_updates:
+            key = Id.parse_id(partial_update["id"]).resource_str()
```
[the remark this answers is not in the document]
FloLey: Do I understand correctly that this is the same issue for which I will create a ticket? (to validate the resources with a pydantic type)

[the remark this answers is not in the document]
arnaudsjs: The problem is that the exporter currently doesn't export any type information regarding the different entities in the model like LSM does. As such, the server cannot compose a pydantic object to properly validate a resource. We discussed this with Bart and the goal is to add this in the future, but not as part of this epic. So at the moment we can only do very basic validation, such as verifying that the id field is present that that it actually represents a resource version id. But you are right about the fact that we have no guarantee here that the key `id` actually exists.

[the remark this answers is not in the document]
bartv: We could always define a pydantic resource that has all the mandatory fields (id, requires??, ...) and allows for additional fields (default pydantic behaviour). This already cleans up the code a lot.

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
arnaudsjs: I would detect this scenario and make the `put_partial()` endpoint return a status code 409 (Conflict) if it occurs. The user/compiler-service can then respond to it by running a full compile for example. @sanderr @wouterdb Wdyt?

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
sanderr: That's what I had in mind. It's a clear check that everything is there and of the appropriate type, and it clearly conveys what has been validated already.

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
sanderr: When we discussed this yesterday I agreed with that, but the more I think about it the more I'm leaning towards the lock. I think it should be relatively simple to implement and it's quite a bit more flexible. It allows for out-of-order `put_version` calls because it guards against the write-since-last-read race.
To be clear, I'm not saying I have an objection against the 409, just that I like it a bit less.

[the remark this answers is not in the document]
sanderr: @FloLey we definitely need to address this as part of this epic, but not necessarily as part of this ticket I think. Creating a new one offers a bit more parallelism so I would go for that. Could you create one once the discussion here has reached some sort of concensus (even if that consensus is that we need to think about it some more)?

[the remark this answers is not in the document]
arnaudsjs: The problem with the lock-based approach is that it becomes quite complex once you take failure scenarios into account. Given the fact this edge case should be rather rare in practice (The server executes compiles in a strictly linear order), I think the locking approach would make things unnecessarily complex.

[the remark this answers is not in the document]
arnaudsjs: It's more complex in the sense that:
* We need to introduce an additional API endpoint
* We need to take into account that a client fails while it has the lock, so we need to make this lock timeout as well.
* It's also a breaking change that will make the previous exporter non-compatible with the new server.
As it's currently not clear whether we actually need this feature (i.e. whether this issue causes a lot of trouble in practice), I think it would be good to go for a simple approach and learn from the experiences we obtain using the initial implementation.

I am going to organize a short meeting to discuss this. Otherwise this discussion is going to take forever.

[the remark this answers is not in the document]
arnaudsjs: @FloLey I would leave this issue as it is for now and move on. I doesn't make sense to block the PR on this.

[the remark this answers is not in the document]
sanderr: I've just had a talk with Arnaud about this. I could leave our (semi-)conclusions here but I might as well just leave them in a new ticket and link there. So I'll create the tickets.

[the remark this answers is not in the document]
FloLey: thanks :)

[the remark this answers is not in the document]
sanderr: I've created #4405 and #4406. The latter is meant as a follow-up ticket to consider at some point in the future (not part of this epic), unless we decide to pick it up now after all. @arnaudsjs with the suggestion I included in the ticket, about dropping reservations for partial compiles only, I actually think it might be feasible (and semantically more clear) to implement it even in the short term. We'll need to discuss this further, then pick up one or the other.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -59,6 +62,161 @@
...
+                **res.attributes,
+            }
+            old_resources[res.resource_id] = ResourceWithResourceSet(resource, res.resource_set)
+        return old_resources, old_resource_sets
+
+    def _merge_resources(self, old_resources: Dict[ResourceIdStr, ResourceWithResourceSet]) -> List[Any]:
+        assert self.paired_resources is not None
+        paired_resources: List[PairedResource] = self.paired_resources
+        updated_resource_sets: Set[str] = set(
+            res.new_resource.resource_set for res in paired_resources if res.new_resource.resource_set
```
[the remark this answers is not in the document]
sanderr: inmanta/std#359 (asked Arnaud's and Wouter's review because this was not agreed upon yet).

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -76,6 +83,10 @@ def is_shared_resource(self) -> bool:
 
 
 class PairedResource:
+    """
+    Paires 2 ResourceWithResourceSet together.
```
sanderr, line 87:
```suggestion
    Pairs an old and a new ResourceWithResourceSet for the same id. Offers methods to inspect the difference.
```
A bit more context

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -134,77 +160,74 @@ def pair_resources_partial_update_to_old_version(
...
         )
-        changed_resource_sets: List[str] = list(updated_resource_sets) + self.removed_resource_sets
+        changed_resource_sets: Set[str] = set(list(updated_resource_sets) + list(self.removed_resource_sets))
         unchanged_resource_sets: Dict[ResourceIdStr, Optional[str]] = {
             k: v for k, v in old_resource_sets.items() if v not in changed_resource_sets
         }
         return {**unchanged_resource_sets, **self.resource_sets}
 
-    async def merge_partial_with_old(self) -> Tuple[List[object], Dict[ResourceIdStr, Optional[str]]]:
+    async def merge_partial_with_old(self) -> Tuple[List[Dict[str, Any]], Dict[ResourceIdStr, Optional[str]]]:
```
sanderr, line 230: Here too, why not stick with `object`? We really don't know anything more than that, do we?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -613,14 +636,23 @@ async def put_partial(
...
+        removed_resource_sets: Optional[List[str]] = None,
     ) -> None:
+        if resource_state is None:
+            resource_state = {}
+        if unknowns is None:
+            unknowns = []
+        if resource_sets is None:
+            resource_sets = {}
+        if removed_resource_sets is None:
+            removed_resource_sets = []
```
sanderr, line 652: Nice. I admit it's less readable but this is the convention (and rightly so).

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -134,77 +160,74 @@ def pair_resources_partial_update_to_old_version(
 
     async def _get_old_resources(
```
sanderr, line 161: I think this method and `_merge_resources` could still benefit from a docstring. How/where does this method get the old resources? What does `_merge_resources` do and return?

sanderr (changes requested):

src/inmanta/protocol/methods.py
```diff
@@ -59,7 +59,7 @@ async def convert_resource_version_id(rvid: model.ResourceVersionIdStr, metadata
         raise exceptions.BadRequest(f"Invalid resource version id: {rvid}")
 
 
-ENV_OPTS: Mapping[str, ArgOption] = {
+ENV_OPTS: Dict[str, ArgOption] = {
```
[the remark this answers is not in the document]
FloLey: It made mypy unhappy at other places, I started changing it at those places but each time I changed it at one place I needed to fix it somewhere else so I gave up as I don't think it is that important. If you insist I will try to fix it everywhere :p

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -134,77 +160,74 @@ def pair_resources_partial_update_to_old_version(
...
-            for r in list(old_resources.values())
+        to_keep: Sequence[ResourceMinimal] = [
+            incremented_ressouce_version(r.resource)
+            for r in old_resources.values()
             if r.resource_set not in self.removed_resource_sets
             and (r.is_shared_resource() or r.resource_set not in updated_resource_sets)
         ]
 
-        merged_resources: Dict[ResourceIdStr, Dict[str, object]] = {r["id"]: r for r in to_keep}
+        merged_resources: Dict[ResourceVersionIdStr, Dict[str, Any]] = {r.id: r.dict() for r in to_keep}
```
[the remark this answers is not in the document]
FloLey: this is a mistake,thanks

src/inmanta/protocol/methods.py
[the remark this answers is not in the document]
sanderr: No, this one is fine. It surprises me but it's not really important. The ones about `object` are because the argument doesn't fly there: if the change makes mypy complain, it means there is an issue you were hiding with the `Any`.

arnaudsjs (changes requested):

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -170,12 +174,23 @@ async def _get_old_resources(
...
-            res.new_resource.resource_set for res in paired_resources if res.new_resource.resource_set is not None
+            res.new_resource.resource_set for res in paired_resources if not res.new_resource.is_shared_resource()
         )
+        """
+        Merges the resources of the partial compile with the old resources. To do so it keeps the old resources that are not in
+        the removed_resource_sets, that are in the shared resource_set and that are not being updated. It then adds the
+        resources coming form the partial compile if they don't break any rule:
+        - cannot move a resource to another resource set
+        - cannot update resources without a resource set.
+        """
```
sanderr, line 187: Looks like you put this a couple lines too low?

sanderr (changes requested):
Last one, tiny detail.

src/inmanta/data/model.py
```diff
@@ -308,9 +308,30 @@ class ResourceMinimal(BaseModel):
 
     id: ResourceVersionIdStr
 
+    @classmethod
+    @validator("id")
+    def name_must_contain_space(cls, v):
```
sanderr, line 313: Method name does not seem accurate.

src/inmanta/data/model.py
```diff
@@ -308,9 +308,30 @@ class ResourceMinimal(BaseModel):
...
+        from inmanta.resources import Id  # prevent circular import
```
sanderr, line 314: I'm going to have a quick look to see if we can prevent this loop instead of importing locally. I'll get back to you.

sanderr (changes requested):

src/inmanta/data/model.py
sanderr: This seems to do it:
```diff
diff --git a/src/inmanta/data/model.py b/src/inmanta/data/model.py
index 5de29173..d00d67ce 100644
--- a/src/inmanta/data/model.py
+++ b/src/inmanta/data/model.py
@@ -28,7 +28,7 @@ from pydantic.fields import ModelField
 
 import inmanta
 import inmanta.ast.export as ast_export
-from inmanta import const, protocol
+from inmanta import const, protocol, resources
 from inmanta.stable_api import stable_api
 from inmanta.types import ArgumentTypes, JsonType, SimpleTypes, StrictNonIntBool
 
@@ -311,9 +311,7 @@ class ResourceMinimal(BaseModel):
     @classmethod
     @validator("id")
     def name_must_contain_space(cls, v):
-        from inmanta.resources import Id  # prevent circular import
-
-        if Id.is_resource_version_id(v):
+        if resources.Id.is_resource_version_id(v):
             return v
         raise ValueError(f"id {v} is not of type ResourceVersionIdStr")
 
@@ -325,9 +323,7 @@ class ResourceMinimal(BaseModel):
         takes a resource and return the same resource with it version incremented
         (the input resource is modified)
         """
-        from inmanta.resources import Id  # prevent circular import
-
-        old_res = Id.parse_id(self.id)
+        old_res = resources.Id.parse_id(self.id)
         new_res = Id(old_res.entity_type, old_res.agent_name, old_res.attribute, old_res.attribute_value, old_res.version + 1)
         self.id = new_res.resource_version_str()
         return self
```

sanderr: (`git apply <file-with-the-diff>`)

src/inmanta/data/model.py
FloLey: oups I'll change it, sorry

arnaudsjs (approved):

PR #4356, opened 2022-06-13
Add environment setting to set the auto-recompile-wait time instead of the config option

### LINKED ISSUE #4332: update  auto-recompile-wait
The option [auto-recompile-wait](https://docs.inmanta.com/inmanta-service-orchestrator/4/reference/config.html#server.auto-recompile-wait) is currently causing performance issues in many places.

What should be done 

- [ ] make this an environment setting
- [ ] set the default to 0

andraskvr:
# Description

closes #4332 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/4332-auto-recompile-wait-env-setting.yml
+++ b/changelogs/unreleased/4332-auto-recompile-wait-env-setting.yml
@@ -1,0 +1,7 @@
+description: Add environment setting to set the auto-recompile-wait time instead of the config option
+issue-nr: 4332
+change-type: patch
+destination-branches: [master]
+sections:
+  feature: "{{description}}",
+  deprecation-note: "The auto-recompile-wait option in the server configuration is now deprecated in favor of the environment setting with the same name"
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -1955,6 +1955,12 @@ def convert_int(value: Union[float, int, str]) -> Union[int, float]:
     return f_value
 
 
+def convert_float(value: Union[float, int, str]) -> float:
+    if isinstance(value, float):
+        return value
+    return float(value)
+
+
 def convert_agent_map(value: Dict[str, str]) -> Dict[str, str]:
     if not isinstance(value, dict):
         raise ValueError("Agent map should be a dict")
@@ -1988,7 +1994,14 @@ def convert_agent_trigger_method(value: object) -> str:
     return value
 
 
-TYPE_MAP = {"int": "integer", "bool": "boolean", "dict": "jsonb", "str": "varchar", "enum": "varchar"}
+TYPE_MAP = {
+    "int": "integer",
+    "bool": "boolean",
+    "dict": "jsonb",
+    "str": "varchar",
+    "enum": "varchar",
+    "float": "double precision",
+}
 
 AUTO_DEPLOY = "auto_deploy"
 PUSH_ON_AUTO_DEPLOY = "push_on_auto_deploy"
@@ -2009,6 +2022,7 @@ PURGE_ON_DELETE = "purge_on_delete"
 PROTECTED_ENVIRONMENT = "protected_environment"
 NOTIFICATION_RETENTION = "notification_retention"
 AVAILABLE_VERSIONS_TO_KEEP = "available_versions_to_keep"
+AUTO_RECOMPILE_WAIT = "auto_recompile_wait"
 
 
 class Setting(object):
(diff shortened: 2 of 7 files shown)


changelogs/unreleased/4332-auto-recompile-wait-env-setting.yml
```diff
@@ -0,0 +1,7 @@
+description: Add environment setting to set the auto-recompile-wait time instead of the config option
+issue-nr: 4332
+change-type: patch
```
arnaudsjs, line 3: This is a breaking change:
```suggestion
change-type: major
```

changelogs/unreleased/4332-auto-recompile-wait-env-setting.yml
```diff
@@ -0,0 +1,7 @@
...
+destination-branches: [master]
+sections:
+  feature: "{{description}}",
+  deprecation-note: "The auto-recompile-wait option in the server configuration is now deprecated in favor of the environment setting with the same name"
```
arnaudsjs, line 7: I don't think it makes much sense to have both a `feature` and `deprecation-note` entry.
```suggestion
  deprecation-note: "The auto-recompile-wait option in the server configuration is now deprecated in favor of the environment setting with the same name"
```

src/inmanta/data/__init__.py
```diff
@@ -2261,6 +2275,14 @@ def to_dto(self) -> m.Environment:
             validator=convert_int,
             doc="The number of days to retain notifications for",
         ),
+        AUTO_RECOMPILE_WAIT: Setting(
```
arnaudsjs, line 2278: Not necessarily a change request, but just a proposal: Shouldn't we take this as an opportunity to improve the name of this config option? I don't like the current name for the following reasons:
* `AUTO_RECOMPILE_WAIT` sounds like a boolean option, but it isn't.
* The `AUTO` part of `AUTO_RECOMPILE_WAIT` hints to the fact that this option only applies to compiles that were triggered automatically (by the update of a fact for example). But in reality it applies on all requested compiles.

Wouldn't it be better to call this option `RECOMPILE_BACKOFF` or something similar. That clearly conveys the purpose of the setting.

src/inmanta/server/config.py
```diff
@@ -175,15 +175,6 @@ def validate_fact_renew(value: object) -> int:
     "server", "fact-resource-block", 60, "Minimal time between subsequent requests for the same fact", is_time
 )
 
-server_autrecompile_wait = Option(
```
arnaudsjs, line 178: I am wondering whether we should make the transition to the new setting a bit more smooth. I mean:
* When the global server setting is set, use it and log a deprecation message.
* When the global server setting is not set, use the environment setting.

We can then remove the global server config setting in ISO7.

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
andraskvr: I agree, that's a better name

arnaudsjs: @wouterdb What is your opinion on this?

src/inmanta/server/services/compilerservice.py
```diff
@@ -638,7 +638,17 @@ def _calculate_recompile_wait(
         return wait
 
     async def _auto_recompile_wait(self, compile: data.Compile) -> None:
-        wait_time = opt.server_autrecompile_wait.get()
+        auto_recompile_wait = opt.server_autrecompile_wait.get()
```
arnaudsjs, line 641: I think we should invoke the `is_set()` method here: https://github.com/inmanta/inmanta-core/blob/6b07442415f8a706c5ce3310cb5e80ea714fdbca/src/inmanta/config.py#L148

src/inmanta/server/services/compilerservice.py
```diff
@@ -638,7 +638,17 @@ def _calculate_recompile_wait(
...
+        if auto_recompile_wait > 0:
+            wait_time = auto_recompile_wait
+            log_message = (
+                f"The server-auto-recompile-wait is enabled and set to {wait_time} seconds. "
+                f"This option is deprecated in favor of the recompile_backoff environment setting."
```
arnaudsjs, line 646: We should log the deprecation notice at level `warning`.

src/inmanta/data/__init__.py
```diff
@@ -2261,6 +2275,14 @@ def to_dto(self) -> m.Environment:
             validator=convert_int,
             doc="The number of days to retain notifications for",
         ),
+        RECOMPILE_BACKOFF: Setting(
+            name=RECOMPILE_BACKOFF,
+            default=0.1,
+            typ="float",
```
arnaudsjs, line 2281: For this setting only values `>=0` make sense. I think it would make sense to introduce a type `positive_float` and use that one instead.

arnaudsjs (changes requested):

wouterdb (approved):

src/inmanta/server/services/compilerservice.py
```diff
@@ -638,7 +639,18 @@ def _calculate_recompile_wait(
         return wait
 
     async def _auto_recompile_wait(self, compile: data.Compile) -> None:
-        wait_time = opt.server_autrecompile_wait.get()
+        auto_recompile_wait = opt.server_autrecompile_wait.get()
+        if Config.is_set("server", "auto-recompile-wait") and auto_recompile_wait > 0:
```
arnaudsjs, line 643: Why add the check for `auto_recompile_wait > 0`?

Issue #3470, opened 2021-11-17
Unserializable items in the 'changes' set make the CRUD handler get stuck in 'updating'


wouterdb:
When reading back a resource, if any of the fields is set to a Set, this cause the crud handler to fail, in such a way that the error is not reported (i.e. this goes straight through all surrounding safeties) 

# Problem
```
2021-11-17 20:02:05,997 ERROR    inmanta.util Unable to serialize set()
2021-11-17 20:02:05,997 ERROR    inmanta.agent.agent.hashicorp-helm-2.4.1 resource terraform::Resource[hashicorp-helm-2.4.1,id=hashicorp-helm-2.4.1-helm_release-test],v=31: An error occurred during deployment of terraform::Resource[hashicorp-helm-2.4.1,id=hashicorp-helm-2.4.1-helm_release-test],v=31 (exception: Exception('Exception during serializing log message arguments'))
Traceback (most recent call last):
  File "/home/wouter/.virtualenvs/uc2/lib64/python3.8/site-packages/inmanta/agent/handler.py", line 366, in log_msg
    json_encode(kwargs)
  File "/home/wouter/.virtualenvs/uc2/lib64/python3.8/site-packages/inmanta/protocol/common.py", line 902, in json_encode
    return json.dumps(value, default=custom_json_encoder).replace("</", "<\\/")
  File "/usr/lib64/python3.8/json/__init__.py", line 234, in dumps
    return cls(
  File "/usr/lib64/python3.8/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.8/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/home/wouter/.virtualenvs/uc2/lib64/python3.8/site-packages/inmanta/protocol/common.py", line 886, in custom_json_encoder
    return util.api_boundary_json_encoder(o)
  File "/home/wouter/.virtualenvs/uc2/lib64/python3.8/site-packages/inmanta/util.py", line 259, in api_boundary_json_encoder
    return custom_json_encoder(o)
  File "/home/wouter/.virtualenvs/uc2/lib64/python3.8/site-packages/inmanta/util.py", line 291, in custom_json_encoder
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: set() is not JSON serializable

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #4364, opened 2022-06-14
Issue/3470 unserializable items in the changes set make the crud handler get stuck in updating

Hugo-Inmanta:
# Description

Cast sets to string during JSON serialization

Part of #3470 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/3470-unserializable-items-in-the-changes-set-make-the-CRUD-handler-get-stuck-in-updating.yml
+++ b/changelogs/unreleased/3470-unserializable-items-in-the-changes-set-make-the-CRUD-handler-get-stuck-in-updating.yml
@@ -1,6 +1,7 @@
-description: Cast sets to string during JSON serialization.
+description: Cast items in the 'changes' set that aren't JSON-serializable to str.
 issue-nr: 3470
 change-type: patch
 destination-branches: [iso5, master]
 sections: {
+  bugfix: "{{description}}"
 }
--- a/changelogs/unreleased/3470-unserializable-items-in-the-changes-set-make-the-CRUD-handler-get-stuck-in-updating.yml
+++ b/changelogs/unreleased/3470-unserializable-items-in-the-changes-set-make-the-CRUD-handler-get-stuck-in-updating.yml
@@ -1,7 +1,7 @@
 description: Cast items in the 'changes' set that aren't JSON-serializable to str.
 issue-nr: 3470
 change-type: patch
-destination-branches: [iso5, master]
+destination-branches: [iso4, iso5, master]
 sections: {
   bugfix: "{{description}}"
 }
--- a/changelogs/unreleased/3470-unserializable-items-in-the-changes-set-make-the-CRUD-handler-get-stuck-in-updating.yml
+++ b/changelogs/unreleased/3470-unserializable-items-in-the-changes-set-make-the-CRUD-handler-get-stuck-in-updating.yml
@@ -1,0 +1,6 @@
+description: Cast sets to string during JSON serialization.
+issue-nr: 3470
+change-type: patch
+destination-branches: [iso5, master]
+sections: {
+}
--- a/changelogs/unreleased/3470-unserializable-items-in-the-changes-set-make-the-CRUD-handler-get-stuck-in-updating.yml
+++ b/changelogs/unreleased/3470-unserializable-items-in-the-changes-set-make-the-CRUD-handler-get-stuck-in-updating.yml
@@ -1,4 +1,4 @@
-description: Cast items in the 'changes' set that aren't JSON-serializable to str.
+description: Fix bug that fails the CRUDHandler when a changed attribute is of type set.
 issue-nr: 3470
 change-type: patch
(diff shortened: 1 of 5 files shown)


src/inmanta/util.py
```diff
@@ -310,6 +310,9 @@ def _custom_json_encoder(o: object) -> Union[ReturnTypes, "JSONSerializable"]:
         # Logs can push exceptions through RPC. Return a string representation.
         return str(o)
 
+    if isinstance(o, set):
+        return str(o)
```
wouterdb, line 314: This I think is way overbroad: it changes the serialization behavior of the entire server.

src/inmanta/agent/handler.py
```diff
@@ -909,7 +909,15 @@ def execute(self, ctx: HandlerContext, resource: resources.Resource, dry_run: Op
                     changes["purged"] = dict(desired=desired.purged, current=True)
 
             for field, values in changes.items():
-                ctx.add_change(field, desired=values["desired"], current=values["current"])
+                for status in ["desired", "current"]:
```
arnaudsjs, line 912: Wouldn't it be better to move this logic into the `AttributeStateChange` class? (I might be missing something here)

arnaudsjs (changes requested):

src/inmanta/agent/handler.py
wouterdb: or even in the add_change method, which is another codepath that needs this check.

wouterdb (changes requested):

src/inmanta/agent/handler.py
arnaudsjs: @wouterdb I don't think that just adding it to the `add_change()` method would be sufficient. The way I see it, is the following: Any code path that updates the `self._changes` field of the `HandlerContext` class should be protected against unserializable things (add_change(), add_changes(), update_changes(), etc). All these code paths use the `AttributeStateChange()` class, so adding guard there would ensure that all code paths are safe. Do you agree?

arnaudsjs:
Shouldn't we also protect the code path to the `log_msg()` method? Because that's the code path mentioned in the stack trace of the ticket.

changelogs/unreleased/3470-unserializable-items-in-the-changes-set-make-the-CRUD-handler-get-stuck-in-updating.yml
```diff
@@ -0,0 +1,7 @@
+description: Cast items in the 'changes' set that aren't JSON-serializable to str.
```
arnaudsjs, line 1: Keep in mind that changelog are intended to be read by end-user. A good changelog entry should describe what has been fixed without going in depth about the implementation details.
```suggestion
description: Fix bug that fails the CRUDHandler when a changed attribute is of type set.
```

src/inmanta/agent/handler.py
```diff
@@ -921,8 +921,8 @@ def execute(self, ctx: HandlerContext, resource: resources.Resource, dry_run: Op
                         self.delete_resource(ctx, desired)
 
                 elif not desired.purged and len(changes) > 0:
-                    ctx.debug("Calling update_resource", changes=changes)
-                    self.update_resource(ctx, dict(changes), desired)
+                    ctx.debug("Calling update_resource", changes=ctx.changes)
+                    self.update_resource(ctx, dict(ctx.changes), desired)
```
arnaudsjs, line 925: Why was this changed?

arnaudsjs (changes requested):

src/inmanta/agent/handler.py
Hugo-Inmanta: Don't we want to pass the sanitized changes to these methods ?

wouterdb: mhh, this is a delicate issue:

1. we have two mechanisms to pass on changes (via the context and via the return value). 
2. if I look at all the surrounding code, we only act on those that come from the return value
3. the server however uses the once from the context to trigger reloads and other events 
4. we correctly log what we act on 
5. the corresponding server log may contain many other things.

While the current behavior is not optimal, changing the input to update_resource is a MAJOR change to the behavior of the agent.

src/inmanta/agent/handler.py
```diff
@@ -909,7 +909,7 @@ def execute(self, ctx: HandlerContext, resource: resources.Resource, dry_run: Op
                     changes["purged"] = dict(desired=desired.purged, current=True)
 
             for field, values in changes.items():
-                ctx.add_change(field, desired=values["desired"], current=values["current"])
+                ctx.add_change(field, desired=changes[field]["desired"], current=changes[field]["current"])
```
wouterdb, line 912: Why was this changed?

wouterdb (changes requested):

src/inmanta/agent/handler.py
Hugo-Inmanta: This is an artifact from a previous iteration, reverting the change now.

PR #4400, opened 2022-06-21
issue/4388-deprecation-warning-setuptool

### LINKED ISSUE #4388: deprecation warnings in iso build
The iso build pipeline for both iso4 and iso5 show some setuptools deprecation warnings related to packaging and likely caused by auto-discovery. Most likely this means that the pipeline is trying to include directories it shouldn't in the package. If so this should be fixed by updating the respective metadata files (see inmanta/inmanta#83 and inmanta/irt#1074). If this is not the case, we should make sure to be compliant with setuptools' expectations.

FloLey:
# Description

Fixes a deprecation warning related to setuptools. The error is the following one and has occurrences in both inmanta-core and 
test_common

```
/tmp/tmpas1m3135/env/lib64/python3.9/site-packages/setuptools/command/build_py.py:153: SetuptoolsDeprecationWarning:     Installing 'inmanta_tests.data' as data is deprecated, please list it in `packages`.
    !!


    ############################
    # Package would be ignored #
    ############################
    Python recognizes 'inmanta_tests.data' as an importable package,
    but it is not listed in the `packages` configuration of setuptools.

    'inmanta_tests.data' has been automatically added to the distribution only
    because it may contain data files, but this behavior is likely to change
    in future versions of setuptools (and therefore is considered deprecated).

    Please make sure that 'inmanta_tests.data' is included as a package by using
    the `packages` configuration field or the proper discovery methods
    (for example by using `find_namespace_packages(...)`/`find_namespace:`
    instead of `find_packages(...)`/`find:`).

    You can read more about "package discovery" and "data files" on setuptools
    documentation page.


!!
```

closes #4388

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-from setuptools import setup, find_packages
+from setuptools import setup, find_namespace_packages
 from os import path
 
 requires = [
@@ -70,7 +70,7 @@ setup(
     },
     # Packaging
     package_dir={"": "src"},
-    packages=find_packages("src") + namespace_packages,
+    packages=find_namespace_packages(where="src") + namespace_packages,
     # https://www.python.org/dev/peps/pep-0561/#packaging-type-information
     package_data={"": ["misc/*", "docs/*"], "inmanta": ["py.typed"]},
     zip_safe=False,
--- a/setup.py
+++ b/setup.py
@@ -30,8 +30,6 @@ requires = [
     "typing_inspect",
 ]
 
-# Package a dummy extensions so that the namespace package for extensions is not empty
-namespace_packages = ["inmanta_ext.core"]
 
 # read the contents of your README file
 this_directory = path.abspath(path.dirname(__file__))
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ setup(
     },
     # Packaging
     package_dir={"": "src"},
-    packages=find_namespace_packages(where="src") + namespace_packages,
+    packages=find_namespace_packages(where="src"),
     # https://www.python.org/dev/peps/pep-0561/#packaging-type-information
     package_data={"": ["misc/*", "docs/*"], "inmanta": ["py.typed"]},
     zip_safe=False,
--- a/setup.py
+++ b/setup.py
@@ -70,9 +70,11 @@ setup(
     },
     # Packaging
     package_dir={"": "src"},
+    # find_namespace_packages scans all directories for packages and namespace_packages,
+    # even those without an __init__.py file where find_packages will not consider directories without
+    # the __init__.py file
     packages=find_namespace_packages(where="src"),
     # https://www.python.org/dev/peps/pep-0561/#packaging-type-information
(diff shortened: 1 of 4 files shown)


arnaudsjs (changes requested):
Setting the status to "request changes"  as such that I get a notification when the necessary changes are implemented.

sanderr (changes requested):
Same

FloLey:
I tried it with `python -m build` in `tests_common` and the warning was gone.

sanderr (approved):

arnaudsjs (changes requested):
Could you add a description to the PR describing what was changed and why. Like this it's quite hard to do a review.

setup.py
```diff
@@ -70,7 +70,7 @@
     },
     # Packaging
     package_dir={"": "src"},
-    packages=find_packages("src") + namespace_packages,
+    packages=find_namespace_packages(where="src") + namespace_packages,
```
arnaudsjs, line 73: Do we still need this `namespace_packages` with this new implementation?

wouterdb (commented):
I have no idea what this is about, as the warnings are not in the ticket

setup.py
FloLey: it seems we don't. I removed it.

setup.py
```diff
@@ -70,7 +70,7 @@
     },
     # Packaging
     package_dir={"": "src"},
-    packages=find_packages("src") + namespace_packages,
+    packages=find_namespace_packages(where="src"),
```
wouterdb, line 73: This needs to be documented why we use this here, as it is not obvious (sander had to explain)

setup.py
```diff
@@ -70,7 +70,7 @@
...
     # https://www.python.org/dev/peps/pep-0561/#packaging-type-information
     package_data={"": ["misc/*", "docs/*"], "inmanta": ["py.typed"]},
```
wouterdb, line 75: I think this can go entirely.

The first part `"": ["misc/*", "docs/*"]` doesn't do anything
The second part is a duplicate of what is in Manifest.in

wouterdb (changes requested):
I think we have to update Manifest.in to not package the tests,docs and misc at all, to make this warning go away

FloLey:
@wouterdb I'm not sure this is what you mean about the MANIFEST.in file. But just adding those excludes wasn't enough.

arnaudsjs (changes requested):
Setting status to "request changes" until Wouter his comments have been processed.

wouterdb (approved):

setup.py
```diff
@@ -70,9 +70,10 @@
     },
     # Packaging
     package_dir={"": "src"},
-    packages=find_packages("src") + namespace_packages,
```
arnaudsjs, line 73: The `namespace_packages` list is still defined in this file. As it serves no purpose anymore, we can remove it.

arnaudsjs (changes requested):

PR #4419, opened 2022-06-22
Issue/4408 cache invalidation

### LINKED ISSUE #4408: Update cache invalidation mechanism for cfc files
The current caching mechanism for `.cf` files uses the modification date of the `.cf` file and the `.cfc` file to determine whether a cache file is still valid. Since #4407 stores these cache files in the root of the inmanta project, this check is not longer sufficient. The modification timestamp of the file being cached should be stored in `.cfc` file to determine whether the `.cfc` file is still valid.

wouterdb:
# Description

More precise  invalidation for the compiler cache
closes #4408 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/parser/cache.py
+++ b/src/inmanta/parser/cache.py
@@ -28,9 +28,9 @@ LOGGER = logging.getLogger(__name__)
 
 
 class CacheEnvelope:
-    """Every cached file gets the exact modification to of the file it is caching, to have cheap, accurate invalidation"""
+    """Every cached file gets the exact modification time of the file it is caching, to have cheap, accurate invalidation"""
 
-    def __init__(self, timestamp: float, statements: List[Statement]):
+    def __init__(self, timestamp: float, statements: List[Statement]) -> None:
         self.timestamp = timestamp
         self.statements = statements
 
--- a/src/inmanta/parser/cache.py
+++ b/src/inmanta/parser/cache.py
@@ -27,7 +27,9 @@ from inmanta.util import get_compiler_version
 LOGGER = logging.getLogger(__name__)
 
 
-class CacheEnveloppe:
+class CacheEnvelope:
+    """Every cached file gets the exact modification to of the file it is caching, to have cheap, accurate invalidation"""
+
     def __init__(self, timestamp: float, statements: List[Statement]):
         self.timestamp = timestamp
         self.statements = statements
@@ -86,7 +88,7 @@ class CacheManager:
                 return None
             with open(cache_filename, "rb") as fh:
                 result = ASTUnpickler(fh, namespace).load()
-                if not isinstance(result, CacheEnveloppe):
+                if not isinstance(result, CacheEnvelope):
                     # old cache format
                     self.misses += 1
                     return None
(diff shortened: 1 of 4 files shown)


src/inmanta/parser/cache.py
```diff
@@ -27,6 +27,12 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class CacheEnveloppe:
```
sanderr, line 30:
```suggestion
class CacheEnvelope:
```
Two p's is the Dutch word.

src/inmanta/parser/cache.py
```diff
@@ -74,13 +80,22 @@ def un_cache(self, namespace: Namespace, filename: str) -> Optional[List[Stateme
             if not os.path.exists(cache_filename):
                 self.misses += 1
                 return None
+            mtime = os.path.getmtime(filename)
             if os.path.getmtime(filename) > os.path.getmtime(cache_filename):
                 self.misses += 1
                 return None
```
sanderr, line 86: Is this still relevant? I think we can just consider the envelope timestamp the authority and ignore the cache mtime altogether.

sanderr: Or is this meant as a performance boost in the most common scenario so we don't need to read the file?

tests/compiler/test_parser_cache.py
```diff
@@ -0,0 +1,61 @@
+"""
+    Copyright 2018 Inmanta
```
sanderr, line 2:
```suggestion
    Copyright 2022 Inmanta
```

tests/compiler/test_parser_cache.py
```diff
@@ -0,0 +1,61 @@
...
+
+def test_caching(snippetcompiler):
+    # reset counts
+    parser.cache_manager = CacheManager()
+    snippetcompiler.setup_for_snippet(
+        """
+a=1
+"""
+    )
+    # don't know hit count, may very on previous testcases
```
sanderr, line 34:
```suggestion
    # don't know hit count, may vary on previous testcases
```
Might not be what you meant. If not, I don't understand.

src/inmanta/parser/cache.py
wouterdb: done

src/inmanta/parser/cache.py
```diff
@@ -27,6 +27,14 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class CacheEnvelope:
+    """Every cached file gets the exact modification to of the file it is caching, to have cheap, accurate invalidation"""
```
arnaudsjs, line 31:
```suggestion
    """Every cached file gets the exact modification time of the file it is caching, to have cheap, accurate invalidation"""
```

src/inmanta/parser/cache.py
```diff
@@ -27,6 +27,14 @@
...
+
+    def __init__(self, timestamp: float, statements: List[Statement]):
```
arnaudsjs, line 33:
```suggestion
    def __init__(self, timestamp: float, statements: List[Statement]) -> None:
```

arnaudsjs (changes requested):

arnaudsjs (approved):

sanderr (approved):

PR #4425, opened 2022-06-23
[FIX] wrap database accesses in a transaction

Hugo-Inmanta:
# Description

## Problem:

The `test_compileservice_cleanup` test depends on 2 fixtures:
- `server_with_frequent_cleanups` that periodically cleans up old `Compile` and `Report` database entries.
- `old_and_new_compile_report` that inserts data in the `Compile` and `Report` databases.

The issue arises when events happen in this order:

1) Insert in `Compile`
2) Server clean up 
3) Insert in `Report`  ==> **FAILS** because `Compile` has been cleaned up

## Solution
Execute all queries in the `old_and_new_compile_report` fixture in a transaction as such that the change is applied atomically.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/server/test_compilerservice.py
+++ b/tests/server/test_compilerservice.py
@@ -820,10 +820,8 @@ async def old_and_new_compile_report(server_with_frequent_cleanups, environment_
         "handled": True,
         "version": 1,
     }
-    await Compile(**old_compile).insert()
     compile_id_new = uuid.uuid4()
     new_compile = {**old_compile, "id": compile_id_new, "requested": now, "started": now, "completed": now}
-    await Compile(**new_compile).insert()
 
     report1 = {
         "id": uuid.UUID("2baa0175-9169-40c5-a546-64d646f62da6"),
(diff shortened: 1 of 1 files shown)


tests/server/test_compilerservice.py
```diff
@@ -843,8 +841,14 @@ async def old_and_new_compile_report(server_with_frequent_cleanups, environment_
         "name": "Recompiling configuration model",
         "compile": compile_id_new,
     }
-    await Report(**report1).insert()
-    await Report(**report2).insert()
+
+    async with Compile.get_connection() as con:
```
Hugo-Inmanta, line 845: I'm not 100% sure about this line.
Is this an ok way to retrieve the connection ?

tests/server/test_compilerservice.py
```diff
@@ -843,8 +841,14 @@ async def old_and_new_compile_report(server_with_frequent_cleanups, environment_
...
+        async with con.transaction():
+            await Compile(**old_compile).insert()
+            await Compile(**new_compile).insert()
+            await Report(**report1).insert()
+            await Report(**report2).insert()
```
arnaudsjs, line 850: You need to pass on the connection object to the insert() method. Otherwise the insert() method will create a new connection, causing it to execute outside of the transaction.
```suggestion
            await Compile(**old_compile).insert(connection=con)
            await Compile(**new_compile).insert(connection=con)
            await Report(**report1).insert(connection=con)
            await Report(**report2).insert(connection=con)
```

tests/server/test_compilerservice.py
arnaudsjs: Looks good to me. I don't think we currently have a more generic way to retrieve the connection.

arnaudsjs (changes requested):

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #4429, opened 2022-06-23
Issue/build master fix inmanta core postgresql versions issue2 wrap api call in retry limited

Hugo-Inmanta:
# Description

Wrap `get_resources_for_agent` API call with a retry_limited so that the test no longer fails if the agent didn't obtain a session yet by the time the call is made.

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/build-master-fix-inmanta-core-postgresql-versions-issue2-wrap-API-call-in-retry-limited.yml
+++ b/changelogs/unreleased/build-master-fix-inmanta-core-postgresql-versions-issue2-wrap-API-call-in-retry-limited.yml
@@ -1,4 +1,3 @@
 description: Fix race condition in test_get_resource_for_agent.
 change-type: patch
 destination-branches: [master, iso5, iso4]
-sections:


tests/test_server.py
```diff
@@ -384,9 +384,12 @@ async def test_get_resource_for_agent(server_multi, client_multi, environment_mu
     assert result.result["model"]["released"]
     assert result.result["model"]["result"] == "deploying"
 
-    result = await aclient.get_resources_for_agent(environment_multi, "vm1.dev.inmanta.com")
-    assert result.code == 200
-    assert len(result.result["resources"]) == 3
+    async def wait_for_session() -> bool:
+        result = await aclient.get_resources_for_agent(environment_multi, "vm1.dev.inmanta.com")
+        assert len(result.result["resources"]) == 3
+        return result.code == 200
```
arnaudsjs, line 390: We want to cover the scenario where a 409 status code is returned, so we have to include it in the returned value.
```suggestion
        return result.code == 200 and len(result.result["resources"]) == 3
```

arnaudsjs (changes requested):

tests/test_server.py
```diff
@@ -384,9 +384,12 @@ async def test_get_resource_for_agent(server_multi, client_multi, environment_mu
     assert result.result["model"]["released"]
     assert result.result["model"]["result"] == "deploying"
 
-    result = await aclient.get_resources_for_agent(environment_multi, "vm1.dev.inmanta.com")
-    assert result.code == 200
-    assert len(result.result["resources"]) == 3
+    async def wait_for_session() -> bool:
+        result = await aclient.get_resources_for_agent(environment_multi, "vm1.dev.inmanta.com")
```
arnaudsjs, line 388: Maybe also add a comment saying why this retry-limited call is required

arnaudsjs (approved):

PR #4434, opened 2022-06-23
improved transparency of get_connection in inmanta.data

sanderr:
# Description

Suggestion for improvement of the `get_connection` method in `inmanta.data.BaseDocument`. This class is part of the stable API and I changed the method signature in the sense that it can now only be used with a context manager where previously it could also be just awaited. But I'm pretty sure that this method isn't used outside of core, and if it is I expect it's used as a context manager.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -29,6 +29,7 @@ import warnings
 from abc import ABC, abstractmethod
 from collections import abc, defaultdict
 from configparser import RawConfigParser
+from contextlib import AbstractAsyncContextManager
 from itertools import chain
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


src/inmanta/data/__init__.py
```diff
@@ -3064,8 +3054,8 @@ async def persist_on_resume(cls, env: uuid.UUID, connection: Optional[asyncpg.co
         Restores default halted state. Returns a list of agents that should be unpaused.
         """
 
-        async def query_with_connection(connection: asyncpg.connection.Connection) -> List[str]:
-            async with connection.transaction():
+        async with cls.get_connection(connection) as con:
+            async with con.transaction():
```
arnaudsjs, line 3058: We should use `con` instead of `connection` in the transaction block.

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
sanderr: Oh, thanks, good catch.

sanderr: Looks like I already fixed this yesterday.

arnaudsjs: You didn't push this yet, I guess?

sanderr: Oh, I misunderstood, I fixed the `connection.transaction()` -> `con.transaction()` but I didn't fix the body... I fixed it now.

arnaudsjs (approved):

PR #4464, opened 2022-06-27
Add byte compile option

bartv:
# Description

This PR adds a bytecompile option to the build command. This option will do two things:

- byte compile all python and remove all python source
- include a dummy so created with cython to force an abi specifc wheel

Closes inmanta/irt#1190

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


bartv:
Do we need something in the release notes?

arnaudsjs (changes requested):

arnaudsjs:
> Do we need something in the release notes?

I am not sure whether this feature will be useful for end-users, but I also don't see a reason why we wouldn't mention it in the release notes.

src/inmanta/moduletool.py
```diff
@@ -901,23 +910,71 @@ def __init__(self, module_path: str) -> None:
...
+            for current_file in [f for f in files if f.endswith(".py")]:
+                py_file = os.path.join(root, current_file)
+                pyc_file = f"{py_file}c"
+                py_compile.compile(file=py_file, cfile=pyc_file, doraise=True)
+                os.remove(py_file)
+
+        # Make sure there is a pyc line in the manifest.in
+        with open(os.path.join(build_path, "MANIFEST.in"), "r+", encoding="utf-8") as fh:
+            content = fh.read()
+            # Make sure that there is at least one .pyc include
```
sanderr, line 948: Could you elaborate on this? It really needs to be `*.pyc`, doesn't it?

src/inmanta/moduletool.py
```diff
@@ -901,23 +910,71 @@ def __init__(self, module_path: str) -> None:
...
+            if ".pyc" not in content:
```
sanderr, line 949: This could also be "global-exclude *.pyc`, but I don't see how we can reliably handle all such edge cases without overcomplicating this.

src/inmanta/moduletool.py
```diff
@@ -901,23 +910,71 @@ def __init__(self, module_path: str) -> None:
...
+                # Add it to the manifest. The read has moved the pointer to the end of the file
+                LOGGER.info("Adding .pyc include line to the MANIFEST")
+                fh.write("\nrecursive-include inmanta_plugins *.pyc\n")
```
sanderr, line 952:
```suggestion
                fh.write("\nrecursive-include inmanta_plugins *.pyc\n")
                fh.write("\nglobal-exclude */__pycache__/*\n")
```

src/inmanta/moduletool.py
sanderr: Unless we just write it regardless? This is only the builder so nothing we write is actually persisted.

src/inmanta/moduletool.py
```diff
@@ -901,23 +910,71 @@ def __init__(self, module_path: str) -> None:
...
+
+        # For wheel to build a non universal python wheel we need to trick it into thinking it contains
+        # compiled extensions against a specific python ABI. If not we might install this wheel on a python with incompatible
+        # python python code
+        dummy_file = os.path.join(build_path, "inmanta_plugins", self._module.name, "_cdummy.c")
+        with open(dummy_file, "w") as fd:
+            fd.write("// Dummy python file")
+
+        with open(os.path.join(build_path, "setup.py"), "w") as fd:
```
sanderr, line 961: Should we document/warn that `setup.py` is overwritten?

src/inmanta/moduletool.py
```diff
@@ -901,23 +910,71 @@ def __init__(self, module_path: str) -> None:
...
+            fd.write(
+                f"""
+from distutils.core import setup
+from Cython.Build import cythonize
+setup(name="{self._module.name}",
```
sanderr, line 966: Isn't `ModuleV2.name` the inmanta module name? I think you might want `ModuleV2Source.get_package_name_for(self_module.name)`.

src/inmanta/moduletool.py
```diff
@@ -1074,7 +1131,10 @@ def _do_update(self, output_directory: str, setup_cfg: ConfigParser, warn_on_mer
                 + "\n"
             )
 
-    def get_pyproject(self, in_folder: str, warn_on_merge: bool = False) -> str:
+    @classmethod
+    def get_pyproject(
+        cls, in_folder: str, warn_on_merge: bool = False, build_requires: list[str] = ["setuptools", "wheel"]
```
sanderr, line 1136:
```suggestion
        cls, in_folder: str, warn_on_merge: bool = False, build_requires: Optional[list[str]] = None
```
Nice with the built-in `list`!

sanderr (changes requested):
Pretty cool! Couple of concerns about some details.

src/inmanta/moduletool.py
```diff
@@ -507,6 +508,14 @@ def modules_parser_config(cls, parser: ArgumentParser) -> None:
             default=None,
             dest="output_dir",
         )
+        build.add_argument(
+            "-b",
+            "--byte-code",
+            help="Produce a module wheel that only contains bytecode.",
```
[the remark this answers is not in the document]
bartv: I updated the wording

src/inmanta/moduletool.py
bartv: I first checked for *.pyc but, if someone want to limit what is included they will write "something.pyc". In that case we should not add a *.pyc anymore.

So the reasoning is that if there is something with .pyc in there, the module maintainer knows what they are doing.

src/inmanta/moduletool.py
bartv: see also the previous comment, then they know what they are doing :-)

src/inmanta/moduletool.py
bartv: done

src/inmanta/moduletool.py
bartv: but then we break the api. Should I set the default inside the method then?

src/inmanta/moduletool.py
bartv: done

src/inmanta/moduletool.py
```diff
@@ -511,7 +511,7 @@ def modules_parser_config(cls, parser: ArgumentParser) -> None:
         build.add_argument(
             "-b",
             "--byte-code",
-            help="Produce a module wheel that only contains bytecode.",
+            help="Produce a module wheel that only contains only python bytecode for the plugins.",
```
sanderr, line 514:
```suggestion
            help="Produce a module wheel that contains only python bytecode for the plugins.",
```

src/inmanta/moduletool.py
sanderr: I see, I guess that makes sense.

src/inmanta/moduletool.py
sanderr: I'd like to go back on this. Arnaud just pointed out to me that all this happens in a temporary directory so we don't actually change anything in the project directory.

src/inmanta/module.py
```diff
@@ -2385,6 +2385,16 @@ def get_plugin_dir(self) -> Optional[str]:
         """
         raise NotImplementedError()
 
+    def _list_python_files(self, plugin_dir: str) -> Generator[None, str, None]:
```
sanderr, line 2388:
```suggestion
    def _list_python_files(self, plugin_dir: str) -> Iterator[str]:
```
No strong requirement for me but I find the `Iterator` type definition a lot more readable than the `Generator` one. And since this is actually an iterator we could just as well be specific.

src/inmanta/plugins.py
```diff
@@ -213,7 +213,12 @@ def __init__(self, namespace: Namespace) -> None:
 
         filename: Optional[str] = inspect.getsourcefile(self.__class__.__function__)
         assert filename is not None
-        line: int = inspect.getsourcelines(self.__class__.__function__)[1] + 1
+        try:
+            line: int = inspect.getsourcelines(self.__class__.__function__)[1] + 1
+        except OSError:
+            # In case of bytecompiled code there is no source line
+            line = 1
```
sanderr, line 220:
```suggestion
            line = -1
```
I'm not completely sure, this might break tests, but I believe we already use `-1` in some places where line number does not make sense, e.g. [here](https://github.com/inmanta/inmanta-core/blob/5eb1419d2e142c8f9bb8ba2a28721801c94711dd/src/inmanta/module.py#L1648).

sanderr (changes requested):
Do I understand correctly you would like someone to have a look at this to make compilation and agent code loading work before you merge it?

src/inmanta/plugins.py
bartv: That does not work. It excepts here: https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/ast/__init__.py#L155

src/inmanta/module.py
bartv: ah, is that correct? did not know that

bartv:
> Do I understand correctly you would like someone to have a look at this to make compilation and agent code loading work before you merge it?

I am still trying myself. However, I have to be realistic on how much time I can invest. So at some point I might need to hand it over to someone else.

src/inmanta/loader.py
```diff
@@ -167,9 +167,16 @@ def get_types(self) -> Iterable[Tuple[str, List[SourceInfo]]]:
 @dataclass(frozen=True)
 class ModuleSource:
     name: str
-    source: str
+    source: bytes
     hash_value: str
 
+    def is_bytecode(self) -> bool:
+        try:
+            self.source.decode("utf-8")
```
wouterdb, line 175: I was thinking, would it not be easier to use the magic number to recognize the file. 

It turns out it is now, because the magic changes often. 

https://github.com/python/cpython/blob/main/Lib/importlib/_bootstrap_external.py#L233

Perhaps add a comment here to prevent this same investigation from being done again.

src/inmanta/module.py
```diff
@@ -2401,6 +2401,15 @@ def get_plugin_dir(self) -> Optional[str]:
         """
         raise NotImplementedError()
 
+    def _list_python_files(self, plugin_dir: str) -> Iterator[str]:
+        """Generate a list of all python files"""
+        for file_name in glob.iglob(os.path.join(plugin_dir, "**", "*.py"), recursive=True):
+            yield file_name
+
+        for file_name in glob.iglob(os.path.join(plugin_dir, "**", "*.pyc"), recursive=True):
+            if "__pycache__" not in file_name:
```
wouterdb, line 2410: This could do with a comment, I think.

src/inmanta/module.py
```diff
@@ -2401,6 +2401,15 @@ def get_plugin_dir(self) -> Optional[str]:
         """
         raise NotImplementedError()
 
+    def _list_python_files(self, plugin_dir: str) -> Iterator[str]:
```
wouterdb, line 2404: What if both the pyc and py file are present? 

I think this can happen by accident and should be tested

wouterdb (changes requested):

src/inmanta/module.py
bartv: pyc has precedence over py and pyc is invalidated based on timestamp.

Normally pyc files are always created in a __pycache__, so this will only happen when created without our build command.

What else would you like to test?

src/inmanta/server/services/codeservice.py
```diff
@@ -125,7 +104,6 @@ async def get_code(self, env: data.Environment, code_id: int, resource: str) ->
         sources = {}
         if code.source_refs is not None:
             for code_hash, (file_name, module, req) in code.source_refs.items():
-                content = self.file_slice.get_file_internal(code_hash)
-                sources[code_hash] = (file_name, module, content.decode(), req)
+                sources[code_hash] = (file_name, module, "", req)
```
sanderr, line 107: I think this could use a comment. We just don't use the column anymore but we keep it for backwards compatibility? Perhaps we should also add this to the method docstring?

src/inmanta/module.py
sanderr: So suppose a module v1.0.0 is byte-compiled but v2.0.0 isn't (or the other way around), both the compiler and the agent code loader should load the latest version as expected?

src/inmanta/loader.py
```diff
@@ -230,6 +237,11 @@ def install_source(self, module_source: ModuleSource) -> bool:
                 os.path.join(all_modules_dir, pathlib.PurePath(pathlib.PurePath(relative_module_path).parts[0]))
             )
 
+            if module_source.is_bytecode():
+                init_file = "__init__.pyc"
+            else:
+                init_file = "__init__.py"
+
```
sanderr, line 244: I'm a bit concerned about this. What if a module contains only one byte compiled file. e.g.
```
inmanta_plugins/mymod/__init__.py  # contains some plugins
inmanta_plugins/mymod/resources.pyc
```
If `install_source` is called in this order, then first `__init__.py` will be written, followed by `resources.pyc`. The loading of the latter will create `__init__.pyc`, which will be an empty file that gets precendence over the actual `__init__.py` file.

sanderr: We could update this logic but we might just as well consider not doing anything if it's a byte-compiled file, because `touch_inits` is only relevant for orchestrator versions of the modules v1 era anyway. At least, that's how it's documented.

sanderr: A second remark here is that I don't think an empty `pyc` file is allowed. I get
```
  File "/home/sander/downloads/bytecompiletest/main.py", line 1, in <module>
    import mymod
ImportError: bad magic number in 'mymod': b''
```

So if we would keep this `touch_inits`, we probably need to byte-compile the empty string first.

sanderr (changes requested):

src/inmanta/agent/agent.py
```diff
@@ -1204,12 +1205,17 @@ async def ensure_code(self, environment: uuid.UUID, version: int, resource_types
...
-                        await self._install(
-                            [
-                                (ModuleSource(name, content, hash_value), requires)
-                                for hash_value, (path, name, content, requires) in result.result["sources"].items()
-                            ]
-                        )
+                        modules = []
+                        for hash_value, (path, name, _, requires) in result.result["sources"].items():
+                            # We do not use the provided source because this might break any binary or non-utf8 content
+                            response: protocol.Result = await self._client.get_file(hash_value)
```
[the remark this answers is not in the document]
bartv: That is also dangerous because we might overload the server with that

src/inmanta/server/services/codeservice.py
bartv: I keep it here to make sure that we do not accidentally break code or tests. The code that actually uses this, the agent code loading, has been updated.

src/inmanta/loader.py
bartv: Fixed now

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
wouterdb: if we don't check if the file already exists on disk, I think this is a step backwards. 

We still download all the files every time (same amount of data*4/3 for the base 64), but we do it in multiple calls (more chatter)

There are two opportunities here:
1. make a raw, binary get_file api (no base64)
2. first check disk, then download. 

We could ticketize these.

src/inmanta/module.py
wouterdb: By where is this filtering done? 

if I look around: 
1. on the exporter code path, I think both the pyc and py files are sent to the server
2. on the loading path, this will cause duplicate loading in 'load_plugins' I think


I also think that if both the py and pyc file are sent over, but the mtime of the files are not also sent, the pyc file will always be recycled. (as per @arnaudsjs research in this matter, iirc, the mtime of the py file is inside the pyc file)

So I would indeed only use the pyc if there is no py file

src/inmanta/server/services/codeservice.py
wouterdb: but it is in fact a breaking change that will break older agents silently? 

I wonder if it would not be better to make it very obvious that this is broken? 

or even make a new API endpoint, and leave this one for backward compatibility and get rid of it in iso6?

wouterdb (changes requested):

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
bartv: yes, that is what I suggested in the standup at some point

src/inmanta/module.py
sanderr: I'm not completely following. Suppose the yang module is used on an orchestrator somewhere right now. When this is merged we create a byte-compiled artifact for it, that also contains some new functionality. How do we make sure this new functionality will be available if we prefer the `.py` files? There are no `.py` files anymore for the new version, but the ones from the old version are not cleaned up.

bartv: As it is implemented now, we will select a loader that loads either pyc or py depending on what the server returned. So that means that if v+1 switches to python it will use the python code.

src/inmanta/module.py
bartv: done

src/inmanta/module.py
bartv: done + tests

src/inmanta/loader.py
```diff
@@ -250,14 +252,19 @@ def touch_inits(directory: str) -> None:
...
                 touch_inits(os.path.dirname(normdir))
 
             # ensure correct package structure
             os.makedirs(module_dir, exist_ok=True)
             touch_inits(os.path.dirname(module_dir))
             source_file = os.path.join(module_dir, init_file)
 
+            if os.path.exists(os.path.join(module_dir, "__init__.py")) and source_file[-4:] == ".pyc":
+                # we seem to have create an __init__.py before, we now have a pyc file so we should remove it before
+                # creating the pyc to keep the structure clean
```
sanderr, line 265: Nice, this does indeed resolve my concerns in this direction. I see how the finder prefers `.py` over `.pyc`, which resolves the concern in the other direction. But wouldn't it be safer to just do the opposite here as well? Then we can never have both files simultaneously, and no assumptions are required between the finder and the loader.

src/inmanta/module.py
```diff
@@ -2438,14 +2438,21 @@ def get_plugin_dir(self) -> Optional[str]:
...
+        files: Dict[str, str] = {}
 
         for file_name in glob.iglob(os.path.join(plugin_dir, "**", "*.pyc"), recursive=True):
+            # Filter out pyc files in the default cache dir. Only support our compiled pyc files.
             if "__pycache__" not in file_name:
-                yield file_name
+                files[file_name[:-3]] = file_name
+
+        for file_name in glob.iglob(os.path.join(plugin_dir, "**", "*.py"), recursive=True):
+            # store the python source file if we do not have a python file
```
sanderr, line 2451: Any reason for this ordering? It seems inconsistent with the finder's behavior. This way if both files exist the compiler would use the `.py` file but we'd upload the `.pyc` file to the agent.

sanderr (changes requested):

src/inmanta/loader.py
sanderr: This comment is not wholly accurate: the finder is not used on the agent iirc, it's just native Python. But I still think it might be best to play safe.

src/inmanta/module.py
sanderr: I don't, that would be fine by me.

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
wouterdb: @bartv I completed the nr2, could you link the tickets?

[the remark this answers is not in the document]
bartv: I never created tickets for this. I started working on them to fix them here.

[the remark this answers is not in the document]
bartv: ah no, the binary get api does not have a ticket yet and no work was done on it

src/inmanta/server/services/codeservice.py
wouterdb: @bartv I changed this to raise a 400 when the source is not valid utf8

src/inmanta/data/model.py
```diff
@@ -661,3 +661,12 @@ class Notification(BaseModel):
     uri: str
     read: bool
     cleared: bool
+
+
+class Source(BaseModel):
+    """Model for source code"""
+
+    hash: str
+    file_name: str
```
wouterdb, line 670: @bartv I wonder if it would not make more sense to not serve the file_name at all and instead have boolean field 'is_compiled' 

We don't use the file_name anyways.

bartv: fine for me

src/inmanta/server/services/codeservice.py
bartv: that is in the case it is byte compiled?

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
wouterdb: https://github.com/inmanta/inmanta-core/issues/4658

src/inmanta/server/services/codeservice.py
wouterdb: or any other binary file, yes 

I added it to the testcase

src/inmanta/data/model.py
wouterdb: done

src/inmanta/module.py
wouterdb: @sanderr @bartv I don't follow this entirely, does this require further action?

bartv: This was already handled by me

sanderr: I think so. The issue here is that there is an inconsistency between this method and the finder if both a `.py` and a `.pyc` file are present. Here we prefer the `.pyc` file while the finder does the opposite (unless Bart changed that already, I haven't checked). Bart proposed to keep the behavior here and change the behavior of the finder to match this. I'm on board with that.

sanderr: Ah, nevermind then.

wouterdb (approved):

src/inmanta/server/services/codeservice.py
```diff
@@ -106,6 +107,29 @@ async def get_code(self, env: data.Environment, code_id: int, resource: str) ->
             for code_hash, (file_name, module, req) in code.source_refs.items():
                 # To support non utf-8 encoded files, the agent uses the file API to load the source code. For backwards
                 # compatability the method returns and empty string.
```
sanderr, line 109: This is no longer accurate.

sanderr (changes requested):

src/inmanta/server/services/codeservice.py
wouterdb: removed

src/inmanta/protocol/endpoints.py
```diff
@@ -392,10 +393,36 @@ class SyncClient(object):
...
+        they can not be used at the same time
+
+        :param name: name of the configuration to use for this endpoint. The config section used is "{name}_rest_transport"
+        :param client: the client to use for this sync_client
+        :param timeout: http timeout on all requests
+
+        :param ioloop: the specific (running) ioloop to schedule this request on.
+        if no ioloop is passed,we assume there is no running ioloop in the context where this syncclient is used.
+        """
+        if name is None and client is None:
```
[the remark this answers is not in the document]
wouterdb: fixed

src/inmanta/loader.py
```diff
@@ -239,10 +260,12 @@ def install_source(self, module_source: ModuleSource) -> bool:
...
 
-            if module_source.is_bytecode():
+            if module_source.is_byte_code:
                 init_file = "__init__.pyc"
+                alternate_file = "__init__.py"
             else:
                 init_file = "__init__.py"
+                alternate_file = "__init__.pyc"
 
             def touch_inits(directory: str) -> None:
```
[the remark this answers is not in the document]
wouterdb: fixed

sanderr (approved):

PR #4467, opened 2022-06-28
dropped dependency on requests[chardet]

sanderr:
# Description

With the context given in https://github.com/psf/requests/issues/6177#issuecomment-1168148782, I think it should be safe to drop this requirement again. Do you agree?

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -398,7 +398,7 @@ caused by:
     def exec(*cmd):
         process = do_run([sys.executable, "-m", "inmanta.app"] + list(cmd), cwd=snippetcompiler.project_dir)
         _, err = process.communicate(timeout=30)
-        assert err.decode() == output
+        assert output in err.decode()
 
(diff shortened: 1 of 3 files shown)


tests/test_app.py
```diff
@@ -494,5 +493,4 @@ def test_init_project(tmpdir):
     assert os.path.exists(test_project_path)
     (stdout, stderr, return_code) = run_without_tty(args, killtime=15, termtime=10)
     assert return_code != 0
-    assert len(stderr) == 1
-    assert "already exists" in stderr[0]
+    assert any(["already exists" in error for error in stderr])
```
arnaudsjs, line 496: With these brackets it looks weird to me.
```suggestion
    assert any("already exists" in error for error in stderr)
```

arnaudsjs (changes requested):

arnaudsjs (approved):

sanderr:
Merged in iso4 in 03997ced

PR #4468, opened 2022-06-28
Added support for development builds of V2 modules

arnaudsjs:
# Description

This PR adds support to create development builds of V2 modules. 

Part of inmanta/irt#1184

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/moduletool.py
+++ b/src/inmanta/moduletool.py
@@ -17,6 +17,7 @@
 """
 import argparse
 import configparser
+import datetime
 import inspect
 import logging
 import os
@@ -507,6 +508,14 @@ mode.
             default=None,
             dest="output_dir",
         )
+        build.add_argument(
+            "--dev",
+            dest="dev_build",
+            help="Perform a development build of the module. This adds the build tag `.dev<timestamp>` to the "
+            "package name. The timestamp has the form %%Y%%m%%d%%H%%M%%S.",
+            default=False,
+            action="store_true",
+        )
 
         subparser.add_parser("v1tov2", help="Convert a V1 module to a V2 module in place")
 
@@ -557,7 +566,7 @@ mode.
             raise ModuleVersionException(f"Expected a v1 module, but found v{module.GENERATION.value} module")
         ModuleConverter(module).convert_in_place()
 
-    def build(self, path: Optional[str] = None, output_dir: Optional[str] = None) -> str:
+    def build(self, path: Optional[str] = None, output_dir: Optional[str] = None, dev_build: bool = False) -> str:
         """
         Build a v2 module and return the path to the build artifact.
         """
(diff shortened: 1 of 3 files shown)


src/inmanta/moduletool.py
```diff
@@ -507,6 +508,14 @@ def modules_parser_config(cls, parser: ArgumentParser) -> None:
...
             dest="output_dir",
         )
+        build.add_argument(
+            "--dev",
+            dest="dev_build",
+            help="Perform a development build of the module. This adds the build tag `.dev<timestamp>` to the "
+            "package name. The timestamp has the form %%Y%%m%%d%%H%%M%%S.",
+            default=False,
+            action="store_true",
+        )
```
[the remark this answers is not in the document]
arnaudsjs: * This `--dev` options is indeed not used at the moment. The idea is to use it in https://github.com/inmanta/irt/issues/1186 for modules for which we don't create a release tag on every commit. Next to that it will be useful for end users to have the possibility to create dev releases.
* The parsing of the arguments happens here https://github.com/inmanta/inmanta-core/blob/01f957fa5be0df0ba6d67d4094837c5e0f9b82fe/src/inmanta/app.py#L765

src/inmanta/moduletool.py
```diff
@@ -507,6 +508,14 @@ def modules_parser_config(cls, parser: ArgumentParser) -> None:
             default=None,
             dest="output_dir",
         )
+        build.add_argument(
+            "--dev",
+            dest="dev_build",
+            help="Perform a development build of the module. This adds the build tag `.dev<timestamp>` to the "
+            "package name. The timestamp has the form %%Y%%m%%d%%H%%M%%S.",
```
sanderr, line 515: Out of curiosity, how does this work with the double percent signs? I got the impressino this is only required if you actually use the `%` operator on it. But the help string actually shows the expected output. Does `ArgumentParser` use `%`?
```python
>>> "%%s%s" % "test"
'%stest'
>>> "%stest"
'%stest'
>>> "%%stest"
'%%stest'
```

src/inmanta/moduletool.py
```diff
@@ -912,12 +921,32 @@ def build(self, output_directory: str) -> str:
             # Copy module to temporary directory to perform the build
             build_path = os.path.join(tmpdir, "module")
             shutil.copytree(self._module.path, build_path)
+            if dev_build:
+                self._add_dev_build_tag_to_setup_cfg(build_path)
```
sanderr, line 925: If this is a build on a v2 module, doesn't this persist the change? Is there no `-Db my_tag` alternative for `setup.cfg` based builds?

sanderr (approved):

src/inmanta/moduletool.py
arnaudsjs: Yes. The `%` character is used by argparse to reference for example the default value of the option in the help string. If you want to use a literal `%` character, it has to be escaped with a double `%` character.

src/inmanta/moduletool.py
arnaudsjs: No. The module is first copied into a temporary build directory. Only the `setup.cfg` from this temporary directory is adjusted. The original one is left untouched.

PR #4486, opened 2022-06-29
Issue/4273 partial compile compilerservice

### LINKED ISSUE #4273: Support incremental compiles via the CompilerService
Tasks:

- [ ] Add a method to the `CompilerService` to request an incremental compile.

FloLey:
# Description

 Add a method to the CompilerService run an partial compile.
 Add support to run a partial compile that removes a resource set.

closes #4273 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/inmanta/db/versions/v202229060.py
```diff
@@ -0,0 +1,32 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    Contact: code@inmanta.com
+"""
+
+from asyncpg import Connection
+
+DISABLED = False
```
sanderr, line 21:
```suggestion
```

I didn't think we included this anymore in the latest versions. It's redundant since our new versioning scheme and it just adds clutter. If we take it out of this one, it will probably persist (since the next one will likely be copied from this).

src/inmanta/data/__init__.py
```diff
@@ -3346,6 +3348,9 @@ class Compile(BaseDocument):
 
     compile_data: Optional[JsonType] = None
 
+    partial: bool = False
+    removed_resource_sets: Optional[List[str]] = None
```
sanderr, line 3352: Out of curiosity, did you try to use the built-in `list`? I hadn't considered this before but it wouldn't surprise me if our data framework (and our typed methods) isn't compatible with that. It would be a good thing to be aware of.

sanderr, line 3352: I think you will also need to add these fields to the corresponding class in `inmanta.data.model`, and potentially update the `to_dao` and/or `to_dto` methods.

src/inmanta/server/services/compilerservice.py
```diff
@@ -376,6 +376,14 @@ async def setup() -> AsyncIterator[Awaitable[Optional[data.Report]]]:
                 compile_data_json_file.name,
             ]
 
+            if self.request.partial:
+                cmd.append("-p")
```
sanderr, line 380:
```suggestion
                cmd.append("--partial")
```
I think we only support a long option, don't we?

sanderr (changes requested):
Intermediary release of my comments. Still need to look at the tests.

src/inmanta/data/__init__.py
FloLey: I did not. Should I try it? :p

tests/server/test_compilerservice.py
```diff
@@ -525,6 +525,70 @@ def assert_report(uid):
...
+    )
+
+    async def wait_for_report():
+        report = await client.get_report(compile_id)
+        return report.code == 200
+
+    await retry_limited(wait_for_report, 10)
+    report = await client.get_report(compile_id)
+    assert report.result["report"]["partial"]
+    assert report.result["report"]["removed_resource_sets"] == []
```
sanderr, line 575: This might not really test what you think it does. `report.result["report"]` just contains the data from the compile request (see [here](https://github.com/inmanta/inmanta-core/blob/fb8fdc340c2a79f315c6d05da54ebd5e93569500/src/inmanta/data/__init__.py#L3367)). It would be nice if we could test somehow if a partial export was actually run.

sanderr (changes requested):

src/inmanta/data/__init__.py
sanderr: If you feel like it, otherwise I'll give it a go myself sometime.

FloLey: mypy doesn't complain. I'll keep `list`

FloLey: Ho, that we'll see when I push then

tests/db/dumps/v202205250.sql
```diff
@@ -444,8 +444,8 @@ COPY public.dryrun (id, environment, model, date, total, todo, resources) FROM s
...
-eeaabbf0-029d-40df-ace8-885562378ae5	dev-2	c80e1f37-771d-4a78-8e61-4eca4807f709			{}	0	f
-527be28f-0f16-40b6-bb6d-8cc03b1bbca9	dev-1	c80e1f37-771d-4a78-8e61-4eca4807f709			{"auto_deploy": true, "server_compile": true, "purge_on_delete": false, "autostart_agent_map": {"internal": "local:", "localhost": "local:"}, "push_on_auto_deploy": true, "autostart_agent_deploy_interval": 0, "autostart_agent_repair_interval": 600, "autostart_agent_deploy_splay_time": 0, "autostart_agent_repair_splay_time": 0, "agent_trigger_method_on_auto_deploy": "push_incremental_deploy"}	2	f
+eeaabbf0-029d-40df-ace8-885562378ae5	dev-2	c80e1f37-771d-4a78-8e61-4eca4807f709			{}	0	f		
+527be28f-0f16-40b6-bb6d-8cc03b1bbca9	dev-1	c80e1f37-771d-4a78-8e61-4eca4807f709			{"auto_deploy": true, "server_compile": true, "purge_on_delete": false, "autostart_agent_map": {"internal": "local:", "localhost": "local:"}, "push_on_auto_deploy": true, "autostart_agent_deploy_interval": 0, "autostart_agent_repair_interval": 600, "autostart_agent_deploy_splay_time": 0, "autostart_agent_repair_splay_time": 0, "agent_trigger_method_on_auto_deploy": "push_incremental_deploy"}	2	f
```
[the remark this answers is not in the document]
sanderr: #4527 should prevent this happening in the future.

src/inmanta/server/services/compilerservice.py
sanderr: Hmm, then I think we diverged from the document, I don't see `-p` in the code except for the profiler, which might actually be the reason we couldn't use it for this.

sanderr: Though I think the profiler is an option of the `inmanta` command, not `inmanta compile`.

arnaudsjs: The `-p` option was already taken to enable profiling. I updated the tickets, but I didn't update the design document.

src/inmanta/server/services/compilerservice.py
```diff
@@ -547,12 +555,16 @@ async def request_recompile(
         remote_id: uuid.UUID,
         metadata: Optional[JsonType] = None,
         env_vars: Optional[Mapping[str, str]] = None,
+        partial: bool = False,
+        removed_resource_sets: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
arnaudsjs: I am not convinced this is a good idea. We currently don't have a use case for this and I don't think it's a good idea to implement something we don't really need. Once this is part of the API, it will be very hard to remove. @sanderr Could you elaborate on why you think it would be good to have support for this?

tests/server/test_compilerservice.py
```diff
@@ -525,6 +525,70 @@ def assert_report(uid):
...
+    with open(os.path.join(project_dir, "main.cf"), "w", encoding="utf-8") as fd:
+        fd.write("")
+    env = await data.Environment.get_by_id(environment)
+    compilerslice: CompilerService = server.get_slice(SLICE_COMPILER)
+    remote_id1 = uuid.uuid4()
+
+    # Do a compile
+    compile_id, _ = await compilerslice.request_recompile(env, force_update=True, do_export=False, remote_id=remote_id1)
+
+    async def wait_for_report():
```
arnaudsjs, line 554:
```suggestion
    async def wait_for_report() -> bool:
```

tests/server/test_compilerservice.py
```diff
@@ -525,6 +525,70 @@ def assert_report(uid):
...
+
+    # add main.cf
+    with open(os.path.join(project_dir, "main.cf"), "w", encoding="utf-8") as fd:
+        fd.write("")
+    env = await data.Environment.get_by_id(environment)
+    compilerslice: CompilerService = server.get_slice(SLICE_COMPILER)
+    remote_id1 = uuid.uuid4()
+
+    # Do a compile
+    compile_id, _ = await compilerslice.request_recompile(env, force_update=True, do_export=False, remote_id=remote_id1)
```
arnaudsjs, line 552: A `force_update` would only make the test case slower I think:
```suggestion
    compile_id, _ = await compilerslice.request_recompile(env, force_update=False, do_export=False, remote_id=remote_id1)
```

tests/server/test_compilerservice.py
```diff
@@ -525,6 +525,70 @@ def assert_report(uid):
...
+        return report.code == 200
+
+    await retry_limited(wait_for_report, 10)
+    report = await client.get_report(compile_id)
+    assert not report.result["report"]["partial"]
+    assert report.result["report"]["removed_resource_sets"] == []
+
+    # Do a partial compile
+    compile_id, _ = await compilerslice.request_recompile(
+        env, force_update=True, do_export=False, remote_id=remote_id1, partial=True
```
arnaudsjs, line 565:
```suggestion
        env, force_update=False, do_export=False, remote_id=remote_id1, partial=True
```

tests/server/test_compilerservice.py
```diff
@@ -525,6 +525,70 @@ def assert_report(uid):
...
+    )
+
+    async def wait_for_report():
+        report = await client.get_report(compile_id)
+        return report.code == 200
```
arnaudsjs, line 570: Let's define one `wait_for_report` method and parameterize it. You can use `functools.partial` to pass the parameters to it.

tests/server/test_compilerservice.py
```diff
@@ -525,6 +525,70 @@ def assert_report(uid):
...
+
+    await retry_limited(wait_for_report, 10)
+    report = await client.get_report(compile_id)
+    assert report.result["report"]["partial"]
+    assert report.result["report"]["removed_resource_sets"] == []
+
+    # Do a partial compile with removed resource_sets
+    compile_id, _ = await compilerslice.request_recompile(
+        env, force_update=True, do_export=False, remote_id=remote_id1, partial=True, removed_resource_sets=["a", "b", "c"]
```
arnaudsjs, line 579:
```suggestion
        env, force_update=False, do_export=False, remote_id=remote_id1, partial=True, removed_resource_sets=["a", "b", "c"]
```

tests/server/test_compilerservice.py
arnaudsjs: I See. Checking the `command` field of the `Report` for the expected arguments would be a solution.

arnaudsjs (changes requested):

src/inmanta/server/services/compilerservice.py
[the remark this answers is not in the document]
sanderr: How will lsm request partial compiles if it's not part of the API? Am I missing something?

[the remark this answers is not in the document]
arnaudsjs: lsm directly interacts with the compiler service: https://github.com/inmanta/inmanta-lsm/blob/b726fb80038192c9fb80da7aaaf6f72ae6de651c/src/inmanta_lsm/service_inventory.py#L1356. These requests don't go over the API.

arnaudsjs (approved):

PR #4496, opened 2022-07-01
temporarily disable strict dependency checking

sanderr:
# Description

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


wouterdb (approved):

FloLey (approved):
I think a lot of test will fail, won't that prevent the merge?

sanderr:
Right, there are tests for this... I'll need to address that, good catch.

sanderr:
Right, there are tests for this... I'll need to address that, good catch.

sanderr:
Right, there are tests for this... I'll need to address that, good catch.

sanderr:
An option to disable it has been added.

PR #4508, opened 2022-07-04
 Take into account the requirements.txt on the project

### LINKED ISSUE #4410: Add requirements file to project
As a follow-up of #3383, especially with the issues described in #4279 in mind, it could be helpful if the project could declare additional constraints on modules to help guide the compiler to install appropriate versions in such edge case scenarios. Support should be added for a `requirements.txt` file that is used as a requirements file (`-r`).

The file is already there but only v2 module requirements are actually installed. Other than that it is completely ignored. See https://github.com/inmanta/inmanta-core/blob/5eb1419d2e142c8f9bb8ba2a28721801c94711dd/src/inmanta/module.py#L1833.

# Concrete
- [x] At the start of the project installation the whole file should be taken into account instead of only its v2 modules. It should be installed with `-r`
- [ ] For all subsequent `pip install` commands triggered by the project (i.e. module installations) it should be added as well in order to make sure these constraints are really protected. Here it should be used with `-r` only if its performance impact is limited. A project should be created with all public v2 modules and lsm v2 in its `requirements.txt` to compare performance between `-r` and `-c` for this task.
- [x] The file should be included in the constraints passed to `inmanta.env.check` after project installation

wouterdb:
# Description

This concludes the ticket 4410

**I wonder if this is safe to add, because it can break scenarios that happen to work ?**

closes #4410 
# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/4410-installation-constraints.yml
+++ b/changelogs/unreleased/4410-installation-constraints.yml
@@ -1,0 +1,4 @@
+description: Take into account the requirements.txt on the project when installing or updating
+issue-nr: 4410
+change-type: patch
+destination-branches: [master, iso5]
--- a/changelogs/unreleased/4410-installation-constraints.yml
+++ b/changelogs/unreleased/4410-installation-constraints.yml
@@ -2,3 +2,5 @@ description: Take into account the requirements.txt on the project when installi
 issue-nr: 4410
 change-type: patch
 destination-branches: [master, iso5]
+sections:
+    upgrade-note: "`inmanta project install` and `inmanta project update` now always take into account the `requirements.txt` of the project to provide additional version constraints to pip"
--- a/docs/model_developers/projects.rst
+++ b/docs/model_developers/projects.rst
@@ -1,0 +1,17 @@
+Understanding Projects
+======================
+
+A project is the basic unit of orchestration. It contains:
+
+* ``main.cf``: the entry point for the compiler to start executing
+* ``project.yml``: the project meta data, defines where to find modules and which versions to use. For detailed documentation see: :ref:`project_yml`.
+* ``requirements.txt``: the python dependencies of the project, defines which python dependencies to install and which versions to use.
+
+.. code-block:: sh
+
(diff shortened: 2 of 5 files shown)


arnaudsjs (changes requested):
Shouldn't we also add some documentation regarding this requirements.txt file?

changelogs/unreleased/4410-installation-constraints.yml
```diff
@@ -0,0 +1,6 @@
+description: Take into account the requirements.txt on the project when installing or updating
+issue-nr: 4410
+change-type: patch
+destination-branches: [master, iso5]
+sections:
+    upgrade-note: "`inmanta project install` and `inmanta project update` now always take into account the `requirements.txt` of the project to provide additional version constraints to pip"
```
wouterdb, line 6: Is this OK? I'm not too happy about the formulation?

docs/model_developers/projects.rst
```diff
@@ -0,0 +1,17 @@
+Understanding Projects
```
wouterdb, line 1: Does this add more than confusion?

changelogs/unreleased/4410-installation-constraints.yml
sanderr: Looks good to me. I don't have any suggestions for a better formulation.

docs/model_developers/projects.rst
sanderr: I'm not sure. It's a nicer overview than but some duplication of what's already in `model_developers/project_creation.rst`. I think it might be better to just add a mention of the requirements file there. But I have no real preference.

sanderr (approved):

docs/model_developers/projects.rst
arnaudsjs: This is fine for me. It's consistent with the structure we already have for modules.

arnaudsjs (changes requested):

PR #4534, opened 2022-07-07
fixed bug in eager promises on unset optional relations

sanderr:
# Description

This PR moves evaluation of attribute references from `requires_emit` to `execute` (`ResultVariable.get_value()`) by introducing a result variable proxy. In addition to slightly improved semantics this allows for propagating unset variables. This is then used to fix the bug in the eager promises mechanism where a promise on an unset variable in a block that never gets executed would cause an exception. By propagating the unset variables, no execution units are left dangling (simply ignoring the exception would cause their requires to hang but by propagating the requires resolve to the unset variable one by one).

A couple of side effects:
- The `VariableReader` class has been greatly simplified. Some of its complexity has shifted to `IsDefinedGradualHelper`, the rest just ceased to be relevant.
- some error messages have slightly changed in the sense that failure now only occurs during execution rather than requires_emit. See tests changes for more details. I believe this is acceptable. I think the only fragile one is the `OptionalValueException` in `DynamicProxy` and that one has not changed (I've added a test for it).

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/ast/statements/__init__.py
+++ b/src/inmanta/ast/statements/__init__.py
@@ -20,7 +20,16 @@ from itertools import chain
 from typing import TYPE_CHECKING, Dict, Iterator, List, Optional, Sequence, Tuple
 
 import inmanta.execute.dataflow as dataflow
-from inmanta.ast import Anchor, DirectExecuteException, Location, Named, Namespace, Namespaced, OptionalValueException, RuntimeException
+from inmanta.ast import (
+    Anchor,
+    DirectExecuteException,
+    Location,
+    Named,
+    Namespace,
+    Namespaced,
+    OptionalValueException,
+    RuntimeException,
+)
 from inmanta.execute.dataflow import DataflowGraph
 from inmanta.execute.runtime import (
     ExecutionUnit,
@@ -30,7 +39,6 @@ from inmanta.execute.runtime import (
     RawUnit,
     Resolver,
     ResultCollector,
-    ResultVariable,
     ResultVariable,
     Typeorvalue,
     VariableABC,
@@ -296,9 +304,7 @@ class VariableReferenceHook(RawResumer):
                 variable = unset
             else:
                 # all requires are present, execute instance
-                instance: object = self.instance.execute(
-                    instance_requires, resolver, queue
-                )
+                instance: object = self.instance.execute(instance_requires, resolver, queue)
 
                 if isinstance(instance, list):
                     raise RuntimeException(
@@ -332,7 +338,11 @@ class VariableReferenceHook(RawResumer):
 
     def __repr__(self) -> str:
         return "%s(%r, %s, %r, propagate_unset=%r)" % (
-            self.__class__.__name__, self.instance, self.name, self.variable_resumer, self.propagate_unset
+            self.__class__.__name__,
+            self.instance,
+            self.name,
+            self.variable_resumer,
+            self.propagate_unset,
         )
 
 
--- a/src/inmanta/ast/statements/__init__.py
+++ b/src/inmanta/ast/statements/__init__.py
@@ -235,6 +235,10 @@ class VariableReferenceHook(RawResumer):
     attributes. Calls variable resumer with the variable object as soon as it's available.
     This class is not a full AST node, rather it is a Resumer only. It is meant to delegate common resumer behavior that would
     otherwise need to be implemented as custom resumer logic in each class that needs it.
+
+    :param instance: The instance this variable is an attribute of, if any. Can be a reference to a nested relation instance.
+    :param name: The name of the variable or instance attribute.
+    :param variable_resumer: The resumer that should be called when a value becomes available.
     """
 
     __slots__ = ("instance", "name", "variable_resumer")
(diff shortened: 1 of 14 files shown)


src/inmanta/ast/variables.py
```diff
@@ -178,15 +171,7 @@ def variable_resume(
     ) -> None:
         if self.resultcollector:
             variable.listener(self.resultcollector, self.owner.location)
-
-        if variable.is_ready():
-            self.write_target(variable)
-        else:
-            # reschedule on the variable's completeness
-            RawUnit(queue_scheduler, resolver, {self: variable}, self, override_exception_location=False)
```
sanderr, line 186: I quite like that we now no longer need to reschedule for plain variable dereferencing. Only `is defined` still needs it because it needs to actually process the value rather than just copy it.

src/inmanta/execute/runtime.py
```diff
@@ -267,6 +267,51 @@ def get_dataflow_node(self) -> dataflow.AssignableNodeReference:
         return self._node
 
 
+class ResultVariableProxy(VariableABC[T]):
```
wouterdb, line 270: cunning.

src/inmanta/ast/statements/__init__.py
```diff
@@ -270,7 +274,12 @@ def resume(self, requires: Dict[object, VariableABC], resolver: Resolver, queue:
         variable: ResultVariable[object]
         if self.instance is not None:
             # get the Instance
-            instance: object = self.instance.execute({k: v.get_value() for k, v in requires.items()}, resolver, queue)
+            try:
+                instance: object = self.instance.execute({k: v.get_value() for k, v in requires.items()}, resolver, queue)
+            # TODO: only catch OptionalValueException
+            except Exception as e:
+                breakpoint()
```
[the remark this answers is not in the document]
sanderr: Yes, this is not stable yet.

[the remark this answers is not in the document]
sanderr: (Still in the experimentation phase here so I didn't bother adding any parameters yet.)

src/inmanta/execute/runtime.py
sanderr: I guess it would be a step closer to a dataflow graph based execution, but I think we would need to be really careful about semantics: can you set a value on a proxy? Does it set it on that RV only or does it set it on the one it is connected to as well? What if a proxy is frozen? ...

tests/compiler/test_is_defined.py
```diff
@@ -411,3 +411,25 @@ def test_3026_is_defined_gradual(snippetcompiler, capsys):
...
+A.other [0:1] -- A
+implement A using std::none
+
+a = A(other=null)
+
+isdef = a.other.other is defined
+        """,
+        shouldbe=(
+            "Optional variable accessed that has no value (attribute `__config__::A.other` of `__config__::A (instantiated at"
+            " {dir}/main.cf:7)`) (reported in a.other.other ({dir}/main.cf:9))"
```
sanderr, line 433: This used to be "reported in a.other" rather than "report in a.other.other". I assume this change is acceptable?

tests/compiler/test_relations.py
```diff
@@ -701,3 +701,45 @@ def test_2689_relation_unset_lower(snippetcompiler) -> None:
...
+end
+A.other [0:1] -- A
+implement A using std::none
+
+a = A(other=null)
+other = a.other
+        """,
+        shouldbe=(
+            "Optional variable accessed that has no value (attribute `__config__::A.other` of `__config__::A (instantiated at"
+            " {dir}/main.cf:7)`) (reported in other = a.other ({dir}/main.cf:7))"
```
sanderr, line 722: Same for these two tests: the failure has been shifted a level to where the value is actually accessed, so this is now reported in the assignment rather than the reference. For the one below it's reported in `a.other.other` instead of `a.other`.

tests/compiler/test_is_defined.py
wouterdb: yes

tests/compiler/test_list.py
```diff
@@ -242,8 +242,9 @@ def test_608_opt_to_list(snippetcompiler):
 b1.a = c1.a
 """
     )
-    with pytest.raises(OptionalValueException):
+    with pytest.raises(AttributeException) as exc_info:
```
wouterdb, line 245: Why did this change?

wouterdb (approved):

src/inmanta/ast/statements/__init__.py
```diff
@@ -258,32 +276,52 @@ def schedule(self, resolver: Resolver, queue: QueueScheduler) -> None:
...
-                )
-
-            # get the attribute result variable
-            variable = instance.get_attribute(self.name)
+            instance_requires: dict[object, object] = {}
+            unset: Optional[VariableABC] = None
+            for k, v in requires.items():
+                try:
+                    instance_requires[k] = v.get_value()
+                except (OptionalValueException,) if self.propagate_unset else ():
```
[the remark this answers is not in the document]
sanderr: Me neither, I knew you could pass tuples so I thought I'd try this. It keeps the stack trace cleaner than a catch and reraise.

tests/compiler/test_list.py
sanderr: The error is now raised only when we try to fetch the value from the attribute, rather when we try to dereference it. So it is now caught and reraised my `SetAttributeHelper`.

PR #4554, opened 2022-07-12
Issue/4500 improve module install logging

### LINKED ISSUE #4500: improve module install logging
When module version checks fail it can be very hard to debug the root cause, especially when this happens on Jenkins. Improving the logging of where modules are installed from could be a minor help in this case.

- when a module is installed, it should be logged that it is being installed, which version is being installed, and whether it's a v1 or v2 installation, and the module source (for v1).

The relevant code is at `inmanta.module.ModuleSourceV{1,2}.get_module(*args, **kwargs, install=True)` and/or where those methods are called in `inmanta.module.Module.load_module()`.

Additionally, add as debug logging:
- pip install output
- git clone output

This should be piped from the subprocess output (rather than collected and logged at the end). Something similar is done in `inmanta.server.services.compilerservice`.

Hugo-Inmanta:
# Description

Add debug logging on module install

closes #4500 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


changelogs/unreleased/4500-improve-module-install-logging.yml
```diff
@@ -0,0 +1,7 @@
+description: Improve logging on module installation.
+change-type: patch
+destination-branches: [master, iso5, iso4]
```
Hugo-Inmanta, line 3: Not too sure about iso4 ?

src/inmanta/module.py
```diff
@@ -2279,6 +2282,15 @@ def requires(self) -> "List[InmantaModuleRequirement]":
...
         return reqs
 
+    def log_install_information(self, project: "Project", module_name: str) -> None:
+        """
+        Display information about this module's installation
+        :param project: The project associated with the module.
+        :param module_name: The module's name.
+        """
+        source = f" from {project.downloadpath}" if self.GENERATION == ModuleGeneration.V1 else ""
+        LOGGER.debug(f"Installing module {module_name} (v{self.GENERATION.value}) version {self.version}{source}.")
```
sanderr, line 2292: Perhaps we could include the constraints as additional debugging information? e.g. "Installing module mymod (v2) version 1.0.0 (from constraints mymod~=1.0,mymod>=1.0.0)."

sanderr (changes requested):
- Do you think it would be feasible to test the tee-like behavior? Checking the output should be trivial but checking that they appear in real-time is a bit harder. Perhaps we can use `LogRecord.asctime` on the first and last output line to verify that there's a time gap there? Perhaps with a subprocess like `sh -c 'echo one && sleep 1 && echo two'`? Not worth investing too much effort but it would be nice to have it.
- Could you add a test for the v1 module installation message?

src/inmanta/env.py
```diff
@@ -391,8 +391,8 @@ def _run_pip_install_command(
             if "versions have conflicting dependencies" in stderr:
                 raise ConflictingRequirements(stderr)
             raise e
-        except Exception:
-            raise
+        except Exception as e:
+            raise e
```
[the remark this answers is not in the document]
Hugo-Inmanta: Reverting this one. It was an artifact from a previous implementation.

src/inmanta/module.py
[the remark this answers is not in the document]
Hugo-Inmanta: Good idea :+1:

FloLey (approved):

sanderr (approved):
Passing on my review to Arnaud.

src/inmanta/module.py
```diff
@@ -611,6 +649,19 @@ def __init__(self, local_repo: "ModuleRepo", remote_repo: "ModuleRepo") -> None:
         self.local_repo: ModuleRepo = local_repo
         self.remote_repo: ModuleRepo = remote_repo
 
+    def log_pre_install_information(self, project: "Project", module_name: str) -> None:
+        local_repo = self.local_repo.path_for(module_name)
+
+        assert local_repo is not None, f"Failed to install module {module_name} because its local repository is not defined."
+        remote_repo = gitprovider.get_remote(local_repo).strip()
+        LOGGER.info(f"Installing module {module_name} (v1) from {remote_repo}.")
```
arnaudsjs, line 657:
```suggestion
        remote_repo = gitprovider.get_remote(local_repo).strip()
        LOGGER.info(f"Installing module %s (v1) from %s.", module_name, remote_repo)
```

src/inmanta/module.py
```diff
@@ -611,6 +649,19 @@ def __init__(self, local_repo: "ModuleRepo", remote_repo: "ModuleRepo") -> None:
...
+
+    def log_post_install_information(self, project: "Project", module_name: str) -> None:
+        module = self.get_installed_module(project, module_name)
+
+        assert module is not None
+        LOGGER.info(f"Successfully installed module {module_name} (v1) version {module.version}")
```
arnaudsjs, line 663:
```suggestion
        LOGGER.info(f"Successfully installed module %s (v1) version %s", module_name, module.version)
```

src/inmanta/module.py
```diff
@@ -611,6 +649,19 @@ def __init__(self, local_repo: "ModuleRepo", remote_repo: "ModuleRepo") -> None:
         self.local_repo: ModuleRepo = local_repo
         self.remote_repo: ModuleRepo = remote_repo
 
+    def log_pre_install_information(self, project: "Project", module_name: str) -> None:
+        local_repo = self.local_repo.path_for(module_name)
+
+        assert local_repo is not None, f"Failed to install module {module_name} because its local repository is not defined."
```
arnaudsjs, line 655: If the V1 module is not present on the local filesystem yet, this check will fail right? Or am I missing something?

tests/moduletool/test_install.py
```diff
@@ -904,3 +906,86 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+            "minimalv2module",
+            [
+                ("Installing module minimalv2module (v2)", logging.INFO),
+                ("Successfully installed module minimalv2module (v2) version", logging.INFO),
+            ],
+            "v2",
+        ),
+        (
+            "minimalv1module",
+            [("Installing module std (v1)", logging.INFO), ("Successfully installed module std (v1) version", logging.INFO)],
```
arnaudsjs, line 924: Why do we check for std in the output?

src/inmanta/env.py
```diff
@@ -378,19 +378,23 @@ def _run_pip_install_command(
...
-                raise ConflictingRequirements(stderr)
-            raise e
+            return_code, full_output = self._run_command_and_stream_output(cmd)
+
+            if return_code:
+                not_found: List[str] = []
+                conflicts: bool = False
+                for line in full_output:
+                    if "No matching distribution found for " in line:
+                        # Add missing package name to not_found list
```
arnaudsjs, line 388: As far as I can see we don't have a test case to verify the log lines when no matching distribution if found. I think we need to add one.

arnaudsjs (changes requested):

src/inmanta/module.py
```diff
@@ -556,6 +587,13 @@ def install(self, project: "Project", module_spec: List[InmantaModuleRequirement
             raise InvalidModuleException(f"{python_package} does not contain a {namespace_package} module.")
         return self.from_path(project, module_name, path)
 
+    def log_pre_install_information(self, project: "Project", module_name: str) -> None:
+        LOGGER.info(f"Installing module {module_name} (v2) version PLACEHOLDER.")
```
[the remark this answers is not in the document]
Hugo-Inmanta: The PLACEHOLDER part needs to go, I forgot to remove it.

tests/moduletool/test_install.py
Hugo-Inmanta: The std module is the only one that get properly installed, the minimalv1module gets copied in the current workdir and so it won't trigger the install logging part. Maybe there's a better way to test for this ?

arnaudsjs: I see. So the std module will get installed for each of the two runs of this test case. Wouldn't it then be better to make this a non-parameterized test case and check for the two log lines at the end of the test case. That would make the test case run faster without losing anything else, right?

Hugo-Inmanta: Indeed :+1:

src/inmanta/module.py
Hugo-Inmanta: (I can't remember if we talked about this during one of our calls)
I'm assuming that after the install the v1 module should be present on the local filesystem, is it ok to do so ?

src/inmanta/module.py
```diff
@@ -611,6 +667,35 @@ def __init__(self, local_repo: "ModuleRepo", remote_repo: "ModuleRepo") -> None:
...
+
+        assert module is not None
+        LOGGER.info("Successfully installed module %s (v1) version %s from %s", module_name, module.version, remote_repo)
+
+        version_snapshot = {name: mod.version for name, mod in project.modules.items()}
+        if version_snapshot:
+            LOGGER.debug("Snapshot of modules versions post-install:")
+            LOGGER.debug("\n".join(f"{mod}: {version}" for mod, version in version_snapshot.items()))
+
+        LOGGER.debug(f"Snapshot difference:\n{set(modules_pre_install.items()) ^ set(version_snapshot.items())}")
```
Hugo-Inmanta, line 697: This needs further work, the symetric difference is not explicit enough

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,163 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+
+
+def test_no_matching_distribution(local_module_package_index: str, snippetcompiler_clean, caplog):
+    """ """
+    caplog.set_level(logging.DEBUG)
+
+    v2_module = "no_matching_dependency"
+
+    v2_requirements = [Requirement.parse(module.ModuleV2Source.get_package_name_for(v2_module))]
+
```
Hugo-Inmanta, line 1025: Rewrite this using the module_from_template() method to avoid polluting the modules_v2 directory.

src/inmanta/env.py
Hugo-Inmanta: Added test : `test_no_matching_distribution`

arnaudsjs (changes requested):
* The ticket also mentions that `git clone` output has to be logged. I don't think this was implemented
* This implementation logs a message each time a module is installed. At the end of the installation we also need to provide an overview of all the modules that are installed, because some modules may have been installed as a transitive dependency of a certain module. The overview should be logged in the log level `INFO`. The messages for the installation of a specific package in `DEBUG`. That way the user can decide whether he/she wants to see all the details or not.

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,218 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+    # ensure we can compile
+    compiler.do_compile()
+
+    expected_logs = [
+        ("Installing module minimalv2module (v2)", logging.INFO),
+        ("Successfully installed module minimalv2module (v2) version", logging.INFO),
+        ("Installing module std (v1)", logging.INFO),
+        ("Successfully installed module std (v1) version", logging.INFO),
+    ]
```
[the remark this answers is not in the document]
Hugo-Inmanta: Are the versions pinned somewhere ? I was reluctant to include a version that might make the test fail in the future.

src/inmanta/module.py
```diff
@@ -427,9 +435,51 @@ def get_module(
...
+    ) -> None:
+        """
+        Display information about this module's installation after the actual installation.
+
+        :param project: The project associated with the module.
+        :param module_name: The module's name.
+        """
+        raise NotImplementedError("Abstract method")
+
+    @staticmethod
```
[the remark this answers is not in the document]
arnaudsjs: No. I was just wondering if there was a specific reason why you added it like that. I have trouble seeing why the concept of a `staticmethod` exists and in which situation it offers benefits. But I have no objections against this.

tests/moduletool/test_install.py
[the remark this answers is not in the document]
arnaudsjs: These modules exist in the `data/modules` or the `data/module_v2` directory with a specific version. These version are intended to remain table over time.

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,218 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+
+
+def test_module_install_logging(local_module_package_index: str, snippetcompiler_clean, caplog) -> None:
+    """
+    Make sure the module's informations are displayed when it is being installed for both v1 and v2 modules.
+    """
+
+    caplog.set_level(logging.INFO)
+
+    v1_module = "minimalv1module"
```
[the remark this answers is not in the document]
Hugo-Inmanta: Indeed, I'll just replace it there: `961         "\n".join(f"import {mod}" for mod in ["std", v1_module, v2_module]),`

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,218 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+    snippetcompiler_clean.setup_for_snippet(
+        f"""
+        import {module.ModuleV2.get_name_from_metadata(modone)}
+        import {module.ModuleV2.get_name_from_metadata(modtwo)}
+        """,
+        autostd=False,
+        python_package_sources=[local_module_package_index],
+        python_requires=v2_requirements,
+        install_project=True,
+    )
```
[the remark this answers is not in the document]
Hugo-Inmanta: Fixed this with changing `install_project=True,` to `False`.
Out of curiosity I tried the other way round (By changing install to False in both `module_from_template` and keeping install_project to True) but I couldn't make it work, what am I missing?

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,218 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+
+    os.chdir(project.path)
+
+    # autostd=True reports std as an import for any module, thus requiring it to be v2 because v2 can not depend on v1
+    module.Project.get().autostd = False
+    ProjectTool().execute("install", [])
+
+    # ensure we can compile
+    compiler.do_compile()
```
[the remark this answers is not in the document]
Hugo-Inmanta: Removing this :+1:

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,218 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+
+    log_contains(
+        caplog,
+        "inmanta.env",
+        logging.DEBUG,
+        "No matching distribution found for inmanta-module-child-module==3.3.3",
+    )
+
+    shutil.rmtree(os.path.join(str(tmpdir), "parent_module"), ignore_errors=True)
+    shutil.rmtree(os.path.join(str(tmpdir), "child_module"), ignore_errors=True)
```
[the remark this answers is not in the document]
Hugo-Inmanta: Removing this :+1:

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,218 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+        )
+
+    log_contains(
+        caplog,
+        "inmanta.env",
+        logging.DEBUG,
+        "No matching distribution found for inmanta-module-child-module==3.3.3",
+    )
+
+    shutil.rmtree(os.path.join(str(tmpdir), "parent_module"), ignore_errors=True)
```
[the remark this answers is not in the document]
Hugo-Inmanta: Removing this :+1:

src/inmanta/env.py
```diff
@@ -500,6 +516,33 @@ def _run_command_and_log_output(
...
+
+        assert process.stdout is not None  # Make mypy happy
+
+        for line in process.stdout:
+            # Eagerly consume the buffer to avoid a deadlock in case the subprocess fills it entirely.
+            output = line.decode()
+            full_output.append(output)
+            LOGGER.debug(output)
+
+        return_code = process.wait()
```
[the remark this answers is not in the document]
Hugo-Inmanta: Added a 10s timeout

src/inmanta/env.py
```diff
@@ -380,19 +390,25 @@ def _run_pip_install_command(
...
-            if "versions have conflicting dependencies" in stderr:
-                raise ConflictingRequirements(stderr)
-            raise e
+            return_code, full_output = self._run_command_and_stream_output(cmd)
+
+            if return_code:
+                not_found: List[str] = []
+                conflicts: bool = False
+                for line in full_output:
+                    if "No matching distribution found for " in line:
```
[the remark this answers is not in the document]
Hugo-Inmanta: Indeed :+1:  Changed to regex

src/inmanta/module.py
```diff
@@ -611,6 +681,30 @@ def __init__(self, local_repo: "ModuleRepo", remote_repo: "ModuleRepo") -> None:
...
+        self, project: "Project", module_name: str, modules_pre_install: Dict[str, "Version"]
+    ) -> None:
+        local_repo = self.local_repo.path_for(module_name)
+
+        assert local_repo is not None, f"Failed to install module {module_name} because its local repository is not defined."
+        remote_repo = gitprovider.get_remote(local_repo).strip()
+        module = self.get_installed_module(project, module_name)
+
+        assert module is not None
+        LOGGER.info("Successfully installed module %s (v1) version %s from %s", module_name, module.version, remote_repo)
```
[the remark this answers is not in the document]
Hugo-Inmanta: I'm not sure where to find this information. Is it `local_repo`?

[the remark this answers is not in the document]
arnaudsjs: The module object has a `path` property that contains the path to the module.

tests/moduletool/test_install.py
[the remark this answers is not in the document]
arnaudsjs: It doesn't work the other way around because the modules, generated from the templates, are not stored in a python index. As such the install on the project can't find the requested packages. That makes me realize it's probably better to use the install from the `setup_for_snippet` everywhere instead of the install from the `module_from_template`. The former is a simulation of the `inmanta project install` command while the latter is one of the `inmanta module install` command. This debugging information is mainly relevant for the `inmanta project install` command, so it's probably better for the tests to also use that logic.

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,386 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+
+    snippetcompiler_clean.setup_for_snippet(
+        f"""
+        import {module.ModuleV2.get_name_from_metadata(modone)}
+        import {module.ModuleV2.get_name_from_metadata(modtwo)}
+        """,
+        autostd=False,
+        python_package_sources=[local_module_package_index],
+        python_requires=v2_requirements,
+        install_project=False,
```
arnaudsjs, line 1058: Why set `install_project=False` here and `install=True` on the templates?

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,386 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+        import std
+        """,
+        autostd=False,
+        python_package_sources=[local_module_package_index],
+        install_project=True,
+    )
+
+    expected_logs = [
+        ("Cloning into %s..." % os.path.join(project.downloadpath, "std"), logging.DEBUG),
+    ]
```
arnaudsjs, line 1092:
```suggestion
def test_git_clone_output(snippetcompiler_clean, caplog, modules_v2_dir, tmpdir):
    """
    This test checks that git clone output is correctly logged on module install.
    """
    caplog.set_level(logging.DEBUG)

    project = snippetcompiler_clean.setup_for_snippet(
        """
        import std
        """,
        autostd=False,
        install_project=True,
    )

    expected_logs = [
        ("Cloning into %s..." % os.path.join(project.downloadpath, "std"), logging.DEBUG),
    ]
```

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,386 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+    )
+    parent_module = module_from_template(
+        os.path.join(modules_v2_dir, "minimalv2module"),
+        os.path.join(str(tmpdir), "parent_module"),
+        new_version=version.Version("1.2.3"),
+        new_name="parent_module",
+        install=False,
+        new_requirements=[InmantaModuleRequirement.parse("child_module==3.3.3")],
+        publish_index=index,
+    )
```
arnaudsjs, line 1164: This one was already published to the index earlier. No need to push it again.

```suggestion
```

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,386 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+
+    parent_module = module_from_template(
+        os.path.join(modules_v2_dir, "minimalv2module"),
+        os.path.join(str(tmpdir), "parent_module"),
+        new_version=version.Version("1.2.3"),
+        new_name="parent_module",
+        install=False,
+        new_requirements=[InmantaModuleRequirement.parse("child_module==3.3.3")],
+        publish_index=index,
+    )
```
arnaudsjs, line 1208: This one was already published to the index earlier. No need to push it again.
```suggestion
    parent_module = module_from_template(
        os.path.join(modules_v2_dir, "minimalv2module"),
        os.path.join(str(tmpdir), "parent_module"),
        new_version=version.Version("1.2.3"),
        new_name="parent_module",
        install=False,
        new_requirements=[InmantaModuleRequirement.parse("child_module==3.3.3")],
        publish_index=index,
    )
```

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,386 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+    # We expect the latest version for a and b to be newly added
+    log_contains(
+        caplog,
+        "inmanta.module",
+        logging.INFO,
+        (
+            """
+Snapshot of modules versions post-install:\n
++inmanta-module-module-a: 5.0.0\n
++inmanta-module-module-b: 1.2.3"""
```
arnaudsjs, line 1295: Why is there line break character there? That means that there are two line breaks after eachother?

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,386 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+
+    # We expect :
+    # - a to be downgraded to a compatible version
+    # - b to remain unchanged
+    # - c to be installed with the latest version
+
+    log_contains(
+        caplog,
+        "inmanta.module",
+        logging.INFO,
```
arnaudsjs, line 1319:
```suggestion
        logging.DEBUG,
```

arnaudsjs (changes requested):

arnaudsjs:
When an `inmanta compile` or an `inmanta export` is executed. We should also print a full list of the V1 and V2 modules that are installed.

tests/moduletool/test_install.py
wouterdb: I would suspect that this test will still pass if you remove this whole `setup_for_snippet` as the module_from_template will do an install already

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,386 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
     """.strip() in str(
         excinfo.value
     )
+
+
+def test_module_install_logging(local_module_package_index: str, snippetcompiler_clean, caplog) -> None:
```
wouterdb, line 947: I think all these test will have to be marked as slow

src/inmanta/module.py
```diff
@@ -427,9 +452,60 @@ def get_module(
...
+                prefix = " "
+                suffix = ""
+                if package_name not in modules_pre_install.keys():
+                    prefix = "+"
+                elif package_name in [elmt[0] for elmt in updates_and_additions]:
+                    prefix = "~"
+                    suffix = f" (was previously {modules_pre_install[package_name]})"
+
+                out.append(prefix + package_name + ": " + str(package_version) + suffix)
+            LOGGER.info("\n".join(out))
```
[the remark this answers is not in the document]
Hugo-Inmanta: >  This implementation logs a message each time a module is installed. At the end of the installation we also need to provide an overview of all the modules that are installed, because some modules may have been installed as a transitive dependency of a certain module. The overview should be logged in the log level INFO. The messages for the installation of a specific package in DEBUG. That way the user can decide whether he/she wants to see all the details or not.

I logged this overview at the info level because of this :point_up:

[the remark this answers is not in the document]
Hugo-Inmanta: Oh nevermind, I missed your other comment about addind a proper overview on `inmanta export` or `inmanta compile`

tests/moduletool/test_install.py
Hugo-Inmanta: Changing this test to `install_project=True` and `install=False` on the templates for consistency

src/inmanta/module.py
```diff
@@ -611,6 +707,36 @@ def __init__(self, local_repo: "ModuleRepo", remote_repo: "ModuleRepo") -> None:
         self.local_repo: ModuleRepo = local_repo
         self.remote_repo: ModuleRepo = remote_repo
 
+    def log_pre_install_information(self, project: "Project", module_name: str) -> Dict[str, "Version"]:
+
+        LOGGER.debug("Installing module %s (v1).", module_name)
+
+        version_snapshot = env.PythonWorkingSet.get_packages_in_working_set(inmanta_modules_only=True)
```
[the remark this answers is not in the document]
arnaudsjs: Indeed. I overlooked this because I thought a V2 module might get installed as a dependency of the V1 modules. But this call only installs the V1 module itself and not its dependencies.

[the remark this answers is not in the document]
Hugo-Inmanta: Removed for v1 modules

changelogs/unreleased/4500-improve-module-install-logging.yml
wouterdb: I wouldn't do iso4

src/inmanta/module.py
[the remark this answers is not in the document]
wouterdb: Do we have consistent logging for V1 and V2 then? 

I think the testcases should also include some V1's to verify consistency.

wouterdb (changes requested):
Do we have consistent logging for V1 and V2 then?

I think the testcases should also include some V1's to verify consistency.

src/inmanta/module.py
```diff
@@ -556,6 +596,53 @@ def install(self, project: "Project", module_spec: List[InmantaModuleRequirement
...
+        LOGGER.debug("Installing module %s (v2).", module_name)
+
+    @staticmethod
+    def take_v2_modules_snapshot() -> Dict[str, "Version"]:
+        """
+        Log and return a dictionary containing currently installed v2 modules and their versions
+        """
+        version_snapshot = env.PythonWorkingSet.get_packages_in_working_set(inmanta_modules_only=True)
+        if version_snapshot:
+            out = ["Snapshot of modules versions pre-install:"]
```
sanderr, line 609:
```suggestion
            out = ["Snapshot of modules versions before installation:"]
```

src/inmanta/module.py
```diff
@@ -556,6 +596,53 @@ def install(self, project: "Project", module_spec: List[InmantaModuleRequirement
...
+            out.extend(f"{mod}: {version}" for mod, version in version_snapshot.items())
+            LOGGER.debug("\n".join(out))
+        return version_snapshot
+
+    @staticmethod
+    def log_snapshot_difference_v2_modules(modules_pre_install: Dict[str, "Version"]) -> None:
```
sanderr, line 615: Nice, I like it!

src/inmanta/module.py
sanderr, line 609: This message is specific to the pre-installation check. Either it shouldn't be, or it should mention this as a usage constraint in the docstring.

src/inmanta/module.py
```diff
@@ -556,6 +596,53 @@ def install(self, project: "Project", module_spec: List[InmantaModuleRequirement
...
+
+    @staticmethod
+    def take_v2_modules_snapshot() -> Dict[str, "Version"]:
+        """
+        Log and return a dictionary containing currently installed v2 modules and their versions
+        """
+        version_snapshot = env.PythonWorkingSet.get_packages_in_working_set(inmanta_modules_only=True)
+        if version_snapshot:
+            out = ["Snapshot of modules versions pre-install:"]
+            out.extend(f"{mod}: {version}" for mod, version in version_snapshot.items())
```
sanderr, line 610: We might want to work with the inmanta module names rather than the Python package names here. There should be a conversion method somewhere in this file. Same for other logging methods.

src/inmanta/module.py
```diff
@@ -611,6 +698,25 @@ def __init__(self, local_repo: "ModuleRepo", remote_repo: "ModuleRepo") -> None:
         self.local_repo: ModuleRepo = local_repo
         self.remote_repo: ModuleRepo = remote_repo
 
+    def log_pre_install_information(self, project: "Project", module_name: str) -> None:
+        LOGGER.debug("Installing module %s (v1).", module_name)
```
sanderr, line 702: I think this would be the perfect place to add [constraints information](https://github.com/inmanta/inmanta-core/pull/4554#discussion_r923126078) (same for v2). I'll get back to your question there ASAP.

sanderr (changes requested):
Still need to review the tests. I quite like how this has evolved since I last looked at it!

src/inmanta/app.py
```diff
@@ -363,6 +363,8 @@ def compile_project(options: argparse.Namespace) -> None:
         do_compile()
         LOGGER.debug("Compile time: %0.03f seconds", time.time() - t1)
 
+    project_instance.log_installed_modules()
```
sanderr, line 366: Wouldn't it make more sense to log this before starting the compile? Same for export.

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,372 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+        )
+    log_contains(
+        caplog,
+        "inmanta.env",
+        logging.DEBUG,
+        "No matching distribution found for inmanta-module-child-module==3.3.3",
+    )
+
+    # Scenario 2
+    # parent_module requires child_module v3.3.3 which is installed with v1.1.1
```
sanderr, line 1148:
```suggestion
    # parent_module requires child_module v3.3.3 but the index only has v1.1.1
```
Is this what you meant? I found the wording pretty confusing.

tests/moduletool/test_install.py
```diff
@@ -941,3 +942,372 @@ def test_install_project_with_install_mode_master(tmpdir: py.path.local, snippet
...
+        caplog,
+        "inmanta.env",
+        logging.DEBUG,
+        "No matching distribution found for inmanta-module-child-module==3.3.3",
+    )
+
+    shutil.rmtree(os.path.join(str(tmpdir), "child_module"))
+
+    # Scenario 3
+    # parent_module requires child_module v3.3.3 which is installed with v3.3.3
```
sanderr, line 1182:
```suggestion
    # parent_module requires child_module v3.3.3 which is present in the index.
```
Same here.

sanderr (changes requested):

src/inmanta/env.py
```diff
@@ -498,6 +498,43 @@ def _run_command_and_log_output(
...
+
+            if "versions have conflicting dependencies" in output:
+                conflicts = True
+
+        return_code = process.wait()
+
+        if not_found:
+            raise PackageNotFound("Packages %s were not found in the given indexes." % ", ".join(not_found))
+        if conflicts:
+            raise ConflictingRequirements("/n".join(full_output))
```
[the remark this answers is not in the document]
sanderr: Could you see if it's feasible to fetch only the relevant lines? I think we definitely don't want the full pip output (it's already streamed as debug logging). If it's not feasible to get the relevant lines, perhaps we should just point to the debug logging? I'm not sure about this. But if it's feasible I think that's clearly the best approach.

src/inmanta/app.py
Hugo-Inmanta: Agreed :+1:

src/inmanta/module.py
sanderr: The `module_spec` argument of `ModuleSource.install` is a list of `Requirement` instances on that module. You should be able to easily extract the set of constraints from that, e.g. in `log_pre_install_information`.

wouterdb (approved):

tests/moduletool/test_install.py
```diff
@@ -1305,9 +1305,77 @@ def test_version_snapshot(local_module_package_index: str, snippetcompiler, capl
...
+            module.InmantaModuleRequirement.parse("module_a<=18.8.8"),
+            module.InmantaModuleRequirement.parse("module_a<18.8.8"),
+        ],
+    )
+    log_contains(
+        caplog,
+        "inmanta.module",
+        logging.DEBUG,
+        (
+            "Installing module module_a (v2) (from constraints module_a>0.0 module_a>=0.0 module_a==8.8.8 module_a<=18.8.8 "
```
sanderr, line 1351: Perhaps also add a v1 and a v2 module that have a constraint on module_a? Should suffice to do this advanced test only here, no need to do the same for the v1 module test below I think.

src/inmanta/module.py
```diff
@@ -458,6 +476,52 @@ def _should_install_module() -> bool:
...
 
+    def _get_constraints(self, module_name: str, module_spec: List[InmantaModuleRequirement]) -> str:
+        """
+        Returns the constraints on a given inmanta module as a string.
+
+        :param module_name: The name of the module.
+        :param module_spec: List of inmanta requirements in which to look for the module.
+        """
+        constraints_on_module: List[str] = [str(req) for req in module_spec if module_name == req.key and req.specs]
+        from_constraints = f" (from constraints {' '.join(constraints_on_module)})" if constraints_on_module else ""
```
sanderr, line 487: Return the empty string if `not constraints_on_module`?

sanderr, line 487:
```suggestion
        from_constraints = f" (with constraints {' '.join(constraints_on_module)})" if constraints_on_module else ""
```
Not a hard requirement, but this reads better to me.

sanderr (changes requested):

src/inmanta/module.py
Hugo-Inmanta: Yes, if there are no constraints it will read:
```bash
"Installing module std (v1)
```
I think it's better to be explicit and to go with:
```bash
"Installing module std (v1) (with no version constraint)
```

sanderr: Oh wow, not sure how I missed that. My comment was actually a suggestion to do that. I guess I read the line to quickly, my apologies.

sanderr: I have no strong preference as to the explicit "no version constraint".

sanderr (approved):

PR #4559, opened 2022-07-12
Issue/4455: removed resource sets are present in partial diff

### LINKED ISSUE #4455: undefined behavior when removed resource sets are present in partial diff
When a partial export is done with `--remove-resource-sets myset` but `myset` is still present in the resources that are exported, the server does not reject this. It should. This behavior should be well defined and documented.

Corner case: if the set does exist in the list but it is empty, there's no need to reject it.

FloLey:
# Description

When a partial export is done with --remove-resource-sets myset but myset is still present in the resources that are exported, the server does not reject this. It should. This behavior should be well defined and documented.

Corner case: if the set does exist in the list but it is empty, there's no need to reject it.

closes #4455 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/4455-removed-resource-sets-present-in-export.yml
+++ b/changelogs/unreleased/4455-removed-resource-sets-present-in-export.yml
@@ -1,5 +1,5 @@
 ---
-description: Removed resource sets are present in partial diff raises excpetion
+description: raises exception if removed resource sets are present in partial diff
 issue-nr: 4455
 issue-repo: inmanta-core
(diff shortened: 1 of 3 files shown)


sanderr (changes requested):
I'm afraid there's been a miscommunication here. The purpose of this ticket was to perform this check on the server side, in `put_partial` (or the methods it calls). I believe the check there should make this one redundant. Could you double check that? If it doesn't we can keep both.

PR #4569, opened 2022-07-15
Issue/4384 regular timeouts on core tests

### LINKED ISSUE #4384: regular timeouts on core tests
The core tests and Pytest in Docker regularly timeout (Jenkins timeout, not python timeout) on `tests/server/test_server_status.py`. This seems to happen every few days (both on slow and fast runs). So far we've been unable to reproduce it consistently. This ticket is meant to keep track of what has been tried to investigate this, as well as collect helper scripts for future investigation.

# Reliable reproduction (sort of)
Based on the "Preceding tests" section below I managed to set up a Jenkins job for a reliable reproduction. It executes pytest 300 times on four tests in a given order. This takes about 45 minutes if successful. However, luckily most runs aren't successful and the timeout occurs somewhere down the road. The Jenkins job can be found [here](https://jenkins.inmanta.com/job/core/job/inmanta-core/job/timeout-experiments-2/). I didn't change anything on Jenkins' side, only modified tox a bit on the `timeout-experiments-2` branch.

I have not yet investigated whether the run of four tests used here are actually a minimal set to reproduce this (I expect not). Feel free to experiment as you like with the branch. At the time of writing the tox command config used is the following:
```
commands=sh -c 'seq 300 | xargs -I_i_ sh -c "echo run _i_ && py.test --log-level DEBUG --junitxml=junit-{envname}.xml -vvv {env:INMANTA_EXTRA_PYTEST_ARGS:} --randomly-dont-reorganize tests/agent_server/test_evolution.py::test_resource_evolution tests/server/test_server_status.py::test_server_status_database_unreachable tests/server/test_server_status.py::test_server_status tests/compiler/test_is_defined.py::test_is_defined_block_scoping[if\ in\ implementation\ self]"'
```

# First occurrence
This issue was first observed in Pytest in Docker build 1905 on 2022-06-07. The build logs at the time of writing only go back until 2022-05-23. Within that window no other instances of this issue were found. Given the infrequency with which this occurs, nothing can be said with certainty about when exactly the root cause was introduced, but it seems likely (because of the large amount of tests run in Pytest in Docker every day) that it was on or a few days before 2022-06-07.

# Preceding tests
A potential cause could be that one or more of the preceding tests occasionalmultiple timesly result in a state where the timeout occurs. If this is indeed the case, we might be able to pinpoint which tests are responsible by collecting data from each build that timed out this way.

Below is an overview of all relevant builds so far:
- [936.txt](https://app.zenhub.com/files/67439560/630bcc78-b31e-4dd4-9274-e6dd99eb4b57/download)
- [943.txt](https://app.zenhub.com/files/67439560/c4efba45-4bfc-48b1-b1fc-c3ddebc41401/download)
- [docker-1905-master-fedora-latest.txt](https://app.zenhub.com/files/67439560/afe9fef9-ed8b-4314-861e-86265f7d6308/download)
- [docker-1907-iso5-stable-fedora-latest.txt](https://app.zenhub.com/files/67439560/c33b96c2-012e-4433-9e94-996d840d2dca/download)
- [docker-1911-master-centos8.txt](https://app.zenhub.com/files/67439560/c0a6221e-c5f3-401b-8665-0c30a6e08770/download)
- [docker-1912-iso5-ubuntu-rolling.txt](https://app.zenhub.com/files/67439560/e23c680b-a860-45c3-9fab-c38493bc7e02/download)

The files list all (non-skipped) tests up to the timed out one, in the order in which they were run. From this the tests that occurred in all runs can be determined: [all-6.txt](https://app.zenhub.com/files/67439560/85f7fd48-8cf5-472e-bcd2-e947b7df547d/download) (number in the name indicates number of builds included when file was generated). These can then be run with `xargs -a <file> pytest --randomly-dont-reorganize`. So far I did not manage to reproduce the issue using this approach (about 200 runs at the time of writing, part on my local machine, part on Jenkins).

The single build files above were obtained using the following procedure:
- copy all test lines from Jenkins output
- `grep -v " SKIPPED " <input> | sed 's/^\(..:..:.. \)\?\(.*\) \(PASSED\|FAILED\) .*/\2/'  > <output>`

The common tests were then determined with `grep -Ff <one> <two> | grep -Ff <three> | ...`.

# Randomly seed
Attempts were made to run with the same randomly-seed on the same commit as one of the timed out builds. So far (10+ builds at the time of writing) this did not manage to reproduce the issue.

# Jenkins agent
Both builds so far ran on jenkins runner 2. This runner does not result in consistent timeouts, even when combined with the randomly seed approach (at the time of writing only forced one such build on jenkins runner 2).

# Stack dump
This has not been tried yet but is the next path to be explored. If we could temporarily set up the Pytest in Docker and/or core test jobs to detect timeouts like this and export a dump, this could give some more information as soon as the issue pops up again.

Hugo-Inmanta:
# Description

Fix deadlock in the database locking mechanism by shielding the queries from cancellation.

closes #4384 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/inmanta/server/services/compilerservice.py
```diff
@@ -517,7 +517,9 @@ async def _cleanup(self) -> None:
         )
         LOGGER.info("Cleaning up compile reports that are older than %s", oldest_retained_date)
         try:
-            await data.Compile.delete_older_than(oldest_retained_date)
+            # Shield this task from cancellation to make sure locks will be released properly if the encasing task
+            # is cancelled in flight. https://github.com/inmanta/inmanta-core/issues/4384
+            await shield(data.Compile.delete_older_than(oldest_retained_date))
```
Hugo-Inmanta, line 522: Not sure if this should be done for some other database query as well.

changelogs/unreleased/4384-regular-timeouts-on-core-tests.yml
```diff
@@ -0,0 +1,6 @@
+description: Fix deadlock in the database locking mechanism.
```
sanderr, line 1: Let's make the changelog message a bit less alarming:
```suggestion
description: Fix rare deadlock in the database locking mechanism when tasks are cancelled, mostly affects test environments.
```

src/inmanta/server/services/compilerservice.py
sanderr: I'm not 100% sure either but it looks fine to me. Let's merge it now and I'll make sure to point Arnaud here when he's back.

sanderr (dismissed):

src/inmanta/server/services/compilerservice.py
arnaudsjs: The problem is a lot bigger than this. Each background task that interacts with the database can conflict with the truncate query when it's cancelled. The above-mentioned query is one such background task, but we have several others as well. I think it would be good to do the shielding at the lowest in the database layer. That way all queries are protected.

An alternative I think about now, is to add support to the Scheduler to await tasks when the scheduler is stopped instead of cancelling them. We already have the equivalent functionality in the `add_background_task()` method using the `cancel_on_stop` argument. The advantage of this solution is that all queries are finished by the time we shutdown the DatabaseService. By shielding the task, it keeps executing. That way we might end up in the scenario where the database connection is shut down before the task ends, which again could cause other issues. But I would like to get the opinion of Wouter before we start implementing this. I remember he had a strong opinion on how this should be implemented. But I don't exactly remember the rationale.

arnaudsjs (changes requested):

src/inmanta/server/services/compilerservice.py
wouterdb: I lforgot we have the 'cancel on stop' option. 

So it would seem like a good solution to set 'cancel on stop' to false on all the database queries, this makes the shutdown wait for them.

src/inmanta/util.py
```diff
@@ -236,14 +236,18 @@ def __init__(self, name: str) -> None:
...
-    def _add_to_executing_tasks(self, action: TaskMethod, task: asyncio.Task[object]) -> None:
+    def _add_to_executing_tasks(self, action: TaskMethod, task: asyncio.Task[object], cancel_on_stop: bool = True) -> None:
         """
         Add task that is currently executing to `self._executing_tasks`.
         """
         if action in self._executing_tasks and self._executing_tasks[action]:
             LOGGER.warning("Multiple instances of background task %s are executing simultaneously", action.__name__)
         self._executing_tasks[action].append(task)
+        if not cancel_on_stop:
+            self._await_tasks[action].append(task)
```
arnaudsjs, line 250: The tasks in the `_await_tasks` dictionary are not properly cleaned up when the task finishes execution. As such, they will grow until infinity.

arnaudsjs (changes requested):

arnaudsjs (approved):

wouterdb (approved):

Issue #4412, opened 2022-06-22
concurrency control and version reservation for partial exports


sanderr:
*This ticket was created as a replacement of #4405 and #4406 after a meeting to discuss the desired approach*

In the current (planned) implementation (#4348) of partial exports (`put_partial`) the usage is very similar to that of the full export: the client first reserves a version, then stores the resources for that version in a second API call (`put_version`). For partial compiles it was concluded that it makes more sense to generate the version dynamically when it is stored. This allows for very intuitive semantics where a partial export is applied relative to the latest version (more advanced options in #4413), and stored as the new latest version. Some concurrency control is required to ensure no (in-scope) race conditions can occur.

There are two (interconnected) parts:
- Concurrency control for resource read and write: `put_version` and `put_partial` both write a new collection of resources to the database. Additionally, `put_partial` also determines which is the current latest version and uses that as a reference to apply the partial on. To prevent `a:read -> b:write -> a:write` races, a lock should be acquired for this logic.
- `put_partial` should no longer accept a `version` parameter. Instead it should reserve a parameter right before applying the partial. This too should be an atomic operation.

In pseudo-code:
```python
with resources_lock():  # must also be acquired by `put_version`, can be Python lock or database lock
    with version_lock():  # must also be acquired by reserve version endpoint -> lock is probably already there implicitly because of transactional operation
        version = env.reserve_version()
    old_resources = get_resources(get_latest_version())  # read within lock
    resources = apply_diff(old_resources, partial_resources)
    store_version(version, resources)  # write within lock
```
And drop the `version` argument from `put_partial`.

# Tasks
- [x] Implement the change as described above
- [x] Document any remaining limitations and/or responsibilities of the caller/exporter with respect to potential race conditions
- [x] Remove mentions of experimental in `--partial` and `put_partial` help and docstring respectively (see #4631)
- [x] Store which version is used as the base version (or create a ticket)
- [ ] Add change entry for partial export

PR #4578, opened 2022-07-20
disabled put_partial method to exclude it from release

sanderr:
# Description

Further discussion is required on #4412. To unblock the release I propose to temporarily disable this method.

If this is rejected, the follow-up ticket #4579 should be closed.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


arnaudsjs:
Removing the endpoint will cause issues as the CLI tools depend on it. We will go for the approach where we keep all the functionality but mark the partial compile feature as experimental (See: #4612)

PR #4597, opened 2022-07-26
Issue/4455: removed resource sets are present in partial diff

FloLey:
# Description

When a partial export is done with --remove-resource-sets myset but myset is still present in the resources that are exported, the server does not reject this. It should. This behavior should be well defined and documented.

Corner case: if the set does exist in the list but it is empty, there's no need to reject it.

closes https://github.com/inmanta/inmanta-core/issues/4455

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/agent_server/test_resource_sets.py
+++ b/tests/agent_server/test_resource_sets.py
@@ -993,7 +993,7 @@ async def test_put_partial_version(server, client, environment, clienthelper):
 
 async def test_put_partial_removed_rs_in_rs(server, client, environment, clienthelper):
     """
-    A test that throws and exception if a removed resource set is still present int the resourcesets
+    Test that an exception is thrown when a resource being exported belongs to a resource set that is being deleted.
     """
     version = await clienthelper.get_version()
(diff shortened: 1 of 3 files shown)


tests/agent_server/test_resource_sets.py
```diff
@@ -989,3 +989,83 @@ async def test_put_partial_version(server, client, environment, clienthelper):
         "test::Resource[agent1,key=key1],v=3 does not match the version argument "
         "(version: 2)"
     )
+
+
+async def test_put_partial_removed_rs_in_rs(server, client, environment, clienthelper):
+    """
+    A test that throws and exception if a removed resource set is still present int the resourcesets
```
arnaudsjs, line 996:
```suggestion
    Test that an exception is thrown when a resource being exported belongs to a resource set that is being deleted.
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #4603, opened 2022-07-28
Issue/927 add env settings

FloLey:
# Description

add the possibility to add extra settings to the env from outside of inmanta-core.

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -2430,6 +2430,9 @@ RETURNING last_version;
     async def register_setting(self, setting: Setting) -> None:
         """
         Adds a new setting in this environment.
+        This can be used to add a setting to the env from outside inmanta-core.
+        For example, inmanta-lsm can use this methode to add settings that are only
+        relevant for inmanta-lsm but that are needed in the environment.
 
         :param setting: the setting that should be added to the existing settings
         """
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -2431,7 +2431,7 @@ RETURNING last_version;
         """
         Adds a new setting in this environment.
         This can be used to add a setting to the env from outside inmanta-core.
-        For example, inmanta-lsm can use this methode to add settings that are only
+        As example, inmanta-lsm can use this methode to add settings that are only
         relevant for inmanta-lsm but that are needed in the environment.
 
         :param setting: the setting that should be added to the existing settings
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -2427,6 +2427,17 @@ RETURNING last_version;
         self.last_version = version
         return version
 
+    async def add_new_setting(self, setting: Setting) -> None:
+        """
+        Adds a new setting in this environment.
+
+        :param setting: the setting that should be added to the existing settings
+        """
+        assert isinstance(setting, Setting)
+        if setting.name in self._settings:
+            raise KeyError()
+        self._settings[setting.name] = setting
+
     @classmethod
     async def get_list(
         cls: Type[TBaseDocument],
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -2433,7 +2433,6 @@ RETURNING last_version;
 
         :param setting: the setting that should be added to the existing settings
         """
-        assert isinstance(setting, Setting)
         if setting.name in self._settings:
             raise KeyError()
         self._settings[setting.name] = setting
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -2427,17 +2427,18 @@ RETURNING last_version;
         self.last_version = version
         return version
 
-    async def register_setting(self, setting: Setting) -> None:
+    @classmethod
+    async def register_setting(cls, setting: Setting) -> None:
         """
-        Adds a new setting to this environment from outside inmanta-core.
+        Adds a new setting to the environments from outside inmanta-core.
         As example, inmanta-lsm can use this methode to add settings that are only
-        relevant for inmanta-lsm but that are needed in the environment.
+        relevant for inmanta-lsm but that are needed in the environments.
 
         :param setting: the setting that should be added to the existing settings
         """
-        if setting.name in self._settings:
+        if setting.name in cls._settings:
             raise KeyError()
-        self._settings[setting.name] = setting
+        cls._settings[setting.name] = setting
 
     @classmethod
     async def get_list(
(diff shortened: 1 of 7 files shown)


src/inmanta/data/__init__.py
```diff
@@ -2427,6 +2427,17 @@ async def get_next_version(self) -> int:
         self.last_version = version
         return version
 
+    async def add_new_setting(self, setting: Setting) -> None:
+        """
+        Adds a new setting in this environment.
+
+        :param setting: the setting that should be added to the existing settings
+        """
+        assert isinstance(setting, Setting)
```
FloLey, line 2436: I don't know if those checks are really needed or if something else should be checker

tests/test_data.py
```diff
@@ -347,6 +347,37 @@ async def test_environment_set_setting_parameter(init_dataclasses_and_load_schem
...
+        default=False,
+        typ="bool",
+        validator=convert_boolean,
+        doc="a new setting",
+    )
+
+    await env.register_setting(new_setting)
+    assert (await env.get("a new boolean setting")) is False
+    await env.set("a new boolean setting", True)
+    assert (await env.get(data.AUTO_DEPLOY)) is True
```
arnaudsjs, line 368: I would verify this actions via the API instead.

src/inmanta/data/__init__.py
arnaudsjs: Indeed. This is not really needed. If we violate this constraint, the type checker should warn us.

arnaudsjs (changes requested):

tests/test_data.py
FloLey: you mean using `await client.list_settings(tid=environment_default)`? why do you prefer that?

arnaudsjs: Yes. More specifically `client.get_setting` and `client.set_setting`. It's better to use the API, because API's are intended to remain stable over time, which prevents changes to the test case in the future. Next to that it also turn it into an end-to-end test. And it's not really more work to use the API instead.

src/inmanta/data/__init__.py
```diff
@@ -2427,6 +2427,16 @@ async def get_next_version(self) -> int:
         self.last_version = version
         return version
 
+    async def register_setting(self, setting: Setting) -> None:
+        """
+        Adds a new setting in this environment.
```
arnaudsjs, line 2432: I would also explain the use case for this method.

arnaudsjs (changes requested):
Don't we also need a `register_setting()` method in the `SLICE_ENVIRONMENT`, that can be used by the lsm extension to register the setting?

FloLey:
> Don't we also need a `register_setting()` method in the `SLICE_ENVIRONMENT`, that can be used by the lsm extension to register the setting?

@arnaudsjs 
is this what you mean? I think there was indeed something missing because I didn't find how to add my setting in inmanta-lsm :p

src/inmanta/data/__init__.py
```diff
@@ -2427,6 +2427,18 @@ async def get_next_version(self) -> int:
         self.last_version = version
         return version
 
+    async def register_setting(self, setting: Setting) -> None:
+        """
+        Adds a new setting to this environment from outside inmanta-core.
+        As example, inmanta-lsm can use this methode to add settings that are only
```
arnaudsjs, line 2433:
```suggestion
        As example, inmanta-lsm can use this method to add settings that are only
```

src/inmanta/server/services/environmentservice.py
```diff
@@ -601,3 +601,18 @@ async def notify_listeners(
                     await listener.environment_action_updated(updated_env, original_env)
             except Exception:
                 LOGGER.warning(f"Notifying listener of {action} failed with the following exception", exc_info=True)
+
+    async def register_setting(self, env: data.Environment, setting: Setting) -> None:
+        """
+        Should only be called during pre-start
+        Adds a new setting to this environment from outside inmanta-core.
+        As example, inmanta-lsm can use this methode to add settings that are only
```
arnaudsjs, line 609:
```suggestion
        As example, inmanta-lsm can use this method to add settings that are only
```

src/inmanta/server/services/environmentservice.py
```diff
@@ -601,3 +601,18 @@ async def notify_listeners(
...
+        relevant for inmanta-lsm but that are needed in the environment.
+
+        :param env: the environment in which the setting will be registered
+        :param setting: the setting that should be added to the existing settings
+        """
+        try:
+            await env.register_setting(setting)
+        except KeyError:
+            raise
```
arnaudsjs, line 618: This except-clause is redundant.
```suggestion
        await env.register_setting(setting)
```

tests/agent_server/test_server_agent.py
```diff
@@ -719,6 +727,26 @@ async def test_get_set_param(resource_container, environment, client, server):
     assert result.code == 200
 
 
+async def test_register_setting(resource_container, environment, client, server):
+    """
+    Test getting and setting params
+    """
+    resource_container.Provider.reset()
```
arnaudsjs, line 734: I don't think you need the `resource_container`.
```suggestion
async def test_register_setting(environment, client, server):
    """
    Test registering a new setting.
    """
```

arnaudsjs (changes requested):

arnaudsjs (approved):

wouterdb (changes requested):
I think it would be good to test what happens when a setting exists in the database for which the definition no longer exists. 


(e.g we start the server with all extensions disabled). 

I would expect the api to not return any unknown settings, but just keep them in the database.

tests/server/test_env_settings.py
```diff
@@ -358,3 +359,19 @@ async def test_environment_add_new_setting_parameter(server, client, environment
...
+    env_id = UUID(environment)
+    env = await data.Environment.get_by_id(env_id)
+    project_id = env.project
+    setting_db_query = "UPDATE environment SET settings=jsonb_set(settings, $1::text[], to_jsonb($2::boolean), TRUE) WHERE name=$3 AND project=$4"
+    values = [["a new setting"], True, "dev", project_id]
+    await Environment._execute_query(setting_db_query, *values)
+
+    result = await client.get_setting(tid=environment, id="a new setting")
+    assert result.code == 404
+    assert result.result["message"] == "Request or referenced resource does not exist"
```
wouterdb, line 377: Could you also verify it is not returned by a listing either and that if you register the option, it is returned by both?

wouterdb (changes requested):

wouterdb (approved):

PR #4717, opened 2022-08-23
Ensure that the test suite cleans up editable installs

arnaudsjs:
# Description

This PR:

* Ensure that `sys.path_hooks`, `sys.path_importer_cache.clear` and `sys.modules` are properly reset after each test case.
* Ensure that modules installed in editable more are correctly removed from `sys.modules`.
* Consolidate the different method that unload modules from `sys.modules`

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -443,6 +443,7 @@ def deactive_venv():
     old_prefix = sys.prefix
     old_path = list(sys.path)
     old_meta_path = sys.meta_path.copy()
+    old_path_hooks = sys.path_hooks.copy()
     old_pythonpath = os.environ.get("PYTHONPATH", None)
     old_os_venv: Optional[str] = os.environ.get("VIRTUAL_ENV", None)
     old_process_env: str = env.process_env.python_path
@@ -459,6 +460,10 @@ def deactive_venv():
     # reset sys.meta_path because it might contain finders for editable installs, make sure to keep the same object
     sys.meta_path.clear()
     sys.meta_path.extend(old_meta_path)
+    sys.path_hooks.clear()
+    sys.path_hooks.extend(old_path_hooks)
+    # Clear cache for sys.path_hooks
+    sys.path_importer_cache.clear()
     pkg_resources.working_set = old_working_set
     # Restore PYTHONPATH
     if old_pythonpath is not None:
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -459,6 +460,13 @@ def deactive_venv():
     # reset sys.meta_path because it might contain finders for editable installs, make sure to keep the same object
     sys.meta_path.clear()
     sys.meta_path.extend(old_meta_path)
+    sys.path_hooks.clear()
+    sys.path_hooks.extend(old_path_hooks)
+    # Clear cache for sys.path_hooks
+    sys.path_importer_cache.clear()
+    sys.modules.clear()
+    sys.modules.update(old_sys_modules)
```
sanderr, line 468: I've considered this in the past but I decided against it because it requires reloading everything, even submodules of the `inmanta-core` package, test modules, third party libraries, ... I wonder it wouldn't be better to be a bit more directed here and just keep the `unload_inmanta_plugins` and `unload_modules_for_path` calls (but consolidated and improved as you did). Wdyt?

sanderr (changes requested):
One question/suggestion, otherwise looks good to me.

tests/conftest.py
arnaudsjs: Fine by me.

tests/conftest.py
```diff
@@ -972,6 +975,7 @@ def setup_func(self, module_dir: Optional[str]):
         self._keep = False
         self.project_dir = tempfile.mkdtemp()
         self.modules_dir = module_dir
+        loader.unload_modules_for_path(self.env)
```
sanderr, line 978: Shouldn't this be in `tearDownClass`? Normal `snippetcompiler` has a shared venv.

sanderr (changes requested):

tests/conftest.py
arnaudsjs: But not every test case is using the `snippetcompiler`, so shouldn't we unload the modules after each test case (`tear_down_func`)? Some also use the `snippetcompiler_clean` or one of the `tmpvenv*` fixtures.

sanderr: Right. We could put them in both the set up and tear down for the class but I admit that's a bit cumbersome. So I agree now that this is probably the best approach.

sanderr (approved):

PR #4725, opened 2022-08-24
Issue/4402 resource actions joint table

wouterdb:
# Description

- Added resource action join table to the read path
- I didn't remove the resource_version_ids column from the ResourceAction table. It is not super heavily used, but we have no easy way to support reconstructing the field to maintain a stable external API. If the removal is deemed necessary, I would not add it to this PR, which has take long enough as it is. 

Performance impact on `resource_did_dependency_change` seems to be very large on large databases, over 25 calls, in seconds

|  | avg | 99.9% | 
|--|------|----------|
|iso5| 3.6 | 3.7|
| this PR | 0.004 | 0.011 |

# Self Check:

- [x] Changelog entry


tests/server/resource_deploy_performance.py
```diff
@@ -103,214 +108,265 @@ async def run_batch(index):
...
+              ) ? r1.resource_version_id
+    """
+    async with ResourceAction.get_connection() as con:
+        stmt = await con.prepare(query_2)
+        print(json.dumps(await stmt.explain(env_id, 2, "test::Resource[agent1,key=key3],v=1", analyze=True)))
+
 
 """
-Baseline output
+Baseline output iso5-stable
```
arnaudsjs, line 154: I think we should add a clarifying comment on what the purpose of this baseline is.

tests/server/resource_deploy_performance.py
```diff
@@ -103,214 +108,265 @@ async def run_batch(index):
 
     print(json.dumps(global_registry().dump_metrics(), indent=4))
 
+    key_query = """
```
arnaudsjs, line 111: What is the goal of these key queries?

src/inmanta/data/__init__.py
```diff
@@ -1368,9 +1412,27 @@ async def get_by_id(
             return result[0]
         return None
 
+    @classmethod
+    async def lock_by_id(
```
arnaudsjs, line 1416: This method is not used anywhere, is it?

arnaudsjs (changes requested):

tests/server/resource_deploy_performance.py
wouterdb: I removed it

tests/server/resource_deploy_performance.py
wouterdb: I removed it

arnaudsjs (approved):

src/inmanta/data/__init__.py
```diff
@@ -4203,6 +4277,116 @@ async def query_resource_actions(
...
+        cls, env: Environment, resource_id: "resources.Id", exclude_change: Optional[const.Change] = None
+    ) -> Dict[ResourceIdStr, List["ResourceAction"]]:
+        """
+        Get all events that should be processed by this specific resource, for the current deployment
+
+        This method searches across versions!
+
+        This means:
+        1. assure a deployment is ongoing
+        2. get the time range between the start of this deployment and the last successfull deploy
```
sanderr, line 4291:
```suggestion
        2. get the time range between the start of this deployment and the last successful deploy
```

src/inmanta/data/__init__.py
```diff
@@ -4203,6 +4277,116 @@ async def query_resource_actions(
...
+                FROM public.resourceaction_resource as jt
+                    INNER JOIN public.resourceaction as ra
+                        ON ra.action_id = jt.resource_action_id
+                    WHERE jt.environment=$1 AND ra.environment=$1 AND jt.resource_id=$2::varchar AND ra.action='deploy'
+                    ORDER BY ra.started DESC
+            )
+        SELECT
+            (SELECT started from base_ra ORDER BY started DESC LIMIT 1) as begin_started,
+            (SELECT status from base_ra ORDER BY started DESC LIMIT 1) as begin_status,
+            COALESCE((SELECT started from base_ra where status='deployed' ORDER BY started DESC LIMIT 1), NULL) as started
```
sanderr, line 4328: Why the duplicate `ORDER BY` statements here? Is this negligible if it is already sorted?

src/inmanta/data/__init__.py
```diff
@@ -4203,6 +4277,116 @@ async def query_resource_actions(
...
+        """
+
+        async with cls.get_connection() as connection:
+            result = await connection.fetchrow(end_query, env.id, resource_id.resource_str())
+
+            if not result or result["begin_status"] is None:
+                raise BadRequest(
```
sanderr, line 4335: I don't think we should raise `inmanta.protocol` exceptions here.

src/inmanta/data/__init__.py
```diff
@@ -4203,6 +4277,116 @@ async def query_resource_actions(
...
+            if result["begin_status"] != const.ResourceState.deploying:
+                raise BadRequest(
+                    "Fetching resource events only makes sense when the resource is currently deploying. Current deploy state"
+                    f" for resource {resource_id_str} is {result['begin_status']}."
+                )
+            current_deploy_start = result["begin_started"]
+            last_deploy_start = result["started"]
+
+            # Step3: Get the resource
+            resource: Optional[Resource] = await Resource.get_one(environment=env.id, resource_version_id=resource_id_str)
```
sanderr, line 4348: This is not run on the same connection, is that intentional? Since you don't start a transaction, any reason you create a connection in the first place rather than just using `cls._fetchrow()` and `cls._fetchquery()`?

src/inmanta/data/__init__.py
```diff
@@ -4314,14 +4498,10 @@ async def get_resources(
         """
         if not resource_version_ids:
             return []
-        resource_version_ids_statement = ", ".join(["$" + str(i) for i in range(2, len(resource_version_ids) + 2)])
-        (filter_statement, values) = cls._get_composed_filter(environment=environment)
-        values = values + cls._get_value(resource_version_ids)
-        query = (
-            f"SELECT * FROM {cls.table_name()} "
-            f"WHERE {filter_statement} AND resource_version_id IN ({resource_version_ids_statement})"
+        query = f"SELECT * FROM {cls.table_name()} " f"WHERE environment=$1 AND resource_version_id = ANY($2)"
```
sanderr, line 4501:
```suggestion
        query = f"SELECT * FROM {cls.table_name()} WHERE environment=$1 AND resource_version_id = ANY($2)"
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -68,6 +68,9 @@
 
 LOGGER = logging.getLogger(__name__)
 
+PERFORM_CLEANUP = True
```
sanderr, line 71:
```suggestion
PERFORM_CLEANUP: Literal[True] = True
```
If I understand correctly this should always be true except for temporary use cases?

sanderr (approved):
Some very minor suggestions only.

sanderr (changes requested):
I forgot about the `inmanta.protocol` remark, that one is an actual change request.

src/inmanta/data/__init__.py
```diff
@@ -4134,13 +4205,16 @@ async def query_resource_actions(
...
+        query = """SELECT DISTINCT ra.*
+                    FROM public.resource as r
+                    INNER JOIN public.resourceaction_resource as jt
+                        ON r.environment = jt.environment
+                        AND r.resource_id = jt.resource_id
+                        AND r.model = jt.resource_version
+                    INNER JOIN public.resourceaction as ra
+                        ON ra.action_id = jt.resource_action_id
+                        WHERE r.environment=$1 AND ra.environment=$1"""
+        values: list[object] = [cls._get_value(environment)]
```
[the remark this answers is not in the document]
wouterdb: I agree that this is in line with the other code, but I also think it reduces readability. 

It is not like any other value would work.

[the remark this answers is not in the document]
sanderr: Fair enough.

src/inmanta/data/__init__.py
wouterdb: I did it to be sure, assuming it to be negligible.

src/inmanta/data/__init__.py
wouterdb: I agree, but failed to come up with a better solution or rationale why this would be bad.

src/inmanta/server/services/orchestrationservice.py
wouterdb: yes, but I'm going to make it a bool anyways

src/inmanta/data/__init__.py
sanderr: That's an interesting observation. I was wondering about this yesterday but I don't think I saw the full implications in this context. You could maybe even deadlock (not PG deadlock but Python deadlock) on this: N methods calls (to the same or different methods) having acquired a lock, then waiting to acquire another one for a query for which the connection object wasn't wired through but that lives (as far as Python is concerned) within the context of the acquired connection.

src/inmanta/server/services/orchestrationservice.py
sanderr: I can live with that.

sanderr (approved):

PR #4729, opened 2022-08-24
when converting v1 to v2, transform build tags

sanderr:
# Description

When converting a v1 module to v2, if any version tags are present, don't put them in the version field but in the `tag_build` field. This prevents artifacts with versions like `2.11.1.dev0.dev202208182021`.

(In hindsight I probably should've ticketized this rather than to implement a fix, but what's done is done.)

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,6 +13,9 @@ charset = utf-8
 [*.py]
 max_line_length = 128
 
+[*.yml]
+indent_size = 2
+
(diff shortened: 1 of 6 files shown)


arnaudsjs (approved):
Apart from a missing changelog file. It looks ok to me.

PR #4757, opened 2022-08-30
fixed put_partial bug rejecting existing shared resources

sanderr:
# Description

Fixed bug where including a shared resource that already exists in a partial would be rejected because the version was not excluded from the update check comparison.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/put-partial-shared-resource-in-partial.yml
+++ b/changelogs/unreleased/put-partial-shared-resource-in-partial.yml
@@ -1,0 +1,5 @@
+description: Fixed bug where including a shared resource that already exists in a partial would be rejected
+change-type: patch
+destination-branches:
+  - master
+  - iso5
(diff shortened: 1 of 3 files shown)


arnaudsjs:
I noticed that my PR (https://github.com/inmanta/inmanta-core/pull/4759) contains a fix for exactly the same bug. As this PR contains a subset of the fixes present in my PR, I guess it's best to drop this PR. Do you agree?

sanderr:
I do. Do you also have a test for it? If not, perhaps include the one here.

arnaudsjs:
> I do. Do you also have a test for it? If not, perhaps include the one here.

I had a test that verifies whether all the resources have the expected values, but no test for the specific scenario you consider here. I added your test to my PR.

PR #4776, opened 2022-09-01
install_from_index should override PIP_EXTRA_INDEX_URL

### LINKED ISSUE #4723: install_from_index should override PIP_EXTRA_INDEX_URL
The contract of `env.PythonEnvironment.install_from_index` is that it uses the index urls it receives as arguments (for project installs these are the index urls in the `project.yml`). While it does set `--index-url` and `--extra-index-url`, it does not set `PIP_EXTRA_INDEX_URL=""`, so if it is set in the environment, it is leaked to the pip install.

FloLey:
# Description

The contract of env.PythonEnvironment.install_from_index is that it uses the index urls it receives as arguments (for project installs these are the index urls in the project.yml). While it does set --index-url and --extra-index-url, it does not set PIP_EXTRA_INDEX_URL="", so if it is set in the environment, it is leaked to the pip install. This commit will prevent this leak.

closes #4723

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/4723-fix--override-pip-extra-index-url.yml
+++ b/changelogs/unreleased/4723-fix--override-pip-extra-index-url.yml
@@ -1,6 +1,8 @@
 ---
-description: make sure that the index present in PIP_EXTRA_INDEX_URL is not leaked to the pip install_from_index
+description: make sure that the index present in PIP_INDEX_URL or PIP_EXTRA_INDEX_URL is not leaked to pip when using install_from_index
 issue-nr: 4723
 issue-repo: inmanta-core
 change-type: patch
 destination-branches: [master, iso5]
+sections:
+  bugfix: "{{description}}"
--- a/changelogs/unreleased/4723-fix--override-pip-extra-index-url.yml
+++ b/changelogs/unreleased/4723-fix--override-pip-extra-index-url.yml
@@ -1,0 +1,6 @@
+---
+description: make sure that the index present in PIP_EXTRA_INDEX_URL is not leaked to the pip install_from_index
+issue-nr: 4723
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master, iso5]
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -474,8 +474,9 @@ class PythonEnvironment:
         process_env = os.environ.copy()
         if index_urls is not None and "PIP_EXTRA_INDEX_URL" in process_env:
             del process_env["PIP_EXTRA_INDEX_URL"]
-        if "PIP_INDEX_URL" in process_env and index_urls is not None:
+        if index_urls is not None and "PIP_INDEX_URL" in process_env:
             del process_env["PIP_INDEX_URL"]
+
         return_code, full_output = self.run_command_and_stream_output(cmd, env_vars=process_env)
 
         if return_code != 0:
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -470,7 +470,10 @@ class PythonEnvironment:
             requirements_files=requirements_files,
         )
         process_env = os.environ.copy()
-        del process_env["PIP_EXTRA_INDEX_URL"]
+        if "PIP_EXTRA_INDEX_URL" in process_env:
+            del process_env["PIP_EXTRA_INDEX_URL"]
+        if "PIP_INDEX_URL" in process_env:
+            del process_env["PIP_INDEX_URL"]
         return_code, full_output = self.run_command_and_stream_output(cmd, env_vars=process_env)
 
         if return_code != 0:
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -522,7 +522,6 @@ class PythonEnvironment:
     ) -> None:
         if len(requirements) == 0:
             raise Exception("install_from_index requires at least one requirement to install")
-
         constraint_files = constraint_files if constraint_files is not None else []
         inmanta_requirements = self._get_requirements_on_inmanta_package()
         self._run_pip_install_command(
@@ -606,8 +605,6 @@ class PythonEnvironment:
         Similar to the _run_command_and_log_output method, but here, the output is logged on the fly instead of at the end
         of the sub-process.
         """
-        print("===================")
-        print(env_vars)
         process = subprocess.Popen(
             cmd,
             stdout=subprocess.PIPE,
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -469,10 +469,12 @@ class PythonEnvironment:
             constraints_files=constraints_files,
             requirements_files=requirements_files,
         )
+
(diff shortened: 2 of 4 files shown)


FloLey:
I think my test is a bit counter intuitive but it does the job. Still I'm open for other ideas I you have one.

src/inmanta/env.py
```diff
@@ -469,7 +469,9 @@ def _run_pip_install_command(
             constraints_files=constraints_files,
             requirements_files=requirements_files,
         )
-        return_code, full_output = self.run_command_and_stream_output(cmd)
+        process_env = os.environ.copy()
+        del process_env["PIP_EXTRA_INDEX_URL"]
```
arnaudsjs, line 473: Don't we also have to remove thee `PIP_INDEX_URL` environment variable as well?

tests/moduletool/test_install.py
```diff
@@ -780,6 +780,55 @@ def test_project_install_requirement_not_loaded(
...
+    """
+    Test that PIP_EXTRA_INDEX_URL is not set in the subprocess doing an install_from_index
+    and that it is not changed in the active env.
+
+    The installation fails with an ModuleNotFoundException
+    as the index https://pypi.org/simple is needed to install lorem~=0.0.1,
+    but it is only present in the active env in PIP_EXTRA_INDEX_URL which is not know by the
+    subprocess doing the pip install.
+    """
+    os.environ["PIP_EXTRA_INDEX_URL"] = "https://pypi.org/simple"
```
arnaudsjs, line 797: Setting environment variables should be done using the monkeypatch fixture to make sure the original values get restored at the end of the test case.

arnaudsjs (changes requested):

arnaudsjs (approved):

changelogs/unreleased/4723-fix--override-pip-extra-index-url.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: make sure that the index present in PIP_EXTRA_INDEX_URL is not leaked to the pip install_from_index
```
arnaudsjs, line 2: I think this bug exists in the stable release. Shouldn't we add a bugfix changelog entry?

arnaudsjs (changes requested):

tests/moduletool/test_install.py
```diff
@@ -780,6 +780,58 @@ def test_project_install_requirement_not_loaded(
...
+    """
+    Test that PIP_EXTRA_INDEX_URL/PIP_INDEX_URL is not set in the subprocess doing an install_from_index
+    and that it is not changed in the active env.
+
+    The installation fails with an ModuleNotFoundException
+    as the index https://pypi.org/simple is needed to install lorem~=0.0.1,
+    but it is only present in the active env in PIP_EXTRA_INDEX_URL/PIP_INDEX_URL which is not know by the
+    subprocess doing the pip install.
+    """
+    monkeypatch.setenv(env_var, "https://pypi.org/simple")
```
sanderr, line 800: This is not safe: if someone were to upload a malicious `inmanta-module-minimalv2module` package to pypi its `setup.py` would get executed on our Jenkins hosts and developer machines. So either we need to do this test with a module we have ownership over, or we must not use a public index, e.g. the `local_module_package_index` (name could be off, didn't check) fixture.

sanderr (changes requested):

src/inmanta/env.py
```diff
@@ -469,7 +469,12 @@ def _run_pip_install_command(
             constraints_files=constraints_files,
             requirements_files=requirements_files,
         )
-        return_code, full_output = self.run_command_and_stream_output(cmd)
+        process_env = os.environ.copy()
+        if "PIP_EXTRA_INDEX_URL" in process_env:
+            del process_env["PIP_EXTRA_INDEX_URL"]
+        if "PIP_INDEX_URL" in process_env:
+            del process_env["PIP_INDEX_URL"]
```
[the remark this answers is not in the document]
arnaudsjs: Fine by me.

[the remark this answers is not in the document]
FloLey: @sanderr I'm not sure I understand what I should write as test for `install_from_source`? just verify `install_from_source` still works after my change? I mean there already are test using `ModuleTool().install`. Or is there something more to test?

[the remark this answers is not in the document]
sanderr: Test that it still takes `PIP_INDEX_URL` etc into account. Then again, the tests you're referring to would probably fail as well if we disabled this so you can drop it.

sanderr (approved):
Looks good to me now, the only thing that remains now is what we just discussed about the pip config file.

tests/moduletool/test_install.py
```diff
@@ -780,6 +780,58 @@ def test_project_install_requirement_not_loaded(
...
+
+    # set up project
+    snippetcompiler_clean.setup_for_snippet(
+        f"""
+        import {module.ModuleV2.get_name_from_metadata(v2mod1)}
+        """,
+        autostd=False,
+        install_project=False,
+        python_package_sources=["unknown"],
+        python_requires=[
```
arnaudsjs, line 821:
```suggestion
        install_project=False,
        # Installing a V2 module requires a python package source.
        python_package_sources=["unknown"],
        python_requires=[
```

src/inmanta/env.py
```diff
@@ -469,7 +469,15 @@ def _run_pip_install_command(
             constraints_files=constraints_files,
             requirements_files=requirements_files,
         )
-        return_code, full_output = self.run_command_and_stream_output(cmd)
+
+        # if index_urls are set, only use those. Otherwise, use the one from the environment
+        process_env = os.environ.copy()
```
arnaudsjs, line 474: Naming that variable `process_env` is a bit confusing. The `env.py` has a global variable that's also called `process_env`, which refers to the Python environment of this process. It would be good to rename this one.

arnaudsjs (changes requested):

src/inmanta/env.py
```diff
@@ -471,13 +471,16 @@ def _run_pip_install_command(
...
         # if index_urls are set, only use those. Otherwise, use the one from the environment
-        process_env = os.environ.copy()
-        if index_urls is not None and "PIP_EXTRA_INDEX_URL" in process_env:
-            del process_env["PIP_EXTRA_INDEX_URL"]
-        if index_urls is not None and "PIP_INDEX_URL" in process_env:
-            del process_env["PIP_INDEX_URL"]
+        sub_env = os.environ.copy()
+        if index_urls is not None:
+            # setting this env_var to os.devnull disables the loading of all pip configuration files
+            sub_env["PIP_CONFIG_FILE"] = os.devnull
```
sanderr, line 478: Yes, way better than what we came up with yesterday!

sanderr (approved):

arnaudsjs (approved):

PR #4779, opened 2022-09-01
Uninstall inmanta packages from compiler service on update

### LINKED ISSUE #4718: server should uninstall inmanta packages before update
# Description
Since `inmanta modules update` protects the inmanta package versions that are currently installed, if any are installed in the venv itself (rather than in its parent environment), it will never be updated and it will always shadow the parent. To prevent this, when the server does an update, it should first uninstall any inmanta packages that live in the venv.

- [ ] implement the change
- [ ] add an upgrade note stating that an update & recompile is required. Alternatively, we can do the uninstall on any compile but that seems expensive.

# Rationale
The following is copied from the initial post about this issue on [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1660662318032809?thread_ts=1660662318.032809&cid=CKRF0C8R3)

<details>
  <summary>Slack post</summary>
  
### The issue on the management orchestrator
The management orchestrator runs on the following version:
```
Inmanta Service Orchestrator (Standard Edition): 5.1.0
Compiler version: 2022.2
Extensions:
    * lsm: 2.1.0
    * ui: 2.0.2
    * support: 2.0.2
    * license: 2.0.2
    * core: 6.1.0
```
This version does not yet have the fix that protects the inmanta packages against upgrading when `pip install` is run. The newly released iso5 does have this. So what happens now that OSS+iso5 have been released is that the management orchestrator's compiler venv runs a `pip install -U --upgrade-strategy eager ...` which upgrades `inmanta-core` to `6.2.0` in the compiler venv. As a side effect the compiler determines it is now compatible with the latest lsm (which has a lower bound constraint on the compiler version). In practice it is not because it actually requires `inmanta-lsm-2.1.0`, which the venv cannot upgrade to because (I'm assuming) it uses PyPi and the package is not available there. Therefore the management orchestrator is left in a broken state

### The issue for production orchestrators running iso 5.1.0
The first stage of the issue will be the same: when "update & recompile" is triggered the eager upgrade will upgrade the inmanta packages in the venv. As a result, the model will be compiled with the latest inmanta-core (within the iso5 range apparently, I'm not sure exactly what constrains this) rather than the version installed in `/opt/inmanta`. While this is definitely not good, it is contained to the compiler, and it will not actually cause the orchestrator to fail because production environments will have all module versions pinned.

### The issue for non-production orchestrators running iso 5.1.0
If the pip index used for the pip install done by the compiler has access to the `inmanta-lsm` package, it will be installed and the lsm module will be upgraded to the latest version. The root of the issue is the same as for production orchestrators. If however inmanta-lsm is not available the root of the issue becomes the same as that of the management orchestrator.

## Long-term consequences
Because the newly released version does have a protection against this sort of behavior, this can not happen again. However, when we release iso 5.3.0 (or 5.2.1), if this issue has already presented itself by then, the compiler venv will have an `inmanta-core` that shadows the one in `/opt/inmanta`. The protection we put in place will then prevent that version to be upgraded. Therefore the compiler venv will be stuck on 6.2.0 for all eternity. There are two conclusions to be drawn here
1. we need to make sure we allow the upgrade to proceed as planned
2. preferably we do so in a way that also works if a human at some point manually shadows any inmanta packages in the compiler environment.
Options that are currently under consideration (but it hasn't been given a lot of thought yet) are:
- have the server perform an uninstall on the inmanta packages in the compiler venv before proceeding
- have the server pass explicit versions of the inmanta packages to the update command rather than having the compiler auto-detect them

</details>

arnaudsjs:
# Description

Make the compiler service remove inmanta packages from the compiler venv on update.

closes #4718 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -43,7 +43,7 @@ from inmanta import config, const, data, protocol, server
 from inmanta.config import Config
 from inmanta.data import APILIMIT, InvalidSort, QueryType
 from inmanta.data.paging import CompileReportPagingCountsProvider, CompileReportPagingHandler, QueryIdentifier
-from inmanta.env import PythonEnvironment, VenvCreationFailedError, VirtualEnv
+from inmanta.env import PythonEnvironment, VenvCreationFailedError, VirtualEnv, PipCommandBuilder
 from inmanta.protocol import encode_token, methods, methods_v2
 from inmanta.protocol.common import ReturnValue
 from inmanta.protocol.exceptions import BadRequest, NotFound
@@ -271,11 +271,19 @@ class CompileRun(object):
                 else:
                     return await self._end_stage(returncode=0)
 
+            async def uninstall_protected_inmanta_packages() -> data.Report:
+                """
+                Ensure that no protected Inmanta packages are installed in the compiler venv.
+                """
+                cmd: List[str] = PipCommandBuilder.compose_uninstall_command(
+                    python_path=PythonEnvironment.get_python_path_for_env_path(venv_dir),
+                    pkg_names=PythonEnvironment.get_protected_inmanta_packages(),
+                )
+                return await self._run_compile_stage(
+                    name="Uninstall inmanta packages from the compiler venv", cmd=cmd, cwd=project_dir
+                )
+
             async def update_modules() -> data.Report:
-                # Make sure that no Inmanta packages are installed in the compiler venv
-                python_env = PythonEnvironment(env_path=venv_dir)
-                python_env.uninstall(pkg_names=PythonEnvironment.get_protected_inmanta_packages())
-                # Update modules
                 return await run_compile_stage_in_venv("Updating modules", ["-vvv", "-X", "project", "update"], cwd=project_dir)
 
             async def install_modules() -> data.Report:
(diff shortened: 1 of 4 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -272,6 +272,10 @@ async def ensure_venv() -> Optional[data.Report]:
                     return await self._end_stage(returncode=0)
 
             async def update_modules() -> data.Report:
+                # Make sure that no Inmanta packages are installed in the compiler venv
+                python_env = PythonEnvironment(env_path=venv_dir)
+                python_env.uninstall(pkg_names=PythonEnvironment.get_protected_inmanta_packages())
```
arnaudsjs, line 277: Do I turn this into a separate stage of the compiler service or do we do it like this (in the background)?

sanderr: I hadn't considered we could just do it in the background. I think I have a slight preference for making it a separate stage, but I'm really not sure.

sanderr (approved):

src/inmanta/server/services/compilerservice.py
sanderr: I don't think so, it's still done synchronously, just hidden from the user.

wouterdb: We have to log the output correctly, to the compile report, so this won't fly.

wouterdb (approved):

PR #4785, opened 2022-09-05
Improved assertions on warnings

wouterdb:
# Description

Improved assertions on warnings to prevent warnings to leak from one testcase to another

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (approved):
Approving this, with a tiny caveat: It's probably ok to ignore all warnings coming from inmanta.config and inmanta.util for these tests, but could it be that we silence a warning that should actually make a test fail ?

wouterdb:
@Hugo-Inmanta it could, but I don't think that is the case here.

PR #4788, opened 2022-09-05
Issue/1291 irt archive stdout mypy

FloLey:
# Description

modify tox.ini file to write output of mypy to a file.

related to https://github.com/inmanta/irt/pull/1303
closes [irt/1291](https://github.com/inmanta/irt/issues/1291)

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


tox.ini
```diff
@@ -52,8 +52,9 @@ deps=
     -rrequirements.dev.txt
     -rrequirements.txt
 commands_pre = mkdir -p coverage
-whitelist_externals = */mkdir
+allowlist_externals = */mkdir
+                      */bash
 setenv = MYPYPATH=stubs:src
 commands =
-    python -m mypy --soft-error-limit=-1 --junit-xml mypy.xml --cobertura-xml-report coverage -p inmanta
+    bash -c "python -m mypy --soft-error-limit=-1 --junit-xml mypy.xml --cobertura-xml-report coverage -p inmanta >> mypy_output.txt"
```
sanderr, line 59:
```suggestion
    sh -c "python -m mypy --soft-error-limit=-1 --junit-xml mypy.xml --cobertura-xml-report coverage -p inmanta > mypy_output.txt"
```
Nitpick, but we don't actually need bash, and we'd want to clear the file if it already exists.

tox.ini
```diff
@@ -52,8 +52,9 @@ deps=
     -rrequirements.dev.txt
     -rrequirements.txt
 commands_pre = mkdir -p coverage
-whitelist_externals = */mkdir
+allowlist_externals = */mkdir
+                      */bash
```
sanderr, line 56: Out of curiosity, do you know what the "*/" is for? I know I once used it without and that worked as well. Is it a regex match on the full path of the executable? (tox wiki seems to be down so I can't check)

changelogs/unreleased/1291-irt-archive-mypy-stdout.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: change tox.ini file to store the output ou mypy into a file
```
sanderr, line 2:
```suggestion
description: change tox.ini file to store the output of mypy into a file
```

sanderr (changes requested):
Could you also add the file to the gitignore?

changelogs/unreleased/1291-irt-archive-mypy-stdout.yml
arnaudsjs, line 2:
```suggestion
description: change tox.ini file to store the output of mypy into a file
```

arnaudsjs (changes requested):

tox.ini
FloLey: I tried just "bash" and it was still complaining and in the warning it was showing something about "/usr/bin/bash" so I just tried it like it was done for mkdir and since it didn't complain anymore I was happy to :p

sanderr: All right, guess I was doing something wrong then. Thanks!

sanderr (approved):

PR #4815, opened 2022-09-08
Issue/scheduler backward compatibility

wouterdb:
# Description

Improve backward compatibility of add_action

I did not add the second argument (initial_delay) back, nor did I make the named arguments work.
That would be too messy

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -280,7 +280,7 @@ class Scheduler(object):
     def add_action(
         self,
         action: TaskMethod,
-        schedule: TaskSchedule,
+        schedule: Union[TaskSchedule, int],  # int for backward compatibility,
         cancel_on_stop: bool = True,
     ) -> ScheduledTask:
         """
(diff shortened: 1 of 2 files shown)


src/inmanta/util.py
```diff
@@ -280,7 +280,7 @@ def remove_action_from_task_dict(task_dict: Dict[TaskMethod, List[asyncio.Task[o
     def add_action(
         self,
         action: TaskMethod,
-        schedule: TaskSchedule,
+        schedule: Union[TaskSchedule, int],  # int for backward compatibility,
```
sanderr, line 283: We could optionally make the two signatures a bit more clear by using `typing.overload`.

sanderr (approved):

src/inmanta/util.py
wouterdb: never done that before, let me look into it.

sanderr: Example: https://github.com/inmanta/lsm/blob/26b2a4e33d1da5923bd1fc0adf30053be66e76e3/plugins/allocation_v2/dict_path.py#L384

PR #4823, opened 2022-09-08
Docs/lsm 482 partial compile guidelines

sanderr:
# Description

Added modeling guidelines for safe use of partial compiles. I decided to put it in core rather than lsm because it could pretty much all be generalized and it all applies to core-only use of partial compiles as well.

Initial technical review. In the next stage I'll see if Pouriya is available for a review on the quality of the documentation.

- I had to move some files and GitHub does not show the diff properly. Review is probably easiest with `git diff master --color-moved` or `git diff master --find-renames=5%`

closes inmanta/lsm#482

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/docs/model_developers/resource_sets.rst
+++ b/docs/model_developers/resource_sets.rst
@@ -117,14 +117,16 @@ In this guide, we only cover models where each set of independent resources is d
 will refer to as the "service" or "service entity" (as in ``LSM``). We will use the term "identity" to refer to any set of
 attributes that uniquely identify an instance. In the model this usually corresponds to an index.
 
-When building an inmanta model, all potential instances of a service entity must refine to compatible (low level) config.
-In the model this config is represented by the resources. Therefore these guidelines will focus on creating valid and
+All potential instances of a service entity must be refined to compatible (low level) configuration when creating an Inmanta
+model. In the model this config is represented by the resources. Therefore these guidelines will focus on creating valid and
 compatible resources. With well-designed resources, valid and compatible config will follow.
 
 To safely make use of partial compiles, each service must be the sole owner of its resources and any shared resources must be
 identical across service instances. The graph below pictures a valid service for partial compiles. Each arrow represents a
 refinement: one entity creating another in one of its implementations. The valid service results in fully separate resource
-sets for each instance. Additionally, the one shared resource is created consistently between service instances.
+sets for each instance. Additionally, the one shared resource is created consistently between service instances. For each
+entity type, the ``id`` attribute is assumed to be an identifying attribute for the instance (i.e. there is an index on the
+attribute).
 
 .. digraph:: resource_sets_generic_good
     :caption: A good service for partial compiles.
@@ -315,9 +317,9 @@ because two networks with the same id are defined.
 Ownership
 ---------
 A resource can safely be considered owned by a service instance if it could never be created by another service instance. There
-are two main mechanisms that can be used to provide this guarantee. One is the use of indexes on appropriate locations, the
-other is the use of some external distributor of unique values (e.g. a plugin to generate a UUID or to allocate values in an
-inventory).
+are two main mechanisms that can be used to provide this guarantee, both of which will be described in their own subsection
+below. One is the use of indexes on appropriate locations, the other is the use of some external allocator of unique values
+(e.g. a plugin to generate a UUID or to allocate values in an inventory).
 
 In either case, the goal is to make sure that any object that is marked as owned by a service instance, is unique to that
 instance. In the index case we do so by making sure the object's identity is in fact completely and uniquely derived from the


docs/resource_sets.rst
```diff
@@ -91,3 +93,168 @@ Limitations
...
+TODO: "On a high level" -> remove?
+TODO: should I simplify and just talk about resources rather than generic nodes? Less accurate but also less abstract
+
+On a high level, to allow safe use of partial compiles, it is imperative that each service's refinements (through
+implementations) form a tree that can only intersect between service instances on shared nodes. The whole subtree below such a
+shared node should be considered shared and any resources in it must not be part of a resource set. All shared resources should
+be consistent between any two service instances that might create the object (see `Constraints and rules`_). All other nodes
+should generally be considered owned by the service and all their resources be part of the service's resource set. For more
+details on what it means to own a a child node in the tree (e.g. a resource) and how to ensure two service instance's trees can
+not intersect on owned nodes, see the `Ownership`_ subsection.
```
wouterdb, line 122: Would it not be easier to explain from the notion of ownership?

```suggestion
On a high level, to allow safe use of partial compiles, each service must be the sole owner if its resources and shared resources must be identical across service instances. More precisely: each service's refinements (through
implementations) form a tree that can only intersect between service instances on shared nodes. The whole subtree below such a shared node should be considered shared and any resources in it must not be part of a resource set. All shared resources should be consistent between any two service instances that might create the object (see `Constraints and rules`_). All other nodes should generally be considered owned by the service and all their resources be part of the service's resource set. For more details on what it means to own a a child node in the tree (e.g. a resource) and how to ensure two service instance's trees can not intersect on owned nodes, see the `Ownership`_ subsection.
```

docs/resource_sets.rst
```diff
@@ -91,3 +93,168 @@ Limitations
...
+considered to have the same identity by the model.
+
+Ownership
+*********
+A resource can safely be considered owned by a service instance if it could never be created by another service instance. There
+are two main mechanisms that can be used to provide this guarantee. One is the use of indexes on appropriate locations, the
+other is the use of some external distributor of unique values (e.g. a plugin to generate a UUID or to allocate values in an
+inventory).
+
+In either case, the goal is to make sure that any object that is marked as owned by a service instance, is unique to that
```
wouterdb, line 148: 1. The index here is in fact not required, the condition below it is sufficient:
`we do so by making sure the object's identity is in fact completely derived from the identity of the service instance` 
The index helps verifying this behavior when doing a full compile, but it not required.

2. completely derived is not sufficient, if we say `the name is the first character of the service id`, the name is completely derived from the service id, but not unique

3. the service id is not the only valid option: if the service instance has other unique (combinations of fields), that are provided as unique (see previous section), they can also serve to make unique id's 


(point 2 and 3 can be considered too precise to bother)

docs/resource_sets.rst
```diff
@@ -91,3 +93,168 @@ Limitations
...
+instance. In the index case we do so by making sure the object's identity is in fact completely derived from the identity of
+the service instance. In the case where unique values are externally produced/allocated, responsibility for uniqueness falls
+to the plugin that produces the values.
+
+Generally, for every index on a set of attributes of an owned resource, at least one of the fields must be either derived from
+the identity of the service instance, or allocated in a safe manner by a plugin as described above. The same goes for every
+pair of resource id and agent. If the former constraint is not met, a full compile might fail, while if the latter is not met,
```
wouterdb, line 155: simplified english is better 
```suggestion
pair of resource id and agent. If the first constraint is not met, a full compile might fail, while if the second is not met,
```

docs/resource_sets.rst
```diff
@@ -91,3 +93,168 @@ Limitations
...
+the export will be rejected because two services are trying to configure the same resources.
+
+For example, consider the example model from before. If two networks with two routers each would be created, they would result
```
wouterdb, line 158: This confuses me, because routers tend to connect networks, they can never be contained in them. (interfaces of routers can be in networks)

wouterdb (changes requested):

docs/resource_sets.rst
```diff
@@ -91,3 +93,168 @@ Limitations
...
+care to respect the `Constraints and rules`_.
+
+For simplicity, we will assume a model where each set of independent resources is ultimately defined by a single top-level
+entity, which we will refer to as the "service" or "service entity" (as in ``LSM``). While the same concepts apply to the
+use case where a set of independent resources is defined by more than one service entity, the guidelines currently do not
+cover that use case.
+
+TODO: check how trivial it would be to extend to the multi-service use case
+TODO: "On a high level" -> remove?
+TODO: should I simplify and just talk about resources rather than generic nodes? Less accurate but also less abstract
```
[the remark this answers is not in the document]
sanderr: Still, this is a pretty big thing we'd be glancing over. I think that if I was reading documentation like this I'd like it to be covered. But perhaps I'm not the average end-user in that regard.

docs/resource_sets.rst
```diff
@@ -91,3 +93,168 @@ Limitations
...
+
+On a high level, to allow safe use of partial compiles, it is imperative that each service's refinements (through
```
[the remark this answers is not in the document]
sanderr: Like the ones below, just move it up? Or would you like something more generic, not tied to the example?

docs/resource_sets.rst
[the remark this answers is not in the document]
wouterdb: In any case, we should structure it to make the main point understood, then refine to the details

docs/resource_sets.rst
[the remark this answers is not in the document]
wouterdb: Moving it up might be sufficient, examples are good.

docs/resource_sets.rst
sanderr: 2. I did take that into account but I see how my wording is quite ambiguous in that respect. I'll pay special attention to this.
3. I don't think I ever mention "service id" directly (I've tried to explicitly avoid that). I mostly speak generically of "identity", which corresponds to what you're saying. I'll see if I can better define what I mean by that somewhere.

docs/resource_sets.rst
sanderr: I'll modify the example to use `Host` rather than `Router`.

docs/resource_sets.rst
```diff
@@ -61,7 +61,7 @@ as the index example shows. This can result in undefinable behavior.
 Constraints and rules
 ************************
 
-Some restrictions and guidelines were implemented to guarantee that partial compilations function properly:
```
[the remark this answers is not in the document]
sanderr: On the other hand, it might be easier to manage while the model is still small, then with each small change make sure to meet these constraints?

docs/resource_sets.rst
```diff
@@ -91,3 +93,291 @@ Limitations
...
+        fontcolor="grey";
+        label = "Shared and consistent among all service instances";
+        labelloc = "bottom";
+    }
+
+    # force rendering on multiple ranks
+    {"Resource(id=0)" "Resource(id=1)" "Resource(id=2)" "Resource(id=3)"} -> "SharedResource(id=0, value=0)" [style="invis"]
+
+
+.. digraph:: resource_sets_generic_bad
```
wouterdb, line 166: I would add the explanation of the bad one after the good one, this has a more natural flow.

wouterdb: It may also help to split both bad things into their own example

docs/resource_sets.rst
```diff
@@ -91,3 +93,291 @@ Limitations
...
+a resource set. All shared resources should be consistent between any two service instances that might create the object (see
+`Constraints and rules`_). All other nodes should generally be considered owned by the service and all their resources be part
+of the service's resource set. For more details on what it means to own a resource (or any child node in the tree) and how
+to ensure two service instance's trees can not intersect on owned nodes, see the `Ownership`_ subsection.
+
+
+Service instance uniqueness
+***************************
+With full compiles, indexes serve as the identity of a service instance in the model. The compiler then validates that no conflicting
+service instances can exist. With partial compiles this validation is lost because only one service instance will be present
```
wouterdb, line 213:
```suggestion
service instances exist. With partial compiles this validation is lost because only one service instance will be present
```

wouterdb (approved):

docs/resource_sets.rst
sanderr: Done

docs/resource_sets.rst
sanderr: Done

docs/resource_sets.rst
sanderr: 2. I tried to improve this a bit.
3. Added a definition of what the document means when it says "identity".

sanderr:
@wouterdb I finalized the document, review is for real this time. Changes since your last review:
- added warning to top of the document
- moved below model developer section
- added some links within the document
- improved on uniqueness and ownership section
- added testing section
- processed review comments
- updated PR description

wouterdb (approved):

Hugo-Inmanta (changes requested):

docs/model_developers/resource_sets.rst
```diff
@@ -0,0 +1,478 @@
...
+            network.hosts += Host(id=host["id"])
+        end
+    end
+
+
+.. code-block:: yaml
+
+    networks:
+        - id: 0
+          routers:
```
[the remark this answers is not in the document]
sanderr: Oh thanks, they used to be called routers...

docs/model_developers/resource_sets.rst
```diff
@@ -0,0 +1,478 @@
...
+a resource set. All shared resources should be consistent between any two service instances that might create the object (see
+`Constraints and rules`_). All other nodes should generally be considered owned by the service and all their resources be part
+of the service's resource set. For more details on what it means to own a resource (or any child node in the tree) and how
+to ensure two service instance's trees can not intersect on owned nodes, see the `Ownership`_ subsection.
+
+
+Service instance uniqueness
+---------------------------
+With full compiles, indexes serve as the identity of a service instance in the model. The compiler then validates that no conflicting
+service instances exist. With partial compiles this validation is lost because only one service instance will be present
```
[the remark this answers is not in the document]
sanderr: It is possible but the document introduction states that it will not be covered here. Also lsm does not at the moment support it.

docs/model_developers/resource_sets.rst
```diff
@@ -0,0 +1,478 @@
...
+            - id: 4
+        - id: 0
+          routers:
+            - id: 0
+            - id: 1
+
+
+The ``read_from_yaml()`` plugin would have to verify that no two networks with the same id are defined. After this validation,
+if doing a partial, it may return a list with only the relevant network in it. For the yaml given above validation would fail
+because two routers with the same id are defined.
```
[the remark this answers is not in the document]
sanderr: Double typo even: there are no routers, only hosts and networks, and you're right that it's about networks here.

docs/model_developers/resource_sets.rst
```diff
@@ -0,0 +1,478 @@
...
+For a guide on how to design a model in order to take these into account, see `Modeling guidelines`_.
+
+
+Exporting a partial model to the server
+---------------------------------------
+
+Two arguments can be passed to the ``inmanta export`` command in order to export a partial model to the server:
+
+- ``--partial`` To specify that the model being compiled only contains the resources that need to be updated in relation to the previous version of the model.
+- ``--delete-resource-set <resource-set-name>`` This option, which may be used more than once, instructs the model to remove the resource set with the specified name. Only in conjunction with the preceding choice may this option be utilized. Note that utilizing a ``std::ResourceSet`` that includes no resources allows resource sets to be implicitly deleted during a partial compilation.
```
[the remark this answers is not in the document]
sanderr: Good remark, I'll make sure to fix it here or create a ticket.

docs/model_developers/resource_sets.rst
```diff
@@ -0,0 +1,478 @@
...
+            - id: 4
+        - id: 0
+          hosts:
+            - id: 0
+            - id: 1
+
+
+The ``read_from_yaml()`` plugin would have to verify that no two networks with the same id are defined. After this validation,
+if doing a partial, it may return a list with only the relevant network in it. For the yaml given above validation would fail
+because two routers with the same id are defined.
```
sanderr, line 312:
```suggestion
because two networks with the same id are defined.
```

docs/model_developers/resource_sets.rst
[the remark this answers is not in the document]
sanderr: Turns out this is already specified in the "Resource sets" subsection. Do you think that suffices or do you have a suggestion to improve it?

Hugo-Inmanta (approved):

PR #4829, opened 2022-09-09
Issue/4803 compilerservice notifications

### LINKED ISSUE #4803: compilerservice: configure notification behavior for compile request
- add an option to configure if notification is required
- add message option to use when notification is used
- add sane defaults for both to ensure backwards compatibility

FloLey:
# Description

- add an option to configure if notification is required
- add message option to use when notification is used
- add sane defaults for both to ensure backwards compatibility

closes #4803

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


tests/db/test_v202208180_to_v202209090.py
```diff
@@ -0,0 +1,51 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    Contact: code@inmanta.com
+"""
+import os
+from typing import Awaitable, Callable, List
```
sanderr, line 19: Small reminder: we recently decided we would do `from collections import abc` and use these types in a qualified manner, i.e. `abc.Awaitable`.

changelogs/unreleased/4803-compiler-service-notification.yml
```diff
@@ -0,0 +1,7 @@
+description: configure notification behavior for compile request
```
sanderr, line 1: When we use this as a message in the actual changelog we should be a bit more verbose.
```suggestion
description: Added options to compiler service to configure notification behavior
```

src/inmanta/data/__init__.py
```diff
@@ -3482,6 +3482,8 @@ class Compile(BaseDocument):
         to this one that actually got compiled.
     :param partial: True if the compile only contains the entities/resources for the resource sets that should be updated
     :param removed_resource_sets: indicates the resource sets that should be removed from the model
+    :param notify_failed_compile: if true use the notification service to notify that a compile has failed
```
sanderr, line 3485:
```suggestion
    :param notify_failed_compile: if true use the notification service to notify that a compile has failed.
        By default, notifications are enabled only for exporting compiles.
```

src/inmanta/db/versions/v202209090.py
```diff
@@ -0,0 +1,30 @@
...
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    schema = """
+
+    ALTER TABLE public.compile
+        ADD COLUMN notify_failed_compile boolean DEFAULT FALSE,
```
sanderr, line 26:
```suggestion
        ADD COLUMN notify_failed_compile boolean DEFAULT NULL,
```
See my comment about making it optional above.

src/inmanta/db/versions/v202209090.py
```diff
@@ -0,0 +1,30 @@
...
+        ADD COLUMN failed_compile_message varchar DEFAULT '';
```
sanderr, line 27:
```suggestion
        ADD COLUMN failed_compile_message varchar NOT NULL DEFAULT '';
```
Not completely sure about the syntax or the need but I think it is allowed to be `null` otherwise.

src/inmanta/server/services/compilerservice.py
```diff
@@ -74,6 +74,18 @@ async def compile_done(self, compile: data.Compile) -> None:
...
 
+    @abc.abstractmethod
+    async def compile_failed(self, compile: data.Compile) -> None:
+        """Receive notification of the compilerService
+
+        1- Notifications are delivered at least once (retry until all listeners have returned or raise an exception other
+        than CancelledError)
+        2- Notification are delivered out-of-band (i.e. the next compile can already start, multiple notifications can be in
+        flight at any given time, out-of-order delivery is possible but highly unlikely)
+        3- Notification are cancelled upon shutdown
```
sanderr, line 85: Maybe move this to the class docstring to avoid duplication?

tests/db/test_v202208180_to_v202209090.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps/v202208180.sql"))
+async def test_added_resource_set_column(
+    migrate_db_from: abc.Callable[[], abc.Awaitable[None]],
+    postgresql_client,
+    get_columns_in_db_table: abc.Callable[[str], abc.Awaitable[List[str]]],
+) -> None:
+    """
+    Test the database migration script that adds the `notify_failed_compile` and 'failed_compile_message' column to the
+    database.
+
```
sanderr, line 34:
```suggestion
```

sanderr (changes requested):
- I think the compiler- and notificationservice changes might need some rework
- Another thing to take into account is compile queue compacting

Let's have a call to discuss these.

src/inmanta/server/services/notificationservice.py
```diff
@@ -63,14 +63,22 @@ async def start(self) -> None:
     async def _cleanup(self) -> None:
         await data.Notification.clean_up_notifications()
 
-    async def compile_done(self, compile: data.Compile) -> None:
```
sanderr, line 66: This is still redundant, isn't it? It's part of the `compile` object.

src/inmanta/server/services/notificationservice.py
```diff
@@ -63,14 +63,22 @@ async def start(self) -> None:
...
+    async def compile_done(self, compile: data.Compile, notify_failed_compile: bool = False) -> None:
         if not compile.success:
             compile_report = await data.Compile.get_report(compile_id=compile.id)
             reports = compile_report["reports"] if compile_report else []
             failed_pull_stage = next(
                 (report for report in reports if report["name"] == "Pulling updates" and report["returncode"] != 0), None
             )
-            if failed_pull_stage:
+            if notify_failed_compile:
```
sanderr, line 73: If `compile.notify_failed_compile is False` just return without doing anything?

src/inmanta/server/services/notificationservice.py
```diff
@@ -63,14 +63,22 @@ async def start(self) -> None:
...
+                await self.notify(
+                    compile.environment,
+                    title="Compile request failed",
```
sanderr, line 76:
```suggestion
                    title="Compilation failed",
```
Stay consistent with the existing behavior where we can?

src/inmanta/server/services/notificationservice.py
```diff
@@ -63,14 +63,22 @@ async def start(self) -> None:
...
+                    message=compile.failed_compile_message,
```
sanderr, line 77: Add a default if this message is not set.

sanderr (changes requested):

src/inmanta/server/services/notificationservice.py
FloLey: hmm no this is wrong, i'll change it.

src/inmanta/server/services/compilerservice.py
```diff
@@ -746,6 +750,9 @@ async def _run(self, compile: data.Compile) -> None:
         compile_data: Optional[model.CompileData]
         success, compile_data = await runner.run(force_update=any(c.force_update for c in chain([compile], merge_candidates)))
 
+        if compile.notify_failed_compile and not success:
+            self.add_background_task(self._notify_listeners(compile, True))
+
```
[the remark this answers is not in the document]
FloLey: it does, I didn't think this through enough

src/inmanta/server/services/notificationservice.py
sanderr: I think you may have misunderstood me here. I think what was there before was perfectly fine, and perhaps even a bit more readable with the `elif` than with the `else: if`. What I meant to say is you don't handle the case where `notify_failed_compile is False` (rather than `None`).

sanderr (changes requested):
See one unresolved thread.

src/inmanta/server/services/notificationservice.py
FloLey: I'm not sure I understand, I thought we still wanted to notify for the `failed_pull_stage` and the `compile.do_export` even if `notify_failed_compile` is false. Is that not the case?

sanderr: We want that to be the default (`notify_failed_compile is None`), but if the user requests `notify_failed_compile=False` I think it would be counterintuitive to still notify for exporting compiles.

FloLey: Ok now I see :p Indeed I agree and this makes more sense. thanks

sanderr (approved):

src/inmanta/server/services/compilerservice.py
```diff
@@ -63,7 +63,7 @@
 
 class CompileStateListener(object):
     @abc.abstractmethod
-    async def compile_done(self, compile: data.Compile) -> None:
+    async def compile_done(self, compile: data.Compile, notify_failed_compile: bool = False) -> None:
```
arnaudsjs, line 66: I guess this needs to be removed, right? That information is part of the Compile object.

```suggestion
    async def compile_done(self, compile: data.Compile) -> None:
```

src/inmanta/server/services/notificationservice.py
```diff
@@ -70,7 +70,17 @@ async def compile_done(self, compile: data.Compile) -> None:
             failed_pull_stage = next(
                 (report for report in reports if report["name"] == "Pulling updates" and report["returncode"] != 0), None
             )
-            if failed_pull_stage:
+            if compile.notify_failed_compile == False:
```
arnaudsjs, line 73:
```suggestion
            if compile.notify_failed_compile is False:
```

tests/db/test_v202208180_to_v202209090.py
```diff
@@ -0,0 +1,52 @@
...
+    # Assert state after running the DB migration script
+    assert "notify_failed_compile" in (await get_columns_in_db_table("compile"))
+    assert "failed_compile_message" in (await get_columns_in_db_table("compile"))
+
+    compiles = await postgresql_client.fetch("SELECT * FROM public.compile;")
+    for c in compiles:
+        assert "notify_failed_compile" in c
+        assert c["notify_failed_compile"] is False
+        assert "failed_compile_message" in c
+        assert c["failed_compile_message"] == ""
```
arnaudsjs, line 52:
```suggestion
    compiles = await postgresql_client.fetch("SELECT * FROM public.compile;")
    assert compiles
    for c in compiles:
        assert "notify_failed_compile" in c
        assert c["notify_failed_compile"] is None
        assert "failed_compile_message" in c
        assert c["failed_compile_message"] == ""
```

tests/db/test_v202208180_to_v202209090.py
```diff
@@ -0,0 +1,52 @@
...
+    Contact: code@inmanta.com
+"""
+import os
+from collections import abc
+from typing import List
+
+import pytest
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps/v202208180.sql"))
```
arnaudsjs, line 25:
```suggestion
@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps", "v202208180.sql"))
```

tests/server/test_compilerservice.py
```diff
@@ -1153,6 +1153,53 @@ async def test_compileservice_api(client, environment):
     assert result.code == 200
 
 
+@pytest.mark.parametrize(
+    "message",
+    ["a custom message", None],
+)
+async def test_notification_failed_compile_with_message(
+    server, client, environment_factory: EnvironmentFactory, message
+) -> None:
```
arnaudsjs, line 1162:
```suggestion
async def test_notification_failed_compile_with_message(
    server, client, environment_factory: EnvironmentFactory, message: Optional[str]
) -> None:
```

src/inmanta/server/services/notificationservice.py
```diff
@@ -70,7 +70,17 @@ async def compile_done(self, compile: data.Compile) -> None:
...
+                return
+            elif compile.notify_failed_compile:
+                await self.notify(
+                    compile.environment,
+                    title="Compilation failed",
+                    message=compile.failed_compile_message if compile.failed_compile_message else "A compile has failed",
+                    severity=const.NotificationSeverity.error,
+                    uri=f"/api/v2/compilereport/{compile.id}",
```
arnaudsjs, line 81: The notifications mentioned below this one in the `elif` sequence have a more specific message. Putting this condition at the start might cause a specific notification message to be overriden by a generic one, which is a step backwards in my opinion. An attempt to improve it:

```
if compile.notify_failed_compile and compile.failed_compile_message is not None:
    # Use specific message provided in request
    await self.notify()
elif failed_pull_stage:
    await self.notify()
elif compile.do_export:
    await self.notify()
elif compile.notify_failed_compile:
    # Send notification with generic message as fallback
    await self.notify()
``` 
I am not entirely sure what is best here. It might even be better to remove the first if condition. That way we force a specific notification message over the one provided in the request. @sanderr Do you have an opinion on this?

arnaudsjs (changes requested):

src/inmanta/server/services/notificationservice.py
sanderr: I think you're suggestion is an improvement of behavior. The logic is slightly more complicated but I think that's an acceptable price to pay. About dropping the first condition, I had thought about this before but I concluded that it could be counterintuitive if the user requests a specific message and we override it in certain circumstances. I don't feel too strongly about this but I'd prefer to keep the first condition.

src/inmanta/data/__init__.py
```diff
@@ -3511,6 +3514,9 @@ class Compile(BaseDocument):
     partial: bool = False
     removed_resource_sets: list[str] = []
 
+    notify_failed_compile: Optional[bool] = None
+    failed_compile_message: str = ""
```
arnaudsjs, line 3518: Why not default to None here?

src/inmanta/data/model.py
```diff
@@ -151,6 +151,9 @@ class CompileRunBase(BaseModel):
     partial: bool = False
     removed_resource_sets: list[str] = []
 
+    notify_failed_compile: Optional[bool] = None
+    failed_compile_message: str = ""
```
arnaudsjs, line 155: Why not default to None?

src/inmanta/server/services/compilerservice.py
```diff
@@ -570,6 +570,8 @@ async def request_recompile(
         env_vars: Optional[Mapping[str, str]] = None,
         partial: bool = False,
         removed_resource_sets: Optional[List[str]] = None,
+        notify_failed_compile: Optional[bool] = None,
+        failed_compile_message: str = None,
```
arnaudsjs, line 574:
```suggestion
        failed_compile_message: Optional[str] = None,
```

src/inmanta/db/versions/v202209090.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    schema = """
+    ALTER TABLE public.compile
+        ADD COLUMN notify_failed_compile boolean DEFAULT NULL,
+        ADD COLUMN failed_compile_message varchar NOT NULL DEFAULT '';
```
arnaudsjs, line 26:
```suggestion
        ADD COLUMN failed_compile_message varchar DEFAULT NULL;
```

src/inmanta/server/services/compilerservice.py
arnaudsjs, line 574: We should also document these arguments here. It should definitely describe the behavior when notify_failed_compile is set to None as this is not very intuitive.

arnaudsjs (changes requested):

src/inmanta/data/model.py
FloLey: I should, I forgot to change it

tests/db/test_v202208180_to_v202209090.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+import os
+from collections import abc
+from typing import List
+
+import pytest
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps", "v202208180.sql"))
+async def test_added_resource_set_column(
```
arnaudsjs, line 26:
```suggestion
async def test_added_notification_related_columns_to_compile_table(
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #4841, opened 2022-09-15
Split query to improve performance

bartv:
# Description

The original query first calculates the deploy status for all resources in the database before it actually filters. It might be possible to do this correct in a single query but it would make the query even more complex than it already is.

This PR does the following:

- fetch the resource without requires status
- do a second query that calculates the resource status only for the requires

The correct solution would be to normalize the data so that there is a table that contains the requires. This could be used to improve this query but can also replace the current provides field.

This solution is not perfect but it does reduce the query time with more than 10x for a serious size.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/improve-query-performance.yml
+++ b/changelogs/unreleased/improve-query-performance.yml
@@ -1,0 +1,3 @@
+description: Improve query performance by splitting the query into two different queries.
+change-type: patch
+destination-branches: [master, iso5]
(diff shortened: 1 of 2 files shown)


wouterdb (approved):
needs a changelog file

sanderr (approved):

PR #4847, opened 2022-09-16
Improve logs query by using like

bartv:
# Description

Use like instead of position + fix the name of the lock

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -3992,6 +3992,10 @@ class ResourceAction(BaseDocument):
         resource_id: m.ResourceIdStr,
         offset: int,
     ) -> Tuple[str, List[object]]:
+        # The query uses a like query to match resource id with a resource_version_id. This means we need to escape the % and _
+        # characters in the query
+        resource_id = resource_id.replace("%", "#%").replace("_", "#_") + "%"
+
         query = f"""{select_clause}
                     FROM
                     (SELECT action_id, action, (unnested_message ->> 'timestamp')::timestamptz as timestamp,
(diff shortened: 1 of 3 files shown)


src/inmanta/data/__init__.py
```diff
@@ -3992,14 +3992,18 @@ def _get_resource_logs_base_query(
         resource_id: m.ResourceIdStr,
         offset: int,
     ) -> Tuple[str, List[object]]:
+        # The query uses a like query to match resource id with a resource_version_id. This means we need to escape the % and _
+        # characters in the query
+        resource_id = resource_id.replace("%", "#%").replace("_", "#_") + "%"
```
wouterdb, line 3997: We probably have to escape itself # as well

src/inmanta/data/__init__.py
```diff
@@ -4787,6 +4794,7 @@ async def get_released_resources(
             db_query = f"""SELECT * FROM ({db_query}) AS matching_records
                         ORDER BY matching_records.{order_by_column} {order}, matching_records.resource_version_id {order}"""
 
+        print(db_query, values)
```
wouterdb, line 4797: This has to go I guess

src/inmanta/data/__init__.py
wouterdb: Might be good to test with a resource with a sequence of escaped chars

wouterdb (changes requested):

src/inmanta/data/__init__.py
bartv: I added an escape for the escape. I did a test on a database and that was indeed required

sanderr (approved):

wouterdb (approved):

PR #4865, opened 2022-09-20
issue/4649-Improve-module-installation-logging-format

### LINKED ISSUE #4649: Improve module installation logging format
### Follow-up ticket to https://github.com/inmanta/inmanta-core/issues/4500
---

2022-08-05 15-02-02 project install

The logging of the modules' versions before and after install is not indented. It could cause the user to glance over this information.

Ideally, we would fix this generically in the logging formatter (in `inmanta.app` iirc).

If too much of a hassle / too difficult -> log on separate lines

FloLey:
# Description

The logging of the modules' versions before and after install is not indented. It could cause the user to glance over this information. This commit logs all the lines on separated lines with the right indentation 

closes #4649 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


FloLey:
I didn't write any test for this as I don't think this really deserves a test but if one if you is angry I'll add one :)

src/inmanta/module.py
```diff
@@ -501,30 +501,33 @@ def log_pre_install_information(self, module_name: str, module_spec: List[Inmant
...
-
+        set_pre_install: Set[Tuple[str, "Version"]] = set(previous_snapshot.items())
+        set_post_install: Set[Tuple[str, "Version"]] = set(version_snapshot.items())
+        updates_and_additions: Set[Tuple[str, "Version"]] = set_post_install - set_pre_install
+        kept: Set[Tuple[str, "Version"]] = set_pre_install.intersection(set_post_install)
+
+        if header:
+            LOGGER.debug(header)
+        for inmanta_module_name, package_version in sorted(kept):
+            LOGGER.debug(inmanta_module_name + ": " + str(package_version))
```
Hugo-Inmanta, line 521: I could be wrong, but I think it was suggested at some point to only log a "real" diff view (aka only things that got actually changed)
```suggestion

```

Hugo-Inmanta (changes requested):

src/inmanta/app.py
```diff
@@ -72,6 +72,34 @@
...
+                name=record.name,
+                level=record.levelno,
+                pathname=record.pathname,
+                lineno=record.lineno,
+                msg="",
+                args=(),
+                exc_info=None,
+            )
+        )
+        len_header = len(header) - 18
```
FloLey, line 91: I know this is not great but I didn't find a way to do it better.

FloLey:
@sanderr I changed it to make it generic but there  is still one part of the code that I don't like but that I can immediately make better. I could use regexes I guess but i'm not sure it is worth it.

Hugo-Inmanta (approved):

src/inmanta/app.py
```diff
@@ -72,6 +72,34 @@
...
+        # 18 is the total length of all the codes used to change the colors in the logs (2 colors "\x1b[3xm" and 2 reset
+        # "\x1b[0m" codes) https://tforgione.fr/posts/ansi-escape-codes/
+        return len_header
+
+    def format(self, record):
+        """Format a record with added indentation."""
+        indent = " " * self.get_header_length(record)
+        head, *trailing = super().format(record).splitlines(True)
```
sanderr, line 99: Traditionally called `tail`, FYI (at least I've often seen `head, *tail` or the inverse `*init, last`). Not a change request.

sanderr, line 99:
```suggestion
        head, *trailing = super().format(record).splitlines(keepends=True)
```
Prevents people having to go to the docs to see what this is about (like I just did).

src/inmanta/app.py
```diff
@@ -72,6 +72,34 @@
...
+        return head + "".join(indent + line for line in trailing)
```
sanderr, line 100: Looks good to me, no objections. If you ever need this again though you might find [`textwrap.indent`](https://docs.python.org/3/library/textwrap.html#textwrap.indent) interesting.

src/inmanta/module.py
```diff
@@ -509,9 +509,9 @@ def _log_version_snapshot(self, header: Optional[str], version_snapshot: Dict[st
     def _log_snapshot_difference(
         self, version_snapshot: Dict[str, "Version"], previous_snapshot: Dict[str, "Version"], header: Optional[str]
     ) -> None:
-        set_pre_install = set(previous_snapshot.items())
-        set_post_install = set(version_snapshot.items())
-        updates_and_additions = set_post_install - set_pre_install
+        set_pre_install: Set[Tuple[str, "Version"]] = set(previous_snapshot.items())
+        set_post_install: Set[Tuple[str, "Version"]] = set(version_snapshot.items())
+        updates_and_additions: Set[Tuple[str, "Version"]] = set_post_install - set_pre_install
```
sanderr, line 514:
```suggestion
        set_pre_install: Set[tuple[str, "Version"]] = set(previous_snapshot.items())
        set_post_install: Set[tuple[str, "Version"]] = set(version_snapshot.items())
        updates_and_additions: Set[tuple[str, "Version"]] = set_post_install - set_pre_install
```

sanderr (changes requested):
Looks good to me, only thing I'd suggest having a quick go at is the color characters.

src/inmanta/app.py
```diff
@@ -75,9 +77,32 @@
...
+        super().__init__(
+            fmt, datefmt, style, log_colors, reset, secondary_log_colors, validate, stream, no_color, force_color, defaults
+        )
+        self.timed = timed
+
     def get_header_length(self, record):
         """Get the header length of a given record."""
-        header = super().format(
+        log_format = "%(asctime)s " if self.timed else ""
+        log_format += "%(name)-25s%(levelname)-8s%(message)s"
```
sanderr, line 103: I don't really get this. This is not stable, is it? How about using `ColoredFormatter(log_colors=False, reset=False, force_color=False, ...)` where `...` represents this instance's attributes?

sanderr (changes requested):

src/inmanta/app.py
FloLey: I created a Formatter that matches the one currently used (https://github.com/inmanta/inmanta-core/blob/df161b561b1d880c65db6d8ff5c496242e25e27b/src/inmanta/app.py#L742)
but removing the color parts. I'll try your approach

sanderr: My main issue with it is that the class seems to be generic but then there's this hardcoded assumption in there. It would be great if we could get rid of that. Related to your link: I now just noticed that we only use this multiline logger when `timed` is true. Was this a conscious decision? If so, could you elaborate?

FloLey: we don't? we use the `MultiLineFormatter` when `_is_on_tty` is true. 
timed just prepends the time to the log_format

sanderr: Right, my bad. Same question though, why only on a tty? What's the alternative scenario? Is this the compile reports on the dashboard? I can seem why we might not want indentation there? Could you perhaps add a comment as to why we use a plain `Formatter`?

FloLey: I was already like that when it was a `colorlog.ColoredFormatter` So I think it is more about the color codes than the indentations. Maybe I could call the `MultiLineFormatter` in the else without colors but I don't know what this if/else had as purpose neither

sanderr: @arnaudsjs do you know what [this](https://github.com/inmanta/inmanta-core/blob/df161b561b1d880c65db6d8ff5c496242e25e27b/src/inmanta/app.py#L752) is about?

FloLey: I've added the `MultiLineFormatter` to the else clause but without colors for now

sanderr: Great, Arnaud and I concluded that we'd like to go with that for now. I would suggest not merging it until after we have branched off for the releases though, so we have some time to let this mature.

FloLey: I'll wait with the review request until then :)

src/inmanta/app.py
```diff
@@ -90,18 +90,15 @@ def __init__(
...
         )
-        self.timed = timed
+        self.fmt = fmt
 
     def get_header_length(self, record):
         """Get the header length of a given record."""
-        log_format = "%(asctime)s " if self.timed else ""
-        log_format += "%(name)-25s%(levelname)-8s%(message)s"
-        formatter = logging.Formatter(fmt=log_format)
+        formatter = colorlog.ColoredFormatter(fmt=self.fmt, reset=False, force_color=False, no_color=True)
```
sanderr, line 101: I think we should additionally pass all other parameters as is.

sanderr (changes requested):

src/inmanta/app.py
FloLey: indeed good catch :)

FloLey: No if we used the color codes here the length is not right:
"\x1B[30mHello" would return a length of more then 5 but the displayed string has only length 5 and we only care about that part as the color codes are not displayed.

sanderr: I don't mean these parameters. For these I agree we should hard set them, that's the whole purpose here (perhaps add a comment about this?). But for all other parameters that we don't care about, shouldn't we make sure to set them to be the same?

FloLey: Ho I see! :) I'll add them, it can do no harm :)

PR #4873, opened 2022-09-22
Set dependency between docker containers

arnaudsjs:
# Description

Set the startup/shutdown order between the Inmanta server and the database in the docker-compose file.

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


docs/install/install-server-with-docker/install-server.inc
```diff
@@ -38,6 +38,8 @@ Here is a minimalist docker-compose file content that can be used to deploy the
             networks:
                 inm_net:
                     ipv4_address: 172.30.0.3
+            depends_on:
+                - "postgres"
```
edvgui, line 42: I didn't knew that worked for the shutdown, very nice!

edvgui (approved):

PR #4879, opened 2022-09-22
Skip migration tests

FloLey:
# Description

Add a marker for migration test to skip migration test older than 30 days when in fast mode

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


pyproject.toml
```diff
@@ -32,5 +32,6 @@ markers = [
     "slowtest",
     "link_check",
     "parametrize_any: only execute one of the parameterized cases when in fast mode (see documentation in conftest.py)",
-    "db_restore_dump(dump): mark the db dump to restore. To be used in conjunction with the `migrate_db_from` fixture."
+    "db_restore_dump(dump): mark the db dump to restore. To be used in conjunction with the `migrate_db_from` fixture.",
+    "migration_test: mark a test as a migration test to skip it in fast runs once it is older than 30 days"
```
arnaudsjs, line 36: Let's make it very clear that we are talking about a *database* migration script. It's likely that we will have other types of migrations in the future.

```suggestion
    "db_migration_test: mark a test as a migration test to skip it in fast runs once it is older than 30 days"
```

changelogs/unreleased/add-test-marker-migration.yml
```diff
@@ -0,0 +1,3 @@
+change-type: patch
+description: Add a marker for migration test to skip migration test older than 30 days when in fast mode
```
arnaudsjs, line 2:
```suggestion
description: Add a marker to skip database migration tests older than 30 days when in fast mode
```

tests/conftest.py
```diff
@@ -28,7 +28,12 @@
 We analyze performance and place the @slowtest in the best places.
 It is often harder to correctly judge what is slow up front, so we do it in bulk when we have all the (historical) data.
 This also allows test to run a few hundred times before being marked as slow.
-"""
+
+The @migration_test annotation is only added on db migration tests. Those Tests should be placed in files following the right
```
arnaudsjs, line 32:
```suggestion
The @migration_test annotation is only added on db migration tests. Those tests should be placed in files following the right
```

tests/conftest.py
```diff
@@ -196,6 +204,16 @@ def pytest_runtest_setup(item: "pytest.Item"):
         return
     if any(True for mark in item.iter_markers(name="slowtest")):
         pytest.skip("Skipping slow tests")
+    if any(True for mark in item.iter_markers(name="migration_test")):
+        try:
+            file_name: str = item.location[0]
+            date: str = re.search("_v(.*)_to_v(.*)[0-9].py", file_name).group(2)
+            test_creation_date: datetime.datetime = datetime.datetime(int(date[0:4]), int(date[4:6]), int(date[6:8]))
```
arnaudsjs, line 211: This is slightly more efficient, because the regex doesn't have to search where in the string the pattern starts. Next to that this provides guarantees that the cast to int() will succeed.
```suggestion
            date: str = re.fullmatch("test_v[0-9]{9}_to_v([0-9]{9}).py", file_name).group(1)
            test_creation_date: datetime.datetime = datetime.datetime(int(date[0:4]), int(date[4:6]), int(date[6:8]))
```

tests/conftest.py
```diff
@@ -196,6 +204,16 @@ def pytest_runtest_setup(item: "pytest.Item"):
...
+            elapsed_days: int = (datetime.datetime.today() - test_creation_date).days
+            if elapsed_days > 30:
+                pytest.skip("Skipping old migration test")
+        except ValueError:
+            pytest.fail("The name of the test file might be incorrect: Should be test_v<old_version>_to_v<new_version>.py")
```
arnaudsjs, line 216: Why don't we check if `re.fullmatch()` returns None and based on that raise the exception?

arnaudsjs (changes requested):

tests/conftest.py
FloLey: that would indeed be better :) I'll change it!

Hugo-Inmanta (approved):

tests/conftest.py
FloLey: I needed to change this a bit but the idea was kept :)

tests/conftest.py
```diff
@@ -196,6 +205,20 @@ def pytest_runtest_setup(item: "pytest.Item"):
         return
     if any(True for mark in item.iter_markers(name="slowtest")):
         pytest.skip("Skipping slow tests")
+    if any(True for mark in item.iter_markers(name="db_migration_test")):
```
arnaudsjs, line 208: I just realize that this approach is rather verbose. You always need to add the `db_migration_test` annotation to any migration test we will add in the future. It's very likely that we will forget this at some point. What do you think about the following approach:

* Remove the `db_migration_test` annotation and distinguish the migration tests for the old and the new versioning schema based on the filename. When the filename of the test case matches the versioning schema of the regex, skip it when the timestamp is older than 30 days.
* Enforce the constraint that any database migration test need either the `slowtest` annotation or it's filename needs to match the new versioning schema. That way we protect ourselves against typos in the name of the file.

Sorry, I noticed this problem only now.

@sanderr Do you have an opinion on this?

arnaudsjs (changes requested):

tests/conftest.py
```diff
@@ -196,6 +205,20 @@ def pytest_runtest_setup(item: "pytest.Item"):
...
+        file_name: str = item.location[0]
+        match: str = re.fullmatch("tests/db/test_v[0-9]{9}_to_v([0-9]{8})[0-9].py", file_name)
+        print(match)
```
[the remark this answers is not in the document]
FloLey: indeed :p

tests/conftest.py
sanderr: I agree with your suggestion.

FloLey: Do I understand correctly that you would add a mechanism before the run of each test that would look a the path of the file to determine if it is a migration test or not, then use the regex to see what do to/which annotation is needed?

arnaudsjs: Yes

FloLey: This shouldn't be to hard to change :) And I agree it wouuld make life easier. I'll change it

tests/db/test_db_schema.py
```diff
@@ -36,6 +36,7 @@
 from utils import log_contains
 
 
+@pytest.mark.slowtest
```
arnaudsjs, line 39: I didn't realize that the db directory contained other test files (like `tests/db/test_db_schema.py` and `tests/db/test_restore_tool.py`) next to the migration test files. These former aren't slow so they shouldn't get the slowtest annotation. Maybe we should move all database migration related test files to a separate subdirectory to separate them from the other database related test files.

arnaudsjs (changes requested):

tests/db/test_db_schema.py
FloLey: I'll do that :)

arnaudsjs (changes requested):
Let's also move the `dump_tool.py` and the `dumps` directory into the `db/migration_tests` directory. Those very strongly belong together. That change shouldn't require a change to the `pytest_runtest_setup` method because it only processes test cases defined in a file that starts with `test_`.

arnaudsjs (approved):

PR #4882, opened 2022-09-23
WIP: Cache the environment to reduce queries

bartv:
# Description

This PR tries to cache environments because on every API call we actually query the database using a new connection and a new transaction.

There are still quite some test cases failing and I have the feeling this is getting out of control. A cleaner solution, but a lot more work, is to create an environment object that lazily loads all data.

In most calls we do two things:
- Verify that the environment exists
- Use the `.id` attribute to use in other methods call or queries.

Only caching the existence of the environment might be a an easier solution. I first want to get feedback on this before continuing. 

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -2439,7 +2439,7 @@ class Environment(BaseDocument):
 
     @classmethod
     def flush_cache(cls, env_id: Optional[uuid.UUID] = None) -> None:
-        if env_id:
+        if env_id is not None:
             if env_id in cls.__instance_cache:
                 del cls.__instance_cache[env_id]
         else:
@@ -2547,8 +2547,8 @@ class Environment(BaseDocument):
             await self.delete()
 
     async def delete(self, connection: Optional[asyncpg.connection.Connection] = None) -> None:
-        if self.id in self.__instance_cache:
-            del self.__instance_cache[self.id]
+        if self.id in Environment.__instance_cache:
+            del Environment.__instance_cache[self.id]
 
         await super().delete(connection)
 
@@ -2566,6 +2566,7 @@ RETURNING last_version;
             self.id,
             connection=connection,
         )
+        Environment.flush_cache()
         version = cast(int, record[0])
         self.last_version = version
         return version
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -2045,7 +2045,7 @@ class Project(BaseDocument):
 
     async def delete(self, connection: Optional[asyncpg.connection.Connection] = None) -> None:
         # flush the environment cache
-        Environment.flush_cache(self.id)
+        Environment.flush_cache()
         await super().delete(connection)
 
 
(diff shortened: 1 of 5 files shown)


src/inmanta/data/__init__.py
```diff
@@ -2043,6 +2043,11 @@ class Project(BaseDocument):
     def to_dto(self) -> m.Project:
         return m.Project(id=self.id, name=self.name, environments=[])
 
+    async def delete(self, connection: Optional[asyncpg.connection.Connection] = None) -> None:
+        # flush the environment cache
+        Environment.flush_cache()
```
sanderr, line 2048: Should we put the cache behind a lock? As in, prevent concurrent read-write, and hold the lock until `delete` is done?

src/inmanta/data/__init__.py
```diff
@@ -2630,21 +2656,33 @@ async def get_list_without_details(
...
         connection: Optional[asyncpg.connection.Connection] = None,
         details: bool = True,
-    ) -> Optional[TBaseDocument]:
+    ) -> Optional["Environment"]:
         """
         Get a specific environment based on its ID
 
         :return: An instance of this class with its fields filled from the database.
         """
+        if doc_id in cls.__instance_cache and details in cls.__instance_cache[doc_id]:
```
sanderr, line 2669:
```suggestion
        if connection is None and doc_id in cls.__instance_cache and details in cls.__instance_cache[doc_id]:
```
I'm not sure about this but I wonder if we could get strange inconsistency errors if we allow the cache to live accross transactions.

src/inmanta/data/__init__.py
```diff
@@ -2630,21 +2656,33 @@ async def get_list_without_details(
...
+            return cls.__instance_cache[doc_id][details]
+
         result = await cls.get_list(id=doc_id, connection=connection, details=details)
         if len(result) > 0:
+            if doc_id not in cls.__instance_cache:
```
sanderr, line 2674:
```suggestion
            if connection is None and doc_id not in cls.__instance_cache:
```
Same here

src/inmanta/data/__init__.py
```diff
@@ -2233,6 +2238,8 @@ class Environment(BaseDocument):
     :param icon: An icon for the environment
     """
 
+    __instance_cache: Dict[uuid.UUID, Dict[bool, "Environment"]] = {}
```
sanderr, line 2241: Add a comment on why we cache this? I think it's not trivial: one might think we want to avoid db calls but that part should actually be pretty cheap. The part we want to avoid it the acquisition of a new connection, right?

sanderr (commented):
This seems like a sensible approach to me. I of course have no context on what is currently going wrong, but from where I'm standing it looks as if this should be feasible.

About your suggestion to only cache the existence, I don't really see how that would work.

arnaudsjs (approved):
This approach is fine by me. Do I understand correctly that the main incentive for the alternative approach, where we only cache existence, is that we have to adjust less tests to make the test suite succeed?

PR #4889, opened 2022-09-23
deadlock fix

sanderr:
# Description

Fixes deadlock introduced by loosening explicit lock in #4871 without losing the huge performance gain that came with that PR.

Still in draft because there might be one more deadlock (a lot rarer). I would propose to merge the full PR before the release, except for `tests/test_postgres_concurrency.py`, which I'd like to do in a separate PR. If anyone thinks we should split it up further to have better confidence in the release, leave a comment about that and we'll discuss it.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


src/inmanta/server/services/orchestrationservice.py
```diff
@@ -760,7 +760,10 @@ async def put_version(
         async with data.Resource.get_connection() as con:
             async with con.transaction():
                 # Acquire a lock that conflicts with the lock acquired by put_partial.
-                await data.Resource.lock_table(data.TableLockMode.ROW_EXCLUSIVE, connection=con)
+                await con.execute(
```
sanderr, line 763: Take advisory lock instead of table lock:
- no deadlocks possible due to table locking order
- don't lock full table, only for this environment (sort of, the full uuid is too long to use as key, so we only use part of the timestamp).

src/inmanta/data/__init__.py
```diff
@@ -5578,29 +5579,25 @@ async def delete_cascade(self) -> None:
                 # Delete all code associated with this version
                 await Code.delete_all(connection=con, environment=self.environment, version=self.version)
 
-                # Acquire explicit lock to avoid deadlock. See ConfigurationModel docstring
-                await ResourceAction.lock_table(TableLockMode.ROW_EXCLUSIVE, connection=con)
-                await Resource.delete_all(connection=con, environment=self.environment, model=self.version)
+                # Delete ConfigurationModel and cascade delete on connected tables
+                await self.delete(connection=con)
```
sanderr, line 5583: This is the main fix, the rest are some additional easy wins.

tests/test_data_concurrency.py
```diff
@@ -0,0 +1,25 @@
...
+    Contact: code@inmanta.com
+"""
+
+"""
+This module contains tests related to database concurrency issues. Whenever we fix a concurrency issue, be it performance
+or deadlock related, a test should be added to ensure this occurence can not accidentally be introduced again.
+"""
+
+
+# TODO: add tests for known deadlocks
```
sanderr, line 25: I'll probably move this to a separate PR so this one can be merged ASAP.

src/inmanta/data/__init__.py
arnaudsjs: Regarding 2: By splitting this into two different transactions we introduce a database leak, right? When the second query would be executed for some reason, the Parameter records will remain in the db until the end of times.

src/inmanta/data/__init__.py
```diff
@@ -112,8 +112,10 @@ class TableLockMode(enum.Enum):
     Table level locks as defined in the PostgreSQL docs:
     https://www.postgresql.org/docs/13/explicit-locking.html#LOCKING-TABLES. When acquiring a lock, make sure to use the same
     locking order accross transactions to prevent deadlocks and to otherwise respect the consistency docs:
-    https://www.postgresql.org/docs/13/applevel-consistency.html#NON-SERIALIZABLE-CONSISTENCY. See relevant data classes'
-    docstrings for appropriate lock orderings.
+    https://www.postgresql.org/docs/13/applevel-consistency.html#NON-SERIALIZABLE-CONSISTENCY.
+
+    In general, locks should be acquired consistently with delete cascade lock order, which is top down. See relevant data
```
arnaudsjs, line 117: Wouldn't it be better to have one central place where we define the total locking order. Now this information is spread across many docstring, which makes it difficult to determine the total locking order and opens the door to inconsistencies in the documentation.

arnaudsjs (changes requested):

wouterdb (approved):

wouterdb (approved):

src/inmanta/data/__init__.py
sanderr: Done

arnaudsjs (approved):

PR #4909, opened 2022-09-28
Fix path db dump tool

FloLey:
# Description

PR https://github.com/inmanta/inmanta-core/pull/4879 moved some files around put paths weren't adapted as they should have.
This commit will fix this.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests_common/copy_files_from_core.py
+++ b/tests_common/copy_files_from_core.py
@@ -53,7 +53,7 @@ simple_project_dir = os.path.join(data_dir, "simple_project")
 db_common_py = os.path.join(db_dir, "common.py")
 dest_db_common_py = os.path.join(dest_db_dir, "common.py")
 dump_tool_py = os.path.join(db_dir, "dump_tool.py")
-dest_dump_tool_py = os.path.join(dest_db_dir, "dump_tool.py")
+dest_dump_tool_py = os.path.join(dest_db_dir, "migration_tests", "dump_tool.py")
 dest_simple_project_dir = os.path.join(dest_db_dir, "simple_project")
 
 
--- a/tests_common/copy_files_from_core.py
+++ b/tests_common/copy_files_from_core.py
@@ -52,7 +52,7 @@ enduser_certs_dir = os.path.join(data_dir, "ca", "enduser-certs")
 simple_project_dir = os.path.join(data_dir, "simple_project")
 db_common_py = os.path.join(db_dir, "common.py")
 dest_db_common_py = os.path.join(dest_db_dir, "common.py")
-dump_tool_py = os.path.join(db_dir, "dump_tool.py")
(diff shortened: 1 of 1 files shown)


FloLey:
I think this is the issue but I'm not sure if I should change `dest_dump_tool_py`? I guess dest has the same structure?
@arnaudsjs could you also remind me / tell me what this is used for?

tests_common/copy_files_from_core.py
```diff
@@ -52,8 +52,8 @@ def cleanup_dir(directory: str):
 simple_project_dir = os.path.join(data_dir, "simple_project")
 db_common_py = os.path.join(db_dir, "common.py")
 dest_db_common_py = os.path.join(dest_db_dir, "common.py")
-dump_tool_py = os.path.join(db_dir, "dump_tool.py")
-dest_dump_tool_py = os.path.join(dest_db_dir, "dump_tool.py")
+dump_tool_py = os.path.join(db_dir, "migration_tests", "dump_tool.py")
+dest_dump_tool_py = os.path.join(dest_db_dir, "migration_tests", "dump_tool.py")
```
arnaudsjs, line 56: I don't see a reason why we want to package the `dump_tool.py` in `pytest-inmanta-extensions`. I think we can just remove it. @sanderr Do you see a reason why we need this?

arnaudsjs: I think the same is true for the `common.py` actually.

arnaudsjs: Ah no. This is not true for the `common.py`. It's used by inmanta-lsm. Maybe we just have to keep this in as it doesn't hurt to keep it in.

tests_common/copy_files_from_core.py
```diff
@@ -52,8 +52,8 @@ def cleanup_dir(directory: str):
 simple_project_dir = os.path.join(data_dir, "simple_project")
 db_common_py = os.path.join(db_dir, "common.py")
 dest_db_common_py = os.path.join(dest_db_dir, "common.py")
-dump_tool_py = os.path.join(db_dir, "dump_tool.py")
-dest_dump_tool_py = os.path.join(dest_db_dir, "dump_tool.py")
+dump_tool_py = os.path.join(db_dir, "migration_tests", "dump_tool.py")
```
arnaudsjs, line 55: Executing this command results in a failure, because the `migration_tests` directory doesn't exist.

arnaudsjs (changes requested):

tests_common/copy_files_from_core.py
sanderr: I'm not aware of it being used anywhere.

arnaudsjs: Ok. Let's kick it out then.

arnaudsjs (approved):

PR #4947, opened 2022-10-05
fixed colorlog compatibility

sanderr:
# Description

Fixed some minor issues with our usage of the `colorlog` library:
- bumped constraint from `~=6.0` to `~=6.4` because all versions in between are pre-releases we're not necessarily compatible with
- dropped fields we don't need that were only added in later versions of the library than the one in our constraint
- import underscore-prefixed type only when type checking and with a safe fallback
- added some typing

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -86,9 +86,7 @@ class MultiLineFormatter(colorlog.ColoredFormatter):
         reset: bool = True,
         no_color: bool = False,
     ):
-        super().__init__(
-            fmt, log_colors=log_colors, reset=reset, no_color=no_color
-        )
+        super().__init__(fmt, log_colors=log_colors, reset=reset, no_color=no_color)
         self.fmt = fmt
 
     def get_header_length(self, record: logging.LogRecord) -> int:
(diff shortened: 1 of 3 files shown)


FloLey (approved):

src/inmanta/app.py
```diff
@@ -71,6 +71,12 @@
 except ImportError:
     rpdb = None
 
+if typing.TYPE_CHECKING:
+    try:
+        from colorlog.formatter import _FormatStyle
+    except ImportError:
+        _FormatStyle = typing.Literal["%", "{", "$"]
+
```
arnaudsjs, line 79: I am missing a comment here that describes why we need this

arnaudsjs (changes requested):

src/inmanta/app.py
```diff
@@ -80,40 +80,23 @@ class MultiLineFormatter(colorlog.ColoredFormatter):
     def __init__(
         self,
         fmt: typing.Optional[str] = None,
-        datefmt: typing.Optional[str] = None,
-        style: _FormatStyle = "%",
+        *,
```
FloLey, line 83: the asterix means that the following argument should be set with the keyword right?

sanderr: Indeed. I added this to prevent incorrect usage based on the assumption that this class would have the same interface as its parent.

FloLey (approved):

PR #4956, opened 2022-10-06
Ensure cleanup of temporary directories

arnaudsjs:
# Description

* This PR ensures that temporary directories, created by the test suite, are always cleaned up. 
* This PR removes the `src/inmanta/doctools.py` file. The assumption is that this file is not required anymore.

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/inmanta/doctools.py
```diff
@@ -1,118 +0,0 @@
-"""
```
sanderr, line 1: Do we really want to delete this for iso4 as well?

sanderr (approved):

src/inmanta/doctools.py
arnaudsjs: Well. I don't see where it's used at the moment, so I don't see the problem with removing it. If we do use it, the CI pipeline will point that out soon.

Hugo-Inmanta (approved):

src/inmanta/doctools.py
sanderr: Fair enough

Issue #1950, opened 2020-03-21
nullable attribute deprecation warning raised even if null explicitly assigned


sanderr:
```
entity A:
    number? n
end

implement A using std::none

A(n = null)
```
raises a deprecation warning

sanderr:
I guess I was a bit too quick in creating this issue. The above model does not produce the warning. The below model does:
```
entity A:
    number? n = null
end

entity B extends A:
end

entity C extends A, B:
end

implement A using std::none
implement B using std::none
implement C using std::none

C(n = null)
```

A similar inheritance structure is used in `ubiquiti::Switch`

sanderr:
The model
```
entity A:
    number? n
end

entity B extends A:
    number? n = null
end

implement A using std::none
implement B using std::none

B()
``` 
also produces the error. So basically any model where a nullable attribute is referenced multiple times in the inheritance hierarchy.

PR #4959, opened 2022-10-07
issue/1888-remove-depracated-OptionVariable

### LINKED ISSUE #1888: remove support for leaving nullable attribute unassigned
Currently nullable attributes are represented by `OptionVariable`. They should instead be represented by a plain `ResultVariable` with a nullable type. See #1775

- [x] Fix std
- [ ] Fix vyos
- [x] Ask solutions team to look into vcenter and yang (tickets created by solutions team)
- [ ] Implement the change

this is only for iso6

This a grep report from the jenkins jobs:
[https://app.zenhub.com/files/67439560/5f494500-f23d-4767-93e8-f3f57a0107db/download](https://app.zenhub.com/files/67439560/5f494500-f23d-4767-93e8-f3f57a0107db/download)

FloLey:
# Description

Currently nullable attributes are represented by OptionVariable. They will now be represented by a plain ResultVariable with a nullable type.

closes #1888 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
+++ b/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
@@ -1,0 +1,7 @@
+---
+description: Remove support for leaving nullable attribute unassigned, an exception will now be raised.
+change-type: major
+issue-nr: 1888
+destination-branches: [master, iso5]
+sections:
+  deprecation-note: "{{description}}"
--- a/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
+++ b/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
@@ -2,6 +2,6 @@
 description: Remove support for leaving nullable attribute unassigned, an exception will now be raised. You should make sure optional variables are always assigned a value.
 change-type: major
 issue-nr: 1888
-destination-branches: [master, iso5]
+destination-branches: [master]
 sections:
   deprecation-note: "{{description}}"
--- a/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
+++ b/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
@@ -1,5 +1,5 @@
 ---
-description: Remove support for leaving nullable attribute unassigned, an exception will now be raised.
+description: Remove support for leaving nullable attribute unassigned, an exception will now be raised. You should now make sure optional variables are always assigned a value.
 change-type: major
 issue-nr: 1888
 destination-branches: [master, iso5]
--- a/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
+++ b/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
@@ -1,5 +1,5 @@
 ---
-description: Remove support for leaving nullable attribute unassigned, an exception will now be raised. You should now make sure optional variables are always assigned a value.
+description: Remove support for leaving nullable attribute unassigned, an exception will now be raised. You should make sure optional variables are always assigned a value.
 change-type: major
 issue-nr: 1888
 destination-branches: [master, iso5]
--- a/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
+++ b/changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
@@ -1,5 +1,5 @@
 ---
-description: Remove support for leaving nullable attribute unassigned, an exception will now be raised. You should make sure optional variables are always assigned a value.
+description: Remove support for leaving nullable attribute unassigned, an exception will now be raised. You should make sure optional variables are always assigned a value. This changes the behaviour of 'is defined' in Jinja templates. You should now use 'is not none' instead
 change-type: major
 issue-nr: 1888
 destination-branches: [master]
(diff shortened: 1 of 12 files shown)


tests/compiler/test_null.py
```diff
@@ -41,39 +41,6 @@ def test_null(snippetcompiler):
...
-            end
-            implement A using std::none
-            a = A()
-            b = a.a
-        """
-    )
-    with pytest.raises(OptionalValueException):
-        (_, scopes) = compiler.do_compile()
-
-
```
FloLey, line 76: I think those 2 tests can just be removed?

tests/compiler/test_exception.py
```diff
@@ -86,30 +86,6 @@ def test_direct_execute_error(snippetcompiler):
...
-Test()
-        """,
-        "Could not set attribute `m` on instance `__config__::Test (instantiated at {dir}/main.cf:13)` (reported in self.m ="
-        " self.n ({dir}/main.cf:8))"
-        "\ncaused by:"
-        "\n  Optional variable accessed that has no value (attribute `n` of `__config__::Test (instantiated at"
-        " {dir}/main.cf:13)`) (reported in self.m = self.n ({dir}/main.cf:8))",
-    )
-
-
```
FloLey, line 112: I think this test can just be removed?

FloLey:
I still need to fix this one before I merge:
https://github.com/inmanta/vyos/pull/263

sanderr (commented):
The change looks right to me but depends heavily on what I still need to investigate. Haven't looked at the tests yet.

Hugo-Inmanta (approved):

changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
```diff
@@ -0,0 +1,7 @@
+---
+description: Remove support for leaving nullable attribute unassigned, an exception will now be raised. You should make sure optional variables are always assigned a value.
+change-type: major
+issue-nr: 1888
+destination-branches: [master, iso5]
```
sanderr, line 5:
```suggestion
destination-branches: [master]
```
The merge tool wouldn't even allow this: major changes can only go into master

src/inmanta/ast/attribute.py
```diff
@@ -117,17 +108,7 @@ def validate(self, value: object) -> None:
...
-        if self.is_optional():
-            # be a 0-1 relation
-            self.end = None
-            self.low = 0
-            self.high = 1
-            out = DeprecatedOptionVariable(self, instance, queue)
-        else:
-            out = ResultVariable()
-
+        out: ResultVariable["Instance"] = ResultVariable()
```
sanderr, line 111: I don't think this is typed correctly (as in, it has been typed incorrectly for the past four years). Could you fix it and run `mypy-diff`? I think `ResultVariable[object]` is probably the best we can do.

tests/compiler/test_null.py
sanderr: I would keep the first and update the check to assert that the value is `null` rather than expect the exception.

tests/compiler/test_warnings.py
```diff
@@ -151,7 +152,7 @@ def test_1918_shadow_warning_for_loop(snippetcompiler):
         assert str(w1.message) == message % (2, 4)
 
 
-def test_deprecation_warning_nullable(snippetcompiler):
+def test_deprecation_exception_nullable(snippetcompiler):
```
sanderr, line 155: The contents of this test look good to me but I would rename it (it doesn't have anything to do with deprecation anymore) and move it to `tests/compiler/test_null.py` (it doesn't have anything to do with warnings anymore).

tests/compiler/test_warnings.py
```diff
@@ -164,17 +165,18 @@ def test_deprecation_warning_nullable(snippetcompiler):
...
+        message: str = (
+            f"The object __config__::A (instantiated at {snippetcompiler.project_dir}/main.cf:8) is not "
+            f"complete: attribute n ({snippetcompiler.project_dir}/main.cf:3:13) is not set"
+        )
+        assert e.msg == message
 
 
 @pytest.mark.parametrize("assign", [True, False])
-def test_1950_deprecation_warning_nullable_diamond_inheritance(snippetcompiler, assign: bool):
+def test_1950_deprecation_exception_nullable_diamond_inheritance(snippetcompiler, assign: bool):
```
sanderr, line 179: Looking at #1950 and #1984 I think we can just drop this test.

sanderr (changes requested):
Also still waiting on the discussion on Slack but I think that will reach a conclusion soon.

tests/compiler/test_null.py
FloLey: you mean keep `test_null_unset`
and changing `string? a` to `string? a = null` and then verifying it is null? 
cause the test just above(`test_null`) test exactly that.

changelogs/unreleased/1888-remove-depracated-OptionVariable.yml
```diff
@@ -0,0 +1,7 @@
+---
+description: Remove support for leaving nullable attribute unassigned, an exception will now be raised. You should make sure optional variables are always assigned a value.
```
sanderr, line 2: We should probably also include something about the change of behavior of `OptionalValueException`, but that will have to wait until the discussion on Slack has reached a conclusion.

tests/compiler/test_null.py
sanderr: Perfect, in that case I withdraw my comment.

sanderr (approved):

PR #4963, opened 2022-10-07
Issue/4743 performance of put version

### LINKED ISSUE #4743: improve performance of partial resource application

wouterdb:
# Description

Removal of resource_version_id from the db

- read path: mostly contained to the data layer, `Resource` serializes the same as before, affects increment calculation
- write path: `put_version`  filters requires
- data layer itself: changes all over the place. 
-  database migration: rewrite requires and provides 

closes #4743 
# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/inmanta/data/__init__.py
```diff
@@ -4469,7 +4470,11 @@ class Resource(BaseDocument):
     # internal field to handle cross agent dependencies
     # if this resource is updated, it must notify all RV's in this list
     # the list contains full rv id's
-    provides: List[m.ResourceVersionIdStr] = []
+    provides: List[m.ResourceIdStr] = []
+
+    @property
+    def resource_version_id(self):
```
wouterdb, line 4476: hide the change to the programmer

src/inmanta/data/__init__.py
```diff
@@ -4481,23 +4486,23 @@ async def get_last_non_deploying_state_for_dependencies(
...
             FROM resource AS r1
             WHERE r1.environment=$1
                   AND r1.model=$2
                   AND (
                       SELECT (r2.attributes->'requires')::jsonb
                       FROM resource AS r2
-                      WHERE r2.environment=$1 AND r2.model=$2 AND r2.resource_version_id=$3
-                  ) ? r1.resource_version_id
+                      WHERE r2.environment=$1 AND r2.model=$2 AND r2.resource_id=$3
+                  ) ? r1.resource_id
```
sanderr, line 4497: For this to work, don't we need to rewrite existing rows to store `resource_id` (rather than `resource_version_id` in `attributes.requires`? And if we do, then the `cleaned_requires` collection in `server/services/orchestrationservice.py` might be redundant.

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -5342,19 +5372,36 @@ async def update_fields(self, connection: Optional[asyncpg.connection.Connection
...
+        dct["resource_version_id"] = rvid
         dct["id"] = dct["resource_version_id"]
+        if "requires" in dct["attributes"]:
+            dct["attributes"]["requires"] = [Id.set_version_in_id(id, version) for id in dct["attributes"]["requires"]]
+        dct["provides"] = [Id.set_version_in_id(id, version) for id in dct["provides"]]
+
         return dct
 
     def to_dto(self) -> m.Resource:
+        attributes = self.attributes
```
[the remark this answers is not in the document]
arnaudsjs: Oh. I didn't see the line `attributes = dict(attributes)`. This code is quite confusing. Wouldn't it be better to do the copy from the start?

src/inmanta/data/__init__.py
```diff
@@ -4522,9 +4527,29 @@ async def get_resources(
...
+
+        def convert_or_ignore(rvid):
+            """Method to retain backward compatibility, ignore bad ID's"""
+            try:
+                return Id.parse_resource_version_id(rvid)
+            except Exception:
+                return None
+
+        parsed_rv = (convert_or_ignore(id) for id in resource_version_ids)
+        effective_parsed_rv = [id for id in parsed_rv if id is not None]
```
sanderr, line 4539: Could you elaborate on why this is required/correct? Is it just for the unlikely case where we get invalid inputs we didn't reject before? If we merge this into iso6 only, can't we afford to break backwards compatibility?

arnaudsjs (approved):

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
wouterdb: agreed

src/inmanta/data/__init__.py
```diff
@@ -4659,8 +4684,10 @@ async def get_resources_for_version(
                     if no_obj:
                         record = dict(record)
                         record["attributes"] = json.loads(record["attributes"])
+                        parsed_id = resources.Id.parse_id(record["resource_id"])
+                        parsed_id.set_version(record["model"])
+                        record["resource_version_id"] = parsed_id.resource_version_str()
```
sanderr, line 4689: A comment on why we add `resource_version_id` could be useful here. It's obvious in the context of this PR but it might be less so in a couple of months.

src/inmanta/data/__init__.py
```diff
@@ -5342,19 +5372,36 @@ async def update_fields(self, connection: Optional[asyncpg.connection.Connection
...
         dct = super(Resource, self).to_dict()
+
+        # Patch up id, resource_version_id, requires and provides for backward compat with older internal representation
+        # in the DB, it is all resource_id, externally resource_version_id
+        version = self.model
+        rvid = Id.set_version_in_id(self.resource_id, version)
+        dct["resource_version_id"] = rvid
         dct["id"] = dct["resource_version_id"]
+        if "requires" in dct["attributes"]:
+            dct["attributes"]["requires"] = [Id.set_version_in_id(id, version) for id in dct["attributes"]["requires"]]
```
sanderr, line 5383: I hadn't considered this before: could be slightly confusing (because it's not apparent in the typing of this field) that we have resource ids in the database and version ids in the DTO objects. I'm not sure if we can avoid it but we'll have to be very careful with this.

sanderr (changes requested):
I've only gone through a fraction of the PR so far. Releasing intermediary comments.

sanderr (commented):
No further comments.

src/inmanta/data/__init__.py
wouterdb: 1. yes, we need to rewrite the DB
2. cleaned_requires is there precisely because we want to have resource_ids in the DB, while the api does RVID's

src/inmanta/data/__init__.py
wouterdb: This is for backward compatibility, if you pass an invalid ID, it will simply be ignored.

src/inmanta/data/__init__.py
wouterdb: I added it to the `__mangle_dict` method

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -350,6 +371,54 @@ def get_accessor(self, column_name: str, table_prefix: Optional[str] = None) ->
         return super().get_accessor(column_name, table_prefix) + "::" + self.forced_type
 
 
+class ResourceVersionIdColumnType(ColumnType):
+    def __init__(self) -> None:
+        self.nullable = False
```
arnaudsjs, line 376: Why is it not required to call the constructor from the super class here?

arnaudsjs (changes requested):

arnaudsjs (approved):

changelogs/unreleased/4743-remove-rvid.yml
```diff
@@ -0,0 +1,5 @@
+description: Remove resource.resource_version_id fields from the database and use resource id instead 
+change-type: patch
+destination-branches: [master, iso6]
```
wouterdb, line 3: @sanderr this is correct?

sanderr: Yes

sanderr (approved):

src/inmanta/db/versions/v202211230.py
```diff
@@ -0,0 +1,98 @@
...
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    schema = f"""
+    ALTER TABLE ONLY public.resource
+        DROP CONSTRAINT resource_pkey;
+
+    ALTER TABLE ONLY public.resource
+        ADD PRIMARY KEY (environment, model, resource_id);
```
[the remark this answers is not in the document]
wouterdb: good question, what is you opinion after the training?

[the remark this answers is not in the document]
arnaudsjs: The existing index is: `CREATE UNIQUE INDEX resource_env_resourceid_index ON resource (environment, resource_id, model DESC);`. I did some research from curiosity and it seems that you cannot define the order (ASC/DESC) using a primary key but you can when you define an index manually. However, it's possible to indicate that a primary key should use a certain index. So we should be able to do something like this: `ALTER TABLE ONLY public.resource ADD PRIMARY KEY USING INDEX resource_env_resourceid_index`.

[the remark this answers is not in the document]
sanderr: Interesting, I didn't know that.

To answer Wouter's question, we didn't really go in-depth on this, but as far as I can tell for the purposes of the primary key using the existing index should suffice, and for the purposes of index performance it should give the same behavior that we had before this PR. That said, I can't say for sure if there might be any queries that might benefit more from this ordering. Since we don't have a specific query in mind that would, I would be inclined to just use the existing index the way Arnaud commented above.

PR #4964, opened 2022-10-07
issue/4905-remove-static-file-handler

### LINKED ISSUE #4905: Remove dashboard from release
In ISO6 the old dashboard should not longer be included. This means:

- [x] Removing it from the builds (RPM specfile + pyproject.toml in product repo)
- [ ] Removing the static file handler from the server
- [ ] Adding a redirecting from dashboard to console
- [ ] Remove the docs

Only in iso6 and OSS

FloLey:
# Description

1.  Removing the static file handler from the server
2.  Adding a redirecting from dashboard to console
3.  Remove the docs

closes #4905

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


src/inmanta/server/server.py
```diff
@@ -67,51 +63,14 @@ def get_dependencies(self) -> List[str]:
...
-    def setup_dashboard(self) -> None:
-        """
-        If configured, set up tornado to serve the dashboard
-        """
-        if not opt.dash_enable.get() or not self.feature_manager.enabled(dashboard_feature):
-            return
-
-        dashboard_path = opt.dash_path.get()
-        if dashboard_path is None:
-            LOGGER.warning("The dashboard is enabled in the configuration but its path is not configured.")
```
arnaudsjs, line 89: We should also remove all dashboard-related config options from the server.

arnaudsjs (changes requested):
We should also remove the dashboard-related config options from `inmanta-core/misc/inmanta.cfg`

src/inmanta/server/server.py
FloLey: I think this is done now?

tests/test_server.py
```diff
@@ -1387,3 +1386,22 @@ async def test_start_location_no_redirect(server):
...
+    result_url = "http://localhost:%s/console%s" % (port, path)
+    http_client = AsyncHTTPClient()
+    request = HTTPRequest(
+        url=base_url,
+    )
+    response = await http_client.fetch(request, raise_error=False)
+    if path != "error":
+        assert result_url == response.effective_url
+    else:
+        assert "http://localhost:%s/dashboarderror" % port == response.effective_url
```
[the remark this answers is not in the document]
arnaudsjs: Ah. I see. An `HTTPRequest` has `follow_redirects` set to True by default. That's why you get the 404 everywhere. When you disable it, you should get a 3xx when there is a redirect and a 404 when there isn't.

arnaudsjs (changes requested):
The `docs/install/10-features.rst` and the `docs/install/install-server/configure-server.inc` still contain references to the Dashboard. Is this intentional? Or is this part of the ticket that will update the `docs/troubleshooting.rst` file.

docs/install/10-features.rst
```diff
@@ -1,23 +0,0 @@
...
-
- - core.server::dashboard
-
-An example feature file is:
-
-.. code-block:: yaml
-
-    slices:
-        core.server:
-            dashboard: false
```
FloLey, line 23: I think this one can be removed entirely?  as there are no features left? :p

arnaudsjs: Good question. It's true that there are no features left for OSS, but I would only remove everything from line 13 and keep the rest. It's still nice to document that our server supports the concept of a feature. But I think it no longer makes sense to make this a part of the install documentation. Maybe we should move this to `Platform developer documentation`. @sanderr Any opinion on this?

Issue #4886, opened 2022-09-23
Create data collector


Hugo-Inmanta:
- [ ] Setup influxdb database 
- [ ] Setup automatic data collection 
    - [ ] Phase I: test failure rates
    - [ ] Phase II: all other features

PR #4969, opened 2022-10-10
Issue/add data collector

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/inmanta-core/issues/4886
First steps towards adding data collection for predictive test selection.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


predictive_test_selection/collect_test_failure_data.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def send_influx_db_data(test_file: str, failed: bool, time_stamp: int) -> None:
+    if not test_file:
+        # Skipped test file have an empty test_file name
+        return
+    influx_db_string = f"test_result,fqn={test_file} failed={failed} {time_stamp}"
+    print(influx_db_string)
+
+    url = "http://mon.ii.inmanta.com:8086/write?db=predictive_test_selection"
```
Hugo-Inmanta, line 15: This db has not been setup yet

predictive_test_selection/collect_pts_data.py
```diff
@@ -0,0 +1,93 @@
+import os
+import subprocess
+from datetime import datetime
+from enum import IntFlag, auto
+from typing import Any, List, Mapping
+
+# Eveything below is work in progress for part 2 when we already have historic data for the failure rates.
```
Hugo-Inmanta, line 7: This file is not being used at the moment but will be used during phase 2 of the data collection

wouterdb:
@arnaudsjs is it OK to add this in core for now? It can move out in a later stage, but not it is kind of convenient here.

wouterdb (changes requested):
Changes requested offline

predictive_test_selection/collect_test_failure_data.py
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        # Get latest commit hash:
+        data["commit_hash"] = subprocess.check_output(["git", "log", "--pretty=%H", "-1"]).strip()
+
+        # Get current branch
+        cmd = ["git", "rev-parse", "--abbrev-ref", "HEAD"]
+        current_branch = subprocess.check_output(cmd).strip().decode()
+
+        cmd = ["git", "diff", f"master...{current_branch}", "--name-only"]
```
arnaudsjs, line 70: This code doesn't take into account the situation where the diff is empty. In that case we don't want to collect any data, I guess.

predictive_test_selection/collect_test_failure_data.py
```diff
@@ -0,0 +1,168 @@
...
+from datetime import datetime
+from enum import IntFlag, auto
+from typing import Any, List, Mapping, Sequence, Tuple
+
+import requests
+
+LOGGER = logging.getLogger(__name__)
+
+
+class CommonFileExtension(IntFlag):
```
arnaudsjs, line 15: I am curious what the impact is if we at some point decide to add extra extensions. Does historical data we collected become useless then?

predictive_test_selection/collect_test_failure_data.py
```diff
@@ -0,0 +1,168 @@
...
+                f" failed_as_int={failed},n_changes={self.code_change_data['n_changes']},"
+                f"file_extension={self.code_change_data['file_extension']},"
+                f"file_cardinality={self.code_change_data['file_cardinality']}"
+            )
+            for test_fqn, failed in self.test_result_data.items()
+        ]
+        return "\n".join(data_points)
+
+    def send_influxdb_data(self, dry_run: bool = True) -> None:
+        url = "http://mon.ii.inmanta.com:8086/write?db=predictive_test_selection&precision=s"
```
arnaudsjs, line 150: I would move this URL to a global variable in the file.

predictive_test_selection/collect_test_failure_data.py
```diff
@@ -0,0 +1,168 @@
...
+                except KeyError:
+                    idx = known_exts.index(CommonFileExtension["other"])
+                out[idx] = 1
+            return out
+
+        data["file_extension"] = get_extension_vector()
+
+        def get_changes() -> List[int]:
+            n_changes: List[int] = []
+            for n_days_ago in [3, 14, 56]:
```
arnaudsjs, line 92: The list `[3, 14, 56]` can be considered a configuration parameter of this script. As such, It would make it into a global variable to make that clear.

predictive_test_selection/collect_test_failure_data.py
```diff
@@ -0,0 +1,168 @@
...
+                if test_case.find("skipped") is not None:  # test got skipped -> we don't record anything
+                    continue
+
+                test_file: str = test_case.get("classname")
+                test_name: str = test_case.get("name")
+                test_fqn: str = ".".join((test_file, test_name))
+
+                test_failed: int = int(test_case.find("failure") is not None)
+
+                self.test_result_data[test_fqn] = test_failed
```
arnaudsjs, line 128:
```suggestion
                self.test_result_data[test_fqn] = test_failed
        assert self.test_result_data
```

predictive_test_selection/collect_test_failure_data.py
```diff
@@ -0,0 +1,168 @@
...
+        data["file_cardinality"] = len(changed_files)
+
+        def get_extension_vector() -> List[int]:
+            extensions = set([os.path.splitext(file)[1].replace(".", "") for file in changed_files])
+
+            out = [0] * 8
+            known_exts = [ext for ext in CommonFileExtension]
+            for ext in extensions:
+                try:
+                    idx = known_exts.index(CommonFileExtension[ext])
```
arnaudsjs, line 82: This operation is rather slow, because it will search through the list every time. It would be better to change to a data structure that allows constant time lookups.

arnaudsjs (changes requested):

arnaudsjs:
> @arnaudsjs is it OK to add this in core for now? It can move out in a later stage, but not it is kind of convenient here.

I think it's fine to keep it in here for now.

predictive_test_selection/collect_test_failure_data.py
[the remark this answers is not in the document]
Hugo-Inmanta: Ah, great point, I need to think a bit more about this.

predictive_test_selection/collect_test_failure_data.py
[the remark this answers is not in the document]
Hugo-Inmanta: Oops indeed, IntFlag is an artifact from a previous implementation, I'll remove it.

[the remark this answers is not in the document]
arnaudsjs: I think you executed this command on a project containing a venv. That would explain why the list contains so many extensions of binary files that are not present in our repository.

predictive_test_selection/README.md
```diff
@@ -0,0 +1,26 @@
...
+    x n_changes_56 int NOT NULL,
+    x file_cardinality int NOT NULL,
+    x file_extension text NOT NULL,
+      target_cardinality int NOT NULL,
+      failure_rate_7 real NOT NULL,
+      failure_rate_14 real NOT NULL,
+      failure_rate_28 real NOT NULL,
+      failure_rate_56 real NOT NULL,
+      test_failed int NOT NULL
+    """
```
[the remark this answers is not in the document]
Hugo-Inmanta: Indeed, this needs a lot more documentation

predictive_test_selection/collect_test_failure_data.py
Hugo-Inmanta: What could cause this situation to arise ? And indeed, I don't think we want to collect any data in this case.

predictive_test_selection/collect_test_failure_data.py
arnaudsjs: I think that's a good idea. That way decouple the data collection part from the model generation part, which gives us more flexibility. And I guess that the store overhead on the database will be limited.

predictive_test_selection/collect_test_failure_data.py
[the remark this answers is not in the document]
Hugo-Inmanta: The script will be called from [there](https://github.com/inmanta/irt/pull/1355/files#diff-62e96281f2c1e5b9441fe9936a16fa518a04ccfb2c95101fad5a0c154bbcd014): 
Would it work to fetch the branch name and pass it from there with `env.BRANCH_NAME` ?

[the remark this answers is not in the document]
arnaudsjs: The `env.BRANCH_NAME` environment variable of the Jenkinsfile contains the name of the feature branch, but here we want to know the name of the development branch. So I don't think this is going to work.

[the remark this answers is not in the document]
Hugo-Inmanta: Ok I will parse the changelog file then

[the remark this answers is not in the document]
arnaudsjs: I think we will have to travel up in the commit history of the feature branch until we reach the first commit that is also present in a development branch. There might be an easier way, but I don't really see how.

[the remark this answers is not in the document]
Hugo-Inmanta: I added the checkout of the full history over there: https://github.com/inmanta/irt/pull/1355/files#diff-62e96281f2c1e5b9441fe9936a16fa518a04ccfb2c95101fad5a0c154bbcd014R151-R153

predictive_test_selection/data_collector.py
```diff
@@ -0,0 +1,236 @@
...
+        modification_count: List[int] = []
+        for n_days_ago in CHANGE_LOOK_BACK:
+            acc: int = 0
+            for file in self.changed_files:
+                ps = subprocess.Popen(
+                    ["git", "log", f"--after='{n_days_ago} days ago'", "--format=oneline", f"{file}"],
+                    stdout=subprocess.PIPE,
+                )
+                try:
+                    acc += int(subprocess.check_output(["wc", "-l"], stdin=ps.stdout))
```
wouterdb, line 114: In python, I would not go for the pipe, but for

```python
    acc = sum(1 for line in ps.stdout.readline())
```

but that is only a slight preference.

predictive_test_selection/data_collector.py
```diff
@@ -0,0 +1,236 @@
...
+        Computes the set of file extensions present in this code change
+        """
+        self.file_extensions = set([os.path.splitext(file)[1].replace(".", "") for file in self.changed_files])
+
+    def get_modifications_data_format(self) -> str:
+        """
+        Converts the modification_count (List[int]) to a condensed csv format.
+        e.g [1, 2, 3] becomes "1,2,3"
+        """
+        return str(self.modification_count).replace(" ", "")[1:-1]
```
wouterdb, line 133: string manipulation to format structured data is something that scares me
```suggestion
        return ",".join((str(count) for count in self.modification_count))
```

predictive_test_selection/data_collector.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def get_file_extensions_data_format(self) -> str:
+        """
+        Converts the file_extensions (Set[str]) to a condensed csv format.
+        e.g {'txt', 'md', 'py'} becomes 'txt,md,py'
+        """
+        return str(self.file_extensions).replace(" ", "").replace("'", "").replace('"', "")[1:-1]
```
wouterdb, line 140: string manipulation to format structured data is something that scares me
```suggestion
                return ",".join(self.file_extensions)
```

predictive_test_selection/data_collector.py
```diff
@@ -0,0 +1,236 @@
...
+
+                self.test_result_data[test_fqn] = test_failed
+
+        assert self.test_result_data
+
+    def __iter__(self):
+        yield from self.test_result_data.items()
+
+
+@dataclass
```
wouterdb, line 179: Why is this a dataclass?

predictive_test_selection/data_collector.py
```diff
@@ -0,0 +1,236 @@
...
+class DataParser:
+    """
+    This class fetches all information regarding a specific code change and the tests that are run on this specific commit
+    and sends it all to the influxdb database.
+    """
+
+    code_change_data: CodeChange = field(default_factory=lambda: CodeChange())
+    test_result_data: TestResult = field(default_factory=lambda: TestResult())
+
+    def parse(self, dry_run: bool):
```
wouterdb, line 189:
```suggestion
    def run(self, dry_run: bool):
```

wouterdb (changes requested):

predictive_test_selection/data_collector.py
```diff
@@ -0,0 +1,236 @@
...
+    '$ py.test --junit-xml=junit-py39.xml'):
+    - test_result_data -> Maps the fully qualified test name <test_file_name>.<test_case> to an integer denoting whether
+      this test failed or not.
+    """
+
+    test_result_data: MutableMapping[str, int] = field(init=False)
+
+    def parse(self) -> None:
+        self.test_result_data = {}
+        path: str = "junit-py39.xml"
```
arnaudsjs, line 156: We cannot hard-code this to `junit-py39.xml`. ISO4 runs on Python36, so for iso4 this will be `junit-py36.xml`.

arnaudsjs (changes requested):

predictive_test_selection/data_collector.py
```diff
@@ -0,0 +1,236 @@
...
+        self._count_modifications()
+
+    def _compute_changed_files(self, current_branch: str) -> None:
+        """
+        Finds the development branch that is the closest to this code change and sets the relevant attributes accordingly.
+        The distance metric used is the total number of files in the diff with the current branch
+        """
+        max_cardinality: Union[float, int] = float("inf")
+
+        for dev_branch in DEV_BRANCHES:
```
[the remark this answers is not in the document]
bartv: I think needs need to get merged and start running asap. it makes no sense to make improvements to something we are not even sure will bring any benefits. It is still an experiment, we need data before there is anything useful to say about the approach.

predictive_test_selection/data_collector.py
```diff
@@ -0,0 +1,236 @@
...
+        e.g [1, 2, 3] becomes "1,2,3"
+        """
+        return ",".join((str(count) for count in self.modification_count))
+
+    def get_file_extensions_data_format(self) -> str:
+        """
+        Converts the file_extensions (Set[str]) to a condensed csv format.
+        e.g {'txt', 'md', 'py'} becomes 'txt,md,py'
+        """
+        return ",".join(self.file_extensions)
```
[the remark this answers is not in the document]
Hugo-Inmanta: As far as I can tell, no.
This string will be unraveled and turned into a fixed-size bit vector by the data pre-processor and the order won't matter there.

predictive_test_selection/data_collector.py
Hugo-Inmanta: No valid reason to use a dataclass there, removing it.

predictive_test_selection/data_collector.py
Hugo-Inmanta: I didn't manage to make it work this way in a reasonable time unfortunately.

wouterdb: ok!

wouterdb (approved):

arnaudsjs (approved):

PR #4977, opened 2022-10-13
issue/4908-add-module-plugin-deprecation

### LINKED ISSUE #4908: Add module and plugin deprecation mechanism
The goal of this ticket is to add a deprecation mechanism to core for modules and plugins. The proposal is to:

- [x] Add a deprecated boolean field to module.yml When this is set to true, the compiler should issue a deprecation warning when the module is imported
- [x] Add a deprecation and replacedby field to the @plugin decorator. When set to true it should print out a warning that the plugin is deprecated. When replaceby is set the message should also say something like `This plugin has been replaced by {replacedby}`

If the current code does not support unkown keyword arguments it is best to create a new decorator so that we do not make modules incompatible with older releases. Similar for the additional keyword in the module.yml If required we need to support it as well in ISO5.

ISO4 modules are not updated so that should not be an issue.

FloLey:
# Description

The goal of this commit is to add a deprecation mechanism to core for modules and plugins.

 - Add a deprecated boolean field to module.yml When this is set to true, the compiler should issue a deprecation warning when the module is imported
 - Add a deprecated and replaced_by field to the @plugin decorator. When set to true it should print out a warning that the plugin is deprecated. When replaced_by is set the message will also say by which plugin the current one has been replaced
 

closes #4908

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -75,7 +75,7 @@ that converts a string to uppercase:
 
 The plugin decorator also accept the ``deprecated`` and ``replaced_by`` fields. When the ``deprecated`` field
 is set to ``True`` it prints out a warning that the plugin is deprecated. with the ``replaced_by`` string it is possible
-to tell by which plugin the current one has been replaced. for example if the code below is run
+to tell by which plugin the current one has been replaced. for example if the following plugin is called,
 
 .. code-block:: python
     :linenos:
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -161,7 +161,7 @@ the previous example would then look like this:
     try:
         from inmanta.plugins import deprecated
     except ImportError:
-    deprecated = lambda f=None, **kwargs: f if f is not None else deprecated
+        deprecated = lambda function=None, **kwargs: function if function is not None else deprecated
 
 
     @deprecated(replaced_by="my_new_plugin")
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -73,6 +73,31 @@ that converts a string to uppercase:
         return value.upper()
 
 
+The plugin decorator also accept the ``deprecated`` and ``replaced_by`` fields. When the ``deprecated`` field
+is set to ``True`` it prints out a warning that the plugin is deprecated. with the ``replaced_by`` string it is possible
+to tell by which plugin the current one has been replaced. for example if the code below is run
+
+.. code-block:: python
+    :linenos:
+
+    from inmanta.plugins import plugin
+
+    @plugin(deprecated=True, replaced_by="my_new_plugin")
+    def printf():
+        """
+            Prints inmanta
+        """
+        print("inmanta")
+
+
+it wil give following warning:
+
+.. code-block::
+
+    Plugin 'get_one' in module 'inmanta_plugins.<module_name>' is deprecated. It should be replaced by 'my_new_plugin'
+
+
+
 This plugin can be tested with:
 
 .. code-block:: inmanta
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -151,7 +151,7 @@ Should the replace_by argument be omitted, the warning would look like this:
 If you want your module to stay compatible with older versions of inmanta you will also need to add a little piece of code that changes how
 :func:`~inmanta.plugins.deprecated` is imported as it does not exist in all versions.
 
-The previous example would then look like this. For older inmanta versions, the replace the decorator with a no-op.
+The previous example would then look like this. For older inmanta versions, replace the decorator with a no-op.
 
 .. code-block:: python
     :linenos:
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -142,7 +142,7 @@ it will give following warning:
 
     Plugin 'printf' in module 'inmanta_plugins.<module_name>' is deprecated. It should be replaced by 'my_new_plugin'
 
-Should de replace_by argument be omitted, the warning would look like this:
+Should the replace_by argument be omitted, the warning would look like this:
 
 .. code-block::
 
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -125,7 +125,7 @@ for example if the plugin below is called:
 .. code-block:: python
     :linenos:
 
-    from inmanta.plugins import plugin,deprecated
+    from inmanta.plugins import plugin, deprecated
 
     @deprecated(replaced_by="my_new_plugin")
     @plugin
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -94,7 +94,7 @@ it wil give following warning:
 
 .. code-block::
 
-    Plugin 'get_one' in module 'inmanta_plugins.<module_name>' is deprecated. It should be replaced by 'my_new_plugin'
+    Plugin 'printf' in module 'inmanta_plugins.<module_name>' is deprecated. It should be replaced by 'my_new_plugin'
 
 
 
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -2,6 +2,11 @@
 
 Developing Plugins
 *********************
+
+
+Adding new plugins
+========================
+
 Plugins provide :ref:`functions<lang-plugins>` that can be called from the :term:`DSL`. This is the
 primary mechanism to interface Python code with the orchestration model at compile time. For Example,
 this mechanism is also used for std::template and std::file. In addition to this, Inmanta also registers all
@@ -73,31 +78,6 @@ that converts a string to uppercase:
         return value.upper()
 
 
-The plugin decorator also accept the ``deprecated`` and ``replaced_by`` fields. When the ``deprecated`` field
-is set to ``True`` it prints out a warning that the plugin is deprecated. with the ``replaced_by`` string it is possible
-to tell by which plugin the current one has been replaced. for example if the following plugin is called,
-
-.. code-block:: python
-    :linenos:
-
-    from inmanta.plugins import plugin
-
-    @plugin(deprecated=True, replaced_by="my_new_plugin")
-    def printf():
-        """
-            Prints inmanta
-        """
-        print("inmanta")
-
-
-it wil give following warning:
-
-.. code-block::
-
-    Plugin 'printf' in module 'inmanta_plugins.<module_name>' is deprecated. It should be replaced by 'my_new_plugin'
-
-
-
 This plugin can be tested with:
 
 .. code-block:: inmanta
@@ -128,3 +108,68 @@ see :ref:`moddev-module`.
 
 .. todo:: context
 .. todo:: new statements
+
+
+
+
+Deprecate plugins
+========================
+
+
+To deprecate a plugin the :func:`~inmanta.plugins.deprecated` decorator can be used in combination with the :func:`~inmanta.plugins.plugin`
+decorator. Using this decorator will log a warning message when the function is called. This decorator also accepts an
+optional argument ``replaced_by`` which can be used to potentially improve the warning message by telling which other
+plugin should be used in the place of the current one.
+
+for example if the plugin below is called:
+
+.. code-block:: python
+    :linenos:
+
+    from inmanta.plugins import plugin,deprecated
+
+    @deprecated(replaced_by="my_new_plugin")
+    @plugin
+    def printf():
+        """
+            Prints inmanta
+        """
+        print("inmanta")
+
+
+it will give following warning:
+
+.. code-block::
+
+    Plugin 'printf' in module 'inmanta_plugins.<module_name>' is deprecated. It should be replaced by 'my_new_plugin'
+
+Should de replace_by argument be omitted, the warning would look like this:
+
+.. code-block::
+
+    Plugin 'printf' in module 'inmanta_plugins.<module_name>' is deprecated.
+
+If you want your module to stay backwards compatible you will also need to add a little piece of code that changes how
+:func:`~inmanta.plugins.deprecated` is imported as it does not exist in all versions.
+
+the previous example would then look like this:
+
+.. code-block:: python
+    :linenos:
+
+    from inmanta.plugins import plugin
+
+    try:
+        from inmanta.plugins import deprecated
+    except ImportError:
+    deprecated = lambda f=None, **kwargs: f if f is not None else deprecated
+
+
+    @deprecated(replaced_by="my_new_plugin")
+    @plugin
+    def printf():
+        """
+            Prints inmanta
+        """
+        print("inmanta")
+
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -115,7 +115,6 @@ see :ref:`moddev-module`.
 Deprecate plugins
 ========================
 
-
 To deprecate a plugin the :func:`~inmanta.plugins.deprecated` decorator can be used in combination with the :func:`~inmanta.plugins.plugin`
 decorator. Using this decorator will log a warning message when the function is called. This decorator also accepts an
 optional argument ``replaced_by`` which can be used to potentially improve the warning message by telling which other
--- a/docs/model_developers/plugins.rst
+++ b/docs/model_developers/plugins.rst
@@ -151,7 +151,7 @@ Should the replace_by argument be omitted, the warning would look like this:
 If you want your module to stay compatible with older versions of inmanta you will also need to add a little piece of code that changes how
 :func:`~inmanta.plugins.deprecated` is imported as it does not exist in all versions.
 
-the previous example would then look like this:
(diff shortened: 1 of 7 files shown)


tests/compiler/test_basics.py
```diff
@@ -427,3 +429,62 @@ def test_unpack_null_dictionary(snippetcompiler):
             "std::replace(hello_world,**dct) ({dir}/main.cf:4))"
         ),
     )
+
+
```
FloLey, line 433: Is there another way to test this than creating a project and compiling it?

tests/compiler/test_basics.py
```diff
@@ -427,3 +429,62 @@ def test_unpack_null_dictionary(snippetcompiler):
...
+@pytest.mark.parametrize_any(
+    "deprecated, replaced_by",
+    [(True, None), (True, '"new_function"'), (False, None), (False, '"new_function"')],
+)
+def test_modules_plugin_deprecation(
```
FloLey, line 438: Is there another way to test this than by creating a project and compiling it?

wouterdb: v1 modules are a lot faster, so that would help .

src/inmanta/plugins.py
```diff
@@ -421,6 +426,11 @@ def __call__(self, *args: object, **kwargs: object) -> object:
         The function call itself
         """
         self.check_requirements()
+        if self.deprecated:
```
sanderr, line 429: Perhaps we could move this before `check_requirements` so the user doesn't sink time in fixing requirements only to discover they should use a different plugin (which might not have these requirements).

src/inmanta/plugins.py
```diff
@@ -421,6 +426,11 @@ def __call__(self, *args: object, **kwargs: object) -> object:
...
+            msg: str = f"Plugin '{self.__function_name__}' in module '{self.__module__}' is deprecated. "
+            if self.replaced_by:
+                msg += f"It should be replaced by '{self.replaced_by}'"
```
sanderr, line 432:
```suggestion
            msg: str = f"Plugin '{self.__function_name__}' in module '{self.__module__}' is deprecated."
            if self.replaced_by:
                msg += f" It should be replaced by '{self.replaced_by}'"
```
No trailing space if `replaced_by` is not defined.

tests/compiler/test_basics.py
sanderr: We could create a plugin and call it with plain Python. But given how much the compiler interfaces with this (it triggers the loading of Python files, it collects all registered plugins, ...) I think we get a lot more confidence out of an actual end-to-end test like this.

sanderr (changes requested):

docs/model_developers/plugins.rst
```diff
@@ -128,3 +108,67 @@ see :ref:`moddev-module`.
...
+        print("inmanta")
+
+
+it will give following warning:
+
+.. code-block::
+
+    Plugin 'printf' in module 'inmanta_plugins.<module_name>' is deprecated. It should be replaced by 'my_new_plugin'
+
+Should de replace_by argument be omitted, the warning would look like this:
```
sanderr, line 145:
```suggestion
Should the replace_by argument be omitted, the warning would look like this:
```

docs/model_developers/plugins.rst
```diff
@@ -128,3 +108,67 @@ see :ref:`moddev-module`.
...
+
+.. code-block::
+
+    Plugin 'printf' in module 'inmanta_plugins.<module_name>' is deprecated.
+
+If you want your module to stay backwards compatible you will also need to add a little piece of code that changes how
```
sanderr, line 151:
```suggestion
If you want your module to stay compatible with older versions of inmanta you will also need to add a little piece of code that changes how
```

docs/model_developers/plugins.rst
```diff
@@ -128,3 +108,67 @@ see :ref:`moddev-module`.
...
+
+.. code-block:: python
+    :linenos:
+
+    from inmanta.plugins import plugin
+
+    try:
+        from inmanta.plugins import deprecated
+    except ImportError:
+    deprecated = lambda f=None, **kwargs: f if f is not None else deprecated
```
sanderr, line 164:
```suggestion
        deprecated = lambda function=None, **kwargs: function if function is not None else deprecated
```
Just in case someone would pass `function` as a key-word argument, make sure the name matches the one from core.

docs/model_developers/plugins.rst
```diff
@@ -128,3 +108,67 @@ see :ref:`moddev-module`.
...
+Should de replace_by argument be omitted, the warning would look like this:
+
+.. code-block::
+
+    Plugin 'printf' in module 'inmanta_plugins.<module_name>' is deprecated.
+
+If you want your module to stay backwards compatible you will also need to add a little piece of code that changes how
+:func:`~inmanta.plugins.deprecated` is imported as it does not exist in all versions.
+
+the previous example would then look like this:
```
sanderr, line 154:
```suggestion
The previous example would then look like this. For older inmanta versions, the replace the decorator with a no-op.
```

src/inmanta/plugins.py
```diff
@@ -558,3 +568,16 @@ def wrapper(self, *args: object, **kwargs: object) -> Any:
     elif function is not None:
         fnc = curry_name(commands=commands, emits_statements=emits_statements, allow_unknown=allow_unknown)
         return fnc(function)
+
+
+@stable_api
+def deprecated(
+    function: Optional[Callable] = None, replaced_by: Optional[str] = None, **kwargs: Dict[str, object]
```
sanderr, line 575:
```suggestion
    function: Optional[Callable] = None, *, replaced_by: Optional[str] = None, **kwargs: Dict[str, object]
```
No positional args other than `function`, remember? Otherwise our lambda is not compatible.

src/inmanta/plugins.py
```diff
@@ -177,6 +177,18 @@ def get_functions(cls) -> Dict[str, "Type[Plugin]"]:
         """
         return dict(cls.__functions)
 
+    @classmethod
+    def deprecate_function(cls, fnc: Callable, replaced_by: Optional[str] = None) -> None:
+        name = fnc.__name__
```
sanderr, line 182: Could you add a test case where you use this new decorator together with a plugin with a custom name? i.e.
```python
@deprecated
@plugin("pluginnamedoesnotmatchfunctioname")
def myfunction():
    pass
```

I expect this will currently not work. I don't know how to fix it yet, let's first see if my hypothesis is right.

src/inmanta/plugins.py
FloLey: right, I always forget about the *. thanks :p

src/inmanta/plugins.py
FloLey: indeed : `Can not deprecate plugin test_module::get_one as it does not exist` :(

src/inmanta/plugins.py
```diff
@@ -177,6 +177,18 @@ def get_functions(cls) -> Dict[str, "Type[Plugin]"]:
...
+        ns_parts = str(fnc.__module__).split(".")
+        ns_parts.append(name)
+        full_name = "::".join(ns_parts[1:])
+        if full_name in cls.get_functions():
+            cls.__functions[full_name].deprecated = True
+            cls.__functions[full_name].replaced_by = replaced_by
+        else:
+            raise Exception(f"Can not deprecate plugin {full_name} as it does not exist")
```
wouterdb, line 190: I think this will also be raise if you do 

```
    @plugin
    @deprecated(replaced_by="my_new_plugin")
    def printf():
```

so I don't think this is the best error message.

wouterdb (changes requested):

src/inmanta/plugins.py
FloLey: wouldn't it be possible to add a field to the dictionary with the "true name" and verify the "true name" is present?

FloLey: @sanderr I didn't want to do a reverse lookup so I've added an extra field. What do you think about that? The naming should still be changed but I has no ideas

src/inmanta/plugins.py
```diff
@@ -172,6 +177,28 @@ def get_functions(cls) -> Dict[str, "Type[Plugin]"]:
         """
         return dict(cls.__functions)
 
+    @classmethod
+    def get_originals(cls) -> Dict[str, str]:
+        """
+        Get the original name of all the functions with the associated function name
+        """
+        return dict(cls.__function_name_mapping)
```
sanderr, line 185: I don't think this is used anymore.
```suggestion
    @classmethod
    def get_originals(cls) -> Dict[str, str]:
        """
        Get the original name of all the functions with the associated function name
        """
        return dict(cls.__function_name_mapping)
```

src/inmanta/plugins.py
```diff
@@ -172,6 +177,28 @@ def get_functions(cls) -> Dict[str, "Type[Plugin]"]:
...
+        name = fnc.__name__
+        ns_parts = str(fnc.__module__).split(".")
+        ns_parts.append(name)
+        full_name = "::".join(ns_parts[1:])
+        if hasattr(fnc, "plugin_name"):
+            cls.__functions[fnc.plugin_name].deprecated = True
+            cls.__functions[fnc.plugin_name].replaced_by = replaced_by
+        else:
+            raise Exception(
+                f"Can not deprecate plugin '{full_name}': The '@deprecated' decorator should be used in combination with the "
```
sanderr, line 198: I don't think it makes sense to use the inmanta plugin name in this error message because it's not a plugin, therefore it doesn't have an inmanta name, only a Python name.

src/inmanta/plugins.py
```diff
@@ -544,3 +585,16 @@ def wrapper(self, *args: object, **kwargs: object) -> Any:
     elif function is not None:
         fnc = curry_name(commands=commands, emits_statements=emits_statements, allow_unknown=allow_unknown)
         return fnc(function)
+
+
+@stable_api
+def deprecated(
+    function: Optional[Callable] = None, *, replaced_by: Optional[str] = None, **kwargs: Dict[str, object]
```
sanderr, line 592:
```suggestion
    function: Optional[Callable] = None, *, replaced_by: Optional[str] = None, **kwargs: abc.Mapping[str, object]
```
Could you also add a comment about why we add `**kwargs` in the first place?

src/inmanta/plugins.py
```diff
@@ -544,3 +585,16 @@ def wrapper(self, *args: object, **kwargs: object) -> Any:
...
+) -> Callable:  # noqa: H801
```
sanderr, line 593: I don't find this violation code anywhere, what does it mean and why do we need it?

sanderr (changes requested):
There is currently some duplication between `PluginMeta.add_function`, `PluginMeta.deprecate_function` and the `plugin` decorator. I think that we could keep the logic in the `plugin` decorator, and have the other locations just access `plugin_class.__function_.__plugin_name__`. Then we only need the logic in one place.

Or potentially even better:
1. `Plugin` has attributes `__function_name__` and `__fq_plugin_name__`, both set by the `plugin` decorator.
2. The `plugin` decorator sets `f.__plugin__ = theplugin` instead of `f.__plugin_name__ = thename`
3. The `deprecated` decorator could deprecate the plugin itself without having to delegate to the `PluginMeta`.
4. `PluginMeta.add_function` reads `plugin_class.__fq_plugin_name__` instead of reconstructing it.

What do you think?

src/inmanta/plugins.py
```diff
@@ -543,6 +553,13 @@ def wrapper(self, *args: object, **kwargs: object) -> Any:
             if name is None:
                 name = fnc.__name__
 
+            ns_parts = str(fnc.__module__).split(".")
+            ns_parts.append(name)
+            if ns_parts[0] != const.PLUGINS_PACKAGE:
+                raise Exception("All plugin modules should be loaded in the %s package" % const.PLUGINS_PACKAGE)
+
+            fnc.plugin_name = "::".join(ns_parts[1:])
```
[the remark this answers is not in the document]
FloLey: We discussed it with @sanderr and we both preferred the current option. My only argument is that I find it easier to read but I don't really mind changing it :)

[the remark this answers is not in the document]
sanderr: That was the previous implementation. We then came up with this new idea and considered it slightly cleaner (though the current implementation still needs a bit of refinement): by attaching metadata to the function returned by the `plugin` decorator, we keep the lookup simple: we only need to maintain one mapping, and we only need to convert between Python and inmanta names once.
Essentially, we just pass metadata over the decorators directly rather than having to extend the registry and lookup class to handle Python names.
We decided to give it a go and see what you thought. Do you prefer the old mechanism?

src/inmanta/plugins.py
FloLey: actually it is present on the plugin decorator and I didn't remove it when I copy/pasted its definition. I'll remove it at both places and see if something complains

sanderr: I see, it could be an old violation that no longer exists.

FloLey:
@sanderr is this what you had in mind? I like it better than the original version. I still have a little question that you'll find in the code of the PR

src/inmanta/plugins.py
```diff
@@ -416,10 +417,20 @@ def check_requirements(self) -> None:
...
+    @classmethod
+    def deprecate_function(cls, replaced_by: Optional[str] = None) -> None:
+        cls.deprecated = True
+        cls.replaced_by = replaced_by
+
     def __call__(self, *args: object, **kwargs: object) -> object:
         """
         The function call itself
         """
+        if Plugin.deprecated:
```
FloLey, line 429: @sanderr I noticed If I called `self.deprecated `it also worked. If I understand correctly python will first look for "deprecated" in the instance variables, and if it doesn't find it, it will look at the class variables. So I guess it is better to tell it immediately to look at the class variables or is there a reason to use `self`?

src/inmanta/plugins.py
```diff
@@ -544,3 +564,26 @@ def wrapper(self, *args: object, **kwargs: object) -> Any:
...
+    the kwargs are currently ignored but where added in case we want to add something later on.
+    """
+
+    def inner(fnc: Callable):
+        if hasattr(fnc, "__plugin__"):
+            fnc.__plugin__.deprecate_function(replaced_by)
+        else:
+            raise Exception(
+                f"Can not deprecate '{fnc.__name__}': The '@deprecated' decorator should be used in combination with the "
+                f"'@plugin' decorator and should be placed first."
```
sanderr, line 583: Only noticed this on my nth read of this line but "first" is actually quite ambiguous. It could mean first as we intend or as the inner decorator (the first wrapper). I'm not sure if my suggestion is good.
```suggestion
                f"'@plugin' decorator and should be placed at the top."
```

sanderr (approved):

src/inmanta/plugins.py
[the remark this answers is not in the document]
sanderr: @wouterdb after the latest changes I quite like the current implementation. It definitely feels simpler to me than the previous one.

src/inmanta/plugins.py
sanderr: I have never managed to make my mind up on this. I generally opt for `self` because it's polymorphic. But if (which should never be the case but still) the name is overridden on an instance, you'll get that value instead. A third option is `self.__class__.deprecated` but that's even worse imo.
Concluding, I don't think we have a strong style regarding this, so until we decide to create an ADR for it I think you can take whatever approach you prefer, though perhaps we should try to be consistent at least within each class.

tests/compiler/test_basics.py
```diff
@@ -588,7 +588,7 @@ def get_one() -> "int":
     with pytest.raises(Exception) as e:
         compiler.do_compile()
     assert (
-        "Can not deprecate plugin 'test_module::get_one': The '@deprecated' decorator should be used in combination with the "
+        "Can not deprecate 'get_one': The '@deprecated' decorator should be used in combination with the "
         "'@plugin' decorator and should be placed first." in e.value.msg
```
wouterdb, line 592:
```suggestion
        "'@plugin' decorator and should be placed at the top." in e.value.msg
```

src/inmanta/plugins.py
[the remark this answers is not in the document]
wouterdb: I like it as well, if it is correct. It changes an internal API, so it might break the plugins, but I assume you checked this?

wouterdb (approved):

src/inmanta/plugins.py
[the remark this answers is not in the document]
sanderr: I did. As far as I can see there should be no fallout. We'll only know for sure the morning after it gets merged of course.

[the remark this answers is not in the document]
FloLey: I tested it with a little module and model I've been using to develop this feature. Should I have tested it more than that if so, how?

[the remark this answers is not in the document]
sanderr: I don't think so. It's just a question of what, if anything, we have overlooked. For example, pytest-inmanta does some magic with plugin loading. I've checked the code and I think it shouldn't be affected by this change. But the only way to really know for sure with this many components playing together is to run the full set of integration tests.

PR #4978, opened 2022-10-13
Issue/1888 remove depracated option variable iso5

### LINKED ISSUE #1888: remove support for leaving nullable attribute unassigned

FloLey:
# Description

add a changelog deprecation note stating the deprecation of Jinja's is defined for nullable attributes

closes #1888 

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:


changelogs/unreleased/1888-depreacation-jinja-is-defined.yml
```diff
@@ -0,0 +1,7 @@
+---
+description: Remove support for leaving nullable attribute unassigned. You should make sure optional variables are always assigned a value. This changes the behaviour of 'is defined' in Jinja templates. You should now use 'is not none' instead.
```
sanderr, line 2: "Remove support" is not really accurate. In iso5 we've only deprecated it, for which we already released a change entry (though it might not be a bad idea to repeat this here for context). Also "You should now use" is not completely accurate: you should only use it if you make the change to explicit `null`. Perhaps like this (I'm not sure myself, feel free to object)?
```suggestion
description: When you make the change to use explicit `null` for plain attributes (due to previous deprecation of leaving nullable attributes unassigned), the behavior of `is defined` in Jinja templates changes. You should update Jinja templates to use 'is not none' instead of `is defined` together with the change to explicit `null`.
```

sanderr (changes requested):

changelogs/unreleased/1888-depreacation-jinja-is-defined.yml
FloLey: I'm not sure I understand this part: `(due to previous deprecation of leaving nullable attributes unassigned)` as it has never been mentioned before on iso5 I think?

sanderr: It appears you're right, I assumed we'd added a message when we introduced the deprecation warning. In that case your message was good, except I would rephrase "remove support for" as "deprecated" (because support will only be dropped in iso6). And optionally rephrase the "you should now use" but that's of lesser importance.

sanderr (approved):

PR #4979, opened 2022-10-13
Deprecate V1 modules

### LINKED ISSUE #4941: deprecate v1 modules
While we're currently not ready yet to drop support for v1 modules, it would be good to already mark them as deprecated so we have the freedom to drop them when we are ready.

- [x] add a deprecation note to iso5 and iso6
- [x] add a deprecation warning when v1 modules are used?

arnaudsjs:
# Description

* Show a deprecation message when a V1 module is loaded. 
* Add a deprecation message to the changelog

closes #4941 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/4941-deprecate-v1-modules.yml
+++ b/changelogs/unreleased/4941-deprecate-v1-modules.yml
@@ -1,0 +1,8 @@
+---
+description: V1 modules are deprecated. Supported for V1 modules will be removed in one of the next major releases
+issue-nr: 4941
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master, iso5]
+sections:
+  deprecation-note: "{{description}}"
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -549,7 +549,7 @@ class PythonEnvironment:
             if conflicts:
                 raise ConflictingRequirements("\n".join(conflicts))
             raise PipInstallError(
-                f"Process {cmd} exited with return code {return_code}."
+                f"Process {cmd} exited with return code {return_code}. "
                 "Increase the verbosity level with the -v option for more information."
             )
 
--- a/src/inmanta/module.py
+++ b/src/inmanta/module.py
@@ -28,6 +28,7 @@ import sys
 import tempfile
 import traceback
 import types
+import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from configparser import ConfigParser
@@ -2202,7 +2203,11 @@ class Project(ModuleLike[ProjectMetadata], ModuleLikeWithYmlMetadataFile):
                 f"Could not find module {module_name}. Please make sure to add any module v2 requirements with"
                 " `inmanta module add --v2` and to install all the project's dependencies with `inmanta project install`."
             )
-
+        if isinstance(module, ModuleV1):
+            warnings.warn(
+                f"Loaded V1 module {module.name}. The use of V1 modules is deprecated. Use the equivalent V2 module instead.",
+                category=DeprecationWarning,
+            )
         self.modules[module_name] = module
         return module
 
--- a/tests/data/modules/v1_module_depends_on_v1_and_v2_module/module.yml
+++ b/tests/data/modules/v1_module_depends_on_v1_and_v2_module/module.yml
@@ -1,5 +1,5 @@
-name: v1_module_depends_on_v2_module
+name: v1_module_depends_on_v1_and_v2_module
 license: Apache 2.0
 version: 1.0
-description: A V1 module with a dependency on a V2 module
+description: A V1 module with a dependency on a V1 and a V2 module
 
--- a/tests/test_module_loader.py
+++ b/tests/test_module_loader.py
@@ -44,7 +44,7 @@ from inmanta.module import (
 )
 from inmanta.moduletool import ModuleConverter, ModuleTool, ProjectTool
 from packaging.version import Version
-from utils import PipIndex, create_python_package, module_from_template, v1_module_from_template
+from utils import PipIndex, create_python_package, log_contains, module_from_template, v1_module_from_template
 
 
 @pytest.mark.parametrize_any("editable_install", [True, False])
@@ -130,7 +130,10 @@ def test_v1_and_v2_module_installed_simultaneously(
     assert isinstance(Project.get().modules[module_name], ModuleV2)
 
 
-def test_v1_module_depends_on_v1_and_v2_module(tmpdir: py.path.local, snippetcompiler, capsys, modules_v2_dir: str) -> None:
+def test_v1_module_depends_on_v1_and_v2_module(
+    tmpdir: py.path.local, snippetcompiler, capsys, modules_v2_dir: str, caplog
+) -> None:
+    caplog.set_level(logging.WARNING)
     module_dir = os.path.join(modules_v2_dir, "elaboratev2module")
     module_copy_dir = os.path.join(tmpdir, "module")
     shutil.copytree(module_dir, module_copy_dir)
@@ -147,6 +150,15 @@ def test_v1_module_depends_on_v1_and_v2_module(tmpdir: py.path.local, snippetcom
     output = capsys.readouterr().out
     assert "Print from v2 module" in output
     assert "Hello world" in output
+
+    # Assert deprecation warnings regarding the use of V1 modules
+    for v1_module_name in [module_name, "elaboratev1module"]:
+        log_contains(
+            caplog,
+            "py.warnings",
+            logging.WARNING,
+            f"Loaded V1 module {v1_module_name}. The use of V1 modules is deprecated. Use the equivalent V2 module instead.",
+        )
 
 
 def test_install_module_no_v2_source(snippetcompiler) -> None:


FloLey (approved):

docs/adr/0001-Logging-warnings-using-the-Python-warnings-or-logging-module.md
```diff
@@ -0,0 +1,33 @@
...
+We don't want Deprecation warnings in third-party libraries to be visible to end-users.
+
+## Pros and Cons
+
+* LOGGER.warning():
+   * This logger just writes a message to the log.
+   * When the same message is logged twice, it will appear twice in the log.
+* warnings.warn():
+   * Each message is displayed only once.
+   * Pytest provides a nice overview of all warning logged during the executing of the test suite.
```
sanderr, line 22:
```suggestion
   * Pytest provides a nice overview of all warning logged during the executing of the test suite.
   * Warnings are highly configurable through e.g. `warnings.filterwarnings`.
```

docs/adr/0001-Logging-warnings-using-the-Python-warnings-or-logging-module.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Python has two different ways to report warnings to the user. Warnings can be logged via the logger (`LOGGER.warning()`) or using the Python warnings module (`warnings.warn()`). It was unclear to developers which type of logger has to used in which circumstances.
+
+We don't want Deprecation warnings in third-party libraries to be visible to end-users.
```
sanderr, line 13:
```suggestion
By default we don't want Deprecation warnings in third-party libraries to be visible to end-users.
```

docs/adr/0001-Logging-warnings-using-the-Python-warnings-or-logging-module.md
```diff
@@ -0,0 +1,33 @@
...
+   * Each message is displayed only once.
+   * Pytest provides a nice overview of all warning logged during the executing of the test suite.
+
+## Decision
+
+- The `warnings` module will be configured as such that all warnings not extending from the `warnings.InmantaWarning` class, are not displayed to the user. This ensures that warnings from third-party libraries are always ignored.
+- All warnings logged to `warning.warn()` should extend from the `warnings.InmantaWarning` class. The `inmanta.warnings.warn()` is a wrapper around `warnigs.warn()` that ensures that this constraint is satisfied. Using the latter method is preferred.
+
+- All warnings, for which we expect an action from the end-user, should be logged using the `inmanta.warnings.warn()` method. An example is a log message that announces a feature deprecation.
+- All other warnings should be logged using the `LOGGER.warning()` method.
```
sanderr, line 30: I'm not following with the motivation for this yet. Revived the slack discussion.

src/inmanta/config.py
```diff
@@ -301,9 +301,8 @@ def get(self) -> T:
             has_deprecated_option = cfg.has_option(self.predecessor_option.section, self.predecessor_option.name)
             has_new_option = cfg.has_option(self.section, self.name)
             if has_deprecated_option and not has_new_option:
-                warnings.warn(
-                    "Config option %s is deprecated. Use %s instead." % (self.predecessor_option.name, self.name),
-                    category=DeprecationWarning,
```
sanderr, line 306: There are quite some deprecation warnings. I wonder if it might be worth creating a `class InmantaDeprecationWarning(InmantaWarning, DeprecationWarning)`? If so we might want to have `CompilerDeprecationWarning` inherit from it. The inheritance graph would become a diamond but I think that should be fine.

arnaudsjs: Yes. I though about that. I didn't do it because I was afraid that this might derail into a full mirror on the Python warnings hierarchy. But I can still add it.

sanderr: It sounds like a valid concern. Exception hierarchy is something I generally feel very uncomfortable with, I find it very difficult to say which is best. The main motivation for making it a differentiable exception class would be if we ever want to allow more customization. Perhaps we should leave it as it is now until such a use case arises.

docs/adr/0001-Logging-warnings-using-the-Python-warnings-or-logging-module.md
arnaudsjs: Could you elaborate on what you would like to see changed here?

sanderr: I'm honestly not sure what exactly but it feels a bit ambiguous to me. Bart and I are discussing it on [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1666013643259899?thread_ts=1665730774.107619&cid=CKRF0C8R3), so perhaps let's centralize the discussion there.

docs/adr/0001-Logging-warnings-using-the-Python-warnings-or-logging-module.md
```diff
@@ -0,0 +1,34 @@
+# 1. Logging warnings using the python warnings or logging module
```
bartv, line 1: Why add a number to the title? The madr project explicitly discourages it because it make it more difficult to move it between repos.

src/inmanta/server/config.py
```diff
@@ -105,17 +106,15 @@ def get_bind_port() -> int:
     if Config.is_set("server", "bind-port") or Config.is_set("server", "bind-address"):
         # Use new bind-port option
         if Config.is_set("server_rest_transport", "port"):
-            warnings.warn(
+            inmanta.warnings.warn(
```
bartv, line 109: Is there a particular reason to drop the information that this is a deprecation? One of the benefits of warnings is that you have more information than just "WARNING". You can use [warning filter](https://docs.python.org/3/library/warnings.html#warning-filter) to control which ones you want to see.

With pytest for example you can ask explicitly: only show me deprecation warnings: https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html

bartv (changes requested):

src/inmanta/server/config.py
arnaudsjs: The reason is that we ignore all warnings by default: https://github.com/inmanta/inmanta-core/blob/0aba1b03937bad3900fc46673726e14e1124e81b/src/inmanta/warnings.py#L121 in order not to get DeprecationWarnings from third-party libraries. But I see we can also define a filter on module. @sanderr I see you implemented these warning filters in https://github.com/inmanta/inmanta-core/pull/1928. Do you remember why you didn't filter on module name?

docs/adr/0001-Logging-warnings-using-the-Python-warnings-or-logging-module.md
arnaudsjs: Hmm. I used the `adr-new` CLI tool from adr-tools-python to generate the template. This title was generate automatically. Do I create a ticket at adr-tools-python to creates title without this number.

bartv: https://github.com/adr/madr/blob/2.1.2/docs/adr/0002-do-not-use-numbers-in-headings.md

bartv: In the demo I used pyadr

arnaudsjs: Ok. I will use that tool then.

src/inmanta/server/config.py
sanderr: Honestly it might just have been an oversight. I remember back when I implemented this I was pretty unexperienced with warnings. Filtering on module name would make more sense in hindsight, and we could then probably even get rid of the `InmantaWarning` class.

bartv (approved):

src/inmanta/warnings.py
```diff
@@ -19,47 +19,44 @@
...
 
 class WarningOption:
     """
     An option to manage warnings. Consists of a name and a range of possible values, each tied to a warning rule.
     For example, applying
     WarningOption(
-        "disable-runtime-warnings",
-        {True: WarningRule(WarningBehaviour.IGNORE, CompilerRuntimeWarning)}
+        "disable-inmanta-warnings",
+        {True: WarningRule(WarningBehaviour.IGNORE, module=REGEX_INMANTA_MODULE)}
```
sanderr, line 57: We lose the ability here to filter on a specific warning. Couldn't that be useful still? i.e. allow filtering on both module and type?

sanderr (changes requested):

src/inmanta/warnings.py
arnaudsjs: I removed this, because we are not using this functionality and I don't like to have code in our code base that is not used. Apart from that I cannot think of a use cases where we want to have a different setting for one specific type of warnings. Do you have a specific use case in mind?

sanderr: One thing that comes to mind is raising an error for compiler deprecation warnings. Or ignoring a specific known issue. But for a moment I forgot that this can only be configured through the options we explicitly expose, so you're right that we're not using it at the moment. I withdraw my comment, we can always add it again if/when the need arises.

sanderr (approved):

PR #4993, opened 2022-10-17
issue/4968-soft-hard-retry

### LINKED ISSUE #4968: Add soft and hard timeout on the retry_limited method
# Description
Add hard timeout multiplier for `retry_limited`, configurable through env var and/or pytest option. The idea would be to be able to set the hard limit a lot higher than the soft limit. The algorithm would be as follows:

* Stop waiting when the wait condition is met or when the hard timeout is reached
* Raise an exception at the end of the `retry_limited` method when it took longer than soft_limit to reach the wait condition.

The `retry_limited` method can then show error messages like:

* Timeout: Wait condition was met after X seconds, but soft limit was set to Y seconds
* Timeout: Wait condition was not reach after hard limit of X seconds

## Original description

In general we always use a timeout of 10 seconds on the `retry_limited` method. In some circumstances this timeout is too sharp. When a test case fails with a timeout, it's hard to find out whether the test case would hang indefinitely or whether the condition would have been reached when the limit was set just a little higher. To improve the error reporting of the `retry_limited` method, I would like to propose the idea to add a soft and a hard timeout on that method. The idea would be to set the hard limit a lot higher than the soft limit. The algorithm would be as follows:

* Stop waiting when the wait condition is met or when the hard timeout is reached
* Raise an exception at the end of the `retry_limited` method when it took longer than soft_limit to reach the wait condition.

The `retry_limited` method can then show error messages like:

* Timeout: Wait condition was met after X seconds, but soft limit was set to Y seconds
* Timeout: Wait condition was not reach after hard limit of X seconds

FloLey:
# Description

Add hard timeout multiplier for retry_limited, configurable through env var. The idea would is to be able to set the hard limit a lot higher than the soft limit. 

- Stop waiting when the wait condition is met or when the hard timeout is reached (raise exception)
- Raise an exception at the end of the retry_limited method when it took longer than soft_limit to reach the wait condition.

closes #4968 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -54,12 +54,6 @@ T = TypeVar("T")
 S = TypeVar("S")
 
 
-class Timeout(Exception):
-    """The operation exceeded the given timeout."""
-
-    pass
-
-
 def get_compiler_version() -> str:
     return COMPILER_VERSION
 
@@ -509,14 +503,16 @@ async def retry_limited(
             return fun(*args, **kwargs)
 
     multiplier: int = int(os.environ.get("INMANTA_RETRY_LIMITED_MULTIPLIER", 1))
+    if multiplier < 1:
+        raise ValueError("value of INMANTA_RETRY_LIMITED_MULTIPLIER must be bigger or equal to 1.")
     hard_timeout = timeout * multiplier
     start = time.time()
     while time.time() - start < hard_timeout and not (await fun_wrapper()):
         await asyncio.sleep(interval)
     if not (await fun_wrapper()):
-        raise Timeout(f"Wait condition was not reach after hard limit of {hard_timeout} seconds")
+        raise asyncio.TimeoutError(f"Wait condition was not reach after hard limit of {hard_timeout} seconds")
     if time.time() - start > timeout:
-        raise Timeout(
+        raise asyncio.TimeoutError(
             f"Wait condition was met after {time.time() - start} seconds, but soft limit was set to {timeout} seconds"
         )
 
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -510,7 +510,7 @@ async def retry_limited(
     while time.time() - start < hard_timeout and not (await fun_wrapper()):
         await asyncio.sleep(interval)
     if not (await fun_wrapper()):
-        raise asyncio.TimeoutError(f"Wait condition was not reach after hard limit of {hard_timeout} seconds")
+        raise asyncio.TimeoutError(f"Wait condition was not reached after hard limit of {hard_timeout} seconds")
     if time.time() - start > timeout:
         raise asyncio.TimeoutError(
             f"Wait condition was met after {time.time() - start} seconds, but soft limit was set to {timeout} seconds"
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -498,6 +498,7 @@ async def retry_limited(
             return fun(*args, **kwargs)
 
     multiplier: int = os.environ.get("INMANTA_RETRY_LIMITED_MULTIPLIER", 1)
+    print(multiplier)
     hard_timeout = timeout * multiplier
     start = time.time()
     while time.time() - start < hard_timeout and not (await fun_wrapper()):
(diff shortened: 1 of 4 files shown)


FloLey:
Once merged I'll add the env var in the jenkin files. Is a multiplier of 10 acceptable?

src/inmanta/util.py
```diff
@@ -487,17 +493,32 @@ async def retry_limited(
...
+    the same.
+    """
+
     async def fun_wrapper() -> bool:
         if inspect.iscoroutinefunction(fun):
             return await fun(*args, **kwargs)
         else:
             return fun(*args, **kwargs)
 
+    multiplier: int = int(os.environ.get("INMANTA_RETRY_LIMITED_MULTIPLIER", 1))
```
arnaudsjs, line 511: I would also add a check to verify this value is `>=1` to prevent that someone uses this method incorrectly.

src/inmanta/util.py
```diff
@@ -487,17 +493,32 @@ async def retry_limited(
...
-    while time.time() - start < timeout and not (await fun_wrapper()):
+    while time.time() - start < hard_timeout and not (await fun_wrapper()):
         await asyncio.sleep(interval)
     if not (await fun_wrapper()):
-        raise asyncio.TimeoutError()
+        raise Timeout(f"Wait condition was not reach after hard limit of {hard_timeout} seconds")
+    if time.time() - start > timeout:
+        raise Timeout(
+            f"Wait condition was met after {time.time() - start} seconds, but soft limit was set to {timeout} seconds"
+        )
```
arnaudsjs, line 521: This method is used pretty much everywhere and some test cases might rely on the fact that an `asyncio.TimeoutError` is raised. Changing the type of the exception might break tests all over the place.

arnaudsjs (changes requested):

src/inmanta/util.py
FloLey: You mean outside of core? I'll set the old exception back

arnaudsjs: Yes. This method is packaged into `pytest-inmanta-extensions`. As such, it can be used by the test suite of any extension or module.

src/inmanta/util.py
```diff
@@ -487,17 +487,34 @@ async def retry_limited(
     *args: object,
     **kwargs: object,
 ) -> None:
+    """
+    This function makes use of the INMANTA_RETRY_LIMITED_MULTIPLIER env variable. If set, INMANTA_RETRY_LIMITED_MULTIPLIER
+    serves as multiplier: The timeout given as argument becomes a 'soft limit' and the "soft limit' mutliplied by the
```
arnaudsjs, line 492:
```suggestion
    serves as multiplier: The timeout given as argument becomes a 'soft limit' and the "soft limit' multiplier by the
```

src/inmanta/util.py
```diff
@@ -487,17 +487,34 @@ async def retry_limited(
...
+    multiplier (form the env var) becomes a "hard limit". If the hard limit is reached before the wait condition is fulfilled
```
arnaudsjs, line 493:
```suggestion
    multiplier (from the env var) becomes a "hard limit". If the hard limit is reached before the wait condition is fulfilled
```

arnaudsjs:
> Once merged I'll add the env var in the jenkin files. Is a multiplier of 10 acceptable?

I think so.

src/inmanta/util.py
FloLey: I mean multiplied :p

src/inmanta/util.py
```diff
@@ -487,17 +487,34 @@ async def retry_limited(
...
+    if multiplier < 1:
+        raise ValueError("value of INMANTA_RETRY_LIMITED_MULTIPLIER must be bigger or equal to 1.")
+    hard_timeout = timeout * multiplier
     start = time.time()
-    while time.time() - start < timeout and not (await fun_wrapper()):
+    while time.time() - start < hard_timeout and not (await fun_wrapper()):
         await asyncio.sleep(interval)
     if not (await fun_wrapper()):
-        raise asyncio.TimeoutError()
+        raise asyncio.TimeoutError(f"Wait condition was not reach after hard limit of {hard_timeout} seconds")
```
arnaudsjs, line 513:
```suggestion
        raise asyncio.TimeoutError(f"Wait condition was not reached after hard limit of {hard_timeout} seconds")
```

arnaudsjs (changes requested):

arnaudsjs (approved):

wouterdb (approved):

PR #5031, opened 2022-10-25
FIX: optional field without default value

FloLey:
# Description

Optional fields which were not assigned None as default were causing issues:
when populating a DAO using `__process_kwargs`, if the value of an optional field was 
`None` in the DB, the DAO would not get this field populated. This caused issues when trying
to access the field as it did not exist.
This PR fixes this issue. 

While resolving this, an inconsistency between the DB and the objects caused some tests to fail:
the `repository` and `branch` fields of an environment that equals None in objects are stored as empty strings in the DB

closes https://github.com/inmanta/inmanta-lsm/issues/1078

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/fix-optional-field-no-default.yml
+++ b/changelogs/unreleased/fix-optional-field-no-default.yml
@@ -1,0 +1,4 @@
+---
+description: Fixes an issue about optional values without default values
+change-type: patch
+destination-branches: [master, iso5]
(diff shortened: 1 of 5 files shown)


wouterdb (approved):

changelogs/unreleased/fix-optional-field-no-default.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: Fixes an issue about optional fields without default value
+change-type: patch
+destination-branches: [master, iso5]
```
arnaudsjs, line 4: This issue is present in the release, so we need to add a bugfix section for it.

arnaudsjs (changes requested):

arnaudsjs (approved):

Issue #4780, opened 2022-09-01
Enable timestamp on the log messages produced by the test cases


arnaudsjs:
In order to debug a test case that times out on a wait condition, it's handy to see the timestamps in the log messages. That way it easier to see in which stage a lot of time is being lost.

Adding the timestamps can be achieved by providing the following options to pytest: `--log-format="%(asctime)s.%(msecs)03d %(levelname)s %(message)s"`.

If `pytest.ini` supports it, add it there.

bartv:
jenkins gives you timestamps?

PR #5034, opened 2022-10-25
issue/4780-add-timestamp-to-logs

FloLey:
# Description

Enable timestamp on the log messages produced by the test cases

closes https://github.com/inmanta/inmanta-core/issues/4780

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ deps=
 extras=
     dataflow_graphic
 # Set the environment variable INMANTA_EXTRA_PYTEST_ARGS='--fast' to run in fast mode
-commands=py.test --log-level DEBUG --cov=inmanta --cov-report term --cov-report xml --junitxml=junit-{envname}.xml -vvv {env:INMANTA_EXTRA_PYTEST_ARGS:} --durations=50 tests/server
+commands=py.test --log-level DEBUG --cov=inmanta --cov-report term --cov-report xml --junitxml=junit-{envname}.xml -vvv {env:INMANTA_EXTRA_PYTEST_ARGS:} --durations=50 tests/server/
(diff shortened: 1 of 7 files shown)


tox.ini
```diff
@@ -11,7 +11,7 @@ deps=
 extras=
     dataflow_graphic
 # Set the environment variable INMANTA_EXTRA_PYTEST_ARGS='--fast' to run in fast mode
-commands=py.test --log-level DEBUG --cov=inmanta --cov-report term --cov-report xml --junitxml=junit-{envname}.xml -vvv {env:INMANTA_EXTRA_PYTEST_ARGS:} --durations=50 tests/
+commands=py.test --log-level DEBUG --log-format="%(asctime)s.%(msecs)03d %(levelname)s %(message)s" --cov=inmanta --cov-report term --cov-report xml --junitxml=junit-{envname}.xml -vvv {env:INMANTA_EXTRA_PYTEST_ARGS:} --durations=50 tests/
```
arnaudsjs, line 14: The ticket suggests to add this setting to the `pytest.ini` file to enable this settings for all pytest runs, because this way the setting is only enabled when running the tests using tox. It see it's also possible to configure pytest settings using the `setup.cfg`. I think that's even better, because that way we don't need to introduce yet another config file. @sanderr Do you agree. I see you added the suggestion for the `pytest.ini` file.

arnaudsjs (changes requested):

tox.ini
sanderr: Agreed. And thanks for showing me you can inspect edit history, that's very useful.

arnaudsjs (approved):

PR #5042, opened 2022-10-26
Add support to only load environment settings from all extensions

arnaudsjs:
# Description

I have been struggling at bit to find an elegant solution for this problem here. With this draft PR I would like to get feedback on the approach being used before I polish it. The general idea is that extensions define their custom environment settings via the `register_environment_settings` in the `extensions.py` method of the extension (See: https://github.com/inmanta/inmanta-lsm/pull/1084/files for an example). Inmanta-sphinx can then use the bootloader to discover all environment settings defined by the different extensions (see: https://github.com/inmanta/inmanta-sphinx/pull/57/files).

The implementation in inmanta-core is written as such that we have the possibility to only discover environment settings via the bootloader. This is required because executing the `setup()` of the license extensions results in an Exception when no valid license is available.

I also tried the approach where settings are exposed using a python entrypoint. That also didn't work, because it causes issues with circular dependencies. The following flow happens in that case:

1. `inmanta.data` discovers the entrypoints and invokes the method in the extension to obtain the environment settings.
2. Returning the settings again requires an import of `inmanta.data` which triggers the circular import

closes inmanta/irt#1366

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -383,3 +383,12 @@ sort
 
 .. automodule:: inmanta.protocol.methods_v2
     :members:
+
+Server
+------
+
+.. autoclass:: inmanta.server.extensions.ApplicationContext
+    :show-inheritance:
+
+.. autoclass:: inmanta.server.bootloader.InmantaBootloader
+    :show-inheritance:
--- a/src/inmanta/server/bootloader.py
+++ b/src/inmanta/server/bootloader.py
@@ -21,12 +21,13 @@ import logging
 import pkgutil
 from pkgutil import ModuleInfo
 from types import ModuleType
-from typing import Callable, Dict, Generator, List, Optional
+from typing import Dict, Generator, List, Optional
 
 from inmanta.const import EXTENSION_MODULE, EXTENSION_NAMESPACE
 from inmanta.server import config
 from inmanta.server.extensions import ApplicationContext, FeatureManager, InvalidSliceNameException
 from inmanta.server.protocol import Server, ServerSlice
+from inmanta.stable_api import stable_api
 
 LOGGER = logging.getLogger(__name__)
 
@@ -61,6 +62,7 @@ class ConstrainedApplicationContext(ApplicationContext):
         self.parent.set_feature_manager(feature_manager)
 
 
+@stable_api
 class InmantaBootloader(object):
     """The inmanta bootloader is responsible for:
     - discovering extensions
@@ -125,7 +127,7 @@ class InmantaBootloader(object):
         """Discover all packages that are defined in the inmanta_ext namespace package. Filter available extensions based on
         enabled_extensions and disabled_extensions config in the server configuration.
 
-        :param return_all_available_packages: Return all available plugin packages and independent of whether the extension is
+        :param return_all_available_packages: Return all available plugin packages independent of whether the extension is
                                               enabled or not.
         :return: A list of all subpackages defined in inmanta_ext
         """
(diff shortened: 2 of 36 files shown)


wouterdb (approved):
looks fine to me. 

Does this break the license package?

arnaudsjs:
> looks fine to me.
> 
> Does this break the license package?

No. It doesn't.

src/inmanta/server/bootloader.py
```diff
@@ -164,38 +168,49 @@ def _load_extension(self, name: str) -> Callable[[ApplicationContext], None]:
...
 
-    # Extension loading Phase II: collect slices
-    def _collect_slices(self, extensions: Dict[str, Callable[[ApplicationContext], None]]) -> ApplicationContext:
+    def _collect_slices(self, ext_name: str, ext_module: ModuleType, app_ctx: ApplicationContext) -> None:
+        try:
+            ext_module.setup(app_ctx)
+        except Exception:
+            LOGGER.warning("Could not load extension %s", ext_name, exc_info=True)
+
+    def _collect_environment_settings(self, ext_name: str, ext_module: ModuleType, app_ctx: ApplicationContext) -> None:
```
sanderr, line 194: Could you add a docstring here? "environment settings" is ambiguous, and the other meaning (shell environment variables) is even the more common interpretation.

src/inmanta/server/bootloader.py
```diff
@@ -164,38 +168,49 @@ def _load_extension(self, name: str) -> Callable[[ApplicationContext], None]:
...
+        try:
+            ext_module.register_environment_settings(app_ctx)
```
sanderr, line 196: Perhaps we should not call it in the `try` block, only try the access, or use `hasattr`. If we call it already we can't differentiate between an `AttributeError` within the method (e.g. it trying to access a method on `app_ctx` that does not exist) or the fact that the method is not defined.

src/inmanta/server/bootloader.py
```diff
@@ -164,38 +168,49 @@ def _load_extension(self, name: str) -> Callable[[ApplicationContext], None]:
...
+        except AttributeError:
+            # Extension doesn't defined any environment settings.
+            pass
+        except Exception:
+            LOGGER.warning("Could not load extension %s", ext_name, exc_info=True)
+
+    def load_slices(
+        self, load_all_extensions: bool = False, only_register_environment_settings: bool = False
```
sanderr, line 204: Since there is no logical ordering here, enforce readability by making them kwarg-only?
```suggestion
        self, *, load_all_extensions: bool = False, only_register_environment_settings: bool = False
```

src/inmanta/server/extensions.py
```diff
@@ -206,3 +207,9 @@ def get_product_metadata(self) -> ProductMetadata:
 
     def get_extension_statuses(self) -> List[ExtensionStatus]:
         return ServerSlice.get_extension_statuses(list(self._slices))
+
+    def register_environment_settings(self, settting: data.Setting) -> None:
+        data.Environment.register_setting(settting)
+
+    def get_environment_settings(self) -> List[data.Setting]:
```
sanderr, line 214:
```suggestion
    def get_environment_settings(self) -> list[data.Setting]:
```

src/inmanta/server/extensions.py
```diff
@@ -206,3 +207,9 @@ def get_product_metadata(self) -> ProductMetadata:
 
     def get_extension_statuses(self) -> List[ExtensionStatus]:
         return ServerSlice.get_extension_statuses(list(self._slices))
+
+    def register_environment_settings(self, settting: data.Setting) -> None:
```
sanderr, line 211:
```suggestion
    def register_environment_setting(self, settting: data.Setting) -> None:
```

sanderr (commented):
The general idea looks good to me. I do think we may want to deprecate `EnvironmentService.regsiter_setting` rather than just to drop it (at least for iso5). Otherwise we break compatibility wit older extensions.

docs/reference/api.rst
```diff
@@ -383,3 +383,12 @@ sort
...
     :members:
+
+Server
+------
+
+.. autoclass:: inmanta.server.extensions.ApplicationContext
+    :show-inheritance:
+
+.. autoclass:: inmanta.server.bootloader.InmantaBootloader
+    :show-inheritance:
```
sanderr, line 394: Could you add these to `tests/test_import_entrypoints.py`?

sanderr (changes requested):

sanderr (approved):

PR #5045, opened 2022-10-27
Issue/4928 add compiler finalizers

### LINKED ISSUE #4928: Add compiler finalizers
# Problem

When writing allocators, one of the most involved aspects is making sure all resources are properly flushed back and all connections are properly closed. 

It would greatly help allocator implementation if the compiler had a finalizer mechanism. 

# Expected end result

We want to be able to register finalizers both as decorator and as callback

```python

connection = None

def get_connection():
   global connection
   if connection is None:
        connection = ...
   return connection

@finalizer
def finalize_connection():
   if connection:
      connection.close()
```
with registration
```python

connection = None
def get_connection():
  global connection 
  if not connection:
       connection = ...
       finalizer(connection.close)
   return connection
```

# Solution

1. add a @finalizer decorator (similar to @plugin) that indicates a method as a finalizer
2. when the compiler finishes, prior to export, call the finalizers (even when the compiler produced an exception),
3. make sure that if the finalizer fails, this is reported: if there is already an expection from the compiler, raise that one and log the one from the finalizer. If there was no excpetion, raise the one from the finalizer instead.
4. add documentation with an example

# Rationale

1. registration Vs decoration: we support both, as the underlying code is in fact the same
2. finalize after compile or after export: we want to encourage proper closing as soon as possible, so before export. We can, later on, add a when argument to the finalizer to have multiple finalization points.
3. finalizer exceptions don't replace the compiler exception: bit risky, we may also use the multiexception to add them. I think this approach will be least likely to break e.g. explainers. 
4. don't report to the finalizer if the compile was a success: allows the use of no-args close methods directly. We can add arguments to propagate compiler state later on by adding arguments to the annotation.

FloLey:
# Description

add the finalizer decorator to register finalizer functions which are executed at the end of the compilations

closes #4928 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


src/inmanta/compiler/__init__.py
```diff
@@ -18,7 +18,7 @@
 import logging
 import sys
 from itertools import chain
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Set, Tuple
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Sequence, Set, Tuple
```
sanderr, line 21: We should use `collections.abc.Callable` for new code.
```suggestion
from collections import abc
from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Set, Tuple
```

src/inmanta/compiler/__init__.py
```diff
@@ -314,3 +317,35 @@ def add_trace(exception: CompilerException) -> bool:
...
+            for fnc in cls.__finalizers:
+                fnc()
+        except Exception as e:
+            if should_log:
+                LOGGER.error(f"Finalizers failed: {e}")
+            else:
+                raise
+
+
+@stable_api
```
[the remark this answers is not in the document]
sanderr: Whenever we expose anything outside of core. You'll also need to add a test case to `tests/test_import_entry_points.py`

src/inmanta/compiler/__init__.py
```diff
@@ -68,14 +69,17 @@ def do_compile(refs: Dict[Any, Any] = {}) -> Tuple[Dict[str, inmanta_type.Type],
     except ParserException as e:
         compiler.handle_exception(e)
     sched = scheduler.Scheduler(compiler_config.track_dataflow(), project.get_relation_precedence_policy())
+    raised_complile_excpetion: bool = False
```
sanderr, line 72:
```suggestion
    raised_compile_exception: bool = False
```

sanderr (changes requested):
I think it could be useful to add another test case that does an export with a finalizer. The ticket specifies that "when the compiler finishes, prior to export, call the finalizers"

sanderr:
Oops, didn't mean to submit yet. More thorough review following.

src/inmanta/compiler/__init__.py
```diff
@@ -314,3 +318,44 @@ def add_trace(exception: CompilerException) -> bool:
...
+        if 'should_log' is set to True the exceptions will not be raised but logged instead.
+        """
+        try:
+            for fnc in cls.__finalizers:
+                fnc()
+        except Exception as e:
+            if should_log:
+                LOGGER.error(f"Finalizers failed: {e}")
+            else:
+                raise
```
sanderr, line 348: Just a suggestion for brevity:
```suggestion
        except (Exception if should_log else ()) as e:
            LOGGER.error(f"Finalizers failed: {e}")
```

sanderr (changes requested):

src/inmanta/compiler/__init__.py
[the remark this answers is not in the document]
FloLey: I could create a list of errors and raise/log them all at the end? Do you see an issue with that?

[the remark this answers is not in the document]
sanderr: I think that would be fine.
There is already a multi exception in `inmanta.ast` that may or may not be useful here (I don't recall if it's specific for compile-time exceptions).

wouterdb (approved):

src/inmanta/compiler/__init__.py
```diff
@@ -337,22 +338,28 @@ def call_finalizers(cls, should_log: bool = False) -> None:
...
-        except Exception as e:
+            except Exception as e:
+                excns.append(CompilerException("Finalizer failed: " + str(e)))
+        if excns:
             if should_log:
-                LOGGER.error(f"Finalizers failed: {e}")
+                msg: list[str] = []
+                for exception in excns:
+                    msg.append(exception.msg + "\n")
+                LOGGER.error("".join(msg).strip())
```
sanderr, line 352: Since these are separate failures I have a slight preference for making it separate log statements. Not a strict change requirement.

src/inmanta/compiler/__init__.py
```diff
@@ -337,22 +338,28 @@ def call_finalizers(cls, should_log: bool = False) -> None:
...
             else:
-                raise
+                raise MultiException(excns)
```
sanderr, line 354: I think the `MultiException` is really only meant for 2+ execptions so I'm afraid we'll need another branch here:
```suggestion
                raise excns[0] if len(excns) == 1 else MultiException(excns)
```

sanderr (changes requested):

src/inmanta/compiler/__init__.py
FloLey: shouldn't we change the MultiException than? I mean we likely need to add this condition every time we want to raise a MultiExcpetion

sanderr: I don't think it would really break anything, but its definition is that it's a wrapper around a multitude of exceptions. But I'm not against adding a guard. @wouterdb wdyt?

wouterdb: I don't understand what is exactly suggested here?

sanderr: `inmanta.ast.MultiException`, should it ever be used with a set of 1 child exceptions? If not, do we add a guard to its constructor that prevents it.

sanderr (approved):

PR #5046, opened 2022-10-27
[DEPR] drop support for 'dashboard' and 'web-console' config option s…

Hugo-Inmanta:
…ections in favor of a new 'web-ui' section

# Description

part of [#317](https://github.com/inmanta/inmanta-ui/issues/317)
# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/inmanta/server/config.py
+++ b/src/inmanta/server/config.py
@@ -268,9 +268,9 @@ server_access_control_allow_origin = Option(
 # Dashboard
 #############################
 
-dash_realm = Option("dashboard", "realm", "inmanta", "The realm to use for keycloak authentication.", is_str)
-dash_auth_url = Option("dashboard", "auth_url", None, "The auth url of the keycloak server to use.", is_str)
-dash_client_id = Option("dashboard", "client_id", None, "The client id configured in keycloak for this application.", is_str)
+dash_realm = Option("web-ui", "realm", "inmanta", "The realm to use for keycloak authentication.", is_str)
+dash_auth_url = Option("web-ui", "auth_url", None, "The auth url of the keycloak server to use.", is_str)
+dash_client_id = Option("web-ui", "client_id", None, "The client id configured in keycloak for this application.", is_str)
 
 
 def default_hangtime() -> str:
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -107,7 +107,7 @@ port=1234
 host=host1
 interval=10
 tags=tag1=value1
-[dashboard]
+[web-ui]
 path=/some/directory
 client-id=test
         """
@@ -150,7 +150,7 @@ port=9999
 host=host3
 [influxdb]
 tags=tag2=value2
-[dashboard]
+[web-ui]
 path=/some/other/directory
         """
         )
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/317-add-web-ui-section-to-the-configuration-file-iso6.yml
```diff
@@ -0,0 +1,9 @@
+---
+description: The 'dashboard' and 'web-console' sections are no longer supported for configuration options.
+  The 'web-ui' section should now be used instead for configuration options relative to web interfaces.
```
arnaudsjs, line 3:
```suggestion
description: The 'dashboard' and 'web-console' sections are no longer supported for configuration options. The 'web-ui' section should now be used instead for configuration options related to web interfaces.
```

src/inmanta/server/config.py
```diff
@@ -268,9 +268,9 @@ def validate_fact_renew(value: object) -> int:
 # Dashboard
 #############################
 
-dash_realm = Option("dashboard", "realm", "inmanta", "The realm to use for keycloak authentication.", is_str)
-dash_auth_url = Option("dashboard", "auth_url", None, "The auth url of the keycloak server to use.", is_str)
-dash_client_id = Option("dashboard", "client_id", None, "The client id configured in keycloak for this application.", is_str)
+dash_realm = Option("web-ui", "realm", "inmanta", "The realm to use for keycloak authentication.", is_str)
+dash_auth_url = Option("web-ui", "auth_url", None, "The auth url of the keycloak server to use.", is_str)
+dash_client_id = Option("web-ui", "client_id", None, "The client id configured in keycloak for this application.", is_str)
```
arnaudsjs, line 273: I think we want to define all `web-ui` configuration options in the `inmanta-ui` extension as this extension primarily exist for web-interface related things.

arnaudsjs (changes requested):

arnaudsjs (changes requested):

arnaudsjs (changes requested):
We should also remove the `web-console` section from `misc/inmanta.cfg`

arnaudsjs (changes requested):
We also need to update the docs. For example, the docs related to configuring authentication.

tests/test_config.py
```diff
@@ -148,27 +147,26 @@ def test_configfile_hierarchy(monkeypatch, tmpdir):
...
         """
         )
 
     with open(dot_inmanta_cfg_file, "w", encoding="utf-8") as f:
         f.write(
             """
-[dashboard]
-path=/directory
-client-id=test123
+[database]
```
arnaudsjs, line 159: This test case is updated to remove the `dashboard.path` and the `dashboard.client-id` config options. That's good. But these options are not 1-to-1 replaced with another option. That way the test doesn't test what is used to test. Why was this done?

arnaudsjs (changes requested):

tests/test_config.py
Hugo-Inmanta: Ah indeed, the mapping was wrong

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #5050, opened 2022-10-27
fix-issue-InmantaWarnings

FloLey:
# Description

fix issue about InmantaWarnings that does not exist anymore.

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/fix-inmanta-warnings.yml
+++ b/changelogs/unreleased/fix-inmanta-warnings.yml
@@ -1,0 +1,4 @@
+---
+description: Fixes an issue about InmantaWarnings that does not exist anymore.
+change-type: patch
+destination-branches: [master]
--- a/tests/compiler/test_basics.py
+++ b/tests/compiler/test_basics.py
@@ -474,19 +474,22 @@ def get_one() -> "int":
     )
     with warnings.catch_warnings(record=True) as w:
         compiler.do_compile()
-        # first warning is about the use of V1 modules that are deprecated.
-        assert len(w) == 2 if decorator else len(w) == 1
-        if len(w) == 2:
-            warning = w[1]
-            assert issubclass(warning.category, PluginDeprecationWarning)
-            if replaced_by:
-                replaced_by_name = replaced_by.replace('"', "")
-                assert (
-                    f"Plugin 'get_one' in module 'inmanta_plugins.test_module' is deprecated. It should be "
-                    f"replaced by '{replaced_by_name}'" in str(warning.message)
-                )
-            else:
-                assert "Plugin 'get_one' in module 'inmanta_plugins.test_module' is deprecated." in str(warning.message)
+        if decorator:
+            has_warning: bool = False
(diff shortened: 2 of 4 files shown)


changelogs/unreleased/fix-inmanta-warnings.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: Fixes an issue about InmantaWarnings that does not exist anymore.
+change-type: patch
+destination-branches: [master]
```
FloLey, line 4: the branch for iso5 hasn't been merged yet (it failed on the same issue) so I've directly fixed it in this PR: https://github.com/inmanta/inmanta-core/pull/5049

tests/compiler/test_basics.py
```diff
@@ -474,9 +474,10 @@ def get_one() -> "int":
     )
     with warnings.catch_warnings(record=True) as w:
         compiler.do_compile()
-        assert len(w) == 1 if decorator else len(w) == 0
-        if len(w):
-            warning = w[0]
+        # first warning is about the use of V1 modules that are deprecated.
+        assert len(w) == 2 if decorator else len(w) == 1
+        if len(w) == 2:
+            warning = w[1]
```
arnaudsjs, line 480: These assertions are too strict. If at some point we introduce an additional warning somewhere, this test case will fail because `len(w) == 3` instead of `len(w) == 2`.

tests/compiler/test_basics.py
```diff
@@ -636,7 +637,8 @@ def get_one() -> "int":
     )
     with warnings.catch_warnings(record=True) as w:
         compiler.do_compile()
-        assert len(w) == 1
-        warning = w[0]
+        assert len(w)
+        # first warning is about the use of V1 modules that are deprecated.
+        warning = w[1]
```
arnaudsjs, line 642: Same here. The warning we expect should not necessarily be present at index 1. It could be present at any other position.

arnaudsjs (changes requested):

tests/compiler/test_basics.py
```diff
@@ -474,18 +474,22 @@ def get_one() -> "int":
...
-                )
-            else:
-                assert "Plugin 'get_one' in module 'inmanta_plugins.test_module' is deprecated." in str(warning.message)
+        if decorator:
+            has_warning: bool = False
+            for warning in w:
+                if issubclass(warning.category, PluginDeprecationWarning):
+                    has_warning = True
+                    if replaced_by:
+                        replaced_by_name = replaced_by.replace('"', "")
```
arnaudsjs, line 483: Is there a specific reason why the `replaced_by` contains double quotes?

arnaudsjs (approved):

tests/compiler/test_basics.py
FloLey: I think this is a residue of multiple changes, I removed them :p

PR #5053, opened 2022-10-28
Resolve issue where the status endpoint can block

wouterdb:
# Description

This PR prevents the status endpoint form hanging when the DB is in trouble.

I produces a fully redundant, local cache of the compiler queue size.
I'm in general not a fan of redundant data like this, but the only alternative I see is to bake in a timeout in the status endpoint so it only reports data is can collect .e.g `<10ms` 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -3609,7 +3609,7 @@ class Compile(BaseDocument):
     @classmethod
     async def get_next_compiles_count(cls) -> int:
         """Get the number of compiles in the queue for ALL environments"""
-        result = await cls._fetch_int(f"SELECT count(*) FROM {cls.table_name()} WHERE NOT handled and completed IS NOT NULL")
+        result = await cls._fetch_int(f"SELECT count(*) FROM {cls.table_name()} WHERE NOT handled AND completed IS NULL")
         return result
 
     @classmethod
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -507,9 +507,11 @@ async def retry_limited(
         raise ValueError("value of INMANTA_RETRY_LIMITED_MULTIPLIER must be bigger or equal to 1.")
     hard_timeout = timeout * multiplier
     start = time.time()
-    while time.time() - start < hard_timeout and not (await fun_wrapper()):
+    result = await fun_wrapper()
+    while time.time() - start < hard_timeout and not result:
         await asyncio.sleep(interval)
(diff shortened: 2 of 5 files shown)


src/inmanta/util.py
```diff
@@ -507,9 +507,11 @@ async def fun_wrapper() -> bool:
         raise ValueError("value of INMANTA_RETRY_LIMITED_MULTIPLIER must be bigger or equal to 1.")
     hard_timeout = timeout * multiplier
     start = time.time()
-    while time.time() - start < hard_timeout and not (await fun_wrapper()):
+    result = await fun_wrapper()
```
wouterdb, line 510: if the condition is unstable, re-evaluation may fail

tests/server/test_compilerservice.py
```diff
@@ -316,9 +336,19 @@ async def isdone():
...
+    await compiler_cache_consistent(5)
+
     # run through env1, entire sequence
     for i in range(4):
         await check_compile_in_sequence(env1, e1, i)
+        if i < 2:
+            # First one is never queued, so not counted
+            # Last iteration here doesn't de-queue an item, but allows it to complete
+            # So don't handle last 2
+            compiler_cache_consistent(4 - i)
```
sanderr, line 348:
```suggestion
            await compiler_cache_consistent(4 - i)
```

src/inmanta/data/__init__.py
```diff
@@ -3609,7 +3609,7 @@ async def get_next_compiles_for_environment(cls, environment_id: uuid.UUID) -> "
     @classmethod
     async def get_next_compiles_count(cls) -> int:
         """Get the number of compiles in the queue for ALL environments"""
-        result = await cls._fetch_int(f"SELECT count(*) FROM {cls.table_name()} WHERE NOT handled and completed IS NOT NULL")
+        result = await cls._fetch_int(f"SELECT count(*) FROM {cls.table_name()} WHERE NOT handled AND completed IS NULL")
```
sanderr, line 3612: This looks like a pretty big bug. How comes this didn't bite use before?

sanderr (changes requested):

src/inmanta/data/__init__.py
wouterdb: only used in the status page,....

sanderr (approved):

wouterdb:
moved to #5068

PR #5069, opened 2022-11-02
issue/5066-fix-format-warnings

### LINKED ISSUE #5066: investigate format of warning messages
Currently the format of warnings is not nice: investigate why and try to make them nice again

FloLey:
# Description

 https://github.com/inmanta/inmanta-core/pull/4979 removed InmantaWarnings which use another formatter than the default warnings. As we want to keep both formats, this PR adds the InmantaWarnings back.

closes  #5066 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


wouterdb (approved):

src/inmanta/warnings.py
```diff
@@ -140,15 +151,19 @@ def _showwarning(
...
-        text: str = warnings.formatwarning(
-            # ignore type check because warnings.formatwarning accepts Warning instance but it's type definition doesn't
-            message,  # type: ignore
-            category,
-            filename,
-            lineno,
-            line,
-        )
-        logger: logging.Logger = logging.getLogger("py.warnings")
+        if issubclass(category, InmantaWarning):
```
sanderr, line 154: I don't think we wanted to bring the class back, just the format. We could just change this condition to check where the warning originates from.

sanderr (changes requested):

src/inmanta/warnings.py
FloLey: what do you mean by 'originates from'? do you mean which class/file raised the warning? how can this be done?

sanderr: Mostly based on `filename` I think. But I need to give this some more thought, perhaps you're right after all that bringing the class back is the cleaner solution.

sanderr: Or we only do the custom formatting when `category.__module__` is part of the inmanta namespace?

sanderr: All in all, I think I agree with the proposal to bring the class back, but we would need to fully revert the previous PR then, also bringing back all inheritance from it. But since @arnaudsjs delved a bit into this when he made the change (and this is not urgent I think) I would ask him if he has an opinion on Monday.

arnaudsjs: It would be nice if we can prevent re-adding the `InmantaWarnings` class, to prevent that we build our own warnings hierarchy that is a duplicate of the default Python warnings hierarchy. Something I don't understand here is: which problem we are trying to resolve? The ticket mentions: `Currently the format of warnings is not nice: investigate why and try to make them nice again`. Do we have an answer to the first question? And why do we want a different formatting between the own warnings and all the other warnings?

FloLey: Before this was removed, the InmantaWarning where formated with a different rule (see line 155 below this thread).
We want to have both formatting as the default one shows where in the code the warning was raised which might not be relevant for the user. For example, a PluginDeprecationWarning looks like this:
```
py.warnings              WARNING /home/florent/Desktop/inmanta-core/src/inmanta/warnings.py:169: PluginDeprecationWarning: Plugin 'pl' in module 'inmanta_plugins.modv2' is deprecated. It should be replaced by 'hello'.
                                   warnings.warn(*args, **kwargs)
```
and before the InmantaWarning where removed it looked like:
```
py.warnings              PluginDeprecationWarning: Plugin 'pl' in module 'inmanta_plugins.modv2' is deprecated. It should be replaced by 'hello'.
```

arnaudsjs: I see. In that case I would introduce the convention that Inmanta warnings need to be logged using the `inmanta.warnings.warn()`. That way we prevent the duplicate hierarchy and we can easily distinguish Inmanta warnings from other warnings by check whether `filename` of the warning matches `inmanta.warnings.__file__`. The downside of this approach is that we should not forget the log warnings using our own `warn()` method instead of using the default `warnings.warn()` method. But that problem also exist when we re-introduce the `InmantaWarning` class. In that case we need to make sure that we always log warnings using a category that is a subclass of the `InmantaWarning` class. @sanderr Wdyt?

sanderr: I like it.

sanderr: I think this means we'll need to update the ADR as well.

src/inmanta/warnings.py
```diff
@@ -140,15 +140,19 @@ def _showwarning(
...
-            message,  # type: ignore
-            category,
-            filename,
-            lineno,
-            line,
-        )
-        logger: logging.Logger = logging.getLogger("py.warnings")
+        if filename == __file__:
+            text: str = "%s: %s" % (category.__name__, message)
+            logger: logging.Logger = logging.getLogger("inmanta.warnings")
```
arnaudsjs, line 145: * I would also add some documentation here regarding why we use a different logger for warnings that we generate ourselves.
* We also need to update the documentation of the `inmanta.warnings.warn()` method to indicate that this method should be used to log warnings instead of the `warnings.warn()` method. Currently is still mentions this method is only used by the tests, which is no longer true.
* We also need to update all calls to `warnings.warn()` and change them to use the `inmanta.warnings.warn()` method.

arnaudsjs (changes requested):

src/inmanta/warnings.py
arnaudsjs: * We also need to add the `@stable_api` annotation to the `inmanta.warnings.warn()` method, because this method might be used by handler code as well.
* We also need to add a test to `test_import_entry_points.py`.

arnaudsjs: Let's wait with the implementation of these changes. I would like to discuss this with the team first.

PR #5071, opened 2022-11-03
5004/fix conflict with std

### LINKED ISSUE #5004: [iso4] constrain Jinja
The std module has a constraint on Jinja that is not present on core. Since the compiler environment is activated in-process after this dependency is already loaded in `sys.modules`, core's imcompatible version is used in practice.

- [ ] Constrain Jinja to the same version as it is constrained to in the std module on the iso4 branch
- [ ] Check if there are other dependency conflicts between std and core

FloLey:
# Description

The std module has a constraint on Jinja that is not present on core. Since the compiler environment is activated in-process after this dependency is already loaded in sys.modules, core's imcompatible version is used in practice. This commit will 
constrain Jinja to the same version as it is constrained to in the std module.

closes #5004 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/changelogs/unreleased/5004-constraint-conflict-std.yml
+++ b/changelogs/unreleased/5004-constraint-conflict-std.yml
@@ -1,0 +1,4 @@
+description: "constraints Jinja and email_validator to same version as std"
+change-type: patch
+destination-branches:
+  - iso4
--- a/changelogs/unreleased/5004-constraint-conflict-std.yml
+++ b/changelogs/unreleased/5004-constraint-conflict-std.yml
@@ -1,4 +1,6 @@
-description: "constraints Jinja and email_validator to same version as std"
+description: "adapt constraints to not conflict with those of std"
 change-type: patch
 destination-branches:
   - iso4
+sections:
+  bugfix: "{{description}}"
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/5004-constraint-conflict-std.yml
```diff
@@ -0,0 +1,4 @@
+description: "constraints Jinja and email_validator to same version as std"
+change-type: patch
+destination-branches:
```
sanderr, line 3: Should we add a bugfix entry?

sanderr (changes requested):
We also need to modify the constraint(s?) in `setup.py`.

changelogs/unreleased/5004-constraint-conflict-std.yml
FloLey: I would only add a bugfix entry if it ever broke something, but did it?

sanderr: [It did](https://inmanta.slack.com/archives/CKRF0C8R3/p1666012217831119). Only for Python>3.6 though (see [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1666016740737439?thread_ts=1666012217.831119&cid=CKRF0C8R3)).

sanderr (approved):

PR #5080, opened 2022-11-04
replaced compiler with mock

wouterdb:
# Description

Improve test speed and stability by using mock instead of real compiler 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/server/test_compilerservice.py
+++ b/tests/server/test_compilerservice.py
@@ -25,7 +25,7 @@ import subprocess
 import uuid
 from asyncio import Semaphore
 from collections import abc
-from typing import AsyncIterator, List, Optional, Tuple
+from typing import TYPE_CHECKING, AsyncIterator, List, Optional, Tuple
 
 import pkg_resources
 import pytest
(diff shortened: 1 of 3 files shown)


tests/server/test_compilerservice.py
```diff
@@ -795,6 +800,9 @@ async def run_compile_and_wait_until_compile_is_done(
     # prevent race conditions where compile request is not yet in queue
     await retry_limited(lambda: not compiler_queue.empty(), timeout=10)
     run = compiler_queue.get(block=True)
+    if fail is not None:
```
FloLey, line 803: Is it intended that we can make it fail by passing it "True" or "False"?

FloLey (changes requested):
There seems to be an issue with the import of conftest when the build is run but I don't understand why as it works locally. Do you know?

tests/server/test_compilerservice.py
wouterdb: no, if it is set, it will use that value. 

1. if `fail_on_pull` is None, it will not set the value, using the default set when constructing the mock
2. if `fail_on_pull` is True, it will set it to fail
2. if `fail_on_pull` is False, it will set it to succed

FloLey: woof, I think I convinced myself that something was wrong, sorry for that.

PR #5121, opened 2022-11-17
Issue/5066 fix format warnings iso4

FloLey:
# Description

https://github.com/inmanta/inmanta-core/pull/4979 removed InmantaWarnings which use another formatter than the default warnings. As we want to keep both formats, this PR adds the InmantaWarnings back.

closes https://github.com/inmanta/inmanta-core/issues/5066

# Self Check:

- [x] Changelog entry


src/inmanta/ast/__init__.py
```diff
@@ -587,7 +585,7 @@ class CompilerRuntimeWarning(Warning, RuntimeException):
     """
 
     def __init__(self, stmt: "Optional[Locatable]", msg: str) -> None:
-        Warning.__init__(self)
+        InmantaWarning.__init__(self)
```
arnaudsjs, line 588: The class this class inherits from should be updated to `InmantaWarning` as well.

arnaudsjs (changes requested):

PR #5131, opened 2022-11-17
issue/5054-modulev2-invalid-version

### LINKED ISSUE #5054: module build does not recognize v2 modules with invalid version
When you create a v2 module and you set the version to an invalid python version, the build command will only say: `No module can be found at /home/bart/tmp/test-module`

This is not very informative. Additionally, when migrating from v1 to v2 and you try to add a dev tag in the version instead of the tag_build you will get this.

Steps to reproduce:
- Use/create a v2 module
- Set version to something invalid. For example 0.0.1.dev0
- Run inmanta module build

FloLey:
# Description

When we create a v2 module and we set the version to an invalid python version, the build command will show a error message that doesn't give much information. This commit will improve the error message.

closes #5054

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/module.py
+++ b/src/inmanta/module.py
@@ -64,6 +64,7 @@ import pkg_resources
 import yaml
 from pkg_resources import Distribution, DistributionNotFound, Requirement, parse_requirements, parse_version
 from pydantic import BaseModel, Field, NameEmail, ValidationError, constr, validator
+from pydantic.error_wrappers import display_errors
 
 import packaging.version
 from inmanta import RUNNING_TESTS, const, env, loader, plugins
@@ -85,16 +86,13 @@ try:
 except ImportError:
     TYPE_CHECKING = False
 
-
 if TYPE_CHECKING:
     from pkg_resources.packaging.version import Version  # noqa: F401
-
 
 LOGGER = logging.getLogger(__name__)
 
 Path = NewType("Path", str)
 ModuleName = NewType("ModuleName", str)
-
 
 T = TypeVar("T")
 TModule = TypeVar("TModule", bound="Module")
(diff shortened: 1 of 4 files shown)


wouterdb (approved):

src/inmanta/module.py
```diff
@@ -260,10 +258,9 @@ def __init__(self, msg: str, validation_error: Optional[ValidationError] = None)
 
     @classmethod
     def _extend_msg_with_validation_information(cls, msg: str, validation_error: ValidationError) -> str:
-        for error in validation_error.errors():
-            mgs: str = error["msg"]
-            error_type = error["type"]
-            msg += f"\n{error['loc']}\n\t{mgs} ({error_type})"
+        errors = validation_error.errors()
+        if errors:
+            msg += "\n" + display_errors(errors)
```
sanderr, line 263:
```suggestion
            msg += "\n" + textwrap.indent(display_errors(errors), " " * 2)
```
I think this could improve readability a bit.

src/inmanta/module.py
```diff
@@ -3041,10 +3037,7 @@ def __init__(
     ) -> None:
         self._is_editable_install = is_editable_install
         self._version: Optional[version.Version] = installed_version
-        try:
-            super(ModuleV2, self).__init__(project, path)
-        except InvalidMetadata as e:
-            raise InvalidModuleException(f"The module found at {path} is not a valid V2 module") from e
+        super(ModuleV2, self).__init__(project, path)
```
sanderr, line 3040: This change makes me a bit nervous so I triggered the slow tests on your branch. I wonder if we could make `InvalidMetadata` inherit from `InvalidModuleException` and get the best of both worlds? We could then drop the `except InvalidMetadata` in moduletool as well.

FloLey: @sanderr so the slowbuild suceeded, do you still want me to try to make InvalidMetadata inherit from InvalidModuleException? (I haven't look at it at all yet )

sanderr (approved):

src/inmanta/module.py
FloLey: Ok i'll give it a try :)

PR #5132, opened 2022-11-17
issue/5047-contradicting-log-msg-compiler-service

### LINKED ISSUE #5047: Contradicting log message in compiler service
When the `recompile_backoff` is set to 0, which actually disables the `recompile_backoff`, the following line is printed in the logs:
```
The recompile_backoff environment setting is enabled and set to 0 seconds.
```

FloLey:
# Description

fix a contradicting message in the logs of the compiler service.

closes #5047 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5047-fix-contradicting-log-message.yml
+++ b/changelogs/unreleased/5047-fix-contradicting-log-message.yml
@@ -1,0 +1,4 @@
+description: fix a contradicting message in the logs of the compiler service
+change-type: patch
+destination-branches: [master, iso5]
+issue-nr: 5132
--- a/changelogs/unreleased/5047-fix-contradicting-log-message.yml
+++ b/changelogs/unreleased/5047-fix-contradicting-log-message.yml
@@ -1,4 +1,4 @@
 description: fix a contradicting message in the logs of the compiler service
 change-type: patch
-destination-branches: [master, iso5]
-issue-nr: 5132
+destination-branches: [master, iso5, iso4]
+issue-nr: 5047
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -728,7 +728,8 @@ class CompilerService(ServerSlice):
         else:
             env = await data.Environment.get_by_id(compile.environment)
             wait_time = await env.get(data.RECOMPILE_BACKOFF)
-            LOGGER.info("The recompile_backoff environment setting is enabled and set to %s seconds.", wait_time)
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/5047-fix-contradicting-log-message.yml
```diff
@@ -0,0 +1,4 @@
+description: fix a contradicting message in the logs of the compiler service
+change-type: patch
+destination-branches: [master, iso5]
+issue-nr: 5132
```
arnaudsjs, line 4: This issue number is incorrect.

changelogs/unreleased/5047-fix-contradicting-log-message.yml
```diff
@@ -0,0 +1,4 @@
+description: fix a contradicting message in the logs of the compiler service
+change-type: patch
+destination-branches: [master, iso5]
```
arnaudsjs, line 3: This issue also exists on iso4

src/inmanta/server/services/compilerservice.py
```diff
@@ -728,7 +728,8 @@ async def _auto_recompile_wait(self, compile: data.Compile) -> None:
         else:
             env = await data.Environment.get_by_id(compile.environment)
             wait_time = await env.get(data.RECOMPILE_BACKOFF)
-            LOGGER.info("The recompile_backoff environment setting is enabled and set to %s seconds.", wait_time)
+            if wait_time:
+                LOGGER.info("The recompile_backoff environment setting is enabled and set to %s seconds.", wait_time)
```
arnaudsjs, line 732: I would not log less then what use used to log. I would just make the message more correct:

```suggestion
            if wait_time:
                LOGGER.info("The recompile_backoff environment setting is enabled and set to %s seconds.", wait_time)
            else:
                LOGGER.info("The recompile_backoff environment setting is disabled)
```

arnaudsjs (changes requested):

arnaudsjs (approved):

FloLey:
@arnaudsjs actually I think this issue wasn't present neither on iso5 nor on iso4. To fix the invariant check i'll need to create a PR with just the changelog right?

arnaudsjs:
> @arnaudsjs actually I think this issue wasn't present neither on iso5 nor on iso4. To fix the invariant check i'll need to create a PR with just the changelog right?

You are right. I am not sure how I missed that. Indeed. Creating a PR with just a change entry is good.

PR #5137, opened 2022-11-18
Issue/4376 inmanta workon

### LINKED ISSUE #4376: Add inmanta-workon command to the rpm
# Problem

One of the common steps when diagnosing issues with production orchestrators is to manipulate the module and venv on the server directly via the cli

This is currently a bit involved because
- the folder is deeply nested `/var/lib/inmanta/server/environments/`
- venv needs to be activated
- even when the venv is activated, the inmanta command doesn't work (for iso 5) (see #4259)
-  the environment can not be adressed by name

This also often leads to problem when the root user is used in the folder, making it impossible for the inmanta user to modify them

# Solution

Provide an inmanta-workon command in the rpm that 
- changes the working dir to the project dir
- activates the venv
- registers a inmanta-deactivate alias that, when called, verfies if there are files in the working dir that are not owned by inmanta. If this is the case, a warning is produced. 
- syntax: `inmanta-workon <environment_name>|<environment_id>`, raise an exception if env is not unique + list all environments with uuids.
- if authentication is enabled, need to use uuid
- update RPM specfiles to put the executable in `/usr/bin/`
- make sure `inmanta` command works and operates on the activated environment

sanderr:
# Description

Added the `inmanta-workon` command. I will create separate pull requests on `inmanta` and `inmanta-service-orchestrator` for the RPM installation.

@wouterdb I'm adding you as a reviewer only for the documentation and change entry part. Is the `model_developers` section the best place for it?

part of #4376

# Self Check:

- [x] Changelog entry


misc/inmanta-workon-register.sh
```diff
@@ -0,0 +1,232 @@
...
+#       %files -n python3-%{modname}
+#       %doc PKG-INFO docs
+#       %license LICENSE
+#       %{python3_sitelib}/%{modname}
+#       %{python3_sitelib}/%{modname}-%{version}*
+#       %{_bindir}/virtualenvwrapper.sh
+#       %{_bindir}/virtualenvwrapper_lazy.sh
+#       %{_bindir}/virtualenvwrapper-3.sh
+#       %{_bindir}/virtualenvwrapper_lazy-3.sh
+#       %config(noreplace) %{_sysconfdir}/profile.d/virtualenvwrapper.sh
```
sanderr, line 28: I'm aware of these TODO's. Will remove them when I'm done.

tests/server/conftest.py
```diff
@@ -0,0 +1,91 @@
+"""
```
sanderr, line 1: Extracted all this from `tests/server/test_compilerservice.py`

src/inmanta/main.py
```diff
@@ -342,17 +349,29 @@ def environment_list(client: Client) -> None:
...
     Show details of an environment
 
     ENVIRONMENT: ID or name of the environment to show
     """
     env = client.get_dict("get_environment", "environment", dict(id=client.to_environment_id(environment)))
-    print_table(
-        ["ID", "Name", "Repository URL", "Branch Name"], [[env["id"], env["name"], env["repo_url"], env["repo_branch"]]]
-    )
+    if format_string is not None:
+        print(format_string.format(**env))
```
sanderr, line 370: Just to be sure, in Python there should be no security concerns regarding this, right?

sanderr:
Apparently I missed some details in the requirements. I'll have to address those still:
- raise an exception if env is not unique + list all environments with uuids.
- make sure `inmanta` command works and operates on the activated environment

arnaudsjs (changes requested):

misc/inmanta-workon-register.sh
```diff
@@ -0,0 +1,232 @@
...
+fi
+if [ "$BASH_SOURCE" = "$0" ]; then
+    echo "ERROR: This script is meant to be sourced rather than executed directly: \`source '$0'\`" >&2
+    exit 1
+fi
+
+# Locate paths to required executables. Make sure to use executables that live in the same Python environment.
+if [ -z "$INMANTA_WORKON_CLI" ]; then
+    # get path to inmanta-cli executable
+    INMANTA_WORKON_CLI="$(command which inmanta-cli)"
```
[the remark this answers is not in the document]
sanderr: I didn't know about it either until I started on this ticket!

sanderr:
Processed review comments and addressed missing requirements. I also made a change to the environment service that I think should be acceptable. Could you pay special attention to that?

sanderr:
(@arnaudsjs @Hugo-Inmanta tagging you for the comment above)

arnaudsjs (approved):

misc/inmanta-workon-register.sh
```diff
@@ -0,0 +1,209 @@
...
+    fi
+
+    # if we are in an active venv, deactivate it first: restores PS1 and triggers our custom deactivate logic if it's an inmanta venv
+    declare -F deactivate > /dev/null && deactivate
+    # store PS1 before sourcing activate because we don't care about virtualenv's modifications
+    declare OLD_PS1="$PS1"
+    source "$activate"
+    export PS1="($env_name) $OLD_PS1"
+
+    eval 'inmanta () { python3 -m inmanta.app; }' # workaround for #4259
```
arnaudsjs, line 174: Don't you have to forwards the parameters of the function to the `python3 -m inmanta.app` command?

arnaudsjs (changes requested):

misc/inmanta-workon-register.sh
sanderr: I would think so. I started out with an alias, where that isn't required. Then I moved to the function and forgot to change it. The strange thing is that it actually still works, as you can see by defining the function `function list() { ls; }`. But I must admit I don't really see how.

sanderr: (Investigating)

sanderr: Turns out I'm stupid: I was testing with `ls` rather than `list. Apparently `ls` is so ingrained in my fingers that I didn't even notice. Concluding, you're right, I need to change this. I do wonder how the tests didn't fail on this though.

sanderr: Fixed

docs/model_developers/model_debugging.rst
```diff
@@ -1,5 +1,24 @@
 Model debugging
```
wouterdb, line 1: I think adminstrator documentation is more appropriate.

Hugo-Inmanta (approved):

docs/administrators/operational_procedures.rst
```diff
@@ -189,6 +189,22 @@ This procedure only works when all agents are autostarted by the server.
    * `autostart_agent_repair_interval`
 
 
+Working directly on the orchestrator
```
wouterdb, line 192: 1. I would move this to it's own section (e.g. 'diagnosing problems`)
2. I would also update the `operational_procedures_with_lsm.html` to use the workon command

wouterdb (changes requested):

docs/administrators/operational_procedures.rst
sanderr: 2. Does inmanta/inmanta-service-orchestrator#356 match your expectations?

wouterdb (approved):

PR #5138, opened 2022-11-18
Issue/fix path in migration check

Hugo-Inmanta:
# Description

Relative paths were causing issues in the docker-pytest builds.
This PR changes the paths to absolute.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/tests/db/test_migration_check.py
+++ b/tests/db/test_migration_check.py
@@ -15,6 +15,7 @@
 
     Contact: code@inmanta.com
 """
+import os
 from pathlib import Path
 from typing import List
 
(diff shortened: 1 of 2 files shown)


tests/db/test_migration_check.py
```diff
@@ -24,12 +25,13 @@ def test_migration_check():
     Make sure there is a database dump for the latest version of the db and
     that a migration test exists for this dump.
     """
-    versions_folder: Path = Path(".").absolute() / "src" / "inmanta" / "db" / "versions"
+    inmanta_dir: str = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))
```
sanderr, line 28: Why use `os.path` here and `Path` below?

sanderr (approved):

tests/db/test_migration_check.py
Hugo-Inmanta: I didn't realize what I needed from `os.path` was present in `Path` as well.
I uniformized it, using `Path` everywhere.

PR #5120, opened 2022-11-16
Issue/4984 drop default constructor

FloLey:
# Description
drop default constructors

drop redundant EntityLike ABC (move docstrings to Entity where appropriate).

Drop get_entity() method and simply use the entity where it's currently used

closes #4984

# Self Check:

Strike through any lines that are not applicable (`~~line~~`) then check the box

- [x] Attached issue to pull request
- [x] Changelog entry
- [x] Type annotations are present
- [x] Code is clear and sufficiently documented
- [x] No (preventable) type errors (check using make mypy or make mypy-diff)
- [x] Sufficient test cases (reproduces the bug/tests the requested feature)
- [x] Correct, in line with design
- [ ] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )

PR #5160, opened 2022-11-23
Fix: backwards compatibility EntityLike 

FloLey:
# Description

EntityLike where dropped in (https://github.com/inmanta/inmanta-core/pull/5120) and caused some fallout.
This PR will fix some backwards compatibility issues.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -478,6 +478,7 @@ class Entity(NamedType):
         return self.location
 
 
+# Kept for backwards compatibility. May be dropped from iso7 onwards.
 EntityLike = Entity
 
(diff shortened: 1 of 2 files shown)


src/inmanta/ast/entity.py
```diff
@@ -469,6 +478,9 @@ def get_location(self) -> Location:
         return self.location
 
 
+EntityLike = Entity
```
sanderr, line 481:
```suggestion
# Kept for backwards compatibility. May be dropped from iso7 onwards.
EntityLike = Entity
```

sanderr (approved):

PR #5161, opened 2022-11-23
Fix: increase timeout compiled environments

FloLey:
# Description

increase the timeout in the compiled environments fixture.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/fix-increase-timeout-compiled-environments.yml
+++ b/changelogs/unreleased/fix-increase-timeout-compiled-environments.yml
@@ -1,0 +1,4 @@
+---
+description: increase the timeout for compiled_environments
+change-type: patch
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/fix-increase-timeout-compiled-environments.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: increase the timeout in the compiled_environments fixture
+change-type: patch
+destination-branches: [master]
```
sanderr, line 4:
```suggestion
destination-branches: [master, iso5]
```

sanderr (changes requested):

sanderr (approved):

PR #5178, opened 2022-11-25
Add inmanta module release command

### LINKED ISSUE #5082: implement module release command
Implement the `inmanta module release` command with the following semantics:
- without arguments, it expects a dev version `x.y.z.devx` and converts it to a stable version `x.y.z`. It commits, adds a release tag, then calls `inmanta module release --dev --patch` (described below).
- with the `--dev` argument it converts the current version to a dev version. It optionally accepts `--patch`, `--minor` or `--major`, defaulting to `--patch`. The command results in a dev version that is at least the requested bump ahead of the last version tag. The changelog should be updated accordingly (either add a new section or rename the current one depending on the type of bump that was done). The command should commit the changes. It should not add a git tag.

Similar logic for dev bumping with respect to a bump specifier and the latest tag is currently implemented in `irt.release.release`, which might be of use.

Make sure to take the following into account:
- It should be compatible with both v1 and v2 modules. In particular, v1 modules have tags directly in the version field while v2 modules have it in the `egg_info.tag_version` field in `setup.cfg`. The metadata objects in `inmanta.module` likely already contain all abstractions to interact with tagged versions for both module generations.
- It should not break for modules that have never been released yet (have no last version tag)

Afterwards notify the solutions team.

arnaudsjs:
# Description

Added the `inmanta module release` command.

closes #5082 

# Self Check:

- [x] Changelog entry


tests/moduletool/test_release.py
```diff
@@ -0,0 +1,376 @@
...
+                stable_version_number = expected_version.rsplit(".", maxsplit=1)[0]
+                assert len(re.findall(re.escape(stable_version_number), fh.read())) == 2
+        else:
+            assert not os.path.exists(path_changelog_file)
+        assert get_commit_message_x_commits_ago(path=path_module, nb_previous_commit=0).strip() == expected_commit_message
+
+    monkeypatch.chdir(path_module)
+    module_tool = ModuleTool()
+    module_tool.release(dev=True, patch=True, message="Commit patch increment")
+    assert_module_state("1.1.2.dev0", expected_commit_message="Initial commit")
```
Hugo-Inmanta, line 247: I can't figure out why the latest commit isn't "Commit patch increment"

tests/moduletool/test_release.py
```diff
@@ -0,0 +1,376 @@
...
+    gitprovider.git_init(repo=path_module)
+    gitprovider.commit(repo=path_module, message="Initial commit", add=["*"], commit_all=True)
+    monkeypatch.chdir(path_module)
+    module_tool = ModuleTool()
+    with pytest.raises(click.ClickException) as exc_info:
+        module_tool.release(dev=True, minor=True, message="Commit changes")
+    assert "Version with an epoch value larger than zero are not supported by this tool." in exc_info.value.message
+
+
+def test_no_a_git_repository(tmpdir, modules_dir: str, monkeypatch) -> None:
```
Hugo-Inmanta, line 359:
```suggestion
def test_not_a_git_repository(tmpdir, modules_dir: str, monkeypatch) -> None:
```

Hugo-Inmanta (changes requested):

src/inmanta/moduletool.py
```diff
@@ -923,6 +1066,142 @@ def freeze(self, outfile: Optional[str], recursive: Optional[bool], operator: st
...
+        """
+        Turn the given current_version into a dev version with version_tag dev0 and ensure
+        the version number is at least `minimal_version_bump_to_prev_release` separated
+        from its predecessor in all_existing_stable_version.
+        """
+        if not minimal_version_bump_to_prev_release:
+            # No version bump is required
+            return VersionOperation.set_version_tag(current_version, version_tag="dev0")
+        try:
+            version_previous_release: Version = [v for v in sorted(all_existing_stable_version) if v <= current_version][-1]
```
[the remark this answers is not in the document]
arnaudsjs: True. Thanks!

tests/moduletool/test_module_tool_basics.py
```diff
@@ -145,16 +152,16 @@ def test_rewrite(tmpdir, module_type: Type[module.Module]):
...
             return f"""
 name: mod
 license: ASL
-version: {version}
+version: {version if not version_tag else f"{version}.{version_tag.rstrip('.')}"}
 compiler_version: 2017.2
             """.strip()
         else:
-            return f"""
+            result = f"""
```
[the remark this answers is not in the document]
arnaudsjs: Indeed. This is a leftover from a previous implementation. I will make it consistent.

tests/moduletool/test_module_tool_basics.py
```diff
@@ -167,7 +174,10 @@ def metadata_contents(version: str) -> str:
 
   cookiecutter~=1.7.0
   cryptography>1.0,<3.5
-            """.strip()
+[egg_info]
+tag_build = {version_tag}
+            """
```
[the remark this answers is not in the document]
arnaudsjs: I removed the strip() here, because otherwise the whitespace between the `=` and `{version_tag}` gets removed when `version_tag` is an empty string. The trip in the if-block is redundant, so I removed it there too for consistency.

tests/moduletool/test_release.py
arnaudsjs: The previous stable release of the module is version  `1.1.1` and the current version of the module is `1.1.2.dev0`. So they are already a patch increment separated from each other. When executing `inmanta module release --dev --patch -m "Commit patch increment"`, no commit is required to bump the version. As such, there is no commit with the message `Commit patch increment`. Should I log a warning saying that no version bump was required?

Hugo-Inmanta: I see it now :+1:

Hugo-Inmanta (approved):

src/inmanta/moduletool.py
```diff
@@ -31,9 +32,11 @@
 import zipfile
 from argparse import ArgumentParser
 from configparser import ConfigParser
+from functools import total_ordering
 from types import TracebackType
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Pattern, Sequence, Set, Type
+from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Pattern, Sequence, Set, Type
```
sanderr, line 37: We should preferably use `collections.abc` for new code.

src/inmanta/moduletool.py
```diff
@@ -178,20 +183,12 @@ def determine_new_version(self, old_version, version, major, minor, patch, dev):
...
-                if minor:
-                    parts[1] += 1
-                    parts[2] = 0
-                if major:
-                    parts[0] += 1
-                    parts[1] = 0
-                    parts[2] = 0
-                outversion = ".".join([str(x) for x in parts])
+
+                change_type: Optional[ChangeType] = ChangeType.parse_from_bools(patch, minor, major)
```
sanderr, line 187: This is just the legacy method, right? I think it could be useful to add a comment at the method definitition stating that this is only used by the old `inmanta module commit` command.

src/inmanta/moduletool.py
```diff
@@ -562,6 +663,48 @@ def modules_parser_config(cls, parser: ArgumentParser) -> None:
...
+            help="Do a minor version bump compared to the previous stable release. Ignored when --dev is not set.",
+            action="store_true",
+        )
+        release.add_argument(
+            "--patch",
+            dest="patch",
+            help="Do a patch version bump compared to the previous stable release. Ignored when --dev is not set.",
+            action="store_true",
+        )
+        release.add_argument("-m", "--message", help="Commit message", required=True)
```
sanderr, line 706: Do we still want to require this? For `inmanta module commit` the purpose was really to commit a change, part of which was the version bump. For `inmanta module release` it's usually really nothing more than a release or a version bump (like what `bumpversion` does). I think a default message would make sense.

sanderr (changes requested):
I need to context switch now so releasing intermediary comments. Still need to review everything below `src/inmanta/moduletoo.py:1069` as well as the tests.

tests/moduletool/test_changetype.py
```diff
@@ -0,0 +1,57 @@
...
+    assert ChangeType.diff(low=Version("1.0.0"), high=Version("1.3.1")) == ChangeType.MINOR
+
+    assert ChangeType.diff(low=Version("1.0.0"), high=Version("1.0.1")) == ChangeType.PATCH
+    assert ChangeType.diff(low=Version("1.0.0"), high=Version("1.0.2")) == ChangeType.PATCH
+
+    assert ChangeType.diff(low=Version("1.0.0.dev0"), high=Version("1.0.1.dev0")) == ChangeType.PATCH
+    assert ChangeType.diff(low=Version("1.0.0.dev0"), high=Version("1.1.0.dev0")) == ChangeType.MINOR
+    assert ChangeType.diff(low=Version("1.0.0.dev0"), high=Version("2.0.0.dev0")) == ChangeType.MAJOR
+
+    assert ChangeType.diff(low=Version("1.0.0"), high=Version("1.0.0")) is None
```
sanderr, line 54:
```suggestion
    assert ChangeType.diff(low=Version("1.0.0"), high=Version("1.0.0")) is None
    assert ChangeType.diff(low=Version("1.0.0.dev0"), high=Version("1.0.0")) is None
```

tests/moduletool/test_release.py
```diff
@@ -0,0 +1,377 @@
...
+                               0 is the previous commit, -1 the commit before the previous commit, etc.
+    """
+    if nb_previous_commit < 0:
+        raise Exception("Argument `nb_previous_commit` should be >= 0")
+    return subprocess.check_output(["git", "log", "-1", f"--skip={nb_previous_commit}", "--pretty=%B"], cwd=path).decode()
+
+
+@pytest.mark.parametrize("v1_module", [True, False])
+@pytest.mark.parametrize("changelog_file_exists", [True, False])
+@pytest.mark.parametrize("previous_stable_version_exists", [True, False])
```
sanderr, line 50: Should these tests use `parametrize_any`?

src/inmanta/module.py
```diff
@@ -1144,21 +1243,36 @@ def rewrite_version(cls: Type[TModuleMetadata], source: str, new_version: str) -
...
+                f"Unable to write version in module definition, should be {expected_version} "
+                f"got {new_metadata.version} instead."
+            )
+
+        # Validate whether version_tag field was updated correctly in metadata file
+        if isinstance(new_metadata, ModuleV2Metadata) and version_tag != new_metadata.version_tag:
+            raise Exception(
+                f"Unable to write tag_build in module definition, should be '{version_tag}' got "
+                f"'{new_metadata.version_tag}' instead."
+            )
```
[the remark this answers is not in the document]
arnaudsjs: I see your point, but I am still on doubt whether this is an actual improvement, because this changes makes the error reporting more generic and as such also more vague. The location where the version tag is stored is dependent on the type of module (V1 vs V2). The previous error message pointed to the exact field that was causing issues, with the new implementation that's not possible. So you have to be very much aware about the differences between both types of modules to be able to interpret the error message. In our team that won't be an issue, but for external people it might be. Maybe I have to try to embed this information in the error message. I will give that a try.

src/inmanta/module.py
```diff
@@ -1262,13 +1377,16 @@ def get_version_tag(v: packaging.version.Version) -> Optional[str]:
...
 
         return v.base_version, get_version_tag(v)
 
     @validator("version_tag")
     @classmethod
-    def is_valid_version_tag(cls, v: str) -> str:
+    def is_valid_version_tag(cls, v: Optional[str]) -> str:
+        if v is None:
+            # Represent a missing version_tag using an empty string
+            return ""
```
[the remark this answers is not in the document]
arnaudsjs: True. This is a leftover from a previous implementation. I will remove it.

src/inmanta/module.py
```diff
@@ -2879,16 +3014,16 @@ def update(
...
-        def try_parse(x: str) -> "Version":
+        def try_parse(x: str) -> Optional[version.Version]:
             try:
                 return parse_version(x)
             except Exception:
                 return None
 
-        versions = [x for x in [try_parse(v) for v in versions] if x is not None]
+        versions: List[version.Version] = [x for x in [try_parse(v) for v in versions_str] if x is not None]
         versions = sorted(versions, reverse=True)
```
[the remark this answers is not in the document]
arnaudsjs: Yes. I didn't replace it, because it would break backwards compatibility.

src/inmanta/module.py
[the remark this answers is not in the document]
sanderr: You make a good point. Alternatively we could offer an abstract method on the ABC to get the version and tag in a tuple, then stick with your logic but with polymorphism rather than isinstance?

sanderr (approved):

PR #5183, opened 2022-11-28
fix-issues-python3.11

FloLey:
# Description

 Fix an issue about enum representation which made a test fail for python3.11

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/environmentservice.py
+++ b/src/inmanta/server/services/environmentservice.py
@@ -606,7 +606,7 @@ class EnvironmentService(protocol.ServerSlice):
                 if action == EnvironmentAction.updated and original_env:
                     await listener.environment_action_updated(updated_env, original_env)
             except Exception:
-                LOGGER.warning(f"Notifying listener of %s failed with the following exception", action.value, exc_info=True)
+                LOGGER.warning("Notifying listener of %s failed with the following exception", action.value, exc_info=True)
 
     async def register_setting(self, setting: Setting) -> None:
         """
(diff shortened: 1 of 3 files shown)


src/inmanta/server/services/environmentservice.py
```diff
@@ -606,7 +606,7 @@ async def notify_listeners(
                 if action == EnvironmentAction.updated and original_env:
                     await listener.environment_action_updated(updated_env, original_env)
             except Exception:
-                LOGGER.warning(f"Notifying listener of {action} failed with the following exception", exc_info=True)
+                LOGGER.warning(f"Notifying listener of {action.value} failed with the following exception", exc_info=True)
```
arnaudsjs, line 609:
```suggestion
                LOGGER.warning(f"Notifying listener of %s failed with the following exception", action.value, exc_info=True)
```

arnaudsjs (changes requested):

src/inmanta/server/services/environmentservice.py
FloLey: Is this an important change or is it for coherence/preference?

arnaudsjs: It's a small performance improvement. With the previous implementation, the f-string would always get expanded. Also when the log level is set as such that the line wouldn't appear in the log. The new implementation only does the variable expansion when the log message is logged. The best practice is to always use the %-based notation in log messages.

arnaudsjs (approved):

PR #5189, opened 2022-11-29
Fix issue used port

FloLey:
# Description

Fix an issue where some tests would always use the same port, which caused failures if the port was already used for something else.

# Self Check:

- [x] Changelog entry


arnaudsjs (approved):

arnaudsjs (changes requested):
* I did a grep for `Server()` in out test suite and it seems we have a lot more occurrences of this issue than the two instances fixed in this PR. Do you also fix these?
* I think it would be convenient to create a new fixture that 1) Creates the `Server()` instance, 2) Sets the config in place, 3) yields the server object and 4) in its tear down executes the `server.stop()` method. Or something similar if you see a better abstraction. This way we prevent code duplication and we have guarantees that all config is applied consistently everywhere.

FloLey:
@arnaudsjs I think I should not add a new fixture as there are already quit a few that do similar things. I changed my current solution to use one of those existing fixtures. Do you like it this way?

arnaudsjs:
> @arnaudsjs I think I should not add a new fixture as there are already quit a few that do similar things. I changed my current solution to use an existing helper function (not sure why it isn't a fixture?). Do you like it this way? (I found it by looking at test that were still calling `Server() `)

Nice. I didn't know we already had this functionality. The advantage of using a fixture is that you can include teardown code in the fixture as well. But for me it doesn't have to be a fixture.

Hugo-Inmanta (approved):
Looks good to me :+1:

arnaudsjs (changes requested):
The test case `test_2151_method_header_parameter_in_body` in `tests/test_protocol.py` needs this fix as well, no?

arnaudsjs (approved):

changelogs/unreleased/issue-test-using-used-port.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: Fix an issue where some tests would always use the same port, which caused failures if the port was already used for something else.
+change-type: patch
+destination-branches: [master, iso6, iso5, iso4]
```
FloLey, line 4: @sanderr is this right?

sanderr:
```suggestion
destination-branches: [master, iso6, iso5]
```
See slack.

PR #5198, opened 2022-12-01
[WIP]Issue/5124 add support collect metrics db

### LINKED ISSUE #5124: Add support to write collect metrics to the database
This ticket adds the main building blocks of the metrics collection framework and adds support to write metrics to the database.

**Tasks:**

- [x] Add the `EnvironmentMetricsService` slice
- [x] Add support to register MetricsCollectors (`register_metric_collector()` method).
- [x] Add support to write collected metrics to the database (`flush_metrics()` method).
       -> Ensure that counter and non-counter metrics are supported
- [x] Add a timer trigger that invokes the `flush_metrics()` method every minute.
- [x] Log a warning when the executing of the `flush_metrics` method takes longer than 1 minute.

FloLey:
# Description

This ticket adds the main building blocks of the metrics collection framework and adds support to write metrics to the database.

Tasks:

 

- [x] Add the EnvironmentMetricsService slice
- [x] Add support to register MetricsCollectors (register_metric_collector() method).
- [x] Add support to write collected metrics to the database (flush_metrics() method)
- [x] Ensure that counter and non-counter metrics are supported
- [x] Add a timer trigger that invokes the flush_metrics() method every minute.
- [x] Log a warning when the executing of the flush_metrics method takes longer than 1 minute.

closes #5124 

# Self Check:

- [x] Changelog entry


src/inmanta/db/versions/v202212010.py
```diff
@@ -0,0 +1,43 @@
...
+
+async def update(connection: Connection) -> None:
+    schema = """
+    CREATE TABLE IF NOT EXISTS public.environmentmetricscounter(
+        metric_name VARCHAR NOT NULL,
+        timestamp TIMESTAMP NOT NULL,
+        count INT NOT NULL,
+        PRIMARY KEY (metric_name, timestamp)
+    );
+    CREATE INDEX IF NOT EXISTS environment_metrics_counter_index ON environmentmetricscounter(metric_name);
```
FloLey, line 30: not sure if I should (already) add it and if I should also include the timestamp.

src/inmanta/db/versions/v202212010.py
```diff
@@ -0,0 +1,43 @@
...
+
+    CREATE TABLE IF NOT EXISTS public.environmentmetricsnoncounter (
+        metric_name VARCHAR NOT NULL,
+        timestamp TIMESTAMP NOT NULL,
+        count INT NOT NULL,
+        value INT NOT NULL,
+        PRIMARY KEY (metric_name, timestamp)
+    );
+    CREATE INDEX IF NOT EXISTS environment_metrics_counter_index ON environmentmetricsnoncounter(metric_name);
```
FloLey, line 39: not sure if I should (already) add it and if I should also include the timestamp.

src/inmanta/data/__init__.py
```diff
@@ -5267,6 +5267,59 @@ def to_dto(self) -> m.Notification:
...
+    """
+
+    metric_name: str
+    timestamp: datetime.datetime
+    count: int
+
+    __primary_key__ = ("metric_name", "timestamp")
+
+    @classmethod
+    def new(cls, metric_name: str, timestamp: datetime.datetime, count: int) -> "EnvironmentMetricsCounter":
```
wouterdb, line 5286: What is this for?

src/inmanta/data/__init__.py
```diff
@@ -5267,6 +5267,59 @@ def to_dto(self) -> m.Notification:
...
+        attr = dict(
```
wouterdb, line 5287: why pack the argument in a dict to unpack it again?

src/inmanta/data/__init__.py
```diff
@@ -5267,6 +5267,59 @@ def to_dto(self) -> m.Notification:
...
+        return cls(**attr)
+
+
+class EnvironmentMetricsNonCounter(BaseDocument):
+    """
+    A metric that is not a counter
+
+    :param metric_name: The name of the metric
+    :param timestamp: The timestamps at which a new record is created
+    :param count: the number of occurrences for the given timestamp
```
wouterdb, line 5301:
```suggestion
    :param count: the number of occurrences for in the interval [previous.timestamp, self.timestamp[
```

src/inmanta/db/versions/v202212010.py
wouterdb: What is this index for?

src/inmanta/db/versions/v202212010.py
wouterdb: What is this index for?

changelogs/unreleased/5124-add-support-collect-metrics.yml
```diff
@@ -0,0 +1,6 @@
+description: Add support to write collected metrics to the database
+issue-nr: 5124
+change-type: minor
+destination-branches: [master, iso6]
+sections:
+  minor-improvement: "{{description}}"
```
arnaudsjs, line 6: I wouldn't add a changelog entry for each sub-ticket in this epic, but only a single one at the end of the epic.
```suggestion
```

src/inmanta/db/versions/v202212010.py
```diff
@@ -0,0 +1,43 @@
...
+        count INT NOT NULL,
+        PRIMARY KEY (metric_name, timestamp)
+    );
+    CREATE INDEX IF NOT EXISTS environment_metrics_counter_index ON environmentmetricscounter(metric_name);
+
+    CREATE TABLE IF NOT EXISTS public.environmentmetricsnoncounter (
+        metric_name VARCHAR NOT NULL,
+        timestamp TIMESTAMP NOT NULL,
+        count INT NOT NULL,
+        value INT NOT NULL,
```
wouterdb, line 36: I'm not sure if int is the most logical choice here.

If we do e.g. compile time, which time-unit are we going to use? 
By using int, we fix the resolution to the unit. (e.g. if we use seconds, we can never measure below 1s, if we use ns, we can never measure longer than 2.14s), if we use floats, we have a fixed resolution of N digits, independent of unit. 

I don't have sufficient feeling to know what should be used.

src/inmanta/data/__init__.py
```diff
@@ -5267,6 +5267,59 @@ def to_dto(self) -> m.Notification:
...
+    :param value: the value of the metric for the given timestamp
```
arnaudsjs, line 5302:
```suggestion
    :param value: The sum of the values of the metric for each occurrence.
```

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,120 @@
...
+        collected by the MetricsCollectors in the past metrics collection interval
+        to the database.
+        """
+        now: datetime = datetime.now()
+        metric_count: List[EnvironmentMetricsCounter] = []
+        metric_non_count: List[EnvironmentMetricsNonCounter] = []
+        for metrics_collector in self.metrics_collectors:
+            metric_name: str = metrics_collector.get_metric_name()
+            metric_type: str = metrics_collector.get_metric_type()
+            metric_value: dict[str, int] = await metrics_collector.get_metric_value(now - timedelta(seconds=60), now)
```
wouterdb, line 108: `now - timedelta(seconds=60)` is not a good way of working: it will make the intervals overlap or have holes depending on when you execute. You really have to keep track of the timestamp used the previous time and use that.

src/inmanta/db/versions/v202212010.py
arnaudsjs: This index is redundant. It's implicitly there because of the primary key.

wouterdb (changes requested):

src/inmanta/db/versions/v202212010.py
arnaudsjs: Same here.

src/inmanta/db/versions/v202212010.py
```diff
@@ -0,0 +1,43 @@
...
+        PRIMARY KEY (metric_name, timestamp)
+    );
+    CREATE INDEX IF NOT EXISTS environment_metrics_counter_index ON environmentmetricsnoncounter(metric_name);
+    """
+
+    async with connection.transaction():
+        await connection.execute(schema)
```
arnaudsjs, line 43: The database migration tool starts a transaction behind the scenes. It's not required to start a second one here.

```suggestion
    await connection.execute(schema)
```

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,120 @@
...
+from inmanta.data import EnvironmentMetricsCounter, EnvironmentMetricsNonCounter
+from inmanta.server import SLICE_DATABASE, SLICE_ENVIRONMENT_METRICS, SLICE_TRANSPORT, protocol
+
+LOGGER = logging.getLogger(__name__)
+
+
+class MetricType(str, Enum):
+    count = "count"
+    non_count = "non_count"
+    compile_rate = "compile_rate"
```
arnaudsjs, line 33: In python the convention is to use capital letter for the names of the venv instances.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def get_dependencies(self) -> List[str]:
+        return [SLICE_DATABASE]
+
+    def get_depended_by(self) -> List[str]:
+        return [SLICE_TRANSPORT]
+
+    async def start(self) -> None:
+        await super().start()
+        self.schedule(self.flush_metrics, 60, initial_delay=0, cancel_on_stop=True)
```
arnaudsjs, line 85: Let's define a constant for the collection interval.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def register_metric_collector(self, metrics_collector: MetricsCollector) -> None:
+        """
+        Register the given metrics_collector.
+        """
+        if not any(metric.get_metric_name() == metrics_collector.get_metric_name() for metric in self.metrics_collectors):
+            self.metrics_collectors.append(metrics_collector)
```
arnaudsjs, line 92: Let's use a data structure that can perform this operation in constant time. Like a dictionary.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,120 @@
...
+        else:
+            LOGGER.warning(f"There already is a metric collector with the name {metrics_collector.get_metric_name()}")
```
arnaudsjs, line 94: This case should never happen. Let's raise an exception here instead.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,120 @@
...
+            if metric_type == MetricType.non_count:
+                metric_non_count.append(
+                    EnvironmentMetricsNonCounter.new(metric_name, now, metric_value["count"], metric_value["value"])
+                )
+
+        await EnvironmentMetricsCounter.insert_many(metric_count)
+        await EnvironmentMetricsNonCounter.insert_many(metric_non_count)
+
+        if datetime.now() - now > timedelta(seconds=60):
+            LOGGER.warning("flush_metrics method took more than 1 minute")
```
arnaudsjs, line 120: I would hard code the 60 second here, but use the constant that defines the collection interval.

tests/db/migration_tests/test_v202209160_to_v202212010.py
```diff
@@ -0,0 +1,65 @@
...
+
+@pytest.fixture
+async def migrate_v202209160_to_v202212010(
+    hard_clean_db, hard_clean_db_post, postgresql_client: Connection, server_config
+) -> AsyncIterator[Callable[[], Awaitable[None]]]:
+    """
+    Returns a callable that performs a v202209160 database restore and migrates to v202212010.
+    """
+    # Get old tables
+    with open(os.path.join(os.path.dirname(__file__), "dumps/v202209160.sql"), "r") as fh:
```
arnaudsjs, line 36: Why start from `v202209160`  there is a newer version already `v202211230`?

tests/db/migration_tests/test_v202209160_to_v202212010.py
```diff
@@ -0,0 +1,65 @@
...
+    get_tables_in_db: Callable[[], Awaitable[List[str]]],
+) -> None:
+    """
+    Test whether the environment_metrics_counter table was added
+    """
+
+    # The table is not present before the migration
+    tables = await get_tables_in_db()
+    assert "environmentmetricscounter" not in tables
+    assert "environmentmetricsnon_counter" not in tables
```
arnaudsjs, line 57:
```suggestion
    assert "environmentmetricsnoncounter" not in tables
```

tests/server/test_environement_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+import pytest
+
+from inmanta import data
+from inmanta.server.protocol import Server
+from inmanta.server.services.environment_metrics_service import EnvironmentMetricsService, MetricsCollector, MetricType
+from utils import log_contains
+
+
+@pytest.fixture
+async def env_metrics_service(server_config, init_dataclasses_and_load_schema):
```
arnaudsjs, line 30:
```suggestion
async def env_metrics_service(server_config, init_dataclasses_and_load_schema) -> EnvironmentMetricsService:
```

tests/server/test_environement_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class DummyCountMetric(MetricsCollector):
+    def get_metric_name(self) -> str:
+        return "dummy_count"
+
+    def get_metric_type(self) -> MetricType:
+        return MetricType.count
+
+    async def get_metric_value(self, start_interval: datetime, end_interval: datetime) -> object:
```
arnaudsjs, line 42: Why is the return type object?

tests/server/test_environement_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class DummyNonCountMetric(MetricsCollector):
+    def get_metric_name(self) -> str:
+        return "dummy_non_count"
+
+    def get_metric_type(self) -> MetricType:
+        return MetricType.non_count
+
+    async def get_metric_value(self, start_interval: datetime, end_interval: datetime) -> object:
```
arnaudsjs, line 53: Why is the return type object?

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
FloLey: it is used here: https://github.com/inmanta/inmanta-core/pull/5198/files#diff-8abb50042fc34e38c278ec05a3fc85edfdb092058a78b62a8a2796ddbd6bb708R110

arnaudsjs: You can just create an instance of EnvironmentMetricsCounter using keyword arguments. For example:

```
EnvironmentMetricsCounter(metric_name="test", timestamp=datetime.datetime.now(), count=1)
```

src/inmanta/server/services/environment_metrics_service.py
FloLey: This is not respected at all in the codebase :p

arnaudsjs: Indeed :p

tests/db/migration_tests/test_v202209160_to_v202212010.py
FloLey: I don't see this dump?

arnaudsjs (changes requested):

arnaudsjs (changes requested):

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        now: datetime = datetime.now()
+        metric_count: List[EnvironmentMetricsCounter] = []
+        metric_non_count: List[EnvironmentMetricsNonCounter] = []
+        for mc in self.metrics_collectors:
+            metrics_collector: MetricsCollector = self.metrics_collectors[mc]
+            metric_name: str = metrics_collector.get_metric_name()
+            metric_type: str = metrics_collector.get_metric_type()
+            metric_value: dict[str, int] = await metrics_collector.get_metric_value(self.previous_timestamp, now)
+            self.previous_timestamp = now
```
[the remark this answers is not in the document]
FloLey: but I need it to not be bumped when the call to `metrics_collector.get_metric_value(self.previous_timestamp, now)` is done. 
nevertheless I just noticed that there is a mistake here, it should be bumped outside of the loop.

src/inmanta/db/versions/v202212010.py
```diff
@@ -0,0 +1,40 @@
...
+        timestamp TIMESTAMP NOT NULL,
+        count INT NOT NULL,
+        PRIMARY KEY (metric_name, timestamp)
+    );
+
+    CREATE TABLE IF NOT EXISTS public.environmentmetricsnoncounter (
+        metric_name VARCHAR NOT NULL,
+        timestamp TIMESTAMP NOT NULL,
+        count INT NOT NULL,
+        value FLOAT NOT NULL,
```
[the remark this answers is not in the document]
FloLey: I thought we said both were fine?

src/inmanta/server/services/environment_metrics_service.py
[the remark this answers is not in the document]
arnaudsjs: You can also use a try-finally construction for example.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,130 @@
...
+        self.schedule(self.flush_metrics, COLLECTION_INTERVAL, initial_delay=0, cancel_on_stop=True)
+
+    def register_metric_collector(self, metrics_collector: MetricsCollector) -> None:
+        """
+        Register the given metrics_collector.
+        """
+        if not metrics_collector.get_metric_name() in self.metrics_collectors:
+            self.metrics_collectors[metrics_collector.get_metric_name()] = metrics_collector
+        else:
+            raise ServerError(f"There already is a metric collector with the name {metrics_collector.get_metric_name()}")
```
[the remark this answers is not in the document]
FloLey: I didn't really now which one to use. I changed it to a simple Exception for now. Is that ok?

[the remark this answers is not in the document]
arnaudsjs: Yes. Fine by me.

src/inmanta/db/versions/v202212010.py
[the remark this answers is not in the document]
wouterdb: both are fine for me

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,144 @@
...
+class MetricType(str, Enum):
+    """
+    There are 3 types of metrics: metrics that represent a counter, metrics that doesn't represent a counter and
+    the compile_rate metric which is a special case. An examples of the count type metric is the agent_count metric which counts
+    the number of agents at a timestamp. An example of a non_count metric is the compile_time metric. This metric counts the
+    numer of compiles between the current timestamp and the last one and the total compile time.
+    The compile_rate metric is a special case as it is being calculated using only the count field of the compile_time metric.
+    """
+
+    COUNT = "count"
```
wouterdb, line 42: I think gauge, timer and rate (which is derived from timer)  are the proper names, if we follow the terminology of https://github.com/omergertel/pyformance (which we already use)

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,144 @@
...
+                        EnvironmentMetricsCounter(metric_name=metric_name, timestamp=now, count=metric_value["count"])
+                    )
+                elif metric_type == MetricType.NON_COUNT:
+                    metric_non_count.append(
+                        EnvironmentMetricsNonCounter(
+                            metric_name=metric_name, timestamp=now, count=metric_value["count"], value=metric_value["value"]
+                        )
+                    )
+                elif metric_type == MetricType.COMPILE_RATE:
+                    raise NotImplementedError()
```
wouterdb, line 134: I think this can just say that `COMPILE_RATE` is derived quantity and can not be the type of a reporter

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,144 @@
...
+        raise NotImplementedError()
+
+
+class EnvironmentMetricsService(protocol.ServerSlice):
+    """Slice for the management of metrics"""
+
+    def __init__(self) -> None:
+        super(EnvironmentMetricsService, self).__init__(SLICE_ENVIRONMENT_METRICS)
+        self.metrics_collectors: Dict[str, MetricsCollector] = {}
+        self.previous_timestamp = datetime.now()
```
wouterdb, line 87: this is not, strictly speaking, correct. 

the previous timestamp is the last one available in the DB. 

This may work, but it is tricky.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,144 @@
...
+                else:
+                    raise Exception("Metric type {metric_type.value} is unknown.")
+        finally:
+            self.previous_timestamp = now
+
+        await EnvironmentMetricsCounter.insert_many(metric_count)
+        await EnvironmentMetricsNonCounter.insert_many(metric_non_count)
+
+        if datetime.now() - now > timedelta(seconds=COLLECTION_INTERVAL_IN_SEC):
+            LOGGER.warning("flush_metrics method took more than 1 minute")
```
wouterdb, line 144: I would also explain what that means (i.e. what can happen if this is true) and what can be done about it (e.g. verify the load on the database and the available connection pool size)

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,144 @@
...
+        else:
+            raise Exception(f"There already is a metric collector with the name {metrics_collector.get_metric_name()}")
+
+    async def flush_metrics(self) -> None:
+        """
+        Invoked at the end of the metrics collection interval. Writes the metrics
+        collected by the MetricsCollectors in the past metrics collection interval
+        to the database.
+        """
+        now: datetime = datetime.now()
```
wouterdb, line 114: it might be required to use a single database connection for everything here, to prevent database pool starvation. 

Currently, this implies passing the connection as an argument everywhere.

wouterdb (changes requested):

src/inmanta/server/services/environment_metrics_service.py
FloLey: isn't the rate called meter in this terminology?

wouterdb: well, if it is the only thing you measure: yes

src/inmanta/server/services/environment_metrics_service.py
wouterdb: I think this can work as well, we just have to be a bit careful

wouterdb (approved):

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,181 @@
...
+            for mc in self.metrics_collectors:
+                metrics_collector: MetricsCollector = self.metrics_collectors[mc]
+                metric_type: str = metrics_collector.get_metric_type()
+                metric_values: Sequence[MetricValue] = await metrics_collector.get_metric_value(
+                    old_previous_timestamp, now, connection=con
+                )
+                if metric_type == MetricType.GAUGE:
+                    create_metrics_gauge(metric_values, now)
+                elif metric_type == MetricType.TIMER:
+                    create_metrics_timer(metric_values, now)
```
FloLey, line 165: @sanderr it is here that I'm having trouble with the typing: as reminder `metric_values` is of type `Sequence[MetricValue]`. But once we get here it should always be of type `Sequence[MetricValueTimer]` and I'd like to enforce that.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,181 @@
...
+
+            await EnvironmentMetricsGauge.insert_many(metric_gauge, connection=con)
+            await EnvironmentMetricsTimer.insert_many(metric_timer, connection=con)
+
+        if datetime.now() - now > timedelta(seconds=COLLECTION_INTERVAL_IN_SEC):
+            LOGGER.warning(
+                f"flush_metrics method took more than {COLLECTION_INTERVAL_IN_SEC} seconds: "
+                f"new attempts to flush metrics are fired faster than they resolve."
+                f"Verify the load on the Database and the available connection pool size."
+            )
```
arnaudsjs, line 181:
```suggestion
            LOGGER.warning(
                f"flush_metrics method took more than %d seconds: "
                f"new attempts to flush metrics are fired faster than they resolve. "
                f"Verify the load on the Database and the available connection pool size.",
                COLLECTION_INTERVAL_IN_SEC,
            )
```

tests/db/migration_tests/test_v202211230_to_v202212010.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Returns a callable that performs a v202211230 database restore and migrates to v202212010.
+    """
+    # Get old tables
+    with open(os.path.join(os.path.dirname(__file__), "dumps/v202211230.sql"), "r") as fh:
+        await PGRestore(fh.readlines(), postgresql_client).run()
+
+    ibl = InmantaBootloader()
+
+    # When the bootloader is started, it also executes the migration to v202203140
```
arnaudsjs, line 41:
```suggestion
    # When the bootloader is started, it also executes the migration to v202212010
```

arnaudsjs (changes requested):

arnaudsjs (approved):

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -0,0 +1,199 @@
...
+        if METRIC_NAME_SEPARATOR in metric_name:
+            raise Exception(
+                'The value "%s" can not be used in the metric_name (%s) as it is used as separator'
+                % (METRIC_NAME_SEPARATOR, metric_name)
+            )
+        if grouped_by and METRIC_NAME_SEPARATOR in grouped_by:
+            raise Exception(
+                'The value "%s" can not be used in the grouped_by value (%s) as it is used as separator'
+                % (METRIC_NAME_SEPARATOR, grouped_by)
+            )
```
arnaudsjs, line 71: Very much a detail...
```suggestion
            raise Exception(
                'The character "%s" can not be used in the metric_name (%s)'
                % (METRIC_NAME_SEPARATOR, metric_name)
            )
        if grouped_by and METRIC_NAME_SEPARATOR in grouped_by:
            raise Exception(
                'The character "%s" can not be used in the grouped_by value (%s)'
                % (METRIC_NAME_SEPARATOR, grouped_by)
            )
```

wouterdb (approved):

arnaudsjs (approved):

PR #5212, opened 2022-12-07
improved readability of dataview resource query

sanderr:
# Description
Improved the readability of this recursive query. I did not change anything about its behavior, just restructured it a bit and added some comments.

# Self Check:

- [ ] Changelog entry


tests/conftest.py
```diff
@@ -232,6 +232,19 @@ def postgres_db(request: pytest.FixtureRequest):
     yield request.getfixturevalue(fixture)
 
 
+@pytest.fixture
+async def postgres_db_debug(postgres_db, database_name) -> abc.AsyncIterator[None]:
```
sanderr, line 236: I used this to interact with the database in the context of the test case I was debugging. I think it could be useful to keep it and refer to it from our internal docs. Wdyt?

sanderr: @arnaudsjs rerequesting review for this change. I'm creating a follow-up ticket for the performance fix right now.

arnaudsjs (approved):

Hugo-Inmanta (approved):

PR #5213, opened 2022-12-07
Fix server-side compiles with conflicting SSL config in .inmanta file

### LINKED ISSUE #4640: Compile's started on the server should get the SSL value of the server
Currently, if a `.inmanta` file is present with SSL enabled, the server doesn't correctly override this setting when starting the compiler.

arnaudsjs:
# Description

Fix issue where server-side compiles fail when the SSL configuration on the server doesn't match the SSL configuration defined in the .inmanta file of the project.

closes #4640

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -491,7 +491,7 @@ def export_parser_config(parser: argparse.ArgumentParser) -> None:
     parser.add_argument("--server_address", dest="server", help="The address of the server to submit the model to")
     parser.add_argument("--server_port", dest="port", help="The port of the server to submit the model to")
     parser.add_argument("--token", dest="token", help="The token to auth to the server")
-    parser.add_argument("--ssl", help="Enable SSL", action="store_true", default=False)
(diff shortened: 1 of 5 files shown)


src/inmanta/app.py
```diff
@@ -585,6 +585,8 @@ def export(options: argparse.Namespace) -> None:
 
     if options.ssl:
         Config.set("compiler_rest_transport", "ssl", "true")
+    else:
+        Config.set("compiler_rest_transport", "ssl", "false")
```
wouterdb, line 589: this is not OK, it will make it impossible to start the server with ssl 

if has to be a three way option: yes, no, unset.

wouterdb (changes requested):

src/inmanta/app.py
arnaudsjs: Oh. How did I miss that? Thanks!

wouterdb (approved):

FloLey (approved):

PR #5227, opened 2022-12-08
Issue/5128 add resource count metric

FloLey:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/server/services/environment_metrics_service.py
+++ b/src/inmanta/server/services/environment_metrics_service.py
@@ -207,5 +207,13 @@ class ResourceCountMetricsCollector(MetricsCollector):
     async def get_metric_value(
         self, start_interval: datetime, end_interval: datetime, connection: Optional[asyncpg.connection.Connection]
     ) -> Sequence[MetricValue]:
-        agent_counts_by_state = await data.Resource.get_list()
-        print(agent_counts_by_state)
+        query = """
+            SELECT status,count(*)
+            FROM resource
+            GROUP BY status
+        """
+        metric_values: List[MetricValue] = []
+        result = await connection.fetch(query)
+        for record in result:
+            metric_values.append(MetricValue(self.get_metric_name(), record["count"], record["status"]))
+        return metric_values
--- a/src/inmanta/server/services/environment_metrics_service.py
+++ b/src/inmanta/server/services/environment_metrics_service.py
@@ -24,6 +24,7 @@ from typing import Dict, List, Optional
 
 import asyncpg
 
+from inmanta import data
 from inmanta.data import EnvironmentMetricsGauge, EnvironmentMetricsTimer
 from inmanta.server import SLICE_DATABASE, SLICE_ENVIRONMENT_METRICS, SLICE_TRANSPORT, protocol
 
@@ -194,3 +195,17 @@ class EnvironmentMetricsService(protocol.ServerSlice):
                 "Verify the load on the Database and the available connection pool size.",
                 COLLECTION_INTERVAL_IN_SEC,
             )
+
+
+class ResourceCountMetricsCollector(MetricsCollector):
+    def get_metric_name(self) -> str:
+        return "resource_count"
+
+    def get_metric_type(self) -> MetricType:
+        return MetricType.GAUGE
+
+    async def get_metric_value(
+        self, start_interval: datetime, end_interval: datetime, connection: Optional[asyncpg.connection.Connection]
+    ) -> Sequence[MetricValue]:
+        agent_counts_by_state = await data.Resource.get_list()
+        print(agent_counts_by_state)
--- a/tests/server/test_environement_metrics.py
+++ b/tests/server/test_environement_metrics.py
@@ -33,7 +33,7 @@ from inmanta.server.services.environment_metrics_service import (
     MetricValue,
     MetricValueTimer,
 )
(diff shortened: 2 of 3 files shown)


src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -194,3 +195,29 @@ def create_metrics_timer(metric_values_timer: Sequence[MetricValueTimer], timest
...
+        return "resource_count"
+
+    def get_metric_type(self) -> MetricType:
+        return MetricType.GAUGE
+
+    async def get_metric_value(
+        self, start_interval: datetime, end_interval: datetime, connection: Optional[asyncpg.connection.Connection]
+    ) -> Sequence[MetricValue]:
+        query: str = """
+            SELECT status,count(*)
```
wouterdb, line 215: 1. I believe all resource have to be collected per environment (oops, missed that before)
2. this counts all resources in the DB, which is useless. I think you want to know the resources in the current, active version.

wouterdb (changes requested):

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -194,3 +195,29 @@ def create_metrics_timer(metric_values_timer: Sequence[MetricValueTimer], timest
...
+            FROM resource
+            GROUP BY status
+        """
+        metric_values: List[MetricValue] = []
+        result: Sequence[asyncpg.Record] = await connection.fetch(query)
```
arnaudsjs, line 220: The type annotation for connection is optional, but no is None check is done here. I think we can just make it a mandatory argument. Do you also run mypy on this?

tests/server/test_environement_metrics.py
```diff
@@ -279,3 +282,86 @@ async def test_flush_metrics_mix(env_metrics_service):
...
+        [],
+        {},
+    )
+
+    assert result.code == 200
+
+    # flush the metrics for the second time: 1 old record (3 resources in available state)
+    # + 2 new records (2 resources in available state, 1 in the deployed state)
+    await metrics_service.flush_metrics()
+    result_gauge = await data.EnvironmentMetricsGauge.get_list()
```
arnaudsjs, line 358: If Postgres ever decides the change the order in which they return objects, this test case will break.

tests/server/test_environement_metrics.py
```diff
@@ -279,3 +282,86 @@ async def test_flush_metrics_mix(env_metrics_service):
...
+            "key": "key3",
+            "value": "value3",
+            "id": "test::Resource[agent1,key=key3],v=" + version,
+            "send_event": False,
+            "requires": [],
+            "purged": True,
+        },
+    ]
+
+    result = await client.put_version(
```
arnaudsjs, line 317: I would also add some resources to a different environment to cover the bug that Wouter noticed.

arnaudsjs (changes requested):

src/inmanta/server/services/environment_metrics_service.py
FloLey: @Hugo-Inmanta I see you have taken a ticket about the metrics. To fix the issue about the environment I will add a column to the tables containing the environment_id. This will affect you I think.

Hugo-Inmanta: Isn't it possible to add to the query "WHERE environment = <active_env>" ?

FloLey: How would you do the aggregation afterwards? without knowing which data comes from which env?

Hugo-Inmanta: Ah, I didn't realize we would need to aggregate on different envs at the same time :+1:

wouterdb (approved):

tests/server/test_environement_metrics.py
```diff
@@ -339,3 +342,262 @@ async def get_metric_value(
...
+        tid=env_uuid1,
+        version=version1,
+        resources=resources_env1_v1,
+        unknowns=[],
+        version_info={},
+        compiler_version=get_compiler_version(),
+    )
+    assert result.code == 200
+    version2 = str(await ClientHelper(client, env_uuid1).get_version())
+    resources_env2 = [
```
arnaudsjs, line 572:
```suggestion
    resources_env1_v2 = [
```

tests/server/test_environement_metrics.py
```diff
@@ -339,3 +342,262 @@ async def get_metric_value(
...
+    assert any(x.count == 1 and x.metric_name == "resource_count.deployed" and x.environment == env_uuid1 for x in result_gauge)
+
+    env_uuid2_records = [
+        r for r in result_gauge if r.environment == env_uuid2 and r.metric_name == "resource_count.available" and r.count == 2
+    ]
+
+    assert len(env_uuid2_records) == 2
+
+
+async def test_resource_count_metric_released(clienthelper, client, agent):
```
arnaudsjs, line 518: This is implied by using the client fixture, but I like it when this is made explicit.

```suggestion
async def test_resource_count_metric_released(clienthelper, client, server, agent):
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #5236, opened 2022-12-09
Fix: add support for collecting metrics by environment

### LINKED ISSUE #5124: Add support to write collect metrics to the database

FloLey:
# Description

the possibility to register metrics by environment was not implemented in https://github.com/inmanta/inmanta-core/pull/5224 but should have been added. This PR fixes this.

closes #5124

# Self Check:

- [x] Changelog entry


src/inmanta/db/versions/v202212090.py
```diff
@@ -0,0 +1,34 @@
...
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    await connection.execute(
+        """
+-- adds the environment to which a metric is related
+ALTER TABLE public.environmentmetricsgauge ADD COLUMN environment uuid NOT NULL REFERENCES environment(id) ON DELETE CASCADE;
```
wouterdb, line 26: did we already release the previous version? if we didn't I would just update that one?

wouterdb (changes requested):

src/inmanta/db/versions/v202212090.py
FloLey: Indeed, that would have been easier, it is changed.

wouterdb (approved):

PR #5237, opened 2022-12-12
Fix: passenv in tox.ini file doesn't accept whitespaces

FloLey:
# Description

Some rules changed with the release of the new version of tox. One of those is the separator for the pass_env variable:
https://tox.wiki/en/4.0.3/faq.html#tox-4-changed-ini-rules

this commit fixes the issue

# Self Check:

- [x] Changelog entry


changelogs/unreleased/fix-tox-issue-passenv
```diff
@@ -0,0 +1,7 @@
+description: Fix issue with passenv in tox.ini file
+change-type: patch
+destination-branches:
+  - master
+  - iso6
+  - iso5
+  - iso4
```
sanderr, line 7:
```suggestion
```
iso4 does not get dependency updates so it's still on the old tox.

sanderr (changes requested):
I was investigating this as well, it's really confusing that the Jenkins output said `pass_env` while it's actually `passenv` so I didn't immediately find the issue.

sanderr (approved):

changelogs/unreleased/fix-tox-issue-passenv
```diff
@@ -0,0 +1,6 @@
+description: Fix issue with passenv in tox.ini file
+change-type: patch
+destination-branches:
+  - master
+  - iso6
+  - iso5
```
FloLey, line 6:
```suggestion
  - iso5
  - iso4
```

PR #5259, opened 2022-12-14
Issue/5091 using a bad escape sequence in the model creates a python warning

### LINKED ISSUE #5091: Using a bad escape sequence in the model creates a python warning
The connect model uses a bad escape sequence. However, we only get a warning with the line in the compiler where python detects this:

```
  /home/bart/.virtualenvs/connect5/lib64/python3.9/site-packages/inmanta/parser/plyInmantaLex.py:135: DeprecationWarning: invalid escape sequence '\.'
    value = bytes(t.value[3:-3], "utf-8").decode("unicode_escape")
```

# Proposal

```
try:
    value = bytes(t.value[3:-3], "utf-8").decode("unicode_escape", "strict")
except UnicodeDecodeError:
    # produce warning here
    value = bytes(t.value[3:-3], "utf-8").decode("unicode_escape")

```

Hugo-Inmanta:
closes #5091 

# Description

 A proper inmanta warning is now displayed when an invalid escape sequence is detected in a regular string or a multi-line string.
Tested on the following model:

```py
s1 = r"No warnings in raw strings: \."
s2 = 'Warnings in standard strings: \.'
s3 = "Warnings in standard strings: \."
s4 = """l1
Warnings in MLS:
Bad escape sequence: \.
"""

std::print(s1)
```
Displaying the following warnings:
```term
inmanta.warnings         WARNING ParserWarning: (./main.cf:2) Invalid escape sequence in string:
                                 Warnings in standard strings: \.
inmanta.warnings         WARNING ParserWarning: (./main.cf:3) Invalid escape sequence in string:
                                 Warnings in standard strings: \.
inmanta.warnings         WARNING ParserWarning: (./main.cf:4) Invalid escape sequence in multi-line string:
                                 l1
                                 Warnings in MLS:
                                 Bad escape sequence: \. 
                                 
```

Also fix the display of the warning for the deprecation of v1 modules:

Was previously:
```term
py.warnings              WARNING /home/hugo/work/inmanta/github-repos/inmanta-core/src/inmanta/module.py:2337: DeprecationWarning: Loaded V1 module std. The use of V1 modules is deprecated. Use the equivalent V2 module instead.
                                   warnings.warn(
```
Now:

```term
inmanta.warnings         WARNING InmantaWarning: Loaded V1 module std. The use of V1 modules is deprecated. Use the equivalent V2 module instead.
```

# Self Check:

- [x] Changelog entry


src/inmanta/parser/plyInmantaLex.py
```diff
@@ -162,7 +181,19 @@ def t_INT(t: lex.LexToken) -> lex.LexToken:  # noqa: N802
...
+        with warnings.catch_warnings():
+            warnings.filterwarnings("error", message="invalid escape sequence")
+            value: str = bytes(t.value[1:-1], "utf-8").decode("unicode_escape")
+    except DeprecationWarning:
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore")
+            value = bytes(typing.cast(str, t.value)[1:-1], "utf_8").decode("unicode_escape")
+
+        msg: str = f"({t.lexer.inmfile}:{t.lexer.lineno}) Invalid escape sequence in string:\n{value}"
+        warnings.showwarning(msg, lineno=t.lexer.lineno, filename=t.lexer.inmfile, category=ParserWarning)
```
Hugo-Inmanta, line 194: Should I factor this and the MLS code into a function ? Would that add a performance overhead ?

Hugo-Inmanta:
I tested the changes manually with a model as stated in the description. When I tried to use the snippetcompiler e.g. 
```py
        r'''
test_string_1 = """trippel hel\.lo\nworld"""
'''
```
It seems to me that the first backslash gets escaped somehow somewhere by the time it reaches the parser, and so I couldn't test from within pytest.

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -132,7 +134,24 @@ def t_JCOMMENT(t: lex.LexToken) -> None:  # noqa: N802
...
-    value = bytes(t.value[3:-3], "utf-8").decode("unicode_escape")
+
+    #   Check for the presence of an invalid escape sequence (ex: "\.")
+    #   Python < 3.12 raises a DeprecationWarning when encountering an invalid escape sequence
+    #   Python 3.12 will raise a SyntaxWarning
+    #   Future versions will eventually raise a SyntaxError
+    #   see ( https://docs.python.org/3.12/whatsnew/3.12.html#other-language-changes )
+    try:
+        with warnings.catch_warnings():
+            warnings.filterwarnings("error", message="invalid escape sequence")
```
sanderr, line 145: Let's make sure we only raise what we catch. If we still have this code in 3.12 we wouldn't want it to start raising a `SyntaxWarning` as an exception.
```suggestion
            warnings.filterwarnings("error", category=DeprecationWarning, message="invalid escape sequence")
```

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -132,7 +134,24 @@ def t_JCOMMENT(t: lex.LexToken) -> None:  # noqa: N802
...
 def t_MLS(t: lex.LexToken) -> lex.LexToken:
     r'"{3,5}([\s\S]*?)"{3,5}'
-    value = bytes(t.value[3:-3], "utf-8").decode("unicode_escape")
+
+    #   Check for the presence of an invalid escape sequence (ex: "\.")
+    #   Python < 3.12 raises a DeprecationWarning when encountering an invalid escape sequence
+    #   Python 3.12 will raise a SyntaxWarning
+    #   Future versions will eventually raise a SyntaxError
+    #   see ( https://docs.python.org/3.12/whatsnew/3.12.html#other-language-changes )
+    try:
```
sanderr, line 143: I think some small comments as to what is going on would really help for this block.

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -132,7 +134,24 @@ def t_JCOMMENT(t: lex.LexToken) -> None:  # noqa: N802
...
+        with warnings.catch_warnings():
+            warnings.filterwarnings("error", message="invalid escape sequence")
+            value: str = bytes(t.value[3:-3], "utf_8").decode("unicode_escape")
+    except DeprecationWarning:
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore")
+            value = bytes(typing.cast(str, t.value)[3:-3], "utf_8").decode("unicode_escape")
+
+        msg: str = f"({t.lexer.inmfile}:{t.lexer.lineno}) Invalid escape sequence in multi-line string:\n{value}"
+        warnings.showwarning(msg, lineno=t.lexer.lineno, filename=t.lexer.inmfile, category=ParserWarning)
```
sanderr, line 153: Why use the low-level `showwarning` rather than `warnings.warn(ParserWarning(...))`?

src/inmanta/parser/plyInmantaLex.py
sanderr: If you can make it a clean function I would indeed go for that. I don't think the slight performance overhead should be a concern during parsing. @wouterdb cc.

sanderr (changes requested):
Nice! About the testing, it would be a plus if we could add a test case. I'll have a quick look at the problem you describe.

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -132,7 +134,24 @@ def t_JCOMMENT(t: lex.LexToken) -> None:  # noqa: N802
...
+    try:
+        with warnings.catch_warnings():
+            warnings.filterwarnings("error", message="invalid escape sequence")
+            value: str = bytes(t.value[3:-3], "utf_8").decode("unicode_escape")
+    except DeprecationWarning:
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore")
+            value = bytes(typing.cast(str, t.value)[3:-3], "utf_8").decode("unicode_escape")
+
+        msg: str = f"({t.lexer.inmfile}:{t.lexer.lineno}) Invalid escape sequence in multi-line string:\n{value}"
```
sanderr, line 152: I wonder if it adds value to have the string as part of the warning message. It can get quite verbose and you already have the line number. But I'm really bad at UX. What do you think?

src/inmanta/parser/plyInmantaLex.py
wouterdb: 1. The function call I  would not worry about.
2. The context manager, I would check what it does to parser performance.

sanderr:
@Hugo-Inmanta I don't seem to have any issues with the test case you gave as an example. When I run the test case below with `pytest tests/compiler/test_warnings.py::test_test --log-cli-level=DEBUG` I do see the warning in the output.

```python
def test_test(snippetcompiler):
        r'''
           test_string_1 = """trippel hel\.lo\nworld"""
        '''
    )
    compiler.do_compile()
```

```
10:56:50.742 WARNING ParserWarning: (/tmp/tmpncfmfc8s/main.cf:2) Invalid escape sequence in multi-line string:
trippel hel\.lo
world
```

Hugo-Inmanta:
> 

Ah indeed it is there in CLI with --log-cli-level=DEBUG option.

src/inmanta/parser/plyInmantaLex.py
Hugo-Inmanta: Makes sense

src/inmanta/parser/plyInmantaLex.py
Hugo-Inmanta: I have to agree that it will get ugly for long strings.

src/inmanta/parser/plyInmantaLex.py
Hugo-Inmanta: I had 2 issues:
- I couldn't find a clean way to get the file name and line number in there using `Range()` but looking at it again, I think I can use `Location()`
- As it is a subclass of an `InmantaWarning`, the value itself wasn't properly displayed, but I agree with your other comment that it's preferable not to show the string in the warning message.

Conclusion -> I'll use `warnings.warn(ParserWarning(...))`

src/inmanta/parser/plyInmantaLex.py
Hugo-Inmanta: Here are the results of the performance benchmark I did:
Model |Master | Feature Branch |
|-----|-----|---|
| Small model with 3 bad escapes |  54.8910 ms  | 57.0080 ms 
| Bigger model with 12 bad escapes |  56.7250 ms  | 59.9952 ms 
| Bigger model with no bad escapes |  54.7620 ms  | 55.1883 ms 

I will paste the full results on the main thread.

src/inmanta/parser/__init__.py
```diff
@@ -46,7 +46,7 @@ def export(self) -> ast_export.Error:
 class ParserWarning(InmantaWarning, ParserException):
     """Warning occurring during the parsing of the code"""
 
-    def __init__(self, location: Range, value: object, msg: str) -> None:
+    def __init__(self, location: Location, value: object, msg: str) -> None:
```
Hugo-Inmanta, line 49: This is fine right ? Rationale is that in the safe_decode function, I want to raise a `ParserWarning` but I only have access to enough info for a `Location` (i.e. file name and line number) and not enough for a `Range` (i.e. start and end positions).
As far as I can tell, the argument is only ever used as a `Location`. Am I missing anything ?

src/inmanta/parser/plyInmantaLex.py
Hugo-Inmanta: @wouterdb wdyt ?

src/inmanta/parser/__init__.py
sanderr: I think you're right.

sanderr:
Out of curiosity, how did you generate this benchmark output? I assume you didn't create the tables yourself?

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -246,3 +221,40 @@ def t_ANY_error(t: lex.LexToken) -> lex.LexToken:  # noqa: N802
...
+    try:
+        # This first block will try to decode the value and turn any deprecation warning into an actual Exception.
+        with warnings.catch_warnings():
+            warnings.filterwarnings("error", message="invalid escape sequence", category=DeprecationWarning)
+            value: str = bytes(typing.cast(str, token.value)[start:end], "utf_8").decode("unicode_escape")
+    except DeprecationWarning:
+        # If the first block did actually encounter an invalid escape sequence, we have to decode the value again, this time
+        # ignoring the python warning, and raising a warning of our own.
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message="invalid escape sequence", category=DeprecationWarning)
```
sanderr, line 253: I think you actually got this right before: non-deprecation warnings would have already been emitted (we didn't catch them) so we can safely ignore any warning here.
```suggestion
        # ignoring this or any other python warning that has already been emitted, and raising a warning of our own.
        with warnings.catch_warnings():
            warnings.filterwarnings("ignore")
```

sanderr (changes requested):

src/inmanta/parser/plyInmantaLex.py
Hugo-Inmanta: Ah yes indeed :+1:

Hugo-Inmanta:
> Out of curiosity, how did you generate this benchmark output? I assume you didn't create the tables yourself?

I used [pytest benchmark](https://pypi.org/project/pytest-benchmark/) like so:
```py


def context_manager_performance_benchmark(snippetcompiler):
    snippetcompiler.setup_for_snippet(...)
    compiler.do_compile()


def test_safe_decode(benchmark, snippetcompiler):
    benchmark.pedantic(context_manager_performance_benchmark, args=(snippetcompiler,), iterations=10, rounds=100)

```
And then call it in cli.
I tried the option to create graphs but the result was underwhelming.

sanderr (approved):
pytest-benchmark looks pretty interesting, thanks!

PR #5267, opened 2022-12-15
agent inter module dependencies mark 1

### LINKED ISSUE #5251: Inter module dependencies on the agent mark I
# Goal

Make it possible to have handlers depend on code in other modules.

# Current state of affairs
- When the agent installs a handler, it also installs al its python dependencies
- [except for the ones named `inmanta-module-*`]( https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/loader.py#L109)
-   in modules v1, one handler can not load code from another handler
-    this makes it impossible to build generic base handlers, that are extended from other modules
    -  on v1, this could be bypassed by packing the base handler as python package and adding it to the requirements.txt of the one needing it
    - on v2, because of the filtering, this doesn't work anymore (unless you re-name the package)

# Proposed solution

1. not filter out the depdencies
2. use a loader order where the special code loader (used for handler loading) has priority over the normal one
3. add a feature flag to the project.yml to control this behavior 

This would

1.   cause all inter-module-dependencies to be available as natively installed
2.   cause handler loading to 'overload' the native install (so every handler itself will use the old method)

# Limitations 

1.  may cause trouble if the modules can not actually be pulled from the agents pip index. (covered by the feature flag)
2. may cause issues if the base module contains handlers, they will not get removed when they are no longer used
3. does not fix any of the existing reloading issue, in particular 

    1. stale code would remain (until restart) and be preferred over the fresh one from the pip index.
    2. the module load order is not guaranteed, so the dependency module might be (re)loaded after the one that depends on it, meaning we might always lag one version behind at load-time, which requires a special form of attention when implementing the module
    3. during the first ever load: 
       1. main_module is loaded, imports base from the pip index (because the code for base has not been loaded by the special loader yet.
        2. base is loaded by the special loader: does our current import mechanism just work, or does it need changes to allow loading from a different source (for a module that is already in sys.modules)?

In general, the base module should not contain actual handlers itself. 

# Special attention point

1. loader order is very important here. 

# Context

Some other tickets are related to this, and provide more complete solutions

- https://github.com/inmanta/inmanta-core/issues/3523
- https://github.com/inmanta/inmanta-core/issues/2314

sanderr:
# Description

Adds a project option to install dependencies on other modules when loading code on the agent. Makes sure code loaded by the agent gets precedence on the code installed by pip.

closes #5251

# Self Check:

- [x] Changelog entry
- [x] Add second reviewer

### DIFF
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -25,7 +25,7 @@ import shutil
 import sys
 from collections import abc
 from logging import INFO
-from typing import List, Set
+from typing import List, Optional, Set
 
 import py
 import pytest
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -321,7 +321,7 @@ def test_code_loader_prefer_finder(tmpdir: py.path.local) -> None:
     """
     assert not isinstance(sys.meta_path[0], loader.PluginModuleFinder)
     loader.CodeLoader(code_dir=str(tmpdir))
-    # it suffices to verify that the module finder is firs in the meta path:
+    # it suffices to verify that the module finder is first in the meta path:
     # `test_plugin_module_finder` verifies the actual loader behavior
     assert isinstance(sys.meta_path[0], loader.PluginModuleFinder)
 
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -23,8 +23,8 @@ import inspect
 import os
 import shutil
 import sys
-from collections import abc
 from logging import INFO
+from types import ModuleType
 from typing import List, Optional, Set
 
 import py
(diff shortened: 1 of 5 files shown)


tests/test_loader.py
```diff
@@ -226,6 +241,91 @@ def module_path(tmpdir):
...
+    assert mod.where == "libs" if prefer_finder else "venv"
+
+
+def test_code_loader_prefer_finder(tmpdir: py.path.local) -> None:
+    """
+    Verify that the agent code loader prefers its loaded code over code in the Python venv.
+    """
+    assert not isinstance(sys.meta_path[0], loader.PluginModuleFinder)
+    loader.CodeLoader(code_dir=str(tmpdir))
+    # it suffices to verify that the module finder is firs in the meta path:
```
wouterdb, line 324:
```suggestion
    # it suffices to verify that the module finder is first in the meta path:
```

wouterdb (approved):

Hugo-Inmanta (approved):

PR #5278, opened 2022-12-19
[poc] Unmanaged resource

FloLey:
# Description

this is a Poc for https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/5140

The  endpoint to trigger discovery of resources in called `discover_facts`. 
`discover_facts` will go over all agents in the environment and get the client for each agent.
for each client it will call `discover_resources_client` which will in turn call `discover_resources_instance`
`discover_resources_instance` will go over all the resources of an agent instance and call the `resource_handler.execute_discover_resource ` for each resource. 

the `execute_discover_resource` will call the `pre` function, followed by the `discover_resource` function, followed by the `post` function. If some resources are discovered by `discover_resource`, they are added to the DB using `discovered_resources_create`

Here is an example handler that works with this code:

```
@provider("aws::Subnets", name="ec2")
class SubnetsHandler(AWSHandler):
    def _get_subnets(self):
        return list(
            self._ec2.subnets.all()
        )

    def discover_resource(self, ctx: HandlerContext):
        sns = self._get_subnets()
        result = []
        for sn in sns:
            result.append({
                "vpc_id":str(sn.vpc_id),
                "availability_zone":str(sn.availability_zone),
                "availability_zone_id":str(sn.availability_zone_id),
                "available_ip_address_count":str(sn.available_ip_address_count),
                "cidr_block":str(sn.cidr_block),
                "owner_id":str(sn.owner_id),
                "state":str(sn.state),
                "subnet_arn":str(sn.subnet_arn),
                "subnet_id":str(sn.subnet_id),
                "tags":str(sn.tags),
            })
        return "subnets",result

```

### DIFF
--- a/src/inmanta/agent/handler.py
+++ b/src/inmanta/agent/handler.py
@@ -822,6 +822,22 @@ class ResourceHandler(object):
         except Exception:
             raise Exception("Unable to upload file to the server.")
 
+    def execute_discover_resource(self, ctx: HandlerContext, resource: resources.Resource):
+        ctx.info("execute_discover_resource")
+        self.pre(ctx, resource)
+        result = self.discover_resource(ctx)
+        if result:
+
+            def call():
+                client = protocol.Client("agent")
+                return client.discovered_resources_create(
+                    env=self._agent.environment, discovered_resource_name=result[0], value=result[1]
+                )
+
+            self.run_sync(call)
+
+        self.post(ctx, resource)
+
 
 @stable_api
 class CRUDHandler(ResourceHandler):
@@ -829,6 +845,13 @@ class CRUDHandler(ResourceHandler):
     This handler base class requires CRUD methods to be implemented: create, read, update and delete. Such a handler
     only works on purgeable resources.
     """
+
+    def discover_resource(self, ctx: HandlerContext) -> None:
+        """
+        :param context: Context can be used to get values discovered in the read method. For example, the id used in API
+                calls. This context should also be used to let the handler know what changes were made to the
+                resource.
+        """
 
     def read_resource(self, ctx: HandlerContext, resource: resources.PurgeableResource) -> None:
         """
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5442,6 +5442,22 @@ class EnvironmentMetricsTimer(BaseDocument):
     __primary_key__ = ("environment", "metric_name", "timestamp")
 
 
+class DiscoveredResources(BaseDocument):
+    """
+    discovered resources
+
+    :param environment: the environment of those resources
+    :param discovered_resource_name: The name of the resources
+    :param values: The values
+    """
+
+    environment: uuid.UUID
+    discovered_resource_name: str
+    values: list[object]
+
+    __primary_key__ = ("environment", "discovered_resource_name", "values")
+
+
 _classes = [
     Project,
     Environment,
@@ -5460,6 +5476,7 @@ _classes = [
     Notification,
     EnvironmentMetricsGauge,
     EnvironmentMetricsTimer,
+    DiscoveredResources,
 ]
 
 
--- a/src/inmanta/db/versions/v202212015.py
+++ b/src/inmanta/db/versions/v202212015.py
@@ -1,0 +1,32 @@
+"""
+    Copyright 2022 Inmanta
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    Contact: code@inmanta.com
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    schema = """
+CREATE TABLE IF NOT EXISTS public.discoveredresources (
+    environment uuid NOT NULL REFERENCES environment(id) ON DELETE CASCADE,
+    discovered_resource_name VARCHAR NOT NULL,
+    values jsonb[] NOT NULL,
+    PRIMARY KEY (environment, discovered_resource_name)
+);
+    """
+
+    await connection.execute(schema)
--- a/src/inmanta/protocol/methods.py
+++ b/src/inmanta/protocol/methods.py
@@ -20,7 +20,7 @@
 
 import datetime
 import uuid
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Set, Union
 
 from inmanta import const, data, resources
 from inmanta.data import model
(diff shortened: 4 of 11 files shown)


src/inmanta/protocol/methods.py
```diff
@@ -480,6 +480,41 @@ def resource_action_update(
...
 
+@typedmethod(
+    path="/discovered_resources",
+    validate_sid=False,
+    operation="POST",
+    agent_server=True,
+    arg_options=ENV_OPTS,
+    client_types=[const.ClientType.agent],
+)
+def discovered_resources_create(env: uuid.UUID, discovered_resource_name: str, value: List[Dict[str, str]]) -> None:
```
wouterdb, line 491: What is the form of this value? Why `List[Dict[str, str]]` ?

src/inmanta/protocol/methods.py
```diff
@@ -480,6 +480,41 @@ def resource_action_update(
...
+    :param env: The id of the environment this resource belongs to
+    :param discovered_resource_name: The resource with the given resource_version_id id from the agent
+    """
+
+
+@typedmethod(
+    path="/discovered_resources",
+    validate_sid=False,
+    operation="GET",
+    agent_server=True,
```
wouterdb, line 504: why is this one set?

src/inmanta/protocol/methods.py
```diff
@@ -480,6 +480,41 @@ def resource_action_update(
...
+    arg_options=ENV_OPTS,
+    client_types=[const.ClientType.agent],
```
wouterdb, line 506: why agent? I would expect this to be api? I suspect this will allow the removal of the two other options above I flagged.

src/inmanta/protocol/methods.py
```diff
@@ -480,6 +480,41 @@ def resource_action_update(
...
+    Send discovered resource to the server
+
+    :param env: The id of the environment this resource belongs to
+    :param discovered_resource_name: The resource with the given resource_version_id id from the agent
+    """
+
+
+@typedmethod(
+    path="/discovered_resources",
+    validate_sid=False,
```
wouterdb, line 502: Why is this one set?

src/inmanta/protocol/methods.py
```diff
@@ -781,9 +816,32 @@ def set_parameters(tid: uuid.UUID, parameters: list):
...
+@method(
+    path="/discover_resources_client",
+    operation="POST",
+    arg_options=AGENT_ENV_OPTS,
+    server_agent=True,
+    client_types=[const.ClientType.agent],
+)
+def discover_resources_client(tid: uuid.UUID, agent_name: str, version: int):
+    """
+    discover all resources for an environment and a client
```
wouterdb, line 837: Wht is a  client in this context?

wouterdb (commented):
Pretty cool. 

Raises some interesting questions. 

I have focused purely on architectural issues, not on e.g. logging, exception handling,....

arnaudsjs (changes requested):

src/inmanta/agent/handler.py
```diff
@@ -822,6 +822,22 @@ def call() -> typing.Awaitable[Result]:
...
 
+    def execute_discover_resource(self, ctx: HandlerContext, resource: resources.Resource):
+        ctx.info("execute_discover_resource")
+        self.pre(ctx, resource)
+        result = self.discover_resource(ctx)
+        if result:
+
+            def call():
+                client = protocol.Client("agent")
+                return client.discovered_resources_create(
```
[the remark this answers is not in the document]
sanderr: To add to this, to me this doesn't feel like it fits well with the current responsibilities of the handler. As it is now, the handler is an expert in its domain, but is nothing more than a slave to the agent in other respects, and it doesn't have side effects outside of its domain or the mutable parameters it receives. e.g. the agent asks the handler to deploy, it does, reports changes via the context, then the agent does all the bookkeeping (report facts, mark deploy done, ...). I think it might be more consistent to have this method report what it finds (possibly via the context?), and leave the processing/reporting to the agent. That way responsibilities are kept decoupled: the handler is the domain expert, the agent knows about the orchestration aspect.

Do you see what I mean? Wdyt?

src/inmanta/data/__init__.py
```diff
@@ -5442,6 +5442,22 @@ class EnvironmentMetricsTimer(BaseDocument):
...
+    discovered resources
+
+    :param environment: the environment of those resources
+    :param discovered_resource_name: The name of the resources
+    :param values: The values
+    """
+
+    environment: uuid.UUID
+    discovered_resource_name: str
+    values: list[object]
```
sanderr, line 5456: Not sure if storing them all in a single row is the best approach here, something to think about in the actual design.

src/inmanta/data/__init__.py
```diff
@@ -5442,6 +5442,22 @@ class EnvironmentMetricsTimer(BaseDocument):
...
+
+    __primary_key__ = ("environment", "discovered_resource_name", "values")
```
sanderr, line 5458: We probably need more than just the name if we want to avoid conflicts between different providers. Unless we use a fully qualified name or the provider's name (because that one is already checked for uniqueness irrc). But even then I think we need to at least be able to differentiate the (potential) agent that discovered it because it's part of a resource's identity, right?

Edge cases we should support (or at least handle gracefully) I can think of off the top of my head:
- two modules that have nothing to do with each other that happen to use the same name for one of their resources
- handlers like the ones for `std::Package`: multiple handler types for a single resource type
- multiple agents discover resources that happen to have the same attributes, e.g. the same vlan interface on different hosts

src/inmanta/db/versions/v202212015.py
```diff
@@ -0,0 +1,32 @@
...
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    schema = """
+CREATE TABLE IF NOT EXISTS public.discoveredresources (
+    environment uuid NOT NULL REFERENCES environment(id) ON DELETE CASCADE,
+    discovered_resource_name VARCHAR NOT NULL,
+    values jsonb[] NOT NULL,
+    PRIMARY KEY (environment, discovered_resource_name)
```
sanderr, line 28: The primary key declared here does not match with the one in `inmanta.data`

src/inmanta/protocol/methods.py
```diff
@@ -781,9 +814,32 @@ def set_parameters(tid: uuid.UUID, parameters: list):
     """
 
 
-# Get parameters from the agent
+@method(path="/discover_facts", operation="GET", arg_options=ENV_OPTS, client_types=[const.ClientType.api])
+def discover_facts(tid: uuid.UUID):
```
sanderr, line 818: Naming typo? facts -> resources

src/inmanta/server/services/resourceservice.py
```diff
@@ -995,3 +996,24 @@ async def versioned_resource_details(
...
+    @handle(methods.discovered_resources_create)
+    async def discovered_resources_create(self, env: data.Environment, discovered_resource_name: str, value: Sequence[dict]):
+        discovered_resource = data.DiscoveredResources(
+            environment=env, discovered_resource_name=discovered_resource_name, values=value
+        )
+        (column_names, values) = discovered_resource._get_column_names_and_values()
+        column_names_as_sql_string = ",".join(column_names)
+        values_as_parameterize_sql_string = "$1,$2,$3"
+        query = f"""INSERT INTO {discovered_resource.table_name()} ({column_names_as_sql_string}) VALUES ({values_as_parameterize_sql_string})
+                ON CONFLICT (environment,discovered_resource_name) DO UPDATE SET values = $3;
```
sanderr, line 1009: We need to be careful here not to overwrite too greedily (as implemented here I think). Related to my comment about the primary key.

sanderr (commented):
Overall very nice, I think it allows us to ask the right concrete questions to help with the final design.

Hugo-Inmanta (approved):

src/inmanta/data/__init__.py
FloLey: @sanderr i'm not sure how to handle the 2 first edge cases. Do you have an Idea?
for the third one I will store the UUID of the agent that discovered the resource in the same record as the resource.

wouterdb: We decided to store them by ResourceID, so that would include the name of the agent

sanderr: Perfect.

PR #5289, opened 2022-12-20
Fix bug that makes the `resource_history` endpoint crash

arnaudsjs:
# Description

Fix bug that makes the `resource_history` endpoint crash, because the requires field of a resource was parsed incorrectly.

# Self Check:

- [x] Changelog entry


FloLey (approved):

src/inmanta/data/__init__.py
```diff
@@ -4463,6 +4463,9 @@ async def get(
 
     @classmethod
     def new(cls, environment: uuid.UUID, resource_version_id: m.ResourceVersionIdStr, **kwargs: Any) -> "Resource":
+        if "attributes" in kwargs and "requires" in kwargs["attributes"]:
```
wouterdb, line 4466: This I don't inderstand?

wouterdb (approved):

src/inmanta/data/__init__.py
arnaudsjs: It's a guard to make sure that the `requires` field in the attributes is always a rid and not and rvid.

PR #5307, opened 2022-12-21
added fix for attribute hash stability

### LINKED ISSUE #5306: Increment calculation can be unstable under different dict orders
# Issue

- Currently, the increment calculation relies on the attribute hash.
- This hash depends on the `__str__` of the values involved. 
- This leads to subtle instabilities, because it makes the insertion order of dictionaries change the hash

# Possible solutions

- update the attribute_hash to use json encoding with ordered keys
- make the json encoder/decoder always order the keys

wouterdb:
# Description

Fixes an unstability in the increment calculation when resources contain dicts. 

closes #5306 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5306-improved-attribute-hash.yml
+++ b/changelogs/unreleased/5306-improved-attribute-hash.yml
@@ -1,0 +1,9 @@
+description: Improve stability of incremental deploy for resources containing dicts
+issue-nr: 5306
+sections:
+  feature: "{{description}"
+  upgrade-note: "The first recompile after this upgrade will always perform a full deploy"
+change-type: minor
+destination-branches:
+  - iso6
+  - master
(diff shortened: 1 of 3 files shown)


bartv (approved):
I like the simplicity of using sort_keys from json dumps. One minor remark: it would be good to document here why you are sorting the keys

arnaudsjs (approved):

changelogs/unreleased/5306-improved-attribute-hash.yml
```diff
@@ -0,0 +1,9 @@
+description: Improve stability of incremental deploy for resources containing dicts
+issue-nr: 5306
+sections:
+  feature: "{{description}"
+  upgrade-note: "The first recompile after this upgrade will always perform a full deploy"
+change-type: minor
+destination-branches:
+  - iso6
+  - master
```
arnaudsjs, line 9: Shouldn't this also go into iso5. iso5 has partial compile as well

arnaudsjs (changes requested):

changelogs/unreleased/5306-improved-attribute-hash.yml
wouterdb: This is a fairly intrusive fix for a fairly exotic problem, so I would say no.

arnaudsjs (approved):

PR #5309, opened 2022-12-22
Update LSM quickstart docs

jptrindade:
# Description

[Move this PR to inmanta core](https://github.com/inmanta/lsm/pull/556)


bartv:
Only review from me is enough because we already looked at the content. The core repo has a slightly different merge procedure that you can find in our internal docs. What you basically still need to do is add a changelog file:

For example something like this: https://github.com/inmanta/inmanta-core/blob/master/changelogs/unreleased/use-new-theme.yml

The only thing you need to change here is the description.

bartv (approved):

PR #5310, opened 2022-12-23
Proof of concept version of a user mgmt service

bartv:
# Description

This is a proof of concept version of having a built-in user mangement service. This has been an itch I have been having for a few years. The goal was to check how easy it would be to add this, turns out not that difficult actually. There are still quite some open question:

- How do we configure this?
- How do we get the initial user in the database?
  - A default user is a big no no
  - Add a command that needs to run on the server
  - Add special mode when there are no users, the frontend will make them create one
- When this is merged, do we enable auth by default?

There are still many todo's of which many are actually a ticket on their own.

I also had make some changes to how we construct queries because user also exists in the non public schema.

**Goals**

- Cleanup
- Make sure all tests succeed
- Get it through review

# Self Check:

- [x] Changelog entry


src/inmanta/user_setup.py
```diff
@@ -0,0 +1,155 @@
...
+    click.echo(f"Server authentication:                                  {click.style('enabled', fg='green')}")
+
+    # make sure the method is set to database
+    if server_config.server_auth_method.get() != "database":
+        raise click.ClickException(
+            "The server authentication method should be set to database to continue. Make sure auth_method in the server "
+            "section is set to database"
+        )
+
+    click.echo(f"Server authentication method:                          {click.style('database', fg='green')}")
```
sanderr, line 63: If your goal here is to create two columns that align for both output lines, Python's built-in string formatting has a functionality that might help: `f"{'hello' : <20}world"` left-aligns "hello" in a space-padded substring of 20 characters, followed by "world". So `hello               world`.

This allows you to create aligned columns without having to count spaces: `print(f"{'hello' : <20}world"); print(f"{'hello there' : <20}world")` gives
```
hello               world
hello there         world
```

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,155 @@
...
+            await data.disconnect()
+
+    click.echo("Make sure to (re)start the orchestrator to activate all changes.")
+
+
+@click.command(help="Do the initial user setup")
+@click.option("--reset", help="Reset the password to recover a lost password", is_flag=True)
+def cmd(reset: bool) -> None:
+    # validate the setup so that we can setup a new user
+    validate_server_setup()
```
sanderr, line 144: What if the database schema is outdated? Do we just expect this command to be run only on an up to date server?

sanderr (commented):
Nice!

src/inmanta/user_setup.py
sanderr: I would be in favor of making an attempt to check if the schema is outdated. `inmanta.data.schema` probably has some methods that could help here, but I'm not sure if it requires the server to be running.

I'd ask the question to the team and see if anyone else has an opinion.

arnaudsjs: I mainly would like a clear error message when the schema would be outdated.

sanderr: I would make it broader than that. What if we make a change to the users table, and consequentially to this command? I would be in favor of aborting the command with a clear error message if there are migration scripts that are not applied to the active schema, because this means that the Python source and the database schema are out of sync.

FloLey: Ow I didn't see @sanderr 's message. Could you point me in a direction on how this can be done?

sanderr: `inmanta.data.schema` is that direction. If you need more I'd need to have a look myself, so let me know. But I think we first need to reach consensus. Let's await @arnaudsjs' response first.

arnaudsjs: Fine by me to make the scope broader.

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # hash the password
+        pw_hash = nacl.pwhash.str(password.encode())
+
+        # insert the user
+        user = data.User(
+            username=username,
+            password_hash=pw_hash.decode(),
+            enabled=True,
+            auth_method="password",
```
arnaudsjs, line 59: Here we specify the auth_method `password`, but the documentation of the `server_auth_method` configuration option mentions that the only two possible values are `oidc` or `database`?

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,110 @@
...
+    @protocol.handle(protocol.methods_v2.list_users)
+    async def list_users(self) -> list[model.User]:
+        return [user.to_dao() for user in await data.User.get_list()]
+
+    @protocol.handle(protocol.methods_v2.add_user)
+    async def add_user(self, username: str, password: str) -> model.User:
+        # check if the user already exists
+        existing = await data.User.get_one(username=username)
+        if existing:
+            raise exceptions.Conflict(f"A user with name {username} already exists.")
```
arnaudsjs, line 49: There is a race condition here if the new user is created between the exists check and insert. In that case the endpoint will crash with a 500 response code.

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,110 @@
...
+
+        if not user.password_hash:
+            raise exceptions.UnauthorizedException()
+
+        try:
+            nacl.pwhash.verify(user.password_hash.encode(), password.encode())
+        except nacl.exceptions.InvalidkeyError:
+            raise exceptions.UnauthorizedException()
+
+        # TODO: set an expire
```
arnaudsjs, line 100: TODO?

arnaudsjs (changes requested):
I didn't make it to the end yet, but here are some comments already.

src/inmanta/data/model.py
```diff
@@ -722,3 +722,19 @@ class EnvironmentMetricsResult(BaseModel):
     end: datetime.datetime
     timestamps: List[datetime.datetime]
     metrics: Dict[str, List[Optional[Union[float, Dict[str, float]]]]]
+
+
+class User(BaseModel):
+    """A user"""
+
+    username: str
+    auth_method: str
```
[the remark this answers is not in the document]
sanderr: I agree with you. This is not free-form so I don't think we should model it as such either. I would actually even extend that to the database. If we want to avoid enums there I would go for the approach Markus suggested for this, iirc he said to create a separate table that contains all valid auth_methods together with a numeric id, and to use that id in this table. Then again, I can't say off the top of my head how well this would fit in our current data model.

sanderr (commented):
Looks like @arnaudsjs requested quite some changes already. If you have no objections I'm going to delay review until the bulk of his comments has been addressed to avoid limit duplicates. Rerequest my review if you want me to look it it now after all or when Arnaud's review has stabilized somewhat.

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,163 @@
...
+from inmanta.server import config as server_config
+
+
+def generate_signing_config() -> str:
+    hostname = socket.gethostname()
+    return f"""[auth_jwt_default]
+algorithm=HS256
+sign=true
+client_types=agent,compiler,api
+key={secrets.token_urlsafe(32)}
```
arnaudsjs, line 37: I am wondering whether it's a good idea to hard code the key length here. The advantages of hard coding the value is that our tests can verify that everything works. On the other hand this means that we have to remember to increment this value over the years to make sure that the key length is still safe. The alternative would be to use the default value of the library. According to the documentation, they will increment the default over time to keep a safe default. I am not sure in which direction we should go here. Any opinions?

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,163 @@
...
+    """Perform the user setup that requires the database interaction"""
+    connection = None
+    try:
+        connection = await get_database_connection()
+        DBschema = schema.DBSchema(CORE_SCHEMA_NAME, PACKAGE_WITH_UPDATE_FILES, connection)
+        schema_up_to_date = await DBschema.is_DB_schema_up_to_date()
+        if schema_up_to_date:
+            click.echo(f"{'DB up to date' : <50}{click.style('yes', fg='green')}")
+        else:
+            click.echo(f"{'DB up to date' : <50}{click.style('no: please migrate your DB to the latest version', fg='red')}")
```
arnaudsjs, line 118:
```suggestion
            click.echo(f"{'DB schema up to date' : <50}{click.style('yes', fg='green')}")
        else:
            click.echo(f"{'DB schema up to date' : <50}{click.style('no: please migrate your DB to the latest version', fg='red')}")
```

arnaudsjs, line 118: This error message is not necessarily correct. It's also possible that the version (software version, not db schema version) of the client running the user setup command is not the same as the version of the server. So the database schema of the server might be up to date while you might still get this error message.

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,163 @@
...
+    click.echo(f"{'Server authentication method: ' : <50}{click.style('database', fg='green')}")
+
+    # make sure there is auth config that supports signing tokens
+    cfg = config.AuthJWTConfig.get_sign_config()
+    if cfg is None:
+        click.echo("Error: No signing config available in the configuration.")
+
+        value = None
+        while value not in ["yes", "no"]:
+            value = click.prompt("Do you want to generate a new configuration? yes/no")
```
arnaudsjs, line 73: * Why do we ask this question actually? Why not always output the proposed config and let the user decide whether he/she wants to use the config or not?
* I also think the question is misleading. When I read it, it sounds like the command will write the config to `/etc/inmanta/inmanta.d/` right away if do say yes. But in fact it only outputs it to stdout, which is a very harmless operation.

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,163 @@
...
+
+        if value == "yes":
+            click.echo("Add the following to the configuration in /etc/inmanta/inmanta.d/auth.cfg:\n")
+            click.echo(generate_signing_config())
+
+        raise click.ClickException("Make sure signing configuration is added to the config. See the documentation for details.")
+
+    click.echo(f"{'Authentication signing config: ' : <50}{click.style('enabled', fg='green')}")
```
arnaudsjs, line 81: `enabled` is not really the right word here for a check that verifies the presence of a config. I am not sure whether `found` is the perfect word here, but it's definitely more accurate than enabled.

```suggestion
    click.echo(f"{'Authentication signing config: ' : <50}{click.style('found', fg='green')}")
```

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,163 @@
...
+        connection_pool_max_size=connection_pool_max_size,
+        connection_timeout=connection_timeout,
+    )
+
+
+async def do_user_setup() -> None:
+    """Perform the user setup that requires the database interaction"""
+    connection = None
+    try:
+        connection = await get_database_connection()
```
arnaudsjs, line 112: Not being able to connect to the database sound like a very plausible issue here. I would:

* Output which database we are trying to connect to. This will make it easier for the user to debug obvious configuration issues.
* I would also make sure there is a clear error message when the connection to the database doesn't work properly. At the moment we will just get a stack trace I guess.

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,163 @@
...
+
+        click.echo(f"{'User %s: ' %username : <50}{click.style('created', fg='green')}")
+    finally:
+        if connection is not None:
+            await data.disconnect()
+
+    click.echo("Make sure to (re)start the orchestrator to activate all changes.")
+
+
+@click.command(help="Do the initial user setup")
```
arnaudsjs, line 149: I think we should clearly mention that this command is intended to be executed on the orchestrator itself and not remotely. Maybe we should also ask this as a question to the user on the setup checks.

tests/test_usersetup.py
```diff
@@ -0,0 +1,106 @@
...
+    # init_dataclasses_and_load_schema and why we open (and close) a connection ourself.
+    # cli.run() calls cmd which will open and close a connection in the same way.
+    connection = await get_database_connection()
+
+    users = await data.User.get_list()
+    assert len(users) == 1
+    assert users[0].username == "new_user"
+
+    if connection is not None:
+        await data.disconnect()
```
arnaudsjs, line 92: We should use a try-finally here. Otherwise this code is skipped when the assertion fails.

tests/test_usersetup.py
```diff
@@ -0,0 +1,106 @@
...
+):
+    setup_config(tmpdir, postgres_db, database_name)
+
+    dump_path = os.path.join(os.path.dirname(__file__), "db/migration_tests/dumps/v202211230.sql")
+    with open(dump_path, "r") as fh:
+        await PGRestore(fh.readlines(), postgresql_client).run()
+
+    cli = CLI_user_setup()
+    result = await cli.run("new_user", "password")
+    assert "please migrate your DB to the latest version" in result.output
```
arnaudsjs, line 106: We also need to assert that the exit-code is non-zero.

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -3131,7 +3134,7 @@ async def log_instance_creation(
                 {cls.table_name()}
                 (id, tid, process, name, expired)
                 VALUES ($1, $2, $3, $4, null)
-                ON CONFLICT ON CONSTRAINT {cls.table_name()}_unique DO UPDATE
+                ON CONFLICT ON CONSTRAINT {cls.table_name(False)}_unique DO UPDATE
```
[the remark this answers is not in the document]
FloLey: Because the name of the constrain doesn't contain the schema. More than that, I think a constraint name can't contain a '.'. But I'm not sure of the later one.

src/inmanta/data/schema.py
```diff
@@ -120,6 +120,7 @@ async def _legacy_migration_table(self) -> None:
         async with self.connection.transaction():
             # lock table
             await self.connection.execute(f"LOCK TABLE {SCHEMA_VERSION_TABLE} IN ACCESS EXCLUSIVE MODE")
+            table_name_without_schema = SCHEMA_VERSION_TABLE.split(".")[1]
```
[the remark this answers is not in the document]
FloLey: because in the `information_schema.columns` table, the `table_name` column doesn't contain the schema neither, it's just the name. So the Where clause never matched anymore.

src/inmanta/data/__init__.py
```diff
@@ -5533,6 +5536,21 @@ class EnvironmentMetricsTimer(BaseDocument):
...
 
 
+class User(BaseDocument):
+    """A user that can authenticate against inmanta"""
+
+    __primary_key__ = ("id",)
+
+    id: uuid.UUID
+    username: str
+    password_hash: Optional[str] = None
```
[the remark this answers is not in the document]
FloLey: To me it a user should always have a password but and even in the DB the password_hash is mandatory but this was done by bart. I don't know if there was a reason for this or not. I'll make it non-optional for now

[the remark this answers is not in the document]
arnaudsjs: @bartv Do we want the password to be optional or not?

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
arnaudsjs: I see.

src/inmanta/server/services/userservice.py
FloLey: I guess this is database? I'll change it to database for now

src/inmanta/server/services/userservice.py
FloLey: This should be done in a single transaction right?

src/inmanta/server/services/userservice.py
FloLey: I'll create tickets for all the todo's right before I merge this. I think this was the plan of this ticket?
`There are still many todo's of which many are actually a ticket on their own.`

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,110 @@
...
+    @protocol.handle(protocol.methods_v2.delete_user)
+    async def delete_user(self, username: str) -> None:
+        user = await data.User.get_one(username=username)
+        if user is None:
+            raise exceptions.NotFound(f"User with name {username} does not exist.")
+
+        await user.delete()
+
+    @protocol.handle(protocol.methods_v2.set_password)
+    async def set_password(self, username: str, password: str) -> None:
```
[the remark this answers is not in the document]
FloLey: I guess so to, I'll put a minimum length of 8 for now.

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def get_depended_by(self) -> list[str]:
+        return [SLICE_TRANSPORT]
+
+    @protocol.handle(protocol.methods_v2.list_users)
+    async def list_users(self) -> list[model.User]:
+        return [user.to_dao() for user in await data.User.get_list()]
+
+    @protocol.handle(protocol.methods_v2.add_user)
+    async def add_user(self, username: str, password: str) -> model.User:
```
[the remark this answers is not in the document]
FloLey: No idea what the plan is for that part. I'll create a discussion ticket about it right before I merge

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,110 @@
...
+        if not user.password_hash:
+            raise exceptions.UnauthorizedException()
+
+        try:
+            nacl.pwhash.verify(user.password_hash.encode(), password.encode())
+        except nacl.exceptions.InvalidkeyError:
+            raise exceptions.UnauthorizedException()
+
+        # TODO: set an expire
+        token = common.encode_token([str(const.ClientType.api.value)], expire=None)
```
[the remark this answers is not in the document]
FloLey: Indeed I think we don't want that :p 
I'll create a ticket for that to

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,163 @@
...
+@click.option("--reset", help="Reset the password to recover a lost password", is_flag=True)
```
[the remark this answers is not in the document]
FloLey: No it is never used, I think it is there for a follow up ticket. I will remove it for now

src/inmanta/user_setup.py
FloLey: I think using the default value is the easiest and safest option

src/inmanta/user_setup.py
FloLey: you mean asking "Are you on the orchestrator?" 
=> yes: continue
=> no: abort
?

src/inmanta/data/model.py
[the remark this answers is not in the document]
arnaudsjs: Yes. Indeed. I am not sure either whether this will fit our current model well. Maybe we have to go for enums in the database for now.

src/inmanta/server/services/userservice.py
arnaudsjs: We could go for a transaction here, but that would mean we need to use the transaction isolation level repeatable reads. Those types of transactions can fails with an exception when the database fails to serialize the actions on the database. That means we have to implement a retry mechanism as well, which feels like overkill to me. Alternatively we can just perform the insert and catch the exception that occurs when there is a primary key conflict.

src/inmanta/server/services/userservice.py
arnaudsjs: I would create the tickets and remove the TODO's from the PR. The goal of the review is also the verify that none of these TODO's are left behind. Now this is not possible and I don't see a reason to delay the creation of these tickets until right before the merge of the PR.

src/inmanta/user_setup.py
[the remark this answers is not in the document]
arnaudsjs: The thing that annoys me here is that we have an option that is not backed by an implementation. When people from the solutions team are running against a dev version and they see this options, they will expect it to work. So I either should make sure that the implementation behind this options works or alternatively we can remove the option, update that error message and create a ticket to do the implementation later on. But right now we are somewhere in between.

src/inmanta/user_setup.py
arnaudsjs: Essentially yes. But I would provide more context in the question. Something like: `This command should be execute locally on the orchestrator you want to configure. Are you running this command locally?`

src/inmanta/data/model.py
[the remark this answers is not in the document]
sanderr: I'm inclined to agree.

src/inmanta/server/services/userservice.py
FloLey: I'll go for the try/catch :p

FloLey: I added the try/except (not pushed yet) put it seems hard to write a test for this

src/inmanta/server/services/userservice.py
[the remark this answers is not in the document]
FloLey: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/5643

src/inmanta/server/services/userservice.py
FloLey: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/5643

src/inmanta/server/services/userservice.py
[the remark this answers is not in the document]
FloLey: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/5644

src/inmanta/user_setup.py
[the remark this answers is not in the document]
FloLey: Ok I'll remove any mention to it and create a ticket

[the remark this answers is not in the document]
FloLey: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/5645

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,163 @@
...
+
+        if value == "yes":
+            click.echo("Add the following to the configuration in /etc/inmanta/inmanta.d/auth.cfg:\n")
+            click.echo(generate_signing_config())
+
+        raise click.ClickException("Make sure signing configuration is added to the config. See the documentation for details.")
+
+    click.echo(f"{'Authentication signing config: ' : <50}{click.style('enabled', fg='green')}")
+
+    # TODO: verify web-console config (if any)
```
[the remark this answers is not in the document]
FloLey: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/5646

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,114 @@
...
+from inmanta import const, data, protocol
+from inmanta.data import model
+from inmanta.protocol import common, exceptions
+from inmanta.server import SLICE_TRANSPORT, SLICE_USER
+from inmanta.server import protocol as server_protocol
+
+LOGGER = logging.getLogger(__name__)
+
+
+class UserService(server_protocol.ServerSlice):
```
arnaudsjs, line 33: The user management endpoints that write to the database should raise an exception when they are invoked when authentication is disabled on the server. The error message should reference to the tool to setup the initial user.

arnaudsjs (changes requested):

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,181 @@
...
+        DBschema = schema.DBSchema(CORE_SCHEMA_NAME, PACKAGE_WITH_UPDATE_FILES, connection)
+        schema_up_to_date = await DBschema.is_db_schema_up_to_date()
+        if schema_up_to_date:
+            click.echo(f"{'DB schema up to date' : <50}{click.style('yes', fg='green')}")
+        else:
+            click.echo(
+                f"{'DB schema up to date' : <50}"
+                f"{click.style('no: please make sure your DB version and software version are aligned', fg='red')}"
+            )
+            raise click.ClickException("The database version and software version are not aligned")
```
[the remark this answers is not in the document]
arnaudsjs: I think I see what you mean now. It's also important that the version of the client used is the same as the version on the server. I am starting to think that it wasn't a good idea to add this very strong constraint. It will be rather hard for end-user to use this command properly. Users have to make sure that:

* They use the `inmanta-initial-user-setup` command in the venv of the inmanta server or they need to create a venv themselves with the exact same version as the inmanta server.
* And, they need to make sure that the server executed all the database migration scripts to the latest version. 

We will need to do quite some explaining in the documentation of this command and in the error messages to make sure that users can use it without much trouble. @sanderr Are you still in favor of adding this strong constraint?

src/inmanta/protocol/methods_v2.py
```diff
@@ -1328,3 +1328,50 @@ def get_environment_metrics(
...
     """
+
+
+@typedmethod(path="/login", operation="POST", client_types=[ClientType.api], enforce_auth=False, api_version=2)
+def login(username: str, password: str) -> ReturnValue[model.LoginReturn]:
+    """Login a new user. When the login succeeds an authentication header is returned with the Bearer token set.
+
+    :param username: The user to login
+    :param password: The password of this user
+    :raises NotFound: Raised when the user does not exist or is disabled
```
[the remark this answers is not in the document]
FloLey: ho I wouldn't have think of that, I'll change it

src/inmanta/data/__init__.py
```diff
@@ -5533,6 +5537,21 @@ class EnvironmentMetricsTimer(BaseDocument):
...
 
+class User(BaseDocument):
+    """A user that can authenticate against inmanta"""
+
+    __primary_key__ = ("id",)
+
+    id: uuid.UUID
+    username: str
+    password_hash: str
+    auth_method: str
```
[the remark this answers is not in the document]
FloLey: I tried to change that but I get an issue with the insert:
```
        try:
            user = data.User(
                username=username,
                password_hash=pw_hash.decode(),
                auth_method=AuthMethod.DATABASE,
            )
            await user.insert()
```
It tries to insert the key of the enum ('DATABASE') instead of the value ('database') which isn't a valid value in the DB.

src/inmanta/user_setup.py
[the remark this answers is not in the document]
FloLey: @sanderr can I assume your thumps up on the previous comment means that you are still in favor of adding the constraint if the message in the exception is made clearer?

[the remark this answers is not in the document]
sanderr: I am but @arnaudsjs is more hesitant and he does make some valid points. We decided we wanted to involve @wouterdb in this. Forgot to notify you, sorry about that.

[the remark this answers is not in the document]
wouterdb: Multiple remarks here:
1. the error message should guide the user to a solution, i.e. it has to tell the user what to do.
2. I think using a venv is unlikely, but not impossible.
3. I would not check the DB version but try the query and see if it sticks. If there are DB failures that are not due to authentication, I would assume the schema is not up to date an prompt the user to start the orchestrator first.

src/inmanta/protocol/methods_v2.py
```diff
@@ -1328,3 +1328,52 @@ def get_environment_metrics(
     :param end_interval: The end of the time window for which the metrics should be returned.
     :param nb_datapoints: The amount of datapoint that will be returned within the given time interval for each metric.
     """
+
+
+@typedmethod(path="/login", operation="POST", client_types=[ClientType.api], enforce_auth=False, api_version=2)
+def login(username: str, password: str) -> ReturnValue[model.LoginReturn]:
+    """Login a new user. When the login succeeds an authentication header is returned with the Bearer token set.
```
sanderr, line 1335:
```suggestion
    """Login a user. When the login succeeds an authentication header is returned with the Bearer token set.
```
Not a new user, right?

src/inmanta/protocol/methods_v2.py
```diff
@@ -1328,3 +1328,52 @@ def get_environment_metrics(
     :param end_interval: The end of the time window for which the metrics should be returned.
     :param nb_datapoints: The amount of datapoint that will be returned within the given time interval for each metric.
     """
+
+
+@typedmethod(path="/login", operation="POST", client_types=[ClientType.api], enforce_auth=False, api_version=2)
+def login(username: str, password: str) -> ReturnValue[model.LoginReturn]:
```
sanderr, line 1334: Why `ReturnValue`? Isn't that the old approach? I could be mistaken.

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
arnaudsjs: I see that all other enums used in DAO classes resolve this by using small letter in the key of the enum instead of capital letters. This goes against the Python convention of using capital letters for the key of an enum, but I think that the benefits of improved readability by using the enum instead of the str type outweighs the downside of violating that naming convention.

src/inmanta/protocol/methods_v2.py
```diff
@@ -1328,3 +1328,52 @@ def get_environment_metrics(
...
+
+
+@typedmethod(path="/user", operation="GET", client_types=[ClientType.api], api_version=2)
+def list_users() -> List[model.User]:
+    """List all users
+
+    :return: A list of all users"""
+
+
+@typedmethod(path="/user/<username>", operation="DELETE", client_types=[ClientType.api], api_version=2)
```
sanderr, line 1350: Shouldn't this require authentication? Same question for the methods below.

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
FloLey: fixed :)

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,125 @@
...
+from inmanta.server import protocol as server_protocol
+
+LOGGER = logging.getLogger(__name__)
+
+
+def verify_authentication_enabled() -> None:
+    """raises an UnauthorizedException exception if server authentication is not enabled"""
+    if not server_config.server_enable_auth.get():
+        raise exceptions.UnauthorizedException(
+            "Server authentication should be enabled. To setup the initial user use the user_setup tool."
```
arnaudsjs, line 38:
```suggestion
            "Server authentication should be enabled. To setup the initial user use the inmanta-initial-user-setup tool."
```

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,179 @@
...
+            server_config.db_name.get(),
+            database_username,
+            database_password,
+            create_db_schema=False,
+            connection_pool_min_size=connection_pool_min_size,
+            connection_pool_max_size=connection_pool_max_size,
+            connection_timeout=connection_timeout,
+        )
+    except Exception:
+        raise ConnectionPoolException
```
arnaudsjs, line 119:
```suggestion
        raise ConnectionPoolException()
```

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,179 @@
...
+        click.echo(f"{'Connection to database' : <50}" f"{click.style('success', fg='red')}")
+        users = await data.User.get_list()
+
+        if len(users):
+            raise click.ClickException("There are already users in the database")
+
+        username = click.prompt("What username do you want to use?", default="admin")
+        password = click.prompt("What password do you want to use?", hide_input=True)
+
+        pw_hash = nacl.pwhash.str(password.encode())
```
arnaudsjs, line 140: We should also enforce the minimal password length here.

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,179 @@
...
+            password_hash=pw_hash.decode(),
+            auth_method="database",
+        )
+        await user.insert()
+
+        click.echo(f"{'User %s: ' %username : <50}{click.style('created', fg='green')}")
+    except ConnectionPoolException as e:
+        click.echo(f"{'Connection to database {server_config.db_host.get()}' : <50}" f"{click.style('failed', fg='red')}")
+        raise e
+    except UndefinedTableError:
```
arnaudsjs, line 154: I think it's better to catch the PostgresError exception here. This is the base class for all PostgreSQL-related errors. If we change the schema of the user table in the future, we might get a different error when the schema is out of date.
```suggestion
    except PostgresError:
```

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def get_dependencies(self) -> list[str]:
+        return [SLICE_DATABASE]
+
+    def get_depended_by(self) -> list[str]:
+        return [SLICE_TRANSPORT]
+
+    @protocol.handle(protocol.methods_v2.list_users)
+    async def list_users(self) -> list[model.User]:
+        return [user.to_dao() for user in await data.User.get_list()]
```
sanderr, line 56: We should probably sort this

tests/server/test_user.py
```diff
@@ -0,0 +1,146 @@
...
+    config.Config.set("auth_jwt_default", "sign", "true")
+    config.Config.set("auth_jwt_default", "client_types", "agent,compiler,api")
+    config.Config.set("auth_jwt_default", "key", "eciwliGyqECVmXtIkNpfVrtBLutZiITZKSKYhogeHMM")
+    config.Config.set("auth_jwt_default", "expire", "0")
+    config.Config.set("auth_jwt_default", "issuer", "https://localhost:8888/")
+    config.Config.set("auth_jwt_default", "audience", "https://localhost:8888/")
+
+
+@pytest.fixture
+def auth_client():
```
arnaudsjs, line 41:
```suggestion
def auth_client(server_pre_start):
```

tests/server/test_user.py
```diff
@@ -0,0 +1,146 @@
...
+import pytest
+
+from inmanta import config, const
+from inmanta.protocol import common, endpoints
+from inmanta.server import SLICE_USER, protocol
+
+
+@pytest.fixture
+def server_pre_start(server_config):
+    """This fixture is called by the server. Override this fixture to influence server config"""
```
arnaudsjs, line 28:
```suggestion
    """Ensure that the server started by the server fixtures have authentication enabled with auth_method database"""
```

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,125 @@
...
+        if user is None:
+            raise exceptions.NotFound(f"User with name {username} does not exist.")
+
+        await user.delete()
+
+    @protocol.handle(protocol.methods_v2.set_password)
+    async def set_password(self, username: str, password: str) -> None:
+        verify_authentication_enabled()
+        if not password or len(password) < 8:
+            raise exceptions.BadRequest("the password should be at least 8 characters long")
```
sanderr, line 92: Perhaps we should make this a constant in case we ever change it.

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,125 @@
...
+from inmanta.server import config as server_config
+from inmanta.server import protocol as server_protocol
+
+LOGGER = logging.getLogger(__name__)
+
+
+def verify_authentication_enabled() -> None:
+    """raises an UnauthorizedException exception if server authentication is not enabled"""
+    if not server_config.server_enable_auth.get():
+        raise exceptions.UnauthorizedException(
```
arnaudsjs, line 37: Shouldn't this be a `BadRequest`? An UnauthorizedException means that the user didn't provide credentials, which is not really the problem here. BadRequest indicates that the client made an error. That is applicable here, because the client shouldn't make these API calls the authentication is disabled on the server.

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,125 @@
...
+        # check if the user exists
+        user = await data.User.get_one(username=username)
+        if not user or not user.password_hash:
+            raise exceptions.UnauthorizedException()
+        try:
+            nacl.pwhash.verify(user.password_hash.encode(), password.encode())
+        except nacl.exceptions.InvalidkeyError:
+            raise exceptions.UnauthorizedException()
+
+        token = common.encode_token([str(const.ClientType.api.value)], expire=None)
```
sanderr, line 116: Is it good practice to never expire this? I honestly don't know mysefl.

arnaudsjs (changes requested):

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,183 @@
...
+import click
+
+import nacl.pwhash
+from inmanta import config, data
+from inmanta.data import CORE_SCHEMA_NAME, PACKAGE_WITH_UPDATE_FILES, schema
+from inmanta.server import config as server_config
+
+
+class ConnectionPoolException(Exception):
+    """Exception that is raised it the connection to the connection pool fails"""
```
sanderr, line 33:
```suggestion
    """Exception that is raised if the connection to the connection pool fails"""
```

sanderr (commented):
No major remarks. Most of my comments are speculation or suggestion more than change requests. I haven't reviewed the user setup yet because I assume it's changed since I started review. I'll get started on that now.

src/inmanta/db/versions/v202302270.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+async def update(connection: Connection) -> None:
+    schema = """
+    CREATE TYPE auth_method AS ENUM('database', 'oidc');
+
+    CREATE TABLE IF NOT EXISTS public.user(
+        id uuid PRIMARY KEY,
+        username VARCHAR NOT NULL UNIQUE ,
+        password_hash VARCHAR NOT NULL,
```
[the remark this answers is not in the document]
arnaudsjs: Is there a specific reason why the default collation that PostgreSQL uses wouldn't be sufficient here?

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
arnaudsjs: > Looks like this has been there forever (since the migration to Postgres). @arnaudsjs do you know why?

I missed this message. Probably as issue with by browser cache. I don't recall why this was required, but probably asyncpg doesn't know what to do if the type in the database is an enum and you pass in a enum as a parameter to the query.

src/inmanta/user_setup.py
sanderr: I learned recently that both are valid, though I still prefer the instance over the class.

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,180 @@
...
+async def do_user_setup() -> None:
+    """Perform the user setup that requires the database interaction"""
+    connection = None
+    try:
+        click.echo(
+            f"{'Trying to connect to DB:': <50}"
+            f"{('%s (%s:%s)' % ( server_config.db_name.get(), server_config.db_host.get(), server_config.db_port.get()))}"
+        )
+        connection = await get_connection_pool()
+        click.echo(f"{'Connection to database' : <50}" f"{click.style('success', fg='red')}")
```
sanderr, line 131: Is this meant to be green? Also, this is an f-string with only this statement in there.

src/inmanta/user_setup.py
sanderr: In addition to this, I think you'll either need to use raw postgres queries, or also catch a pydantic error for the `data.User.get_list()`, e.g. if we add a field to the users table but the database has not been upgraded yet.

sanderr (commented):
Final comments

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
arnaudsjs: Indeed.

src/inmanta/data/__init__.py
```diff
@@ -1618,7 +1621,9 @@ async def insert_many(
             records.append(tuple(current_record))
 
         async with cls.get_connection(connection) as con:
-            await con.copy_records_to_table(table_name=cls.table_name(), columns=columns, records=records)
+            await con.copy_records_to_table(
+                table_name=cls.table_name(include_schema=False), columns=columns, records=records, schema_name="public"
```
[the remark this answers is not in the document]
FloLey: that's how the `copy_records_to_table` function works: one argument with the name and another with the `schema_name` the second one being optional. But the function doesn't work if the table name contains the schema name.

src/inmanta/data/model.py
```diff
@@ -722,3 +722,29 @@ class EnvironmentMetricsResult(BaseModel):
...
+    username: str
+    auth_method: AuthMethod
+
+
+class LoginReturn(BaseModel):
+    """
+    Login information
+
+    :param token: A token representing the user's authentication session
+    :param expiry: The timestamp at which the token will expire
```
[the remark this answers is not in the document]
FloLey: I've removed `expiry` completely as it isn't used for now.

src/inmanta/data/schema.py
```diff
@@ -31,7 +31,7 @@
 
 LOGGER = logging.getLogger(__name__)
 
-SCHEMA_VERSION_TABLE = "schemamanager"
+SCHEMA_VERSION_TABLE = "public.schemamanager"
```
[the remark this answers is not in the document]
FloLey: I think this will answer your question: https://inmanta.slack.com/archives/CKRF0C8R3/p1671789024753869?thread_ts=1671787545.275579&cid=CKRF0C8R3

src/inmanta/protocol/methods_v2.py
FloLey: I think they all do by default?
only the` /login ` has `enforce_auth=False`

src/inmanta/protocol/methods_v2.py
FloLey: indeed, a user that exists.

src/inmanta/protocol/methods_v2.py
FloLey: I have no idea. What would be the new approach? @arnaudsjs do you know about this?

sanderr: Sorry, I meant to withdraw this comment. The new approach is just returning the object, but if you want to explicitly include a bearer token (as you do) that doesn't suffice.

src/inmanta/protocol/methods_v2.py
sanderr: Oh, I had missed that it defaulted to `True`. Makes sense though.

src/inmanta/server/services/userservice.py
FloLey: there is a follow up ticket for that: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/5643

src/inmanta/protocol/methods_v2.py
sanderr: But how does that work for agent methods? I don't see them being excluded, do they have their own form of auth or is that an oversight?

src/inmanta/server/services/userservice.py
sanderr: Ok, that'll do!

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,180 @@
...
+def generate_signing_config() -> str:
+    hostname = socket.gethostname()
+    return f"""[auth_jwt_default]
+algorithm=HS256
+sign=true
+client_types=agent,compiler,api
+key={secrets.token_urlsafe()}
+expire=0
+issuer=https://{hostname}:{server_config.server_bind_port.get()}/
+audience=https://{hostname}:{server_config.server_bind_port.get()}/
```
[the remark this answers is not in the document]
FloLey: I don't think so but I'm not sure. I didn't change this part of the code as it was 'working'.

src/inmanta/data/schema.py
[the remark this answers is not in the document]
sanderr: Thanks, I don't get it but I can let it slide.

src/inmanta/db/versions/v202302270.py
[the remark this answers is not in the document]
sanderr: I honestly don't know. I do seem to recall the "default" was a very loose concept here. I think PostgreSQL itself doesn't have one and it falls back to something ill defined, Markus wasn't positive about it but I don't recall the details. My takeaway was that generally you should always use an explicit collation, but I could just as well have misunderstood that. You didn't understand that the same way?

src/inmanta/protocol/methods_v2.py
FloLey: I think that the agents need a token to: 

- the `create_token` method uses the `client_type` to create a token and one of the possible types is agent. 

- then `authorize_request` function verifies that the correct client_type is used.

sanderr: What I mean to say is that enabling authentication also enables it for the agents. I have no clue how difficult it is to set up for the agent, and whether that is acceptable or not.

sanderr: I guess it makes sense.

src/inmanta/data/model.py
```diff
@@ -746,5 +746,4 @@ class LoginReturn(BaseModel):
     """
 
     token: str
-    expiry: int
```
sanderr: You deleted this but it's still in the docstring.

sanderr (approved):

src/inmanta/db/versions/v202302270.py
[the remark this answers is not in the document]
arnaudsjs: I mainly understood that you have to use collations if you want to have the output sorted in a specific way. I am not sure what the default collation does, but I don't really see why we would want to have the usernames sorted in a specific way. To me it's important that the API returns the usernames in a consistent order. Whether capital letters come before small letter or the other way around, etc. is not that important to me.

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def get_depended_by(self) -> list[str]:
+        return [SLICE_TRANSPORT]
+
+    @protocol.handle(protocol.methods_v2.list_users)
+    async def list_users(self) -> list[model.User]:
+        return sorted([user.to_dao() for user in await data.User.get_list()], key=lambda u: u.username)
+
+    @protocol.handle(protocol.methods_v2.add_user)
+    async def add_user(self, username: str, password: str) -> model.User:
```
arnaudsjs, line 60: Shouldn't we also make sure that the username is not the empty string? I guess that having an empty string as a username can cause all sorts of issues. We should also validate this in the user setup script.

src/inmanta/server/services/userservice.py
```diff
@@ -0,0 +1,125 @@
...
+        user = await data.User.get_one(username=username)
+        if user is None:
+            raise exceptions.NotFound(f"User with name {username} does not exist.")
+
+        await user.delete()
+
+    @protocol.handle(protocol.methods_v2.set_password)
+    async def set_password(self, username: str, password: str) -> None:
+        verify_authentication_enabled()
+        if not password or len(password) < MIN_PASSWORD_LENGTH:
```
arnaudsjs, line 92: We should all this constraint on the password length to the API documentation. The front-end, for example, will find that information rather useful.

arnaudsjs (changes requested):

src/inmanta/server/services/userservice.py
FloLey: In the user setup script, if no username is provided it defaults to Admin. So I don't think a check is needed there. But I've added it to the test

src/inmanta/user_setup.py
```diff
@@ -0,0 +1,184 @@
...
+        user = data.User(
+            username=username,
+            password_hash=pw_hash.decode(),
+            auth_method=AuthMethod.database,
+        )
+        await user.insert()
+
+        click.echo(f"{'User %s: ' %username : <50}{click.style('created', fg='green')}")
+    except ConnectionPoolException as e:
+        click.echo(f"{'Connection to database {server_config.db_host.get()}' : <50}" f"{click.style('failed', fg='red')}")
```
arnaudsjs, line 156:
```suggestion
        click.echo(f"{f'Connection to database {server_config.db_host.get()}' : <50}" f"{click.style('failed', fg='red')}")
```

src/inmanta/db/versions/v202302270.py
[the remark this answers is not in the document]
sanderr: I agree with that sentiment, so since I can't come up with a concrete reason to use the collation, let's leave it out for now. If it turns out to have been important it will come back up.

arnaudsjs (changes requested):

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #5333, opened 2023-01-06
Issue/5127 add agent count metric

### LINKED ISSUE #5127: Add agent_count metrics collector
This ticket adds the metrics collector that tracks the amount of agent instances grouped by agent state.

Tasks:

- [x] Add metrics collector
- [x] Register the metrics collector to the `EnvironmentMetricsService` slice.

FloLey:
# Description

This PR adds the metrics collector that tracks the amount of agent instances grouped by agent state.

- [x] Add metrics collector
- [x]  Register the metrics collector to the EnvironmentMetricsService slice.

closes #5127 

# Self Check:

- [x] Changelog entry


src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -243,3 +244,37 @@ async def get_metric_value(
...
         return metric_values
+
+
+class AgentCountMetricsCollector(MetricsCollector):
+    """
+    This Metric will track the number of agents (grouped by agent status).
+    """
+
+    def get_metric_name(self) -> str:
+        return "agent_count"
```
arnaudsjs, line 255: According to the design this name should be `resource.agent_count`.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -217,7 +218,7 @@ class ResourceCountMetricsCollector(MetricsCollector):
     """
 
     def get_metric_name(self) -> str:
-        return "resource_count"
+        return "resource.resource_count"
```
FloLey, line 221: @arnaudsjs this is not related to this PR but is also not right, so i'll fix it here

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -243,3 +244,37 @@ async def get_metric_value(
...
         return metric_values
+
+
+class AgentCountMetricsCollector(MetricsCollector):
+    """
+    This Metric will track the number of agents (grouped by agent status).
+    """
+
+    def get_metric_name(self) -> str:
+        return "resource.agent_count"
```
FloLey, line 255: this https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/server/services/environment_metrics_service.py#L63
prevents me form changing the name to `resource.agent_count`. I'm not really sure what I should do here? something like this?

```
    def __init__(self, target: string, metric_name: str, count: int, environment: uuid.UUID, grouped_by: Optional[str] = None) -> None:
        if METRIC_NAME_SEPARATOR in metric_name:
            raise Exception('The character "%s" can not be used in the metric_name (%s)' % (METRIC_NAME_SEPARATOR, metric_name))
        if METRIC_NAME_SEPARATOR in target:
            raise Exception('The character "%s" can not be used in the target (%s)' % (METRIC_NAME_SEPARATOR, target))
        if grouped_by and METRIC_NAME_SEPARATOR in grouped_by:
            raise Exception(
                'The character "%s" can not be used in the grouped_by value (%s)' % (METRIC_NAME_SEPARATOR, grouped_by)
            )
        self.metric_name = ".".join([target, metric_name, grouped_by]) if grouped_by else  ".".join([target, metric_name]) 
        self.count = count
```
and add `resource` via `target`? 
or diverge from the design and use an underscore `resource_agent_count or some other separator?`

arnaudsjs: Why not use a another separator characters like `#` for example? That way it's clear that the first part is the name and the second part the subdivision.

FloLey: that works for me to :) I'll do that

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, metric_name: str, count: int, environment: uuid.UUID, grouped
             raise Exception(
                 'The character "%s" can not be used in the grouped_by value (%s)' % (METRIC_NAME_SEPARATOR, grouped_by)
             )
-        self.metric_name = ".".join([metric_name, grouped_by]) if grouped_by else metric_name
+        self.metric_name = METRIC_NAME_SEPARATOR.join([metric_name, grouped_by]) if grouped_by else metric_name
```
arnaudsjs, line 69: Why do we use string concatenation between the name of a metric and the sub-category instead of using a separate field (both in the code as well as in the database). Wouldn't the latter make the code more easy to read?

arnaudsjs (approved):

Hugo-Inmanta (approved):

PR #5335, opened 2023-01-09
added ADR about timezone semantics

sanderr:
# Description

I decided to add an ADR about this because I was once again confused about the fine points of how we decided to handle timezones.

Additionally, the ADR contains a new decision regarding new API endpoints. Please pay special attention to that. I already violated it in inmanta/designs#113 (before I double checked the current semantics), so I'll need to make minor changes there if this is accepted.

I've added the three of you as decision drivers and therefore reviewers because you were the reviewers of the PR that implemented this back in the day. Feel free to opt out if you like (as long as you don't do so all three of you).

# Self Check:

- [x] Changelog entry
- [x] ~~Code is clear and sufficiently documented~~
- [x] ~~No (preventable) type errors (check using make mypy or make mypy-diff)~~

### DIFF
--- a/docs/adr/XXXX-timezones.md
+++ b/docs/adr/XXXX-timezones.md
@@ -32,9 +32,12 @@ timezone information for these datetime objects?
 
 ## Decision Outcome
 
-Chosen option: option 3 with timezone-aware internal representation, because it is as explicit as we can make (decision driver
-3) it without breaking backwards compatibility. In contrast to option 4, it remains simple: a single contract that applies to
-all API endpoints.
+Chosen option: option 3 with timezone-aware internal representation. Support both timezone aware and naive timestamps as input,
+use timezone-naive timestamps in API return values. Internally, always use timezone-aware timestamps (objects are converted at
+the API boundary). Any timezone-naive timestamps represent a timestamp in UTC.
+
+Rationale: it is as explicit as we can make it (decision driver 3) without breaking backwards compatibility. In contrast to
+option 4, it remains simple: a single contract that applies to all API endpoints.
 
 Option 2 (fully timezone-aware) has a lot of merit and we may want to migrate to it at some point, though that would be a
 breaking change.
--- a/docs/adr/XXXX-timezones.md
+++ b/docs/adr/XXXX-timezones.md
@@ -16,7 +16,7 @@ How do we deal with timezones when it comes to datetime data? This data is passe
 various places in the database and manipulated throughout the code. Which invariants do we uphold with regards to explicit
 timezone information for these datetime objects?
 
-## Decision Drivers <!-- optional -->
+## Decision Drivers
 
 * Behavior should be consistent regardless of the timezone participating hosts (e.g. server, agent, web-console client) live in.
 * Timezone handling over the API is non-trivial: we need to present a clear contract to the end-user.
--- a/docs/adr/XXXX-timezones.md
+++ b/docs/adr/XXXX-timezones.md
@@ -32,9 +32,12 @@ timezone information for these datetime objects?
 
 ## Decision Outcome
 
-Chosen option: option 3 with timezone-aware internal representation. Support both timezone aware and naive timestamps as input,
-use timezone-naive timestamps in API return values. Internally, always use timezone-aware timestamps (objects are converted at
-the API boundary). Any timezone-naive timestamps represent a timestamp in UTC.
+Chosen option: option 3 with timezone-aware internal representation. 
+* Support both timezone aware and naive timestamps as API input,
+* Any timezone-naive timestamps represent a timestamp in UTC.
+* Use timezone-naive timestamps in API return values. 
+* Internally, always use timezone-aware timestamps (objects are converted at
+the API boundary). 
 
 Rationale: it is as explicit as we can make it (decision driver 3) without breaking backwards compatibility. In contrast to
 option 4, it remains simple: a single contract that applies to all API endpoints.
--- a/docs/adr/XXXX-timezones.md
+++ b/docs/adr/XXXX-timezones.md
@@ -36,7 +36,7 @@ Chosen option: option 3 with timezone-aware internal representation.
 * Support both timezone aware and naive timestamps as API input,
 * Any timezone-naive timestamps represent a timestamp in UTC.
 * Use timezone-naive timestamps in API return values. 
-* Internally, always use timezone-aware timestamps (objects are converted at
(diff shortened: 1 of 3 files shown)


docs/adr/XXXX-timezones.md
```diff
@@ -0,0 +1,77 @@
...
+
+* Use UTC everywhere
+* Use timezone-aware datetimes everywhere
+* Support both timezone aware and naive timestamps as input, return naive timestamps. All naive timestamps are implicitly
+    assumed to be in UTC. Internally both a naive and an aware representation are considered.
+* Like the previous (mixed) option but make new end-points (where backwards compatibility is irrelevant) timezone-aware only.
+
+## Decision Outcome
+
+Chosen option: option 3 with timezone-aware internal representation, because it is as explicit as we can make (decision driver
```
wouterdb, line 35: I would write it out in full, this is the most important part, so it should not have references (which may lead to confusion)

sanderr:
Should I already add a changelog note that we'll eventually migrate to fully timezone-aware? This would give API clients the time to make sure they are compatible so we can make the change from iso7 on.

docs/adr/XXXX-timezones.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Context and Problem Statement
+
+How do we deal with timezones when it comes to datetime data? This data is passed in both directions over the API, stored at
+various places in the database and manipulated throughout the code. Which invariants do we uphold with regards to explicit
+timezone information for these datetime objects?
+
+## Decision Drivers
+
+* Behavior should be consistent regardless of the timezone participating hosts (e.g. server, agent, web-console client) live in.
```
wouterdb, line 21: These are bulleted lists, but referred to as numbered lists?

docs/adr/XXXX-timezones.md
```diff
@@ -0,0 +1,77 @@
...
+### Negative Consequences
+
+* API output must remain timezone-naive
+* Difference between aware/naive on the API boundary vs internally can be confusing
+
+## Pros and Cons of the Options
+
+### option 1: UTC everywhere
+
+* Good, because it presents a clear contract for the end user (all timestamps are implicitly in UTC)
```
wouterdb, line 60: I like this notation

sanderr: The "good, ...", "bad, ..."? It's part of the ADR template.

wouterdb: good template, that,

docs/adr/XXXX-timezones.md
sanderr: I wasn't sure about the desired format. The template uses bullets, then refers to the options as `[option x]`. I think the intent is to describe the option there but that seems very verbose to me. Perhaps a numbered list is the way to go, but then we might get other ADRs that do it differently. I'm not sure how much of an issue this would be.

docs/adr/XXXX-timezones.md
```diff
@@ -0,0 +1,80 @@
...
+2. Use timezone-aware datetimes everywhere
+3. Support both timezone aware and naive timestamps as input, return naive timestamps. All naive timestamps are implicitly
+    assumed to be in UTC. Internally both a naive and an aware representation are considered.
+4. Like the previous (mixed) option but make new end-points (where backwards compatibility is irrelevant) timezone-aware only.
+
+## Decision Outcome
+
+Chosen option: option 3 with timezone-aware internal representation. Support both timezone aware and naive timestamps as input,
+use timezone-naive timestamps in API return values. Internally, always use timezone-aware timestamps (objects are converted at
+the API boundary). Any timezone-naive timestamps represent a timestamp in UTC.
```
wouterdb, line 37:
```suggestion
Chosen option: option 3 with timezone-aware internal representation. 
* Support both timezone aware and naive timestamps as API input,
* Any timezone-naive timestamps represent a timestamp in UTC.
* Use timezone-naive timestamps in API return values. 
* Internally, always use timezone-aware timestamps (objects are converted at
the API boundary). 
```

wouterdb (approved):

docs/adr/XXXX-timezones.md
```diff
@@ -0,0 +1,83 @@
...
+    https://github.com/inmanta/inmanta-core/issues/2919,
+    https://github.com/inmanta/inmanta-core/pull/2953,
+    https://github.com/inmanta/inmanta-lsm/pull/621,
+]
+
+## Context and Problem Statement
+
+How do we deal with timezones when it comes to datetime data? This data is passed in both directions over the API, stored at
+various places in the database and manipulated throughout the code. Which invariants do we uphold with regards to explicit
+timezone information for these datetime objects?
```
arnaudsjs, line 17: I think the problem statement is not sufficiently clear. This document often talks about backwards compatibility issues. So basically we describe here how we went from implementation X to implementation Y regarding the handling of timezones. I think we should clearly describe here what the implementation X was. That will make it easier for the reader to interpret the backwards compatibility constraints discussed later on in the document.

arnaudsjs (changes requested):

docs/adr/XXXX-timezones.md
```diff
@@ -0,0 +1,83 @@
...
+    assumed to be in UTC. Internally both a naive and an aware representation are considered.
+4. Like the previous (mixed) option but make new end-points (where backwards compatibility is irrelevant) timezone-aware only.
+
+## Decision Outcome
+
+Chosen option: option 3 with timezone-aware internal representation. 
+* Support both timezone aware and naive timestamps as API input,
+* Any timezone-naive timestamps represent a timestamp in UTC.
+* Use timezone-naive timestamps in API return values. 
+* Internally, always use timezone-aware timestamps (objects are converted at
```
arnaudsjs, line 39:
```suggestion
* Internally (database + code), always use timezone-aware timestamps (objects are converted at
```

arnaudsjs (changes requested):

docs/adr/XXXX-timezones.md
sanderr: I added some more context, I think it should be clear now.

arnaudsjs (approved):

docs/adr/XXXX-timezones.md
```diff
@@ -0,0 +1,90 @@
+# timezones
+
+* Status: proposed
```
sanderr, line 3: note to self: set to accepted

bartv (approved):
Looks good. I do  think at some point we will need to add an "API intro" to our docs and then make it explicit there as well.

PR #5339, opened 2023-01-10
issue/5126-add-compile-wait-time-metrics

### LINKED ISSUE #5126: Add compile_waiting_time metrics collector
This ticket adds the metrics collector that tracks the amount of time a compile request spends in the compile queue before being executed by the compiler service.

Tasks:

- [x] Add metrics collector
- [x] Register the metrics collector to the `EnvironmentMetricsService` slice.

FloLey:
# Description

This PR adds the metrics collector that tracks the amount of time a compile request spends in the compile queue before being executed by the compiler service.

Tasks:
- [x] Add metrics collector
- [x] Register the metrics collector to the EnvironmentMetricsService slice.

closes #5126 

# Self Check:

- [x] Changelog entry


arnaudsjs (changes requested):

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -318,3 +312,42 @@ async def get_metric_value(
...
+
+    async def get_metric_value(
+        self, start_interval: datetime, end_interval: datetime, connection: asyncpg.connection.Connection
+    ) -> Sequence[MetricValueTimer]:
+        query: str = f"""
+            SELECT count(*)  as count,environment,sum(started-requested) as compile_waiting_time
+            FROM {Compile.table_name()}
+            WHERE started >= '{start_interval}'
+            AND started < '{end_interval}'
+            GROUP BY environment
```
arnaudsjs, line 336: We should never use string concatenation in queries. This opens the door for SQL-injection attacks.

arnaudsjs (changes requested):

src/inmanta/server/services/environment_metrics_service.py
Hugo-Inmanta: Nice catch, @FloLey can you also fix this in the compile time metrics query ?
How can we fix this, should we use query parameters instead ?

arnaudsjs: Yes. You write the query with `$<number>` placeholders in the locations where you have a parameter. Asyncpg will then make sure that the parameter is correctly escaped to prevent an incorrect interpretation of the value. This is not possible for the `{Compile.table_name()}` substitution, but that is not bad because the value is directly derived from the name of a class. So we know it's harmless.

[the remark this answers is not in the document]
arnaudsjs: You can ignore this comment. `requested` is always set.

src/inmanta/db/versions/v202301100.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Contact: code@inmanta.com
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    schema = """
+    ALTER TABLE public.environmentmetricsgauge ADD COLUMN grouped_by TEXT DEFAULT 'None';
```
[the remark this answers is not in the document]
arnaudsjs: 2. Let's add a comment about this in the migration script.

tests/server/test_environement_metrics.py
```diff
@@ -780,3 +782,120 @@ async def add_compiles(environment: uuid.UUID, compile_times: Sequence[float]):
...
+    assert len(envs) == 2
+
+    wait_times: List[float] = [2.1, 4.3]
+    await add_compiles(env_uuid2, wait_times)
+    await metrics_service.flush_metrics()
+
+    result_gauge = await data.EnvironmentMetricsTimer.get_list()
+
+    expected_count = len(wait_times)
+    expected_total_compile_time = sum(wait_times)
```
Hugo-Inmanta, line 872:
```suggestion
    expected_total_wait_time = sum(wait_times)
```

tests/server/test_environement_metrics.py
```diff
@@ -780,3 +782,120 @@ async def add_compiles(environment: uuid.UUID, compile_times: Sequence[float]):
...
+    # Add another set of compiles to the first environment.
+    wait_times: List[float] = [1.1, 2.2, 3.3, 4.4]
+    await add_compiles(env_uuid1, wait_times)
+
+    await metrics_service.flush_metrics()
+
+    result_gauge = await data.EnvironmentMetricsTimer.get_list()
+
+    expected_count = len(wait_times)
+    expected_total_compile_time = sum(wait_times)
```
Hugo-Inmanta, line 892:
```suggestion
    expected_total_wait_time = sum(wait_times)
```

tests/server/test_environement_metrics.py
```diff
@@ -780,3 +782,120 @@ async def add_compiles(environment: uuid.UUID, compile_times: Sequence[float]):
...
+
+    assert len(result_gauge) == 2
+    assert any(
+        x.count == expected_count
+        and x.metric_name == "orchestrator.compile_waiting_time"
+        and x.environment == environment2.id
+        and x.value == expected_total_compile_time
```
Hugo-Inmanta, line 879:
```suggestion
        and x.value == expected_total_wait_time
```

tests/server/test_environement_metrics.py
```diff
@@ -780,3 +782,120 @@ async def add_compiles(environment: uuid.UUID, compile_times: Sequence[float]):
...
+
+    expected_count = len(wait_times)
+    expected_total_compile_time = sum(wait_times)
+
+    assert len(result_gauge) == 3
+    assert any(
+        x.count == expected_count
+        and x.metric_name == "orchestrator.compile_waiting_time"
+        and x.environment == environment1.id
+        and x.value == expected_total_compile_time
```
Hugo-Inmanta, line 899:
```suggestion
        and x.value == expected_total_wait_time
```

Hugo-Inmanta (approved):
Looks good to me, apart from one variable renaming :+1:

src/inmanta/db/versions/v202301100.py
```diff
@@ -0,0 +1,34 @@
...
+    Contact: code@inmanta.com
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    # grouped_by has as default value 'None' as it is part of the PRIMARY KEY and can therefore not be NULL.
+    schema = """
+    ALTER TABLE public.environmentmetricsgauge ADD COLUMN grouped_by VARCHAR DEFAULT 'None';
```
arnaudsjs, line 25: I would use a default value here for which it is very unlikely that it would ever be the name of a group. Like `__None__` for example. `None` sound like something that could become a group at some point.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -166,7 +155,11 @@ def create_metrics_gauge(metric_values_gauge: Sequence[MetricValue], timestamp:
             for mv in metric_values_gauge:
                 metric_gauge.append(
                     EnvironmentMetricsGauge(
-                        metric_name=mv.metric_name, timestamp=timestamp, count=mv.count, environment=mv.environment
+                        metric_name=mv.metric_name,
+                        grouped_by=mv.grouped_by if mv.grouped_by else "None",
```
arnaudsjs, line 159: Let's turn this special value into constant. We can then give it a clear name and add a comment to it to make it clear why we have it.

```suggestion
                        grouped_by=mv.grouped_by if mv.grouped_by else "None",
```

src/inmanta_ext/core/extension.py
```diff
@@ -50,3 +51,4 @@ def setup(application: ApplicationContext) -> None:
     application.register_slice(orchestrationservice.OrchestrationService())
     application.register_slice(dryrunservice.DyrunService())
     application.register_slice(notificationservice.NotificationService())
+    application.register_slice(environment_metrics_service.EnvironmentMetricsService())
```
FloLey, line 54: @arnaudsjs Am I right that is was missing? I needed this in LSM

arnaudsjs: Yes. Indeed. This is required. Otherwise the server won't start the slice.

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #5342, opened 2023-01-10
Fix race condition in resource_action_consistency_check

arnaudsjs:
# Description

Fix race condition in the `resource_action_consistency_check` method. The assertion would fail when more resource actions are added in between the execution of both SELECT queries. This PR ensures that both queries are executed in a transaction with the `repeatable_read` isolation level to prevent this race condition.

# Self Check:

- [x] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -2186,6 +2189,27 @@ def to_dict(self) -> JsonType:
...
+        """
+        async with cls.get_connection() as postgresql_client:
+            max_retries = 3
+            while True:
+                try:
+                    async with postgresql_client.transaction(isolation=tx_isolation_level):
+                        return await fnc(postgresql_client)
+                except SerializationError:
+                    if max_retries >= 0:
+                        max_retries -= 1
```
sanderr, line 2209: Shouldn't we include a delay? Markus even recommended an increasing delay but I think we can safely simplify that for testing purposes.

sanderr (approved):

src/inmanta/data/__init__.py
arnaudsjs: Indeed. I didn't include it because it's a test suite here. But since I moved this logic to `src/inmanta/data/__init__.py`, it makes sense to add the backoff logic as well.

Hugo-Inmanta (approved):

src/inmanta/data/__init__.py
```diff
@@ -2191,24 +2191,28 @@ def to_dict(self) -> JsonType:
...
                         return await fnc(postgresql_client)
                 except SerializationError:
-                    if max_retries >= 0:
-                        max_retries -= 1
+                    if attempt > 3:
+                        raise Exception("Failed to execute transaction after 3 attempts.")
                     else:
-                        raise Exception("Failed to execute transaction after 3 retries.")
+                        # Exponential backoff
+                        await asyncio.sleep(10 * pow(2, attempt))
```
sanderr, line 2214: 20-80 seconds seems like a lot to me. I was expecting more something like 0.1-2 seconds or so. But I must admit I can't say with confidence what would be an appropriate interval here.

sanderr (approved):

src/inmanta/data/__init__.py
arnaudsjs: I put it to 10ms, 100ms, 1000ms

sanderr: Looks good to me.

PR #5353, opened 2023-01-11
Fix interval first metrics collection after starting the server

arnaudsjs:
# Description

Ensure that we only write metrics to the database after the server has been up for `COLLECTION_INTERVAL_IN_SEC` seconds.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/environment_metrics_service.py
+++ b/src/inmanta/server/services/environment_metrics_service.py
@@ -139,7 +139,6 @@ class EnvironmentMetricsService(protocol.ServerSlice):
         self.register_metric_collector(ResourceCountMetricsCollector())
         self.register_metric_collector(AgentCountMetricsCollector())
         self.register_metric_collector(CompileTimeMetricsCollector())
-        self.previous_timestamp = datetime.now()
         self.schedule(
             self.flush_metrics, COLLECTION_INTERVAL_IN_SEC, initial_delay=COLLECTION_INTERVAL_IN_SEC, cancel_on_stop=True
         )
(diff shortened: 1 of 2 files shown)


sanderr (approved):

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -139,7 +139,10 @@ async def start(self) -> None:
         self.register_metric_collector(ResourceCountMetricsCollector())
         self.register_metric_collector(AgentCountMetricsCollector())
         self.register_metric_collector(CompileTimeMetricsCollector())
-        self.schedule(self.flush_metrics, COLLECTION_INTERVAL_IN_SEC, initial_delay=0, cancel_on_stop=True)
+        self.previous_timestamp = datetime.now()
```
Hugo-Inmanta, line 142: What's the rationale behind setting this here as well as in the `__init__` ?

Hugo-Inmanta (approved):

src/inmanta/server/services/environment_metrics_service.py
arnaudsjs: The one in `__init__` is actually not required. Only the one here is required. But I didn't want to make `self.previous_timestamp` optional, so I set it in both locations.

arnaudsjs: Giving it a second though, this line is not really required here. The one in `__init__.py` should be sufficient. I will remove this one.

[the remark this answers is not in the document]
arnaudsjs: You can create a ticket for this.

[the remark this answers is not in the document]
sanderr: #5356

PR #5360, opened 2023-01-12
Issue/5356 metrics collectors should use timezone aware datetime objects

### LINKED ISSUE #5356: metrics collectors should use timezone-aware datetime objects
Since all collectors run in the same process as the server, I think they can all be considered internal and we should really make them all work with timezone-aware datetimes to match the ADR from #5335.

The docstring of the ABC should probably mention this.

Hugo-Inmanta:
# Description

closes #5356

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/db/migration_tests/dumps/v202301120.sql
+++ b/tests/db/migration_tests/dumps/v202301120.sql
@@ -10,7 +10,7 @@ SET lock_timeout = 0;
 SET idle_in_transaction_session_timeout = 0;
 SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
-SELECT pg_catalog.set_config('search_path', '', false);
+--SELECT pg_catalog.set_config('search_path', '', false);
 SET check_function_bodies = false;
 SET xmloption = content;
 SET client_min_messages = warning;
--- a/tests/db/migration_tests/dumps/v202301120.sql
+++ b/tests/db/migration_tests/dumps/v202301120.sql
@@ -1,0 +1,1196 @@
+--
+-- PostgreSQL database dump
+--
+
+-- Dumped from database version 12.12 (Ubuntu 12.12-0ubuntu0.20.04.1)
+-- Dumped by pg_dump version 12.12 (Ubuntu 12.12-0ubuntu0.20.04.1)
+
+SET statement_timeout = 0;
+SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
+SET client_encoding = 'UTF8';
+SET standard_conforming_strings = on;
+SELECT pg_catalog.set_config('search_path', '', false);
+SET check_function_bodies = false;
+SET xmloption = content;
+SET client_min_messages = warning;
+SET row_security = off;
+
+--
+-- Name: change; Type: TYPE; Schema: public; Owner: -
+--
+
+CREATE TYPE public.change AS ENUM (
+    'nochange',
+    'created',
+    'purged',
+    'updated'
+);
+
+
+--
+-- Name: non_deploying_resource_state; Type: TYPE; Schema: public; Owner: -
+--
+
+CREATE TYPE public.non_deploying_resource_state AS ENUM (
+    'unavailable',
+    'skipped',
+    'dry',
+    'deployed',
+    'failed',
+    'available',
+    'cancelled',
+    'undefined',
+    'skipped_for_undefined'
+);
+
+
+--
+-- Name: notificationseverity; Type: TYPE; Schema: public; Owner: -
+--
+
+CREATE TYPE public.notificationseverity AS ENUM (
+    'message',
+    'info',
+    'success',
+    'warning',
+    'error'
+);
+
+
+--
+-- Name: resource_id_version_pair; Type: TYPE; Schema: public; Owner: -
+--
+
+CREATE TYPE public.resource_id_version_pair AS (
+	resource_id character varying,
+	version integer
+);
+
+
+--
+-- Name: resourceaction_type; Type: TYPE; Schema: public; Owner: -
+--
+
+CREATE TYPE public.resourceaction_type AS ENUM (
+    'store',
+    'push',
+    'pull',
+    'deploy',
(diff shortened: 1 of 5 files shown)


Hugo-Inmanta:
This could probably use some tests but I don't see what to test for

tests/db/migration_tests/dumps/v202301120.sql
```diff
@@ -0,0 +1,1196 @@
...
+    'success',
+    'failed',
+    'deploying',
+    'pending'
+);
+
+
+SET default_tablespace = '';
+
+SET default_table_access_method = heap;
```
sanderr, line 118:
```suggestion
--SET default_table_access_method = heap;
```

tests/db/migration_tests/dumps/v202301120.sql
```diff
@@ -0,0 +1,1196 @@
...
+
+-- Dumped from database version 12.12 (Ubuntu 12.12-0ubuntu0.20.04.1)
+-- Dumped by pg_dump version 12.12 (Ubuntu 12.12-0ubuntu0.20.04.1)
+
+SET statement_timeout = 0;
+SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
+SET client_encoding = 'UTF8';
+SET standard_conforming_strings = on;
+SELECT pg_catalog.set_config('search_path', '', false);
```
sanderr, line 13:
```suggestion
--SELECT pg_catalog.set_config('search_path', '', false);
```

tests/db/migration_tests/test_v202301110_to_v202301120.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps/v202301110.sql"))
+async def test_timestamp_timezones(
+    migrate_db_from: abc.Callable[[], abc.Awaitable[None]],
+    postgresql_client: Connection,
+) -> None:
+    """
+    Test that all timestamps are timezone-aware.
+    """
+
+    # Fill the tables with dummy data to be able to fetch the types
```
sanderr, line 32: This is what we talked about on Tuesday: we should try to get out of this habit of prepopulating in the test case and instead do so in the dump tool.

If this proves difficult, for a trivial change like this, I think we can get away with leaving it untested. I wouldn't spend too much time on it.

If you really want to test this, it could be more useful to add a generic (non-migration) test that checks that all tables with time columns use timezones, in the same vein as `test_db_schema_enum_consistency`.

sanderr (changes requested):
Since `datetime` does not make a difference between naive and aware timestamps on the type level, I would add to the `MetricsCollector.get_metric_value` docstring's description of its two datetime parameters that these will be timezone-aware objects.

Other than that and a few minor comments this looks good to me. As to testing, I don't really see any useful scenarios to test either. Operations between a naive and an aware object produce a `TypeError`, so if there's anything wrong in that regard we'll catch it through other tests anyway.

sanderr:
Could you also add a second reviewer?

tests/db/migration_tests/dumps/v202301120.sql
Hugo-Inmanta: Damn it this slipped my mind

tests/db/migration_tests/test_v202301110_to_v202301120.py
Hugo-Inmanta: > This is what we talked about on Tuesday: we should try to get out of this habit of prepopulating in the test case and instead do so in the dump tool.


I don't clearly remember the discussion on Tuesday, prepopulating in the test case was causing some issues ? I don't see what the difference would be to do so in the dump tool.

> If this proves difficult, for a trivial change like this, I think we can get away with leaving it untested. I wouldn't spend too much time on it.

I tend to agree

> If you really want to test this, it could be more useful to add a generic (non-migration) test that checks that all tables with time columns use timezones, in the same vein as test_db_schema_enum_consistency.


I'll look into this :+1:

sanderr: > I don't see what the difference would be to do so in the dump tool.

The main difference is that it then becomes part of the dump, which is static. Then again, I think this approach only starts causing issues if you start using any code from `inmanta` (because that might change to reflect database changes which do not exist yet for this dump, rendering the code incompatible with the pre-state of the database). But you don't do that here, so maybe we should consider it acceptable in this case. @wouterdb wdyt? Should we discuss this special case next week?

wouterdb: I'll add some comments. 

https://github.com/inmanta/inmanta-core/pull/5362

tests/db/migration_tests/test_v202301110_to_v202301120.py
```diff
@@ -0,0 +1,72 @@
...
+            f"""
+                SELECT pg_typeof({column}) as type
+                FROM public.{table};
+            """
+        )
+        assert result[0]["type"] == type
+
+    type_pre = "timestamp without time zone"
+    for table in ["environmentmetricsgauge", "environmentmetricstimer"]:
+        await check_column_type(table, "timestamp", type_pre)
```
wouterdb, line 65: I don't see much value in this, I trust potsgresql to do this correctly.

tests/db/migration_tests/test_v202301110_to_v202301120.py
sanderr: But the migration script that introduced this table in the first place has never been released, so does it matter?

sanderr: > as we have never used this type of conversion before.

We have: https://github.com/inmanta/inmanta-core/blob/9f8de245b21765d064bc61f14e2397ddf64f602f/src/inmanta/db/versions/v202105170.py#L48

sanderr (approved):

wouterdb (approved):

PR #5362, opened 2023-01-12
added adr for DB upgrade

wouterdb:
# Description

added adr for DB upgrade

### DIFF
--- a/docs/adr/0000-database-upgrade-testing.md
+++ b/docs/adr/0000-database-upgrade-testing.md
@@ -6,7 +6,7 @@
 
 ## Context and Problem Statement
 
-When making updates to the database schema, it is import to test this.
+When making updates to the database schema, it is important to test this.
 
 ## Decision Drivers
 
--- a/docs/adr/0000-database-upgrade-testing.md
+++ b/docs/adr/0000-database-upgrade-testing.md
@@ -28,7 +28,7 @@ How do we set up these preconditions so that the tests remain valid and relevant
 
 We chose option 1 and 2.
 
-1. When data is very simple or specific, write data to the database in the testcase, prior to the update
+1. When data is very simple or specific, write data to the database in the testcase, prior to the update. Make sure not to use any inmanta code.
 2. When the write path is non trivial, use the procedure below:
  - creating test data using the old code,
  - dumping it to file and
--- a/docs/adr/0000-database-upgrade-testing.md
+++ b/docs/adr/0000-database-upgrade-testing.md
@@ -2,7 +2,7 @@
 
 * Status: accepted
 * Deciders: [Sander, Arnaud, Wouter]
-* Date: 12/01/2023
+* Date: 2023-01-12
 
 ## Context and Problem Statement
 
--- a/docs/adr/0000-database-upgrade-testing.md
+++ b/docs/adr/0000-database-upgrade-testing.md
@@ -43,7 +43,7 @@ In practice, this means that if you write a database update test:
 
 ### Negative Consequences
 
-* Easy to do wrong. It is more natural to set up the inital state in the testcase itself, as we do everywhere else.
+* Easy to do wrong. It is more natural to set up the initial state in the testcase itself, as we do everywhere else.
 * We accumulate some additional test data in the repo
 
 ## Pros and Cons of the Options
--- a/docs/adr/0000-database-upgrade-testing.md
+++ b/docs/adr/0000-database-upgrade-testing.md
@@ -1,0 +1,60 @@
+# Database Upgrade Testing
+
+* Status: accepted
+* Deciders: [Sander, Arnaud, Wouter]
+* Date: 12/01/2023
+
+## Context and Problem Statement
+
+When making updates to the database schema, it is import to test this.
+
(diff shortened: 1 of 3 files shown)


arnaudsjs (approved):

docs/adr/0000-database-upgrade-testing.md
```diff
@@ -0,0 +1,60 @@
+# Database Upgrade Testing
+
+* Status: accepted
+* Deciders: [Sander, Arnaud, Wouter]
+* Date: 12/01/2023
+
+## Context and Problem Statement
+
+When making updates to the database schema, it is important to test this.
```
sanderr, line 9: I think we should expand a bit on this. We don't just want to test the migrations (we already do that), we want a uniform approach to set up any relevant preconditions for the tests. Since the considered options focus on that part, I think the problem statement should as well.

docs/adr/0000-database-upgrade-testing.md
```diff
@@ -0,0 +1,60 @@
+# Database Upgrade Testing
+
+* Status: accepted
+* Deciders: [Sander, Arnaud, Wouter]
+* Date: 12/01/2023
```
sanderr, line 5: The template contains a date format, I think it could get confusing if we start using different ones across ADRs. I have no objection against changing the format in the template either if you prefer.
```suggestion
* Date: 2023-01-12
```

FloLey (approved):

docs/adr/0000-database-upgrade-testing.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Considered Options
+
+1. Write data corresponding to the old database schema directly in the testcase to the DB.
+2. Generate test data using old code and persist it.
+3. Dynamically check out old code to run the test
+
+## Decision Outcome
+
+We chose option 2 to test upgrades by
```
sanderr, line 24: As we saw in @Hugo-Inmanta 's PR there are two flavors of this option:
a. we always do it like this
b. we are flexible and allow prepopulation in the test case iff it doesn't use any code from `inmanta`.  Directly accessing the database is always safe, since we're operating on a dump that reflects the old data and schema.

Option b could serve to limit pollution of the dump tool for very specific edge cases, and it might simplify the test by making them more directed (they put something in and make sure it comes out as expected, focusing only on that representative subset).
I would consider this to be the best of both worlds.

docs/adr/0000-database-upgrade-testing.md
wouterdb: I personally care much less about uniform and more about correct. 

Could you make a suggestion?

sanderr: That makes sense. Either way, the part where some of us made mistakes against, is the precondition set up, not the testing itself. A suggestion:
```suggestion
When making updates to the database schema, it is important to test this. In case the migration includes any data
transformation, testing this commonly involves setting up some known preconditions, and verifying that they are
converted as expected when the migration script is executed. These preconditions should be set up in a way that
remains both valid and relevant for the test scenario, even when related code may change in the future.

How do we set up these preconditions so that the tests remain valid and relevant in the future?
```

And then perhaps a small addition to the decision drivers. I'll make a separate suggestion there.

docs/adr/0000-database-upgrade-testing.md
```diff
@@ -0,0 +1,60 @@
...
+* Date: 2023-01-12
+
+## Context and Problem Statement
+
+When making updates to the database schema, it is important to test this.
+
+## Decision Drivers
+
+* We want to test that database update code correctly transforms data produced by older code to data that can be handled by the current code
+* But within the code base, this older code no longer exists
```
sanderr, line 14:
```suggestion
* But within the code base, this older code no longer exists or has been modified, so it can not be relied upon in the test case before the migration, where database schema and code might be out of sync.
```
Part of https://github.com/inmanta/inmanta-core/pull/5362/files#r1069134210

docs/adr/0000-database-upgrade-testing.md
wouterdb: added

sanderr: I don't see it. Forgot to push?

wouterdb: :facepalm:

docs/adr/0000-database-upgrade-testing.md
```diff
@@ -0,0 +1,68 @@
...
+
+1. Write data corresponding to the old database schema directly in the testcase to the DB.
+2. Generate test data using old code and persist it.
+3. Dynamically check out old code to run the test
+
+## Decision Outcome
+
+We chose option 1 and 2.
+
+1. When data is very simple or specific, write data to the database in the testcase, prior to the update
```
sanderr, line 31:
```suggestion
1. When data is very simple or specific, write data to the database in the testcase, prior to the update. Make sure not to use any inmanta code.
```

sanderr (approved):

Hugo-Inmanta (approved):

PR #5374, opened 2023-01-13
Issue/5358 metrics always datapoint

### LINKED ISSUE #5358: Metrics collectors should always write a datapoint for each environment

FloLey:
# Description

When the database doesn't contain any record for the monitored element, no datapoint is written to the database. For example, when a given environment does not have any configuration models, the ResourceCountMetricsCollector will not return any MetricValue for that environment, while it should actually return the value 0 instead.

Tasks:

 Fix this bug for all collectors in inmanta-core that suffer from this bug.

closes #5358 

# Self Check:

- [x] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -5444,6 +5444,7 @@ class EnvironmentMetricsGauge(BaseDocument):
     :param environment: the environment to which this metric is related
     :param metric_name: The name of the metric
     :param timestamp: The timestamps at which a new record is created
+    :grouped_by: the field on which the metrics is grouped by
```
sanderr, line 5447: I don't find this message all that clear.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -277,12 +277,20 @@ async def get_metric_value(
...
-            SELECT r.environment, r.status, COUNT(*)
-            FROM latest_released_resources AS r
-            GROUP BY r.environment, r.status
-            ORDER BY r.environment, r.status
-        """
+                WITH {LATEST_RELEASED_MODELS_SUBQUERY}, nonzero_statuses AS (
+                    SELECT r.environment, r.status, COUNT(*) AS count
+                    FROM {Resource.table_name()} AS r
+                    INNER JOIN latest_released_models AS cm
+                    ON r.environment = cm.environment AND r.model = cm.version
```
sanderr, line 284: Sorry about the conflict. I think we can keep a bit more of my changes though, and get the best of both worlds.
```suggestion
                WITH {LATEST_RELEASED_RESOURCES_SUBQUERY}, nonzero_statuses AS (
                    SELECT r.environment, r.status, COUNT(*) AS count
                    FROM latest_released_resources AS r
```

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -360,11 +368,19 @@ async def get_metric_value(
...
-            GROUP BY environment
+            WITH compile_time as (
+                SELECT count(*) as count, environment, sum(completed-started)
+                FROM {Compile.table_name()}
+                WHERE completed >= $1
+                AND completed < $2
+                GROUP BY environment
+            ) SELECT e.id AS environment,
+            COALESCE(compile_time.count, 0) AS count,
+            COALESCE(compile_time.sum, INTERVAL '0' SECOND) as compile_time
```
sanderr, line 379: I think some indentation could help readability here.
```suggestion
            )
            SELECT e.id AS environment,
                COALESCE(compile_time.count, 0) AS count,
                COALESCE(compile_time.sum, INTERVAL '0' SECOND) AS compile_time
```

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -277,12 +277,20 @@ async def get_metric_value(
...
-        """
+                WITH {LATEST_RELEASED_MODELS_SUBQUERY}, nonzero_statuses AS (
+                    SELECT r.environment, r.status, COUNT(*) AS count
+                    FROM {Resource.table_name()} AS r
+                    INNER JOIN latest_released_models AS cm
+                    ON r.environment = cm.environment AND r.model = cm.version
+                    GROUP BY r.environment, r.status
+                )
+                SELECT e.id as environment, s.name as status, COALESCE(r.count, 0) as count
+                FROM public.environment AS e
```
sanderr, line 288:
```suggestion
                FROM {Environment.table_name()} AS e
```

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -400,11 +416,17 @@ async def get_metric_value(
...
     ) -> Sequence[MetricValueTimer]:
         query: str = f"""
-            SELECT count(*)  as count,environment,sum(started-requested) as compile_waiting_time
+        WITH compile_waiting_time as (
+            SELECT count(*) as count,environment,sum(started-requested)
             FROM {Compile.table_name()}
             WHERE started >= $1
             AND started < $2
             GROUP BY environment
+        ) SELECT e.id AS environment, COALESCE(compile_waiting_time.count, 0) AS count, COALESCE(compile_waiting_time.sum, INTERVAL '0' SECOND) as compile_waiting_time
```
sanderr, line 425: In this context 0 feels weird. I also see in the design "The data returned to the client should contain the value NULL for the sub-intervals for which no data is available.", though the context is slightly different here. I wonder if we should use `null` instead of 0 for all metrics (if we do, we should make sure to keep 0 for absent statuses, and only use `null` for environments without data)? For e.g. the count metrics 0 does make a lot of sense. I need to think this over.

@arnaudsjs cc

sanderr (changes requested):

src/inmanta/data/__init__.py
FloLey: how would you explain it? The name is self explanatory to me :p 
Should I add an example, like:
For example the 'agent_count' metric is grouped_by "status".

src/inmanta/server/services/environment_metrics_service.py
FloLey: At first glance I like your proposal.

src/inmanta/data/__init__.py
sanderr: Well, the concept of grouping is just a method to do the aggregation in SQL, so in the context of the API, it took me a while to see what it referred to, because it has little meaning there, does it? For example, a field called `status` would be a lot more meaningful in my opinion, because that is exactly what it is, while the grouping is just a means to an end.

sanderr: It is of course possible I'm missing something. I haven't seen all the code for the metrics.

FloLey: We could be grouping on something else than "status" for another metric. This fields indicates which fields (in addition to the environment) was used in the sql query to aggregate the data. (Maybe this is how I should explain it :p )

sanderr: Yes, I get that, but I did overlook something else that is now clear to me. I still find it very strange but I'm having difficulty explaining it. I'll get back to this next week.

src/inmanta/server/services/environment_metrics_service.py
arnaudsjs: I think the main question here is: How do we want to represent this scenario in the graph on the web-console?

* If we want to represent this using the value 0, we should write a value 0 into the database.
* If we want to represent this using an empty spot in the graph, we should not write a 0.

Personally, I don't have a strong opinion this. But if we would go for the second approach, I am wondering whether we should write anything at all to the database. As far a I can see, we want the same behavior in bullet point 2 as when the server is offline and no data is written to the database. Do I see that correctly?

src/inmanta/data/__init__.py
sanderr: So, with some time to let this sink in, I feel less strongly about it. The main thing that I still find unintuitive about the name (that we can clear up in the docstring) is that semantically, afaik, you generally group by "some generic name/type", which gives you categories with specific values for the thing. e.g. suppose you have something with colors, you say you group by color, giving you categories red, green, blue. You don't say you group by red.

This is why in the context of a single metric value, I find the name non trivial: in natural language it doesn't make sense in this context, so if you don't know up front what it means, it might take a while to make the connection. Therefore, I think a natural language docstring could help clarify this. e.g. "The name of the group/category this metric represents (e.g. red if grouped by color). None iff metrics of this type are not divided in groups."

I realize I wasn't making much sense before. I hope this is more clear. What do you think?

FloLey: yes! now I see what you mean and how it could be confusing. I'll change the docstring

Hugo-Inmanta: I have been following this from a distance, what about "group" or "category" instead of "grouped_by" ?
Also what about multiple aggregation ? Afaik we are supporting it so it might be worth adding to the docstring ?

sanderr: > what about "group" or "category" instead of "grouped_by"

I would honestly prefer that. The only reason I didn't propose it myself is because I didn't want to be too heavy on suggesting changes to existing code. But if it's feasible and everyone agrees I'm definitely in favor.

>  Also what about multiple aggregation ?

What exactly do you mean by this? To have a second axis to group by? I'm not sure we do support that at the moment.

src/inmanta/server/services/environment_metrics_service.py
sanderr: Let me start by saying I agree with your conclusion that if we pick option 2, we don't need to write anything at all.

Now, for some metrics (e.g. any of the count metrics) I would say option 1 makes more sense: a newly created environment really does have 0 instances, resources and relevant agents. For compile rate we could argue the same. For all of those, absence naturally translates to zero. But then for compile time and compile waiting time this is not the case: adding zeros may skew front-end aggregations, potentially giving the false impression of a compile time regression.

Therefore, I would propose we extend the docstring of the method on the ABC with something along the lines of "If, at the time of collection, no metric data exists for an environment and/or group, the implementation should return a meaningful default  (e.g. 0 for count metrics). If no meaningful default exists (e.g. for some time metrics), the data may be left out."

This would allow us to make a guided choice for each metric individually. In our current case, we could leave the data out for the compile time and waiting time metrics.

What do you think?

@Hugo-Inmanta @wouterdb CC, since this is turning into a design discussion rather than just a PR review. If you need more context, let me know.

arnaudsjs: I agree.

src/inmanta/data/__init__.py
FloLey: should be feasible I think, and I have no problem changing it :)

Hugo-Inmanta: > To have a second axis to group by?
Yes, I might be completely off here, but when defining a `get_metric_value` method, what prevents us from returning some  MetricValue(grouped_by="a,b") for example ?

sanderr: Nothing prevents you, but do we want to endorse that? Either we don't officially support multiple aggregation for now, or we need to do it right. In the second case it's probably better to go for a `tuple[str]` than `str`, but since we don't need it at the moment (and I don't immediately see a future use case), I feel like it would be an overengineered solution (even if it would make the migration simpler if ever we do want to add support). So while I agree with you that it could be done, I'm not sure we should.

sanderr: (cc @arnaudsjs, from Hugo's first comment (5 up) on, the rest has been settled)

src/inmanta/server/services/environment_metrics_service.py
FloLey: this seems like to most flexible and right solution. I'll change it.

tests/db/migration_tests/test_v202212010_to_v202301100.py
```diff
@@ -38,8 +38,5 @@ async def test_added_environment_metrics_tables(
     await migrate_db_from()
 
     for table_name in ["environmentmetricsgauge", "environmentmetricstimer"]:
-        assert "grouped_by" in (await get_columns_in_db_table(table_name))
-
         columns_pk = await get_primary_key_columns_in_db_table(table_name)
         assert len(columns_pk) == 4
-        assert "grouped_by" in columns_pk
```
FloLey, line 45: @wouterdb FYI: We decided to rename a column in the metrics tables from "grouped_by" to "category" which broke the migration test where this column was added. I removed the assertion for this reason.

src/inmanta/server/services/environment_metrics_service.py
sanderr: I now realize the compile time and waiting time metrics contain a count as well, so writing zero would not have been so bad after all, since it's up to the collection endpoint to do correct aggregation anyway. Either way, I don't think leaving it out does any harm.

tests/db/migration_tests/test_v202212010_to_v202301100.py
sanderr: (This old migration script has never been released and will only be released together with this new one.)

tests/db/migration_tests/test_v202301110_to_v202301160.py
```diff
@@ -0,0 +1,44 @@
...
+    limitations under the License.
+    Contact: code@inmanta.com
+"""
+import os
+from collections import abc
+
+import pytest
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps/v202301160.sql"))
```
sanderr, line 20:
```suggestion
@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps/v202301100.sql"))
```

Your dump seems to be for the old schema. Did you generate it before you wrote the migration script?

sanderr (changes requested):

tests/db/migration_tests/test_v202301110_to_v202301160.py
FloLey: I generated it on the master branch, as it is what I understood should be done from the adr. But I don't understand why I should use v202301100.sql? I thought the Idea was to create the dump before the changes and to migrate from that dump in the migration test?

sanderr (approved):

changelogs/unreleased/5358-metrics-always-datapoint.yml
```diff
@@ -0,0 +1,6 @@
+description: make sur a datapoint is written to the database even if no records are present for the monitored element.
```
Hugo-Inmanta, line 1:
```suggestion
description: Make sure a datapoint is written to the database even if no records are present for the monitored element.
```

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -153,6 +153,10 @@ async def get_metric_value(
         the database. No in-memory state is being stored by this metrics collector. The provided interval
         should be interpreted as [start_interval, end_interval[
 
+        If, at the time of collection, no metric data exists for an environment and/or group,
```
Hugo-Inmanta, line 156: Maybe for consistency use "category" here instead of "group" ? Not a strong opinion

Hugo-Inmanta (approved):

PR #5379, opened 2023-01-16
issue/5359-flaky-test-metrics

### LINKED ISSUE #5359: flaky test for metrics collectors
currently all the test for metric collector do the following steps (on core and lsm):

1. starts the server
2. flushes metrics manually
3. assert that the metrics are collected

This is potentially flaky as the server will also flush at regular intervals. We can never be sure amount the amount of metrics that are present in the metrics tables of the database when the assertions are executed.

We should find a way to remove this uncertainty.  
maybe we can add an option to disable the flush at regular intervals or a fixture that disables this for a specific test case.

FloLey:
# Description

Currently all the test for metric collector do the following steps:

1. starts the server
2. flushes metrics manuall
3. assert that the metrics are collected
This is potentially flaky as the server will also flush at regular intervals. We can never be sure amount the amount of metrics that are present in the metrics tables of the database when the assertions are executed.

this PR disables the automatic flush done at regular intervals

closes #5359

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/environment_metrics_service.py
+++ b/src/inmanta/server/services/environment_metrics_service.py
@@ -181,9 +181,11 @@ class EnvironmentMetricsService(protocol.ServerSlice):
         self.register_metric_collector(CompileWaitingTimeMetricsCollector())
         self.register_metric_collector(AgentCountMetricsCollector())
         self.register_metric_collector(CompileTimeMetricsCollector())
-        self.schedule(
-            self.flush_metrics, COLLECTION_INTERVAL_IN_SEC, initial_delay=COLLECTION_INTERVAL_IN_SEC, cancel_on_stop=True
-        )
+        if COLLECTION_INTERVAL_IN_SEC != 0:
+            print("OLA")
+            self.schedule(
+                self.flush_metrics, COLLECTION_INTERVAL_IN_SEC, initial_delay=COLLECTION_INTERVAL_IN_SEC, cancel_on_stop=True
+            )
 
     def register_metric_collector(self, metrics_collector: MetricsCollector) -> None:
         """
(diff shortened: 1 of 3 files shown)


src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -181,9 +181,10 @@ async def start(self) -> None:
         self.register_metric_collector(CompileWaitingTimeMetricsCollector())
         self.register_metric_collector(AgentCountMetricsCollector())
         self.register_metric_collector(CompileTimeMetricsCollector())
-        self.schedule(
-            self.flush_metrics, COLLECTION_INTERVAL_IN_SEC, initial_delay=COLLECTION_INTERVAL_IN_SEC, cancel_on_stop=True
-        )
+        if COLLECTION_INTERVAL_IN_SEC != 0:
```
arnaudsjs, line 184:
```suggestion
        if COLLECTION_INTERVAL_IN_SEC > 0:
```

tests/server/test_environement_metrics.py
```diff
@@ -47,6 +48,15 @@ async def env_metrics_service(server_config, init_dataclasses_and_load_schema) -
     yield metrics_service
 
 
+@pytest.fixture
+def server_pre_start(server_config):
+    """This fixture is called before the server starts to disable the auotmatical flush_metrics done by the scheduler"""
```
arnaudsjs, line 53:
```suggestion
    """This fixture is called before the server starts to disable the automatic flush_metrics done by the EnvironmentMetricsService"""
```

arnaudsjs (changes requested):

arnaudsjs (approved):

Hugo-Inmanta (approved):

PR #5380, opened 2023-01-16
Issue/5048 mypy strict equality

### LINKED ISSUE #5048: Enable mypy --strict-equality check
This ticket was spawned by [this discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1666358794418509)

At the time of writing, the following type errors are raised:
```bash
(inmanta-core-39) sander@bedevere:~/documents/projects/inmanta/inmanta-core$ make mypy-diff 
+ src/inmanta/reporter.py:140: error: Non-overlapping identity check (left operand type: "Type[float]", right operand type: "Type[str]")  [comparison-overlap]
+ src/inmanta/config.py:212: error: Non-overlapping equality check (left operand type: "Type[str]", right operand type: "Type[bool]")  [comparison-overlap]
+ src/inmanta/data/__init__.py:5780: error: Non-overlapping container check (element type: "Dict[str, Any]", container item type: "str")  [comparison-overlap]
+ src/inmanta/loader.py:481: error: Non-overlapping container check (element type: "PluginModuleFinder", container item type: "_MetaPathFinder")  [comparison-overlap]
+ src/inmanta/execute/scheduler.py:707: error: Non-overlapping container check (element type: "RelationPrecedenceGraphNode", container item type: "RelationAttribute")  [comparison-overlap]
```
TODO:
- [x] enable mypy --strict-equality check
- [x] fix errors

sanderr:
# Description

- Moved mypy config to pyproject.toml
- Added `strict_equality = true`
- Fixed fallout
  - small bug in cycle detection for relation precedence policy
  - `inmanta.config.is_bool` used a private library method, I changed it to use equivalent behavior through the public API instead.
  - v1 module finder used deprecated `find_module`. I implemented `find_spec` instead.
  - some incorrect type annotations

Resulting diff below (red is good). There is one new error but I decided not to fix it because it is due to incorrect typing in typeshed, which has already been fixed in the latest mypy nightlies, so it should disappear soon.
```diff
- src/inmanta/config.py:214: error: Returning Any from function declared to return "bool"  [no-any-return]
- src/inmanta/config.py:214: error: "ConfigParser" has no attribute "_convert_to_boolean"  [attr-defined]
- src/inmanta/loader.py:483: error: Argument 1 to "remove" of "list" has incompatible type "PluginModuleFinder"; expected "_MetaPathFinder"  [arg-type]
+ src/inmanta/loader.py:510: error: Argument 2 of "find_spec" is incompatible with supertype "MetaPathFinder"; supertype defines the argument type as "Optional[Sequence[Union[bytes, str]]]"  [override]
+ src/inmanta/loader.py:510: note: This violates the Liskov substitution principle
+ src/inmanta/loader.py:510: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
- src/inmanta/data/__init__.py:5172: error: Argument 1 to "list" has incompatible type "Set[str]"; expected "Iterable[Dict[str, Any]]"  [arg-type]
- src/inmanta/data/__init__.py:5184: error: Invalid index type "Dict[str, Any]" for "Dict[str, List[str]]"; expected type "str"  [index]
- src/inmanta/data/__init__.py:5185: error: Argument 1 to "extend" of "list" has incompatible type "List[str]"; expected "Iterable[Dict[str, Any]]"  [arg-type]
- src/inmanta/data/__init__.py:5189: error: Incompatible return value type (got "Tuple[Set[str], Set[Any]]", expected "Tuple[Set[ResourceIdStr], Set[ResourceIdStr]]")  [return-value]
- src/inmanta/server/services/resourceservice.py:316: error: Incompatible types in assignment (expression has type "Tuple[Set[ResourceIdStr], Set[ResourceIdStr]]", variable has type "Optional[Tuple[Set[ResourceVersionIdStr], List[ResourceVersionIdStr]]]")  [assignment]
- src/inmanta/server/services/resourceservice.py:319: error: "None" object is not iterable  [misc]
```

Closes #5048

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5048-mypy-strict-equality.yml
+++ b/changelogs/unreleased/5048-mypy-strict-equality.yml
@@ -1,0 +1,8 @@
+description: "Enabled mypy's `strict_equality` option"
+change-type: patch
+issue-nr: 5380
+sections:
+  bugfix: Fixed cycle detection in experimental relation precedence policy
+destination-branches:
+  - master
+  - iso6
--- a/stubs/pyformance/__init__.pyi
+++ b/stubs/pyformance/__init__.pyi
@@ -1,12 +1,13 @@
-from typing import Callable, Dict, Optional
+from collections import abc
+from typing import Optional
 
 from pyformance.meters import Gauge
 
 
 class MetricsRegistry:
     def __init__(self, clock): ...
-    def dump_metrics(self)-> Dict[str, Dict[str, float]]: ...
+    def dump_metrics(self)-> abc.Mapping[str, abc.Mapping[str, int|float|str]]: ...
 
-global_registry: Callable[[], MetricsRegistry]
+global_registry: abc.Callable[[], MetricsRegistry]
 
 def gauge(key: str, gauge: Optional[Gauge]=None) -> Gauge: ...
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,33 @@ exclude = '''
 )/
 '''
 
+[tool.mypy]
+plugins = [
+    'pydantic.mypy'
+]
+check_untyped_defs = true
+disallow_any_generics = true
+disallow_incomplete_defs = true
+disallow_subclassing_any = true
+disallow_untyped_calls = true
+disallow_untyped_decorators = true
+disallow_untyped_defs = true
+follow_imports = 'silent'
+no_implicit_optional = true
+no_implicit_reexport = true
+show_error_codes = true
+strict_optional = true
+strict_equality = true
+warn_redundant_casts = true
+warn_return_any = true
+warn_unused_configs = true
+warn_unused_ignores = true
+
+[tool.pydantic-mypy]
+init_forbid_extra = true
+init_typed = true
+warn_untyped_fields = true
+
 [tool.irt.build.python]
 hooks = { 'pre' = 'pre-build.sh' }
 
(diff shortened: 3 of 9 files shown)


stubs/pyformance/__init__.pyi
```diff
@@ -1,12 +1,13 @@
...
+from collections import abc
+from typing import Optional
 
 from pyformance.meters import Gauge
 
 
 class MetricsRegistry:
     def __init__(self, clock): ...
-    def dump_metrics(self)-> Dict[str, Dict[str, float]]: ...
+    def dump_metrics(self)-> abc.Mapping[str, abc.Mapping[str, int|float|str]]: ...
```
Hugo-Inmanta, line 9: :exploding_head: I like it much more than Union(...) ! I noticed it is specific to python 3.10 and this sparked a question: how do we decide which version of the product uses which python version ?

Hugo-Inmanta (approved):

stubs/pyformance/__init__.pyi
sanderr: There are two parts to this:
1. afaik stubs are only for the type checker so we can cheat a little: mypy knows about `|` even if Python does not.
2. Generally we decide based on what's available on the lowest RHEL version we need to support for a product version. For both iso5 and iso6 that's RHEL 8, which doesn't include Python 3.10 yet, so we're stuck with 3.9 for now. For iso7, we might be able to upgrade to 3.10 or 3.11 eventually, if RHEL 9 includes them (I expect it will). We always have an officially supported version, and we try to be compatible with higher versions as well but we provide no such guarantees.

changelogs/unreleased/5048-mypy-strict-equality.yml
```diff
@@ -0,0 +1,8 @@
+description: "Enabled mypy's `strict_equality` option"
+change-type: patch
+issue-nr: 5380
```
arnaudsjs, line 3:
```suggestion
issue-nr: 5048
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #5391, opened 2023-01-17
Issue/5347 compile time metrics collector is missing an index in db

### LINKED ISSUE #5347: CompileTimeMetricsCollector is missing index in DB
The `CompileTimeMetricsCollector` executes a query that performs a sequential scan over the `Compile` table. PostgreSQL has no alternative but to do a sequential scan, because there is no index on the completed column.

```
EXPLAIN ANALYZE SELECT count(*) as count, environment, sum(completed-started) as compile_time FROM public.compile WHERE completed >= (now() - INTERVAL '5 MINUTES') AND completed < now() GROUP BY environment;
                                                  QUERY PLAN                                                   
---------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=24.50..24.53 rows=1 width=40) (actual time=0.120..0.120 rows=0 loops=1)
   Group Key: environment
   ->  Sort  (cost=24.50..24.51 rows=1 width=32) (actual time=0.118..0.119 rows=0 loops=1)
         Sort Key: environment
         Sort Method: quicksort  Memory: 25kB
         ->  Seq Scan on compile  (cost=0.00..24.49 rows=1 width=32) (actual time=0.112..0.113 rows=0 loops=1)
               Filter: ((completed < now()) AND (completed >= (now() - '00:05:00'::interval)))
               Rows Removed by Filter: 243
 Planning Time: 0.152 ms
 Execution Time: 0.158 ms
(10 rows)

```

Hugo-Inmanta:
# Description
Add index on public.compile table, for [environment, completed]

closes #5347 
# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/db/versions/v202301170.py
+++ b/src/inmanta/db/versions/v202301170.py
@@ -21,4 +21,4 @@ from asyncpg import Connection
 
 async def update(connection: Connection) -> None:
     # Add index to avoid a sequential scan in CompileTimeMetricsCollector.get_metric_value query
-    await connection.execute("CREATE INDEX ON public.compile (environment, completed DESC)")
+    await connection.execute("CREATE INDEX ON public.compile (completed)")
--- a/src/inmanta/db/versions/v202301170.py
+++ b/src/inmanta/db/versions/v202301170.py
@@ -21,4 +21,4 @@ from asyncpg import Connection
 
 async def update(connection: Connection) -> None:
     # Add index to avoid a sequential scan in CompileTimeMetricsCollector.get_metric_value query
-    await connection.execute("CREATE INDEX ON public.compile (completed)")
+    await connection.execute("CREATE INDEX ON public.compile (completed, environment)")
--- a/src/inmanta/db/versions/v202301170.py
+++ b/src/inmanta/db/versions/v202301170.py
@@ -1,0 +1,24 @@
+"""
+    Copyright 2023 Inmanta
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    Contact: code@inmanta.com
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    # Add index to avoid a sequential scan in CompileTimeMetricsCollector.get_metric_value query
+    await connection.execute("CREATE INDEX ON public.compile (environment, completed DESC)")
--- a/changelogs/unreleased/5347-compile-time-metrics-collector-is-missing-an-index-in-DB.yml
+++ b/changelogs/unreleased/5347-compile-time-metrics-collector-is-missing-an-index-in-DB.yml
@@ -1,4 +1,4 @@
 description: Add index on compile table.
 issue-nr: 5347
-change-type: minor
+change-type: patch
 destination-branches: [master, iso6]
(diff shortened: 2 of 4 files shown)


src/inmanta/db/versions/v202301170.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Contact: code@inmanta.com
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    # Add index to avoid a sequential scan in CompileTimeMetricsCollector.get_metric_value query
+    await connection.execute("CREATE INDEX ON public.compile (environment, completed DESC)")
```
arnaudsjs, line 24: The query will perform better if you turn around the columns in the index. Remember that PostgreSQL first executes the WHERE clause and only then the GROUP BY.
```suggestion
    await connection.execute("CREATE INDEX ON public.compile (completed, environment)")
```

arnaudsjs (changes requested):

src/inmanta/db/versions/v202301170.py
arnaudsjs: I have been thinking about this a bit more and I even think we can remove the `environment` column from the index. This column is only useful when we expect multiple compiles to finish at the same time within a given environment. Given that the compiler services executes all compiles sequentially, this can never happen in practice. On the other hand, if we ever decide to allow concurrent compiles within an environment in the future, we will probably forget to update this index. So that might be an argument to add the environment column after all.

I tested both scenario's using the infra model and the cost of the queries seems to go down when we remove the environment column from the index:

With index on (completed, environment):
```
 GroupAggregate  (cost=8.30..8.32 rows=1 width=40) (actual time=0.071..0.072 rows=0 loops=1)
   Group Key: environment
   ->  Sort  (cost=8.30..8.30 rows=1 width=32) (actual time=0.069..0.070 rows=0 loops=1)
         Sort Key: environment
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using compile_completed_environment_idx on compile  (cost=0.27..8.29 rows=1 width=32) (actual time=0.040..0.040 rows=0 loops=1)
               Index Cond: ((completed >= '2023-01-13 13:06:00.502389'::timestamp without time zone) AND (completed < '2023-01-13 14:06:00.502389'::timestamp without time zone))
 Planning Time: 0.770 ms
 Execution Time: 0.225 ms
(9 rows)
```

with index on (completed):
```
 GroupAggregate  (cost=8.17..8.20 rows=1 width=40) (actual time=0.021..0.021 rows=0 loops=1)
   Group Key: environment
   ->  Sort  (cost=8.17..8.18 rows=1 width=32) (actual time=0.020..0.020 rows=0 loops=1)
         Sort Key: environment
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using compile_completed_idx on compile  (cost=0.14..8.16 rows=1 width=32) (actual time=0.011..0.011 rows=0 loops=1)
               Index Cond: ((completed >= '2023-01-13 13:06:00.502389'::timestamp without time zone) AND (completed < '2023-01-13 14:06:00.502389'::timestamp without time zone))
 Planning Time: 0.247 ms
 Execution Time: 0.066 ms
(9 rows)
```

@sanderr Do you have any opinion on this?

Hugo-Inmanta: Oh wow this looks much much faster indeed. Could it be that it performs worse on small dataset but better on bigger ones ?

Also about `DESC`, my rationale was that we will often (at every flush interval) query using this index with timestamps parameters centered in time around "now()" so it made sense to have the biggest timestamps near the top. Could it be of any use or am I off ?

arnaudsjs: I don't think adding the `DESC` makes a difference. As far as I know, it's only useful if you want the end result to be sorted in a specific way. But we don't need that here.

arnaudsjs: BTW: For the example above, it looks like the single-column index performs a lot better than the multi-column index, but that's not really true. There is just a lot of jitter on the execution time of the query if you execute it repeatedly. So the performance gain is not as big as the example indicates, but in general the single-column index performs better than the multi-column index.

sanderr: 1. about the single  vs multi column index: I think the multi-column one would become more efficient if we had a lot of environments. But since that isn't a realistic use case, I have no objection against using the single column index.
2. about `DESC`: as far as I know Arnaud is right. To add to that, for single column indexes it doesn't even matter at all: an index is as cheap to traverse in either direction. The reason it matters for multi column indexes is that the columns form a composite key: if you have an index on `x, y` you can cheaply traverse `x, y` or `x DESC, y DESC`. If you want `x, y DESC` you'd need an index on either that or on `x DESC, y`.

arnaudsjs (approved):

PR #5395, opened 2023-01-18
Add support to cleanup old environment metrics

### LINKED ISSUE #5130: Add support to cleanup old metrics
This ticket should make sure that old metrics are cleaned up by a background job.

Tasks:

- [x] Add environment setting `metrics_retention_interval` (default value: 1 year)
- [x] Setup a background task that cleans up old metrics every hour.

arnaudsjs:
# Description

Add support to cleanup old environment metrics using the `environment_metrics_retention` environment setting.

closes #5130 

# Self Check:

- [x] Changelog entry


src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -185,6 +188,34 @@ async def start(self) -> None:
...
+                delete_order_than: datetime = datetime.now().astimezone() - timedelta(hours=rentention_time_in_hours)
+                env_id_to_delete_older_than_timestamp[env.id] = delete_order_than
+
+            filter_condition = " OR ".join(
+                f"(environment=${(2*i)+1} AND timestamp < ${(2*i)+2})"
+                for i in range(len(env_id_to_delete_older_than_timestamp))
+            )
+            values = list(itertools.chain.from_iterable(env_id_to_delete_older_than_timestamp.items()))
+            await con.execute(f"DELETE FROM {EnvironmentMetricsGauge.table_name()} WHERE {filter_condition}", *values)
+            await con.execute(f"DELETE FROM {EnvironmentMetricsTimer.table_name()} WHERE {filter_condition}", *values)
```
Hugo-Inmanta, line 218: I wonder if this still behaves well when `env_id_to_delete_older_than_timestamp` only has 0 or 1 element in it and the query reads
```sql
(...) WHERE ;
```

or

```sql
(...) WHERE OR (...) ;
```

Hugo-Inmanta (changes requested):

src/inmanta/server/services/environment_metrics_service.py
arnaudsjs: You are right. This will cause problems where the server doesn't have any environment. Thanks!

Hugo-Inmanta: Just realized I was wrong about the 1 element case

arnaudsjs: Indeed. The 1 element case is not an issue.

Hugo-Inmanta (approved):
Looks good to me :+1: !

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -189,6 +192,36 @@ async def start(self) -> None:
             self.schedule(
                 self.flush_metrics, COLLECTION_INTERVAL_IN_SEC, initial_delay=COLLECTION_INTERVAL_IN_SEC, cancel_on_stop=True
             )
+            # Cleanup metrics once per hour
+            self.schedule(self._cleanup_old_metrics, interval=3600, initial_delay=0, cancel_on_stop=True)
```
sanderr, line 196: Shouldn't we enable this always, even if collection has been disabled? I don't have a strong opinion but it would feel slightly more intuitive to me.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -189,6 +192,36 @@ async def start(self) -> None:
...
+
+    async def _cleanup_old_metrics(self) -> None:
+        """
+        Clean up metrics that are older than the retention time specified in the environment_metrics_retention
+        environment setting.
+        """
+        async with Environment.get_connection() as con:
+            environments = await Environment.get_list(connection=con)
```
sanderr, line 204: Couldn't we do all this in SQL? Or don't you think it would be worth it?

src/inmanta/server/services/environment_metrics_service.py
sanderr, line 196: This will need to be rescheduled when the setting changes. Also, since it's an environment setting, I think it needs to be done for each environment separately. The easiest approach is probably to extract it to a method here, then allow the environment service to manage it. The periodic full compile setting does something similar. When I implemented that I did my best to do it in a generic way so new schedule settings could easily be added.

- `_enable_schedules` [enables all relevant schedules for the environment](https://github.com/inmanta/inmanta-core/blob/6e7e51da8c8683dc0d7bdb47879c9115cf570f2c/src/inmanta/server/services/environmentservice.py#L166). It is called when the environment [is created](https://github.com/inmanta/inmanta-core/blob/6e7e51da8c8683dc0d7bdb47879c9115cf570f2c/src/inmanta/server/services/environmentservice.py#L389), [starts](https://github.com/inmanta/inmanta-core/blob/6e7e51da8c8683dc0d7bdb47879c9115cf570f2c/src/inmanta/server/services/environmentservice.py#L156) or when [a setting is updated](https://github.com/inmanta/inmanta-core/blob/6e7e51da8c8683dc0d7bdb47879c9115cf570f2c/src/inmanta/server/services/environmentservice.py#L202).
- It delegates the actual scheduling to the compiler service, which [keeps track of existing schedules so it can replace them](https://github.com/inmanta/inmanta-core/blob/6e7e51da8c8683dc0d7bdb47879c9115cf570f2c/src/inmanta/server/services/compilerservice.py#L555). I can't say for sure off the top of my head but it's possible interval schedules are replaced automatically.

sanderr (changes requested):

src/inmanta/server/services/environment_metrics_service.py
arnaudsjs: Fine by my. I was in doubt myself.

arnaudsjs: I changed my mind. the reason why this if-statement exists, is to make sure that the test suite can make sure that the server doesn't act on the metrics tables of the database. So from that perspective it's better to put this one under the if-statement as well. I will add a comment.

src/inmanta/server/services/environment_metrics_service.py
arnaudsjs: Good idea. For some reason I didn't think about writing this into a single query.

src/inmanta/server/services/environment_metrics_service.py
sanderr: Ok, I see the reasoning. We'll see if it ever causes trouble. Fine by me to leave it as is for now.

sanderr: Ah but this isn't even an environment setting so it couldn't cause issues anyway.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -206,27 +208,37 @@ async def _cleanup_old_metrics(self) -> None:
...
-                    continue
-                delete_older_than: datetime = datetime.now().astimezone() - timedelta(hours=retention_time_in_hours)
-                env_id_to_delete_older_than_timestamp[env.id] = delete_older_than
-
-            if not env_id_to_delete_older_than_timestamp:
-                return
-            filter_condition = " OR ".join(
-                f"(environment=${(2*i)+1} AND timestamp < ${(2*i)+2})"
-                for i in range(len(env_id_to_delete_older_than_timestamp))
+            query = f"""
```
sanderr, line 211: I don't suppose you have any data on the performance difference between the single query and the previous implementation? Purely out of curiosity. I would be interested to get an idea on just how big of a difference it makes.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -206,27 +208,37 @@ async def _cleanup_old_metrics(self) -> None:
...
+                SELECT id, (CASE WHEN e.settings ? $1 THEN (e.settings->>$1)::integer ELSE $2 END) AS retention_time_in_hours
+                FROM {Environment.table_name()} AS e
+                WHERE e.halted IS FALSE
+            ), env_and_delete_before_timestamp AS (
+                SELECT e.id, ($3::timestamptz - make_interval(hours => e.retention_time_in_hours)) AS delete_before_timestamp
+                FROM env_and_retention_time_in_hours AS e
+                WHERE e.retention_time_in_hours > 0
+            ), delete_gauge AS (
+                DELETE FROM {EnvironmentMetricsGauge.table_name()} AS emg
+                WHERE EXISTS(
```
sanderr, line 222: I would think a `INNER JOIN` would have slightly better performance because the `EXISTS` will continue searching if it hits the right environment but the `AND` condition doesn't match.
That said, I don't think this will be a performance bottleneck, no need to change this on my account. I've just gotten in the habit of trying to analyze all queries I see a practice.

sanderr (changes requested):
New changes look good to me, no remarks. I think you forgot to address [this comment](https://github.com/inmanta/inmanta-core/pull/5395#discussion_r1073695182) though.

src/inmanta/server/services/environment_metrics_service.py
arnaudsjs: I didn't do a comparison. But it would surprise me if this one would be slower, because the previous one did three round trips to the database.

arnaudsjs:
> I think you forgot to address https://github.com/inmanta/inmanta-core/pull/5395#discussion_r1073695182 though.

The cleanup interval is not configurable using an environment setting, so I don't think this comment is relevant. I thought you came to the same conclusion in one of the other threads in this PR. Do you agree?

src/inmanta/server/services/environment_metrics_service.py
sanderr: Me too, I assume this one is faster, I'm just wondering what order of magnitude the difference would be. I might set up an experiment sometime.

sanderr (approved):

PR #5397, opened 2023-01-18
Move DB-related functions and classes to inmanta.db.util module

### LINKED ISSUE #5383: move PGRestore and do_clean_hard from testing to inmanta namespace
inmanta/inmanta-support#475 requires the functionality implemented in `inmanta_tests.conftest.do_clean_hard` and `inmanta_tests.db.common.PGRestore`. They should be moved to the `inmanta` namespace (e.g. `inmanta.db.util`).

- [x] move function and class. Make sure to import under the old names in the old locations for backwards compatibility purposes.
- [x] mark as part of stable API and make sure an import test exists in `tests/test_import_entry_points.py`. Add to programmatic API docs.
- [x] rename `do_clean_hard` to something that is more immediately clear to an end user. Double check other method names are clear and docstrings are present where relevant.
- [x] Add deprecation note stating that this has been moved from `inmanta_tests` to wherever it was moved, and that it will be dropped from the old location in a future version. Create a follow-up ticket for the removal in a future iso version (>=iso7). This might already be the case as a result of iso5 divergence.
- [x] Make sure master/iso6 are at least a minor ahead of their corresponding stable branch so `inmanta-support` can add a meaningful constraint on `inmanta-core`.

arnaudsjs:
# Description

Moved the `do_clean_hard` and `postgres_get_custom_types` functions and the `PGRestore` and the `AsyncSingleton` classes to the `inmanta.db.util` module. This way it can be used from other Inmanta extensions.

closes #5383

# Self Check:

- [x] Changelog entry


stubs/asyncpg/connection.pyi
```diff
@@ -40,6 +41,7 @@ class Connection:
     def terminate(self) -> None: ...
     async def reset(self, *, timeout: Optional[float] = ...) -> None: ...
     async def reload_schema_state(self) -> None: ...
+    async def _copy_in(self, copy_stmt: str, source: AsyncIterable[bytes], timeout: int) -> None: ...
```
sanderr, line 44: I don't like this (the fact that we use it I mean, I do like the fact that we type it if we use it). But I don't see an less brittle way around it either. It does make me wonder if we should include an upper bound on the minor of the `asyncpg` constraint in `setup.py` to make sure released versions don't break on an `asyncpg` update.

sanderr (approved):
I didn't check the implementation in `inmanta.db.util` under the assumption that it remained unchanged. Let me know if that is not the case.

arnaudsjs:
> I didn't check the implementation in `inmanta.db.util` under the assumption that it remained unchanged. Let me know if that is not the case.

I added some code to fix the typing issues. But the logic should remain the name.

stubs/asyncpg/connection.pyi
arnaudsjs: I see your concern, but adding an upper bound on the minor version doesn't provide us any guarantees, right? It only decreases the likelihood of an issue, because we narrow the range.

Hugo-Inmanta (approved):

stubs/asyncpg/connection.pyi
sanderr: You're right, I overstated the impact of that change. We would be betting on them not including big internal refactors in a patch update.

arnaudsjs: I added the upper bound.

sanderr (approved):

PR #5435, opened 2023-01-26
Update incremental deploy wrt to deploying

### LINKED ISSUE #5434: Stuck in repair loop/ infinite loop from increment calculation
# Problem

Give the following scenario

1. we start a periodic repair, which happens to be veeeery slow
2. a resource starts to deploy, it goes to the deploying state
3. a periodic (incremental) deploy starts
4. it calculates the increment, which is empty for that agent, except for the deploying resource
5. the slow periodic deploy is interrupted because there is an increment (with the one deploying resource in there)
6. one slow resource is deployed again
7. the repair is resumed -> goto 1

# Solution

The 'deploying' state should be treated as similar to the 'available' state

- [ ] This should go into ISO4

# Work around 

- increase the deploy interval to be larger than repair time

wouterdb:
# Description

Change the handling of deploying state to prevent infinite loops in repair

1. I went through the commit history and there seems to be no reason why this should not work
2. **question**: should  I make a full scenario test for the bug? I'm afraid it will be slow, brittle and complicated, but I can try. I did leave some comments that should make sure this behavior it never changed inadvertently. 

closes #5434 

# Self Check:

- [x] Changelog entry
- [] Sufficient test cases (reproduces the bug/tests the requested feature)

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5136,15 +5136,14 @@ class ConfigurationModel(BaseDocument):
                 ores = id_to_resource[res["resource_id"]]
 
                 status = ores["status"]
-                # available -> next version
-                if status in [ResourceState.available.name]:
+                # available / deploying -> next version
+                if status in [ResourceState.available.name, ResourceState.deploying.name]:
                     next.append(res)
 
                 # -> increment
                 elif status in [
                     ResourceState.failed.name,
                     ResourceState.cancelled.name,
-                    ResourceState.deploying.name,
                     ResourceState.skipped_for_undefined.name,
                     ResourceState.undefined.name,
                     ResourceState.skipped.name,
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5136,14 +5136,14 @@ class ConfigurationModel(BaseDocument):
                 ores = id_to_resource[res["resource_id"]]
 
                 status = ores["status"]
-                # available : next version
+                # available -> next version
                 if status == ResourceState.available.name:
                     next.append(res)
 
                 # deploying
                 # same hash -> next version
                 # different hash -> increment
-                elif status in [ResourceState.deploying.name]:
+                elif status == ResourceState.deploying.name:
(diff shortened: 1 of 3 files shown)


sanderr (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -5136,15 +5136,23 @@ async def get_increment(cls, environment: uuid.UUID, version: int) -> tuple[set[
                 ores = id_to_resource[res["resource_id"]]
 
                 status = ores["status"]
-                # available -> next version
-                if status in [ResourceState.available.name]:
+                # available : next version
+                if status in [ResourceState.available.name, ResourceState.deploying.name]:
```
sanderr, line 5140: This should be reverted I think.
```suggestion
                if status == ResourceState.available.name:
```

sanderr (changes requested):

arnaudsjs (approved):

src/inmanta/data/__init__.py
```diff
@@ -5136,15 +5136,23 @@ async def get_increment(cls, environment: uuid.UUID, version: int) -> tuple[set[
...
-                # available -> next version
-                if status in [ResourceState.available.name]:
+                # available : next version
+                if status == ResourceState.available.name:
                     next.append(res)
 
+                # deploying
+                # same hash -> next version
+                # different hash -> increment
+                elif status in [ResourceState.deploying.name]:
```
wouterdb, line 5146:
```suggestion
                elif status == ResourceState.deploying.name:
```

src/inmanta/data/__init__.py
```diff
@@ -5136,15 +5136,23 @@ async def get_increment(cls, environment: uuid.UUID, version: int) -> tuple[set[
                 ores = id_to_resource[res["resource_id"]]
 
                 status = ores["status"]
-                # available -> next version
-                if status in [ResourceState.available.name]:
+                # available : next version
```
wouterdb, line 5139:
```suggestion
                # available -> next version
```

sanderr (approved):

Hugo-Inmanta (approved):

PR #5445, opened 2023-01-30
make shutdown more resilient

wouterdb:
# Description

1. Improve testcase stability by making shutdown more robust
2. It fixes a bug where the first exception breaks the waiting  (see example below)

```python
async def test_a():
    a = []
    async def test1():
        raise Exception()

    async def test2():
        await asyncio.sleep(1)
        a.append("A")

    try:
        await asyncio.gather(test1(), test2())
    except Exception:
        a.append("B")

    assert len(a) == 2
```

# Self Check:

- [x] Changelog entry
- [ ]~~ No (preventable) type errors (check using make mypy or make mypy-diff)~~

### DIFF
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -370,7 +370,7 @@ class Scheduler(object):
                 # Ignore this, it is ok to leak a cancel here
                 pass
             if isinstance(result, Exception):
-                logging.error("Exception during shutdown", exc_info=result)
+                LOGGER.error("Exception during shutdown", exc_info=result)
 
     def __del__(self) -> None:
         if len(self._scheduled) > 0:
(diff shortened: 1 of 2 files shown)


arnaudsjs (approved):

src/inmanta/util.py
```diff
@@ -359,7 +359,18 @@ async def stop(self) -> None:
...
+            *[handle for handles in self._await_tasks.values() for handle in handles], return_exceptions=True
+        )
+
+        # Log any exception that happened during shutdown
+        for result in results:
+            if isinstance(result, CancelledError):
+                # Ignore this, it is ok to leak a cancel here
+                pass
+            if isinstance(result, Exception):
+                logging.error("Exception during shutdown", exc_info=result)
```
sanderr, line 373: Does logging suffice?

sanderr (approved):

src/inmanta/util.py
wouterdb: ooooh! thx

PR #5453, opened 2023-01-31
Improve performance of put_partial endpoint

### LINKED ISSUE #4743: improve performance of partial resource application

arnaudsjs:
# Description

Increase the performance of the `put_partial` endpoint by not fetching resources from the database that belong to a resource set that will be removed.

closes #4743 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/4743-increase-performance-put-partial.yml
+++ b/changelogs/unreleased/4743-increase-performance-put-partial.yml
@@ -1,0 +1,8 @@
+---
+description: Increase the performance of the `put_partial` endpoint by not fetching resources from the database that belong to a resource set that will be removed.
+issue-nr: 4743
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master, iso6, iso5]
+sections:
+  minor-improvement: "{{description}}"
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4436,13 +4436,26 @@ class Resource(BaseDocument):
         version: int,
         agent: Optional[str] = None,
         no_obj: bool = False,
+        exclude_resource_sets: Optional[abc.Set[str]] = None,
         *,
         connection: Optional[asyncpg.connection.Connection] = None,
     ) -> List["Resource"]:
+        """
+        Return the resources that are present in the given environment and version.
+
+        :param environment: The environment the resources should belong to.
+        :param version: The version the resources should belong to.
+        :param agent: Is not None, only return resources for this agent.
+        :param exclude_resource_sets: If not None, only return resources that do not belong to these resource sets.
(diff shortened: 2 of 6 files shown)


Hugo-Inmanta (approved):

sanderr (changes requested):
As discussed.

changelogs/unreleased/4743-increase-performance-put-partial.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: Increase the performance of the `put_partial` endpoint by not fetching resources from the database that belong to a resource set that will be removed.
```
wouterdb, line 2: This is not what the ticket intended: the idea it to not pull any resource set in memory, but to copy them in the DB.

arnaudsjs: Yes. I understood that from Sander his review. I am not sure what you mean with the `DB constraints` however.

wouterdb: OK, it wasn't in the comments, so I was  a bit surprised. 

The constraints is not very important.  What I mean is e.g. 

2. every resource has a unique constraint on environment/id/version
3. we copy all resources that have to remain, updating their version field
4. we insert the new resources

Now if a resource is duplicated, we get a constraint violation when we try to insert it a second time. 
So we don't need to check that ourselves. 

(Now I think about it: do we have any rules about cross-resource-set-requires? An update of one set could make a requires of a resource in another set disappear)

sanderr: We discussed it in person because it was easier to get accross that way. That's what the "As discussed" refers to.

arnaudsjs: Yes. I was planning to rely on the database constraints to detect duplicate resources. There is a constraint on the usage of partial compiles that covert your concern regarding `cross-resources-set-requires`:


See https://docs.inmanta.com/inmanta-service-orchestrator/5/model_developers/resource_sets.html#constraints-and-rules:
```
The new version of the model that emerges from a partial compilation should have a dependency graph that is closed within the resource sets that were exported. i.e., it should not depend on any resource sets other than those that were exported.
```

sanderr: Now that I think about it, is this constraint acceptable for lsm's multi-service partial compiles? In all practical cases I think it means that all resources of a slice have to be in the `owned_resources` of a single service instance, not distributed over the services. Something to take into account when writing the docs for the new feature.

arnaudsjs:
I created a new PR for this (See #5580)

PR #5470, opened 2023-02-01
Issue/5179 inmanta workon improvements

### LINKED ISSUE #5179: inmanta-workon improvements (quick wins)
1. print how to deactivate (`deactivate`) and warn the user to do so rather than just to exit the shell to allow the deactivation to check for incorrect permissions.
2. if not run as root or the inmanta user, show a warning to say that the tool should be run as either root or the inmanta user to have write access (to be able to run `pip install` or `inmanta project install`.
3. update the suggested command to fix permissions to just use a plain `chown -R` instead of the current `find ... -exec chown ...`
4. set the `INMANTA_CONFIG_ENVIRONMENT` variable to the id of the activated environment. This should allow compiling lsm-models simply with `inmanta compile`.

Hugo-Inmanta:
# Description

Add minor improvements

closes #5179 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/misc/inmanta-workon-register.sh
+++ b/misc/inmanta-workon-register.sh
@@ -144,6 +144,13 @@ function __inmanta_workon_list {
 
 
 function __inmanta_workon_activate {
+    # Check user
+    current_user = $(whoami)
+    if [ [ $current_user != "root" ] && [ $current_user != ${INMANTA_USER:-inmanta} ] ]; then
+        echo "WARNING: The inmanta-workon tool should be run as either root or the inmanta user to have write access (to be able to run pip install or inmanta project install)." >&2
+        return 1
+    fi
+
     # Activates the environment's venv and registers the deactivate function
 
     declare env_name="$1"
@@ -170,8 +177,11 @@ function __inmanta_workon_activate {
     declare OLD_PS1="$PS1"
     source "$activate"
     export PS1="($env_name) $OLD_PS1"
+    export INMANTA_CONFIG_ENVIRONMENT=$env_id
 
     eval 'inmanta () { python3 -m inmanta.app "$@"; }' # workaround for #4259
+
+    echo "Make sure you exit the current environment by running the 'deactivate' command rather than simply exiting the shell. This ensures the proper permission checks are performed."
     __inmanta_workon_register_deactivate
 }
 
@@ -197,7 +207,7 @@ function __inmanta_workon_register_deactivate {
 
         ownership_issues=$(find "$inmanta_env_dir" \! -user "$user" -print -quit)
         if [ -n "$ownership_issues" ]; then
-            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`find '\''$inmanta_env_dir'\'' ! -user '\''$user'\'' -exec chown '\''$user'\'':'\''$user'\'' {} \;\` as root." >&2
+            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`chown -R\` as root." >&2
         fi
 
         if [ ! "$1" = "nondestructive" ]; then
--- a/changelogs/unreleased/5179-inmanta-workon-improvements.yml
+++ b/changelogs/unreleased/5179-inmanta-workon-improvements.yml
@@ -1,0 +1,6 @@
+description: Minor improvements to the inmanta-workon command.
+change-type: patch
+destination-branches:
+  - master
+  - iso6
+  - iso5


misc/inmanta-workon-register.sh
```diff
@@ -144,6 +144,13 @@ function __inmanta_workon_list {
 
 
 function __inmanta_workon_activate {
+    # Check user
+    current_user = $(whoami)
```
Hugo-Inmanta, line 148: Here I hesitated with `$USER`, not sure if one is better than the other in this context.

misc/inmanta-workon-register.sh
```diff
@@ -144,6 +144,13 @@ function __inmanta_workon_list {
...
+    if [ [ $current_user != "root" ] && [ $current_user != ${INMANTA_USER:-inmanta} ] ]; then
+        echo "WARNING: The inmanta-workon tool should be run as either root or the inmanta user to have write access (to be able to run pip install or inmanta project install)." >&2
```
sanderr, line 150: Since we actually return here without doing anything I would make this an error.
```suggestion
        echo "ERROR: The inmanta-workon tool should be run as either root or the inmanta user to have write access (to be able to run pip install or inmanta project install)." >&2
```

misc/inmanta-workon-register.sh
```diff
@@ -197,7 +207,7 @@ function __inmanta_workon_register_deactivate {
 
         ownership_issues=$(find "$inmanta_env_dir" \! -user "$user" -print -quit)
         if [ -n "$ownership_issues" ]; then
-            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`find '\''$inmanta_env_dir'\'' ! -user '\''$user'\'' -exec chown '\''$user'\'':'\''$user'\'' {} \;\` as root." >&2
+            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`chown -R\` as root." >&2
```
sanderr, line 210:
```suggestion
            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`chown -R '$inmanta_env_dir'\` as root." >&2
```

sanderr: I haven't checked but I think this should expand fine.

misc/inmanta-workon-register.sh
```diff
@@ -170,8 +177,11 @@ function __inmanta_workon_activate {
     declare OLD_PS1="$PS1"
     source "$activate"
     export PS1="($env_name) $OLD_PS1"
+    export INMANTA_CONFIG_ENVIRONMENT=$env_id
```
sanderr, line 180: We should probably unset this one on deactivate. For `PS1` we don't do it because `virtualenv` does it for us.

misc/inmanta-workon-register.sh
sanderr: I think `whoami` is more portable.

misc/inmanta-workon-register.sh
sanderr: That or don't abort in this case. I think aborting is fine though. @wouterdb wdyt?

sanderr (changes requested):

misc/inmanta-workon-register.sh
sanderr: (See the `activate` script in the `bin` dir of any venv you've got lying around.)

misc/inmanta-workon-register.sh
wouterdb: I would not abort and warn

misc/inmanta-workon-register.sh
```diff
@@ -205,9 +204,13 @@ function __inmanta_workon_register_deactivate {
         unset -f inmanta >/dev/null 2>&1
         # no need to restore PS1 because virtualenv_deactivate already does that
 
+        if [ -n "${INMANTA_CONFIG_ENVIRONMENT}" ] ; then
+            unset INMANTA_CONFIG_ENVIRONMENT
```
sanderr, line 208: I actually think we should restore it if it was set before. Something like `PYTHONHOME` in `activate` scripts, but slightly more advanced.

misc/inmanta-workon-register.sh
```diff
@@ -205,9 +204,13 @@ function __inmanta_workon_register_deactivate {
...
+        fi
+
         ownership_issues=$(find "$inmanta_env_dir" \! -user "$user" -print -quit)
         if [ -n "$ownership_issues" ]; then
-            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`chown -R\` as root." >&2
+            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`chown -R $user '$inmanta_env_dir'\` as root." >&2
```
sanderr, line 213: Good catch, I'd forgotten about the user part. Still missing the group though.
```suggestion
            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`chown -R $user:$user '$inmanta_env_dir'\` as root." >&2
```

sanderr (changes requested):

misc/inmanta-workon-register.sh
```diff
@@ -204,13 +211,19 @@ function __inmanta_workon_register_deactivate {
         unset -f inmanta >/dev/null 2>&1
         # no need to restore PS1 because virtualenv_deactivate already does that
 
-        if [ -n "${INMANTA_CONFIG_ENVIRONMENT}" ] ; then
+        if [ -n "${_OLD_INMANTA_CONFIG_ENVIRONMENT:-}" ] ; then
+            # Nested environment case: reset INMANTA_CONFIG_ENVIRONMENT to outer env
```
sanderr, line 215: I don't think nested environments are a thing here because we always call `deactivate` before activating the other (like `virtualenv` does). Keeping track of the old vars would be too complicated otherwise. It's more for the user's own env vars.

misc/inmanta-workon-register.sh
```diff
@@ -204,13 +211,19 @@ function __inmanta_workon_register_deactivate {
...
+            unset _OLD_INMANTA_CONFIG_ENVIRONMENT
+        else
+            # Outermost environment case: simply unset INMANTA_CONFIG_ENVIRONMENT
             unset INMANTA_CONFIG_ENVIRONMENT
         fi
 
         ownership_issues=$(find "$inmanta_env_dir" \! -user "$user" -print -quit)
         if [ -n "$ownership_issues" ]; then
-            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`chown -R $user '$inmanta_env_dir'\` as root." >&2
+            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`chown -R $user:$user '$inmanta_env_dir'\` as root." >&2
```
sanderr, line 226: From the tests I get the impression that the `inmanta_env_dir` variable doesn't get expanded correctly. I think (not sure) this should do it:
```suggestion
            echo "WARNING: Some files in the environment are not owned by the $user user. To fix this, run \`chown -R $user:$user '\''$inmanta_env_dir'\''\` as root." >&2
```
Iirc the naive approach I suggested first closes the `eval` string we open on line 205, evaluates the variable, then opens a new string and concatenates the three together. Only the variable does not exist in the outer scope. So instead we close the string, concatenate a single quote, and open the string again.

tests/server/test_workon.py
```diff
@@ -763,10 +770,100 @@ async def test_workon_deactivate(
...
+        expected_dir=env_dir,
+        invert_python_assert=True,
+        invert_ps1_assert=True,
+        expect_stderr="WARNING: Make sure you exit the current environment by running the 'deactivate' command rather than simply exiting the shell. This ensures the proper permission checks are performed."
+    )
+
+
+
+@pytest.mark.slowtest
+async def test_workon_sets_inmanta_config_environment_empty_outer(
```
sanderr, line 835: For this one, perhaps unset the variable in the pre script, just in case it's set on the host running the tests?

sanderr (changes requested):
One last detail, see `inmanta_env_dir` expansion comment. The others are of lesser importance.

sanderr (approved):

PR #5477, opened 2023-02-02
[FEAT] bump compiler version to 2023.1

Hugo-Inmanta:
# Description

bump compiler version to 2023.1

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/2023-02-02-release-bump-compiler-version.yml
+++ b/changelogs/unreleased/2023-02-02-release-bump-compiler-version.yml
@@ -1,4 +1,4 @@
 ---
 description: Bump COMPILER_VERSION to 2023.1
 change-type: patch
-destination-branches: [master]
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/2023-02-02-release-bump-compiler-version.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: Bump COMPILER_VERSION to 2023.1
+change-type: patch
+destination-branches: [master]
```
sanderr, line 4: For some reason I can't attach a suggestion... We'll need to include iso6 because they are still aligned. Which also means that ideally we want to have OSS and iso6 both in the RC stage before releasing the stable for either. It's not terribly important, but it prevents the creation of second (very small) version for the second of the two.

sanderr (approved):

PR #5483, opened 2023-02-02
added small improvement for release command

wouterdb:
# Description

1. Added some more docs to the release increment calculation
2. I made it log what it does on DEBUG
3. I made sure we never return a lower version (i.e. adding the `.dev0`  tag is moving to a lower version

# Self Check:

- [x] Changelog entry
- [x ] Type annotations are present

### DIFF
--- a/changelogs/unreleased/fix_release_command.yml
+++ b/changelogs/unreleased/fix_release_command.yml
@@ -1,6 +1,6 @@
 description: Ensure the `inmanta module release` command never decreases the version number
 change-type: patch
-destination-branches: [master, iso6, iso5]
+destination-branches: [master, iso6]
 sections:
   minor-improvement:  "{{description}}"
 
--- a/changelogs/unreleased/fix_release_command.yml
+++ b/changelogs/unreleased/fix_release_command.yml
@@ -1,6 +1,5 @@
 description: Ensure the `inmanta module release` command never decreases the version number
 change-type: patch
 destination-branches: [master, iso6]
-sections:
-  minor-improvement:  "{{description}}"
+sections: []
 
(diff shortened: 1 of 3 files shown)


src/inmanta/moduletool.py
```diff
@@ -1107,24 +1112,63 @@ def _get_dev_version_with_minimal_distance_to_previous_stable_release(
...
+                    current_diff,
+                    minimal_version_bump_to_prev_release,
                 )
+                # It is good as it is
+                new_version = current_version
             else:
-                return new_version
-        else:
-            return VersionOperation.set_version_tag(current_version, version_tag="dev0")
+                LOGGER.debug(
```
wouterdb, line 1150: This is different

changelogs/unreleased/fix_release_command.yml
```diff
@@ -0,0 +1,6 @@
+description: Ensure the `inmanta module release` command never decreases the version number
+change-type: patch
+destination-branches: [master, iso6, iso5]
```
sanderr, line 3: I don't think the command exists on iso5
```suggestion
destination-branches: [master, iso6]
```

sanderr (approved):
Nice!

changelogs/unreleased/fix_release_command.yml
```diff
@@ -0,0 +1,6 @@
+description: Ensure the `inmanta module release` command never decreases the version number
```
sanderr, line 1: This sounds more ominous than it is, which might not be ideal if we use it as a changelog section?

changelogs/unreleased/fix_release_command.yml
```diff
@@ -0,0 +1,6 @@
...
+change-type: patch
+destination-branches: [master, iso6]
+sections:
+  minor-improvement:  "{{description}}"
```
wouterdb, line 5:
```suggestion
sections: []
```

changelogs/unreleased/fix_release_command.yml
wouterdb: changed

arnaudsjs (approved):

PR #5498, opened 2023-02-06
Changed default value

wouterdb:
# Description

In a recent support call, I have noticed that de dashboard promote button by default performs a full deploy.
This is not compatible with the upgrade procedures described in the docs. 

1. I updated this default. 
2. I tried to improve the docs

It is strictly speaking a breaking change, but it should not affect many things in practice. 

# Self Check:

- [x] Changelog entry
- [X] No (preventable) type errors (check using make mypy or make mypy-diff)
- [X] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )


arnaudsjs (approved):

FloLey (approved):

changelogs/unreleased/agent_trigger_method_default.yml
```diff
@@ -0,0 +1,6 @@
+description: Changed the default value of `environment_agent_trigger_method` environment setting to `push_incremental_deploy`
+change-type: minor
+destination-branches: [master, iso6, iso5]
```
wouterdb, line 3:
```suggestion
destination-branches: [master, iso6]
```

PR #5548, opened 2023-02-13
Fix for line numbering issue

### LINKED ISSUE #5497: Errors are reported on the wrong line
Constructors are often written in multiple lines. Errors are always reported on the line where the constructor starts. This also results in wrong marks in vscode

image

wouterdb:
# Description

Fixes #5497 

A few remarks:
1. the error is now reported on the key in the assignment (in the testcase that is `d` in `d = "D"` 
2. @sanderr I added a field to not have to re-wire the whole class, wdyt? 
3. @bartv @sanderr I could also report on the RHS of the assignment, which is easier to implement, but I believe less accurate. What should I report on ? 
4. @sanderr do you recall why locatable str is not in fact a child of locatable? 

# Self Check:

- [ ] Changelog entry
```
+ src/inmanta/ast/statements/generator.py:565: error: Argument 1 to "TypingException" has incompatible type "LocatableString"; expected "Optional[Locatable]"  [arg-type]
```

- [X] Sufficient test cases (reproduces the bug/tests the requested feature)

### DIFF
--- a/changelogs/unreleased/5497-errors-on-attributes.yml
+++ b/changelogs/unreleased/5497-errors-on-attributes.yml
@@ -1,0 +1,6 @@
+description: Improve line numbering when reporting non-existing attributes on constructors
+issue-nr: 5497
+change-type: patch
+destination-branches: [master, iso6]
+sections:
+  minor-improvement: "{{description}}"
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -471,6 +471,7 @@ class Constructor(ExpressionStatement):
     __slots__ = (
         "class_type",
         "__attributes",
+        "__attribute_locations",
         "__wrapped_kwarg_attributes",
         "location",
         "type",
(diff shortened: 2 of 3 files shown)


sanderr (approved):
2. Fine by me, the overhead should be fairly limited since these are static objects.
3. I prefer reporting on the lhs, even if it is slightly more complex.
4. Alas not, I do seem to recall @Hugo-Inmanta trying to fix that at some point and running into issues. But I could be misremembering.

sanderr:
4. Perhaps it has to do with the fact that it works with `Range` only?

Hugo-Inmanta:
> 2. Fine by me, the overhead should be fairly limited since these are static objects.
> 
>     3. I prefer reporting on the lhs, even if it is slightly more complex.
> 
>     4. Alas not, I do seem to recall @Hugo-Inmanta trying to fix that at some point and running into issues. But I could be misremembering.

This doesn't ring that big of a bell, the closest thing I can think of is this: https://github.com/inmanta/inmanta-core/pull/5548#pullrequestreview-1295946714 ?

bartv (approved):
Looks alright as far as I can tell

PR #5549, opened 2023-02-13
Improve logging on exceptions in api requests

bartv:
# Description

When we get an http exception during an API request we send the message to the client without logging.

When this error causes a redirect loop in your browser you do not even see the message. This PR also adds it to the logs of the server.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/protocol/rest/server.py
+++ b/src/inmanta/protocol/rest/server.py
@@ -158,7 +158,7 @@ class RESTHandler(tornado.web.RequestHandler):
                 self.respond({"message": "Unable to decode request body"}, {}, 400)
 
             except exceptions.BaseHttpException as e:
-                LOGGER.warning("Got %d status code: %s", e.to_status(), e.to_body())
+                LOGGER.warning("Received an exception with status code %d and message %s", e.to_status(), e.to_body())
                 self.respond(e.to_body(), {}, e.to_status())
 
             except CancelledError:
--- a/src/inmanta/protocol/rest/server.py
+++ b/src/inmanta/protocol/rest/server.py
@@ -158,6 +158,7 @@ class RESTHandler(tornado.web.RequestHandler):
                 self.respond({"message": "Unable to decode request body"}, {}, 400)
 
             except exceptions.BaseHttpException as e:
+                LOGGER.warning("Got %d status code: %s", e.to_status(), e.to_body())
                 self.respond(e.to_body(), {}, e.to_status())
 
             except CancelledError:
--- a/changelogs/unreleased/improve-request-error-logging.yml
+++ b/changelogs/unreleased/improve-request-error-logging.yml
@@ -1,0 +1,4 @@
+---
+description: Improve the logging of exceptions in API requests
+change-type: patch
+destination-branches: [master, iso6, iso5]


sanderr (approved):

src/inmanta/protocol/rest/server.py
```diff
@@ -158,6 +158,7 @@ async def _call(self, kwargs: Dict[str, str], http_method: str, call_config: com
                 self.respond({"message": "Unable to decode request body"}, {}, 400)
 
             except exceptions.BaseHttpException as e:
+                LOGGER.warning("Got %d status code: %s", e.to_status(), e.to_body())
```
wouterdb, line 161: This will log a warning for every non-200 response we return via our API (roughly? ) ? 

As I understand, this is on the dispatcher path, the message suggests we `Got` an exception, but it is more likely that we produce the exception. 

Also, in `_execute_call` there already is : 
```
        except exceptions.BaseHttpException:
            LOGGER.debug("An HTTP Error occurred", exc_info=True)
            raise
 ```

So we will log it twice at different level.

**Review by wouterdb — CHANGES_REQUESTED** [review]

File: src/inmanta/protocol/rest/server.py (hunk 1 above)
Comment (bartv) [reply src/inmanta/protocol/rest/server.py:161]: I did not get errors in my logs. Everything that generates exceptions outside _execute_call will only get a warning log from the tornado with the status code and the url.

Comment (bartv) [reply src/inmanta/protocol/rest/server.py:161]: I am open for any other solution that ensures it is possible to configure authentication based on the logs (and not having to use a debugger)

Comment (wouterdb) [reply src/inmanta/protocol/rest/server.py:161]: What exactly happened?

Comment (bartv) [reply src/inmanta/protocol/rest/server.py:161]: The aud setting was wrong, for which you get the message https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/protocol/common.py#L1032

But the logs will only say: "403 /.... "

Comment (wouterdb) [reply src/inmanta/protocol/rest/server.py:161]: I checked the docs https://internal.inmanta.com/development/core/technical-designs/workmanship.html?highlight=workman


Level | Meaning
-- | --
Trace | Information only relevant to developers of specific subsystems.
Debug | Detailed information, typically of interest only when diagnosing problems, intended audience is developers and support staff.
Info | Information on normal progress, should be understandable to operators and end users.
Warning | Information on potential problems, targeted towards operators and end users.
Error | Information on problems, that should be examined by an operator, but not immediately.
Critical | Information on problems that are a threat to system stability. An  operator should be notified immediately (even in the middle of the  night).

So this can be considered to be in line with that. 

I don't know if logging every 404 as a warning makes sense, but it does improve visibility

**Review by wouterdb — APPROVED** [review]

File: src/inmanta/protocol/rest/server.py (hunk 1 above)
Comment (bartv) [reply src/inmanta/protocol/rest/server.py:161]: That is the thing, tornado logs every 404 but without the message that we send to the client. That makes debugging really difficult.

Alternatively we disable the tornado logger and replace it with this one. Then there is the same amount of log line but with more information

**Comment by bartv** [comment]
after:

image

**Comment by bartv** [comment]
I am going to change the message to something better:

```
Received an exception with status code 403 and message {'message': 'Access denied: Invalid audience'}
```

PR #5561, opened 2023-02-15
[FEAT] move increment calculation to release_version pipeline

### LINKED ISSUE #5508: Calculate increment at release time
Step 1 of #5491, see that ticket for more details.

Should be done all in one transaction.

Hugo-Inmanta:
# Description

move increment calculation to release_version pipeline

closes #5508 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
+++ b/changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
@@ -1,6 +1,7 @@
-description: Move increment calculation to release_version pipeline.
-change-type: patch
+description: Move increment calculation to release_version pipeline to prevent resource state change outside of the agent's deploy lock.
+change-type: minor
 destination-branches:
   - master
   - iso6
-  - iso5
+sections:
+  bugfix: "{{description}}"
--- a/changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
+++ b/changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
@@ -1,0 +1,6 @@
+description: Move increment calculation to release_version pipeline.
+change-type: patch
+destination-branches:
+  - master
+  - iso6
+  - iso5
--- a/changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
+++ b/changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
@@ -4,4 +4,4 @@ destination-branches:
   - master
   - iso6
 sections:
-  bugfix: "{{description}}"
+  bugfix: "Fix race condition that causes the deployment of a resource to fail with the error message: 'Fetching resource events only makes sense when the resource is currently deploying'. This issue happens in rare situations and the orchestrator will automatically recover from it during the next deployment."
--- a/src/inmanta/server/services/orchestrationservice.py
+++ b/src/inmanta/server/services/orchestrationservice.py
@@ -26,7 +26,7 @@ import asyncpg
 import asyncpg.connection
 import pydantic
 
-from inmanta import const, data, util
+from inmanta import const, data
 from inmanta.const import ResourceState
 from inmanta.data import APILIMIT, AVAILABLE_VERSIONS_TO_KEEP, ENVIRONMENT_AGENT_TRIGGER_METHOD, PURGE_ON_DELETE, InvalidSort
 from inmanta.data.dataview import DesiredStateVersionView
@@ -889,7 +889,6 @@ class OrchestrationService(protocol.ServerSlice):
         if model is None:
             return 404, {"message": "The request version does not exist."}
 
-
         if model.total == 0:
             await model.mark_done(connection=connection)
             return 200, {"model": model}
@@ -935,13 +934,14 @@ class OrchestrationService(protocol.ServerSlice):
                     connection=connection,
                 )
 
-        increments: tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]] = await self.resource_service._get_increment(env, version_id)
+        increments: tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]] = await self.resource_service._get_increment(
+            env, version_id
+        )
 
         increment_ids, neg_increment = increments
         await self.resource_service._mark_deployed(env, neg_increment, now, lambda x: True, version_id)
 
         await model.update_fields(released=True, result=const.VersionState.deploying, connection=connection)
-
 
         if push:
             # fetch all resource in this cm and create a list of distinct agents
--- a/src/inmanta/server/services/orchestrationservice.py
+++ b/src/inmanta/server/services/orchestrationservice.py
@@ -930,12 +930,12 @@ class OrchestrationService(protocol.ServerSlice):
                     connection=connection,
                 )
 
-        increments: tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]] = await self.resource_service._get_increment(
+        increments: tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]] = await self.resource_service.get_increment(
             env, version_id
         )
 
         increment_ids, neg_increment = increments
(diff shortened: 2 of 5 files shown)


src/inmanta/server/services/resourceservice.py
```diff
@@ -390,6 +358,53 @@ def in_requires(req: ResourceIdStr) -> bool:
 
         return 200, {"environment": env.id, "agent": agent, "version": version, "resources": deploy_model}
 
+    async def _mark_deployed(
+        self,
+        env: data.Environment,
+        neg_increment: abc.Set[ResourceIdStr],
+        now: datetime.datetime,
+        on_agent: Callable[[ResourceIdStr], bool],
```
wouterdb, line 366: I would rename this to filter to indicate what it does

src/inmanta/server/services/resourceservice.py
```diff
@@ -390,6 +358,53 @@ def in_requires(req: ResourceIdStr) -> bool:
...
+            messages=[logline],
+            change=const.Change.nochange,
+            send_events=False,
+            keep_increment_cache=True,
+        )
+
+    async def _get_increment(
+        self, env: data.Environment, version: int
+    ) -> tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]]:
+        increment: Optional[tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]]] = self._increment_cache.get(env.id, None)
```
wouterdb, line 397: This could use some docs

src/inmanta/server/services/resourceservice.py
```diff
@@ -390,6 +358,53 @@ def in_requires(req: ResourceIdStr) -> bool:
...
         return 200, {"environment": env.id, "agent": agent, "version": version, "resources": deploy_model}
 
+    async def _mark_deployed(
+        self,
+        env: data.Environment,
+        neg_increment: abc.Set[ResourceIdStr],
+        now: datetime.datetime,
+        on_agent: Callable[[ResourceIdStr], bool],
+        version: int,
+    ) -> None:
```
wouterdb, line 368: this could use a short docstring

wouterdb (changes requested):

changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
```diff
@@ -0,0 +1,6 @@
+description: Move increment calculation to release_version pipeline.
```
arnaudsjs, line 1: Don't we want to have a changelog entry for this?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -936,6 +930,19 @@ async def release_version(
...
 
+        increments: tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]] = await self.resource_service._get_increment(
+            env, version_id
+        )
+
+        increment_ids, neg_increment = increments
+        await self.resource_service._mark_deployed(env, neg_increment, now, lambda x: True, version_id)
+
+        await model.update_fields(released=True, result=const.VersionState.deploying, connection=connection)
+
```
arnaudsjs, line 941: Why did we move this call down? I find it a bit strange that we send resource_action_update calls with the action  `ResourceAction.deploy` when the state of the ConfigurationModel hasn't been set to deploying yet.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -936,6 +930,19 @@ async def release_version(
                     connection=connection,
                 )
 
+        increments: tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]] = await self.resource_service._get_increment(
+            env, version_id
+        )
+
+        increment_ids, neg_increment = increments
+        await self.resource_service._mark_deployed(env, neg_increment, now, lambda x: True, version_id)
```
arnaudsjs, line 938: I think we should remove the underscores from these method names, because they belong to the public API of the class while the name suggests otherwise.

src/inmanta/server/services/resourceservice.py
```diff
@@ -390,6 +358,53 @@ def in_requires(req: ResourceIdStr) -> bool:
...
+        increment: Optional[tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]]] = self._increment_cache.get(env.id, None)
+        if increment is None:
+            lock = self._increment_cache_locks[env.id]
+            async with lock:
+                increment = self._increment_cache.get(env.id, None)
+                if increment is None:
+                    increment = await data.ConfigurationModel.get_increment(env.id, version)
+                    self._increment_cache[env.id] = increment
+        increment_ids, neg_increment = increment
+        return increment_ids, neg_increment
```
arnaudsjs, line 406:
```suggestion
        return increment
```

src/inmanta/server/services/resourceservice.py
arnaudsjs, line 366: I would also make this optional. It feel rather strange to have to pass in a lambda function that always returns True when you don't want to apply a filter.

arnaudsjs (changes requested):

changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
wouterdb: Probably a minor at best

src/inmanta/server/services/orchestrationservice.py
wouterdb: This is the moment agents can start to pull in the resources (by setting `released=True`) so either all of this goes into a transaction or this one has to be at the end,...

src/inmanta/server/services/resourceservice.py
wouterdb: I would not make it optional then, but default to the lambda that returns true

changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
arnaudsjs: Then we should exclude iso5, but I don't think that's an issue.

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: I see. I don't think we need a transaction here. But I would definitely add this explanation in the code as a comment.

src/inmanta/server/services/resourceservice.py
arnaudsjs: > I would not make it optional then, but default to the lambda that returns true

That's exactly what I meant. I was referring to the fact that passing the attribute should be optional and you to whether or not the type should be optional :p

wouterdb (approved):

changelogs/unreleased/5508-move-increment-calculation-to-release-version-pipeline.yml
```diff
@@ -0,0 +1,7 @@
+description: Move increment calculation to release_version pipeline to prevent resource state change outside of the agent's deploy lock.
```
arnaudsjs, line 1: I don't think this changelog message is very useful for an end-user. It only makes sense if you know about how the orchestrator works internally. The end-user is more interested in knowing which issue exists, when it can occur and how big the impact is. From that perspective it would be better to write something like this: `Fix race condition that causes the deployment of a resource to fail with the error message: "Fetching resource events only makes sense when the resource is currently deploying". This issue happens in rare situation and the orchestrator will automatically recover from it during the next deployment.`

However, I do like the fact that the changelog contains a short description regarding the technical implementation that was done to resolve the problem. The good news is that our changelog files supports this. We could do the following:

```
description: Move increment calculation to release_version pipeline to prevent resource state change outside of the agent's deploy lock.
change-type: minor
destination-branches:
  - master
  - iso6
sections:
  bugfix: "Fix race condition that causes the deployment of a resource to fail with the error message: "Fetching resource events only makes sense when the resource is currently deploying". This issue happens in rare situation and the orchestrator will automatically recover from it during the next deployment."
```
The `bugfix` section will go into the changelog and the description field is there for internal use.

arnaudsjs (approved):

PR #5569, opened 2023-02-16
Issue/build master investigate flaky test

Hugo-Inmanta:
# Description

Add a retry_limited around the flaky part of the test_create_too_many_versions test.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -209,9 +209,12 @@ async def test_create_too_many_versions(client, server, n_versions_to_keep, n_ve
     assert result.code == 200
 
     # Check value was set
-    result = await client.get_setting(tid=env_1_id, id=data.AVAILABLE_VERSIONS_TO_KEEP)
-    assert result.code == 200
-    assert result.result["value"] == n_versions_to_keep
+    async def wait_for_setting() -> bool:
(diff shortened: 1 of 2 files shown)


tests/test_server.py
```diff
@@ -209,9 +209,11 @@ async def test_create_too_many_versions(client, server, n_versions_to_keep, n_ve
...
 
     # Check value was set
-    result = await client.get_setting(tid=env_1_id, id=data.AVAILABLE_VERSIONS_TO_KEEP)
-    assert result.code == 200
-    assert result.result["value"] == n_versions_to_keep
+    async def wait_for_setting() -> bool:
+        result = await client.get_setting(tid=env_1_id, id=data.AVAILABLE_VERSIONS_TO_KEEP)
+        return result.code == 200 and result.result["value"] == n_versions_to_keep
+
+    await retry_limited(wait_for_setting, 10)
```
arnaudsjs, line 216: I don't get this. `client.set_setting()` makes the changes in the database in a synchronous way. There shouldn't be a need for a retry_limited here, right?

arnaudsjs (changes requested):

tests/test_server.py
Hugo-Inmanta: Ah indeed, this mustn't be the issue.

PR #5578, opened 2023-02-16
Added first POC of type hinting

### LINKED ISSUE #4028: Infer namespaces in constructor calls
In tree structured code, there are often very long namespaces in constructors

```ruby
        resolution_filter = sros::configure::service::epipe::bgp_evpn::mpls::auto_bind_tunnel::ResolutionFilter(
```

We could however simply infer the namespace based on the relation.

There are some questions however:
 1. how/if to handle subtypes

wouterdb:
# Description

First very raw POC of type hinting. 

Turns out the most difficult wiring is done by the constructor trees. 

This cuts off all corners, but it does show we should be able to do it easily 

closes  #4028 

- [x]  docs
- [x]  changelog file

# TBD

 - ~~this makes it possible to break a model by importing a subtype with the same name as a super type in a different namespace. This is a very dangerous thing.~~  -> resolved: local resolution takes precedence 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -401,6 +401,56 @@ attribute names as keys and the desired values as values. For example:
     file1_config = {"path": "/opt/1"}
     f1 = File(host=h1, **file1_config)
 
+
+Refering to instances
+++++++++++++++++++++++
+
+When referring to entities in the same module, a parent model or std, short names can be used
+
+Following code blocks are equivalent and both valid
+
+.. code-block:: inmanta
+
+    std::Host("test")
+
+.. code-block:: inmanta
+
+    Host("test")
+
+
+When constructing entities from other modules, the fully qualified name must be used
+
+.. code-block:: inmanta
+
+   import srlinux
+   import srlinux::interface
+
+   interface = srlinux::Interface(
+        subinterface = srlinux::interface::Subinterface(
+        )
+    )
+
+When nesting constructors, short names can be used for the nested constructors, because their types can be inferred
+
+.. code-block:: inmanta
+
+   import srlinux
+   import srlinux::interface
+
+   interface = srlinux::Interface( # This type is qualified
+        subinterface = Subinterface( # This type is inferred
+        )
+    )
+
+However, when relying on type inference,
+1. avoid creating sibling types with the same name, but different fully qualified name, as they may become indistinguishable, breaking the inference on existing models.
+
+    1. if multiple types exist with the same name, and one is in scope, that one is selected (i.e it is defined in this module, a parent module or `std`)
+    2. if multiple types exist with that are all out of scope, inference fails
+
+2. make sure the type you want to infer is imported somewhere in the model. Otherwise the compiler will not find it.
+
+
 Refinements
 ===========
 
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -527,14 +527,14 @@ class Constructor(ExpressionStatement):
     def _normalize_rhs(self, index_attributes: abc.Set[str]) -> None:
         assert self.type is not None  # Make mypy happy
         for k, v in self.__attributes.items():
-            # don't notify the rhs for index attributes because it won't be able to resolve the reference
-            # (index attributes need to be resolved before the instance can be constructed)
             attr = self.type.get_attribute(k)
             if attr is None:
                 raise TypingException(
                     self.__attribute_locations[k], "no attribute %s on type %s" % (k, self.type.get_full_name())
                 )
             type_hint = attr.get_type().get_base_type()
+            # don't notify the rhs for index attributes because it won't be able to resolve the reference
+            # (index attributes need to be resolved before the instance can be constructed)
             v.normalize(
                 lhs_attribute=AttributeAssignmentLHS(self._self_ref, k, type_hint) if k not in index_attributes else None
             )
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -24,6 +24,7 @@ from collections import abc
 from itertools import chain
 from typing import Dict, Iterator, List, Optional, Set, Tuple
 
+import inmanta.ast.entity
 import inmanta.ast.type as inmanta_type
 import inmanta.execute.dataflow as dataflow
 from inmanta.ast import (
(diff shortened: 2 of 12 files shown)


bartv:
I did a quick test on a yang tree and it does not seem to work.

sanderr:
@bartv did you leave out the index attributes? This implementation does not work for those.

wouterdb:
note from @bartv 

we should consider the normal resolution as a type candidate as well, so it takes the candidate with the right type  for more intuitive. Now we fall back to inference only when there is no normal match, but this leads to complex behavior that is hard to explain .

src/inmanta/ast/statements/generator.py
```diff
@@ -525,12 +526,41 @@ def _normalize_rhs(self, index_attributes: abc.Set[str]) -> None:
         for k, v in self.__attributes.items():
             # don't notify the rhs for index attributes because it won't be able to resolve the reference
             # (index attributes need to be resolved before the instance can be constructed)
-            v.normalize(lhs_attribute=AttributeAssignmentLHS(self._self_ref, k) if k not in index_attributes else None)
+            type_hint = self.type.get_attribute(k).get_type().get_base_type()
+            v.normalize(
+                lhs_attribute=AttributeAssignmentLHS(self._self_ref, k, type_hint) if k not in index_attributes else None
```
sanderr, line 531: This could even be trivially extended to attribute assignment (I think).

sanderr (commented):
Looks generally good to me

src/inmanta/ast/statements/generator.py
sanderr: Ah no, we don't know the type of the lhs yet.

src/inmanta/ast/statements/generator.py
```diff
@@ -522,15 +525,86 @@ def pretty_print(self) -> str:
...
 
     def _normalize_rhs(self, index_attributes: abc.Set[str]) -> None:
+        assert self.type is not None  # Make mypy happy
         for k, v in self.__attributes.items():
             # don't notify the rhs for index attributes because it won't be able to resolve the reference
             # (index attributes need to be resolved before the instance can be constructed)
-            v.normalize(lhs_attribute=AttributeAssignmentLHS(self._self_ref, k) if k not in index_attributes else None)
+            attr = self.type.get_attribute(k)
+            if attr is None:
+                raise TypingException(
```
wouterdb, line 534: This is a duplicate, ...

Perhaps some re-ordering can help

docs/language.rst
```diff
@@ -401,6 +401,56 @@ attribute names as keys and the desired values as values. For example:
     file1_config = {"path": "/opt/1"}
     f1 = File(host=h1, **file1_config)
 
+
+Refering to instances
```
sanderr, line 405:
```suggestion
Referring to instances
```

src/inmanta/ast/statements/generator.py
```diff
@@ -523,45 +525,24 @@ def pretty_print(self) -> str:
         )
 
     def _normalize_rhs(self, index_attributes: abc.Set[str]) -> None:
+        assert self.type is not None  # Make mypy happy
         for k, v in self.__attributes.items():
             # don't notify the rhs for index attributes because it won't be able to resolve the reference
             # (index attributes need to be resolved before the instance can be constructed)
```
sanderr, line 531: Could you move this comment down to right above the `v.normalize` call?

src/inmanta/ast/statements/generator.py
```diff
@@ -574,6 +585,72 @@ def normalize(self, *, lhs_attribute: Optional[AttributeAssignmentLHS] = None) -
...
+                base_types = [type_hint]
+                # Find all correct types with the matching unqualified name
+                candidates = {
+                    entity
+                    for entity in chain(base_types, type_hint.get_all_child_entities())
+                    if entity.name == str(self.class_type)
+                }
+                if len(candidates) > 1:
+                    # To many options, inheritance may cause this to break a working model due to dependency update
+                    raise AmbiguousTypeException(self.class_type, list(candidates))
```
sanderr, line 629:
```suggestion
                    raise AmbiguousTypeException(self.class_type, sorted(candidates))
```
For more consistent error messages?

src/inmanta/ast/type.py
```diff
@@ -93,12 +93,12 @@ def validate(self, value: Optional[object]) -> bool:
         """
         return True
 
-    def type_string(self) -> Optional[str]:
+    def type_string(self) -> str:
         """
         Returns the type string as expressed in the Inmanta :term:`DSL`, if this type can be expressed in the :term:`DSL`.
         Otherwise returns None.
```
sanderr, line 99:
```suggestion
```

src/inmanta/ast/type.py
```diff
@@ -387,6 +390,9 @@ def validate(self, value: Optional[object]) -> bool:
     def type_string_internal(self) -> str:
         return "List"
 
+    def type_string(self) -> str:
+        return "list"
```
sanderr, line 394: I'm somewhat concerned about this change. I have to look into it.

sanderr (commented):
Index attributes are still not supported. Just wanted to make sure this was a conscious decision.

src/inmanta/ast/type.py
wouterdb: That would be nice, I didn't know how this was intended

src/inmanta/ast/statements/generator.py
wouterdb: done

arnaudsjs (approved):

src/inmanta/ast/statements/generator.py
wouterdb: Doesn't work (entities are not ordereable)

But the sort was already in the sub class.

src/inmanta/ast/type.py
sanderr: So the issue is that this method is meant to represent a mapping between a DSL type and its string representation. But not all children of `Type` are DSL types, e.g. `Union` is an internal type used to represent primitives, `List` is an internal type, with `TypedList` and `LiteralList` its concrete DSL types. If I had to rewrite the whole typing hierarchy I'd probably do it differently, but that's beside the point now. I think we should keep this method as is for now. For any user-definable attribute type, the method should return a not-`None` value, so you could go with an assertion.

sanderr (changes requested):
Requesting changes just to ping you about my response.

src/inmanta/ast/type.py
wouterdb: so you propose to not override this here, as it is a ABC, and only override it in the concrete leaf types?

PR #5580, opened 2023-02-17
Improve performance put_partial endpoint

### LINKED ISSUE #4743: improve performance of partial resource application

arnaudsjs:
# Description

Improve performance of the `put_partial()` endpoint by copying of all resources in the base version of the model to the new version of the model in the database instead of on the server.

closes #4743

# Self Check:

- [x] Changelog entry


FloLey (approved):

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -611,72 +576,85 @@ def safe_get(input: object, key: str, default: object) -> object:
...
+                )
 
-        for uk in unknowns:
-            await uk.insert(connection=connection)
+            all_resource_version_ids |= {Id.set_version_in_id(rid, version) for rid in rids_unchanged_resource_sets}
 
-        for agent in agents:
-            await self.agentmanager_service.ensure_agent_registered(env, agent, connection=connection)
+        await data.Resource.insert_many(list(rid_to_resource.values()), connection=connection)
+        await cm.recalculate_total(connection=connection)
```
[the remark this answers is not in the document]
arnaudsjs: No. When a partial compile is done, `rid_to_resource` contains the resources that are part of the partial compile and the remaining number of resources is copied using the `copy_resources_from_unchanged_resource_set` method. When the configurationmodel record is created, we don't know how many records will be created by the `copy_resources_from_unchanged_resource_set`. That's why the recalculate_total method is called.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,204 @@
...
+            rids_deleted_resource_sets,
+        )
 
-    def resource_changed_resource_set(self) -> bool:
+    def merge_updated_and_shared_resources(
+        self, updated_and_shared_resources: abc.Sequence[data.Resource]
+    ) -> Dict[ResourceIdStr, data.Resource]:
         """
-        return true if the resource_set in new_resource and old_resource are not the same
+        :param updated_and_shared_resources: The resources that are part of the partial compile.
```
sanderr, line 160:
```suggestion
        Separates named resource sets from the shared resource set and expands the shared set with the shared resources in the previous model version.

        :param updated_and_shared_resources: The resources that are part of the partial compile.
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,204 @@
...
+        return result
 
-    resources: list[dict[str, object]]
-    resource_states: dict[ResourceIdStr, Literal[ResourceState.available, ResourceState.undefined]]
-    resource_sets: dict[ResourceIdStr, Optional[str]]
-    unknowns: List[data.UnknownParameter]
+    def _validate_constraints(self, new_updated_and_shared_resources: abc.Sequence[data.Resource]) -> None:
+        """
+        Validate whether the new updated and shared resources that results from the merging the old version of the model
+        with resources of the partial compile, are compliant with the constraints of a partial compile.
```
sanderr, line 174: Since we no longer manipulate the named resource sets here, does this still serve a use? Wouldn't it suffice to check that `updated_resources` and `shared_resources_merged` don't have any keys in common in `merge_updated_and_shared_resources`?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,204 @@
...
+            if res.resource_id not in self.updated_and_shared_resources_old:
+                continue
+            matching_resource_old_model = self.updated_and_shared_resources_old[res.resource_id]
 
-class PartialUpdateMerger:
-    """
-    This class is used to merge the result of a partial compile with previous resources and resource_sets. Takes a partial spec
-    that should be applied on a given base version to form the given new version.
+            if res.resource_set != matching_resource_old_model.resource_set:
+                raise BadRequest(f"A partial compile cannot migrate resource {res.resource_id} to another resource set.")
```
sanderr, line 184: Since we no longer manipulate the named resource sets here, does this still serve a use? Wouldn't it suffice to check that `updated_resources` and `shared_resources_merged` don't have any keys in common in `merge_updated_and_shared_resources`?

This would also mean that for the next check we only need to consider the shared resource check, so it suffices to fetch that one in the `create` call, another small performance gain.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -468,117 +395,155 @@ async def delete_version(self, env: data.Environment, version_id: int) -> Apiret
...
+        removed_resource_sets: Optional[List[str]] = None,
+        *,
+        connection: asyncpg.connection.Connection,
+    ) -> None:
+        is_partial_update = partial_base_version is not None
+
+        if resource_sets is None:
+            resource_sets = {}
+
+        if removed_resource_sets is None:
```
sanderr, line 522: If this is set, it should be disjunct from `resource_sets`. I don't think there's a check for that anywhere.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,204 @@
...
-        :param env: The environment in which the partial compile happens.
-        :param base_version: The version number of the old configuration model from which the resources are merged
-                             together with the resources from the partial compile.
-        :param new_version: The version number of the newly generated configuration model.
-        :param partial_updates: The resources part of the partial compile
-        :param resource_states: The resource states of the resources in the partial compile.
-        :param resource_sets: The resource sets of the resources in the partial compile.
-        :param removed_resource_sets: The resources in these resource sets should be present in the new configuration model.
-        :param unknowns: The unknowns that belong to the partial model.
+    def _merge_shared_resources(self, shared_resources_new: Dict[ResourceIdStr, data.Resource]) -> abc.Sequence[data.Resource]:
```
sanderr, line 189: note to self: continue review between here and `_put_version`

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -836,40 +828,49 @@ async def put_partial(
...
+                    base_version=base_version,
+                    version=version,
+                    rids_in_partial_compile=set(rid_to_resource.keys()),
+                    updated_resource_sets=updated_resource_sets,
+                    deleted_resource_sets=set(removed_resource_sets),
+                    connection=con,
                 )
-                merged_model: MergedModel = await merger.apply_partial(connection=con)
+
+                merged_resources = partial_update_merger.merge_updated_and_shared_resources(list(rid_to_resource.values()))
```
sanderr, line 857:
```suggestion
                # add shared resources
                merged_resources = partial_update_merger.merge_updated_and_shared_resources(list(rid_to_resource.values()))
```

sanderr (commented):
Not fully through yet, I'd thought I'd release the first batch already. Please ignore all comments starting with "note to self", I'll get back to those and delete/resolve them when I'm done with them.

src/inmanta/data/__init__.py
```diff
@@ -4728,6 +4807,95 @@ async def get_resource_deploy_summary(cls, environment: uuid.UUID) -> m.Resource
...
+                    r.resource_type,
+                    r.resource_id_value,
+                    r.agent,
+                    (
+                        CASE WHEN r.status='undefined'::resourcestate
+                        THEN 'undefined'::resourcestate
+                        ELSE 'available'::resourcestate
+                        END
+                    ) AS status,
+                    jsonb_set(r.attributes, '{{"version"}}'::text[], to_jsonb($3::integer)) AS attributes,
```
wouterdb, line 4852: this bit I don't understand?

src/inmanta/data/__init__.py
```diff
@@ -4827,6 +5003,131 @@ def done(self) -> int:
...
+                total=total,
+                version_info=version_info,
+                undeployable=undeployable,
+                skipped_for_undeployable=skipped_for_undeployable,
+            )
+            await cm.insert(connection=connection)
+        else:
+            # A partial compile was done. Create the new version of the ConfigurationModel based on the partial_base version.
+            query = f"""
+                WITH rids_undeployable_base_version AS (
```
wouterdb, line 5040: is this correct? 

This works if we assume there are no cross-set-dependencies, but do we enforce that?

wouterdb (commented):
(didn't make it to the end yet)

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -468,117 +395,155 @@ async def delete_version(self, env: data.Environment, version_id: int) -> Apiret
...
+        version: int,
+        rid_to_resource: Dict[ResourceIdStr, data.Resource],
+        unknowns: abc.Sequence[data.UnknownParameter],
+        version_info: Optional[JsonType] = None,
+        resource_sets: Optional[Dict[ResourceIdStr, Optional[str]]] = None,
+        partial_base_version: Optional[int] = None,
+        removed_resource_sets: Optional[List[str]] = None,
+        *,
+        connection: asyncpg.connection.Connection,
+    ) -> None:
```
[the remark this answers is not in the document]
arnaudsjs: Correct

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -611,72 +576,85 @@ def safe_get(input: object, key: str, default: object) -> object:
...
-        await cm.update_fields(total=cm.total + len(resources_to_purge), connection=connection)
+        all_resource_version_ids: abc.Set[ResourceVersionIdStr] = set(
+            Id.set_version_in_id(rid, version) for rid in rid_to_resource.keys()
+        )
+        if is_partial_update:
+            # Make mypy happy
+            assert partial_base_version is not None
+            rids_unchanged_resource_sets: abc.Set[
+                ResourceIdStr
+            ] = await data.Resource.copy_resources_from_unchanged_resource_set(
```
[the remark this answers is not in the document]
arnaudsjs: `connection` is never `None` is it. The type of connection is `connection: asyncpg.connection.Connection`

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -756,7 +738,7 @@ async def put_version(
         unknowns: List[Dict[str, PrimitiveTypes]],
         version_info: JsonType,
         compiler_version: Optional[str] = None,
-        resource_sets: Optional[Dict[ResourceIdStr, Optional[str]]] = None,
+        resource_sets: Dict[ResourceIdStr, Optional[str]] = {},
```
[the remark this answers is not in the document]
arnaudsjs: I just made the signature of this method in-line with the signature of the corresponding method in the methods.py file. The latter is used to by the protocol layer to set the defaults for example. So this should not break anything as this method is never called directly.

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: I don't think this check exists to verify whether all manipulation were done correctly. It exists as an input check to validate whether the constraints of a partial compile are fulfilled.

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: I agree. This check should only be executed on shared resources. I am not sure what you mean with "the create call" however.

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: There is one, but it's [here](https://github.com/inmanta/inmanta-core/blob/76b5e5eb149f82ca3ad156febf502b90216c4e52/src/inmanta/server/services/orchestrationservice.py#L816). Should I copy that check to the `_put_version()` method?

src/inmanta/data/__init__.py
arnaudsjs: [Here](https://github.com/inmanta/inmanta-core/blob/971d2c6e29bb400b358a1a53d1138ac8734c093f/src/inmanta/server/services/orchestrationservice.py#L543) the attributes are filtered from the resource dictionary sent to the server. It only filters out the `id` field and not the version field, while I think it should. I guess this is a bug that already exists for quite a while. Do you agree this is a bug?

wouterdb (commented):
central question: 
1. what is allowed in terms of inter-slice requires-provides?  (and where is this documented)


Details
1. looking at the `_create_dao_resources_from_api_resources` method, I would definitely make test cases that do weird things wrt requires-provides: 
   1.have resource A require a resource in the shared slice, the push an update removing  resource A, verify if provides is correctly updated in the shared slice. 
   1. the reverse (common requiring specific)
   1. have a requires that doesn't exist at all  (api safety: compiler can't do this)
   2. have a requires that doesn't exist in the current slice  (api safety: compiler can't do this)

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,204 @@
...
+        for res in new_updated_and_shared_resources:
+            if res.resource_id not in self.updated_and_shared_resources_old:
+                continue
+            matching_resource_old_model = self.updated_and_shared_resources_old[res.resource_id]
 
-class PartialUpdateMerger:
-    """
-    This class is used to merge the result of a partial compile with previous resources and resource_sets. Takes a partial spec
-    that should be applied on a given base version to form the given new version.
+            if res.resource_set != matching_resource_old_model.resource_set:
```
wouterdb, line 183: This works for movement between sets we have in memory, but what prevents movement between sets we don't have in memory? 

is there a database constraint for this? 

if so, why keep this duplicate? (I'm not implying the duplicate must go, I just want to understand why it exists)

src/inmanta/data/__init__.py
wouterdb: I do, I would lift it out of the 'attributes' and use the one in the resource.

wouterdb (commented):
Very impressive bit of code

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
sanderr: One more thing I find hard to reason on at the moment is which invariants are assumed to hold by this method and which are enforced. I think it would be good if the docstring clearly states the extent of this method's responsibilities. Otherwise I think it will be very hard to follow the flow in a few months.

i.e. does it check any of the partial compile invariants or are those are assumed to be met already (the second I think)? How does it handle requires-provides relations / undeployables / purge-on-delete / ... and which invariants of these properties are a precondition?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,204 @@
...
-        merged_resources: dict[ResourceIdStr, ResourceWithResourceSet] = self._merge_resources(
-            old_resources, paired_resources_partial, updated_resource_sets
-        )
-        return MergedModel(
-            resources=[r.resource.dict() for r in merged_resources.values()],
-            resource_states={rid: r.get_resource_state_for_new_resource() for rid, r in merged_resources.items()},
-            resource_sets=self._merge_resource_sets(old_resource_sets, updated_resource_sets),
-            unknowns=await self._merge_unknowns(merged_resources),
-        )
+        return [*old_unresolved_unknowns_to_keep, *unknowns_in_partial_compile]
```
sanderr, line 263: Why do we treat these differently from the resources? The resources we just copy in `_put_version`, but these we still take through Python. Is there a reason behind that?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -468,117 +395,155 @@ async def delete_version(self, env: data.Environment, version_id: int) -> Apiret
...
-        partial_base_version: Optional[int] = None,
-        *,
-        connection: asyncpg.connection.Connection,
-    ) -> None:
+        resource_sets: Dict[ResourceIdStr, Optional[str]],
+        set_version: Optional[int] = None,
+    ) -> Dict[ResourceIdStr, data.Resource]:
         """
-        :param resources: a list of serialized resources
+        Convert the resources sent to the put_version or put_partial endpoint to dao Resource objects.
```
sanderr, line 407: Could you elaborate a bit on what this means exactly? If I understand correctly, in addition to plainly creating the DAO, it updates the version field in the resource's attributes (if specified), verifies that the requires graph is closed within the provided resources, and sets up the `provides` edge for all cross agent dependencies. The end result are fully configured resource objects, including set up of inter-resource fields.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -611,72 +576,85 @@ def safe_get(input: object, key: str, default: object) -> object:
...
 
-        for agent in agents:
-            await self.agentmanager_service.ensure_agent_registered(env, agent, connection=connection)
+        await data.Resource.insert_many(list(rid_to_resource.values()), connection=connection)
+        await cm.recalculate_total(connection=connection)
+
+        await data.UnknownParameter.insert_many(unknowns, connection=connection)
+
+        for res in rid_to_resource.values():
+            await self.agentmanager_service.ensure_agent_registered(env, res.agent, connection=connection)
```
sanderr, line 646: I think it would be better to create a set of agents and only call the method for each agent once. The method acquires a session lock and does a query for each invocation, so it's not for free.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -468,117 +395,155 @@ async def delete_version(self, env: data.Environment, version_id: int) -> Apiret
...
+        Return the resources that are skipped_for_undeployable given the full set of resources and
+        the resource ids of the resources that are undeployable.
+
+        :param resources: All resources in the model.
+        :param undeployable_ids: The ids of the resource that are undeployable.
+        """
+        # Build up provides tree
+        provides_tree: Dict[ResourceIdStr, List[ResourceIdStr]] = defaultdict(lambda: [])
+        for r in resources:
+            if "requires" in r.attributes:
```
sanderr, line 489: All methods only consider `"requires"`. Don't our DAO's attributes' have a `"provides"` attribute? Do we convert those into requires at some point?

sanderr (commented):
Almost done: the final thing I'm still struggling with is the big picture of what happens where and whether everything is covered (I assume it is). So I need a bit more time to fully get there. Related to my docstring comment on `_put_version`.

src/inmanta/server/services/orchestrationservice.py
sanderr: For me it's fine to have preconditions here. I think we have to if we want to avoid one monolithic code block. But I do think they should be documented. But I left another comment stating just that so you can resolve this conversation as a subset of the other one.

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
arnaudsjs: ok. I will revert this change.

src/inmanta/data/__init__.py
arnaudsjs: I added this check.

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: The resource table has a primary key on `(environment, model, resource_id)`. So we will get a primary key violation when there is movement between resource sets that are not in memory.

src/inmanta/data/__init__.py
```diff
@@ -2942,6 +2942,61 @@ def copy(self, new_version: int) -> "UnknownParameter":
...
+                        AND u1.version=$2
+                        AND u1.resolved IS FALSE
+                        AND (u1.resource_id IS NULL OR u1.resource_id='')
+                )
+                UNION
+                -- Get unknowns that belong to a resource
+                (
+                    SELECT u2.*
+                    FROM {cls.table_name()} AS u2
+                       INNER JOIN {Resource.table_name()} AS r
```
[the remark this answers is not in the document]
arnaudsjs: Oh. That's much better. Don't know how I missed that.

src/inmanta/data/__init__.py
```diff
@@ -4527,6 +4582,30 @@ def new(cls, environment: uuid.UUID, resource_version_id: m.ResourceVersionIdStr
...
+        return Resource(
+            environment=self.environment,
+            model=new_version,
+            resource_id=self.resource_id,
+            resource_type=self.resource_type,
+            resource_id_value=self.resource_id_value,
+            agent=self.agent,
+            last_deploy=None,
+            attributes=attributes_with_version_update,
+            attribute_hash=self.attribute_hash,
```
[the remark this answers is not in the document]
arnaudsjs: This comment has become out of date as Wouter requested to remove the version field from the attributes dictionary in the database.

src/inmanta/data/__init__.py
```diff
@@ -4728,6 +4807,95 @@ async def get_resource_deploy_summary(cls, environment: uuid.UUID) -> m.Resource
...
+                        THEN 'undefined'::resourcestate
+                        ELSE 'available'::resourcestate
+                        END
+                    ) AS status,
+                    jsonb_set(r.attributes, '{{"version"}}'::text[], to_jsonb($3::integer)) AS attributes,
+                    r.attribute_hash,
+                    r.resource_set,
+                    r.provides
+                FROM {cls.table_name()} AS r
+                WHERE r.environment=$1 AND r.model=$2 AND r.resource_set IS NOT NULL AND NOT r.resource_set=ANY($4)
```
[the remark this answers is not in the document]
arnaudsjs: I don't see how this filter would work incorrectly when the resource set name is an empty string.
```
arnaud=> select * FROM (VALUES (''), ('test')) AS tab(val) WHERE NOT val=ANY('{""}');
 val  
------
 test
(1 row)

arnaud=> select * FROM (VALUES (''), ('test')) AS tab(val) WHERE NOT val=ANY('{"test"}');
 val 
-----
 
(1 row)

```

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
sanderr: Even better!

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
sanderr: I mean if `r.resource_set` is the empty string but it's not in `$4`, as in your second example. This query does not exclude it, but other queries do have a `... AND r.resource_set != '' AND ...`. I think either all queries should exclude the empty string, or none of them should.

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: I mainly didn't do this because the `UnknownParameter` table has id (uuid) as primary key. If we would copy the records from the base version to the new version, we would have to generate a new id from each copied unknown. Postgresql seems to have a [function](https://www.postgresql.org/docs/13/functions-uuid.html) for that, but it's only available in PG13+ and not in PG10. As the number of unknowns are usually small, I thought it wouldn't be an issue to do the merging in Python. On the other hand we merge this change only into ISO6+ which requires PG13, so it wouldn't be much of an issue. But it might break some of our internal labs.

sanderr: Your reasoning makes sense to me, you can consider this conversation resolved.

As to your final comment, I'm not sure how much of our infra still runs PG<13. @wouterdb, you added support for PG13 in the postgresql module in inmanta/postgresql#239. Did you also update the test orchestrators and other clients to use PG13, or is that still outstanding?

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: As discussed, resources sent to the server only contain the requires relationship and not the provides relationship. The compiler turns the provides into requires.

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
arnaudsjs: I see. After simplifying the query in `get_unknowns_to_copy_in_partial_compile()`. All queries now consider the resource set with the empty string as a name as an actual resource set.

arnaudsjs:
> 1. what is allowed in terms of inter-slice requires-provides?  (and where is this documented)

It's documented here https://docs.inmanta.com/inmanta-service-orchestrator-dev/6/model_developers/resource_sets.html#constraints-and-rules. Is that sufficient?

src/inmanta/data/__init__.py
```diff
@@ -4875,7 +4869,12 @@ async def copy_resources_from_unchanged_resource_set(
                         ELSE 'available'::resourcestate
                         END
                     ) AS status,
-                    jsonb_set(r.attributes, '{{"version"}}'::text[], to_jsonb($3::integer)) AS attributes,
+                    (
+                        CASE WHEN r.attributes ? 'version'
+                        THEN jsonb_set(r.attributes, '{{"version"}}'::text[], to_jsonb($3::integer))
+                        ELSE r.attributes
+                        END
+                    ) AS attributes,
```
sanderr, line 4877: Why not just delete it in a schema migration script? Or at least delete it here instead of updating it?

src/inmanta/data/__init__.py
```diff
@@ -5030,129 +5039,125 @@ def done(self) -> int:
...
-                version_info=version_info,
-                undeployable=undeployable,
-                skipped_for_undeployable=skipped_for_undeployable,
+        Create and insert a new configurationmodel that is the result of a partial compile. The new ConfigururationModel will
+        contain all the undeployables and skipped_for_undeployables present in the partial_base version that are not part of
+        the partial compile, i.e. not present in rids_in_partial_compile.
+        """
+        query = f"""
+            WITH base_version_exists AS (
+                SELECT COUNT(*) != 0 AS base_version_found
```
sanderr, line 5061: This is equivalent to `EXISTS`, isn't it?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,299 @@
...
-        if self.resource_state is ResourceState.undefined:
-            return ResourceState.undefined
-        else:
-            return ResourceState.available
-
-    def is_shared_resource(self) -> bool:
-        return self.resource_set is None
+        if res.resource_set is None:
+            # Resource is in shared resource set.
+            return False
```
wouterdb, line 95: # pass I
A dependency starting in the shared resource set should remain in the shared resource set.

So I think this would allow the shared set to refer back to others,...

# pass II
Looking deeper into the code, it seems that this IS supported, which is pretty cool, but it will take some docs

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,299 @@
...
+        """
+        This method raises a CrossResourceSetDependencyError when a resource in self.resources that belongs to a non-shared
+        resource set has a dependency (requires/provides) on another resource that belongs to a different non-shared resource
+        set.
+        """
+        for res in self.resources:
+            for req in res.get_requires():
+                if self._is_cross_resource_set_dependency(res, req):
+                    raise CrossResourceSetDependencyError(res.resource_id, req)
+            for prov in res.provides:
```
wouterdb, line 116: Why check both requires and provides? If the one checks out, the other should also check out. 

(and provides it not complete in most cases, as it only contains CAD's in the database?)

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,299 @@
...
-    unknowns: List[data.UnknownParameter]
+        :param new_updated_and_shared_resources: The resources that have to be validated.
+        """
+        for res_id, res in new_updated_and_shared_resources.items():
+            if res.resource_id not in self.updated_and_shared_resources_old:
+                continue
+            matching_resource_old_model = self.updated_and_shared_resources_old[res.resource_id]
 
+            if res.resource_set != matching_resource_old_model.resource_set:
+                raise BadRequest(f"A partial compile cannot migrate resource {res.resource_id} to another resource set.")
```
wouterdb, line 253:
```suggestion
                raise BadRequest(f"A partial compile cannot migrate resources: trying to move {res.resource_id} from resource set {matching_resource_old_model.resource_set} to {res.resource_set} ")
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,204 @@
...
-            paired_resources.append(pair)
-        return paired_resources
+        all_rids_shared_resources = set(self.shared_resources_old.keys()) | set(shared_resources_new.keys())
+        result = []
+        for rid_shared_resource in all_rids_shared_resources:
+            if rid_shared_resource in shared_resources_new and rid_shared_resource in self.shared_resources_old:
+                # Merge requires/provides shared resource
+                old_shared_resource = self.shared_resources_old[rid_shared_resource]
+                new_shared_resource = shared_resources_new[rid_shared_resource]
+                res = self._merge_requires_and_provides_of_shared_resource(old_shared_resource, new_shared_resource)
```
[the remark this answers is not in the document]
wouterdb: it does provide some powerful capabilities!

src/inmanta/server/services/orchestrationservice.py
wouterdb: We still support RHEL7, so I think we need to stay compatible with older PG's for a while.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -471,140 +493,354 @@ async def delete_version(self, env: data.Environment, version_id: int) -> Apiret
     async def reserve_version(self, env: data.Environment) -> int:
         return await env.get_next_version()
 
-    async def _put_version(
+    def _create_dao_resources_from_api_resources(
```
wouterdb, line 496: note: I'm here

src/inmanta/db/versions/v202303010.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    # Create column
+    await connection.execute("ALTER TABLE public.configurationmodel ADD COLUMN is_suitable_for_partial_compiles boolean")
+    # Populate column
+    records = await connection.fetch(f"SELECT environment, version FROM {data.ConfigurationModel.table_name()}")
+    for record in records:
+        environment = record["environment"]
+        version = record["version"]
+        resources = await data.Resource.get_resources_for_version(environment, version, connection=connection)
+        resource_set_validator = ResourceSetValidator(set(resources))
```
sanderr, line 37: I think this isn't necessarily safe: it seems to me to be very similar to [ADR 2](https://github.com/inmanta/inmanta-core/blob/master/docs/adr/0002-database-upgrade-testing.md) but for the actual migration, which is more critical. I think it would be difficult to enforce this check without using these methods though, so perhaps we should just put it to `False` for all old resources and include an upgrade note in the changelog?

wouterdb (commented):
I'm about half way, super cool stuff this

src/inmanta/server/services/orchestrationservice.py
sanderr: I don't think we do support RHEL7 for iso5+ though.

EDIT: man, it's annoying that GitHub doesn't allow you to comment directly while having a review in progress...

tests/test_data.py
```diff
@@ -1506,7 +1506,7 @@ async def insert_resource(env_id, version, agent_name, path, status):
     assert len(result) == 3
     assert sorted([r["agent"] for r in result]) == ["agent1", "agent1", "agent2"]
     for r in result:
-        assert len(r["attributes"]) == 1
+        assert len(r["attributes"]) == 2  # path and version attribute
```
sanderr, line 1509: Could you elaborate on what exactly changed here? I got the impression we inserted the version field for backwards compatibility, but this assertion seems to imply otherwise.

sanderr (commented):
Still need to go over the orchestrationservice changes

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -611,72 +576,85 @@ def safe_get(input: object, key: str, default: object) -> object:
...
                 )
-                resource_objects.append(res_obj)
 
                 previous_requires[res_obj.resource_id] = res.attributes["requires"]
-                resource_version_ids.append(res_obj.resource_version_id)
-                agents.add(res_obj.agent)
-                rv_dict[res_obj.resource_id] = res_obj
+                rid_to_resource[res_obj.resource_id] = res_obj
 
             # invert dependencies on purges
```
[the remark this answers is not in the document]
wouterdb: We do it because most often, deletes have to be done in reverse order (i.e. create folder, then file, delete file, then folder)

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
sanderr: I'm convinced now! I do think a picture of resource sets and their allowed edges to the shared resource set (both for normal relations and requires-provides) in the docs would be very useful. Not necessarily as part of this ticket, since nothing new was introduced really, just refined.

wouterdb (approved):

src/inmanta/server/services/orchestrationservice.py
sanderr: See [iso4 docs](https://docs.inmanta.com/inmanta-service-orchestrator/4/install/1-install-server.html#step-2-install-postgresql-10) vs [iso5 docs](https://docs.inmanta.com/inmanta-service-orchestrator/5/install/1-install-server.html#step-2-install-postgresql-13). Both RHEL versions and Postgres version are mentioned there.

src/inmanta/db/versions/v202302170.py
```diff
@@ -0,0 +1,23 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    Contact: code@inmanta.com
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
```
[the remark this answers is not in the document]
arnaudsjs: Oh. That's not intentional. I am going to merge them together.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -517,6 +624,38 @@ async def _put_version(
...
+                         Also the unknowns for resources that are not present in rid_to_resource.
+        :param partial_base_version: When a partial compile is done, this parameter contains the version of the
+                                     configurationmodel this partial compile was based on. Otherwise this parameter should be
+                                     None.
+        :param removed_resource_sets: When a partial compile is done, this parameter should indicate the names of the resource
+                                      sets that are removed by the partial compile. When no resource sets are removed by
+                                      a partial compile or when a full compile is done, this parameter can be set to None.
+
+        Pre-conditions:
+            * The requires and provides relationships of the resources in rid_to_resource must to be set correctly.
```
sanderr, line 641:
```suggestion
            * The requires and provides relationships of the resources in rid_to_resource must be set correctly. For a partial export, this means it is assumed to be valid with respect to all absolute constraints that apply to partial compiles. Constraints that are relative to the base version will be verified by this method.
```
It wasn't clear to me what this meant exactly. I think this is the essence of it?

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
sanderr: Thanks, I feel like the docstring adds a lot of clarity as to the responsibilities of the different parts.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -548,23 +687,155 @@ async def _put_version(
...
+                )
+                await ra.insert(connection=connection)
+
+        LOGGER.debug("Successfully stored version %d", version)
+
+    async def _get_resources_for_purge_on_delete(
+        self,
+        env: data.Environment,
+        version: int,
+        rid_to_resource: Dict[ResourceIdStr, data.Resource],
```
sanderr, line 798:
```suggestion
        rid_to_resource: abc.Mapping[ResourceIdStr, data.Resource],
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -548,23 +687,155 @@ async def _put_version(
...
+        all_resource_ids: abc.Sequence[ResourceIdStr],
+        version_info: Optional[JsonType] = None,
+        *,
+        connection: asyncpg.connection.Connection,
+    ) -> Dict[ResourceIdStr, data.Resource]:
```
sanderr, line 803:
```suggestion
    ) -> dict[ResourceIdStr, data.Resource]:
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -517,6 +624,38 @@ async def _put_version(
         *,
         connection: asyncpg.connection.Connection,
     ) -> None:
+        """
+        :param rid_to_resource: This parameter should contain all the resources when a full compile is done.
+                                When a partial compile is done, it should contain all the resources that belong to the
+                                updated resource sets or the shared resource sets.
```
sanderr, line 630: The object is mutated by this method, may be useful to mention that?
```suggestion
                                updated resource sets or the shared resource sets.
                                This method updates this object with purge-on-delete resources.
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -517,6 +624,38 @@ async def _put_version(
...
+        :param unknowns: This parameter should contain all the unknowns for all the resources in the new version of the model.
+                         Also the unknowns for resources that are not present in rid_to_resource.
+        :param partial_base_version: When a partial compile is done, this parameter contains the version of the
+                                     configurationmodel this partial compile was based on. Otherwise this parameter should be
+                                     None.
+        :param removed_resource_sets: When a partial compile is done, this parameter should indicate the names of the resource
+                                      sets that are removed by the partial compile. When no resource sets are removed by
+                                      a partial compile or when a full compile is done, this parameter can be set to None.
+
+        Pre-conditions:
```
sanderr, line 640: Perhaps add a section for validations performed by this method as well?
- In case of a full export, checks whether this version has any requires-provides across resource sets and sets the `is_suitable_for_partial_compiles` field appropriately, indicating whether this version is eligible to be used as a base version for a future partial compile.
- In case of a partial export, verifies that no resources moved resource sets.

sanderr (commented):
I only need to review the `ResourceSetValidator` class now.

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
sanderr: I looked it over once more and I draw the following conclusions:
1. My earlier statement about inverting the requires relation twice doesn't hold.
2. Suppose `a` requires `b`, where `a` and `b` live on different agents and only `a` is deleted from the model. Then this logic goes through the following steps
  a. empty `a`'s requires and provides
  b. leave `b`'s requires and provides as is because it is not being purged
  c. invert: set `b.requires = a` and `a.provides = b`
  => `b` still has `b.provides = a`, resulting in a circular (and incorrect) dependency. A similar issue occurs if the dependency would have been the other way around.

Do you agree (2) is a bug? I don't think it's introduced by this PR.

sanderr (commented):
Gone through everything now. No further comments.

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: You are right. It should be sufficient to check on the requires. I will remove provides check and add a comment.

src/inmanta/data/__init__.py
arnaudsjs: Yes. I turned it into an EXISTS query.

tests/test_data.py
arnaudsjs: When the exporter sends resources to the server, the version field is included in two different locations. Once in the version field of the dictionary and once in the `attributes["version"]` field of the dictionary. It was decided that we want to keep the API backwards compatible, but that we don't want to include the redundant version data into the database. The problem is that when we remove the version field from the attributes dictionary in the database, we have no way of knowing whether the version field was there when the resources were sent to the server. This is not an issue in practice because the exporter always does, but this test case doesn't. Maybe I should have fixed this test case the other way around, namely adding the version field to the attributes dictionary.

sanderr: I understand. We don't use the exporter here so we never send the version field. We used to get back exactly what we sent in, but now the version field is added in the layer above the database because we don't persist it anymore, so we always get it out, even if we didn't include it in the insert.

src/inmanta/data/__init__.py
arnaudsjs: Good point. I completely overlooked that. I think the removing the version field using a database migration script is the way to go. I will fix this.

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
wouterdb: looks like a bug to me indeed

src/inmanta/db/versions/v202303010.py
arnaudsjs: It defaults to False now.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -110,12 +110,11 @@ def ensure_no_cross_resource_set_dependencies(self) -> None:
         set.
         """
         for res in self.resources:
+            # It's sufficient to only check the requires relationship. The provides
+            # relationship is always a subset of the provides relationship.
```
sanderr, line 114:
```suggestion
            # It's sufficient to only check the requires relationship. The provides
            # relationship is always a subset of the reverse relationship.
```
I'm not sure if this is what you're trying to say?

sanderr (changes requested):
Last comment, promise!

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: Yes. Indeed. This is what I wanted to day. I will make this more clear.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -63,277 +63,302 @@
...
+    def ensure_no_cross_resource_set_dependencies(self) -> None:
+        """
+        This method raises a CrossResourceSetDependencyError when a resource in self.resources that belongs to a non-shared
+        resource set has a dependency (requires/provides) on another resource that belongs to a different non-shared resource
+        set.
+        """
+        for res in self.resources:
+            # It's sufficient to only check the requires relationship. The provides
+            # relationship is always a subset of the reverse relationship (provides relationship).
+            # The provides relationship only contains the cross-agent dependencies.
```
sanderr, line 115:
```suggestion
            # It's sufficient to only check the requires relationship. The provides
            # field is always a subset of the reverse relationship (provides relationship).
            # The provides field only contains the cross-agent dependencies.
```
Perhaps this makes it even more clear?

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
arnaudsjs: I agree this is a bug.

PR #5587, opened 2023-02-17
Issue/5543 resource file upload should ignore identical files

### LINKED ISSUE #5543: resource file upload should ignore identical files
The current bulk resource file upload may fail when it gets the same file contents more than once in the same request. Because of #2313 this may happen in practice (see its first comment). Identical files should instead be ignored, they are identical after all.

Fix this in the `stat_files` endpoint.

Hugo-Inmanta:
# Description

The stat_file API endpoint now returns a set of files instead of a list.

closes #5543  

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5543-resource-file-upload-should-ignore-identical-files.yml
+++ b/changelogs/unreleased/5543-resource-file-upload-should-ignore-identical-files.yml
@@ -5,4 +5,4 @@ destination-branches:
   - iso6
   - iso5
 sections:
-  bugfix: "{{description}}"
+  bugfix: "Fix issue that may cause the first export for an environment to fail when files with identical content are present."
--- a/changelogs/unreleased/5543-resource-file-upload-should-ignore-identical-files.yml
+++ b/changelogs/unreleased/5543-resource-file-upload-should-ignore-identical-files.yml
@@ -1,4 +1,4 @@
-description: The stat_file API endpoint now returns a set of files instead of a list.
+description: The stat_file API endpoint now returns a list with no duplicates.
 change-type: patch
 destination-branches:
   - master
--- a/changelogs/unreleased/5543-resource-file-upload-should-ignore-identical-files.yml
+++ b/changelogs/unreleased/5543-resource-file-upload-should-ignore-identical-files.yml
@@ -4,3 +4,5 @@ destination-branches:
   - master
   - iso6
   - iso5
+sections:
+  bugfix: "{{description}}"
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/5543-resource-file-upload-should-ignore-identical-files.yml
```diff
@@ -0,0 +1,8 @@
+description: The stat_file API endpoint now returns a list with no duplicates.
+change-type: patch
+destination-branches:
+  - master
+  - iso6
+  - iso5
+sections:
+  bugfix: "{{description}}"
```
Hugo-Inmanta, line 8: I don't have a strong opinion about this changelog entry being present or not

src/inmanta/server/services/fileservice.py
```diff
@@ -131,13 +131,14 @@ def stat_file_internal(self, files: Iterable[str]) -> List[str]:
         """
         Return which files in the list don't exist on the server
         """
-        response: List[str] = []
+        response: Dict[str, object] = {}
```
Hugo-Inmanta, line 134: I decided to use a dict here where only the keys are of interest to have an efficient set-like behavior and preserving insertion order compared to the previous implementation.
I'm not sure whether this is truly necessary or not

changelogs/unreleased/5543-resource-file-upload-should-ignore-identical-files.yml
sanderr: I think the message is good as an internal description of the change. For the end-user message I'd focus more on the symptom: this fixes a bug that may cause the first export for an environment to fail if there are identical files.

src/inmanta/server/services/fileservice.py
sanderr: Looks good to me. Perhaps add a comment.

sanderr, line 134:
```suggestion
        response: dict[str, object] = {}
```

sanderr (approved):

sanderr (changes requested):
Forgot that I requested a change. The one to the change entry, the rest are minor.

changelogs/unreleased/5543-resource-file-upload-should-ignore-identical-files.yml
Hugo-Inmanta: Great point :+1:

sanderr (approved):

sanderr:
@bartv FYI

PR #5596, opened 2023-02-20
Fix race in incremental deploy

arnaudsjs:
# Description

This PR fixes a race condition in the incremental deploy calculation. The following scenario triggers the race:

* `put_version()` is called for model version N and the increment cache is cleared.
* An agent pulls the resources it has to deploy. This populates the increment cache with an increment relative to version N-1
* The `release_version()` method is called for version N. This call hits the increment cache entry created in the previous bullet point, resulting in an incorrect increment calculation.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/orchestrationservice.py
+++ b/src/inmanta/server/services/orchestrationservice.py
@@ -687,8 +687,6 @@ class OrchestrationService(protocol.ServerSlice):
 
         LOGGER.debug("Successfully stored version %d", version)
 
-        self.resource_service.clear_env_cache(env)
-
     async def _trigger_auto_deploy(
         self,
         env: data.Environment,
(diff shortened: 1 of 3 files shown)


sanderr (approved):

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -687,8 +687,6 @@ def safe_get(input: object, key: str, default: object) -> object:
 
         LOGGER.debug("Successfully stored version %d", version)
 
-        self.resource_service.clear_env_cache(env)
-
```
Hugo-Inmanta, line 691: Don't we still have to clear the cache somewhere ?

Hugo-Inmanta (approved):
Just one question, see comment

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: No. The cache is cleared implicitly in the `release_version()` method, because the version number used there is always higher than the currently deployed version. So `release_version` will always have a cache miss on the increment cache, but we never have to clear the cache explicitly.

PR #5666, opened 2023-03-06
Fix: issue with table name in migration test 

FloLey:
# Description

a migration test uses the `get_columns_in_db_table(table_name: str)`. It was called with `table_name = ConfigurationModel.table_name()` this function also returns the schema of the table in a string. By using `ConfigurationModel.table_name(include_schema=False)` this is fixed.

# Self Check:

- [x] Changelog entry


tests/db/migration_tests/test_v202206290_to_v202208180.py
```diff
@@ -34,13 +34,13 @@ async def test_column_add(
     """
 
     # Assert state before running the DB migration script
-    assert "partial_base" not in (await get_columns_in_db_table(ConfigurationModel.table_name()))
+    assert "partial_base" not in (await get_columns_in_db_table(ConfigurationModel.table_name(include_schema=False)))
```
wouterdb, line 37: Why not make the `ConfigurationModel.table_name`  backward compatible by defaulting to false?

wouterdb (changes requested):

tests/db/migration_tests/test_v202206290_to_v202208180.py
arnaudsjs: I was going to make the same remark.

arnaudsjs (changes requested):
See comment

wouterdb (approved):

src/inmanta/data/__init__.py
```diff
@@ -5547,6 +5542,13 @@ class User(BaseDocument):
     password_hash: str
     auth_method: AuthMethod
 
+    @classmethod
+    def table_name(cls) -> str:
+        """
+        Return the name of the user table prepended by "public." to differentiate it from the user table in the pg schema.
+        """
+        return f"public.{cls.__name__.lower()}"
+
```
arnaudsjs, line 5551: I find this code rather confusing. This is a single method that behaves differently for one specific subclass. Why didn't you choose for the approach where the method is parameterized and the default value is set to False?

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
wouterdb: This hadn't struck me, but I do like the use of polymorphism here.

FloLey: because I caused fall outs at multiple places. For example in the `insert` method:
It calls `self.table_name()` but If I want to specify the parameter to it (which I need for the `insert` method of a User) , it will also become a parameter of the `insert` method which I didn't like.

arnaudsjs: It doesn't have to become a parameter of the `insert` method, does it. Can't you just use `table_name(include_schema=True)` in the insert method?

FloLey: Indeed, I made the change at all the other places it was needed. I personally have no strong preference. Both seem readable to me.

wouterdb: @arnaudsjs I do like the use of polymorphism here a lot. 

Because here it is clear why it is different from the others (because of the name of this table).
And it still does what the `table_name` method is expected to do: return the table name to use in queries. 

In the insert method on the parent class, why it needs the prefix is not at all apparent.

arnaudsjs (changes requested):
I set the state to request changes to make sure that I get a notification when the discussed changes are implemented.

arnaudsjs (approved):

PR #5695, opened 2023-03-15
Issue/5458 compiler gradual progress potential

sanderr:
# Description

I still need to consider whether it's complete and I haven't added any tests yet. I thought I'd ask review for the concept of the implementation first before I start refining it.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -182,7 +182,7 @@ class GradualFor(ResultCollector[object]):
         self.seen: set[int] = set()
 
     @classmethod
-    def gradual_only(self) -> bool:
+    def pure_gradual(self) -> bool:
         # freezing the iterable allows dropping eager promises
         return False
 
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -182,8 +182,7 @@ class GradualFor(ResultCollector[object]):
         self.seen: set[int] = set()
 
     # TODO: add test and drop
-    @classmethod
-    def pure_gradual(cls) -> bool:
+    def pure_gradual(self) -> bool:
         # freezing the iterable allows dropping eager promises
         return False
 
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -181,9 +181,9 @@ class GradualFor(ResultCollector[object]):
         self.stmt = stmt
         self.seen: set[int] = set()
 
-    def receive_result(self, value: object, location: ResultVariable) -> None:
+    def receive_result(self, value: object, location: ResultVariable) -> bool:
         if id(value) in self.seen:
-            return
+            return False
         self.seen.add(id(value))
 
         xc = ExecutionContext(self.stmt.module, self.resolver.for_namespace(self.stmt.module.namespace))
@@ -193,6 +193,7 @@ class GradualFor(ResultCollector[object]):
         assert isinstance(loopvar, ResultVariable)
         loopvar.set_value(value, self.stmt.location)
         xc.emit(self.queue)
+        return False
 
 
 class For(RequiresEmitStatement):
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -181,8 +181,9 @@ class GradualFor(ResultCollector[object]):
         self.stmt = stmt
         self.seen: set[int] = set()
 
+    # TODO: add test and drop
     @classmethod
-    def pure_gradual(self) -> bool:
+    def pure_gradual(cls) -> bool:
         # freezing the iterable allows dropping eager promises
         return False
 
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -181,11 +181,6 @@ class GradualFor(ResultCollector[object]):
         self.stmt = stmt
         self.seen: set[int] = set()
 
-    # TODO: add test and drop
-    def pure_gradual(self) -> bool:
-        # freezing the iterable allows dropping eager promises
-        return False
-
     def receive_result(self, value: object, location: ResultVariable) -> bool:
         if id(value) in self.seen:
             return False
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -181,6 +181,11 @@ class GradualFor(ResultCollector[object]):
         self.stmt = stmt
         self.seen: set[int] = set()
 
+    @classmethod
+    def gradual_only(self) -> bool:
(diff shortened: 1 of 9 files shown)


wouterdb (commented):
Cool document!

src/inmanta/execute/runtime.py
```diff
@@ -681,10 +730,9 @@ def __str__(self) -> str:
         return "ListVariable %s %s = %s" % (self.myself, self.attribute, self.value)
 
     def get_progress_potential(self) -> int:
-        """How many are actually waiting for us"""
         # Ensure that relationships with a relation precedence rule cannot end up in the zerowaiters queue
         # of the scheduler. We know the order in which those types can be frozen safely.
-        return len(self.waiters) - len(self.listeners) + int(self.attribute.has_relation_precedence_rules())
+        return super().get_progress_potential() + int(self.attribute.has_relation_precedence_rules())
```
wouterdb, line 735: what does this additional method call do to performance?

src/inmanta/ast/statements/generator.py
```diff
@@ -181,9 +181,14 @@ def __init__(self, stmt: "For", resolver: Resolver, queue: QueueScheduler) -> No
         self.stmt = stmt
         self.seen: set[int] = set()
 
-    def receive_result(self, value: object, location: ResultVariable) -> None:
+    @classmethod
+    def pure_gradual(self) -> bool:
+        # freezing the iterable allows dropping eager promises
+        return False
```
wouterdb, line 187: but it will never cause an additional instance to be added to a relation directly. 

I find this one very weird. 

For loops are the most gradual thing there is, they are pure for each, they only produce something if there is more ouput. 


I would expect this to break many models, because the first relation feeding into a for loop will possible get progress potential. But the one feeding into it does not.

wouterdb (changes requested):

src/inmanta/execute/runtime.py
```diff
@@ -508,10 +527,15 @@ class BaseListVariable(DelayedResultVariable[ListValue]):
...
     value: "List[Instance]"
 
-    __slots__ = ()
+    __slots__ = ("_listeners", "_done_listeners", "_nb_pure_listeners")
 
     def __init__(self, queue: "QueueScheduler") -> None:
-        self.listeners: List[ResultCollector[ListValue]] = []
+        # keep track of listeners and whether they may cause progress when this variable is frozen
+        # use dict for easy lookup with reliable ordering
+        self._listeners: Optional[dict[ResultCollector[ListValue], None]] = {}
```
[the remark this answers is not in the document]
sanderr: I expect this to be more efficient because of the cleanup. I'll try to do some compiler scaling tests with this and your other performance related comment in mind.

src/inmanta/ast/statements/generator.py
sanderr: Added a test case that triggers what you describe in 1200114.

sanderr:
# Performance testing (WIP)

# Unmodified compilerscaling
Negligible execution time difference, slight memory increase:
- master: `Compile with 10000 elines took 276sec. RAM usage=2193.54296875MB:`
- this branch (49bb957): `Compile with 10000 elines took 275sec. RAM usage=2228.96875MB:`

# Compilerscaling with few is defined
*(I ran these on my laptop and the unmodified ones on Jenkins so they're not cross-comparable)*

```diff
diff --git a/libs/eline/model/_init.cf b/libs/eline/model/_init.cf
index a5e38d2..0294543 100644
--- a/libs/eline/model/_init.cf
+++ b/libs/eline/model/_init.cf
@@ -64,6 +64,11 @@ implementation eline for Eline:
     a = mefce::PE[name=self.pe_a]
     z = mefce::PE[name=self.pe_z]
 
+    if evc.network_interfaces is defined:
+    end
+    if evc.network_elements is defined:
+    end
+
     # Create an eline and associate NI's with the eline
     evc = mefce::ELine(
         parent=self,
```

- master: ?
- this branch (49bb957): `Compile with 10000 elines took 241sec. RAM usage=2246.609375MB:`

src/inmanta/execute/runtime.py
```diff
@@ -576,18 +615,36 @@ def set_value(self, value: ListValue, location: Location, recur: bool = True) ->
...
         for value in self.value:
             resultcollector.receive_result(value, location)
         if not self.hasValue:
-            self.listeners.append(resultcollector)
+            assert self._listeners is not None
+            if resultcollector in self._listeners:
+                # may happen in case of a duplicate assignment, e.g. `x.a = [y.a, y.a]`
+                self._done_listeners += 1
+                return
+            assert resultcollector not in self._listeners, "Invalid compiler state: ResultCollector registered twice"
```
sanderr, line 631:
```suggestion
```

src/inmanta/execute/runtime.py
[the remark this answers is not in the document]
sanderr: It requires about 1% more memory in the modified compilerscaling test, and the cleanup doesn't seem to have much of an effect so if I dropped it and made this a plain list again.

src/inmanta/execute/runtime.py
sanderr: Negligible.

wouterdb (commented):
reviewed up to here, looks nice

src/inmanta/execute/runtime.py
```diff
@@ -597,14 +598,12 @@ def _notify_listeners(self, value: ListValue, location: Location) -> None:
...
             done: bool = listener.receive_result(value, location)
             if done:
-                if listener.pure_gradual():
-                    self._nb_pure_listeners -= 1
-                # keep memory footprint minimal
-                del self._listeners[listener]
-                self._done_listeners += 1
+                # TODO: comments
+                if not listener.pure_gradual():
+                    self._nb_pure_listeners += 1
```
sanderr, line 606: TODO: create a test case that catches this bug: listener keeps getting results => count keeps increasing.

src/inmanta/execute/runtime.py
```diff
@@ -18,7 +18,7 @@
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Deque, Dict, Generic, Hashable, List, Optional, Set, TypeVar, Union, cast
 
-import inmanta.ast.attribute  # noqa: F401 (pep8 does not recognize partially qualified access ast.attribute)
+import inmanta.ast.attribute  # noqa: F401 (pyflakes does not recognize partially qualified access ast.attribute)
```
sanderr, line 21: @wouterdb any opinion on this? I don't like the `noqa`, and I'm not sure about the practice in the first place, but I do really prefer `ast.attribute.RelationAttribute` over both `inmanta.ast.attribute.RelationAttribute` and `attribute.RelationAttribute`. Is this too exotic?

wouterdb (approved):

PR #5696, opened 2023-03-15
Add support for 4 digit versioning

hlloreda:
# Description

Added support for 4 digit versioning needed in some repos where we want to follow other products versioning but still able to provide revisions

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/add-four-digit-versioning-module-tool.yml
+++ b/changelogs/unreleased/add-four-digit-versioning-module-tool.yml
@@ -1,5 +1,5 @@
 ---
-description: Add support for four digit versioning for module tool (release).
+description: Add support for four digit versioning for `inmanta module release`.
 change-type: minor
 destination-branches: [master, iso6]
 sections:
--- a/changelogs/unreleased/add-four-digit-versioning-module-tool.yml
+++ b/changelogs/unreleased/add-four-digit-versioning-module-tool.yml
@@ -1,4 +1,6 @@
 ---
-description: Add four digit versioning for module tool (release).
+description: Add support for four digit versioning for module tool (release).
 change-type: minor
 destination-branches: [master, iso6]
+sections:
+  minor-improvement: "{{description}}"
--- a/changelogs/unreleased/add-four-digit-versioning-module-tool.yml
+++ b/changelogs/unreleased/add-four-digit-versioning-module-tool.yml
@@ -1,0 +1,4 @@
+---
+description: Add four digit versioning for module tool (release).
+change-type: minor
+destination-branches: [master, iso6]
--- a/src/inmanta/moduletool.py
+++ b/src/inmanta/moduletool.py
@@ -103,17 +103,21 @@ def add_deps_check_arguments(parser: argparse.ArgumentParser) -> None:
         dest="no_strict_deps_check",
         action="store_true",
         default=False,
-        help="When this option is enabled, only version conflicts in the direct dependencies will result in an error. "
-             "All other version conflicts will result in a warning. This option is mutually exclusive with the "
+        help=(
+            "When this option is enabled, only version conflicts in the direct dependencies will result in an error. ",
+             "All other version conflicts will result in a warning. This option is mutually exclusive with the ",
              "--strict-deps-check option.",
+        ),
     )
     parser.add_argument(
         "--strict-deps-check",
         dest="strict_deps_check",
         action="store_true",
         default=False,
-        help="When this option is enabled, a version conflict in any (transitive) dependency will results in an error. "
+        help=(
+            "When this option is enabled, a version conflict in any (transitive) dependency will results in an error. ",
              "This option is mutually exclusive with the --no-strict-deps-check option.",
+        )
     )
 
 
@@ -158,7 +162,7 @@ class ModuleLikeTool(object):
             project.load()
         return project
 
-    def determine_new_version(self, old_version, version, major, minor, patch, revision, dev):
+    def determine_new_version(self, old_version, version, major, minor, patch, dev):
         """
         Only used by the `inmanta module commit` command.
         """
@@ -192,6 +196,8 @@ class ModuleLikeTool(object):
                     LOGGER.error("You can use only one of the following options: --major, --minor or --patch")
                     return None
 
+                # We do not support revision for deprecated methods
+                revision = False
                 change_type: Optional[ChangeType] = ChangeType.parse_from_bools(revision, patch, minor, major)
                 if change_type:
                     outversion = str(VersionOperation.bump_version(change_type, old_version, version_tag=""))
@@ -217,18 +223,6 @@ class ChangeType(enum.Enum):
     PATCH: str = "patch"
     REVISION: str = "revision"
 
-    def less(self) -> Optional["ChangeType"]:
-        """
-        Returns the change type that is one less than this one.
-        """
-        if self == ChangeType.MAJOR:
-            return ChangeType.MINOR
-        if self == ChangeType.MINOR:
-            return ChangeType.PATCH
-        if self == ChangeType.PATCH:
-            return ChangeType.REVISION
-        return None
-
     def __lt__(self, other: "ChangeType") -> bool:
         order: List[ChangeType] = [ChangeType.REVISION, ChangeType.PATCH, ChangeType.MINOR, ChangeType.MAJOR]
         if other not in order:
@@ -239,7 +233,7 @@ class ChangeType(enum.Enum):
     def diff(cls, *, low: Version, high: Version) -> Optional["ChangeType"]:
         """
         Returns the order of magnitude of the change type diff between two versions.
-        Return None if the versions are less than a patch separated from each other.
+        Return None if the versions are less than a patch / a revision (if 4 digit version number) separated from each other.
         For example, a dev release and a post release for the same version number are less
         than a patch separated from each other.
         """
@@ -254,10 +248,10 @@ class ChangeType(enum.Enum):
         if high.micro > low.micro:
             return cls.PATCH
         if len(high.base_version.split('.')) >= 4:
-            high_revision = high.base_version.split('.')[-1]
+            high_revision = high.base_version.split('.')[3]
             # We are switching from 3 digits to 4
             if len(low.base_version.split('.')) < 4 or \
-                (len(low.base_version.split('.')) >= 4 and high_revision > low.base_version.split('.')[-1]):
+                (len(low.base_version.split('.')) >= 4 and high_revision > low.base_version.split('.')[3]):
                 return cls.REVISION
         raise Exception("Couldn't determine version change type diff: this state should be unreachable")
 
@@ -269,7 +263,7 @@ class ChangeType(enum.Enum):
         than one boolean argument is set to True, a ValueError is raised.
         """
         if sum([revision, patch, minor, major]) > 1:
-            raise ValueError("Only one argument of patch, minor or major can be set to True at the same time.")
+            raise ValueError("Only one argument of revision, patch, minor or major can be set to True at the same time.")
         if revision:
             return ChangeType.REVISION
         if patch:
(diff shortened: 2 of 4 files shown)


src/inmanta/moduletool.py
```diff
@@ -251,17 +253,25 @@ def diff(cls, *, low: Version, high: Version) -> Optional["ChangeType"]:
             return cls.MINOR
         if high.micro > low.micro:
             return cls.PATCH
+        if len(high.base_version.split('.')) >= 4:
+            high_revision = high.base_version.split('.')[-1]
+            # We are switching from 3 digits to 4
+            if len(low.base_version.split('.')) < 4 or \
+                (len(low.base_version.split('.')) >= 4 and high_revision > low.base_version.split('.')[-1]):
```
sanderr, line 260: This part seems redundant to me
```suggestion
            if len(low.base_version.split('.')) < 4 or high_revision > low.base_version.split('.')[-1]:
```

src/inmanta/moduletool.py
```diff
@@ -251,17 +253,25 @@ def diff(cls, *, low: Version, high: Version) -> Optional["ChangeType"]:
...
+                return cls.REVISION
```
sanderr, line 261: The docstring is no longer accurate: it says "Return None if the versions are less than a patch separated ...".

src/inmanta/moduletool.py
```diff
@@ -290,9 +301,20 @@ def bump_version(cls, change_type: ChangeType, version: Version, version_tag: st
             parts[0] += 1
             parts[1] = 0
             parts[2] = 0
+
+        if len(parts) >= 4:
+            size_version = 4
+
+            if change_type is ChangeType.REVISION:
```
sanderr, line 308: This doesn't allow us to do a revision on a 3-digit version, does it? I don't see that case in your tests either. I think we  do need to support it. It might be easier to make `parts` be size 4 always, and at the end of the method drop the fourth digit if it's a 0?

Open question: if you bump `1.2.3.4` with a patch, do we want to go to `1.2.4.0` or `1.2.4`? We should also add a test case for this.

src/inmanta/moduletool.py
```diff
@@ -683,10 +705,10 @@ def modules_parser_config(cls, parser: ArgumentParser) -> None:
   version.
 When a development release is done using the --dev option, this command:
 * Does a commit that updates the current version of the module to a development version that is a patch, minor or major version
-  ahead of the previous stable release. The size of the increment is determined by the --patch, --minor or --major argument
-  (--patch is the default). When a CHANGELOG.md file is present in the root of the module directory then the version number in
-  the changelog is also updated accordingly. The changelog file is always populated with the associated stable version and
-  not a development version.
+  ahead of the previous stable release. The size of the increment is determined by the --revision (only with 4 digits version),
```
sanderr, line 708: I would allow this for 3-digit versions as well. I see no reason not to, do you?

sanderr (changes requested):
Looks mostly good to me. I'm adding Arnaud as a second reviewer because he knows this code more intimately than I do.

tests/moduletool/test_release.py
```diff
@@ -414,6 +414,56 @@ def test_bump_dev_version_distance_already_met(
...
+    )
+    gitprovider.git_init(repo=path_module)
+    gitprovider.commit(repo=path_module, message="Initial commit", add=["*"], commit_all=True)
+    gitprovider.tag(repo=path_module, tag=initial_version)
+
+    monkeypatch.chdir(path_module)
+    module_tool = ModuleTool()
+    module_tool.release(dev=True, revision=True, message="Commit changes")
+    assert str(Module.from_path(path_module).version) == str(Version(after_revision_increment))
+    assert str(Module.from_path(path_module).version) == str(Version(after_revision_increment))
```
arnaudsjs, line 455: Why twice the same assertion? The same for the other assertions.

changelogs/unreleased/add-four-digit-versioning-module-tool.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: Add four digit versioning for module tool (release).
```
arnaudsjs, line 2: I think we want to add an entry to the `minor-improvement` section of the changelog.

arnaudsjs (changes requested):

src/inmanta/moduletool.py
```diff
@@ -159,7 +158,7 @@ def get_project(self, load: bool = False, strict_deps_check: Optional[bool] = No
             project.load()
         return project
 
-    def determine_new_version(self, old_version, version, major, minor, patch, dev):
+    def determine_new_version(self, old_version, version, major, minor, patch, revision, dev):
```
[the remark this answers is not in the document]
hlloreda: Yes, we don't want this change for this method. I made them because this method also used the `parse_from_bools` method which required the revision boolean.

tests/moduletool/test_release.py
hlloreda: That doesn't seem right, indeed. I started from the `test_bump_dev_version_distance_already_met` test, I didn't know if there was a reason to have two assertions there? But I should have asked earlier.

src/inmanta/moduletool.py
```diff
@@ -103,17 +103,21 @@ def add_deps_check_arguments(parser: argparse.ArgumentParser) -> None:
...
     )
     parser.add_argument(
         "--strict-deps-check",
         dest="strict_deps_check",
         action="store_true",
         default=False,
-        help="When this option is enabled, a version conflict in any (transitive) dependency will results in an error. "
+        help=(
+            "When this option is enabled, a version conflict in any (transitive) dependency will results in an error. ",
              "This option is mutually exclusive with the --no-strict-deps-check option.",
```
sanderr:
```suggestion
            "This option is mutually exclusive with the --no-strict-deps-check option.",
```

src/inmanta/moduletool.py
```diff
@@ -254,10 +248,10 @@ def diff(cls, *, low: Version, high: Version) -> Optional["ChangeType"]:
         if high.micro > low.micro:
             return cls.PATCH
         if len(high.base_version.split('.')) >= 4:
-            high_revision = high.base_version.split('.')[-1]
+            high_revision = high.base_version.split('.')[3]
             # We are switching from 3 digits to 4
             if len(low.base_version.split('.')) < 4 or \
-                (len(low.base_version.split('.')) >= 4 and high_revision > low.base_version.split('.')[-1]):
+                (len(low.base_version.split('.')) >= 4 and high_revision > low.base_version.split('.')[3]):
```
sanderr: This can still be simplified by dropping the redundant `len(low.base_version.split('.')) >= 4`.

src/inmanta/moduletool.py
```diff
@@ -289,10 +283,13 @@ def bump_version(cls, change_type: ChangeType, version: Version, version_tag: st
...
         """
         parts = [int(x) for x in version.base_version.split(".")]
-        while len(parts) < 3:
+        while len(parts) < 4:
             parts.append(0)
 
+        if change_type is ChangeType.REVISION:
+            parts[3] += 1
         if change_type is ChangeType.PATCH:
+            parts[3] = 0
```
sanderr, line 292: We need this for minor and major as well, don't we?

src/inmanta/moduletool.py
```diff
@@ -289,10 +283,13 @@ def bump_version(cls, change_type: ChangeType, version: Version, version_tag: st
         If the given version has a different version tag set, it will be ignored.
         """
         parts = [int(x) for x in version.base_version.split(".")]
-        while len(parts) < 3:
+        while len(parts) < 4:
             parts.append(0)
 
+        if change_type is ChangeType.REVISION:
```
sanderr, line 289: This comment got a bit out of hand. A summary of what follows below:
- If I understand correctly this currently always results in a four digit version (or I'm missing something). I don't think we want that. This is the only actual hard change request in this comment. Possible fix included in the snippet below but feel free to approach it differently.
- I don't like the big if-else with assignment to each part individually and I realized how we could clean it up. You didn't introduce it, but by adding another branch it got just a bit worse. Not a strict requirement for me but I think we can clean it up without too much effort. PoC below. Feel free to ignore this part if you want.
- I think we should raise an exception if `ChangeType` is none of the expected. Again, not something you introduced, but easy enough to fix it while we're at it. "assert your invariants, even if (or particularly when) they are trivial." - @wouterdb 2022 - https://inmanta.com/2022/02/15/the-test-case-i-wish-i-had-written/

PoC, feel free to rewrite this whole thing if you want.
```python
bump_index: int
if change_type is ChangeType.REVISION:
    bump_index = 3
elif change_type is ChangeType.PATCH:
    bump_index = 2
...
else:
    raise AppropriateException()

base_parts = [int(x) for x in version.base_version.split(".")]
nb_digits: int = max(bump_index + 1, 3, min(len(base_parts), 4)) # use 4th digit only if it already existed or if it is being bumped
parts = list(
    more_itertools.take(
        nb_digits,
        itertools.chain(
            base_parts[:bump_index + 1],
            itertools.repeat([0]),
        ),
    )
)
parts[bump_index] += 1
```

sanderr (changes requested):

src/inmanta/moduletool.py
hlloreda: I have reunified the string. Otherwise, the test_feature_flags test was failing because string methods were applied to a tuple.

src/inmanta/moduletool.py
hlloreda: The result was always a four-digit version, but if it ended with a 0, I was dropping that number to come back to a three-digit version. But your solution is much nicer than the old one, so I have replaced it with your PoC.

src/inmanta/moduletool.py
sanderr: Ah, I missed the comma. I don't think the way you made it now is good for this use case. It is a multi-line string, meaning it includes all newlines and indentation. We don't want that. We want it to be a single, one-line string, but we don't necessarily want to write it in our source as such. That's where the string concatenation comes in. The only issue with the above is the comma's at the end of the line:

```python
>>> help = (
...     "hello",
...     " world",
... )
>>> help
('hello', ' world')
>>> help = (
...     "hello"
...     " world"
... )
>>> help
'hello world'
>>> 
```

So there should only be a comma after the closing bracket.

src/inmanta/moduletool.py
sanderr: Ah, I must've missed the dropping of the fourth digit somehow. Anyway, I guess it's fine now.

sanderr: I see you made a small change to my snippet that is simpler but only valid if the fourth digit does indeed get dropped. Could you point me to where that happens?

sanderr: (The tests are passing so you must be right but I don't see it)

sanderr: And if it does get dropped we might as well simplify it even more by always taking 4 digits.

src/inmanta/moduletool.py
```diff
@@ -251,17 +248,24 @@ def diff(cls, *, low: Version, high: Version) -> Optional["ChangeType"]:
             return cls.MINOR
         if high.micro > low.micro:
             return cls.PATCH
+        if len(high.base_version.split(".")) >= 4:
+            high_revision = high.base_version.split(".")[3]
+            # We are switching from 3 digits to 4
+            if len(low.base_version.split(".")) < 4 or high_revision > low.base_version.split(".")[3]:
```
arnaudsjs, line 254: This doesn't look correct. You are doing a string comparison, but you actually want to do compare numbers.

arnaudsjs (changes requested):

arnaudsjs (approved):

src/inmanta/moduletool.py
```diff
@@ -308,6 +309,9 @@ def bump_version(cls, change_type: ChangeType, version: Version, version_tag: st
         )
         parts[bump_index] += 1
 
+        while len(parts) > 3 and parts[-1] == 0:
+            parts.pop()
```
sanderr, line 313: This removes the fourth digit always if it's 0, in contrast to what I put in my snippet. I don't have a strong opinion on this, just a slight preference for keeping four digits if there already were 4 or more. The main thing is that if we take this approach, the comment up above is no longer accurate: "use 4th digit only if it already existed or ...".

So either we need to update the comment or we need to change this approach to do as the comment states.

sanderr (approved):
Left one last tiny comment, otherwise looks good to me now.

PR #5697, opened 2023-03-16
Issue/933 show docs hover

FloLey:
# Description

This PR adds the docstrings to the anchormap so that the vscode extension can display them on hover.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/ast/__init__.py
+++ b/src/inmanta/ast/__init__.py
@@ -29,7 +29,7 @@ try:
     from typing import TYPE_CHECKING
 except ImportError:
     TYPE_CHECKING = False
-
+typeof = type
 if TYPE_CHECKING:
     from inmanta.ast.attribute import Attribute  # noqa: F401
     from inmanta.ast.entity import Entity
@@ -143,6 +143,22 @@ class Range(Location):
         return False
 
 
+class AncherTarget(Range):
+    __slots__ = "docstring"
+
+    def __init__(self, file: str, start_lnr: int, start_char: int, end_lnr: int, end_char: int, docstring: str) -> None:
+        """
+        Create a new Range instance.
+        :param file: the file this Range is in
+        :param start_lnr: the line number this Range starts on, 1-based
+        :param start_char: the start character number of the Range, 1-based
+        :param end_lnr: the line number this Range ends on, 1-based
+        :param end_char: the end character number of the Range, exclusive, 1-based
+        """
+        Range.__init__(self, file, start_lnr, start_char, end_lnr, end_char)
+        self.docstring = docstring
+
+
 class Locatable(object):
     __slots__ = ("_location",)
 
@@ -225,7 +241,14 @@ class TypeReferenceAnchor(Anchor):
 
     def resolve(self) -> Location:
         t = self.namespace.get_type(self.type)
-        return t.get_location()
+        location = t.get_location()
+        print("=TypeReferenceAnchor=")
+        print(typeof(location))
+        return location
+        # if location:
+        #     return AncherTarget(location.file, location.lnr, location.start_char, location.end_lnr, location.end_char, "test1")
+        # else:
+        #     return location
 
 
 class AttributeReferenceAnchor(Anchor):
@@ -241,7 +264,11 @@ class AttributeReferenceAnchor(Anchor):
         # assert isinstance(instancetype, Entity)
         entity_attribute: Optional[Attribute] = instancetype.get_attribute(self.attribute)
         assert entity_attribute is not None
-        return entity_attribute.get_location()
+        location = entity_attribute.get_location()
+        print("=AttributeReferenceAnchor=")
+        print(typeof(location))
+        return location
+        # return AncherTarget(location.file, location.start_lnr, location.start_char, location.end_lnr, location.end_char, "test1")
 
 
 class Namespaced(Locatable):
--- a/src/inmanta/ast/__init__.py
+++ b/src/inmanta/ast/__init__.py
@@ -29,7 +29,6 @@ try:
     from typing import TYPE_CHECKING
 except ImportError:
     TYPE_CHECKING = False
-typeof = type
 if TYPE_CHECKING:
     from inmanta.ast.attribute import Attribute  # noqa: F401
     from inmanta.ast.entity import Entity
--- a/src/inmanta/ast/__init__.py
+++ b/src/inmanta/ast/__init__.py
@@ -142,7 +142,9 @@ class Range(Location):
         return False
 
 
-class AncherTarget(Location):
+class AnchorTarget(Location):
+    __slots__ = ("start_char", "end_lnr", "end_char", "docstring")
+
     def __init__(
         self,
         file: str,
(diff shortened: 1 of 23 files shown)


FloLey:
image
 Those are the typing errors That remains: the first two ones is broken Typing that has been moved to another location. Fixing it would probably be a lot of work.
 
 The 2 lasts I didn't know hot to fix the `self.__class__.__function__` is already used multiple times in the function before I try to use it so I think (except for the typing) this is ok?

src/inmanta/ast/__init__.py
```diff
@@ -143,6 +143,38 @@ def __eq__(self, other: object) -> bool:
         return False
 
 
+class AnchorTarget(Location):
+    __slots__ = ("start_char", "end_lnr", "end_char", "docstring")
```
sanderr, line 147: Why not `AnchorTarget(Range)`?

src/inmanta/ast/__init__.py
```diff
@@ -143,6 +143,38 @@ def __eq__(self, other: object) -> bool:
...
+        self,
+        file: str,
+        start_lnr: int,
+        start_char: Optional[int],
+        end_lnr: Optional[int],
+        end_char: Optional[int],
+        docstring: Optional[str],
+    ) -> None:
+        """
+        Create a new AncherTarget instance.
```
sanderr, line 159: I think there has to be an empty line here. Also a typo.
```suggestion
        Create a new AnchorTarget instance.

```

src/inmanta/ast/__init__.py
```diff
@@ -143,6 +143,38 @@ def __eq__(self, other: object) -> bool:
...
+        """
+        Location.__init__(self, file, start_lnr)
+        self.start_char = start_char
+        self.end_lnr = end_lnr
+        self.end_char = end_char
+        self.docstring = docstring
+
+
+class WithComment(object):
+    comment: str = ""
```
sanderr, line 175: Why not `Optional[str] = None`? I feel like this would better convey that there may not be a comment attached.

sanderr (changes requested):

src/inmanta/ast/__init__.py
FloLey: because we don't always have a start_char, end_lnr and end_char which are not optional for Ranges

src/inmanta/ast/__init__.py
FloLey: For me it means the same `""` also means there is nothing attached as it impossible to have otherwise. and I don't like having Optionals everywhere in the code afterwards. But If you think it is wrong or not the "python" way of doing it I will change it :p

FloLey: It was just my first way of doing it, I'm not particularly attached to it. I'll see if I can change it

src/inmanta/ast/__init__.py
FloLey: Actually I think I'm already breaking compatibilty with previous versions of the extension. Some of it's behaviors depend on `if isinstance(t, Location)` or if `isinstance(t, Range)` that I had to change to `if isinstance(t, AnchorTarget)`. I noticed  this morning (it was breaking the code navigation), but I only realize now that this might be bad. I don't think there is another way around this other than to add the docstring on the location/range object which I don't like at all. 
So I think I could have a location field as i'm is already not backward compatible anymore :/

FloLey: do you mean where the anchormap is used?

sanderr: Yes. (When I asked the question I thought it was broader than that but I had another look at the diff and that's what it comes down to.)

FloLey: I think the anchormap is only used to created what is needed for the LS so I shouldn't break anything else.

src/inmanta/ast/__init__.py
```diff
@@ -223,9 +255,20 @@ def __init__(self, namespace: "Namespace", type: LocatableString) -> None:
         self.namespace = namespace
         self.type = type
 
-    def resolve(self) -> Location:
+    def resolve(self) -> Optional[AnchorTarget]:
         t = self.namespace.get_type(self.type)
-        return t.get_location()
+        location = t.get_location()
```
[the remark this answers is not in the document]
FloLey: @wouterdb I changed the AnchorTarget class after the review of @sanderr so I don't think this comment is relevent anymore.
Also my current solution is not backward compatible with older version of our vscode extension. For example it breaks here:
https://github.com/inmanta/vscode-inmanta/blob/master/server/src/inmantals/server.py#L228
as the anchormap now contains `AnchorTargets`. To be backward compatible I would need to add the comment on the `Location/Range` which I dislike.

src/inmanta/ast/__init__.py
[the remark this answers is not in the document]
wouterdb: This is the typing half of the compiler, so less critical, because it has way less instances.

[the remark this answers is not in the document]
sanderr: Great, I thought as much.

src/inmanta/ast/__init__.py
```diff
@@ -65,6 +65,9 @@ def merge(self, other: "Location") -> "Location":
 
         return Location(self.file, min(self.lnr, other.lnr))
 
+    def as_anchor_target(self, docstring: Optional[str]) -> "AnchorTarget":
+        return AnchorTarget(location=self, docstring=docstring)
```
sanderr, line 69: I think we have two options here:
- `AnchorTarget(Location)` and we use this method to create it
- `AnchorTarget` has a location field, no need for this method.

src/inmanta/ast/statements/define.py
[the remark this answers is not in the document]
sanderr: I'm still wondering about this.

src/inmanta/ast/statements/define.py
```diff
@@ -141,8 +140,6 @@ def __init__(
         self.attributes = attributes
         if comment is not None:
             self.comment = str(comment)
-        else:
-            self.comment = ""
```
sanderr: I think we should still set it. If I understand correctly you're currently counting on the default, but that's a class variable rather than an instance variable (for `pydantic` and `dataclasses` we do it like that because they do the magic for us). I think it should work like this (but I'm not even sure) but it is definitely good practice to set it on the instance.

Same for the classes below.

sanderr: Ah no, we don't inherit from `WithComment` so there isn't even the class var.

sanderr (changes requested):

sanderr (approved):

src/inmanta/ast/statements/define.py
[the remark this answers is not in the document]
wouterdb: it extends TypeDefinitionStatement so it does?

wouterdb (approved):

src/inmanta/ast/statements/define.py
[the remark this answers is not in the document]
sanderr: Yes, Florent pointed this out to me as well. I forgot to resolve the thread.

sanderr:
I hadn't noticed iso5 in there. I would leave it out, unless one of you thinks we need it there?

PR #5723, opened 2023-03-22
Issue/5433 list comprehensions

### LINKED ISSUE #5433: list comprehension in the inmanta DSL
Just a suggestion based on my experience in the lsm exercise.

Consider the model below
```
entity A:
end

entity B:
    int id
end

implement A using std::none
implement B using std::none

A.nested [0:1] -- A
A.important_config [0:] -- B
```

Now suppose that an implementation instantiates a tree of `A`, with at some level in the tree, the important config, based on some external input (e.g. another relation, lsm input, ...). The best approach would be to use a for loop after construction of the tree:
```
config_values = std::sequence(10)

a = A(
    nested=A(
        nested=A(
            nested=A()
        )
    )
)
for value in config_values:
    a.nested.nested.nested.important_config += B(id=value)
end
```

A list comprehension could greatly improve this by allowing us to include the important config right in the constructor tree, as you would for any non-list value:
```
config_values = std::sequence(10)
a = A(
    nested=A(
        nested=A(
            nested=A(
                important_config=[B(id=value) for value in config_values]
            )
        )
    )
)
```

Could be expanded for dict comprehensions as well.

sanderr:
# Description

Adds list comprehensions to the DSL. I didn't add any documentation yet. I'm planning to start on it today but I don't want to delay the merging of the implementation on it. Docs ticket: #6030.

closes #5433 

# Self Check:

- [x] Changelog entry
- [x] issue is created for end-user documentation: #6030

# Original description from back when it was a draft

Still in a very rough state, first review request is about the general flow that starts in `ListComprehension.requires_emit`. You can consider all else out of scope for now. I mostly want to know if I'm missing anything obvious here because it got quite a bit more verbose than I initially anticipated, but I think it is the best we can do. I'm asking review now so that I don't put in too much irrelevant work if you have another approach in mind.


src/inmanta/ast/statements/generator.py
```diff
@@ -268,7 +267,151 @@ def nested_blocks(self) -> Iterator["BasicBlock"]:
...
+    def requires_emit(
+        self, resolver: Resolver, queue: QueueScheduler, *, lhs: Optional[ResultCollector[object]] = None
+    ) -> dict[object, VariableABC]:
+        """
+        Sets up gradual execution of the list comprehension. Additionally sets up a resumer for when the iterable is complete.
+        Returns as requires the gradual helper's result variable, which will be frozen by the resumer.
+
+        Flow: requires_emit -> gradual execution -> resume -> execute
+        """
+        base_requires: dict[object, VariableABC] = super().requires_emit(resolver, queue)
```
wouterdb, line 310: this does nothing?

src/inmanta/ast/statements/generator.py
```diff
@@ -268,7 +267,151 @@ def nested_blocks(self) -> Iterator["BasicBlock"]:
...
+
+    def normalize(self, *, lhs_attribute: Optional[AttributeAssignmentLHS] = None) -> None:
+        self.value_expression.normalize(lhs_attribute=lhs_attribute)
+        self.iterable.normalize()
+
+    def requires(self) -> list[str]:
+        # exclude loop var, unless it shadows an occurrence in iterable
+        return list(set(self.iterable.requires()) - {self.loop_var} + set(self.value_expression.requires()))
+
+    def requires_emit(
```
wouterdb, line 301: it feels a bit unnatural for this implementation to be in `requires_emit` and not in `requires_emit_gradual`

src/inmanta/ast/statements/generator.py
```diff
@@ -268,7 +267,151 @@ def nested_blocks(self) -> Iterator["BasicBlock"]:
...
+        value_wrapper.set_value(value, location)
+        value_resolver: VariableResolver = VariableResolver(
+            parent=self.resolver,
+            name=str(self.statement.loop_var),
+            variable=value_wrapper,
+        )
+
+        # execute the value expression
+        requires: dict[object, VariableABC] = (
+            self.statement.value_expression.requires_emit(value_resolver, self.queue)
```
wouterdb, line 406: I think it is safe to just call requires_emit_gradual with a None value a last argument?

src/inmanta/execute/runtime.py
```diff
@@ -308,6 +309,49 @@ def get_dataflow_node(self) -> dataflow.AssignableNodeReference:
         return self._node
 
 
+# TODO: review implementation
+# TODO: double check typing
+class FixedCountVariable(ResultVariable[Union[T, list[T]]]):
```
wouterdb, line 314: This is unique in that it doesn't behave as a set, but as a list?

src/inmanta/ast/statements/generator.py
```diff
@@ -268,7 +267,151 @@ def nested_blocks(self) -> Iterator["BasicBlock"]:
...
+        )
+        iterable_requires: dict[object, VariableABC] = self.iterable.requires_emit_gradual(
+            resolver, queue, resultcollector=gradual_helper
+        )
+
+        # non-gradual mode / finishing up
+        # pass helper to the resumer via the requires object
+        wrapped_helper: VariableABC = WrappedValueVariable(gradual_helper)
+        requires: dict[object, VariableABC] = base_requires | iterable_requires | {self: wrapped_helper}
+        RawUnit(queue, resolver, requires, resumer=self)
```
wouterdb, line 327: Would it not be easier to make the `ListComprehensionGradual`  the resumer as well to get rid of the WrappedValueVariable?

wouterdb (commented):
looks fine, but it's been a while since I did this myself

src/inmanta/ast/statements/generator.py
```diff
@@ -268,7 +267,159 @@ def nested_blocks(self) -> Iterator["BasicBlock"]:
...
+
+        iterable: object = self.iterable.execute({k: v.get_value() for k, v in requires.items()}, resolver, queue)
+        if isinstance(iterable, Unknown):
+            return Unknown(self)
+        if not isinstance(iterable, list):
+            # TODO: allow dict?
+            raise TypingException(
+                self, f"A list comprehension can only be applied to lists and relations, got {type(iterable)}"
+            )
+        if gradual_helper.nb_results_received == 0:
```
wouterdb, line 355: I don't know if a mixed mode can occur, but the fact that we don't de-duplicate makes that rather important.

src/inmanta/ast/statements/generator.py
sanderr: It might: the super class emits eager promises. In this case I'm not sure if there are any but I think it's good practice to call it anyway.

src/inmanta/execute/runtime.py
wouterdb: as it is now, it also breaks propagation of gradual execution

src/inmanta/ast/statements/generator.py
sanderr: I wasn't sure myself which direction felt more natural. This one is an extension to `requires_emit` to optionally make it gradual, the other direction is an extension to `requires_emit_gradual` to optionally make it not gradual. Both felt unnatural to me with this one the lesser of two evils because `requires_emit_gradual` explicitly has `gradual` in the name so dropping it felt even more weird. But I'm willing to swap it if you prefer. Or if you see a third approach that does feel natural I'm all for that.

src/inmanta/ast/statements/generator.py
sanderr: I rejected that for some reason. I'll look it over once more and get back to you.

src/inmanta/ast/statements/generator.py
sanderr: Note to self: double check that all statements that implement `requires_emit_gradual` have a consistent `execute` and document that as a requirement.

wouterdb: I think we concluded that we can't currently think of a mixed scenario, so it doesn't make it worse.

src/inmanta/execute/runtime.py
sanderr: I dropped this, replacing it with the `_results` attribute in the helper.

src/inmanta/ast/statements/generator.py
sanderr: It is already a resumer of itself now so I think the current implementation is the best we can get.

src/inmanta/ast/statements/__init__.py
```diff
@@ -234,7 +252,7 @@ def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.NodeReference:
         raise NotImplementedError()
 
 
-class Resumer(ExpressionStatement):
+class Resumer(Locatable):
```
sanderr, line 255: I didn't get any type errors when I did this and I see no reason why it should be an `ExpressionStatement`. If you think differently, let me know.

src/inmanta/ast/statements/generator.py
```diff
@@ -360,26 +664,33 @@ def get_all_eager_promises(self) -> Iterator["StaticEagerPromise"]:
...
         RawUnit(queue, resolver, self.condition.requires_emit(resolver, queue), resumer)
 
         # Wait for the result variable to be populated
         requires[self] = result
         return requires
 
+    def requires_emit_gradual(
+        self, resolver: Resolver, queue: QueueScheduler, resultcollector: ResultCollector[object]
+    ) -> dict[object, VariableABC[object]]:
+        return self.requires_emit(resolver, queue, lhs=resultcollector)
```
sanderr, line 687: I made the conditional expression gradually executable while I was at it.

wouterdb: what does that even mean? conditional expressions produce only one value, so why is it a benefit to be gradual?

src/inmanta/ast/statements/generator.py
```diff
@@ -427,10 +741,15 @@ def resume(self, requires: Dict[object, VariableABC], resolver: Resolver, queue:
             subexpression: ExpressionStatement = (
                 self.expression.if_expression if self.condition_value else self.expression.else_expression
             )
+            subexpression_requires: dict[object, VariableABC[object]] = (
+                subexpression.requires_emit(resolver, queue)
+                if self.lhs is None
+                else subexpression.requires_emit_gradual(resolver, queue, self.lhs)
```
wouterdb, line 747: Why have this else everywhere instead of making requires_emit_gradual shortcut to requires_emit when the lhs parameter is None?

tests/compiler/test_list_comprehension.py
```diff
@@ -0,0 +1,677 @@
...
+            end
+            Branch.twigs [0:] -- Twig.branch [1]
+            index Twig(branch, length)
+
+            implement Trunk using std::none
+            implement Branch using std::none
+            implement Twig using std::none
+
+            # e.g. from lsm::all()
+            raw_tree = {
```
wouterdb, line 217: I'm here

wouterdb (commented):
tbc

tests/compiler/test_list_comprehension.py
```diff
@@ -0,0 +1,677 @@
...
+                        "twigs": [4, 5, 6],
+                    },
+                    {
+                        "length": 30,
+                        "twigs": [7, 8, 9],
+                    },
+                ]
+            }
+
+            tree = Trunk(
```
wouterdb, line 235: so nice!

wouterdb (approved):
1. super nice
2. testing and implementation look very complete and thorough
3. I've not internalized all the details sufficiently to really get a feeling for all the details, but given the level of testing and your experience with the compiler, this looks great.

src/inmanta/ast/statements/generator.py
sanderr: They resolve to one expression, which may produce more than one value. For example, the conditional expression `x.myrelation = condition ? y.a : y.b` obviously has to wait for the condition to fully execute (we can't be gradual there), but once it has picked a branch it can pass on the result collector to the relevant child expression. With the implementation on master this expression would block the gradual execution of `y.a` or `y.b`. The test case `test_conditional_expression_gradual` I added illustrates this: it fails on master but succeeds on this branch.

src/inmanta/ast/statements/generator.py
```diff
@@ -268,7 +267,312 @@ def nested_blocks(self) -> Iterator["BasicBlock"]:
...
+        self.anchors.extend(
+            itertools.chain(
+                self.value_expression.get_anchors(),
+                self.iterable.get_anchors(),
+                (self.guard.get_anchors() if self.guard is not None else ()),
+            )
+        )
+
+    def normalize(self, *, lhs_attribute: Optional[AttributeAssignmentLHS] = None) -> None:
+        self.value_expression.normalize(lhs_attribute=lhs_attribute)
```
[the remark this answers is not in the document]
sanderr: Indeed

PR #5724, opened 2023-03-22
Fix bug in calculation fact renewal timestamp

arnaudsjs:
# Description

Fix bug that incorrectly calculates the timestamp indicating which facts have to be renewed.

# Self Check:

- [x] Changelog entry


sanderr (approved):
The method name still has `expired`, fyi

FloLey (approved):

PR #5728, opened 2023-03-23
Disable cleanup jobs that removes old model version and agent records for tests

arnaudsjs:
# Description

Disable the cleanup job, that removes old model versions and agent records, when the test suite is ran. If this job is enabled, it causes race conditions in test cases that manually create agent records without an associated model.

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (approved):

arnaudsjs:
> Do the tests that require it already enable it explicitly?

As far as I can see there are no tests that need the scheduled cleanup. The tests invoke the cleanup method directly.

PR #5735, opened 2023-03-24
Fix backward compatibility LS

FloLey:
# Description

Change core to have 2 functions(scheduler_instance.anchormap (old) and scheduler_instance.anchormap_extended (new)) that can be called by the LS to ensure it stays backward compatible.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/execute/scheduler.py
+++ b/src/inmanta/execute/scheduler.py
@@ -290,7 +290,10 @@ class Scheduler(object):
 
     def anchormap(
         self, compiler: "Compiler", statements: Sequence["Statement"], blocks: Sequence["BasicBlock"]
-    ) -> Sequence[Tuple[Location, AnchorTarget]]:
+    ) -> Sequence[Tuple[Location, Location]]:
+        """
+        This methode exists for backward compatibility with inmantals
+        """
         range_to_anchor_target = self.anchormap_extended(compiler, statements, blocks)
         range_to_range = [(f, t.location) for f, t in range_to_anchor_target]
 
--- a/src/inmanta/execute/scheduler.py
+++ b/src/inmanta/execute/scheduler.py
@@ -262,7 +262,7 @@ class Scheduler(object):
 
         self.types = {k: v for k, v in types_and_impl.items() if isinstance(v, Type)}
 
-    def anchormap_extended(
+    def get_anchormap(
         self, compiler: "Compiler", statements: Sequence["Statement"], blocks: Sequence["BasicBlock"]
     ) -> Sequence[Tuple[Location, AnchorTarget]]:
         prev = time.time()
@@ -294,7 +294,7 @@ class Scheduler(object):
         """
         This methode exists for backward compatibility with inmantals
         """
-        range_to_anchor_target = self.anchormap_extended(compiler, statements, blocks)
+        range_to_anchor_target = self.get_anchormap(compiler, statements, blocks)
         range_to_range = [(f, t.location) for f, t in range_to_anchor_target]
 
         return range_to_range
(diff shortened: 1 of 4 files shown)


src/inmanta/execute/scheduler.py
```diff
@@ -280,13 +280,21 @@ def anchormap(
...
+        range_to_anchor_target = [(anchor.get_location(), anchor.resolve()) for anchor in anchors]
+        range_to_anchor_target = [(f, t) for f, t in range_to_anchor_target if t is not None]
 
         now = time.time()
         LOGGER.debug("Anchormap took %f seconds", now - prev)
 
-        return rangetorange
+        return range_to_anchor_target
+
+    def anchormap(
```
arnaudsjs, line 291: We need a docstring here to indicate that this method exists for backwards compatibility with inmanta-vscode.

src/inmanta/execute/scheduler.py
```diff
@@ -280,13 +280,21 @@ def anchormap(
...
+        self, compiler: "Compiler", statements: Sequence["Statement"], blocks: Sequence["BasicBlock"]
+    ) -> Sequence[Tuple[Location, AnchorTarget]]:
```
arnaudsjs, line 293: This type annotation isn't correct. You probably forgot to run mypy on this?

arnaudsjs (changes requested):

tests/test_compiler_entrypoints.py
```diff
@@ -74,6 +76,63 @@ def verify_anchor(flnr, s, e, tolnr):
...
+        autostd=False,
+    )
+    compiler = Compiler()
+    (statements, blocks) = compiler.compile()
+    sched = scheduler.Scheduler()
+    anchormap = sched.anchormap(compiler, statements, blocks)
+
+    assert len(anchormap) == 9
+
+    checkmap = {(r.lnr, r.start_char, r.end_char): t.lnr for r, t in anchormap}
```
arnaudsjs, line 118: anchormap returns something of type `Sequence[Tuple[Location, Location]]`, but `Location` doesn't the fields `start_char` and `end_char`. Am I missing something here?

arnaudsjs (changes requested):

tests/test_compiler_entrypoints.py
FloLey: `Range` inherits from a `Location` and range has those fields.
The typing between `Location` and `Range` is a bit broken and not super nice but it would be to much of an effort to fix it now:

- the resolve function [here](https://github.com/inmanta/inmanta-core/blob/db32611489e71e1064d1c60df7b963200fd31114/src/inmanta/execute/scheduler.py#L283) use to return a Range or a Location before I made any change. Now it returns an AnchorTarget where its location field is a `Location` or a `Range`.

arnaudsjs: Ok. So in this specific case the map happens to only contain elements of type `Range`. In that case we need to add an assert statement that verifies this to make clear to the reader of the remainder of the code assumes that all the elements for the `Range` type.

arnaudsjs (approved):

src/inmanta/execute/scheduler.py
```diff
@@ -262,7 +262,7 @@ def define_types(self, compiler: "Compiler", statements: Sequence["Statement"],
 
         self.types = {k: v for k, v in types_and_impl.items() if isinstance(v, Type)}
 
-    def anchormap(
+    def anchormap_extended(
```
sanderr, line 265: We could go with just `get_anchormap`? Because this name only makes sense with respect to the old one, it's not really standalone.

sanderr (approved):

PR #5736, opened 2023-03-24
Extend functionality openapi type converter

arnaudsjs:
# Description

This PR:

* Extend OpenAPI type converter with features required by lsm extension.
* Bugfix: Ensure OpenAPI enum values always have the type field set.
* Bugfix: Ensure that the OpenAPI type reference resolver populates the `allOf` and `oneOf` properties of the OpenAPI schema.
* Bugfix: Ensure that the OpenApiTypeConverter.get_openapi_type_of_parameter() method populates the default value of a parameter.

Part of inmanta/inmanta-lsm#1169

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/protocol/common.py
+++ b/src/inmanta/protocol/common.py
@@ -63,6 +63,7 @@ from inmanta import config as inmanta_config
 from inmanta import const, execute, types, util
 from inmanta.data.model import BaseModel, validator_timezone_aware_timestamps
 from inmanta.protocol.exceptions import BadRequest, BaseHttpException
+from inmanta.protocol.openapi import model as openapi_model
 from inmanta.stable_api import stable_api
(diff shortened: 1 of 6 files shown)


FloLey (approved):

sanderr (approved):
With the caveat that I hardly know this code I think this is good.

src/inmanta/protocol/common.py
```diff
@@ -878,6 +885,20 @@ def _encode_dict_for_get(
         sub_dict = {f"{query_param_name}.{key}": value for key, value in query_param_value.items()}
         return sub_dict
 
+    def get_openapi_parameter_type_for(self, param_name: str) -> Optional[openapi_model.ParameterType]:
```
sanderr, line 888: I just noticed you use this in your other PR. I think it should be marked with `@stable_api`.

arnaudsjs: I added them

PR #5741, opened 2023-03-27
Remove ensure internal agent code path from AutostartedAgentManager

arnaudsjs:
# Description

This PR removes the code from the AutostartedAgentManager that ensures that the internal agent is present in the `autostarted_agent_map`. The code is removed because it performs a read-modify-write operation in a non-atomic way, that causes a race condition when the tests perform a call to update the `autostart_agent_map` setting.

The code to ensure that internal agent is present in the `autostart_agent_map` was introduced in [this PR](https://github.com/inmanta/inmanta-core/pull/2103),  because in an old version of the server there wasn't a validator on the update path of the setting that verified this. A [validator](https://github.com/inmanta/inmanta-core/blob/db32611489e71e1064d1c60df7b963200fd31114/src/inmanta/data/__init__.py#L2280) that perform this check was introduced in Apr 21, 2020, so before the initial release of ISO4. As such, we can safely remove this code now.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/remove-ensure-internal-agent-code.yml
+++ b/changelogs/unreleased/remove-ensure-internal-agent-code.yml
@@ -1,4 +1,4 @@
 ---
 description: Remove the code in the AutostartedAgentManager that ensure that the internal agent is present in the autostarted_agent_map. This code that safely be removed because a validator was added on the update path of the `autostart_agent_map` in Apr 21, 2020. So this check was included in the initial release of ISO4.
 change-type: patch
-destination-branches: [master, iso6, iso5, iso4]
+destination-branches: [master, iso6, iso5]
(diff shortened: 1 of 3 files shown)


arnaudsjs:
I now realize that this cannot cause the race condition in the tests. Do we still remove this code?

changelogs/unreleased/remove-ensure-internal-agent-code.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: Remove the code in the AutostartedAgentManager that ensure that the internal agent is present in the autostarted_agent_map. This code that safely be removed because a validator was added on the update path of the `autostart_agent_map` in Apr 21, 2020. So this check was included in the initial release of ISO4.
+change-type: patch
+destination-branches: [master, iso6, iso5, iso4]
```
wouterdb, line 4: I would not add this to iso4

wouterdb (approved):
I would not in

sanderr (approved):

PR #5776, opened 2023-03-30
 Autostarted agents don't log in debug mode#5562

### LINKED ISSUE #5562: Autostarted agents don't log in debug mode
Autostarted agents [are started with the `-vvvv` option](https://github.com/inmanta/inmanta-core/blob/d9b6c75cf44694476d60d88e6a743906cb2cf445/src/inmanta/server/agentmanager.py#L1080), because we want it to produce logs with the log level DEBUG. However, as soon as the [`--log-file` option](https://github.com/inmanta/inmanta-core/blob/d9b6c75cf44694476d60d88e6a743906cb2cf445/src/inmanta/app.py#L820) is used, the desired log level should be indicated using the [log-file-level option](https://github.com/inmanta/inmanta-core/blob/d9b6c75cf44694476d60d88e6a743906cb2cf445/src/inmanta/app.py#L680) and not using the `-vvvv` option. The AutostartedAgentManager doesn't set this option. The results is that autostarted agents produce logs at the INFO level instead of the DEBUG level.

Tasks:

- [ ] Make sure that autostarted agents have the `log-file-level` option set to log at DEBUG level.
- [ ] Autostarted agents should not set the `-vvvv` option, because it's never used anyway.

FloLey:
# Description

Autostarted agents [are started with the -vvvv option](https://github.com/inmanta/inmanta-core/blob/d9b6c75cf44694476d60d88e6a743906cb2cf445/src/inmanta/server/agentmanager.py#L1080), because we want it to produce logs with the log level DEBUG. However, as soon as the [--log-file option](https://github.com/inmanta/inmanta-core/blob/d9b6c75cf44694476d60d88e6a743906cb2cf445/src/inmanta/app.py#L820) is used, the desired log level should be indicated using the [log-file-level option](https://github.com/inmanta/inmanta-core/blob/d9b6c75cf44694476d60d88e6a743906cb2cf445/src/inmanta/app.py#L680) and not using the -vvvv option. The AutostartedAgentManager doesn't set this option. The results is that autostarted agents produce logs at the INFO level instead of the DEBUG level.

Tasks:

- [x]  Make sure that autostarted agents have the log-file-level option set to log at DEBUG level.
- [x]   Autostarted agents should not set the -vvvv option, because it's never used anyway.
- [x]  accept DEBUG|WARNING|ERROR|... in addition to the number that is accepted now

closes #5562 

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (approved):

changelogs/unreleased/5562-autostarted-agents-log.yml
```diff
@@ -0,0 +1,6 @@
+description: Autostarted agents will now log in debug mode
+issue-nr: 5562
+change-type: patch
+destination-branches: [master, iso6, iso5]
```
arnaudsjs, line 4: I think this should go into ISO4 as well. Having DEBUG logs on the agent is quite important at debugging time.

src/inmanta/app.py
```diff
@@ -767,9 +776,7 @@ def _get_watched_file_handler(options: argparse.Namespace) -> logging.handlers.W
     return file_handler
 
 
-def _convert_to_log_level(level: int) -> int:
-    if level >= len(log_levels):
-        level = len(log_levels) - 1
```
arnaudsjs, line 772: This check made sure that we wouldn't get an exception when somebody passed in `-vvvvvvvv` on the commandline. I think we should maintain that behavior.

arnaudsjs (changes requested):

arnaudsjs:
Did you somehow test this change?

FloLey:
@arnaudsjs 
> * Test whether autostarted agents actually log in debug mode.  Autostarted agents [log a line in DEBUG level](https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/protocol/endpoints.py#L195) when they start, so it should to rather easy to write a test case for this.

This log line is only present with loglevel trace right?

arnaudsjs:
> @arnaudsjs
> 
> > * Test whether autostarted agents actually log in debug mode.  Autostarted agents [log a line in DEBUG level](https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/protocol/endpoints.py#L195) when they start, so it should to rather easy to write a test case for this.
> 
> This log line is only present with loglevel trace right?

True. Then you should pick another one. [This one](https://github.com/inmanta/inmanta-core/blob/2be65a80d8932a92411780e9c6b932580a45d55f/src/inmanta/agent/agent.py#L1106) is also rather easy to trigger for example. Or you can monkeypatch a method in the agent so that it writes a log line at DEBUG level.

src/inmanta/app.py
FloLey: I put it in the `_convert_cli_log_level` function. I think it is ok?

tests/test_agent_manager.py
```diff
@@ -1272,3 +1272,31 @@ async def test_dont_start_paused_agent(server, client, environment, caplog) -> N
...
+
+
+async def test_auto_started_agent_log_in_debug_mode(server, environment, agent_factory, caplog, monkeypatch):
+    """
+    Test the logging of an autostarted agent
+    """
+
+    async def mock_start_agent(self) -> None:
+        await super(Agent, self).start()
+        LOGGER.debug("test log in debug level")
```
arnaudsjs, line 1284: This removes the functionality present in the `Agent.start()` method. This may cause all sorts of instabilities on the test suite. We should make sure we main that behavior.

src/inmanta/app.py
arnaudsjs: That's not enough. This method is used by the `_get_watched_file_handler()` method. This new implementation crashes when a log level is passed in above the maximum log level while the previous implementation wouldn't. This might cause the orchestrator to crash after an upgrade. We should prevent that.

src/inmanta/app.py
```diff
@@ -767,9 +778,7 @@ def _get_watched_file_handler(options: argparse.Namespace) -> logging.handlers.W
     return file_handler
 
 
-def _convert_to_log_level(level: int) -> int:
```
arnaudsjs, line 770: For me, it took some time to interpret this method and the method below this method. This is not an issue introduced by your changes, but we should grasp the opportunity to improve it now that we are working on this part of the code base. It would be good to add a docstring to both methods, describing what these methods do. We should definitely mention that:

* `_convert_cli_log_level()`: Converts the number of `-v`'s passed on the CLI to the corresponding Inmanta log level.
* `_convert_to_log_level()`: This method converts the Inmanta log level to the Python log level. I think we could just as well rename this method to `convert_inmanta_log_level_to_python_log_level()`.

src/inmanta/app.py
```diff
@@ -662,7 +662,18 @@ def export(options: argparse.Namespace) -> None:
         conn.release_version(tid, version, True, agent_trigger_method)
 
 
-log_levels = {0: logging.ERROR, 1: logging.WARNING, 2: logging.INFO, 3: logging.DEBUG, 4: 2}
+log_levels = {
```
arnaudsjs, line 665: I would also add a docstring here saying that this dictionary maps the Inmanta log levels to the corresponding Python log levels.

tests/test_app.py
```diff
@@ -299,11 +314,53 @@ def test_log_file_set(tmpdir, log_level, with_tty, regexes_required_lines, regex
     log_file = os.path.join(log_dir, log_file)
     with open(log_file, "r") as f:
         log_lines = f.readlines()
+        print(log_lines)
```
arnaudsjs, line 317: `check_logs` already prints the log lines. Why do we print them twice?

tests/test_app.py
```diff
@@ -299,11 +314,53 @@ def test_log_file_set(tmpdir, log_level, with_tty, regexes_required_lines, regex
...
+        (
+            3,
+            [
+                r"[a-z.]*[ ]*INFO[\s]+[a-x\.A-Z]*[\s]Starting server endpoint",
+                r"[a-z.]*[ ]*DEBUG[\s]+[a-x\.A-Z]*[\s]Starting Server Rest Endpoint",
+            ],
+            [],
+        ),
+        (
+            2,
```
arnaudsjs, line 351: Why did you exclude the case where `log_level=1`? That's an interesting case, because it verifies whether the `INFO` log line is filtered out.

arnaudsjs (changes requested):

tests/test_agent_manager.py
```diff
@@ -1272,3 +1273,25 @@ async def test_dont_start_paused_agent(server, client, environment, caplog) -> N
...
+    await agentmanager.ensure_agent_registered(env, "test1")
+    await autostarted_agentmanager._ensure_agents(env, ["test1"])
+
+    logdir = Config.get("config", "log-dir")
+    log_file = f"{logdir}/agent-{environment}.log"  # Path to the log file
+
+    with open(log_file, mode="r") as f:
+        log_content = f.read()
+
+    assert "DEBUG    inmanta.protocol.endpoints Start transport for client agent" in log_content
```
arnaudsjs, line 1297: There is a race condition here. If the agent starts slowly, the log entry might not be there yet. We should use `retry_limited` here.

tests/test_agent_manager.py
```diff
@@ -1272,3 +1273,25 @@ async def test_dont_start_paused_agent(server, client, environment, caplog) -> N
     assert "Started new agent with PID" not in caplog.text
     # Ensure no timeout happened
     assert "took too long to start" not in caplog.text
+
+
+async def test_auto_started_agent_log_in_debug_mode(server, environment, monkeypatch):
```
arnaudsjs, line 1278:
```suggestion
async def test_auto_started_agent_log_in_debug_mode(server, environment):
```

src/inmanta/app.py
```diff
@@ -767,18 +781,24 @@ def _get_watched_file_handler(options: argparse.Namespace) -> logging.handlers.W
...
 
 
-def _convert_to_log_level(level: int) -> int:
-    if level >= len(log_levels):
-        level = len(log_levels) - 1
+def _convert_inmanta_log_level_to_python_log_level(level: str) -> int:
+    """
+    Converts the Inmanta log level to the Python log level
+    """
+    if level.isdigit() and int(level) >= 4:
```
arnaudsjs, line 788:
```suggestion
    if level.isdigit() and int(level) > 4:
```

arnaudsjs (changes requested):

tests/test_agent_manager.py
FloLey: indeed, I didn't think of that. I'll add it. Thanks!

arnaudsjs (approved):

PR #5779, opened 2023-03-30
Produce no warning about download path if it is not used

### LINKED ISSUE #5507: Downloadpath is mandatory even for only v2 modules
When you create a project with only v2 modules you still need to add downloadpath and then you also get a warning that this path needs to be in modulepath.

This check can be disabled if there is no v1 module source configured.

wouterdb:
# Description

Produce no warning about download path if it is not used

closes #5507 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/module.py
+++ b/src/inmanta/module.py
@@ -960,7 +960,7 @@ class ModuleRepo:
 
     def is_empty(self) -> bool:
         """
-        return true of this repo will never produce any repo
+        Return true if this repo will never produce any repo.
 
         Used to distinguish an empty compose repo from a non-empty one
         """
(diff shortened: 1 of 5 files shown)


FloLey (approved):

src/inmanta/module.py
```diff
@@ -956,6 +958,14 @@ def path_for(self, name: str) -> Optional[str]:
         # same class is used for search path and remote repos, perhaps not optimal
         raise NotImplementedError("Abstract method")
 
+    def is_empty(self) -> bool:
+        """
+        return true of this repo will never produce any repo
```
arnaudsjs, line 963:
```suggestion
        Return true if this repo will never produce any repo.
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #5783, opened 2023-03-31
Ensure proper agent shutdown in testcases

wouterdb:
# Description

Ensure proper agent shutdown in testcases

closes #5783

The essential issue here was that there is no way to asynchronously join a threadpool 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -704,7 +704,18 @@ class nullcontext(contextlib.nullcontext[T], contextlib.AbstractAsyncContextMana
 
 
 async def join_threadpools(threadpools: List[ThreadPoolExecutor]) -> None:
-    # idea borrowed from BaseEventLoop.shutdown_default_executor
+    """
+    Asynchronously join a set of threadpools
+
+    idea borrowed from BaseEventLoop.shutdown_default_executor
+
+    We implemented this method because:
+    1. ThreadPoolExecutor.shutdown(wait=True)` is a blocking call, blocking the ioloop.
+       This doesn't work because we often have back-and-forth between the ioloop and the thread
+       due to our `ResourceHandler.run_sync` method.
+    2.The python sdk has no support for async awaiting threadpool shutdown (except for the default pool)
+    """
+
     loop = asyncio.get_event_loop()
     future = loop.create_future()
 
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -33,8 +33,10 @@ import warnings
 from abc import ABC, abstractmethod
 from asyncio import CancelledError, Future, Lock, Task, ensure_future, gather
 from collections import abc, defaultdict
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


tests/test_util.py
```diff
@@ -494,3 +498,35 @@ def test_running_test_fixture():
...
+        block.result()
+        done.put("A")
+
+    for i in range(5):
+        tp.submit(worker)
+
+    tp.shutdown(wait=False)
+    assert done.qsize() == 0
+    hanglock.set()
+    assert done.qsize() == 0
```
Hugo-Inmanta, line 528: I think I don't understand how this assertion can always be true. I played around a bit with the testcase and set the sleep in `cor()` to 0 and the assertion is still always true.

Hugo-Inmanta (approved):

tests/test_util.py
wouterdb: When unblocking the hanglock, all thread get stuck on `block.result()` 
for this to unblock, the co-routine `cor` has to run. 

Which means that the ioloop must be able to execute things. 
But we are sitting on the main thread, so it can't execute until we do an `await` to return control to the ioloop....

src/inmanta/util.py
```diff
@@ -699,3 +701,24 @@ async def __aenter__(self) -> T:
 
     async def __aexit__(self, *excinfo: object) -> None:
         pass
+
+
+async def join_threadpools(threadpools: List[ThreadPoolExecutor]) -> None:
+    # idea borrowed from BaseEventLoop.shutdown_default_executor
```
sanderr, line 707: This comment is good but I think we also need a comment as to why we implement a custom method for this rather than to rely on plain `threadpool.shutdown`.

src/inmanta/util.py
```diff
@@ -699,3 +701,24 @@ async def __aenter__(self) -> T:
...
+
+    def join() -> None:
+        for threadpool in threadpools:
+            try:
+                threadpool.shutdown(wait=True)
+            except Exception:
+                LOGGER.exception("Exception during threadpool shutdown")
+        loop.call_soon_threadsafe(future.set_result, None)
+
+    thread = threading.Thread(target=join)
```
sanderr, line 719: Do I understand correctly that we start a new thread outside of the threadpool to wait (blocking) for shutdown of the threadloop, thereby allowing the event loop(s?) within the threadpool to continue? It is a lot nicer than what we came up with.

tests/test_util.py
Hugo-Inmanta: I see, thanks :+1:

sanderr (approved):

src/inmanta/util.py
wouterdb: You understand correctly. 

It was my alternative idea, only that my idea was to start a threadpool, which is a bit overkill.

sanderr: I think I completely misunderstood your idea

sanderr (approved):

PR #5784, opened 2023-03-31
Issue/5651 ignore plugins for modules that arent loaded

### LINKED ISSUE #5651: ignore plugins for modules that aren't loaded
To avoid errors like this:
```
  File "/home/user/ws/.venv39-connect/lib64/python3.9/site-packages/inmantals/server.py", line 171, in compile_and_anchor
    await asyncio.get_event_loop().run_in_executor(self.threadpool, sync_compile_and_anchor)
  File "/usr/lib64/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/user/ws/.venv39-connect/lib64/python3.9/site-packages/inmantals/server.py", line 140, in sync_compile_and_anchor
    (statements, blocks) = compiler_instance.compile()
  File "/home/user/ws/.venv39-connect/lib64/python3.9/site-packages/inmanta/compiler/__init__.py", line 215, in compile
    raise Exception("Unable to find namespace for plugin module %s" % (cls.__module__))
Exception: Unable to find namespace for plugin module inmanta_plugins.netbox
```

When the compiler encounters such a plugin in `PluginMeta` (see trace above) it should delete it from `PluginMeta`, as if it was never loaded in the first place.

May go in both iso6 and iso7.

Hugo-Inmanta:
# Description

closes #5651 

# Self Check:

- [x] Changelog entry


tests/data/modules_v2/cross_module_dependency/setup.cfg
arnaudsjs, line 1: Is there a specific reason why this module contains: files, templates and several sub-namespaces?

arnaudsjs (changes requested):

tests/data/modules_v2/cross_module_dependency/setup.cfg
Hugo-Inmanta: Cleaned up this module

changelogs/unreleased/5651-ignore-plugins-for-modules-that-arent-loaded.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: Ignore plugins for modules that aren't loaded.
+issue-nr: 5651
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master, iso6]
```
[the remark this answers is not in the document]
arnaudsjs: I also wasn't entirely sure. That's why I added Sander as a second reviewer. To me this issue is a blocker for using v2 modules to its full extend and since v2 modules are part of iso5, it should also be merged there. But we can just as well decided not to merge this into iso5 and say that users should upgrade to iso6 get access to this functionality.

src/inmanta/compiler/__init__.py
```diff
@@ -213,11 +220,16 @@ def compile(self) -> Tuple[List["Statement"], List["BasicBlock"]]:
                 ns = ns.get_child(part)
 
             if ns is None:
-                raise Exception("Unable to find namespace for plugin module %s" % (cls.__module__))
+                # Mark this plugin's module for unregistration so that future iterations on this module can be skipped.
+                marked_for_unregistration.add(cls.__module__)
+                # Un-register this plugin, as it is part of a module that is not loaded
```
arnaudsjs, line 225:
```suggestion
                # Un-register this plugin, as it is part of a module that is not loaded (i.e. not imported in the model)
```

src/inmanta/compiler/__init__.py
```diff
@@ -196,8 +196,15 @@ def compile(self) -> Tuple[List["Statement"], List["BasicBlock"]]:
 
         project.log_installed_modules()
 
+        # This lookup variable provides efficiency in the loop bellow:
+        # only call PluginMeta.clear once per module that is not loaded
```
arnaudsjs, line 200: This sentence is not really correct. I think you are mixing up the loading of the Python module with the loading of the Inmanta module. 
```suggestion
        # This lookup variable provides efficiency in the loop bellow:
        # Make sure that `PluginMeta.clear` is called only once for each loaded Python module that is part of a non-loaded (non-imported) Inmanta module.
```

src/inmanta/compiler/__init__.py
```diff
@@ -213,11 +213,14 @@ def compile(self) -> Tuple[List["Statement"], List["BasicBlock"]]:
                 ns = ns.get_child(part)
 
             if ns is None:
-                raise Exception("Unable to find namespace for plugin module %s" % (cls.__module__))
+                # Un-register this plugin, as it is part of a module that is not loaded
+                # (https://github.com/inmanta/inmanta-core/issues/5651)
+                PluginMeta.clear(".".join(mod_ns))
```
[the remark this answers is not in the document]
arnaudsjs: We could even further optimize this by defining a new `clear()` method that accepts a set of namespaces. Then we can do the clear after the for loop by doing a single pass over the `__functions` dictionary. That would give us O(n) performance in terms of the number of plugin functions.

tests/data/modules_v2/cross_module_dependency/README.md
arnaudsjs, line 1: Why do we need this file?

tests/test_module_loader.py
```diff
@@ -1363,3 +1363,72 @@ async def test_v2_module_editable_with_links(tmpvenv_active: tuple[py.path.local
     module: Optional[ModuleV2] = ModuleV2Source([]).get_installed_module(DummyProject(autostd=False), "minimalv2module")
     assert module is not None
     assert module.path == module_dir
+
+
+def test_cross_module_dependency(local_module_package_index: str, snippetcompiler_clean, capsys) -> None:
```
arnaudsjs, line 1368: Why `snippetcompiler_clean` instead of `snippetcompiler`?

arnaudsjs (changes requested):

src/inmanta/compiler/__init__.py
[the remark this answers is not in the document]
sanderr: I agree. The current `clear` method must remain as is (stable API) but adding a second method to clear/unregister multiple modules sounds like a good idea. You should make sure it also works for submodules, because I think the current `clear` method does not (and is not intended to). We should probably have a test case for that scenario.

Alternatively, an even simpler option that I hadn't considered before is to just leave them in and ignore them. The important thing is we don't add the `PluginStatement`. So we could keep the implementation here mostly as is but drop the `PluginMeta.clear()` call. This may also be more reliable for in-process recompiles like pytest-inmanta because we keep the plugin objects around. And I don't immediately see any bad side effects. Wdyt?

tests/test_module_loader.py
```diff
@@ -1363,3 +1363,72 @@ async def test_v2_module_editable_with_links(tmpvenv_active: tuple[py.path.local
...
+    """
+    This test checks that the python code living in the inmanta_plugins dir of a module ('anothermod' in this test case)
+    that is not loaded, can be used from the plugins of another module. ('cross_module_dependency' in this test case)
+    """
+
+    def check_name_space(name_space: Dict[str, any], includes: Sequence[str], excludes: Sequence[str]) -> None:
```
sanderr, line 1374:
```suggestion
    def check_name_space(name_space: Dict[str, object], includes: Sequence[str], excludes: Sequence[str]) -> None:
```

sanderr (changes requested):
Looks mostly good to me

changelogs/unreleased/5651-ignore-plugins-for-modules-that-arent-loaded.yml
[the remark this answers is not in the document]
sanderr: I would leave it out, partially because it changes a critical component and I'm not sure if we should do that in iso5, partially because of what Bart said on Slack earlier today.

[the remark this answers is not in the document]
sanderr: I'd also delay merging until we've branched off for the release. I wouldn't be surprised if this causes some fallout.

src/inmanta/compiler/__init__.py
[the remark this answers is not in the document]
arnaudsjs: I like the proposal to leave them in and just ignore them.

sanderr (changes requested):
Could you still add a test case for submodules? e.g. if `mymod::submod` and `mymod` but contain a plugin, and `mymod::submod` isn't loaded but `mymod` is, make sure that `mymod::myplugin()` still works.

arnaudsjs (approved):

tests/data/modules_v2/complex_module_dependencies_mod2/inmanta_plugins/submod/__init__.py
sanderr, line 1: This file is in the wrong place: should be `inmanta_plugins/complex_module_dependencies_mod2/submod/__init__.py`

sanderr (changes requested):
Final remark

sanderr (approved):

PR #5786, opened 2023-04-03
Added documentation regarding inter-resource set dependency constraints

### LINKED ISSUE #5679: Improve documentation regarding partial compiles
Partial compiles have the limitation that the resources from one resources cannot have a dependency (requires/provides) on resources from another resource set. This issue should add that limitation to the documentation. It would be good to add some diagrams that clearly describe which dependencies are allowed and which are not.

arnaudsjs:
# Description

Added diagrams to the documentation that explain the limitations regarding inter-resource set dependencies when partial compiles are enabled.

closes #5679

# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/model_developers/resource_sets.rst
+++ b/docs/model_developers/resource_sets.rst
@@ -483,10 +483,12 @@ Inter-resource set dependencies
 -------------------------------
 
 Resources within a resource set can only depend on resources within the same resource set or on shared resources.
-The diagram below provides an example where the resource dependency graph satisfies this requirement.
+Shared resources on the other hand can have dependencies on any resource in the model. The diagram below provides
+an example where the resource dependency graph satisfies these requirements. The arrows in the diagram show the
+requires relationship between entities/resources.
 
 .. digraph:: resource_sets_example_valid_dependencies
-    :caption: Two resource sets satisfying the dependency constraint
+    :caption: Two resource sets satisfying the dependency constraints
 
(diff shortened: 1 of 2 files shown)


sanderr (changes requested):

Hugo-Inmanta (approved):

docs/model_developers/resource_sets.rst
```diff
@@ -480,6 +479,96 @@ behavior because the allocator guarantees uniqueness for the host id:
...
+            label = "Resource set 1";
+            labelloc = "bottom";
+        }
+        color = "green";
+        label = "service 1";
+        labelloc = "top";
+    }
+
+    {"Host(id=269)" "Host(id=31)"} -> "AgentConfig";
+    {"Host(id=269)"} -> "Host(id=31)" [color=red];
```
Hugo-Inmanta, line 569: One minor visual thing: I feel the second diagram would look nicer if both service 0 and service 1 were aligned, but I don't know how difficult this would be. (Or if it would even actually look nicer in the end)
For reference: service1 is slightly below service0:
image

arnaudsjs: I tried to align them, but it seems that the feature to align them has a bug that causes all boxes in the clustergraph to disappear. So I am afraid we will have to keep it like it is, but I do agree it would be better to align them.

sanderr (approved):

PR #5791, opened 2023-04-03
unmanaged resources storage

### LINKED ISSUE #5690: Unmanaged Resources storage
# Goal

Implement a first storage backend for the unmanaged resources

this can be based on the [POC](https://github.com/inmanta/inmanta-core/pull/5278)

API endpoint URL
POST api/v2/unmanaged/<rid>/ (one)   (and get)
POST api/v2/unmanaged/  (batch)           (and get)

## In scope

- Storage of discovered resources
- on a per resource basis
- with associated ID
- paging

## Out of scope 

- typed storage
- cleanup 
- traceability (where was this discovered)
- versions/history
- filtering and custom sorting

FloLey:
# Description

Implement a first storage backend for the unmanaged resources

add following API endpoint URLs:
POST api/v2/unmanaged/<unmanaged_resource_id> (and GET)
POST api/v2/unmanaged/ (batch) (and GET)

closes #5690 

# Self Check:

- [x] Changelog entry


src/inmanta/protocol/methods_v2.py
```diff
@@ -1357,3 +1357,73 @@ def set_password(username: str, password: str) -> None:
...
+@typedmethod(
+    path="/unmanaged/<unmanaged_resource_id>",
+    validate_sid=False,
+    operation="POST",
+    agent_server=True,
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.agent],
+    api_version=2,
+)
+def unmanaged_resource_create(env: uuid.UUID, unmanaged_resource_id: str, values: Dict[str, str]) -> None:
```
arnaudsjs, line 1371: The convention is to use the `tid` to reference the environment id in the API endpoint. We should use consistent naming here and also in the other API endpoints.
```suggestion
def unmanaged_resource_create(tid: uuid.UUID, unmanaged_resource_id: str, values: Dict[str, str]) -> None:
```

src/inmanta/protocol/methods_v2.py
```diff
@@ -1357,3 +1357,73 @@ def set_password(username: str, password: str) -> None:
...
+)
+def unmanaged_resources_get_batch(
+    tid: uuid.UUID, start_resource_id: Optional[str] = None, end_resource_id: Optional[str] = None, limit: Optional[int] = None
+) -> List[model.UnmanagedResource]:
+    """
+    Get the unmanaged resources with paging
+    :param tid: the id of the environment in which to get the unmanaged resources.
+    :param start_resource_id: the unmanaged_resource_id from which to start the paging
+    :param end_resource_id: the unmanaged_resource_id where to finish the paging
+    :param limit: the maximal number of resources to return
```
arnaudsjs, line 1427: This paging API look different from all the other paging APIs we already have. It for example doesn't have a field to indicate the order in which we want to page the output. Is that done on purpose?

src/inmanta/protocol/methods_v2.py
```diff
@@ -1357,3 +1357,73 @@ def set_password(username: str, password: str) -> None:
...
+    path="/unmanaged/<unmanaged_resource_id>",
+    operation="GET",
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.api],
+    api_version=2,
+)
+def unmanaged_resources_get(tid: uuid.UUID, unmanaged_resource_id: str) -> model.UnmanagedResource:
+    """
+    Get a single discovered resource
+    :param tid: the id of the environment in which to get the unmanaged resource.
```
arnaudsjs, line 1407: I seem to recall that the documentation generation tool requires an empty line between the description and parameters. I am not sure, but it doesn't hurt to add the empty line anyway. 
```suggestion
    Get a single discovered resource.

    :param tid: the id of the environment in which to get the unmanaged resource.
```

src/inmanta/server/services/resourceservice.py
```diff
@@ -1038,3 +1039,62 @@ async def versioned_resource_details(
         if not resource:
             raise NotFound("The resource with the given id does not exist")
         return resource
+
+    @handle(methods_v2.unmanaged_resource_create)
+    async def unmanaged_resource_create(
+        self, env: data.Environment, unmanaged_resource_id: str, values: Dict[str, str]
+    ) -> None:
+        unmanaged_resource = UnmanagedResource(unmanaged_resource_id=unmanaged_resource_id, values=values)
```
arnaudsjs, line 1047: Why not have the `UnmanagedResource` as a parameter to the API endpoint, instead of passing in the three individual parameters? We do the same in the batch API call.

src/inmanta/server/services/resourceservice.py
```diff
@@ -1038,3 +1039,62 @@ async def versioned_resource_details(
...
+    async def unmanaged_resources_create_batch(
+        self, env: data.Environment, unmanaged_resources: List[UnmanagedResource]
+    ) -> None:
+        resources: List[data.UnmanagedResource] = []
+        for res in unmanaged_resources:
+            unmanaged_resource = data.UnmanagedResource(
+                environment=env,
+                unmanaged_resource_id=res.unmanaged_resource_id,
+                values=res.values,
+            )
```
arnaudsjs, line 1062: I would add a `to_dao` method to the `UnmanagedResource` class. Then we can reuse that logic across the different API endpoints.

src/inmanta/server/services/resourceservice.py
```diff
@@ -1038,3 +1039,62 @@ async def versioned_resource_details(
...
+            resources.append(unmanaged_resource)
```
arnaudsjs, line 1063: Not necessarily a change request, because it's more personal taste than anything else, but a list comprehension would look bad here.

tests/test_unmanaged_resources.py
```diff
@@ -0,0 +1,129 @@
+"""
+    Copyright 2016 Inmanta
```
arnaudsjs, line 2:
```suggestion
    Copyright 2023 Inmanta
```

tests/test_unmanaged_resources.py
```diff
@@ -0,0 +1,129 @@
...
+    result = await client.unmanaged_resource_create(
+        environment, "test::Resource[agent1,key=key],v=1", {"value1": "test1", "value2": "test2"}
+    )
+    assert result.code == 200
+
+    result = await client.unmanaged_resources_get(environment, "test::Resource[agent1,key=key],v=1")
+    assert result.code == 200
+
+    assert result.result["data"]["unmanaged_resource_id"] == "test::Resource[agent1,key=key],v=1"
+    assert result.result["data"]["values"] == {"value1": "test1", "value2": "test2"}
```
arnaudsjs, line 33:
```suggestion
    unmanaged_resource_id = "test::Resource[agent1,key=key],v=1"
    values = {"value1": "test1", "value2": "test2"}
    result = await client.unmanaged_resource_create(
        environment, unmanaged_resource_id, values
    )
    assert result.code == 200

    result = await client.unmanaged_resources_get(environment, unmanaged_resource_id)
    assert result.code == 200

    assert result.result["data"]["unmanaged_resource_id"] == unmanaged_resource_id
    assert result.result["data"]["values"] == values
```

tests/test_unmanaged_resources.py
```diff
@@ -0,0 +1,129 @@
...
+    assert result.result["data"][0]["unmanaged_resource_id"] == resources[2]["unmanaged_resource_id"]
+    assert result.result["data"][0]["values"] == resources[2]["values"]
+
+
+async def test_discovery_resource_bad_res_id(server, client, agent, environment):
+    """
+    Test that exceptions are raised when creating unmanaged resources with invalid IDs.
+    """
+    result = await client.unmanaged_resource_create(environment, "test", {"value1": "test1", "value2": "test2"})
+    assert result.code == 500
```
arnaudsjs, line 119: wrong input from a user should results in a 400. Or at least something in the range 4xx. 500 means that the server is causing the problem, which is not the case here.

tests/test_unmanaged_resources.py
arnaudsjs, line 1: Do you also test the behavior when you store the same unmanaged resource twice. Or when a call to the batch API contains duplicates.

tests/test_unmanaged_resources.py
```diff
@@ -0,0 +1,129 @@
...
+    result = await client.unmanaged_resources_get_batch(
+        environment,
+        start_resource_id="test::Resource[agent1,key2=key2],v=1",
+        end_resource_id="test::Resource[agent1,key4=key4],v=1",
+        limit=3,
+    )
+    assert result.code == 200
+    assert len(result.result["data"]) == 1
+    assert result.result["data"][0]["unmanaged_resource_id"] == resources[2]["unmanaged_resource_id"]
+    assert result.result["data"][0]["values"] == resources[2]["values"]
```
arnaudsjs, line 111: I don't understand this case. Could you elaborate.

arnaudsjs (dismissed):

src/inmanta/data/__init__.py
```diff
@@ -5782,6 +5782,26 @@ def to_dao(self) -> m.User:
...
 
+class UnmanagedResource(BaseDocument):
+    """
+    :param environment: the environment of the resource
+    :param unmanaged_resource_id: The id of the resource
+    :param values: The values associated with the unmanaged_resource
+    """
+
+    environment: uuid.UUID
+    unmanaged_resource_id: str
```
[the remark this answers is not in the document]
FloLey: I first had done it like this but than I had typing errors in mypy. I'll change it back and see if I can fix the typing errors in an other way

src/inmanta/server/services/resourceservice.py
FloLey: I'll change it :)

src/inmanta/data/__init__.py
```diff
@@ -5782,6 +5782,26 @@ def to_dao(self) -> m.User:
...
+    values: Dict[str, str]
```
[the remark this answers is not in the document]
FloLey: form what I understood the ticket says this (typed storage) is out of scoop?

src/inmanta/server/services/resourceservice.py
FloLey: I though for the single calls we wanted to have the api call as close as possible to this one:
```
@method(
    path="/resource/<id>",
    operation="GET",
    agent_server=True,
    validate_sid=False,
    arg_options=ENV_OPTS,
    api=True,
    client_types=[const.ClientType.api, const.ClientType.agent],
)
def get_resource(
    tid: uuid.UUID, id: str, logs: bool = None, status: bool = None, log_action: const.ResourceAction = None, log_limit: int = 0
):
```

I mean having the Id in the path

tests/test_unmanaged_resources.py
FloLey: We specify the start_resource_id (key2) with a limit of 3, so it could potentially get key3, key4 and key 5.This test verifies that the end_resource_id (key4) is taken into account and that it will stop returning values startin from key4, resulting in only key3 being returned

src/inmanta/data/model.py
```diff
@@ -716,3 +716,20 @@ class LoginReturn(BaseModel):
...
+
+    unmanaged_resource_id: str
+    values: Dict[str, str]
+
+    @validator("unmanaged_resource_id")
+    @classmethod
+    def unmanaged_resource_id_is_resource_version_id(cls, v: str) -> Optional[Any]:
+        if resources.Id.is_resource_version_id(v):
+            return v
+        raise ValueError(f"id {v} is not of type ResourceVersionIdStr")
```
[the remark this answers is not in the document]
wouterdb: that is out-of-scope for now, now it is `ResourceIdStr` not `ResourceVersionIdStr`

src/inmanta/server/services/resourceservice.py
```diff
@@ -1038,3 +1039,56 @@ async def versioned_resource_details(
...
+        if not result:
+            raise NotFound(f"unmanaged_resource with name {unmanaged_resource_id} not found in env {env}")
+        dto = result.to_dto()
+        return dto
+
+    @handle(methods_v2.unmanaged_resources_get_batch, env="tid")
+    async def unmanaged_resources_get_batch(
+        self, env: data.Environment, start_resource_id: str, end_resource_id: str, limit: int
+    ) -> List[UnmanagedResource]:
+        if limit is None:
```
wouterdb, line 1072: Best to use DataView here to be in line with other paged resources

src/inmanta/server/services/resourceservice.py
arnaudsjs: The advantage of the single `UnmanagedResource` parameter is that the protocol layer will do the type validation, while now it has to be done within the body of the API endpoint.

arnaudsjs: Ah. Now I see. It's because the `unmanaged_resource_id` is part of the path of the URL. It makes sense now.

src/inmanta/protocol/methods_v2.py
arnaudsjs: Indeed. It seems our way of paging is not as standardized as I thought it way. That's a shame. You can ignore this comments.

FloLey: well I should use DataView (see wouters comment) so we do have a standard

src/inmanta/data/dataview.py
```diff
@@ -1174,3 +1174,67 @@ def construct_dtos(self, records: Sequence[Record]) -> Sequence[model.Agent]:
...
+    def construct_dtos(self, records: Sequence[Record]) -> Sequence[dict[str, str]]:
+        return [
+            model.UnmanagedResource(
+                unmanaged_resource_id=res.unmanaged_resource_id,
+                values=res.values,
+            ).dict()
+            for res in records
+        ]
+
+    async def get_data(self) -> Tuple[Sequence[T_DTO], Optional[PagingBoundaries]]:
```
wouterdb, line 1229: do you even need to override this one?

src/inmanta/protocol/methods_v2.py
```diff
@@ -1357,3 +1357,93 @@ def set_password(username: str, password: str) -> None:
...
+    path="/unmanaged",
+    operation="GET",
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.api],
+    api_version=2,
+)
+def unmanaged_resources_get_batch(
+    tid: uuid.UUID,
+    limit: Optional[int] = None,
+    first_id: Optional[model.ResourceVersionIdStr] = None,
```
wouterdb, line 1426: do we need both first_id and start? I think with single orders, this is not required?

(not quite sure)

tests/test_unmanaged_resources.py
```diff
@@ -0,0 +1,165 @@
...
+    assert result.result["data"]["unmanaged_resource_id"] == unmanaged_resource_id
+    assert result.result["data"]["values"] == values
+
+    # try to store the same resource a second time
+    result = await agent._client.unmanaged_resource_create(
+        tid=environment,
+        unmanaged_resource_id=unmanaged_resource_id,
+        values=values,
+    )
+    assert result.code == 500
```
wouterdb, line 50: a 500 is NOT what I want here. 409 would be more appropriate

wouterdb (changes requested):

src/inmanta/protocol/methods_v2.py
```diff
@@ -1357,3 +1357,93 @@ def set_password(username: str, password: str) -> None:
...
+
+
+@typedmethod(
+    path="/unmanaged/<unmanaged_resource_id>",
+    operation="GET",
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.api],
+    api_version=2,
+)
+def unmanaged_resources_get(tid: uuid.UUID, unmanaged_resource_id: str) -> model.UnmanagedResource:
```
sanderr, line 1407: Isn't this a `ResourceIdStr`?

sanderr (approved):

changelogs/unreleased/5690-unmanaged-resources-storage.yml
```diff
@@ -0,0 +1,5 @@
+description: Implement storage backend for the unmanaged resources
+issue-nr: 5690
+change-type: minor
+destination-branches: [master]
```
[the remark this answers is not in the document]
FloLey: What do you mean? if it only goes into master or not?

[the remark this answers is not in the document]
FloLey: I don't think we did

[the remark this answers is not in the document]
sanderr: I looked it over once more. I would include iso6

wouterdb (approved):

Issue #5790, opened 2023-04-03
type inference uses is_subclass in wrong direction


sanderr:
`my_entity.is_subclass(other)` seems to return `True` iff `other` is a subclass of `my_entity`. [Here](https://github.com/inmanta/inmanta-core/blob/00d42e775257cb758469c0ebfc367bbddf78ac93/src/inmanta/ast/statements/generator.py#L616) it is used the other way around.

- [x] Verify claim about the semantics of this method + clarify docstrings of this method and `is_parent` and the names of their arguments.
- [x] Create a type inference test case that reproduces the bug, PoC:
```
# mymod:
entity A: end
entity B:
    int n = 0
end
A.b [1] -- B

# main.cf
import mymod

entity A extends mymod::A: end
entity B extends mymod::B:
    int n = 1
end

a = A(b=B())
n = a.b.n
# assert
n = 1
```
- [x] Fix usage and make sure the test succeeds.

PR #5792, opened 2023-04-03
Issue/5790 type inference uses is subclass in wrong direction

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/5790

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/tests/compiler/test_type_hinting.py
+++ b/tests/compiler/test_type_hinting.py
@@ -165,12 +165,3 @@ n = a.b.n
     (_, scopes) = compiler.do_compile()
     root = scopes.get_child("__config__")
     assert 1 == root.lookup("n").get_value()
-
-
-# mymod:
-
-
-# main.cf
-
-# assert
-# n = 1
--- a/tests/compiler/test_type_hinting.py
+++ b/tests/compiler/test_type_hinting.py
@@ -140,3 +140,37 @@ implement A using std::none
         """,
         "Can not assign a value of type A to a variable of type int (reported in Construct(A) ({dir}/main.cf:8))",
     )
+
+
+def test_type_inference_is_subclass_right_direction_5790(snippetcompiler):
+    snippetcompiler.setup_for_snippet(
+        """
+import test_5790
+
+entity A extends test_5790::A:
+end
+
+entity B extends test_5790::B:
+    int n = 1
+end
+
+implement A using std::none
+implement B using std::none
+
+
+a = A(b=B())
+n = a.b.n
+        """,
+    )
+    (_, scopes) = compiler.do_compile()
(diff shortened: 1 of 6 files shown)


tests/compiler/test_type_hinting.py
```diff
@@ -140,3 +140,28 @@ def test_basic_type_hint_attribute_collision(snippetcompiler):
         """,
         "Can not assign a value of type A to a variable of type int (reported in Construct(A) ({dir}/main.cf:8))",
     )
+
+
+def test_type_inference_is_subclass_right_direction_5790(snippetcompiler):
```
sanderr, line 145: Did this indeed fail without your patch?

sanderr (changes requested):
I'm adding @wouterdb as a second reviewer, primarily to make sure he thinks the parameter name change is a good idea. I'm 100% in favor and I don't foresee any issues but I could be missing something.

wouterdb (approved):

tests/compiler/test_type_hinting.py
Hugo-Inmanta: It was failing indeed before the patch with 

```term
E                   inmanta.ast.AmbiguousTypeException: Could not determine namespace for type B. 2 possible candidates exists: [__config__::B, test_5790::B]. To resolve this, use the fully qualified name instead of the short name. (reported in a = Construct(A) (/tmp/tmpr23y_be3/main.cf:15:9))

```

tests/compiler/test_type_hinting.py
```diff
@@ -83,29 +83,6 @@ def test_basic_type_hint_name_collision(snippetcompiler):
...
-implement elaboratev1module::A using std::none
-implement A using std::none
-
-        """,
-        "Could not determine namespace for type A. 2 possible candidates exists: [__config__::A, elaboratev1module::A]. "
-        "To resolve this, use the fully qualified name instead of the short name. "
-        "(reported in Construct(A) ({dir}/main.cf:10:7))",
-    )
-
-
```
Hugo-Inmanta, line 108: I figure this test that is now failing can go now that we fixed the bug ?

sanderr: I think the test as it is now is wrong, but I think we do still need one that tests for this error behavior in a correct way. You can drop it for now (this bug is causing issues for solutions so the faster we merge it the better) but after we've reached a conclusion on Slack we should add a replacement in a second PR.

sanderr (approved):

tests/compiler/test_type_hinting.py
wouterdb: I agree

PR #5801, opened 2023-04-04
Add support keyword-only attributes on Plugins

### LINKED ISSUE #5706: Support keyword-only arguments for plugins
A plugin with keyword-only arguments for plugins is not interpreted correctly by [`inmanta.plugins.Plugin._load_signature`](https://github.com/inmanta/inmanta-core/blob/f8c13bdf7d7410084cfcece5439e1cffb72e24f4/src/inmanta/plugins.py#L222) because it ignores these arguments.

```python
@plugin
def myplugin(*, x: "int", y: "int") -> "int":
    return x * y
```

`std::print(std::myplugin(x=2, y=5))`

```
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/app.py", line 874, in app
    options.func(options)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/app.py", line 412, in compile_project
    do_compile()
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/compiler/__init__.py", line 75, in do_compile
    success = sched.run(compiler, statements, blocks)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/execute/scheduler.py", line 406, in run
    next.execute()
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/ast/statements/call.py", line 284, in execute
    self.function.resume(requires, self.resolver, self.queue, self.result)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/ast/statements/call.py", line 128, in resume
    self.function.call_in_context(arguments, kwargs, resolver, queue, result)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/ast/statements/call.py", line 228, in call_in_context
    no_unknows = self.plugin.check_args(args, kwargs)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/plugins.py", line 339, in check_args
    raise Exception(
Exception: Incorrect number of arguments for myplugin() -> int. Expected at most 0, got 2
```

arnaudsjs:
# Description

Added support for keyword-only attributes on plugins.

closes #5706

# Self Check:

- [x] Changelog entry


src/inmanta/plugins.py
```diff
@@ -228,36 +258,57 @@ def __init__(self, namespace: Namespace) -> None:
...
+                return PluginArgument(argument_name, spec_type, is_kw_only_argument=False, default_value=default_value)
             else:
-                if default_start is not None and default_start <= i:
-                    default_value = arg_spec.defaults[default_start - i]
+                return PluginArgument(argument_name, spec_type, is_kw_only_argument=False)
 
-                    arguments.append((arg, spec_type, default_value))
+        def process_arg(index: int, argument_name: str, is_kwonly_arg: bool) -> None:
+            if argument_name not in arg_spec.annotations:
+                raise Exception(f"All arguments of plugin '{function.__name__}' should be annotated")
```
FloLey, line 295: it's a detail but maybe this code path/exception could be tested aswell?

changelogs/unreleased/5706-add-support-keyword-only-attributes.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: Added support for keyword-only attributes in plugins
```
sanderr, line 2:
```suggestion
description: Added support for keyword-only arguments in plugins
```

src/inmanta/plugins.py
```diff
@@ -345,50 +392,46 @@ def check_args(self, args: List[object], kwargs: Dict[str, object]) -> bool:
...
+            )
         # check for kwargs overlap with positional arguments
-        if not present_kwargs.isdisjoint(present_positional_args):
+        overlapping_args: FrozenSet[str] = present_kwargs.intersection(present_positional_args)
+        if overlapping_args:
             raise RuntimeException(
                 None,
-                "Multiple values for %s in %s()"
-                % (",".join(present_kwargs.intersection(present_positional_args)), self.__class__.__function_name__),
+                "Multiple values for %s in %s()" % (",".join(overlapping_args), self.__class__.__function_name__),
```
FloLey, line 416: this could also be tested

src/inmanta/plugins.py
```diff
@@ -228,36 +258,57 @@ def __init__(self, namespace: Namespace) -> None:
...
+            default_start_for_args = None
 
-        arguments = []
-        for i in range(len(arg_spec.args)):
-            arg = arg_spec.args[i]
+        arguments: List[PluginArgument] = []
 
-            if arg not in arg_spec.annotations:
-                raise Exception("All arguments of plugin '%s' should be annotated" % function.__name__)
+        def process_kw_only_args(argument_name: str, spec_type: object) -> PluginArgument:
```
sanderr, line 279: While we're cleaning up, `annotation` may be more clear than `spec_type`?

src/inmanta/plugins.py
```diff
@@ -345,50 +392,46 @@ def check_args(self, args: List[object], kwargs: Dict[str, object]) -> bool:
...
+                "Multiple values for %s in %s()" % (",".join(overlapping_args), self.__class__.__function_name__),
             )
 
-        def is_valid(expected_arg: Tuple[Optional[Type[object]], str], expected_type: Type[object], arg: object) -> bool:
+        def is_valid(expected_arg: PluginArgument, expected_type: Type[object], arg: object) -> bool:
             if isinstance(arg, Unknown):
                 return False
 
-            if expected_arg[0] is not None and not self._is_instance(arg, expected_type):
+            if not self._is_instance(arg, expected_type):
```
sanderr, line 423: You dropped a part of this condition, could you explain why? Was it just an unreachable branch? `expected_arg[0]` corresponds with `expected_arg.arg_name`, right? But `expected_type` could still be `None` (incorrectly typed here?) because `to_type` returns `Optional`. I'll get back to this comment.

sanderr (commented):
One thing I'm not sure of and need to check tomorrow. Other than that everything looks good to me. Nice improvements on the tuple -> class front.

@edvgui FYI

src/inmanta/plugins.py
arnaudsjs: Yes. Indeed. An argument always has a name so `expected_arg[0] is not None` can never occur.  I don't really see why that condition was here. But please review very carefully, I might be missing something.

src/inmanta/plugins.py
arnaudsjs: This is already tested here: https://github.com/inmanta/inmanta-core/blob/ca9b1725b12bc8fca4aa9f9fcdf33a4b6549d29b/tests/compiler/test_plugins.py#L87

sanderr (approved):

PR #5811, opened 2023-04-05
Fix bug reporting incorrectly length compile queue

arnaudsjs:
# Description

Fix bug that causes the `/serverstatus` endpoint to report an incorrect length of the compiler queue.

# Self Check:

- [x] Changelog entry


src/inmanta/server/services/compilerservice.py
```diff
@@ -621,6 +622,7 @@ async def request_recompile(
             failed_compile_message=failed_compile_message,
         )
         await compile.insert()
+        self._queue_count_cache[env.id] += 1
```
sanderr, line 625: This relates to inmanta/inmanta-lsm#1249, we'll have to make sure the solution we pick takes this into account (e.g. if we add the compile within a transaction we should only increment the counter when it commits.

(Just a note, I'll refer here from the other ticket, no need to do anything about this now)

sanderr (approved):

sanderr (approved):

FloLey (approved):

PR #5817, opened 2023-04-06
Drop support for purge on delete

### LINKED ISSUE #5677: drop support for purge on delete
# Description

Completely drop support for purge on delete. If feasible, drop it from iso7 only but if it requires a lot of refactoring, delay the change for a few months to keep the branches closer together.

# Original ticket description
## incorrect dependency inversion for purge-on-delete
When the server receives a version and purge on delete is enabled, it inverts requires-provides edges for deleted resources and adds them to the model with `purged=True`. Under some circumstances the inversion logic is incorrect, leading to a circular dependency.

The source that does the inversion is part of the `_put_version` method in the orchestration service.

The inversion logic is as follows: it first collects all `requires` for the previous version of each deleted resource, then it empties each deleted resource's `requires` and `provides`. It then iterates over each of these resources, and for each `requires` that is part of the current model, it inverts it by instead adding the dependency to `provides` (if it's a cross-agent dependency) and adds this one as `requires` for the other one.

This is where it can go wrong: suppose `a` requires `b`, where `a` and `b` live on different agents and only `a` is deleted from the model. Then this logic goes through the following steps
  a. empty `a`'s requires and provides
  b. leave `b`'s requires and provides as is because it is not being purged
  c. invert: set `b.requires = a` and `a.provides = b`
  => `b` still has `b.provides = a`, resulting in a circular (and incorrect) dependency. A similar issue occurs if the dependency would have been the other way around.

*This issue was first reported [here](https://github.com/inmanta/inmanta-core/pull/5580#discussion_r1124550278) (linked for a reference to the discussion)*

arnaudsjs:
# Description

This PR removes support for the purge_on_delete feature and the `POST /decommission/<id>`.

closes #5677

# Self Check:

- [x] Changelog entry


wouterdb (approved):
Beautiful

tests/db/migration_tests/dumps/v202303071.sql
```diff
@@ -441,8 +441,8 @@ CREATE TABLE public.unknownparameter (
 --
```
FloLey, line 441: Why did the previous dump get changed?

FloLey (approved):
Just a small question

tests/db/migration_tests/dumps/v202303071.sql
arnaudsjs: I wanted to have an environment in the dump that had `purge_on_delete` enabled, to test the db migration script. No such environment was present in the existing dump.

PR #5826, opened 2023-04-11
Make status endpoint less fragile to timeouts

wouterdb:
# Description
Make status endpoint less fragile to timeouts
# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/server.py
+++ b/src/inmanta/server/server.py
@@ -133,7 +133,12 @@ class Server(protocol.ServerSlice):
             try:
                 return SliceStatus(name=slice_name, status=await asyncio.wait_for(slice.get_status(), 0.1))
             except TimeoutError:
-                return SliceStatus(name=slice_name, status={"error": f"timeout on data collection for {slice_name}, consult the server log for additional information"})
+                return SliceStatus(
+                    name=slice_name,
+                    status={
+                        "error": f"timeout on data collection for {slice_name}, consult the server log for additional information"
+                    },
+                )
             except Exception:
                 LOGGER.error(
                     f"The following error occured while trying to determine the status of slice {slice_name}",
--- a/src/inmanta/server/server.py
+++ b/src/inmanta/server/server.py
@@ -136,7 +136,8 @@ class Server(protocol.ServerSlice):
                 return SliceStatus(
                     name=slice_name,
                     status={
-                        "error": f"timeout on data collection for {slice_name}, consult the server log for additional information"
+                        "error": f"timeout on data collection for {slice_name}, "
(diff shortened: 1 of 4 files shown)


src/inmanta/server/server.py
```diff
@@ -128,15 +129,21 @@ async def get_server_status(self) -> StatusResponse:
...
             )
 
-        slices = []
-        for slice_name, slice in self._server.get_slices().items():
+        async def collect_for_slice(slice_name: str, slice: protocol.ServerSlice) -> SliceStatus:
             try:
-                slices.append(SliceStatus(name=slice_name, status=await slice.get_status()))
+                return SliceStatus(name=slice_name, status=await asyncio.wait_for(slice.get_status(), 0.1))
+            except TimeoutError:
+                return SliceStatus(name=slice_name, status={"error": "timeout after 100ms"})
```
sanderr, line 136: Should we have the frontend retry regularly if this occurs? Also for handling these new error responses in general, should we create a frontend ticket?

sanderr (approved):

src/inmanta/server/server.py
wouterdb:
```suggestion
                return SliceStatus(name=slice_name, status={"error": f"timeout on data collection for {slice_name}, consult the server log for additional information"})
```

wouterdb: I discussed it and this is what we came up with

sanderr: So no retries? I assume this is an exceptional case and indicates something is wrong?

wouterdb: Indeed. (and they naturally retry every 5 seconds)

src/inmanta/server/services/databaseservice.py
```diff
@@ -116,7 +116,7 @@ def start_monitor(self) -> None:
     async def get_connection_status(self) -> bool:
         if self._pool is not None and not self._pool._closing and not self._pool._closed:
             try:
-                async with self._pool.acquire(timeout=10):
+                async with self._pool.acquire(timeout=5):
```
Hugo-Inmanta, line 119: I'm curious, what's the rationale behind this change ?

Hugo-Inmanta (approved):

src/inmanta/server/services/databaseservice.py
wouterdb: oh, right not required, thx

wouterdb, line 119:
```suggestion
                async with self._pool.acquire(timeout=10):
```

PR #5829, opened 2023-04-12
Issue/5755 verbose flags ignored at some locations

### LINKED ISSUE #5755: verbose flags ignored at some locations
`inmanta module install -e . -vvvv` does not log at debug level, `inmanta -vvvv module install -e .` does. This is very unintuitive. Similar issues are likely present for other commands and flags.

No tests required.

If `inmanta module -m` causes trouble, drop it.

Hugo-Inmanta:
# Description

This PR adds the possibility to define verbosity flags on the sub-commands

closes #5755 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -298,7 +298,7 @@ class ExperimentalFeatureFlags:
                 option.set("true")
 
 
-def compiler_config(parser: argparse.ArgumentParser) -> None:
+def compiler_config(parser: argparse.ArgumentParser, shared_args_parser: argparse.ArgumentParser) -> None:
     """
     Configure the compiler of the export function
     """
@@ -420,7 +420,7 @@ def list_commands(options: argparse.Namespace) -> None:
         print(" %s: %s" % (cmd, info["help"]))
 
 
-def help_parser_config(parser: argparse.ArgumentParser) -> None:
+def help_parser_config(parser: argparse.ArgumentParser, shared_args_parser: argparse.ArgumentParser) -> None:
     parser.add_argument("subcommand", help="Output help for a particular subcommand", nargs="?", default=None)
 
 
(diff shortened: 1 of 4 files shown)


src/inmanta/moduletool.py
```diff
@@ -603,14 +625,28 @@ def modules_parser_config(cls, parser: ArgumentParser) -> None:
...
-        subparser.add_parser("verify", help="Verify dependencies and frozen module versions")
+        subparser.add_parser(
+            "verify",
+            help="Verify dependencies and frozen module versions",
+            parents=[verbosity_parser],
+        )
 
-        commit = subparser.add_parser("commit", help="Commit all changes in the current module.")
+        commit = subparser.add_parser(
+            "commit", help="Commit all changes in the current module.", parents=[long_verbosity_parser]
```
Hugo-Inmanta, line 648: A specific verbosity parser that only parses "--verbose" (and not "-v") is required for this sub-command since it already defines a "-v" option `commit.add_argument("-v", "--version", help="Version to use on tag")`

Hugo-Inmanta: As discussed on slack: leave commit command untouched, it is deprecated and will be dropped from iso7 in the near future.

sanderr (changes requested):
Nice work! However, I think it would be better to make the verbosity option the default so that if we add new commands they get it automatically. So I think it would be better to have an opt-out boolean in the `command` than the callable you use now. e.g. `app.cmd_parser` could then be responsible for setting the option, but I have no strong opinion on the implementation details.

src/inmanta/app.py
```diff
@@ -41,11 +41,12 @@
 import time
 import traceback
 import typing
+from argparse import ArgumentParser
 from asyncio import ensure_future
 from configparser import ConfigParser
 from threading import Timer
 from types import FrameType
-from typing import Any, Callable, Coroutine, Dict, Optional
+from typing import Any, Callable, Coroutine, Dict, Optional, Sequence
```
sanderr, line 49: `Sequence` should be used from `collections.abc` as `abc.Sequence`. Same goes for some of the others in here but let's do it right at least for any imports we add.

src/inmanta/command.py
```diff
@@ -66,6 +52,7 @@ def add(
         parser_config: Optional[ParserConfigType],
         require_project: bool = False,
         aliases: List[str] = [],
+        verbose_opt_out: bool = False,
```
sanderr, line 55: I find this a rather strange name. Wouldn't something like `add_verbosity_flag: bool = True` be more to the point? Not a strong change request.

sanderr (approved):

PR #5833, opened 2023-04-14
Issue/5708 import modules agent logger helpers into core

### LINKED ISSUE #5708: Import modules agent logger helpers into core
The logging interface exposed by the `HandlerContext` object is very similar to the one of `logging.Logger`, the definition is in fact the same, [but the behavior is not](https://inmanta.slack.com/archives/CKRF0C8R3/p1675417876753029).  When developing modules, it happens regularly that we want to reuse in the tests of the module, a helper class defined in the plugins, which is meant to be used with the handler logger initially.

Those modules then commonly define an interface representing the part of the `HandlerContext` dealing with logging, and a class, which allows to wrap a `logging.Logger` object to expose the same interface as the handler, while still processing the logs with the python `logging` library. 

Here is where it was originally implemented: https://code.inmanta.com/solutions/modules/restbase/-/blob/master/inmanta_plugins/restbase/handler_logger.py

There was at some point a discussion to make it the other way around, in case someone is interested: https://inmanta.slack.com/archives/CKRF0C8R3/p1675420892413819?thread_ts=1675417876.753029&cid=CKRF0C8R3

It would be nice if those two helpers could be packaged in the inmanta-core package directly, so that we don't have to copy paste it in all the modules we want to use it in.

Hugo-Inmanta:
# Description

Add the HandlerLogger interface and its implementation LoggerWrapper into core

I used the code from the restbase module, only slightly changing the docstrings.

closes #5708 

# Self Check:

- [x] Changelog entry


src/inmanta/agent/handler.py
```diff
@@ -1067,3 +1067,83 @@ class HandlerNotAvailableException(Exception):
     This exception is thrown when a resource handler cannot perform its job. For example, the admin interface
     it connects to is not available.
     """
+
+
+class HandlerLogger(typing.Protocol):
```
sanderr, line 1072: I think that if we port this to core we need to integrate it more tightly:
- I see no harm in this remaining a `Protocol` but I think we should explicitly inherit from it for both the actual handler's logger and the `LoggerWrapper` to make sure we stay compatible with the interface.
- if we lift `log_msg` to this class we can actually implement all of the others. It may then be cleaner to make this an `ABC` instead of a protocol with only `log_msg` an `abstractmethod`. This would simplify both `LoggerWrapper` and `HandlerContext`.
- Depending on how [this discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1681809378693359) goes we may be able to solve the issues you encountered with `exc_info` by refactoring `log_msg` a bit.
- The docstrings are written from the perspective of the module that used this. They should instead be written from the perspective of core.
- The name `LoggerWrapper` isn't very clear in this context. Then again I don't immediately have a better idea. `LoggingLogger`, `PlainLogger`, `SimpleLogger`, `NativeLogger`, `PythonLogger` etc come to mind but I don't know if I really like any of them.

src/inmanta/agent/handler.py
```diff
@@ -1067,3 +1067,83 @@ class HandlerNotAvailableException(Exception):
...
+
+    def __init__(self, logger: logging.Logger) -> None:
+        self.logger = logger
+
+    def log_msg(
+        self,
+        level: int,
+        msg: str,
+        args: typing.Sequence[object],
+        kwargs: typing.Dict[str, object],
```
sanderr, line 1118:
```suggestion
        args: abc.Sequence[object],
        kwargs: abc.Mapping[str, object],
```

src/inmanta/agent/handler.py
sanderr: For the refactor, we could make it `_log_msg` here, then keep `HandlerContext.log_msg` for backwards compatibility to be on the safe side but add a deprecation warning. Once this is merged you can even immediately drop it from `master` (we diverged from iso6 last week).

sanderr (changes requested):

src/inmanta/agent/handler.py
sanderr: - Both of these classes should be marked as part of the stable API and included in the [docs](https://docs.inmanta.com/community/latest/model_developers/handlers.html#logging).

docs/model_developers/handlers.rst
```diff
@@ -125,6 +125,8 @@ For example:
         # ...
         ctx.debug("Creating loadbalancer with security group %(sg)s", sg=sg_id)
 
+In addition, the :class:`~inmanta.agent.handler.KwargsLogger` provides similar functionalities but without being
+restricted to an Agent context.
```
Hugo-Inmanta, line 129: This feel very lightweight as far as documentation goes, but I don't quite see how to make it better. The changelog message could also use some more care.

src/inmanta/agent/handler.py
Hugo-Inmanta: I went with `IBasicLogging` for the interface and `KwargsLogger` for the wrapper wdyt ?

sanderr: The `I<something>` convention is not something we really use (I think it's more Java than Python but I'm not sure). We generally go for something like `<something>ABC` (from Abstract Base Class) if we want to explicitly indicate it's only an interface. I wouldn't say either is better than the other a priori, but I would definitely aim for consistency.

That aside:
- For the interface I prefer `Logger` over `Logging` because it's methods are instance methods so instances of the class are loggers. I'm not sure if `Basic` is required, could be just `Logger` / `LoggerABC` but I don't really care too much about that.
- `KwargsLogger` I would focus less on the kwargs and more on what/where it logs. It's a logger that just bridges the interface with Python's built-in logging. Words like Python, native, simple, ... would describe this better.

docs/model_developers/handlers.rst
sanderr: I agree with the changelog message, I'll add a comment there. As to this being lightweight, yes but I also agree there's not much more to say. One thing I would make sure to mention is the interface. e.g.
```suggestion
This logger implements the `~inmanta.agent.handler.IBasicLogging` logging interface. An alternative implementation that just logs to the Python logger is provided in `~inmanta.agent.handler.KwargsLogger`. This logger is not meant to be used in actual handlers but it can be used for the automated testing of helper methods that accept a `IBasicLogging` instance. In production, these helpers would receive the actual `HandlerContext` and log appropriately, while for testing the `KwargsLogger` can be passed.
```

Just a quick draft, I'll leave it up to you to review and/or cherry-pick what you think appropriate.

src/inmanta/agent/handler.py
```diff
@@ -354,7 +394,10 @@ def update_changes(  # noqa: F811
     def changes(self) -> Dict[str, AttributeStateChange]:
         return self._changes
 
-    def log_msg(self, level: int, msg: str, args: Sequence[object], kwargs: Dict[str, object]) -> None:
+    def log_msg(self, level: int, msg: str, args: abc.Sequence[object], kwargs: abc.Mapping[str, object]) -> None:
+        LOGGER.warning(
```
sanderr, line 398: Are you planning to drop this from master after merging? Otherwise I'd add a comment as to why it's here and when it can be dropped.

src/inmanta/agent/handler.py
```diff
@@ -159,7 +160,46 @@ def bound(**kwds):
...
+        level: int,
+        msg: str,
+        *args,
+        **kwargs,
+    ) -> None:
+        raise NotImplementedError
+
+
+@stable_api
+class HandlerContext(IBasicLogging, object):
```
sanderr, line 202:
```suggestion
class HandlerContext(IBasicLogging):
```
Let's clean this up while we're at it. I think this was required in Python 3.5 or something so it hasn't been for a long time.

src/inmanta/agent/handler.py
```diff
@@ -1069,53 +1071,22 @@ class HandlerNotAvailableException(Exception):
...
-
-class LoggerWrapper:
+@stable_api
+class KwargsLogger(IBasicLogging):
     """
-    This class implements the HandlerLogger interface
-
-    It can be used in the tests of a module to be able to use the handler helpers, outside of the context of an agent.
+    This class implements the IBasicLogging interface and is a standalone wrapper
+    around a logging.Logger to facilitate logging of keyword arguments.
```
sanderr, line 1078: Here too I wouldn't focus on the kwargs aspect. The purpose is more to offer a simple standalone logger compatible with the handler one. Basically what you put in the rest of the docstring so that's good.

src/inmanta/agent/handler.py
Hugo-Inmanta: Yes :+1:

src/inmanta/agent/handler.py
```diff
@@ -394,43 +395,20 @@ def update_changes(  # noqa: F811
...
-            raise Exception("Args not supported")
+    def log_msg(self, level: int, msg: str, args: abc.Sequence[object], kwargs: abc.MutableMapping[str, object]) -> None:
+        LOGGER.warning("Direct calls to the log_msg method are being deprecated, please use the LoggerABC interface instead.")
         if "exc_info" in kwargs:
-            exc_info = kwargs["exc_info"]
+            exc_info = bool(kwargs["exc_info"])
             kwargs["traceback"] = traceback.format_exc()
         else:
             exc_info = False
+        self._log_msg(level, msg, *args, exc_info=exc_info, **kwargs)
```
sanderr, line 405: You're passing it twice now. I don't think it would cause issues but I think it's cleaner to just go with this.
```suggestion
        self._log_msg(level, msg, *args, **kwargs)
```

This also means you can get rid of the `exc_info` variable above, and the traceback part is also redundant because that is done in the `_log_msg` method.

sanderr (changes requested):
One last comment

src/inmanta/agent/handler.py
```diff
@@ -354,16 +395,17 @@ def update_changes(  # noqa: F811
     def changes(self) -> Dict[str, AttributeStateChange]:
         return self._changes
 
-    def log_msg(self, level: int, msg: str, args: Sequence[object], kwargs: Dict[str, object]) -> None:
+    def log_msg(self, level: int, msg: str, args: abc.Sequence[object], kwargs: abc.MutableMapping[str, object]) -> None:
+        LOGGER.warning("Direct calls to the log_msg method are being deprecated, please use the LoggerABC interface instead.")
+        self._log_msg(level, msg, *args, **kwargs)
```
Hugo-Inmanta, line 400: mypy now complains on this because its expecting arg4 to be a bool, is there a way around this ? This would go into iso6 only in the end

sanderr: I'd ignore it. It's the least of our typing errors and as you say it's not present on master. We could circumvent it but it would lead us back to more or less what you had before and in my opinion this is the cleaner approach.

sanderr (approved):

PR #5841, opened 2023-04-18
issue/367 add changelog for id_attribute called id 

FloLey:
# Description

Adds a changelog entry saying that in a future release it will not be possible anymore to use an id_attribute called id

closes https://github.com/inmanta/pytest-inmanta/issues/367

# Self Check:

- [x] Changelog entry


changelogs/unreleased/367-id-attribute-called-id.yml
```diff
@@ -0,0 +1,7 @@
+description: In a future release it will not be possible anymore to use an id_attribute called id
```
arnaudsjs, line 1: Just to add a bit more context.
```suggestion
description: In a future release it will not be possible anymore to use a resource with an id_attribute called id
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #5855, opened 2023-04-18
Fix bug in resource action cleanup job

arnaudsjs:
# Description

Fix bug where the cleanup job, that removes old resource actions, ignores the environment scope of the `resource_action_logs_retention` setting. This way the most restrictive value of the `resource_action_logs_retention` setting across all environments was applied on all environments.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/bugfix-resource-action-cleanup.yml
+++ b/changelogs/unreleased/bugfix-resource-action-cleanup.yml
@@ -1,6 +1,6 @@
 ---
 description: Fix bug where the cleanup job, that removes old resource actions, ignores the environment scope of the `resource_action_logs_retention` setting. This way the most restrictive value of the `resource_action_logs_retention` setting across all environments was applied on all environments.
 change-type: patch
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/bugfix-resource-action-cleanup.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: Fix bug where the cleanup job, that removes old resource actions, ignores the environment scope of the `resource_action_logs_retention` setting. This way the most restrictive value of the `resource_action_logs_retention` setting across all environments was applied on all environments.
```
sanderr, line 2: I find "restrictive" to be slightly ambiguous here. It could be interpreted as a restriction on "compiles should be kept for x hours" but I think you mean "compiles should be cleaned up after x hours". Perhaps "most restrictive value" -> "shortest interval"?

FloLey (approved):

sanderr (approved):

PR #5878, opened 2023-04-19
Issue/5842-dont-run-cleanup-jobs-halted-environment

### LINKED ISSUE #5842: Don't run cleanup jobs on halted environment
The server runs many cleanup jobs in the background. This issue should make sure that no cleanup jobs are run on halted environments.

Rationale: When a support archive is imported into a test orchestrator using the support tool, it first halts all environments to prevent any undesired side-effects. But all cleanup job remain active. When the achieve is old or the cleanup intervals are set short, it can happen that the orchestrator removes information (like resource actions logs) before it can be inspected by the person debugging the issue.

We probably want to halt all scheduled jobs (like for example the metrics collection job) and not only cleanup jobs?

FloLey:
# Description

The server runs many cleanup jobs in the background. This issue should make sure that no cleanup jobs are run on halted environments.

Closes #5842 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5727,7 +5727,6 @@ class Notification(BaseDocument):
 
     @classmethod
     async def clean_up_notifications(cls) -> None:
-        # TODO HALTED
         environments = await Environment.get_list()
         for env in environments:
             time_to_retain_logs = await env.get(NOTIFICATION_RETENTION)
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -3637,6 +3637,7 @@ class Compile(BaseDocument):
     async def delete_older_than(
         cls, oldest_retained_date: datetime.datetime, connection: Optional[asyncpg.Connection] = None
     ) -> None:
+        # todo add tests
         query = f"""
         WITH non_halted_envs AS (
           SELECT id FROM public.environment WHERE NOT halted
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5736,7 +5736,7 @@ class Notification(BaseDocument):
 
     @classmethod
     async def clean_up_notifications(cls) -> None:
-        default_retention_time = Environment._settings[NOTIFICATION_RETENTION].default
+        default_retention_time = str(Environment._settings[NOTIFICATION_RETENTION].default)
         LOGGER.info("Cleaning up notifications")
         query = f"""
                    WITH non_halted_envs AS (
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5736,7 +5736,7 @@ class Notification(BaseDocument):
 
     @classmethod
     async def clean_up_notifications(cls) -> None:
-        default_retention_time = str(Environment._settings["notification_retention"].default)
+        default_retention_time = Environment._settings[NOTIFICATION_RETENTION].default
         LOGGER.info("Cleaning up notifications")
         query = f"""
                    WITH non_halted_envs AS (
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4047,7 +4047,7 @@ class ResourceAction(BaseDocument):
 
     @classmethod
     async def purge_logs(cls) -> None:
-        default_retention_time = "7"
+        default_retention_time = str(Environment._settings["resource_action_logs_retention"].default)
 
         query = f"""
             WITH non_halted_envs AS (
@@ -5736,20 +5736,20 @@ class Notification(BaseDocument):
 
     @classmethod
     async def clean_up_notifications(cls) -> None:
-        # todo fix issue race condition
-        environments = await Environment.get_list(halted=False)
-        for env in environments:
-            time_to_retain_logs = await env.get(NOTIFICATION_RETENTION)
-            keep_notifications_until = datetime.datetime.now().astimezone() - datetime.timedelta(days=time_to_retain_logs)
-            LOGGER.info(
-                "Cleaning up notifications in environment %s that are older than %s", env.name, keep_notifications_until
-            )
-
-            query = f"""
-            DELETE FROM {cls.table_name()}
-            WHERE created < $1 AND environment = $2;
-            """
-            await cls._execute_query(query, cls._get_value(keep_notifications_until), cls._get_value(env.id))
+        default_retention_time = str(Environment._settings["notification_retention"].default)
+        LOGGER.info("Cleaning up notifications")
+        query = f"""
+                   WITH non_halted_envs AS (
+                       SELECT id, (COALESCE(settings->>'notification_retention', $1))::int AS retention_days
+                       FROM public.environment
+                       WHERE NOT halted
+                   )
+                   DELETE FROM {cls.table_name()}
+                   USING non_halted_envs
+                   WHERE environment = non_halted_envs.id
+                       AND created < now() AT TIME ZONE 'UTC' - (non_halted_envs.retention_days || ' days')::interval
+               """
+        await cls._execute_query(query, default_retention_time)
 
     def to_dto(self) -> m.Notification:
         return m.Notification(
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5735,7 +5735,18 @@ class Notification(BaseDocument):
             LOGGER.info(
                 "Cleaning up notifications in environment %s that are older than %s", env.name, keep_notifications_until
             )
-            query = f"DELETE FROM {cls.table_name()} WHERE created < $1 AND environment = $2"
+
+            query = f"""
+            WITH non_halted_envs AS (
+              SELECT id FROM public.environment WHERE NOT halted
+            )
+            DELETE FROM {cls.table_name()}
+            WHERE created < $1
+              AND environment = $2
+              AND environment IN (
+                SELECT id FROM non_halted_envs
+            );
+            """
             await cls._execute_query(query, cls._get_value(keep_notifications_until), cls._get_value(env.id))
 
     def to_dto(self) -> m.Notification:
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4058,7 +4058,7 @@ class ResourceAction(BaseDocument):
             DELETE FROM {cls.table_name()}
             USING non_halted_envs
             WHERE environment = non_halted_envs.id
-                AND started < now() AT TIME ZONE 'UTC' - (non_halted_envs.retention_days || ' days')::interval
+                AND started < now() AT TIME ZONE 'UTC' - make_interval(days => non_halted_envs.retention_days)
         """
         await cls._execute_query(query, default_retention_time)
 
@@ -5747,7 +5747,7 @@ class Notification(BaseDocument):
                    DELETE FROM {cls.table_name()}
                    USING non_halted_envs
                    WHERE environment = non_halted_envs.id
-                       AND created < now() AT TIME ZONE 'UTC' - (non_halted_envs.retention_days || ' days')::interval
+                       AND created < now() AT TIME ZONE 'UTC' - make_interval(days => non_halted_envs.retention_days)
                """
         await cls._execute_query(query, default_retention_time)
 
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -3635,8 +3635,15 @@ class Compile(BaseDocument):
     async def delete_older_than(
         cls, oldest_retained_date: datetime.datetime, connection: Optional[asyncpg.Connection] = None
     ) -> None:
-        # TODO HALTED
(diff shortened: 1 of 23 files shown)


src/inmanta/data/__init__.py
```diff
@@ -4026,7 +4039,17 @@ async def purge_logs(cls) -> None:
...
+            WITH non_halted_envs AS (
+              SELECT id FROM public.environment WHERE NOT halted
+            )
+            DELETE FROM {cls.table_name()}
+            WHERE environment = $1
+              AND started < $2
+              AND environment IN (
+                SELECT id FROM non_halted_envs
+            );
+            """
```
arnaudsjs, line 4052: It would be better to filter out the halted environment in the query that lists all environments.

```
environments = await Environment.get_list(halted=False)
```

src/inmanta/data/__init__.py
```diff
@@ -5711,7 +5734,18 @@ async def clean_up_notifications(cls) -> None:
             LOGGER.info(
                 "Cleaning up notifications in environment %s that are older than %s", env.name, keep_notifications_until
             )
-            query = f"DELETE FROM {cls.table_name()} WHERE created < $1 AND environment = $2"
+
```
arnaudsjs, line 5737: Same here.

arnaudsjs (changes requested):
The following jobs should be disabled as well:
* `OrchestrationService._purge_versions()`
* `EnvironmentMetricsService._cleanup_old_metrics()`

And I think this one as well, although it's not a cleanup job. I am going to ask feedback on this one from the team. 

* `ParameterService.renew_facts()`

arnaudsjs (changes requested):
The full compiles scheduled by the EnvironmentService should be disabled as well.

tests/server/test_environment_metrics.py
```diff
@@ -1406,7 +1406,8 @@ async def test_metric_aggregation_no_date(
     assert result.result["data"]["metrics"]["gauge_metric1"] == [None for _ in range(10)]
 
 
-async def test_cleanup_environment_metrics(init_dataclasses_and_load_schema) -> None:
+@pytest.mark.parametrize("env1_halted, env2_halted", [(True, True), (False, True), (True, False), (False, False)])
```
Hugo-Inmanta, line 1409: If you want to do all possible combinations you can do it this way:
```suggestion
@pytest.mark.parametrize("env1_halted", [True, False])
@pytest.mark.parametrize("env2_halted", [True, False])
```

Hugo-Inmanta (approved):

tests/server/test_environment_metrics.py
FloLey: ow that's easier :p

FloLey:
@arnaudsjs EnvironmentMetricsService._cleanup_old_metrics() was already done, I've added a test

tests/test_data.py
arnaudsjs, line 1: Test cases in this file are intended to test database queries. These test cases should not start an Inmanta server, but now many of them do by adding the `server` fixture to them. By starting a server many kinds of problems can occur. If you want to halt the environment, I would do that via a direct insert into the Environment table, but not via the API.

tests/server/test_environment_metrics.py
```diff
@@ -1406,7 +1406,9 @@ async def test_metric_aggregation_no_date(
     assert result.result["data"]["metrics"]["gauge_metric1"] == [None for _ in range(10)]
 
 
-async def test_cleanup_environment_metrics(init_dataclasses_and_load_schema) -> None:
+@pytest.mark.parametrize("env1_halted", [True, False])
+@pytest.mark.parametrize("env2_halted", [True, False])
+async def test_cleanup_environment_metrics(server, client, env1_halted, env2_halted) -> None:
```
arnaudsjs, line 1411: This test case didn't start a server before, but now it does by adding the `server` fixture. This means that next to the metrics created by this test case, the Inmanta server will create metrics as well. If that happens the test case will fail. This test case was not built to run while an Inmanta server is running and I think we should keep it like that.

arnaudsjs (changes requested):

src/inmanta/server/services/paramservice.py
```diff
@@ -93,16 +93,26 @@ async def renew_facts(self) -> None:
                 await self.agentmanager.request_parameter(param.environment, param.resource_id)
 
         unknown_parameters = await data.UnknownParameter.get_list(resolved=False)
+        environments = await data.Environment.get_list(halted=False)
```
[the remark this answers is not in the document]
FloLey: I forgot to do the other part. I remember thinking I should do do it.

tests/server/test_compilerservice.py
```diff
@@ -1065,6 +1065,58 @@ async def report_cleanup_finished_successfully() -> bool:
...
+    compiles = await data.Compile.get_list()
+    assert len(compiles) == 2
+
+    if halted:
+        result = await client.halt_environment(environment)
+        assert result.code == 200
+
+    oldest_retained_date = datetime.datetime.now().astimezone() - datetime.timedelta(seconds=50)
+
+    await data.Compile.delete_older_than(oldest_retained_date)
```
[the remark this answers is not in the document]
FloLey: here just moving 
```
    if halted:
        result = await client.halt_environment(environment)
        assert result.code == 200
```
this before adding the compiles is enough right?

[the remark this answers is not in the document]
arnaudsjs: Yes. First halt the environment to make sure that the server doesn't do cleanups anymore, then add the data. If you do it the other way around, the server might cleanup the data you add.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -394,7 +394,7 @@ async def _purge_versions(self) -> None:
         Purge versions from the database
         """
         # TODO: move to data and use queries for delete
-        envs = await data.Environment.get_list()
```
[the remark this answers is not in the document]
FloLey: indeed good catch. I will try to fix that

src/inmanta/data/__init__.py
```diff
@@ -4022,13 +4047,20 @@ async def save(self, connection: Optional[asyncpg.connection.Connection] = None)
...
     @classmethod
     async def purge_logs(cls) -> None:
-        environments = await Environment.get_list()
-        for env in environments:
-            time_to_retain_logs = await env.get(RESOURCE_ACTION_LOGS_RETENTION)
-            keep_logs_until = datetime.datetime.now().astimezone() - datetime.timedelta(days=time_to_retain_logs)
-            query = "DELETE FROM " + cls.table_name() + " WHERE environment=$1 AND started < $2"
-            value = cls._get_value(keep_logs_until)
-            await cls._execute_query(query, env.id, value)
+        default_retention_time = "7"
```
FloLey, line 4050: @arnaudsjs I think this should do what we want but I don't know how to get the "default_retention_time" form [here](https://gist.github.com/FloLey/4e9b1883441a8bda3929ec9a820c6300#file-__init__-py-L2252).
Should I create a new global variable that would be used at that location and in the purge_logs function or is there another way?

arnaudsjs: The `Environment` class in the database layer has an class variable `_settings` that contains all the settings and their default values. You can extract them directly from there:

```
Environment._settings[<name_environment_setting>].default
```
Does that make sense?

src/inmanta/data/__init__.py
```diff
@@ -5704,15 +5736,20 @@ class Notification(BaseDocument):
...
-        for env in environments:
-            time_to_retain_logs = await env.get(NOTIFICATION_RETENTION)
-            keep_notifications_until = datetime.datetime.now().astimezone() - datetime.timedelta(days=time_to_retain_logs)
-            LOGGER.info(
-                "Cleaning up notifications in environment %s that are older than %s", env.name, keep_notifications_until
-            )
-            query = f"DELETE FROM {cls.table_name()} WHERE created < $1 AND environment = $2"
-            await cls._execute_query(query, cls._get_value(keep_notifications_until), cls._get_value(env.id))
+        default_retention_time = str(Environment._settings["notification_retention"].default)
+        LOGGER.info("Cleaning up notifications")
```
FloLey, line 5740: I lost a bit of information in this logger but I don't think is it to bad. wdyt?

tests/server/test_notifications.py
```diff
@@ -35,22 +35,51 @@
...
-            created=created.astimezone(),
-            read=i in {2, 4},
-            cleared=i in {4, 5},
-        ).insert()
+async def add_notifications(server, environment: str):
+    """
+    A fixture that returns a method for adding notifications to the specified environment.
+    """
+
+    async def add_notifications_with_timestamp(days_old):
```
arnaudsjs, line 43: I don't see why we would wrap this method into a fixture.

tests/server/test_notifications.py
```diff
@@ -385,3 +414,22 @@ async def notification_cleaned_up(env_id: uuid.UUID, expected_length_after_clean
     # Only the latest one is kept
     assert len(short_retention_notifications) == 1
     assert short_retention_notifications[0].created == timestamps[2]
+
+
+@pytest.mark.parametrize("halted", [True, False])
+async def test_cleanup_notifications(server, postgresql_client, client, halted_env_with_old_notifications, halted):
```
arnaudsjs, line 420:
```suggestion
async def test_cleanup_notifications(server, client, halted_env_with_old_notifications, halted):
```

tests/server/test_environment_metrics.py
```diff
@@ -1406,7 +1406,11 @@ async def test_metric_aggregation_no_date(
     assert result.result["data"]["metrics"]["gauge_metric1"] == [None for _ in range(10)]
 
 
-async def test_cleanup_environment_metrics(init_dataclasses_and_load_schema) -> None:
+@pytest.mark.parametrize("env1_halted", [True, False])
+@pytest.mark.parametrize("env2_halted", [True, False])
+async def test_cleanup_environment_metrics(
+    init_dataclasses_and_load_schema, postgresql_client, env1_halted, env2_halted
```
arnaudsjs, line 1412:
```suggestion
    init_dataclasses_and_load_schema, env1_halted, env2_halted
```

arnaudsjs (changes requested):

tests/test_data.py
arnaudsjs: I think you forgot about this comment.

FloLey: Ow I didn't see it sorry. I will address it

src/inmanta/data/__init__.py
```diff
@@ -4022,13 +4047,20 @@ async def save(self, connection: Optional[asyncpg.connection.Connection] = None)
...
     @classmethod
     async def purge_logs(cls) -> None:
-        environments = await Environment.get_list()
-        for env in environments:
-            time_to_retain_logs = await env.get(RESOURCE_ACTION_LOGS_RETENTION)
-            keep_logs_until = datetime.datetime.now().astimezone() - datetime.timedelta(days=time_to_retain_logs)
-            query = "DELETE FROM " + cls.table_name() + " WHERE environment=$1 AND started < $2"
-            value = cls._get_value(keep_logs_until)
-            await cls._execute_query(query, env.id, value)
+        default_retention_time = str(Environment._settings["resource_action_logs_retention"].default)
```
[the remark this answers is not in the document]
FloLey: I first didn't do it and got :
`asyncpg.exceptions.DataError: invalid input for query argument $1: 7 (expected str, got int)`
But I don't see why it is needed

src/inmanta/server/services/paramservice.py
```diff
@@ -77,20 +77,30 @@ async def renew_facts(self) -> None:
...
+        env_ids = [env.id for env in environments]
+
         for param in params_to_renew:
             if param.environment is None:
                 LOGGER.warning(
                     "Found parameter without environment (%s for resource %s). Deleting it.", param.name, param.resource_id
                 )
                 await param.delete()
-            else:
+            if param.environment.environment in env_ids:
```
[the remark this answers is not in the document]
FloLey: I removed the first if statement as it seemed to be an impossible scenario.

tests/test_data.py
```diff
@@ -2072,13 +2092,15 @@ async def test_parameter(init_dataclasses_and_load_schema):
     updated_before = await data.Parameter.get_updated_before(datetime.datetime(2018, 7, 12, 12, 30))
     assert len(updated_before) == 0
     updated_before = await data.Parameter.get_updated_before(datetime.datetime(2018, 7, 14, 12, 30))
-    assert len(updated_before) == 1
-    assert (updated_before[0].environment, updated_before[0].name) == (parameters[2].environment, parameters[2].name)
+    assert len(updated_before) == (0 if halted else 1)
```
arnaudsjs, line 2095: By reading this test case I now realize that the name of the method (`get_updated_before`) doesn't make much sense anymore. We should give it a better name.

arnaudsjs (changes requested):

tests/test_data.py
arnaudsjs: ok. I would embed that info in the name of the method. But I am fine with a docstring as well.

arnaudsjs (approved):

PR #5883, opened 2023-04-20
Issue/5392 event loops in threads

### LINKED ISSUE #5389: Python 3.10 deprecated implicit event loop creation on main thread
On Python 3.9, when no current event loop is set, `asyncio.get_event_loop()` [creates a new one if it is running on the main thread](https://github.com/python/cpython/blob/595f9ccb0c059f2fb5bf13643bfc0cdd5b55a422/Lib/asyncio/events.py#L639).

For Python 3.10 this behavior is kept but [a deprecation warning was added](https://github.com/python/cpython/blob/1dd9be6584413fbfa823f37a224f101b819505d1/Lib/asyncio/events.py#L666).

For Python 3.12 it looks like [it will become an error](https://github.com/python/cpython/blob/3d5d3f7af6498effbc60a3db1d2b5f41ae6c0a75/Lib/asyncio/events.py#L676).

Once this is fixed, see if the `asyncio_thread_wrapper` can be removed from the lsm module's and core's tests. See inmanta/lsm#564 and #5392.

What does this mean for us? All our code that relies on implicit event loop creation through `asyncio.get_event_loop()` will have to be revisited. One thing that comes to mind is the implicit creation of tornado's `IOLoop` through `IOLoop.current()`. It's docstring now states explicitly that it should only be called when an event loop exists.

### LINKED ISSUE #5392: replace deprecated AnyThreadEventLoopPolicy
Similar to inmanta/lsm#564 for inmanta-core.

sanderr:
# Description

- Refactored bulk of the codebase to no longer depend on implicit event loop creation on the main thread because it has been deprecated. Migrated from `tornado` to `asyncio` event loops in those places
- Dropped `AnyThreadLoopEventPolicy` because with the change mentioned above all threads are treated equally, therefore it is no longer required.
- Fixed some race conditions in test cases that triggered consistenly with my changes.
- `inmanta.app` still uses `tornado` event loops because it wasn't as straightforward to factor it out there. I did make sure the implementation doesn't depend on implicit event loop creation on the main thread anymore.

Mypy diff (red is good)
```diff
- src/inmanta/protocol/endpoints.py:438: error: Returning Any from function declared to return "Result"  [no-any-return]
- src/inmanta/plugins.py:138: error: Module "tornado.ioloop" does not explicitly export attribute "TimeoutError"  [attr-defined]
- src/inmanta/plugins.py:141: error: Returning Any from function declared to return "T"  [no-any-return]
- src/inmanta/server/protocol.py:445: error: Incompatible types in assignment (expression has type "object", variable has type "None")  [assignment]
```

closes #5389 
closes #5392 
Related to inmanta/lsm#539, assumed to allow major simplification of [inmanta/lsm#564@cd9c3e97](https://github.com/inmanta/lsm/pull/564/files/cd9c3e972c8395c186e0362a695512f251be03b4): `AnyThreadLoopEventPolicy` can just be dropped without the need for the `asyncio_thread_wrapper`

# Self Check:

- [ ] Changelog entry


tests/test_agent_manager.py
```diff
@@ -1208,7 +1208,6 @@ async def test_are_agents_active(server, client, environment, agent_factory) ->
     Ensure that the `AgentManager.are_agents_active()` method returns True when an agent
     is in the up or the paused state.
     """
-    agent_config.use_autostart_agent_map.set("True")
```
sanderr, line 1211: Similar problem to the other test case but here the solution is simpler: we don't need to use the autostarted agent map.

tests/test_protocol.py
```diff
@@ -139,14 +132,7 @@ def do_test():
...
 
-    thread = threading.Thread(target=do_test)
-    thread.start()
-
-    while len(done) == 0 and limit > 0:
-        await gen.sleep(sleep)
-        limit -= 1
-
-    thread.join()
+    await asyncio.get_running_loop().run_in_executor(None, do_test)
```
sanderr, line 135: I think this is equivalent but please review this carefully.

src/inmanta/protocol/endpoints.py
```diff
@@ -427,21 +426,16 @@ def __init__(
...
+                    # no loop is running: create a loop for this thread if it doesn't exist already and run it
+                    return util.ensure_event_loop().run_until_complete(with_timeout)
                 else:
-                    # a specific IOloop is passed
-                    # we unwrap the tornado loop to get the native python loop
-                    # and safely tap into it using run_coroutine_threadsafe
-                    assert isinstance(self._ioloop, BaseAsyncIOLoop)  # make mypy happy
-                    return run_coroutine_threadsafe(method_call(), self._ioloop.asyncio_loop).result(self.timeout)
+                    # loop is running on different thread
+                    return run_coroutine_threadsafe(with_timeout, self._ioloop).result()
```
sanderr, line 438: I think this is equivalent but please review this carefully.

tests/test_protocol.py
```diff
@@ -139,14 +132,7 @@ def do_test():
...
 
-    thread = threading.Thread(target=do_test)
-    thread.start()
-
-    while len(done) == 0 and limit > 0:
-        await gen.sleep(sleep)
-        limit -= 1
-
-    thread.join()
+    await asyncio.wait_for(asyncio.get_running_loop().run_in_executor(None, do_test), timeout=1)
```
sanderr, line 135: I think this is equivalent but please review this carefully.

tests/conftest.py
```diff
@@ -990,12 +989,6 @@ def pytest_runtest_makereport(item, call):
             )
 
 
-async def off_main_thread(func):
```
wouterdb, line 993: This method exists elsewhere, but that may be copy paste, unsure.

tests/test_protocol.py
wouterdb: I agree

wouterdb (approved):
Looks very nice, but inherently tricky stuff

arnaudsjs (approved):

PR #5885, opened 2023-04-20
Issue/5816 inmanta module release output release tag

### LINKED ISSUE #5816: Make inmanta module release output release tag
When the `inmanta module release` command is executed, it doesn't produce any output. That way you have to execute a `git log` do figure out which Git tag was created. This information is required to push the tag later on using the `git push origin <tag>` command. It would be more user friendly if the release command would output the tag it created.

FloLey:
# Description

When the inmanta module release command is executed, it doesn't produce any output. That way you have to execute a git log to figure out which Git tag was created. This information is required to push the tag later on using the git push origin <tag> command. Now the release command will output the tag it created.

closes #5816 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5816-module-release-tag.yml
+++ b/changelogs/unreleased/5816-module-release-tag.yml
@@ -1,5 +1,5 @@
 description: The 'inmanta module release' command now outputs the release tag
-issue-nr: 58116
+issue-nr: 5816
 change-type: patch
 destination-branches: [master, iso6, iso5]
 sections:
--- a/changelogs/unreleased/5816-module-release-tag.yml
+++ b/changelogs/unreleased/5816-module-release-tag.yml
@@ -1,7 +1,7 @@
 description: The 'inmanta module release' command now outputs the release tag
 issue-nr: 5816
 change-type: patch
-destination-branches: [master, iso6, iso5]
+destination-branches: [master, iso6]
 sections:
   minor-improvement: "{{description}}"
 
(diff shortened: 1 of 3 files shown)


src/inmanta/moduletool.py
```diff
@@ -1336,6 +1336,7 @@ def release(
             gitprovider.tag(repo=module_dir, tag=str(release_tag))
             # bump to the next dev version
             self.release(dev=True, message="Bump version to next development version", patch=True)
+            print(f"Tag created successfully: {release_tag}")
```
arnaudsjs, line 1339: Why did you put this print here and not right after the `gitprovider.tag()` call? This can cause the output to be ordered in a strange way when we decide produce more output in the future.

arnaudsjs (changes requested):
Let's also verify this output log line in a test case to make sure that we don't break this functionality in the future.

changelogs/unreleased/5816-module-release-tag.yml
```diff
@@ -0,0 +1,7 @@
+description: The 'inmanta module release' command now outputs the release tag
+issue-nr: 5816
+change-type: patch
+destination-branches: [master, iso6, iso5]
```
arnaudsjs, line 4: The `inmanta module release` command doesn't exist on iso5, is it?

arnaudsjs (changes requested):

src/inmanta/moduletool.py
FloLey: I thought I was nice to have it at the end even if more things are printed to find it easily as we will need it. but indeed i didn't think of the exception scenario. I'll change it.

arnaudsjs (approved):

PR #5891, opened 2023-04-21
Bump compiler version to 2024.0

arnaudsjs:
# Description

Bump compiler version to 2024.0

Part of inmanta/lsm#592

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/__init__.py
+++ b/src/inmanta/__init__.py
@@ -16,7 +16,7 @@
     Contact: code@inmanta.com
 """
 
-COMPILER_VERSION = "2023.2"
+COMPILER_VERSION = "2023.3"
 RUNNING_TESTS = False
 """
     This is enabled/disabled by the test suite when tests are run.
--- a/src/inmanta/__init__.py
+++ b/src/inmanta/__init__.py
@@ -16,7 +16,7 @@
     Contact: code@inmanta.com
 """
 
-COMPILER_VERSION = "2023.3"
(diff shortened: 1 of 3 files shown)


Hugo-Inmanta (approved):

src/inmanta/__init__.py
```diff
@@ -16,7 +16,7 @@
     Contact: code@inmanta.com
 """
 
-COMPILER_VERSION = "2023.2"
+COMPILER_VERSION = "2023.3"
```
sanderr, line 19: I think we need to diverge the compiler versions now that we've diverged the branches. This is the first time we've diverged this early in the year, so our internal docs doesn't cover it yet. By far the easiest would be to already move to the 2024 range for OSS, but that may not be ideal. I'm not sure how else to move forward though.

For iso6 this version is definitely good.

sanderr (approved):
I'm adding Wouter in case he has an objection to moving to 2024 already.

wouterdb (approved):

sanderr:
@arnaudsjs we should also update the [internal docs](https://internal.inmanta.com/development/core/technical-designs/compiler_version.html) if this gets approved. Do you do it or shall I?

arnaudsjs:
> @arnaudsjs we should also update the [internal docs](https://internal.inmanta.com/development/core/technical-designs/compiler_version.html) if this gets approved. Do you do it or shall I?

I will do it.

PR #5900, opened 2023-04-24
Improve documentation of get_resource() endpoint

arnaudsjs:
# Description

Improve the documentation of the `api/v1/resource/<id>` endpoint and return a clear error message if the given id is not a valid resource version id.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/resourceservice.py
+++ b/src/inmanta/server/services/resourceservice.py
@@ -224,8 +224,8 @@ class ResourceService(protocol.ServerSlice):
         # Validate resource version id
         try:
             Id.parse_resource_version_id(resource_id)
-        except ValueError as exc:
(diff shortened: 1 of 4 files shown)


src/inmanta/server/services/resourceservice.py
```diff
@@ -221,6 +221,12 @@ async def get_resource(
         log_action: const.ResourceAction,
         log_limit: int,
     ) -> Apireturn:
+        # Validate resource version id
+        try:
+            Id.parse_resource_version_id(resource_id)
+        except ValueError as exc:
+            return 404, {"message": str(exc)}
```
wouterdb, line 228: What does this look like?

wouterdb (approved):

Hugo-Inmanta (approved):

PR #5901, opened 2023-04-24
Expose the logging setup through the stable api

### LINKED ISSUE #5815: Expose core logging setup through the stable api
As discussed here: https://inmanta.slack.com/archives/CKRF0C8R3/p1680678625072959

It would be nice to expose (at least some part) of the logging setup done by core through the stable api, to allow other python tools developed internally to reuse it.  Simplifying their development and making them more consistent with core.

The method which would be handy are:
```
inmanta.app._get_default_stream_handler()
inmanta.app._get_log_formatter_for_stream_handler(timed=True)
inmanta.app._convert_to_log_level(verbose)
```

FloLey:
# Description

Expose core logging setup through the stable api

closes #5815

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/logging.py
+++ b/src/inmanta/logging.py
@@ -18,7 +18,7 @@
 import logging
 import os
 import sys
-import typing
+from typing import Optional
 
 import colorlog
 from colorlog.formatter import LogColors
@@ -47,55 +47,36 @@ log_levels = {
 }
 
 
-class MultiLineFormatter(colorlog.ColoredFormatter):
-    """Multi-line formatter."""
-
-    def __init__(
-        self,
-        fmt: typing.Optional[str] = None,
-        *,
-        # keep interface minimal: only include fields we actually use
-        log_colors: typing.Optional[LogColors] = None,
-        reset: bool = True,
-        no_color: bool = False,
-    ):
-        super().__init__(fmt, log_colors=log_colors, reset=reset, no_color=no_color)
-        self.fmt = fmt
-
-    def get_header_length(self, record: logging.LogRecord) -> int:
-        """Get the header length of a given record."""
-        # to get the length of the header we want to get the header without the color codes
-        formatter = colorlog.ColoredFormatter(
-            fmt=self.fmt,
-            log_colors=self.log_colors,
-            reset=False,
-            no_color=True,
-        )
-        header = formatter.format(
-            logging.LogRecord(
-                name=record.name,
-                level=record.levelno,
-                pathname=record.pathname,
-                lineno=record.lineno,
-                msg="",
-                args=(),
-                exc_info=None,
-            )
-        )
-        return len(header)
-
-    def format(self, record: logging.LogRecord) -> str:
-        """Format a record with added indentation."""
-        indent: str = " " * self.get_header_length(record)
-        head, *tail = super().format(record).splitlines(True)
-        return head + "".join(indent + line for line in tail)
-
-
 class InmantaLogs:
-    _handler: logging.Handler
+    """
+    A class that provides logging functionality for Inmanta projects.
+
+    Usage:
+    To use this class, you first need to call the `setup_handler` method to configure the logging handler. This method
+    takes a `stream` argument that specifies where the log messages should be sent to. If no `stream` is provided,
+    the log messages will be sent to standard output.
+
+    You can then call the `apply_options` method to configure the logging options. This method takes an `options`
+    argument that should be an object with the following attributes:
+    - `log_file`: if this attribute is set, the logs will be written to the specified file instead of the stream
+      specified in `setup_handler`.
+    - `log_file_level`: the logging level for the file handler (if `log_file` is set).
+    - `verbose`: the verbosity level of the log messages.
+    - 'timed':if true,  adds the time to the formatter in the log lines.
+
+    This is not done in one step as we want logs for the cmd_parser, which will provide the options needed to configure
+    the 'final' logger with apply_options.
+    """
+
+    _handler: Optional[logging.Handler] = None
 
     @classmethod
-    def setup_handler(cls, stream=sys.stdout) -> logging.StreamHandler:
+    def create_default_handler(cls, stream: Optional[str] = sys.stdout) -> None:
+        """
+        Set up the logging handler for Inmanta.
+
+        :param stream: The stream to send log messages to. Default is standard output (sys.stdout).
+        """
         cls._handler = logging.StreamHandler(stream=stream)
         cls.set_log_level(logging.INFO)
         formatter = cls._get_log_formatter_for_stream_handler(timed=False)
@@ -106,7 +87,17 @@ class InmantaLogs:
         logging.root.setLevel(0)
 
     @classmethod
-    def apply_options(cls, options):
+    def apply_options(cls, options, stream: Optional[str] = sys.stdout) -> None:
+        """
+        Apply the logging options to the current handler. If no handler yet, first start on with setup_handler
+
+        :param options: the option object coming from the command line. This function use the following
+            attribute: log_file, log_file_level, verbose, timed
+        :param stream: The stream to send log messages to. Default is standard output (sys.stdout). This is only used
+            if a handler needs to be setup. If there is already a handler this param is ignored.
+        """
+        if cls._handler is None:
+            cls.create_default_handler(stream=stream)
         if options.log_file:
             cls.set_logfile_location(options.log_file)
             formatter = logging.Formatter(fmt="%(asctime)s %(levelname)-8s %(name)-10s %(message)s")
@@ -121,21 +112,45 @@ class InmantaLogs:
             cls.set_log_level(log_level)
 
     @classmethod
-    def set_log_level(cls, log_level: int):
+    def set_log_level(cls, log_level: int) -> None:
+        """
+        Set the logging level.
+
+        :param log_level: The logging level.
+        """
         cls._handler.setLevel(log_level)
 
     @classmethod
-    def set_log_formatter(cls, formatter: logging.Formatter):
+    def set_log_formatter(cls, formatter: logging.Formatter) -> None:
+        """
+        Set the log formatter.
+
+        :param formatter: The log formatter.
+        """
         cls._handler.setFormatter(formatter)
-        return
 
     @classmethod
-    def set_logfile_location(cls, location: str):
+    def set_logfile_location(cls, location: str) -> None:
+        """
+        Set the location of the log file. Be carefull taht this function will replace the current handler with a new one
+        This means that configurations done on the previous handler will be lost. It might be a good idea to
+        call this function first.
+
+        :param location: The location of the log file.
+        """
         file_handler = logging.handlers.WatchedFileHandler(filename=location, mode="a+")
         logging.root.removeHandler(cls._handler)
         cls._handler = file_handler
         logging.root.addHandler(cls._handler)
-        return
+
+    @classmethod
+    def get_handler(cls) -> logging.Handler:
+        """
+        Get the logging handler instance used by this class
+
+        :return: The logging handler instance used by this class
+        """
+        return cls._handler
 
     @classmethod
     def _convert_cli_log_level(cls, level: int) -> int:
@@ -175,3 +190,47 @@ class InmantaLogs:
                 no_color=True,
             )
         return formatter
+
+
+class MultiLineFormatter(colorlog.ColoredFormatter):
+    """Multi-line formatter."""
+
+    def __init__(
+        self,
+        fmt: Optional[str] = None,
+        *,
+        # keep interface minimal: only include fields we actually use
+        log_colors: Optional[LogColors] = None,
+        reset: bool = True,
+        no_color: bool = False,
+    ):
+        super().__init__(fmt, log_colors=log_colors, reset=reset, no_color=no_color)
+        self.fmt = fmt
+
+    def get_header_length(self, record: logging.LogRecord) -> int:
+        """Get the header length of a given record."""
+        # to get the length of the header we want to get the header without the color codes
+        formatter = colorlog.ColoredFormatter(
+            fmt=self.fmt,
+            log_colors=self.log_colors,
+            reset=False,
+            no_color=True,
+        )
+        header = formatter.format(
+            logging.LogRecord(
+                name=record.name,
+                level=record.levelno,
+                pathname=record.pathname,
+                lineno=record.lineno,
+                msg="",
+                args=(),
+                exc_info=None,
+            )
+        )
+        return len(header)
+
+    def format(self, record: logging.LogRecord) -> str:
+        """Format a record with added indentation."""
+        indent: str = " " * self.get_header_length(record)
+        head, *tail = super().format(record).splitlines(True)
+        return head + "".join(indent + line for line in tail)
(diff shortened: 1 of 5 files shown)


src/inmanta/logging.py
```diff
@@ -0,0 +1,245 @@
...
+        cls.set_log_level("INFO")
+        formatter = cls._get_log_formatter_for_stream_handler(timed=False)
+        cls.set_log_formatter(formatter)
+
+        logging.root.handlers = []
+        logging.root.addHandler(cls._handler)
+        logging.root.setLevel(0)
+
+    @classmethod
+    def apply_options(cls, options: object) -> None:
```
FloLey, line 91: @sanderr I don't know how to type options here. Could you help?

sanderr: Yes but it'll have to be for tomorrow I'm afraid, or perhaps even Friday.

FloLey: No issue :) Thanks!

tests/test_logging.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # test that with if no log_file is given, the stream will be used with the specified verbose option
+    # For verbose level 1, WARNINGs are shown INFOs not
+    options1 = Options(log_file=None, log_file_level="INFO", verbose="1")
+    InmantaLogs.apply_options(options1)
+    logger.info("info: This is the first test")
+    logger.warning("warning: This is the second test")
+    log_output = stream.getvalue().strip()
+    "INFO     test_logger info: This is the first test" not in log_output
+    "WARNING  test_logger warning: This is the second test" in log_output
```
Hugo-Inmanta, line 132: Did you mean 
```suggestion
    assert "INFO     test_logger info: This is the first test" not in log_output
    assert "WARNING  test_logger warning: This is the second test" in log_output
``` ?

**Review by Hugo-Inmanta — CHANGES_REQUESTED** [review]
The tickets suggests these methods to be added to the stable api:
```inmanta.app._get_default_stream_handler()
inmanta.app._get_log_formatter_for_stream_handler(timed=True)
inmanta.app._convert_to_log_level(verbose)```

I think only `_get_log_formatter_for_stream_handler` has been added ?
And I think they should be annotated with @stable_api decorator ?

File: src/inmanta/logging.py (hunk 1 above)
Comment (arnaudsjs) [reply src/inmanta/logging.py:95]: Why not make options into a separate object?

File: src/inmanta/logging.py [hunk 3]
```diff
@@ -0,0 +1,249 @@
...
+    def set_logfile_location(cls, location: str) -> None:
+        """
+        Set the location of the log file. Be careful that this function will replace the current handler with a new one
+        This means that configurations done on the previous handler will be lost.
+
+        :param location: The location of the log file.
+        """
+        file_handler = logging.handlers.WatchedFileHandler(filename=location, mode="a+")
+        if cls._handler:
+            logging.root.removeHandler(cls._handler)
```
arnaudsjs, line 175: Shouldn't we call `close()` on the old handler? Otherwise we don't close the file if the old handler would be a FileHandler for example.

src/inmanta/logging.py
```diff
@@ -0,0 +1,249 @@
...
+    "4": 2,
+    "ERROR": logging.ERROR,
+    "WARNING": logging.WARNING,
+    "INFO": logging.INFO,
+    "DEBUG": logging.DEBUG,
+    "TRACE": 2,
+}
+
+
+class InmantaLogs:
```
arnaudsjs, line 50: This class has a bit of an unconventional style, because all the methods are class methods. Normally a singleton contains a class method `get_instance()` that returns the only instance of the class. That way all other methods can be instance methods. This makes the structure of the class more closely related or non-singleton classes and it also makes it clear on the user side that you are applying state changes on the same object. Now, there is just a bunch of calls to static methods on the user side, which makes it harder to interpret what is going on.

arnaudsjs (changes requested):

src/inmanta/logging.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @classmethod
+    def set_logfile_location(cls, location: str) -> None:
+        """
+        Set the location of the log file. Be careful that this function will replace the current handler with a new one
+        This means that configurations done on the previous handler will be lost.
+
+        :param location: The location of the log file.
+        """
+        file_handler = logging.handlers.WatchedFileHandler(filename=location, mode="a+")
```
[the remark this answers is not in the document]
FloLey: As it was already used in the [code of core](https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/app.py#L796) before I don't think it is

src/inmanta/logging.py
```diff
@@ -0,0 +1,249 @@
...
+    To use this class, you first need to call the `setup_handler` method to configure the logging handler. This method
+    takes a `stream` argument that specifies where the log messages should be sent to. If no `stream` is provided,
+    the log messages will be sent to standard output.
+
+    You can then call the `apply_options` method to configure the logging options. This method takes an `options`
+    argument that should be an object with the following attributes:
+    - `log_file`: if this attribute is set, the logs will be written to the specified file instead of the stream
+      specified in `setup_handler`.
+    - `log_file_level`: the logging level for the file handler (if `log_file` is set).
+    - `verbose`: the verbosity level of the log messages.
```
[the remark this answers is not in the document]
FloLey: I'm not sure I understand what you mean. You'd like to have the set_log_level function but just used by us (via the CLI). So in that case I would rename it.
And than you would like a 'easier' function that set the loglevel directly and that takes a python log level as input? is that it?

FloLey:
> The tickets suggests these methods to be added to the stable api:
> 
> ```
> inmanta.app._get_default_stream_handler()
> inmanta.app._get_log_formatter_for_stream_handler(timed=True)
> inmanta.app._convert_to_log_level(verbose)
> ```
> 
> I think only `_get_log_formatter_for_stream_handler` has been added ? And I think they should be annotated with @stable_api decorator ?

src/inmanta/logging.py
```diff
@@ -0,0 +1,300 @@
...
+        :param options: the option object coming from the command line. This function use the following
+            attributes: log_file, log_file_level, verbose, timed
+        """
+        if not self._handler:
+            raise Exception("No handler to apply options to. Please use the get_instance method before calling this one")
+        if options.log_file:
+            self.set_logfile_location(options.log_file)
+            formatter = logging.Formatter(fmt="%(asctime)s %(levelname)-8s %(name)-10s %(message)s")
+            self.set_log_formatter(formatter)
+            self.set_log_level(options.log_file_level, cli=False)
```
arnaudsjs, line 141: `log_file_level` can be None, but `set_log_level()` accepts `str`?

src/inmanta/logging.py
```diff
@@ -0,0 +1,300 @@
...
+        :param cli: True if the logs will be outputed to the CLI.
+        """
+        if not self._handler:
+            raise Exception("No handler to apply options to. Please use the get_instance method before calling this one")
+        # maximum of 4 v's
+        if inmanta_log_level.isdigit() and int(inmanta_log_level) > 4:
+            inmanta_log_level = "4"
+
+        # The minimal log level on the CLI is always WARNING
+        if cli and inmanta_log_level == "ERROR" or (inmanta_log_level.isdigit() and int(inmanta_log_level) < 1):
```
arnaudsjs, line 174:
```suggestion
        if cli and (inmanta_log_level == "ERROR" or (inmanta_log_level.isdigit() and int(inmanta_log_level) < 1)):
```

tests/conftest.py
```diff
@@ -1763,3 +1764,9 @@ def disable_version_and_agent_cleanup_job():
     orchestrationservice.PERFORM_CLEANUP = False
     yield
     orchestrationservice.PERFORM_CLEANUP = old_perform_cleanup
+
+
+@pytest.fixture(scope="function")
+async def cleanup_log_instance():
+    InmantaLoggerConfig.clean_instance()
+    yield
```
arnaudsjs, line 1772: Shouldn't we call this from the `reset_all_objects()` method instead? Then it runs before and after each test case. I don't think it's a good idea to leave singleton in a dirty state across runs of test cases.

src/inmanta/logging.py
```diff
@@ -0,0 +1,300 @@
...
+
+    for more fine-grained configuration the following functions can be used as well:
+        - `set_log_level`
+        - `set_log_formatter`
+        - `set_logfile_location`
+    """
+
+    _instance: Optional["InmantaLoggerConfig"] = None
+
+    def __init__(self, stream: TextIO = sys.stdout) -> None:
```
arnaudsjs, line 88: Just a question: Do we have a use case for passing something else than `sys.stdout` here?

arnaudsjs (changes requested):

tests/test_logging.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # test that with if no log_file is given, the stream will be used with the specified verbose option
+    # For verbose level 4, WARNINGs and INFOs are shown
+    options2 = Options(log_file=None, log_file_level="INFO", verbose="4")
```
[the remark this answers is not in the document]
arnaudsjs: Yes. You are right. I must have overlooked that.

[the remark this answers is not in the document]
arnaudsjs: Yes. You are right.

src/inmanta/logging.py
```diff
@@ -0,0 +1,300 @@
...
+    @stable_api
+    def apply_options(self, options: Options) -> None:
+        """
+        Apply the logging options to the current handler. A handler should have been created before
+
+        :param options: the option object coming from the command line. This function use the following
+            attributes: log_file, log_file_level, verbose, timed
+        """
+        if not self._handler:
+            raise Exception("No handler to apply options to. Please use the get_instance method before calling this one")
```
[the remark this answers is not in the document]
FloLey: Indeed I will remove this.

tests/conftest.py
FloLey: It is better, I'll change it :)

src/inmanta/logging.py
FloLey: I saw here that Guillaume use stderr once:
https://code.inmanta.com/solutions/tools/yang-module-generator/-/blob/master/src/yang_module_generator/cli.py#L77-97

src/inmanta/logging.py
```diff
@@ -0,0 +1,300 @@
...
+    @stable_api
+    def get_instance(cls, stream: TextIO = sys.stdout) -> "InmantaLoggerConfig":
+        """
+        This method should be used to obtain an instance of this class, because this class is a singleton.
+
+        :param stream: The stream to send log messages to. Default is standard output (sys.stdout)
+        """
+        if not cls._instance:
+            cls._instance = cls(stream)
+        return cls._instance
```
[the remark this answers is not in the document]
FloLey: Indeed, I didn't think of that!

src/inmanta/logging.py
arnaudsjs: I see. Thanks!

src/inmanta/logging.py
FloLey: I fixed it by changing
`log_file_level: str = "INFO"
`
so it can't be None anymore. This is also the default value in the CLI

src/inmanta/logging.py
```diff
@@ -0,0 +1,323 @@
...
+            "3": logging.DEBUG,
+            "4": 2,
+            "ERROR": logging.ERROR,
+            "WARNING": logging.WARNING,
+            "INFO": logging.INFO,
+            "DEBUG": logging.DEBUG,
+            "TRACE": 2,
+        default is 'INFO'
+    - `verbose`: the verbosity level of the log messages. can be a number from 0 to 4.
+        if a bigger number is provided, 4 will be used. refer to log_file_level for the explanation of each level.
```
Hugo-Inmanta, line 71:
```suggestion
        if a bigger number is provided, 4 will be used. Refer to log_file_level for the explanation of each level.
```

src/inmanta/logging.py
```diff
@@ -0,0 +1,323 @@
...
+            if options.timed:
+                formatter = self._get_log_formatter_for_stream_handler(timed=True)
+                self.set_log_formatter(formatter)
+            self.set_log_level(str(options.verbose))
+
+    @stable_api
+    def set_log_level(self, inmanta_log_level: str, cli: bool = True) -> None:
+        """
+        Set the logging level. A handler should have been created before.
+        Below the supported Inmanta log levels and there equivalent and their equivalent Python log levels are shown:
```
Hugo-Inmanta, line 180:
```suggestion
        Below the supported Inmanta log levels and their equivalent Python log levels are shown:
```

tests/test_logging.py
```diff
@@ -0,0 +1,194 @@
...
+
+    if log_file:
+        log_file = tmpdir.join("test.log")
+
+    options1 = Options(log_file=log_file, log_file_level=log_file_level, verbose=verbose)
+    inmanta_logger.apply_options(options1)
+    logger.debug("debug: This is the first test")
+    logger.info("info: This is the second test")
+    logger.warning("warning: This is the third test")
+    logger.error("error: This is the forth test")
```
Hugo-Inmanta, line 134:
```suggestion
    logger.error("error: This is the fourth test")
```

tests/test_logging.py
```diff
@@ -0,0 +1,194 @@
...
+)
+def test_apply_options(tmpdir, log_file, log_file_level, verbose):
+    stream = StringIO()
+    inmanta_logger = InmantaLoggerConfig.get_instance(stream)
+    logger = logging.getLogger("test_logger")
+
+    if log_file:
+        log_file = tmpdir.join("test.log")
+
+    options1 = Options(log_file=log_file, log_file_level=log_file_level, verbose=verbose)
```
Hugo-Inmanta, line 129: I think the `verbose` parameter is typed as an `int` in the `Options` class (here a string is passed)

tests/test_logging.py
```diff
@@ -0,0 +1,194 @@
...
+        inmanta_logger.apply_options(options1)
+        options2 = Options(log_file=None, log_file_level="INFO", verbose="2")
+        inmanta_logger.apply_options(options2)
+    message = "Options can only be applied once to a handler."
+    assert message in str(e.value)
+
+
+def test_logging_cleaned_after_apply_options(tmpdir):
+    # verifies that when changing the stream with apply_option, the old stream is properly cleaned up
+    # and not user anymore.
```
Hugo-Inmanta, line 173:
```suggestion
    # and not used anymore.
```

Hugo-Inmanta (commented):
Some typos here and there, feel free to re-request review when the changes requested by Arnaud are in!

tests/test_logging.py
FloLey: indeed i'll change it

arnaudsjs (changes requested):

src/inmanta/logging.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        This method should be used to obtain an instance of this class, because this class is a singleton.
+
+        :param stream: The stream to send log messages to. Default is standard output (sys.stdout)
+        """
+        if cls._instance and stream != sys.stdout:
+            raise Exception("Instance already exists: cannot set the stream argument")
+        if not cls._instance:
+            cls._instance = cls(stream)
+        return cls._instance
```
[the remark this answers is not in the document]
FloLey: Now it should be fixed

arnaudsjs (approved):

PR #5917, opened 2023-04-26
Issue/5757 add f strings to the language

### LINKED ISSUE #5757: Add f-strings to the language
The goal is to add a python like f-string to the language. Currently we only have string interpolation. The current proposal is to add a new f-string that uses the same syntax as the python f-strings.

The python string.py module offers a parsers that returns both the variables and the formatters. This should make the implementation straightforward.

Open question: what do we do with the existing string interpolation.

Hugo-Inmanta:
# Description

Add support for f-strings to the language

closes #5757 

# Self Check:

- [x] Changelog entry


src/inmanta/ast/statements/assign.py
```diff
@@ -555,16 +555,36 @@ def __repr__(self) -> str:
...
+        return super().execute(requires, resolver, queue)
+
+    def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.NodeReference:
+        return dataflow.NodeStub("StringFormat.get_node() placeholder for %s" % self).reference()
+
+    def __repr__(self) -> str:
+        return "Format(%s)" % self._format_string
+
+
+class StringInterpolationFormat(FormattedString):
```
wouterdb, line 579:
```suggestion
class StringFormat(FormattedString):
```

I would not change the name if there is no strong reason. This might be used in code somewhere.

src/inmanta/ast/statements/assign.py
```diff
@@ -555,16 +555,36 @@ def __repr__(self) -> str:
...
 
-    def __init__(self, format_string: str, variables: typing.List[typing.Tuple["Reference", str]]) -> None:
-        ReferenceStatement.__init__(self, [k for (k, _) in variables])
+    def __init__(self, format_string: str, variables: Sequence["Reference"]) -> None:
+        super().__init__(variables)
         self._format_string = format_string
+
+    def execute(self, requires: typing.Dict[object, object], resolver: Resolver, queue: QueueScheduler) -> object:
+        return super().execute(requires, resolver, queue)
+
```
wouterdb, line 571:
```suggestion
```

I don't know if it adds any value to have execute here? (definitely not if there is no documentation on it)

src/inmanta/ast/statements/assign.py
```diff
@@ -581,8 +601,45 @@ def execute(self, requires: typing.Dict[object, object], resolver: Resolver, que
...
-    def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.NodeReference:
-        return dataflow.NodeStub("StringFormat.get_node() placeholder for %s" % self).reference()
 
-    def __repr__(self) -> str:
-        return "Format(%s)" % self._format_string
+class FStringFormatter(Formatter):
+    def __init__(self) -> None:
+        Formatter.__init__(self)
+
+    def get_field(self, key: str, args: Sequence[object], kwds: Mapping[str, object]) -> Tuple[object, str]:
```
wouterdb, line 609: Nice!

src/inmanta/ast/statements/assign.py
```diff
@@ -581,8 +601,45 @@ def execute(self, requires: typing.Dict[object, object], resolver: Resolver, que
...
+    __slots__ = ()
+
+    def __init__(self, format_string: str, variables: Sequence[typing.Tuple["Reference", str]]) -> None:
+        only_refs: Sequence["Reference"] = [k for (k, _) in variables]
+        super().__init__(format_string, only_refs)
+        self._variables = only_refs
+
+    def execute(self, requires: typing.Dict[object, object], resolver: Resolver, queue: QueueScheduler) -> object:
+        super().execute(requires, resolver, queue)
+        formatter: FStringFormatter = FStringFormatter()
```
wouterdb, line 631: Sadly the underlying formatter has no ability to cache the parsed string, so we have to re-parse every time. We might be able to squeeze a bit more performance out of it, but alas. 

Perhaps you could add a comment here to clarify why we don't cache the `formatter`

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -847,6 +855,36 @@ def p_constant_string(p: YaccProduction) -> None:
...
+    formatter = string.Formatter()
+
+    # formatter.parse returns an iterable of tuple (literal_text, field_name, format_spec, conversion)
+    parsed: Iterable[Tuple[str, Optional[str], Optional[str], Optional[str]]] = formatter.parse(str(p[1]))
+
+    start_lnr = p[1].location.lnr
+    start_char_pos = p[1].location.start_char + 2  # FSTRING tokens begin with `f"` or `f'` of length 2
+
+    locatable_matches: List[Tuple[str, LocatableString]] = []
+    for match in parsed:
```
wouterdb, line 869: I would add some testcase to check this line numbering logic is correct. 

(it is easy to test here, and very annoying in vscode if it is off)

cfr test_parser

wouterdb (changes requested):

src/inmanta/parser/plyInmantaParser.py
Hugo-Inmanta: Added a `test_fstring_numbering_logic` test

src/inmanta/ast/statements/assign.py
Hugo-Inmanta: I went with `        # We can't cache the formatter because it has no ability to cache the parsed string`, does this make sense ? I must admit it I don't fully get your point still.

arnaudsjs (changes requested):
We should document this feature on our documentation pages.

wouterdb (approved):

Hugo-Inmanta:
Added support for nested formatting e.g. `f"result: {arg:{width}.{precision}f}"` + some documentation

docs/language.rst
```diff
@@ -643,6 +643,33 @@ To prevent string interpolation, use raw strings
     motd = r"Welcome to {{hostname}}\n"
 
 
+String formatting
```
arnaudsjs, line 646: When this is compiled, we get errors like this. And the titles are not shown in the docs.

```
/home/arnaud/inmanta-core/docs/language.rst:623: CRITICAL: Title level inconsistent:

String interpolation
--------------------
/home/arnaud/inmanta-core/docs/language.rst:646: CRITICAL: Title level inconsistent:

String formatting
-----------------
/home/arnaud/inmanta-core/docs/language.rst:673: CRITICAL: Title level inconsistent:

Templates
---------
```

This is an issue that already existed before you added your changes, but I do think we need to fix this.

wouterdb (approved):

docs/language.rst
arnaudsjs, line 646: I recall from the demo meeting that our f-string support had some limitations with respect to the f-string support in Python. Is that true? If yes, we need to mention these limitations here.

arnaudsjs (changes requested):

docs/language.rst
Hugo-Inmanta: Added the limitation regarding `f"{arg=}"`

arnaudsjs (approved):

PR #5928, opened 2023-04-27
[FEAT] drop deprecated log_msg method

Hugo-Inmanta:
# Description

This is a follow-up to this ticket https://github.com/inmanta/inmanta-core/pull/5833, the log_msg method was only kept in iso6 for backwards compatibility.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/drop-deprecated-log_msg-method.yml
+++ b/changelogs/unreleased/drop-deprecated-log_msg-method.yml
@@ -1,0 +1,4 @@
+description: Drop deprecated log_msg method
+change-type: patch
+destination-branches: [master]
+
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/drop-deprecated-log_msg-method.yml
```diff
@@ -0,0 +1,5 @@
+description: Drop deprecated log_msg method in the handler.
+change-type: patch
+destination-branches: [master]
+sections:
+  deprecation-note: "{{description}}"
```
Hugo-Inmanta, line 5: On the fence about including a deprecation-note or not.

sanderr: Yes, definitely. Keep in mind this will only be released (disregarding OSS) in about half a year, so the deprecation warning will have been there for that amount of time, therefor this is an end-user visible change.

sanderr (approved):

PR #5948, opened 2023-05-03
Add support for compile request inside a database transaction

arnaudsjs:
# Description

This PR extends the API of the compiler service with functionality to request a compile inside a database transaction.

Part of inmanta/inmanta-lsm#1249

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
+++ b/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
@@ -1,0 +1,8 @@
+---
+description: Fixed race condition between the lifecycle state transfer and the associated compile request of a service instance. This bug can cause a service instance to get stuck in a certain state of its lifecycle. If you hit this bug, the issue can be resolved by executing a recompile.
+issue-nr: 1249
+issue-repo: inmanta-lsm
+change-type: patch
+destination-branches: [master, iso6, iso5, iso4]
+sections:
+  bugfix: "{{description}}"
--- a/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
+++ b/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
@@ -2,7 +2,7 @@
 description: Add support to the compilerservice to request a compile that is part of a database transaction.
 issue-nr: 1249
 issue-repo: inmanta-lsm
-change-type: patch
+change-type: minor
 destination-branches: [master, iso6]
 sections:
   minor-improvement: "{{description}}"
--- a/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
+++ b/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
@@ -3,6 +3,6 @@ description: Add support to the compilerservice to request a compile that is par
 issue-nr: 1249
 issue-repo: inmanta-lsm
 change-type: minor
-destination-branches: [master, iso6]
+destination-branches: [master]
 sections:
   minor-improvement: "{{description}}"
--- a/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
+++ b/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
@@ -3,6 +3,6 @@ description: Fixed race condition between the lifecycle state transfer and the a
 issue-nr: 1249
 issue-repo: inmanta-lsm
 change-type: patch
-destination-branches: [master, iso6, iso5, iso4]
+destination-branches: [master, iso6]
 sections:
   bugfix: "{{description}}"
--- a/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
+++ b/changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
@@ -1,8 +1,8 @@
 ---
-description: Fixed race condition between the lifecycle state transfer and the associated compile request of a service instance. This bug can cause a service instance to get stuck in a certain state of its lifecycle. If you hit this bug, the issue can be resolved by executing a recompile.
+description: Add support to the compilerservice to request a compile that is part of a database transaction.
 issue-nr: 1249
 issue-repo: inmanta-lsm
 change-type: patch
 destination-branches: [master, iso6]
 sections:
-  bugfix: "{{description}}"
+  minor-improvement: "{{description}}"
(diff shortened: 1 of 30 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -601,13 +602,104 @@ async def request_recompile(
...
+            metadata,
+            env_vars,
+            partial,
+            removed_resource_sets,
+            exporter_plugin,
+            notify_failed_compile,
+            failed_compile_message,
+        )
+
+    async def request_recompile_in_transaction(
```
sanderr, line 625: Why a separate method instead of an optional parameter?

src/inmanta/server/services/compilerservice.py
```diff
@@ -601,13 +602,104 @@ async def request_recompile(
...
+        connection: Optional[Connection] = None,
+    ) -> Tuple[Optional[uuid.UUID], Warnings]:
+        """
+        :param in_db_transaction: True iff the given connection is part of an ongoing database transaction.
+        :param connection: An optional database connection. A connection must be provided when in_db_transaction is True.
         """
+        if in_db_transaction and not connection:
+            raise Exception("A connection should be provided when in_db_transaction is True.")
+        if in_db_transaction and connection and not connection.is_in_transaction():
+            raise Exception("in_db_transaction is True, but the given connection is not executing a transaction.")
```
sanderr, line 702: Doesn't this make the `in_db_transaction` argument completely redundant? `in_db_transaction = connection is not None and connection.is_in_transaction()`?

src/inmanta/server/services/compilerservice.py
```diff
@@ -601,13 +602,104 @@ async def request_recompile(
...
+        partial: bool = False,
+        removed_resource_sets: Optional[List[str]] = None,
+        exporter_plugin: Optional[str] = None,
+        notify_failed_compile: Optional[bool] = None,
+        failed_compile_message: Optional[str] = None,
+    ) -> Tuple[Optional[uuid.UUID], Warnings]:
+        """
+        This method is similar to the `request_recompile()` method, but should be used when the compile request
+        is part of an ongoing database transaction. The given `connection` attribute should be within the scope
+        of that database transaction. It is required to call `CompileService.notify_compile_request_committed()`
```
sanderr, line 643: With this approach we do break compatibility between core-dev and inmanta-lsm-stable. I had not considered that before. The other approaches don't have that problem. Then again, when installed through the RPM that's never an issue.

src/inmanta/server/services/compilerservice.py
```diff
@@ -601,13 +602,104 @@ async def request_recompile(
...
+
+        This method is safe with respect to server restarts. If the server restarts after the transaction was committed,
+        but before this method was invoked, the server will automatically recover from this and run the requested compile
+        without any need to call this method.
+        """
+        async with self._queue_count_cache_lock:
+            self._queue_count_cache += 1
+        compile_obj: data.Compile = await data.Compile.get_by_id(compile_id)
+        if compile_obj.started:
+            raise Exception(f"Compile with id {compile_obj.id} was already started at {compile_obj.started}.")
```
sanderr, line 676: I don't think we should raise an exception here. This can happen when another compile finished more or less simultaneously with the transaction commit. Then this method call and the start of the compile would race with each other. Since both have the same purpose that's a success scenario.

src/inmanta/server/services/compilerservice.py
```diff
@@ -601,13 +602,104 @@ async def request_recompile(
...
+            partial,
+            removed_resource_sets,
+            exporter_plugin,
+            notify_failed_compile,
+            failed_compile_message,
+            connection=connection,
+            in_db_transaction=True,
+        )
+
+    async def notify_compile_request_committed(self, compile_id: uuid.UUID) -> None:
```
sanderr, line 662: I have no strict objection but I think we could get rid of the `compile_id` parameter if we wanted to. All that needs to be done here is to check whether there's a compile running and if not whether there's a compile queued. It's very similar to `_dequeue()`, the only difference being that that method has as a precondition that no compile is running but we could easily extract the common part if we wanted to.

changelogs/unreleased/1249-fix-race-condition-compile-requests.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: Add support to the compilerservice to request a compile that is part of a database transaction.
+issue-nr: 1249
+issue-repo: inmanta-lsm
+change-type: patch
```
sanderr, line 5: I would label this a minor, and bump the version now so we can depend on the right increment in inmanta-lsm.

sanderr (changes requested):

sanderr:
@wouterdb I'm adding you as reviewer just to confirm that the approach taken is acceptable. I don't think you need to review the implementation, just approve (/ reject) that we went for approach 1a described in #1249. While it is the simplest approach to take now it is the least future focused one and I seem to recall you had a strong opinion about that.

arnaudsjs:
> @wouterdb I'm adding you as reviewer just to confirm that the approach taken is acceptable. I don't think you need to review the implementation, just approve (/ reject) that we went for approach 1a described in #1249. While it is the simplest approach to take now it is the least future focused one and I seem to recall you had a strong opinion about that.

I discussed this with Wouter and it was decided not to do any premature optimizations here. So we went for the simplest approach that works.

src/inmanta/server/services/compilerservice.py
wouterdb: I wonder as well.

src/inmanta/server/services/compilerservice.py
wouterdb: I would not do that, because the additional argument comes with an additional obligation (call notify).

sanderr: Which would be a clear part of the contract, but you're right that it would be less explicit.

Hugo-Inmanta (changes requested):
I'm not 100% confident reviewing this part of the code, I'm requesting changes here so that I can see what it looks like in the end

src/inmanta/server/services/compilerservice.py
arnaudsjs: I wanted to make it sufficiently clear that this call would require the notify, but maybe the presence of the `in_db_transaction` argument is sufficient. I will change it.

src/inmanta/server/services/compilerservice.py
arnaudsjs: Then you want to replace `compile_id` by `environment_id`. Or would you execute that logic for every environment?

sanderr: `environment_id` indeed.

Hugo-Inmanta (approved):

PR #5972, opened 2023-05-11
Issue/enable web console

edvgui:
# Description

A client reported that the latest install of the rpm (iso6) was giving `404` when trying to reach the web console or the dashboard.  It looks like by default the web console is not reachable, because the default config doesn't enable it.  This MR fixes it.

# Self Check:

- [x] Changelog entry


arnaudsjs (changes requested):
The web-console section was removed in iso6 (see: https://docs.inmanta.com/inmanta-service-orchestrator/6/CHANGELOG.html#inmanta-ui-release-4-0-0-2022-11-30).

PR #5973, opened 2023-05-12
Fix heartbeat returning too fast

wouterdb:
# Description

Fix issue where heartbeat doesn't sleep 

I did not look through all other uses of .time()

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/agent_server/test_server_agent.py
+++ b/tests/agent_server/test_server_agent.py
@@ -310,7 +310,6 @@ async def test_spontaneous_deploy(
     """
     start = time.time()
     with caplog.at_level(DEBUG):
-
         resource_container.Provider.reset()
 
         env_id = environment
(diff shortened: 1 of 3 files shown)


sanderr (approved):

tests/agent_server/test_server_agent.py
```diff
@@ -302,76 +303,87 @@ async def test_server_restart(
...
+        assert resource_container.Provider.get("agent1", "key2") == "value2"
+        assert not resource_container.Provider.isset("agent1", "key3")
+
+    duration = time.time() - start
+
+    # approximate check, the number of heartbeats can vary, but not by a factor of 10
+    beats = [message for logger_name, log_level, message in caplog.record_tuples if "Received heartbeat from" in message]
+    assert (
+        len(beats) < duration / 10
+    ), f"Sent {len(beats)} heartbeats over a time period of {duration} seconds, sleap mechanism is broken"
```
arnaudsjs, line 386:
```suggestion
    ), f"Sent {len(beats)} heartbeats over a time period of {duration} seconds, sleep mechanism is broken"
```

arnaudsjs (changes requested):

PR #5987, opened 2023-05-16
fix-issue-timedelta

FloLey:
# Description

There seems to be an issue with the _set_timeout function of Tornado. If timeout is a timedelta with value zero it evaluates to false [here](https://github.com/tornadoweb/tornado/blob/master/tornado/queues.py#L62) which causes issues

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/protocol.py
+++ b/src/inmanta/server/protocol.py
@@ -502,11 +502,9 @@ class Session(object):
             if no_hang:
                 timeout = 0.1
             else:
-                timeout = self._interval
-
-            if timeout < 0.1:
-                timeout = 0.1
-
+                timeout = self._interval if self._interval > 0.1 else 0.1
+                # timeout should always been at least 0.1, this to work around an issue with Tornado:
+                # https://github.com/tornadoweb/tornado/issues/3271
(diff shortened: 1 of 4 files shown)


src/inmanta/server/protocol.py
```diff
@@ -504,7 +504,7 @@ async def get_calls(self, no_hang: bool) -> Optional[List[common.Request]]:
             else:
                 timeout = self._interval
 
-            call = await self._queue.get(timeout=timedelta(seconds=timeout))
+            call = await self._queue.get(timeout=timedelta(seconds=timeout, microseconds=1))
```
arnaudsjs, line 507: * We must add a clear comment regarding why adding one microsecond is required. I created an issue for this at Tornado, so you can also reference https://github.com/tornadoweb/tornado/issues/3271 in the comment. That way we know when we can remove the workaround.
* I would only add one microsecond when `timeout==0`. That way we keep the old behavior in production and only use the workaround when a specific test case requires it.

arnaudsjs (changes requested):

wouterdb (approved):

PR #5996, opened 2023-05-17
issue/5976-use-pip-config-file

### LINKED ISSUE #5976: improved support for pip config
`trusted-host`, credentials, pip index url through config, ... Which do we want to support, which do we want to avoid (how stand-alone do we want the project to be, how dependent on environment?

FloLey:
# Description

improved support for pip config:

1. add project option to use pip config from system -> pip: { use_config_file: bool }
2. if set, no "package" source is required for v2 to work
3. if set together with a "package" source the "package" source become extras
4. if not set, keep current behavior: override relevant pip env vars.

related pr: https://github.com/inmanta/inmanta-project-template/pull/14

closes #5976 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5976-use-pip--config-file.yml
+++ b/changelogs/unreleased/5976-use-pip--config-file.yml
@@ -1,5 +1,4 @@
-description: improved support for pip config files by adding use_config_file to project.yml.
-
+description: improved support for pip config files by adding the use_config_file option to project.yml. If set the pip config specified in the PIP_CONFIG_FILE env var will be used.
 issue-nr: 5976
 change-type: minor
 destination-branches: [master, iso6, iso5]
--- a/changelogs/unreleased/5976-use-pip--config-file.yml
+++ b/changelogs/unreleased/5976-use-pip--config-file.yml
@@ -1,0 +1,7 @@
+description: improved support for pip config by adding use_config_file to project.yml
+issue-nr: 5976
+change-type: minor
+destination-branches: [master, iso6, iso5]
+sections:
+  minor-improvement: "{{description}}"
+
--- a/changelogs/unreleased/5976-use-pip--config-file.yml
+++ b/changelogs/unreleased/5976-use-pip--config-file.yml
@@ -1,4 +1,4 @@
-description: improved support for pip config files by adding the use_config_file option to project.yml. If set the pip config specified in the PIP_CONFIG_FILE env var will be used.
+description: improved support for pip config files by adding the use_config_file option to project.yml. If set the pip config file will be used.
 issue-nr: 5976
 change-type: minor
 destination-branches: [master, iso6, iso5]
--- a/changelogs/unreleased/5976-use-pip--config-file.yml
+++ b/changelogs/unreleased/5976-use-pip--config-file.yml
@@ -1,4 +1,5 @@
-description: improved support for pip config by adding use_config_file to project.yml
+description: improved support for pip config files by adding use_config_file to project.yml.
+
 issue-nr: 5976
 change-type: minor
 destination-branches: [master, iso6, iso5]
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -345,7 +345,7 @@ class PipCommandBuilder:
         :param allow_pre_releases: Allow the installation of packages with pre-releases and development versions.
         :param constraints_files: Files that should be passed to pip using the `-c` option.
         :param requirements_files: Files that should be passed to pip using the `-r` option.
-        :param use_pip_config: whether the pip config should be used
+        :param use_pip_config: Whether the pip config file specified in the PIP_CONFIG_FILE env var should be used
         """
         requirements = requirements if requirements is not None else []
         paths = paths if paths is not None else []
@@ -893,7 +893,7 @@ class ActiveEnv(PythonEnvironment):
         :param requirements_list: List of requirement strings to install.
         :param upgrade: Upgrade requirements to the latest compatible version.
         :param upgrade_strategy: The upgrade strategy to use for requirements' dependencies.
-        :param use_pip_config: whether the pip config should be used
+        :param use_pip_config: Whether the pip config file specified in the PIP_CONFIG_FILE env var should be used
         """
         if not upgrade and self.are_installed(requirements_list):
             # don't fork subprocess if requirements are already met
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -615,7 +615,6 @@ class PythonEnvironment:
         self,
         paths: List[LocalPackagePath],
         constraint_files: Optional[List[str]] = None,
-        use_pip_config: bool = False,
     ) -> None:
         """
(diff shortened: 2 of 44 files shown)


src/inmanta/env.py
```diff
@@ -353,13 +355,29 @@ def compose_install_command(
...
+        index_args: List[str] = []
+
+        if use_pip_config:
+            if index_urls:
+                # Use only --extra-index-url arguments
+                for url in index_urls:
+                    index_args.append("--extra-index-url")
+                    index_args.append(url)
+            else:
+                index_args = []
```
Hugo-Inmanta, line 367: This can be removed I think
```suggestion
```

src/inmanta/module.py
```diff
@@ -712,7 +712,7 @@ def get_namespace_package_name(cls, module_name: str) -> str:
 
     def install(self, project: "Project", module_spec: List[InmantaModuleRequirement]) -> Optional["ModuleV2"]:
         module_name: str = self._get_module_name(module_spec)
-        if not self.urls:
+        if not self.urls and not project.metadata.pip.use_config_file:
```
Hugo-Inmanta, line 715: We could add something to the exception message like "or set the use_config_file project option to true to use the pip config file defined in the PIP_CONFIG_FILE env variable." ?

Hugo-Inmanta (changes requested):

tests/moduletool/test_install.py
```diff
@@ -893,6 +911,126 @@ def test_install_from_index_dont_leak_pip_index(
...
+            for metadata in [v2mod1, v2mod2]
+        ],
+    )
+
+    # install project
+    project_path = module.Project.get().path
+    os.chdir(project_path)
+    with caplog.at_level(logging.DEBUG):
+        ProjectTool().execute("install", [])
+    assert f"--extra-index-url {index2.url}" in caplog.text
```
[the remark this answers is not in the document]
FloLey: I'll add it :)

changelogs/unreleased/5976-use-pip--config-file.yml
```diff
@@ -0,0 +1,7 @@
+description: improved support for pip config by adding use_config_file to project.yml
```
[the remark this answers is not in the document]
FloLey: Isn't the point of release notes to briefly mention a change so that user know what to look for in the docs? I feel other descriptions aren't more descriptive than this one. Or do you think it is not clear?

src/inmanta/module.py
FloLey: good idea!

changelogs/unreleased/5976-use-pip--config-file.yml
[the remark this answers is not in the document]
Hugo-Inmanta: > Isn't the point of release notes to briefly mention a change so that user know what to look for in the docs?

I agree but then we should add more documentation about this change e.g. mention the fact that we can now use the config file defined in PIP_CONFIG_FILE env when setting use_config_file to true. 

> Or do you think it is not clear? 

I think mentioning the PIP_CONFIG_FILE here as well wouldn't hurt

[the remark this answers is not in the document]
FloLey: concerning the docs:
this is done here: https://github.com/inmanta/inmanta-core/pull/5996/files#diff-b5d17e866e545b1dc547769774bc95987546ecbbb5c7decb1035c4e6f16299acR22

which will show: https://github.com/inmanta/inmanta-core/pull/5996/files#diff-69cc23dbeac56bc1a4d681d8faf95f7da4dfc3587347c54d39c068e805b238a4R348

I'll try to add the PIP_CONFIG_FILE to the note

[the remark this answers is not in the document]
Hugo-Inmanta: > which will show: https://github.com/inmanta/inmanta-core/pull/5996/files#diff-69cc23dbeac56bc1a4d681d8faf95f7da4dfc3587347c54d39c068e805b238a4R348

Aah I completely missed the fact that this would be shown. Adding PIP_CONFIG_FILE to the note will make it easy to find :+1:

tests/moduletool/test_repos.py
```diff
@@ -27,6 +27,7 @@ def test_file_co(git_modules_dir, modules_repo):
     result = """name: mod6
 license: Apache 2.0
 version: '3.2'
+pip: {use_config_file: False}
```
sanderr, line 30: Same question here.

tests/moduletool/test_install.py
```diff
@@ -893,6 +911,127 @@ def test_install_from_index_dont_leak_pip_index(
...
+        new_name="v2mod1",
+        publish_index=index,
+    )
+
+    pip_config_file = os.path.join(tmpdir, "pip.conf")
+    with open(pip_config_file, "w+", encoding="utf-8") as f:
+        f.write("[global]\n")
+        f.write("timeout = 60\n")
+        f.write("index-url = file://" + index.url + "\n")
+    monkeypatch.setenv("PIP_CONFIG_FILE", pip_config_file)
```
sanderr, line 943: I'm still missing a test case that doesn't have `PIP_CONFIG_FILE` set (monkeypatch delete it from the environment). This test ideally shouldn't try to install a non-existing module though because that could open us op to dependency confusion attacks.

FloLey: You mean if PIP_CONFIG_FILE is not set it will use the default and will try to get the module from pypi and so I should use an existing module like inmanta-module-dummy-module?

FloLey: I tried it like this:
```
def test_install_with_use_config_but_PIP_CONFIG_FILE_not_set(    tmpvenv_active_inherit: env.VirtualEnv,
    modules_v2_dir: str,
    snippetcompiler_clean,
    monkeypatch,
    caplog,
) -> None:
    monkeypatch.delenv("PIP_CONFIG_FILE", False)

    # set up project
    snippetcompiler_clean.setup_for_snippet(
        """
        import dummy_module
        """,
        autostd=False,
        install_project=False,
        use_pip_config_file=True,
    )

    # install project
    project_path = module.Project.get().path
    os.chdir(project_path)
    with caplog.at_level(logging.DEBUG):
        ProjectTool().execute("install", [])
    print(caplog.text)
    assert tmpvenv_active_inherit.are_installed(requirements=["inmanta-module-dummy-module"])
```

but the dummy gets installed as a V1 module. I Don't know how to fix this. Is it something wrong with my setup_for_snippet or should I change `repo: https://github.com/inmanta/` in `project.yml` to something else?

sanderr: You forgot to add it to the `python_requires`. For the other tests you did do so.

FloLey: indeed, sorry for that I don't know how I missed it

sanderr: No problem, it happens!

sanderr (approved):

Hugo-Inmanta (approved):

PR #6057, opened 2023-05-26
Make agent backoff configurable

arnaudsjs:
# Description

This PR adds a configuration option for the agent backoff time. I will create a follow up PR to change the default value on  master only.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/config.py
+++ b/src/inmanta/agent/config.py
@@ -132,6 +132,17 @@ Each subsequent repair deployment will start agent-repair-interval seconds after
     is_time,
 )
 
+agent_get_resource_backoff = Option(
+    "config",
+    "agent-get-resource-backoff",
+    3,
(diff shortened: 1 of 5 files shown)


src/inmanta/agent/config.py
```diff
@@ -132,6 +132,17 @@
     is_time,
 )
 
+agent_get_resource_backoff = Option(
+    "config",
+    "agent-get-resource-backoff",
+    5,
```
wouterdb, line 138: are we going to change the default on master?

wouterdb (approved):

src/inmanta/agent/config.py
arnaudsjs: Yes. I will create a separate PR for that as we don't want to change the default on the iso6 branch. This one goes into `master` and `iso6`.

Issue #5933, opened 2023-04-28
investigate timeout on license component status


sanderr:
There was a timeout on the server status page of the management orchestrator for the license component. The log contained this trace:

```
2023-04-27 08:43:18,213 ERROR    inmanta.server.server The following error occured while trying to determine the status of slice license.license
Traceback (most recent call last):
  File "src/inmanta_license/license.py", line 806, in get_status
  File "src/inmanta_license/license.py", line 788, in get_license_metrics
  File "/opt/inmanta/lib64/python3.9/site-packages/inmanta/data/__init__.py", line 4534, in get_resources_report
    async for record in con.cursor(query, *values):
  File "/opt/inmanta/lib64/python3.9/site-packages/asyncpg/cursor.py", line 223, in __anext__
    buffer = await self._bind_exec(self._prefetch, self._timeout)
  File "/opt/inmanta/lib64/python3.9/site-packages/asyncpg/cursor.py", line 129, in _bind_exec
    buffer, _, self._exhausted = await protocol.bind_execute(
  File "asyncpg/protocol/protocol.pyx", line 201, in bind_execute
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.9/asyncio/tasks.py", line 490, in wait_for
    return fut.result()
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/inmanta/lib64/python3.9/site-packages/inmanta/server/server.py", line 134, in collect_for_slice
    return SliceStatus(name=slice_name, status=await asyncio.wait_for(slice.get_status(), 0.1))
  File "/usr/lib64/python3.9/asyncio/tasks.py", line 492, in wait_for
    raise exceptions.TimeoutError() from exc
asyncio.exceptions.TimeoutError
```

The timeout should be investigated and appropriate action taken.

sanderr:
Discussed in planning meeting at 2023-05-02: probably a badly constructed query.

PR #6094, opened 2023-06-06
[FEAT] add get_resource_type_count_for_latest_version query

Hugo-Inmanta:
# Description
The `get_license_metrics` was calling `get_resources_report` but only using it to get the resource_type count. This PR adds a query to retrieve only this.

TODO: 

- [x] use this newly added get_resource_type_count_for_latest_version query instead of get_resources_report in inmanta-license
-> PR [here](https://github.com/inmanta/inmanta-license/pull/516)
- [x] [bump inmanta core 1 minor up](https://github.com/inmanta/inmanta-core/pull/6094#issuecomment-1580000173)

closes https://github.com/inmanta/inmanta-core/issues/5933

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/5933-investigate-timeout-on-license-component.yml
+++ b/changelogs/unreleased/5933-investigate-timeout-on-license-component.yml
@@ -1,4 +1,4 @@
-description: Add get_latest_resources_resource_type_count to get the count for each resource_type over all resources in their latest version.
+description: Add get_latest_resources_resource_type_count to get the count for each resource_type over all resources in the model's latest version.
 issue-nr: 5933
 change-type: patch
 destination-branches: [master, iso6]
--- a/changelogs/unreleased/5933-investigate-timeout-on-license-component.yml
+++ b/changelogs/unreleased/5933-investigate-timeout-on-license-component.yml
@@ -1,4 +1,4 @@
 description: Add get_latest_resources_resource_type_count to get the count for each resource_type over all resources in the model's latest version.
 issue-nr: 5933
-change-type: patch
+change-type: minor
 destination-branches: [master, iso6]
--- a/changelogs/unreleased/5933-investigate-timeout-on-license-component.yml
+++ b/changelogs/unreleased/5933-investigate-timeout-on-license-component.yml
@@ -1,4 +1,4 @@
-description: Add get_latest_resources_resource_type_count to get the count for each resource_type over all resources in the model's latest version.
+description: Add get_resource_type_count_for_latest_version to get the count for each resource_type over all resources in the model's latest version.
 issue-nr: 5933
 change-type: minor
 destination-branches: [master, iso6]
--- a/changelogs/unreleased/5933-investigate-timeout-on-license-component.yml
+++ b/changelogs/unreleased/5933-investigate-timeout-on-license-component.yml
@@ -1,0 +1,4 @@
+description: Add get_latest_resources_resource_type_count to get the count for each resource_type over all resources in their latest version.
+issue-nr: 5933
+change-type: patch
+destination-branches: [master, iso6]
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4536,10 +4536,11 @@ class Resource(BaseDocument):
             GROUP BY resource_type;
         """
         values = [cls._get_value(environment)]
-        result: JsonType = {}
+        result: dict[str, int] = {}
         async with cls.get_connection() as con:
             async with con.transaction():
                 async for record in con.cursor(query, *values):
+                    assert isinstance(record["count"], int)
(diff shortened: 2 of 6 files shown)


arnaudsjs (changes requested):
I am setting this to "request changes", so that I get notified when this PR is ready for review again.

src/inmanta/data/__init__.py
```diff
@@ -4519,6 +4519,34 @@ async def get_resources_in_latest_version(
                         result.append(resource)
         return result
 
+    @classmethod
+    async def get_latest_resources_resource_type_count(cls, environment: uuid.UUID) -> JsonType:
```
arnaudsjs, line 4523: I find the name a bit confusing. It starts with `get_latest_resources`, which makes me believe that this method will return resources, but in fact it returns information about resource types.
```suggestion
    async def get_resource_type_count_for_latest_version(cls, environment: uuid.UUID) -> JsonType:
```

src/inmanta/data/__init__.py
```diff
@@ -4519,6 +4519,34 @@ async def get_resources_in_latest_version(
...
+        """
+        query = f"""
+            SELECT resource_type, count(*) as count
+            FROM (
+                SELECT DISTINCT resource_id, resource_type
+                FROM {Resource.table_name()}
+                WHERE environment=$1
+                AND model=({query_latest_model})
+            ) AS r1
+            GROUP BY resource_type;
```
arnaudsjs, line 4540: This query is more complex than it needs to be. It's important to realize that resource_ids are unique within a given model version of a given environment. So there is no need for the DISTINCT here.

```suggestion
            SELECT resource_type, count(*) as count
            FROM {Resource.table_name()}
            WHERE environment=$1 AND model=({query_latest_model})
            GROUP BY resource_type;
```

src/inmanta/data/__init__.py
```diff
@@ -4519,6 +4519,34 @@ async def get_resources_in_latest_version(
                         result.append(resource)
         return result
 
+    @classmethod
+    async def get_latest_resources_resource_type_count(cls, environment: uuid.UUID) -> JsonType:
+        """
+        Returns the count for each resource_type over all resources in the model's latest version
+        """
+        query_latest_model = f"""
```
arnaudsjs, line 4527: Nice! The new query executes in +-5ms, while the old one took >100ms

src/inmanta/data/__init__.py
arnaudsjs, line 4523: JsonType is a very generic type. It matches `Dict[str, Any]`. We can make this more specific `dict[str, int]` for this method.

```suggestion
    async def get_latest_resources_resource_type_count(cls, environment: uuid.UUID) -> dict[str, int]:
```

changelogs/unreleased/5933-investigate-timeout-on-license-component.yml
```diff
@@ -0,0 +1,4 @@
+description: Add get_latest_resources_resource_type_count to get the count for each resource_type over all resources in the model's latest version.
+issue-nr: 5933
+change-type: patch
```
arnaudsjs, line 3: This PR adds new functionality to inmanta-core, so it should be minor instead.

```suggestion
change-type: minor
```

arnaudsjs (changes requested):

arnaudsjs:
We also need to bump the version of inmanta-core one minor up. So that inmanta-license has the correct version constraint.

src/inmanta/data/__init__.py
```diff
@@ -4519,6 +4519,30 @@ async def get_resources_in_latest_version(
...
+            WHERE environment=$1
+        """
+        query = f"""
+            SELECT resource_type, count(*) as count
+            FROM {Resource.table_name()}
+            WHERE environment=$1 AND model=({query_latest_model})
+            GROUP BY resource_type;
+        """
+        values = [cls._get_value(environment)]
+        result: JsonType = {}
```
arnaudsjs, line 4539: Did you run mypy?

```suggestion
        result: dict[str, int] = {}
```

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
Hugo-Inmanta: That's weird, this wasn't caught by mypy. When I change this to `dict[str, int]` it does complain that `record["count"]` is an object and not an int though

arnaudsjs: Ah. True. That's because `JsonType` matches `Dict[str, Any]`. The use of `Any` basically switches of type validation.

Hugo-Inmanta: I see, didn't know that !

arnaudsjs (approved):

PR #6095, opened 2023-06-08
Issue/5884 improve logging regarding resource sets

### LINKED ISSUE #5884: Improve logging regarding resource sets
Tasks:

- [x] Log in the exporter which resource belongs to which resource set. Probably already fixed by now
- [ ] [This error message](https://github.com/inmanta/inmanta-core/blob/1caf8916161a53f7e493fa356bdfe3203ec2428b/src/inmanta/server/services/orchestrationservice.py#L751) should be improve. The error message only says that the resources moved to a different resource set, but it doesn't say from which resource set to which other resource set each resource moved.

Hugo-Inmanta:
# Description

 Improve error reporting when attempting to move a resource to another resource set in a partial compile. 

closes #5884 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4863,7 +4863,7 @@ class Resource(BaseDocument):
                 destination_version,
                 updated_resource_sets | deleted_resource_sets,
             )
-            return {record["resource_id"]: record["resource_set"] for record in result}
+            return {str(record["resource_id"]): str(record["resource_set"]) for record in result}
 
     @classmethod
     async def get_resources_in_resource_sets(
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4814,7 +4814,7 @@ class Resource(BaseDocument):
         deleted_resource_sets: abc.Set[str],
         *,
         connection: Optional[asyncpg.connection.Connection] = None,
-    ) -> abc.Set[m.ResourceIdStr]:
+    ) -> dict[m.ResourceIdStr, str]:
         """
         Copy the resources that belong to an unchanged resource set of a partial compile,
         from source_version to destination_version. This method doesn't copy shared resources.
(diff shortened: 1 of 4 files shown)


src/inmanta/server/services/orchestrationservice.py
```diff
@@ -736,8 +736,9 @@ async def _put_version(
             if is_partial_update:
                 # Make mypy happy
                 assert partial_base_version is not None
-                rids_unchanged_resource_sets: abc.Set[
-                    ResourceIdStr
+                # This dict maps a resource id to its resource set for unchanged resource sets.
+                rids_unchanged_resource_sets: dict[
```
Hugo-Inmanta, line 740: This could probably be better named now that it changed types but I couldn't find a good name

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -746,13 +747,19 @@ async def _put_version(
                     deleted_resource_sets=set(removed_resource_sets),
                     connection=connection,
                 )
-                resources_that_moved_resource_sets = rids_unchanged_resource_sets & set(rid_to_resource.keys())
+                resources_that_moved_resource_sets = rids_unchanged_resource_sets.keys() & set(rid_to_resource.keys())
```
arnaudsjs, line 750: Why do we convert `rid_to_resource.keys()` to a set here?

arnaudsjs (approved):

src/inmanta/server/services/orchestrationservice.py
Hugo-Inmanta: I think this can go

src/inmanta/data/__init__.py
```diff
@@ -4863,7 +4863,7 @@ async def copy_resources_from_unchanged_resource_set(
                 destination_version,
                 updated_resource_sets | deleted_resource_sets,
             )
-            return {record["resource_id"] for record in result}
+            return {record["resource_id"]: record["resource_set"] for record in result}
```
Hugo-Inmanta, line 4866: I noticed mypy complains about several return values e.g. `src/inmanta/data/__init__.py:5286: error: Incompatible return value type (got "List[object]", expected "List[str]")  [return-value]` [here](https://github.com/inmanta/inmanta-core/blob/b1cd58a6351a467b6c0c69ef560d1d1f15172d57/src/inmanta/data/__init__.py#LL5286C22-L5286C22) 
Should we cast before returning ?
```suggestion
            return {str(record["resource_id"]): str(record["resource_set"]) for record in result}

```

arnaudsjs: That's a common issue. Data that comes from the database always has the type object, so you need to cast it yourself.

PR #6097, opened 2023-06-08
issue/6023-unmanaged-resources-storage-fixes

### LINKED ISSUE #6023: unmanaged resources storage backend fixes
- [x] Make sure duplicate entries are overridden instead of raising an (implicit) exception
- [x] Add a timestamp column to indicate when a resource was discovered
- [x] rename methods according to the [design](https://github.com/inmanta/designs/blob/master/agent-server/UnmanagedResources.md)

FloLey:
# Description

1. Make sure duplicate entries are overridden instead of raising an (implicit) exception
2. Add a timestamp column to indicate when a resource was discovered
3. Rename methods according to the design

closes #6023 

# Self Check:

- [x] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -5859,24 +5859,22 @@ def to_dao(self) -> m.User:
...
     :param environment: the environment of the resource
-    :param unmanaged_resource_id: The id of the resource
-    :param values: The values associated with the unmanaged_resource
+    :param discovered_resource_id: The id of the resource
+    :param values: The values associated with the discovered_resource
     """
 
     environment: uuid.UUID
-    unmanaged_resource_id: m.ResourceIdStr
+    discovered_time: datetime.datetime
```
arnaudsjs, line 5870: I don't think the name `discovered_time` very clear. The `_time` part makes me think it expresses a certain duration, while the type indicates it's a timestamp. I think that `discovered_at` would be better for example..
```suggestion
    discovered_at: datetime.datetime
```

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -1571,6 +1571,26 @@ async def insert(self, connection: Optional[asyncpg.connection.Connection] = Non
...
+        """
+        Insert a new document based on the instance passed. If the document already exists, overwrite it.
+        """
+        (column_names, values) = self._get_column_names_and_values()
+        column_names_as_sql_string = ",".join(column_names)
+
+        values_as_parameterize_sql_string = ",".join(["$" + str(i) for i in range(1, len(values) + 1)])
+        primary_key_fields = self._get_names_of_primary_key_fields()
+        primary_key_string = ",".join(primary_key_fields)
+        update_set = list(set(column_names) - set(self._get_names_of_primary_key_fields()))
```
Hugo-Inmanta, line 1584: I think you don't need to cast to a list here

src/inmanta/data/__init__.py
```diff
@@ -1571,6 +1571,26 @@ async def insert(self, connection: Optional[asyncpg.connection.Connection] = Non
...
+        update_set_string = ",\n".join([f"{item} = EXCLUDED.{item}" for item in update_set])
+        query = f"""INSERT INTO {self.table_name()}
+            ({column_names_as_sql_string})
+            VALUES ({values_as_parameterize_sql_string})
+            ON CONFLICT ({primary_key_string})
+            DO UPDATE SET
+            {update_set_string};"""
```
Hugo-Inmanta, line 1591: This is nitpicking, but I find this can be made even more readable: (same for the other query)
```suggestion
        query = f"""
            INSERT INTO {self.table_name()}
            ({column_names_as_sql_string})
            VALUES ({values_as_parameterize_sql_string})
            ON CONFLICT ({primary_key_string})
            DO UPDATE SET
            {update_set_string};
        """
```

src/inmanta/data/__init__.py
```diff
@@ -1571,6 +1571,26 @@ async def insert(self, connection: Optional[asyncpg.connection.Connection] = Non
...
         await self._execute_query(query, *values, connection=connection)
 
+    async def insert_with_overwrite(self, connection: Optional[asyncpg.connection.Connection] = None) -> None:
+        """
+        Insert a new document based on the instance passed. If the document already exists, overwrite it.
+        """
+        (column_names, values) = self._get_column_names_and_values()
+        column_names_as_sql_string = ",".join(column_names)
+
+        values_as_parameterize_sql_string = ",".join(["$" + str(i) for i in range(1, len(values) + 1)])
```
arnaudsjs, line 1581:
```suggestion
        values_as_parameterized_sql_string = ",".join(["$" + str(i) for i in range(1, len(values) + 1)])
```

src/inmanta/data/__init__.py
```diff
@@ -1571,6 +1571,26 @@ async def insert(self, connection: Optional[asyncpg.connection.Connection] = Non
         )
         await self._execute_query(query, *values, connection=connection)
 
+    async def insert_with_overwrite(self, connection: Optional[asyncpg.connection.Connection] = None) -> None:
```
arnaudsjs, line 1574: There is a lot of code duplication between the `insert_with_overwrite` method and the `insert_many_with_overwrite` method. The latter is just an extended version of the former. If we let the `insert_with_overwrite` method delegate to `insert_many_with_overwrite`, by passing in just one document, we can get rid of a lot of code.

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -1634,6 +1654,48 @@ async def insert_many(
...
+        start_value = 1
+        placeholders = []
+        for _ in range(len(values)):
+            sublist = []
+            step = len(values[0])
+            for i in range(start_value, start_value + step):
+                sublist.append(f"${i}")
+            placeholders.append("(" + ", ".join(sublist) + ")")
+            start_value += step
+        values_placeholder = ", ".join(placeholders)
```
Hugo-Inmanta, line 1688: This could maybe use a comment as to what it is doing ? And/or I find this more readable with list comprehension: (not a strong opinion at all, but the comment would still be nice)
```suggestion
n_docs = ...
n_cols = ...
placeholders = ', '.join(
    [
        '(' + ', '.join([f'${doc*n_cols+col}' for col in range(1, n_cols+1)]) + ')' for doc in range(n_docs)
    ]
)

```

Hugo-Inmanta (changes requested):

src/inmanta/data/__init__.py
FloLey: Do you still find a comment is required? I find the variable name clear enough and I don't know what to put in the comment

src/inmanta/data/__init__.py
```diff
@@ -1634,6 +1640,46 @@ async def insert_many(
...
+        column_names = cls.get_field_names()
+        primary_key_fields = cls._get_names_of_primary_key_fields()
+        primary_key_string = ",".join(primary_key_fields)
+        update_set = set(column_names) - set(cls._get_names_of_primary_key_fields())
+        update_set_string = ",\n".join([f"{item} = EXCLUDED.{item}" for item in update_set])
+
+        values: List[List[object]] = []
+        for document in documents:
+            col_names, dao_values = document._get_column_names_and_values()
+            values.append(dao_values)
```
arnaudsjs, line 1661:
```suggestion
        values: List[List[object]] = [
        document._get_column_names_and_values()[1] for document in documents
        ]
```

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
Hugo-Inmanta: No, it's fine without a comment with the list comprehension syntax :+1:

Hugo-Inmanta (changes requested):
There are still a couple of changes that weren't addressed, maybe you didn't push them all?

Hugo-Inmanta (approved):
Please disregard latest comment, I was not the reviewing the latest version!

src/inmanta/data/__init__.py
```diff
@@ -1634,6 +1640,46 @@ async def insert_many(
...
+
+        query = f"""INSERT INTO {documents[0].table_name()}
+                    ({column_names_as_sql_string})
+                    VALUES {placeholders}
+                    ON CONFLICT ({primary_key_string})
+                    DO UPDATE SET
+                    {update_set_string};"""
+
+        flattened_values = [item for sublist in values for item in sublist]
+        await documents[0]._execute_query(query, *flattened_values)
```
Hugo-Inmanta, line 1678: @arnaudsjs we should change it here as well right ?
```suggestion
        await cls._execute_query(query, *flattened_values)
```

arnaudsjs (approved):

Issue #6117, opened 2023-06-15
Deadlock in environment cleanup


wouterdb:
```
09:52:08.330 INFO 200 GET /api/v1/agent (127.0.0.1) 7.16ms
09:52:08.334 DEBUG Getting config in section client_rest_transport
09:52:08.334 DEBUG Calling server DELETE http://localhost:59227/api/v1/environment/0e3fb94f-3d20-48bd-b31f-1541547f645a
09:52:08.336 DEBUG HTTP version of request: HTTP/1.1
09:52:08.336 DEBUG Calling method <bound method EnvironmentService.delete_environment of <inmanta.server.services.environmentservice.EnvironmentService object at 0x7f30247d1be0>>(environment_id='0e3fb94...')
09:52:08.339 DEBUG Stopping all autostarted agents for env 0e3fb94f-3d20-48bd-b31f-1541547f645a
09:52:08.501 DEBUG Expiring all sessions for 0e3fb94f-3d20-48bd-b31f-1541547f645a
09:52:08.501 DEBUG Expired session with id 87149f6e-0b51-11ee-a1c8-525400bb0bf2, last seen 0 seconds ago
09:52:08.502 DEBUG Heartbeat wait expired for jenkins-slave-4.ii.inmanta.com, returning. (long poll)
09:52:08.502 INFO 200 POST /api/v1/heartbeat (127.0.0.1) 206.53ms
09:52:08.502 DEBUG expiring session 87149f6e-0b51-11ee-a1c8-525400bb0bf2
09:52:09.343 ERROR An exception occured during the request.
Traceback (most recent call last):
  File "/home/jenkins/workspace/core_inmanta-core_master/project/inmanta-core/.tox/py39/lib/python3.9/site-packages/inmanta/protocol/rest/__init__.py", line 509, in _execute_call
    result = await config.handler(**arguments.call_args)
  File "/home/jenkins/workspace/core_inmanta-core_master/project/inmanta-core/.tox/py39/lib/python3.9/site-packages/inmanta/server/services/environmentservice.py", line 252, in delete_environment
    await self.environment_delete(environment_id)
  File "/home/jenkins/workspace/core_inmanta-core_master/project/inmanta-core/.tox/py39/lib/python3.9/site-packages/inmanta/server/services/environmentservice.py", line 480, in environment_delete
    await asyncio.gather(self.autostarted_agent_manager.stop_agents(env), env.delete_cascade())
[...]

PR #6120, opened 2023-06-15
added fixture to capture deadlock logs

wouterdb:
# Description

In relation to #6117 and #6090 

1. I was unable to reproduce 
1. it is unclear to me how postgresql takes locks when doing a delete cascade 
 - In general, locks should be acquired consistently with delete cascade lock order, which is top down.  This means
   -  agentprocess -> agentinstance
   - environment -> agentprocess
   - environment -> agent
  2. the agent manager assumes agentprocess -> agentinstance -> agent
  3. I found one ordering bug in the agent manager, but I don't know if that was it
  4. I made it so the postgresql log part after the deadlock is logged, that is like
  
  ```15:37:49.068 WARNING 2023-06-15 15:29:43.830 CEST [75118] DETAIL:  Process 75118 waits for AccessExclusiveLock on relation 16571 of database 16385; blocked by process 75121.
15:37:49.068 WARNING 	Process 75121 waits for RowExclusiveLock on relation 16599 of database 16385; blocked by process 75118.
15:37:49.068 WARNING 	Process 75118: LOCK TABLE public.agentprocess
15:37:49.068 WARNING 	Process 75121: LOCK TABLE agent IN ROW EXCLUSIVE MODE
15:37:49.068 WARNING 2023-06-15 15:29:43.830 CEST [75118] HINT:  See server log for query details.
15:37:49.068 WARNING 2023-06-15 15:29:43.830 CEST [75118] STATEMENT:  LOCK TABLE public.agentproces

```

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/agentmanager.py
+++ b/src/inmanta/server/agentmanager.py
@@ -411,10 +411,10 @@ class AgentManager(ServerSlice, SessionListener):
         now = datetime.now().astimezone()
         async with data.AgentProcess.get_connection() as connection:
             async with connection.transaction():
+                await data.AgentProcess.update_last_seen(session.id, now, connection)
                 await data.AgentInstance.log_instance_creation(session.tid, session.id, endpoints_to_add, connection)
                 await data.AgentInstance.log_instance_expiry(session.id, endpoints_to_remove, now, connection)
                 await data.Agent.update_primary(session.tid, endpoints_with_new_primary, now, connection)
-                await data.AgentProcess.update_last_seen(session.id, now, connection)
 
     # Session registration
     async def _register_session(self, session: protocol.Session, endpoint_names_snapshot: Set[str], now: datetime) -> None:
--- a/changelogs/unreleased/6090-deadlock.yml
+++ b/changelogs/unreleased/6090-deadlock.yml
@@ -1,0 +1,4 @@
+description: Added more log info when detecting a deadlock
+issue-nr: 6090
+change-type: patch
+destination-branches: [master, iso6]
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/inmanta/server/agentmanager.py [hunk 1]
```diff
@@ -411,10 +411,10 @@ async def _log_session_seen_to_db(
         now = datetime.now().astimezone()
         async with data.AgentProcess.get_connection() as connection:
             async with connection.transaction():
+                await data.AgentProcess.update_last_seen(session.id, now, connection)
```
arnaudsjs, line 414: If I understand everything correctly. The locking order is: `environment->agent->agentprocess->agentinstance`. Do you agree? In that case we should move the last line to the top as well.

arnaudsjs (changes requested):

src/inmanta/server/agentmanager.py
wouterdb: agentprocess->agentinstance -> agent 

is what is used in the rest of the agentmanager.

For the cascading delete, I don't know.

This at least makes it consistent with the rest of the file

arnaudsjs: Hmm. In the `Environment.delete_cascade()` method, the `Agent` is delete first and only then the `AgentProcess`. I am not convinced the locking order is right now, but maybe it's better to move forward with little steps and to see what happens.

arnaudsjs (approved):

src/inmanta/server/agentmanager.py
wouterdb: In the first part of the method: yes, but no transaction there. 

The second part: I don;t know.

arnaudsjs: Indeed. I completely missed the fact that it doesn't run in a transaction.

PR #6125, opened 2023-06-16
Issue/5944 produce warnings when database pool is exhausted

### LINKED ISSUE #5944: Produce warnings when database pool is exhausted
Warning in the server log when it's exhausted. Make sure the warning doesn't get repeated too much (Python may or may not already take care of this). Once a day would be ideal but not too much time should be invested.

this may have to be a regular poll as this is not supported.

Hugo-Inmanta:
# Description

Add the `ExhaustedPoolWatcher` class to count database pool  exhaustion events
- counter is increased when db exhaustion is detected (a recurring task is scheduled every 200ms in the `DatabaseService` startup)
- counter is logged as a warning every 24h and reset if any events were recorded (a recurring task is scheduled every 24h in the `DatabaseService` startup)

closes #5944 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -1296,12 +1296,7 @@ class BaseDocument(object, metaclass=DocumentMeta):
             return util.nullcontext(connection)
         # Make pypi happy
         assert cls._connection_pool is not None
-        try:
-            return cls._connection_pool.acquire(timeout=1.0)
-        except asyncio.TimeoutError:
-            # record exhausted pool event
-            util.ExhaustedPoolWatcher.record_event()
-            return cls._connection_pool.acquire()
+        return cls._connection_pool.acquire()
 
     @classmethod
     def table_name(cls) -> str:
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -1296,7 +1296,13 @@ class BaseDocument(object, metaclass=DocumentMeta):
             return util.nullcontext(connection)
         # Make pypi happy
         assert cls._connection_pool is not None
-        return cls._connection_pool.acquire()
+        try:
+            return cls._connection_pool.acquire(timeout=1.0)
+        except asyncio.TimeoutError:
+            # log pool event
+            LOGGER.warning("Connection pool is exhausted")
+            # exhausted_connections_logger.log()
+            return cls._connection_pool.acquire()
 
     @classmethod
     def table_name(cls) -> str:
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -1299,9 +1299,8 @@ class BaseDocument(object, metaclass=DocumentMeta):
         try:
             return cls._connection_pool.acquire(timeout=1.0)
         except asyncio.TimeoutError:
-            # log pool event
-            LOGGER.warning("Connection pool is exhausted")
-            # exhausted_connections_logger.log()
+            # record exhausted pool event
+            util.ExhaustedPoolWatcher.record_event()
             return cls._connection_pool.acquire()
 
     @classmethod
--- a/src/inmanta/server/services/databaseservice.py
+++ b/src/inmanta/server/services/databaseservice.py
@@ -51,7 +51,7 @@ class DatabaseService(protocol.ServerSlice):
         # Schedule database pool exhaustion watch
         interval: int = 60 * 5  # Check for pool exhaustion every 5 min
         self.schedule(self._check_database_pool_exhaustion, interval=interval, cancel_on_stop=True)
-        interval: int = 3_600 * 24  # Report for pool exhaustion every 24h
+        interval = 3_600 * 24  # Report for pool exhaustion every 24h
         self.schedule(self._report_database_pool_exhaustion, interval=interval, cancel_on_stop=True)
 
     async def stop(self) -> None:
@@ -136,4 +136,5 @@ class DatabaseService(protocol.ServerSlice):
 
     async def _check_database_pool_exhaustion(self) -> None:
         max_size: int = opt.db_connection_pool_max_size.get()
+        assert self._pool  # Make mypy happy
         util.ExhaustedPoolWatcher.check_for_pool_exhaustion(self._pool, max_size, LOGGER)
--- a/src/inmanta/server/services/databaseservice.py
+++ b/src/inmanta/server/services/databaseservice.py
@@ -48,8 +48,11 @@ class DatabaseService(protocol.ServerSlice):
             self.schedule(
                 self._purge_agent_processes, interval=agent_process_purge_interval, initial_delay=0, cancel_on_stop=False
             )
-        interval: int = 3_600 * 24  # Check for pool exhaustion every 24h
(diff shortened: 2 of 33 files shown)


arnaudsjs (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -1296,7 +1296,12 @@ def get_connection(
             return util.nullcontext(connection)
         # Make pypi happy
         assert cls._connection_pool is not None
-        return cls._connection_pool.acquire()
+        try:
+            return cls._connection_pool.acquire(timeout=1.0)
+        except asyncio.TimeoutError:
+            # record exhausted pool event
+            util.ExhaustedPoolWatcher.record_event()
+            return cls._connection_pool.acquire()
```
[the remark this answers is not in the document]
Hugo-Inmanta: > But I don't really know how asyncpg behaves in that situation. Did you investigate this?

I'm not following 100% what do you mean ?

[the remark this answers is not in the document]
Hugo-Inmanta: >  It would be nice if we can also detect the situation where the pool size on the database server is configured to be lower than the pool size on the inmanta server

I added a check in the `check_for_pool_exhaustion` method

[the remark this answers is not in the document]
arnaudsjs: PostgreSQL has a config option to indicate the maximum amount of connections it accepts (See: max_connections config option [here](https://www.postgresql.org/docs/13/runtime-config-connection.html)). So it's possible that max_connection < connection-pool-max-size. But this is something that should be handled by the monitoring/alerting of the PostgreSQL server, so you can ignore that comment.

src/inmanta/util.py
```diff
@@ -747,3 +748,38 @@ def ensure_event_loop() -> asyncio.AbstractEventLoop:
...
+        cls._exhausted_pool_events_count = 0
+
+    @classmethod
+    def report_and_reset(cls, logger: logging.Logger) -> None:
+        """
+        Log how many exhausted pool events were recorded since the last time the counter
+        was reset, if any, and reset the counter.
+        """
+        if cls._exhausted_pool_events_count > 0:
+            logger.warning("Database pool was empty %d times in the past 24h." % cls._exhausted_pool_events_count)
```
arnaudsjs, line 775:
```suggestion
            logger.warning("Database pool was exhausted %d times in the past 24h.", cls._exhausted_pool_events_count)
```

src/inmanta/util.py
```diff
@@ -747,3 +748,38 @@ def ensure_event_loop() -> asyncio.AbstractEventLoop:
         new_loop = asyncio.new_event_loop()
         asyncio.set_event_loop(new_loop)
         return new_loop
+
+
+class ExhaustedPoolWatcher:
```
arnaudsjs, line 753: Let's not make this a singleton. We also have a connection pool on the lsm extension. So we will need two instances of this class in the end.

arnaudsjs (changes requested):

src/inmanta/server/services/databaseservice.py
```diff
@@ -48,6 +48,11 @@ async def start(self) -> None:
             self.schedule(
                 self._purge_agent_processes, interval=agent_process_purge_interval, initial_delay=0, cancel_on_stop=False
             )
+        # Schedule database pool exhaustion watch
+        interval: int = 60 * 5  # Check for pool exhaustion every 5 min
```
arnaudsjs, line 52: The amount of connections being used varies continuously. If we only check this once every five minutes, it might take a very long time until we notice that we have connection pool issues. I checked with Wouter and he proposed to run the check five times per second because the check itself  is very light weight.

arnaudsjs (changes requested):

arnaudsjs (changes requested):

arnaudsjs (changes requested):
Do you also add a simple test case?

tests/server/test_databaseservice.py
```diff
@@ -40,3 +44,56 @@ async def _wait_until_expire_is_finished():
...
+    opt.db_connection_pool_min_size.set(str(save_min_size))
+    opt.db_connection_pool_max_size.set(str(save_max_size))
+
+
+@pytest.mark.slowtest
+async def test_pool_exhaustion_watcher(set_pool_size_to_one, server, caplog):
+    """
+    Test the basic functionalities of the ExhaustedPoolWatcher class
+    """
+    with caplog.at_level(logging.WARNING, logging.getLogger("inmanta.server.services.databaseservice").name):
```
arnaudsjs, line 71:
```suggestion
    with caplog.at_level(logging.WARNING, "inmanta.server.services.databaseservice"):
```

tests/server/test_databaseservice.py
```diff
@@ -40,3 +44,56 @@ async def _wait_until_expire_is_finished():
...
+        database_slice = server.get_slice(databaseservice.SLICE_DATABASE)
+
+        assert database_slice._db_pool_watcher._exhausted_pool_events_count == 0
```
arnaudsjs, line 74: This assertion is not always true. This test case runs an Inmanta server with a database pool size of 1. The server executes several background task, that might need a database connection. So at this point there might already be a pool exhaustion event.

tests/server/test_databaseservice.py
```diff
@@ -40,3 +44,56 @@ async def _wait_until_expire_is_finished():
...
+
+        # Check that _report_database_pool_exhaustion resets the counter:
+        assert database_slice._db_pool_watcher._exhausted_pool_events_count == 0
+
+        log_contains(
+            caplog,
+            "inmanta.server.services.databaseservice",
+            logging.WARNING,
+            f"Database pool was exhausted {n_events} times in the past 24h",
+        )
```
arnaudsjs, line 99: This assertion is not always true. Background tasks may have triggered additional exhaustion events after `n_events` was recorded.

arnaudsjs (changes requested):

tests/server/test_databaseservice.py
```diff
@@ -40,3 +44,56 @@ async def _wait_until_expire_is_finished():
...
+            await connection.close()
+
+        n_events: int = database_slice._db_pool_watcher._exhausted_pool_events_count
+        assert n_events > 0
+
+        # Call _report_database_pool_exhaustion manually (scheduled to run every 24h)
+        await database_slice._report_database_pool_exhaustion()
+
+        # Check that _report_database_pool_exhaustion resets the counter:
+        assert database_slice._db_pool_watcher._exhausted_pool_events_count == 0
```
Hugo-Inmanta, line 92: This one has to go as well, some new events might get recorded between this check and the call to report just above

tests/server/test_databaseservice.py
```diff
@@ -40,3 +44,57 @@ async def _wait_until_expire_is_finished():
...
+    opt.db_connection_pool_min_size.set("1")
+    opt.db_connection_pool_max_size.set("1")
+
+    yield
+
+    opt.db_connection_pool_min_size.set(str(save_min_size))
+    opt.db_connection_pool_max_size.set(str(save_max_size))
+
+
+@pytest.mark.slowtest
```
Hugo-Inmanta, line 66: this can go as well I think now

arnaudsjs (approved):

PR #6138, opened 2023-06-20
Fix sequencing issue stop procedure of database slice

arnaudsjs:
# Description

Make sure the background tasks of the database slice are stopped first, before the slice disconnects from the database. This is required because the background tasks need the database connection.

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (approved):
Nice catch :+1:

sanderr (approved):

PR #6144, opened 2023-06-21
Document pip configuration section in project.yml file

arnaudsjs:
# Description

Added documentation regarding the pip configuration section in the project.yml file.

Part of #5976

# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/reference/projectyml.rst
+++ b/docs/reference/projectyml.rst
@@ -16,11 +16,7 @@ The ``project.yml`` file defines the following settings:
 
 .. autoclass:: inmanta.module.ProjectMetadata
 
-.. autoclass:: inmanta.module.ModuleRepoInfo
-    :show-inheritance:
-
-.. autoclass:: inmanta.module.ModuleRepoType
-    :show-inheritance:
+.. autoclass:: inmanta.module.ProjectPipConfig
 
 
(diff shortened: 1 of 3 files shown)


FloLey (approved):

docs/reference/projectyml.rst
```diff
@@ -16,11 +16,7 @@ The ``project.yml`` file defines the following settings:
 
 .. autoclass:: inmanta.module.ProjectMetadata
 
-.. autoclass:: inmanta.module.ModuleRepoInfo
-    :show-inheritance:
-
-.. autoclass:: inmanta.module.ModuleRepoType
-    :show-inheritance:
```
sanderr, line 23: Should we really get rid of this? I thought we'd just deprecate them once Hugo's PR is in?

sanderr (changes requested):

docs/reference/projectyml.rst
arnaudsjs: If you look at the rendered documentation, you see that these two don't really provide any useful information to the end-user. As such, I though it was better to remove it entirely.

sanderr: I see, and the `repo` field itself still has documentation.

sanderr (approved):

PR #6145, opened 2023-06-21
Print deadlock issues

FloLey:
# Description

add an assertion to the postgres_db to make if fail is there is a deadlock. This in the hope we will be able to get the logs.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -244,9 +244,15 @@ def postgres_db(request: pytest.FixtureRequest):
     logfile = os.path.join(initial_cwd, "pg.log")
     if os.path.exists(logfile):
         with open(logfile, "r") as fh:
+            has_deadlock = False
             for line in fh:
                 if "deadlock" in line:
-                    assert False
+                    has_deadlock = True
+                    break
+            sublogger = logging.getLogger("pytest.postgresql.deadlock")
+            for line in fh:
+                sublogger.warning("%s", line)
+            assert not has_deadlock
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -247,7 +247,9 @@ def postgres_db(request: pytest.FixtureRequest):
             for line in fh:
                 if "deadlock" in line:
                     break
+            sublogger = logging.getLogger("pytest.postgresql.deadlock")
             for line in fh:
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -246,10 +246,7 @@ def postgres_db(request: pytest.FixtureRequest):
         with open(logfile, "r") as fh:
             for line in fh:
                 if "deadlock" in line:
-                    break
-            sublogger = logging.getLogger("pytest.postgresql.deadlock")
-            for line in fh:
-                sublogger.warning("%s", line)
+                    assert False
```
FloLey, line 249: @sanderr for the deadlock issue, is this what you proposed [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1687164561125909?thread_ts=1687160441.637899&cid=CKRF0C8R3)? because I'm not sure I see how this will help

sanderr: Not completely. I meant `assert False` after this code block, not as a replacement. But I'm not sure it would help either, I think it might.

sanderr: Hmm, the lines you deleted did not seem to do what I thought they did.... What I meant by my proposal was to log the warning if there is a deadlock and after logging the warning `assert False`.

FloLey: so something like this:
```
        with open(logfile, "r") as fh:
            has_deadlock = False
            for line in fh:
                if "deadlock" in line:
                    has_deadlock = True
                    break
            for line in fh:
                sublogger.warning("%s", line)
            if has_deadlock:
                assert False
```
but what will the assert false accomplish?

sanderr: I have to admit I only followed this with one eye so I may be missing something, but what I understood is that we log these warnings but they are not shown because pytest only shows log messages when a test fails. If this is indeed the problem we're trying to solve, `assert False` serves to fail this fixture cleanup, and I hope that will also lead to the log messages being shown.

sanderr: One small style note: if you already have the `has_deadlock` variable, `assert not has_deadlock` provides way more clarity to the end user than `assert False`.

FloLey: I see! let's try that :)

sanderr (approved):

PR #6148, opened 2023-06-21
Bugfix on import

wouterdb:
# Description

Fix issue where we would not correctly import `Requirement` 

This requires a very specific setup, that was not covered by the testcases yet

related to https://github.com/inmanta/inmanta-core/pull/4508
# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/moduletool/test_install.py
+++ b/tests/moduletool/test_install.py
@@ -478,7 +478,9 @@ def test_project_install(
         "\n".join(f"import {mod}" for mod in ["std", *install_module_names]),
         autostd=False,
         python_package_sources=[local_module_package_index],
-        python_requires=[Requirement.parse(module.ModuleV2Source.get_package_name_for(mod)) for mod in install_module_names],
+        # We add tornado, as there is a code path in update for the case where the project has python requires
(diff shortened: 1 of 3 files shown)


tests/moduletool/test_install.py
```diff
@@ -488,6 +490,8 @@ def test_project_install(
     # autostd=True reports std as an import for any module, thus requiring it to be v2 because v2 can not depend on v1
     module.Project.get().autostd = False
     ProjectTool().execute("install", [])
+    # Make sure update works in all cases where install passed
+    ProjectTool().execute("update", [])
```
sanderr, line 494: By always doing the update, don't we lose the assertions on the install?

sanderr (commented):
We should really move to `packaging` at some point, but for consistency I wouldn't make the change now for this instance.

tests/moduletool/test_install.py
wouterdb: I'll move it

sanderr (approved):

Hugo-Inmanta (approved):

Issue #5993, opened 2023-05-16
add dedicated pip index url config for full project


sanderr:
Follow-up for #5976.

- [x] deprecate package source config
- [x] introduce `pip: { index_url: str[] }` config option: applies to all the project's pip installs, not only v2
- [ ] Open question: what with agent installs?
- [ ] make sure pytest-inmanta etc still work + create follow-up tickets for migration to non-deprecated config
- [ ] update project template: also make sure there is no confusion on repo_url vs package_repo, renaming them and adding comments clarifying what they are and why they're there, then pointing to the docs.

Hugo-Inmanta:
As discussed in planning meeting don't rename the options in the project template

PR #6150, opened 2023-06-21
Issue/5993 add dedicated pip index url config for full project

Hugo-Inmanta:
# Description

- [x] deprecate package source config
- [x] introduce `pip: { index_url: str[] }` config option: applies to all the project's pip installs, not only v2
~~- [ ] Open question: what with agent installs?~~
        - Out of scope for now -> TODO -> create follow-up ticket ?
- [x] make sure pytest-inmanta etc still work + create follow-up tickets for migration to non-deprecated config
I had to make a few changes to check if it still worked. Looks like it is working: all tests pass when running `make test`
 -> PR [here](https://github.com/inmanta/pytest-inmanta/pull/396) 
- [x] update project template: also make sure there is no confusion on repo_url vs package_repo, renaming them and adding comments clarifying what they are and why they're there, then pointing to the docs.
        - PR [here](https://github.com/inmanta/inmanta-project-template/pull/15)

closes https://github.com/inmanta/inmanta-core/issues/5993

# Self Check:

- [x] Changelog entry


src/inmanta/util.py
```diff
@@ -750,6 +750,52 @@ def ensure_event_loop() -> asyncio.AbstractEventLoop:
         return new_loop
 
 
+class OnAccess(enum.EnumMeta):
```
Hugo-Inmanta, line 753: Both this class and `DeprecatedEnum` are still raising a lot of mypy errors, I could use some help with dealing with those

docs/reference/projectyml.rst
```diff
@@ -52,7 +50,9 @@ The code snippet below provides an example of a complete ``project.yml`` file:
     freeze_recursive: true
     freeze_operator: ~=
     pip:
-       use_config_file: False
+      use_config_file: False
```
sanderr, line 53:
```suggestion
      use_config_file: false
```
I think both work but let's be consistent.

src/inmanta/module.py
```diff
@@ -1657,7 +1666,7 @@ def get_relation_precedence_rules(self) -> List[RelationPrecedenceRule]:
         return [RelationPrecedenceRule.from_string(rule_as_str) for rule_as_str in self.relation_precedence_policy]
 
     def get_index_urls(self) -> List[str]:
-        return [repo.url for repo in self.repo if repo.type == ModuleRepoType.package]
+        return self.pip.index_url
```
sanderr, line 1669: We must not ignore the old config: it's deprecated but still supported. In addition to that, the old and the new configs have different semantics, which is a big part of the ticket that's still missing in this PR. The old one is an index for v2 modules, the new one is an index for the entire project. The current implementation still only uses it for v2 modules, and the error message you updated above is also specific for v2 modules.

src/inmanta/module.py
```diff
@@ -1486,9 +1486,16 @@ def to_config(self, inp: Optional[configparser.ConfigParser] = None) -> configpa
 
 
 @stable_api
-class ModuleRepoType(enum.Enum):
+class ModuleRepoType(DeprecatedEnum):
```
sanderr, line 1489: Wouldn't a pydantic validator be way simpler?

src/inmanta/util.py
sanderr: I can have a look but let's first reach a conclusion on this vs a pydantic validator and then come back to this.

sanderr (changes requested):
Could you add a second reviewer once the dust has settled a bit?

src/inmanta/module.py
Hugo-Inmanta: - added support for the old config back
- log a warning if indexes are set through both options

I might be wrong but I think the indexes are used for v1 modules requirements as well [here](https://github.com/inmanta/inmanta-core/blob/be410459e1ff30671be67e829a5202dc2044b0ab/src/inmanta/module.py#L2078). As I understand it, the v1 modules are loaded above, and then requirements are collected and installed

src/inmanta/module.py
Hugo-Inmanta: Indeed it is, I removed all of it and when with a validator :+1:

src/inmanta/util.py
Hugo-Inmanta: Removed them both

src/inmanta/module.py
```diff
@@ -1654,6 +1647,12 @@ def validate_repo_field(cls, v: object) -> List[Dict[Any, Any]]:
                 # Ensure backward compatibility with the version of Inmanta that didn't have support for the type field.
                 result.append({"url": elem, "type": ModuleRepoType.git})
             elif isinstance(elem, dict):
+                msg = (
+                    "Setting a pip index through the project.yml `repo -> url` option with type `package` is deprecated. "
+                    "Please set the pip index url through the project.yml `pip -> index_url` option instead."
+                )
```
sanderr, line 1653: Why put this in a variable?

src/inmanta/module.py
```diff
@@ -1666,7 +1665,16 @@ def get_relation_precedence_rules(self) -> List[RelationPrecedenceRule]:
         return [RelationPrecedenceRule.from_string(rule_as_str) for rule_as_str in self.relation_precedence_policy]
 
     def get_index_urls(self) -> List[str]:
-        return self.pip.index_url
+        """
+        Once setting repos with type package is no longer supported, this method can return self.pip.index_url alone.
+        """
```
sanderr, line 1670: A docstring is meant for the users of the method. This message is meant for the maintainers of the method.
```suggestion
        # Once setting repos with type package is no longer supported, this method can return self.pip.index_url alone.
```

src/inmanta/module.py
```diff
@@ -1666,7 +1665,16 @@ def get_relation_precedence_rules(self) -> List[RelationPrecedenceRule]:
...
+        index_urls_deprecated_option: List[str] = [repo.url for repo in self.repo if repo.type == ModuleRepoType.package]
+        if all([index_urls_deprecated_option, self.pip.index_url]):
+            LOGGER.warning(
+                "Pip indexes are configured in two places. Setting them through the `repo -> url` option is being "
```
sanderr, line 1674:
```suggestion
                "Pip indexes are configured in two places. Setting them through the `repo -> url` option has been "
```

src/inmanta/module.py
sanderr: > I might be wrong but I think the indexes are used for v1 modules requirements as well [here](https://github.com/inmanta/inmanta-core/blob/be410459e1ff30671be67e829a5202dc2044b0ab/src/inmanta/module.py#L2078). As I understand it, the v1 modules are loaded above, and then requirements are collected and installed

You're right, it's been like that for almost a year, I had no clue.

sanderr (approved):
Wait with the merge until after the release.

src/inmanta/module.py
Hugo-Inmanta: no particular reason, removed it

src/inmanta/module.py
```diff
@@ -1657,7 +1664,14 @@ def get_relation_precedence_rules(self) -> List[RelationPrecedenceRule]:
...
     def get_index_urls(self) -> List[str]:
-        return [repo.url for repo in self.repo if repo.type == ModuleRepoType.package]
+        # Once setting repos with type package is no longer supported, this method can return self.pip.index_url alone.
+        index_urls_deprecated_option: List[str] = [repo.url for repo in self.repo if repo.type == ModuleRepoType.package]
+        if all([index_urls_deprecated_option, self.pip.index_url]):
+            LOGGER.warning(
+                "Pip indexes are configured in two places. Setting them through the `repo -> url` option has been "
+                "deprecated in favour of the `pip -> index_url` option."
+            )
+        return self.pip.index_url + index_urls_deprecated_option
```
arnaudsjs, line 1674: Don't we need to make sure that we don't have any duplicates in this list.

src/inmanta/module.py
```diff
@@ -1657,7 +1664,14 @@ def get_relation_precedence_rules(self) -> List[RelationPrecedenceRule]:
...
 
     def get_index_urls(self) -> List[str]:
-        return [repo.url for repo in self.repo if repo.type == ModuleRepoType.package]
+        # Once setting repos with type package is no longer supported, this method can return self.pip.index_url alone.
+        index_urls_deprecated_option: List[str] = [repo.url for repo in self.repo if repo.type == ModuleRepoType.package]
+        if all([index_urls_deprecated_option, self.pip.index_url]):
+            LOGGER.warning(
+                "Pip indexes are configured in two places. Setting them through the `repo -> url` option has been "
+                "deprecated in favour of the `pip -> index_url` option."
+            )
```
arnaudsjs, line 1673: I don't really see why we need this log message. The warning about the deprecated option will be logged in the `validate_repo_field()` method, right? With this one we will have two warning messages, with very little added value in my opinion.

src/inmanta/module.py
```diff
@@ -1537,10 +1537,12 @@ def __str__(self) -> str:
 class ProjectPipConfig(BaseModel):
     """
     :param use_config_file: Indicates whether the pip configuration files have to be taken into account when installing
-                            Python packages.
+        Python packages.
+    :param index_url: List of pip indexes to use project-wide.
     """
 
     use_config_file: bool = False
+    index_url: List[str] = []
```
arnaudsjs, line 1545: Why do we use the singular word here, while it actually has the type list?
```suggestion
    index_urls: List[str] = []
```

arnaudsjs (changes requested):

src/inmanta/module.py
Hugo-Inmanta: I don't have a strong opinion about this, I went with the naming suggested in the ticket, and it actually is consistent with the naming of other options with type List in https://docs.inmanta.com/inmanta-service-orchestrator/latest/reference/projectyml.html#project-yml (modulepath, repo, relation_precedence_policy)

src/inmanta/module.py
Hugo-Inmanta: I agree, I'm in favour of removing this one.
 @sanderr any objections against this ? I seem to recall you asked for a warning if indexes were set in both places but I can't find it anywhere

sanderr: I don't recall, it's possible that was assuming that we would ignore one if the other was set. I agree with your remark.

src/inmanta/module.py
Hugo-Inmanta: Added a cast to `set` and then back to `list`

src/inmanta/module.py
arnaudsjs: Yes. I see. But I still think it would be more clear to make it plural. I would not care about the consistency issue.

* `modulepath` and `repo` will be removed when we remove support for V1 modules. 
* `relation_precedence_policy` should be singular I think. One policy can consist of multiple rules.

So if we get this one right, we have an easy migration path to a config file that is consistent.

@sanderr @wouterdb Any opinions on this?

Hugo-Inmanta:
Version bump to 9.3.0 because of this [comment](https://github.com/inmanta/pytest-inmanta/pull/396#discussion_r1242107953)

sanderr (changes requested):

src/inmanta/module.py
sanderr: I would not do this because it affects which is the `--index-url` and which are `--extra-index-url`. Afaik pip doesn't really treat them differently but I wouldn't count on that. Going through `dict` instead could work because it keeps insertion order. I don't know off the top of my head how it treats duplicate keys and how it affects insertion order so that should be checked.

Hugo-Inmanta: Ok, I'll make it so order is preserved and document clearly the behaviour regarding `--index-url` vs `--extra-index-url`

sanderr: I don't think we need to document it. That might strengthen the impression that there is a difference. I think if we make sure to document that we strongly discourage using more than one index in the first place we're good. And I think we already have that, or we have a ticket for it.

Hugo-Inmanta: Checked just now, in case of duplicate keys, the value is indeed updated but it has no impact on the recorded insertion order:
```python
import itertools

l1 = ["a1", "b2", "c3"]
l2 = ["c3", "d4", "a1"]

a = list({value: None for value in itertools.chain(l1, l2)})

print(a) #  ['a1', 'b2', 'c3', 'd4']

out = {}
for i, value in enumerate(itertools.chain(l1, l2)):
    out[value] = i

print(out) #  {'a1': 5, 'b2': 1, 'c3': 3, 'd4': 4}
print(list(out)) #  ['a1', 'b2', 'c3', 'd4']


```

arnaudsjs (approved):

sanderr (approved):

PR #6156, opened 2023-06-22
Issue race condition test_deploy_no_code

FloLey:
# Description

Fix race condition in test 'test_deploy_no_code'

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/agent_server/test_server_agent.py
+++ b/tests/agent_server/test_server_agent.py
@@ -3094,7 +3094,6 @@ async def test_deploy_no_code(resource_container, client, clienthelper, environm
 
     await _wait_until_deployment_finishes(client, environment, version)
     # The version, resource state and its logs are not set atomically. This call prevents a race condition.
-    await wait_until_resource_done_state(client, environment, resource_id)
     await wait_until_logs_are_available(client, environment, resource_id, expect_nr_of_logs=4)
 
     response = await client.get_resource(environment, resource_id, logs=True)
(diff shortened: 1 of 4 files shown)


FloLey:
I'm not sure this is the right fix. It fixes the test (previously I always saw at least one failure when running the test 30 times in a row, now I don't anymore), but maybe it is the code that needs a fix.

tests/agent_server/test_server_agent.py
```diff
@@ -3092,7 +3093,8 @@ async def test_deploy_no_code(resource_container, client, clienthelper, environm
     await clienthelper.put_version_simple(resources, version)
 
     await _wait_until_deployment_finishes(client, environment, version)
-    # The resource state and its logs are not set atomically. This call prevents a race condition.
+    # The version, resource state and its logs are not set atomically. This call prevents a race condition.
+    await wait_until_resource_done_state(client, environment, resource_id)
```
arnaudsjs, line 3097: This cannot be the cause of the race condition. `_wait_until_deployment_finishes()` only returns when all the resources are in a non-transient state, while the test case fails because a resource is in the "deploying" state, which is a transient state.

The fact that the resource goes to deploying again, indicates that the deploy or the repair timer has fired. The `autostarted_agent` fixture, [disables the deploy and repair timer](https://github.com/inmanta/inmanta-core/blob/09930d94f66ab663282158b8ea1d74450af96c5d/tests/conftest.py#L659). To activate this setting, the agent needs to be restarted. It seems that [the logic to restart the agent is only present in the code path of the API](https://github.com/inmanta/inmanta-core/blob/09930d94f66ab663282158b8ea1d74450af96c5d/src/inmanta/server/services/environmentservice.py#L203), but this test case doesn't go over the API. It goes directly to the database layer.

arnaudsjs (changes requested):

arnaudsjs (approved):

Hugo-Inmanta (approved):

Issue #6146, opened 2023-06-21
re-add generic CRUDHandler but under a different name


sanderr:
c9fb8842 reverted #6129 because it was causing issues for `CRUDHandler`'s children. We do still want the generic CRUD handler so we will have to add it again but under a different name. We can then simply add an alias `CRUDHandler = GenericCRUDHandler[PurgeableResource]` or something similar.

PR #6166, opened 2023-06-23
Make CRUDHandler generic (attempt 2)

### LINKED ISSUE #5555: Make CRUDHandler generic to its resource
Currently the entire typing of the CRUDHandler (and normal handler) is to the Resource type. However, in handler we often override this to the actual resource. However, type wise this is not correct.

The goal of this ticket it to make the CRUDHandler generic (in a non breaking way). For example:

```python
class LicenseResource(resources.Resource, base.ApiBase):
    fields = ...


class LicenseHandler(handler.CRUDHandler[LicenseResource]):
    def read_resource(self, ctx: handler.HandlerContext, resource: LicenseResource) -> None: ...
```

arnaudsjs:
# Description

This PR adds the CRUDHandlerGeneric class. This class contains the same implementation as the CRUDHandler class, but is generic with respect to the specific PurgeableResource it acts on.

closes #5555 and #6146

# Self Check:

- [x] Changelog entry


src/inmanta/agent/handler.py
```diff
@@ -932,6 +933,14 @@ def execute(self, ctx: HandlerContext, resource: resources.Resource, dry_run: Op
                 )
 
 
+# Keep the CRUDHandler class for backwards compatibility. Making the CRUDHandler class generic can cause MRO issues
+# if a subclass of CRUDHandler is generic as well. The code below makes a copy of the CRUDHandlerGeneric class
+# and removes Generic as a base class.
+CRUDHandler = types.new_class(
```
sanderr, line 939: Iirc my testing revealed that it isn't a problem per se that it inherits from `Generic`, more that it inherited from a non-parametrized generic. I think we should be fine with
```python
class CRUDHandler(CRUDHandlerGeneric[resources.PurgeableResource]): pass

**Review by sanderr — CHANGES_REQUESTED** [review]

File: src/inmanta/agent/handler.py (hunk 1 above)
Comment (arnaudsjs) [reply src/inmanta/agent/handler.py:939]: No. The issue is the method resolution order (MRO) that cannot be determined. The inheritance structure looks like this:

```mermaid
classDiagram
    class Generic
    class CRUDHadlerGeneric
    class CRUDHandler
    class BaseHandler
    class CFSOrderResourceHandler

    Generic <|-- CRUDHadlerGeneric
    CRUDHadlerGeneric <|-- CRUDHandler
    CRUDHandler <|-- BaseHandler
    BaseHandler <|-- CFSOrderResourceHandler
    Generic <|-- CFSOrderResourceHandler
```

The `CFSOrderResourceHandler` class is defined as:

```
class CFSOrderResourceHandler(Generic[Order], BaseHandler):
```

which makes it impossible to find a consistent ordering. If the signature would be changed to:

```
class CFSOrderResourceHandler(BaseHandler, Generic[Order])
```

it works. But that is not going to help us here.

I tested with your proposal and the MRO issue is still there.

sanderr (commented):
As discussed

src/inmanta/agent/handler.py
```diff
@@ -952,6 +952,14 @@ def delete_resource(self, ctx: HandlerContext, resource: TPurgeableResource) ->
     def update_resource(self, ctx: HandlerContext, changes: Dict[str, Dict[str, Any]], resource: TPurgeableResource) -> None:
         pass
 
+    def calculate_diff(
+        self, ctx: HandlerContext, current: TPurgeableResource, desired: TPurgeableResource
+    ) -> typing.Dict[str, typing.Dict[str, typing.Any]]:
+        super().calculate_diff(ctx, current, desired)
```
sanderr, line 958: Could just use `typing.overload` as @wouterdb proposed. But I don't think this one additional super call should have any real effect on performance in practice.

sanderr (approved):

PR #6167, opened 2023-06-23
Add custom log location for logs of psql in tests

FloLey:
# Description

adds a custom log location for the logs produced by postgres in the postgres_db fixture.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -260,6 +260,7 @@ def postgres_db(request: pytest.FixtureRequest):
             for line in fh:
                 sublogger.warning("%s", line)
             assert not has_deadlock
+        os.remove(pg_logfile)
 
 
 @pytest.fixture
@@ -538,10 +539,6 @@ def reset_all_objects():
     compiler.Finalizers.reset_finalizers()
     AuthJWTConfig.reset()
     InmantaLoggerConfig.clean_instance()
-    # truncate the logs produced by postgres
-    if os.path.exists(pg_logfile):
-        with open(pg_logfile, "w") as file:
-            file.truncate()
 
 
 @pytest.fixture()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -229,7 +229,7 @@ def pytest_runtest_setup(item: "pytest.Item"):
 
 
 # adds a custom log location for postgres
-postgresql_proc_with_log = factories.postgresql_proc(startparams=f"--log={pg_logfile}")
+postgresql_proc_with_log = factories.postgresql_proc(startparams=f"--log='{pg_logfile}'")
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -225,25 +228,29 @@ def pytest_runtest_setup(item: "pytest.Item"):
             pytest.skip("Skipping old migration test")
 
 
+# adds a custom log location for postgres
+postgresql_proc_with_log = factories.postgresql_proc(startparams=f"--log={pg_logfile}")
```
arnaudsjs, line 232: I think we have to quote this to keep the people happy that put white spaces in the names of their files and directories.
```suggestion
postgresql_proc_with_log = factories.postgresql_proc(startparams=f"--log='{pg_logfile}'")
```

tests/conftest.py
```diff
@@ -225,25 +228,29 @@ def pytest_runtest_setup(item: "pytest.Item"):
...
+        fixture = "postgresql_proc_with_log"
 
     logger.info("Using database fixture %s", fixture)
     pg = request.getfixturevalue(fixture)
     yield pg
 
-    logfile = os.path.join(initial_cwd, "pg.log")
-    if os.path.exists(logfile):
-        with open(logfile, "r") as fh:
+    if os.path.exists(pg_logfile):
```
arnaudsjs, line 252: The log file is not deleted at the end of the test suite. Is this something we want? Also, new logs are appended to the old log file. So if we don't clean up at the end of the test suite, we at least need to truncate the file at the start of the test case.

arnaudsjs (changes requested):

tests/conftest.py
FloLey: I added a  os.remove(pg_logfile) if the file exists

arnaudsjs: The log file is not deleted when there is a deadlock. This could be a feature or a bug. In any case, if it's a feature we should truncate the file the next time the test is run.

arnaudsjs (changes requested):

arnaudsjs (approved):

tests/conftest.py
```diff
@@ -225,34 +228,39 @@ def pytest_runtest_setup(item: "pytest.Item"):
...
+        with open(pg_logfile, "r") as fh:
             for line in fh:
                 if "deadlock" in line:
                     has_deadlock = True
                     break
             sublogger = logging.getLogger("pytest.postgresql.deadlock")
             for line in fh:
                 sublogger.warning("%s", line)
-            assert not has_deadlock
+        os.remove(pg_logfile)
```
wouterdb, line 262: I would leave it behind, ...

wouterdb (approved):

PR #6173, opened 2023-06-26
Issue deadlock

FloLey:
# Description

test_postgres_cascade_locking_order create a deadlock on purpose which will conflict with the
"assert not has deadlock" check in the postgres_db fixture.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -269,7 +269,9 @@ async def run_without_keeping_psql_logs(postgres_db):
         # Store the original content of the logfile
         with open(pg_logfile, "r") as file:
             original_content = file.read()
-        yield
+    yield
+
+    if os.path.exists(pg_logfile):
         # Restore the original content of the logfile
         with open(pg_logfile, "w") as file:
             file.write(original_content)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -250,17 +250,14 @@ def postgres_db(request: pytest.FixtureRequest):
     yield pg
 
     if os.path.exists(pg_logfile):
-        has_deadlock = False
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -263,6 +263,20 @@ def postgres_db(request: pytest.FixtureRequest):
...
+    if os.path.exists(pg_logfile):
+        # Store the original content of the logfile
+        with open(pg_logfile, "r") as file:
+            original_content = file.read()
+    yield
+
+    if os.path.exists(pg_logfile):
+        # Restore the original content of the logfile
+        with open(pg_logfile, "w") as file:
+            file.write(original_content)
```
arnaudsjs, line 277: The problem with this implementation is that it's rather slow. After every test, it will read the entire log file into memory (including the logs from all previous tests) and write it out again. So this operation becomes exponentially slower with the number of tests. I think it's better to copy the content of the file to a different location and then truncate the old log file. That's something that can be done in constant time. Or are these logs really small, so that it doesn't cause problems. Did you measure the overhead of this implementation?

arnaudsjs (changes requested):

tests/conftest.py
FloLey: it will only do it for the 2 tests that have the fixture. I run the full testsuit here and I don't see that much of a degradation: 
https://jenkins.inmanta.com/job/core/job/inmanta-core/job/issue-deadlock/

arnaudsjs (approved):

Hugo-Inmanta (approved):

PR #6191, opened 2023-06-28
Add support to expose API methods using multiple URLs

arnaudsjs:
# Description

Add support to expose API methods using multiple URLs. If the path property of the typedmethod annotation has type list, the inmanta client will always use the first path to perform its API calls.

closes  inmanta/inmanta-lsm#1274

# Self Check:

- [x] Changelog entry
- [ ] ~~If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)

### DIFF
--- a/src/inmanta/protocol/common.py
+++ b/src/inmanta/protocol/common.py
@@ -366,7 +366,7 @@ class MethodProperties(object):
         if (properties.path, properties.api_version) in current_list:
             raise Exception(
                 f"Method {properties.function.__name__} already has a "
-                f"method definition for api version path {properties.path} and API version {properties.api_version}"
+                f"method definition for api path {properties.path} and API version {properties.api_version}"
             )
 
         cls.methods[properties.function.__name__].append(properties)
--- a/src/inmanta/protocol/common.py
+++ b/src/inmanta/protocol/common.py
@@ -358,12 +358,15 @@ class MethodProperties(object):
 
     @classmethod
     def register_method(cls, properties: "MethodProperties") -> None:
-        """Register new method properties. Multiple properties on a method is supported but versions have to be unique."""
-        current_list = [x.api_version for x in cls.methods[properties.function.__name__]]
-        if properties.api_version in current_list:
+        """
+        Register new method properties. Multiple properties on a method is supported but the (URL, API version) combination has
+        to be unique.
+        """
+        current_list = [(x.path, x.api_version) for x in cls.methods[properties.function.__name__]]
+        if (properties.path, properties.api_version) in current_list:
             raise Exception(
                 f"Method {properties.function.__name__} already has a "
-                f"method definition for api version {properties.api_version}"
+                f"method definition for api version path {properties.path} and API version {properties.api_version}"
             )
 
         cls.methods[properties.function.__name__].append(properties)
@@ -422,6 +425,7 @@ class MethodProperties(object):
             validate_sid = agent_server and not api
 
         self._path = UrlPath(path)
+        self.path = path
         self._operation = operation
         self._reply = reply
         self._arg_options = arg_options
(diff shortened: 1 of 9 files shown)


src/inmanta/protocol/common.py
```diff
@@ -358,12 +358,15 @@ class MethodProperties(object):
...
+        """
+        Register new method properties. Multiple properties on a method is supported but the (URL, API version) combination has
+        to be unique.
+        """
+        current_list = [(x.path, x.api_version) for x in cls.methods[properties.function.__name__]]
+        if (properties.path, properties.api_version) in current_list:
             raise Exception(
                 f"Method {properties.function.__name__} already has a "
-                f"method definition for api version {properties.api_version}"
+                f"method definition for api version path {properties.path} and API version {properties.api_version}"
```
sanderr, line 369:
```suggestion
                f"method definition for api path {properties.path} and API version {properties.api_version}"
```
(I think)

src/inmanta/protocol/endpoints.py
```diff
@@ -363,9 +363,7 @@ def _select_method(self, name) -> Optional[common.MethodProperties]:
         elif self._version_match is VersionMatch.highest:
             return max(methods, key=lambda x: x.api_version)
         elif self._version_match is VersionMatch.exact:
-            for method in methods:
-                if method.api_version == self._exact_version:
-                    return method
+            return next((m for m in methods if m.api_version == self._exact_version), None)
```
sanderr, line 366: Cool, I didn't know `next`  had a fallback parameter.

src/inmanta/protocol/rest/__init__.py
```diff
@@ -123,15 +123,34 @@ def metadata(self) -> Dict[str, object]:
         return self._metadata
 
     def _is_header_param(self, arg: str) -> bool:
+        """
+        Return True of the given call argument can be provided using a header parameter.
```
sanderr, line 127:
```suggestion
        Return True if the given call argument can be provided using a header parameter.
```

sanderr (approved):

src/inmanta/protocol/rest/__init__.py
```diff
@@ -183,52 +202,61 @@ async def process(self) -> None:
         if self._argspec.defaults is not None:
             defaults_start = len(args) - len(self._argspec.defaults)
 
+        # Make sure that an arguments is not passed both using the header and a non-header value with a different value
```
Hugo-Inmanta, line 205:
```suggestion
        # Make sure that an argument is not passed both using the header and a non-header value with a different value
```

src/inmanta/protocol/rest/__init__.py
```diff
@@ -183,52 +202,61 @@ async def process(self) -> None:
...
+                else:
+                    value = self._message[arg]
+                all_fields.remove(arg)
+            elif arg_type and self._properties.operation == "GET" and self._is_dict_or_optional_dict(arg_type):
+                # Argument is dictionary-based expression in query parameters of GET operation
+                dict_prefix = f"{arg}."
+                dict_with_prefixed_names = {
+                    param_name: param_value
+                    for param_name, param_value in self._message.items()
+                    if param_name.startswith(dict_prefix) and len(param_name) > len(dict_prefix)
```
Hugo-Inmanta, line 244: >  len(param_name) > len(dict_prefix)
I'm curious, what is this for ? To prevent matching exactly f"{arg}." with no attribute behind it ?

Hugo-Inmanta (approved):

src/inmanta/protocol/rest/__init__.py
arnaudsjs: yes

PR #6194, opened 2023-06-28
Issue/392 document config private python package repo

FloLey:
# Description

Document the proper way to configure a project to access a private python package repository

closes https://github.com/inmanta/inmanta-service-orchestrator/issues/392

# Self Check:

- [x] Changelog entry


docs/model_developers/modules.rst
```diff
@@ -329,3 +329,41 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
 The only exception to this rule is when using the ``inmanta export`` command. It exports a project and all its modules'
 ``inmanta_plugins`` packages to the orchestrator server. When this method is used, the orchestrator does not install any modules
 from the Python package repository but instead contains all Python code as present in the local Python environment.
+
+Configure a project to install modules form a private python package repository
+--------------------------------------------------------------------------------
+
+Modules v2 are Python packages that can be installed from private Python package repositories, which may require authentication. This section explains the steps to install such modules.
+
+Create a file named ``.netrc`` in the orchestrator's file system, specifically in the "inmanta" directory.
```
arnaudsjs, line 338: We need to be more specific about the directory where this file needs to be created. There are many "inmanta" directories on the orchestrator.
```suggestion
Create a file named ``/var/lib/inmanta/.netrc`` in the orchestrator's file system. 
```

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,41 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+
+  [global]
+  timeout = 60
+  index-url = <url of the repository >
+
+Set the ``PIP_CONFIG_FILE`` environment variable to the path of the ``pip.conf`` file. You can do this by running the following command:
+
+.. code-block:: bash
+
+  $ export PIP_CONFIG_FILE=<path/to/pip.conf>
```
arnaudsjs, line 361: I think this procedure should describe how the orchestrator should be configured so that compiles started by the server can access the private python package repository. Setting the environment variable in a bash shell will only work for compiles started using the CLI tool.

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,41 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+
+Run the following command to install the modules:
+
+.. code-block:: bash
+
+  $ inmanta project install
+
+By following these steps, you will be able to configure your project to install modules from a private Python package repository.
```
arnaudsjs, line 369: The compiles service will execute these, so no need to mention them here. Unless we want to have a separate section for compiles on the CLI, but I am not sure whether we need that.

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,41 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+
+Create a file named ``pip.conf`` and add the following content to it:
+
+.. code-block:: text
+
+  [global]
+  timeout = 60
+  index-url = <url of the repository >
+
+Set the ``PIP_CONFIG_FILE`` environment variable to the path of the ``pip.conf`` file. You can do this by running the following command:
```
arnaudsjs, line 357: I am not sure whether we want to create a new pip config file here and set the environment variable or whether it would be better to change the global pip config file. I guess the latter is better, but I am not sure.

arnaudsjs (changes requested):

arnaudsjs:
@edvgui I added you as a reviewer, because you created the ticket and have a better understanding about how this is configured in practice.

docs/model_developers/modules.rst
FloLey: I'll remove that part

docs/model_developers/modules.rst
FloLey: I could also speak about booth options as they both work?

arnaudsjs: That's a possibility. Regarding the approach with the environment variable: I would mention the procedure to make sure that the environment variable is automatically set when a bash shell is opened. Having to run the export command every time a user logs into the orchestrator is not a very practical solution.

FloLey: but this part should be removed completely right? or do I misunderstand [your other comment](https://github.com/inmanta/inmanta-core/pull/6194#discussion_r1245274243)?

docs/model_developers/modules.rst
FloLey: I don't think I know what actions should be taken on the orchestrator. I expect if we use the global pip config file nothing needs to be done?

docs/model_developers/modules.rst
arnaudsjs: I don't really see how the referenced comment is linked to this. If we use the global pip config instead of a custom config file, this can indeed be removed.

docs/model_developers/modules.rst
arnaudsjs: True. If we use the global config, we don't need to environment variable. Otherwise, we needs to set the environment variable in the `/etc/sysconfig/inmanta-server` file (See: https://docs.inmanta.com/inmanta-service-orchestrator-dev/6/administrators/credentials.html)

arnaudsjs:
@FloLey Do you also make sure you test this procedure?

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,41 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+
+v2 modules are Python packages that can be installed from private Python package repositories, which may require authentication. This section explains the steps to install such modules.
+
+Create a file named ``/var/lib/inmanta/.netrc`` in the orchestrator's file system. 
+Add the following content to the file:
+
+.. code-block:: text
+
+  machine <url of the private repository>
+  login token
```
Hugo-Inmanta, line 344: Why not ?
```suggestion
  login <login>
```

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,41 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
 The only exception to this rule is when using the ``inmanta export`` command. It exports a project and all its modules'
 ``inmanta_plugins`` packages to the orchestrator server. When this method is used, the orchestrator does not install any modules
 from the Python package repository but instead contains all Python code as present in the local Python environment.
+
+Configure a project to install modules from a private python package repository
+--------------------------------------------------------------------------------
+
```
Hugo-Inmanta, line 335: I think this section could use 2 sub-sections: 
1. The one you already described with `use_config_file` to True
2. Another one with the pip -> index_urls ([ticket](https://github.com/inmanta/inmanta-core/issues/5993) adding this for reference)
The option is used like this in the `project.yml`:
```yml
pip:
    index_urls:
        - <url of the private repository>
```
I don't know how this second section sits with the .netrc file though (i.e. maybe this is a common part to both sub-sections)

Hugo-Inmanta (changes requested):

docs/model_developers/modules.rst
FloLey: because that's how guillaume described it here: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-service-orchestrator/392 :p 
I haven't tested it yet

docs/model_developers/modules.rst
FloLey: what happens when `use_config_file` is true and a `index_urls` is provided?

docs/model_developers/project_creation.rst
```diff
@@ -44,12 +44,12 @@ An example ``project.yml`` could be:
...
       - url: https://github.com/inmanta/
         type: git
-      - url: https://pypi.org/simple
-        type: package
   install_mode: release
   requires:
   pip:
       use_config_file: False
+      index_urls:
+          - https://pypi.org/simple
```
FloLey, line 52: is this correct @Hugo-Inmanta ?

docs/model_developers/modules.rst
Hugo-Inmanta: Both are used: see [here](https://github.com/inmanta/inmanta-core/blob/c79cc1e1376a13367fa2a3c1949cd4b864ea7fdd/src/inmanta/env.py#L360) and the indexes specified by index_urls will be passed to pip as `--extra-index-url`'s

docs/model_developers/modules.rst
FloLey: it has been tested and token is indeed correct

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,61 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+V2 modules are Python packages that can be installed from private Python package repositories, which may require authentication.This section explains the steps to install such modules.
+
+Create a file named ``/var/lib/inmanta/.netrc`` in the orchestrator's file system.
+Add the following content to the file:
+
+.. code-block:: text
+
+  machine <url of the private repository>
```
arnaudsjs, line 343:
```suggestion
  machine <hostname of the private repository>
```

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,61 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+  login token
+  password <the token to access the private repository>
+
+Next choose one of the following two options to specify to the inmanta server where to get the modules:
+
+1. Use the pip config file
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Hugo-Inmanta, line 350:
```suggestion
1. Use the pip config file
^^^^^^^^^^^^^^^^^^^^^^^^^^
```

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,61 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+
+In the ``project.yml`` (see: :ref:`project_yml`) file of your project, ensure that the ``use_config_file`` option under the ``pip`` section is set to ``True``.
+
+.. code-block:: text
+
+    pip:
+        use_config_file: True
+
+Then add the following content the global pip config file:
```
Hugo-Inmanta, line 359:
```suggestion
Then add the following content to the global pip config file:
```

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,61 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+  [global]
+  timeout = 60
+  index-url = <url of the repository >
+
+.. note::
+
+    You can also create a new pip config file ``pip.conf``, add the content there and then set the ``PIP_CONFIG_FILE`` environment variable in the inmanta server to the path of the newly created ``pip.conf`` file (See: :ref:`env_vars`).
+
+2. Set the index_urls in the project
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Hugo-Inmanta, line 372:
```suggestion
2. Set the index_urls in the project
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,61 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
 The only exception to this rule is when using the ``inmanta export`` command. It exports a project and all its modules'
 ``inmanta_plugins`` packages to the orchestrator server. When this method is used, the orchestrator does not install any modules
 from the Python package repository but instead contains all Python code as present in the local Python environment.
+
+Configure a project to install modules from a private python package repository
+--------------------------------------------------------------------------------
```
Hugo-Inmanta, line 334:
```suggestion
Configure a project to install modules from a private python package repository
-------------------------------------------------------------------------------
```

Hugo-Inmanta (changes requested):

docs/model_developers/modules.rst
arnaudsjs: `token` is the username that is used by Cloudsmith when you use an entitlement token. I don't think we should document the procedure for Cloudsmith here. Customers probably want to connect to their own internal Python package repository that requires authentication, so I would keep it generic.
```suggestion
  login <username>
```

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,61 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+V2 modules are Python packages that can be installed from private Python package repositories, which may require authentication.This section explains the steps to install such modules.
+
+Create a file named ``/var/lib/inmanta/.netrc`` in the orchestrator's file system.
+Add the following content to the file:
+
+.. code-block:: text
+
+  machine <url of the private repository>
+  login token
+  password <the token to access the private repository>
```
arnaudsjs, line 345:
```suggestion
  password <password>
```

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,61 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+
+Next choose one of the following two options to specify to the inmanta server where to get the modules:
+
+1. Use the pip config file
```
arnaudsjs, line 349: Now that I think about it, the two section mentioned below describe the difference between 1) using a index URL defined in the global pip config and 2) using a index URL defined in the project.yml file. This is completely unrelated to whether or not the Python package repository requires authentication. I see that it's interesting to mention this information somewhere in the documentation, but I don't think it should be part of this section.

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,61 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
 The only exception to this rule is when using the ``inmanta export`` command. It exports a project and all its modules'
 ``inmanta_plugins`` packages to the orchestrator server. When this method is used, the orchestrator does not install any modules
 from the Python package repository but instead contains all Python code as present in the local Python environment.
+
+Configure a project to install modules from a private python package repository
```
arnaudsjs, line 333: We don't really configure the project here. The configuration of the project is independent from whether or not authentication is required on the remote Python package repository. It's the configuration of the server that has to be updated.
```suggestion
Configure the Inmanta server to install modules from a private python package repository
```

arnaudsjs (changes requested):

docs/model_developers/modules.rst
FloLey: I changed the location but i'm still not sure it is the best place. what do you think?

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,23 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+Create a file named ``/var/lib/inmanta/.netrc`` in the orchestrator's file system.
+Add the following content to the file:
+
+.. code-block:: text
+
+  machine <hostname of the private repository>
+  login <username>
+  password <password>
+
+You will also need to specify the url of the repository to the inmanta server. This can be done in the ``project.yml`` file of your project. (See: :ref:`specify_location_pip`)
```
arnaudsjs, line 347:
```suggestion
You will also need to specify the url of the repository in the ``project.yml`` file of your project. (See: :ref:`specify_location_pip`)
```

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,23 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+
+By following these steps, you will be able to install modules from a private Python package repository.
```
arnaudsjs, line 349:
```suggestion
By following the previous steps, the Inmanta server will be able to install modules from a private Python package repository.
```

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,34 @@ The code snippet below provides an example of a complete ``project.yml`` file:
       index_urls:
         - https://pypi.org/simple/
 
+
+.. _specify_location_pip:
+
+Specify locations from where V2 modules will be installed
```
arnaudsjs, line 60: This section needs a lot more details and structure if we want to make sure that it's clear to the end user. Or we can also drop it for now and create a ticket for this, because it's technically outside of the scope of this ticket.

* We should split this in sub-sections for each of the different alternatives with a section above these three sub-section that gives an short overview of the three different approaches.
* We should include snippets when we refer to setting a specific config option.

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,34 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+
+  [global]
+  timeout = 60
+  index-url = <url of the repository >
+
+.. note::
+
+    You can also create a new pip config file ``pip.conf``, add the content there and then set the ``PIP_CONFIG_FILE`` environment variable to the path of the newly created ``pip.conf`` file (See: :ref:`env_vars`).
+
+Another option is to use the  ``index_urls`` option. There you can directly add the urls of repositories you are interested in.
```
arnaudsjs, line 76: We should make it clear that the `index_urls` config option is a config option of the project.yml file. Maybe also refer to it as `pip.index_urls`, because index_urls is a sub-config option of the pip section.

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,34 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+
+
+.. note::
+
+    Both methods can also be used together.
```
arnaudsjs, line 81: If we mention this, we should also mention what the behavior is if you do this. Namely, which one takes precedence.

arnaudsjs (changes requested):

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,34 @@ The code snippet below provides an example of a complete ``project.yml`` file:
       index_urls:
         - https://pypi.org/simple/
 
+
+.. _specify_location_pip:
+
+Specify locations from where V2 modules will be installed
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Hugo-Inmanta, line 61:
```suggestion
Specify locations from where V2 modules will be installed
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

docs/reference/projectyml.rst
Hugo-Inmanta, line 76:
```suggestion
Another option is to set the urls of the repositories in the ``index_urls`` option of the ``pip`` section.
```

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,34 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+
+By setting ``use_config_file`` option to ``True``, the project will use the global pip config file when installing V2 modules.
```
Hugo-Inmanta, line 63:
```suggestion
By setting the ``use_config_file`` option of the ``pip`` section to ``True``, the project will use the global pip config file when installing V2 modules.
```

Hugo-Inmanta (changes requested):

docs/reference/projectyml.rst
FloLey: I will add it now.

docs/reference/projectyml.rst
FloLey: you mention in your previous 3 options, this is the third one right?

arnaudsjs: I was thinking: 1) global pip config, 2) specific custom pip config, 3) index urls in project.yml file. But that's a detail. The most important part is that the different options are clearly structured.

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,66 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+  timeout = 60
+  index-url = <url of a python package repository>
+  extra-index-url = <url of a second python package repository>
+                    <url of a third python package repository>
+
+Creating a new pip config file
+""""""""""""""""""""""""""""""
+
+Alternatively, a new ``pip.conf`` can be created and the ``index-url`` can be specified there as explained in the previous section.
+To make this work, the ``PIP_CONFIG_FILE`` environment variable needs to be set to the path of the newly created ``pip.conf`` file (See: :ref:`env_vars`).
```
arnaudsjs, line 91:
```suggestion
Alternatively, a pip config file can be used at a custom location.
The ``index-url`` can be specified in this file as explained in the previous section.
To make this work, the ``PIP_CONFIG_FILE`` environment variable needs to be set to the path of the newly created file (See: :ref:`env_vars`).
```

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,66 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+.. _specify_location_pip:
+
+
+Specify locations from where V2 modules will be installed
+---------------------------------------------------------
+This section explains how to configure your project in order to download v2 modules from any python package repository.
+By default, the project is configured to install packages from ``https://pypi.org/simple/``. There are multiple ways
+to change this behaviour.
+
+Using the global pip config file
```
arnaudsjs, line 67: The global config file is just one type of pip config file, there are multiple of them (https://pip.pypa.io/en/stable/topics/configuration/#configuration-files)
```suggestion
Using pip config file at a standard location
```

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,66 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+      index_urls:
+        - <url of a python package repository>
+        - <url of a second python package repository>
+
+
+.. note::
+    The pip config file can also be used in combination with ``index-urls`` specified in the ``pip`` section of the ``project.yml`` file:
+
+    * If the pip config is used (by setting ``use_config_file`` to ``True``), the ``index-url`` specified in the pip config file will take precedence and the ``index-urls`` specified in the ``pip`` section of the ``project.yml`` file will be used as ``extra-index-urls`` when installing with pip.
+    * If the pip config is not used (by setting ``use_config_file`` to False), then the first ``index_url`` specified in the project.yml will be used as an ``index_url`` and all the following ones will be used as ``extra-index-urls`` when installing with pip.
```
arnaudsjs, line 111: For consistency with the previous bullet point.

```suggestion
    * If the pip config is not used (by setting ``use_config_file`` to ``False``), then the first ``index_url`` specified in the project.yml will be used as an ``index_url`` and all the following ones will be used as ``extra-index-urls`` when installing with pip.
```

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,66 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+Specify locations from where V2 modules will be installed
+---------------------------------------------------------
+This section explains how to configure your project in order to download v2 modules from any python package repository.
+By default, the project is configured to install packages from ``https://pypi.org/simple/``. There are multiple ways
+to change this behaviour.
+
+Using the global pip config file
+""""""""""""""""""""""""""""""""
+
+By setting the ``use_config_file`` option of the pip section to ``True``, the project will use the global pip config file.
```
arnaudsjs, line 70:
```suggestion
By setting the ``use_config_file`` option of the pip section to ``True``, the project will use the pip config files.
```

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,66 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+
+.. code-block:: text
+
+  [global]
+  timeout = 60
+  index-url = <url of a python package repository>
+  extra-index-url = <url of a second python package repository>
+                    <url of a third python package repository>
+
+Creating a new pip config file
```
arnaudsjs, line 87:
```suggestion
Creating a pip config file at a custom location
```

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,66 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+""""""""""""""""""""""""""""""
+
+Alternatively, a new ``pip.conf`` can be created and the ``index-url`` can be specified there as explained in the previous section.
+To make this work, the ``PIP_CONFIG_FILE`` environment variable needs to be set to the path of the newly created ``pip.conf`` file (See: :ref:`env_vars`).
+
+Specify the locations in the project.yml file
```
arnaudsjs, line 93:
```suggestion
Specify the index-urls in the project.yml file
```

arnaudsjs (changes requested):

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,23 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+Create a file named ``/var/lib/inmanta/.netrc`` in the orchestrator's file system.
+Add the following content to the file:
+
+.. code-block:: text
+
+  machine <hostname of the private repository>
+  login <username>
+  password <password>
+
+You will also need to specify the url of the repository in the ``project.yml`` file of your project. (See: :ref:`specify_location_pip`)
```
Hugo-Inmanta, line 347:
```suggestion
You will also need to specify the url of the repository in the ``project.yml`` file of your project (See: :ref:`specify_location_pip`).
```

Hugo-Inmanta (changes requested):

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,66 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
 
+
+.. _specify_location_pip:
+
+
+Specify locations from where V2 modules will be installed
+---------------------------------------------------------
+This section explains how to configure your project in order to download v2 modules from any python package repository.
+By default, the project is configured to install packages from ``https://pypi.org/simple/``. There are multiple ways
+to change this behaviour.
```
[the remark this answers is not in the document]
FloLey: what I meant is that in the template of a project it is configured with ``https://pypi.org/simple/`` as the default.

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,67 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+
+
+Specify locations from where V2 modules will be installed
+---------------------------------------------------------
+This section explains how to configure your project in order to download v2 modules from any python package repository.
+By default, a project created using the :ref:`project-creation-guide` is configured to install packages from ``https://pypi.org/simple/``. There are multiple ways
+to change this behaviour.
+
+Using pip config file at a standard location
+""""""""""""""""""""""""""""""""
```
arnaudsjs, line 68: Do you make sure that the length of the title and the length of the title marker match. At the moment we get many ugly warnings regarding this in the compile log.
```suggestion
Using pip config file at a standard location
""""""""""""""""""""""""""""""""""""""""""""""""
```

arnaudsjs (changes requested):

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,67 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+
+To specify the url of a pip repository, add the following to the pip config file:
+
+.. code-block:: text
+
+  [global]
+  timeout = 60
+  index-url = <url of a python package repository>
+  extra-index-url = <url of a second python package repository>
+                    <url of a third python package repository>
```
sanderr, line 85: I would drop the extra index urls because we strongly discourage it, I would not give the impression this is a typical approach

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,67 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+
+Creating a pip config file at a custom location
+""""""""""""""""""""""""""""""
+
+Alternatively, a pip config file can be used at a custom location.
+The ``index-url`` can be specified in this file as explained in the previous section.
+To make this work, the ``PIP_CONFIG_FILE`` environment variable needs to be set to the path of the newly created file (See: :ref:`env_vars`).
```
sanderr, line 92: Wouldn't it be easier to just refer to the appropriate page on the pip documentation?

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,67 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+      use_config_file: False
+      index_urls:
+        - <url of a python package repository>
+        - <url of a second python package repository>
+
+
+.. note::
+    The pip config file can also be used in combination with ``index-urls`` specified in the ``pip`` section of the ``project.yml`` file:
+
+    * If the pip config is used (by setting ``use_config_file`` to ``True``), the ``index-url`` specified in the pip config file will take precedence and the ``index-urls`` specified in the ``pip`` section of the ``project.yml`` file will be used as ``extra-index-urls`` when installing with pip.
```
sanderr, line 111:
```suggestion
    * If the pip config is used (by setting ``use_config_file`` to ``true``), the ``index-url`` specified in the pip config file will take precedence and the ``index-urls`` specified in the ``pip`` section of the ``project.yml`` file will be used as ``extra-index-urls`` when installing with pip.
```

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,67 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+
+Another option is to use the  ``index_urls`` option in the ``pip`` section of the ``project.yml`` file:
+
+.. code-block:: yaml
+
+    pip:
+      use_config_file: False
+      index_urls:
+        - <url of a python package repository>
+        - <url of a second python package repository>
```
sanderr, line 105: Same here

FloLey: what do you mean here? I mean we still want to tell the user that he can provide multiple urls in the project.yml file right? otherwise I don't like that we called the variable 'urls' and not 'url'

sanderr: I mean, it's a list so any user who cares to have more than one will see it, but I am strongly opposed to advertising it any more than we need to. As far as our examples go, I'm of the opinion they should all use a single-element list here. If you (or one of the reviewers) disagree(s) you can raise it in the planning meeting tomorrow and we can figure out what we want there.

docs/reference/projectyml.rst
FloLey: I added the link to the docs but I didn't remove anything. I find it relevant to refer to our env_vars doc as I didn't now how to set them when I did the testing.

docs/reference/projectyml.rst
FloLey: Ok I see what you mean

arnaudsjs (approved):

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,63 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+"""""""""""""""""""""
+
+By setting the ``use_config_file`` option of the pip section to ``True``, the project will use the pip config files.
+
+.. code-block:: yaml
+
+    pip:
+      use_config_file: True
+
+To specify the url of a pip repository, add the following to the pip config file:
```
edvgui, line 77: If you show the content of the file, shouldn't you also specify the path to that file? I know there is a link to the full doc below.  But as a user, I think I would be a bit frustrated to see the simple step that needs to be done in here, and actually have to click one more link and search through a different doc to find the place where that change needs to be applied.

docs/model_developers/project_creation.rst
```diff
@@ -44,12 +44,12 @@ An example ``project.yml`` could be:
   repo:
       - url: https://github.com/inmanta/
         type: git
-      - url: https://pypi.org/simple
-        type: package
```
edvgui, line 48: Is this way of working deprecated then? Can it still work in combination with the new behavior? Or is it one or the other?

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,23 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
 The only exception to this rule is when using the ``inmanta export`` command. It exports a project and all its modules'
 ``inmanta_plugins`` packages to the orchestrator server. When this method is used, the orchestrator does not install any modules
 from the Python package repository but instead contains all Python code as present in the local Python environment.
+
+Configure the Inmanta server to install modules from a private python package repository
+----------------------------------------------------------------------------------------
+
+V2 modules can be installed from a Python package repository that requires authentication. This section explains how the Inmanta server should be configured to install v2 modules from such a Python package repository.
+
+Create a file named ``/var/lib/inmanta/.netrc`` in the orchestrator's file system.
```
edvgui, line 338: I would reference the documentation of netrc here.

edvgui (changes requested):

docs/reference/projectyml.rst
FloLey: do you mean the location of the pip config file? because if I understand correctly it could be located at multiple places.

docs/model_developers/project_creation.rst
FloLey: This PR deprecated it: https://github.com/inmanta/inmanta-core/commit/c79cc1e1376a13367fa2a3c1949cd4b864ea7fdd
@Hugo-Inmanta do you know if both can be used together?

docs/reference/projectyml.rst
edvgui: That is what I meant yes, I ever used it so I don't know how it works.  Which also means that I wouldn't know what to do for this step.

edvgui: I guess there is a user level pip config file? Maybe you can advice to edit this one?

FloLey: Ok i'll do that :)

docs/reference/projectyml.rst
```diff
@@ -54,6 +54,63 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+"""""""""""""""""""""
+
+By setting the ``use_config_file`` option of the pip section to ``True``, the project will use the pip config files.
+
+.. code-block:: yaml
+
+    pip:
+      use_config_file: True
+
+To specify the url of a pip repository, add the following to the pip config file located at ``$HOME/.config/pip/pip.conf``:
```
edvgui, line 77:
```suggestion
To specify the url of a pip repository, add the following to the pip config file of the ``inmanta`` user, located at ``/var/lib/inmanta/.config/pip/pip.conf``:
```

docs/model_developers/project_creation.rst
Hugo-Inmanta: If I understand this correctly:
- on **master** a warning is raised when using an `url` with type `package` but it can be used in combination with the new behaviour
- on **iso6** we are currently reverting the pip index PR's (see discussion [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1688132643347189))
And a separate documentation PR should be opened for iso6 specifically (see Sander's comment [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1688394987395209?thread_ts=1688394690.058399&cid=CKRF0C8R3)

Hugo-Inmanta (approved):

docs/model_developers/modules.rst
```diff
@@ -329,3 +329,25 @@ for supplying the agents with the appropriate ``inmanta_plugins`` packages.
...
+Add the following content to the file:
+
+.. code-block:: text
+
+  machine <hostname of the private repository>
+  login <username>
+  password <password>
+
+for more information see the  doc about`pip authentication <https://pip.pypa.io/en/stable/topics/authentication/>`_.
```
edvgui, line 347:
```suggestion
For more information see the doc about`pip authentication <https://pip.pypa.io/en/stable/topics/authentication/>`_.
```

edvgui (approved):
One last minor comment

PR #6196, opened 2023-06-29
Add support to expose API methods using multiple URLs (iso6)

arnaudsjs:
# Description

**This is the same PR as #6191 but applied on the iso6 branch due to a merge conflict.**

Add support to expose API methods using multiple URLs. If the path property of the typedmethod annotation has type list, the inmanta client will always use the first path to perform its API calls.

closes  inmanta/inmanta-lsm#1274

# Self Check:

- [x] Changelog entry
- [ ] ~~If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 8.5.0
+current_version = 8.6.0
 
 [bumpversion:file:setup.py]
 search = version = "{current_version}"
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 8.6.0
+current_version = 8.5.0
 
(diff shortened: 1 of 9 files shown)


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 8.5.0
+current_version = 8.6.0
```
sanderr, line 2: As noted on [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1688024923128089) it appears we've already been a bit too eager in bumping this.

sanderr (changes requested):

sanderr (approved):

PR #6201, opened 2023-07-03
Prevent repair loops

### LINKED ISSUE #6202: Prevent repair loops
# Problem

1. incremental deploys interrupt repairs (unless the increment is empty)
2. if you have resources that are in a failing state (and can't be fixed by the orchestrator), the increment will never be empty
3. if the deploy interval is smaller than the time it takes to run a full repair
4. we get into an infinite retry loop

# Related

When looking at the log, this looks just like https://github.com/inmanta/inmanta-core/issues/5434
But it is different in what causes it. 

# Work Around

1. Increase deploy interval to be larger than one full reapir
2. Make sure heave loaded agents have no failing resources
3. Split out agent with many resources over multiple agent to shorten repair time

wouterdb:
# Description
 
Prevent repair loops by ignoring period deploys if a periodic deploy is running 

closes #6202 

# Self Check:

- [x] Changelog entry


tests/agent_server/test_server_agent.py
```diff
@@ -2489,29 +2500,30 @@ def get_resources(version, value_resource_three):
     version2 = await clienthelper.get_version()
     resources_version_2 = get_resources(version2, "value3")
     await _deploy_resources(client, environment, resources_version_2, version2, False)
-    await myagent_instance.get_latest_version_for_agent(reason="Second Deploy", incremental_deploy=False, is_repair_run=False)
```
wouterdb, line 2492: False/False was  a bug I think

tests/agent_server/test_server_agent.py
```diff
@@ -2563,7 +2575,9 @@ def get_resources(version, value_resource_three):
 
     # Initial deploy
     await _deploy_resources(client, environment, resources_version_1, version1, False)
-    await myagent_instance.get_latest_version_for_agent(reason="Initial Deploy", incremental_deploy=False, is_repair_run=False)
```
wouterdb, line 2566: False/False was a bug I think

src/inmanta/agent/agent.py
```diff
@@ -334,6 +336,78 @@ def notify(
...
+    :param is_periodic: is this deploy triggered by a timer?
+    :param reason: textual description of the deployment
+    """
+
+    is_repair: bool
+    is_periodic: bool
+    reason: str
+
+    def interrupt(self, other: "DeployRequest") -> "DeployRequest":
+        """Interrupt this deploy and for the other and produce a new request"""
```
sanderr, line 354:
```suggestion
        """Interrupt this deploy and for the other and produce a new request for future rescheduling of this deploy"""
```

changelogs/unreleased/6202-repair-loop.yml
```diff
@@ -0,0 +1,6 @@
+description: Prevent repairs from restarting indefinitely.
```
sanderr, line 1: If we include it in the changelog perhaps we should make it sound a bit less alarming by specifying when it triggers?
```suggestion
description: Prevent repairs from restarting indefinitely when a short deploy interval is set
```

src/inmanta/agent/agent.py
```diff
@@ -334,6 +336,78 @@ def notify(
             self.future.set_result(ResourceActionResult(cancel=False))
 
 
+@dataclasses.dataclass
+class DeployRequest:
+    """
+    A request to perform a deploy
+
+    :param is_repair: is this a full deploy/repair run or incremental?
```
sanderr, line 344: The name does not seem to correspond with the docstring.

src/inmanta/agent/agent.py
```diff
@@ -334,6 +336,78 @@ def notify(
...
+
+# This matrix describes what do when a new DeployRequest enters before the old one is done
+# Format is old_is_repair, old_is_periodic, new_is_repair, new_is_periodic
+# The underlying idea is that
+# 1. periodic deploys have no time pressure, they can be delayed
+# 2. non-periodic deploy should run as soon as possible
+# 3. non-periodic incremental deploy take precedence over repairs (as they are smaller)
+# 4. periodic deploys should not interrupt each other to prevent restart loops
+deploy_response_matrix = {
+    # Periodic restart loops: Full periodic is never interrupted by periodic
```
sanderr, line 383:
```suggestion
    # (old_is_repair, old_is_periodic, new_is_repair, new_is_periodic)
    # Periodic restart loops: Full periodic is never interrupted by periodic
```
I think it's more readable to repeat it here.

src/inmanta/agent/agent.py
```diff
@@ -334,6 +336,78 @@ def notify(
...
+
+
+# This matrix describes what do when a new DeployRequest enters before the old one is done
+# Format is old_is_repair, old_is_periodic, new_is_repair, new_is_periodic
+# The underlying idea is that
+# 1. periodic deploys have no time pressure, they can be delayed
+# 2. non-periodic deploy should run as soon as possible
+# 3. non-periodic incremental deploy take precedence over repairs (as they are smaller)
+# 4. periodic deploys should not interrupt each other to prevent restart loops
+deploy_response_matrix = {
```
sanderr, line 382: I still have to review this matrix, but it depends heavily on the meaning of `is_repair`, for which I left a comment on the docstring.

sanderr (commented):
Intermediate comments

src/inmanta/agent/agent.py
sanderr: Ah, you mean it's either
- a repair or a full deploy
- an incremental deploy
Is that right? I find the "deploy/repair" part difficult to parse. I read it as either
- a full deploy or a full repair
- an incremental deploy or an incremental repair

Even though that doesn't make sense if you know the fine points of deploys and repairs, when you just read the docstring I think the mistake is easily made. I think renaming it `is_full`/`is_incremental` and rewording the docstring to be less ambiguous would make it all more readable.

wouterdb: updated

src/inmanta/agent/agent.py
sanderr, line 382: I found this very difficult to interpret. I had to convert it to conditions to be able to see the full picture. If I understand correctly you started out with a conditional implementation and then converted it to this matrix. Do you find it more readable? What if we split up the conditions for which takes precedence and whether or not the non-preferred needs to be run afterwards? e.g. (bad naming and docstrings, some more cleanup needed)
```python
    def take_action(self, other: "DeployRequest") -> tuple["DeployRequestAction", Optional["DeployRequest"]]:
        """
        Returns the action to take and the deploy request to schedule afterwards, if appliccable
        """
        carry_on: bool = self.pick(other)
        if not carry_on:
            if self.run_after(other):
                return DeployRequestAction.interrupt, self.interrupt(other)
            else:
                return DeployRequestAction.terminate, None
        if other.run_after(self):
            return DeployRequestAction.defer, other
        else:
            return DeployRequestAction.ignore, None

    def pick(self, other: "DeployRequest") -> bool:
        """
        Return True iff self should be preferred over other.
        """
        if self.is_periodic:
            return (
                # if both are periodic, pick the full one
                self.is_repair
                if (
                    other.is_periodic
                    # if neither is full, pick the new one
                    and self.is_repair != other.is_repair
                )
                # otherwise prefer non-periodic
                else False
            )
        elif other.is_periodic:
            # periodic may only interrupt full if both are incremental
            return not self.is_repair and not other.is_repair
        elif self.is_repair != other.is_repair:
            # neither is periodic: incremental takes precedence over full
            return not self.is_repair
        else:
            # pick newest when both are incremental or both are full
            return False

    def run_after(self, other: "DeployRequest") -> bool:
        """
        Should self still run after other has finished?
        """
        # TODO: this is tightly coupled to the logic in `pick`
        return (
            # incremental can always be skipped
            self.is_repair
            # no need to run if other was already a full deploy
            and not other.is_repair
        )
```

wouterdb: I think this approach has some advantages:

1. it is clear that all cases are covered. 
2. it is clear how each case is covered 
3. some properties are very easy to see (e.g. the jobs that can be resumed are all full repairs)

What drove me to this is that the 4 level deep if statement allows for some very nasty bugs, like confusing `is_repair` with `self.is_repair`. There are so many code paths that contain similar code that testing against mixups is very hard. Tests must traverse all branches.  Here, there is basically one code path, so every mistake is systemic.

wouterdb: Perhaps we have to compact the tuples to 

`(periodic_full, increment)`
or
`(PF,I)`

sanderr: For 1-3 we could have a test case that builds the full matrix and asserts it is equal to the one you built by hand (I meant to include that suggestion in my earlier comment but it seems I forgot). In my opinion that would give us the best of both worlds: code with readable decision drivers, and a matrix for verifying specific cases.

tests/agent_server/test_server_agent.py
```diff
@@ -2573,26 +2598,179 @@ def get_resources(version, value_resource_three):
...
+    resource_container.Provider.reset()
+    config.Config.set("config", "agent-deploy-interval", "0")
+    config.Config.set("config", "agent-repair-interval", "0")
+    agent_name = "agent1"
+
+    myagent_instance = agent._instances[agent_name]
+
+    resource_container.Provider.set("agent1", "key1", "value1")
+    resource_container.Provider.set("agent1", "key1", "value1")
+    resource_container.Provider.set("agent1", "key1", "value1")
```
arnaudsjs, line 2699:
```suggestion
    resource_container.Provider.set("agent1", "key1", "value1")
    resource_container.Provider.set("agent1", "key2", "value1")
    resource_container.Provider.set("agent1", "key3", "value1")
```

tests/agent_server/test_server_agent.py
```diff
@@ -2573,26 +2598,179 @@ def get_resources(version, value_resource_three):
...
+                "purged": False,
+                "requires": ["test::Wait[agent1,key=key2],v=%d" % version],
+            },
+        ]
+
+    version1 = await clienthelper.get_version()
+    resources_version_1 = get_resources(version1, "value2")
+
+    # Initial deploy
+    await _deploy_resources(client, environment, resources_version_1, version1, False)
```
arnaudsjs, line 2733:
```suggestion
    await _deploy_resources(client, environment, resources_version_1, version1, push=False)
```

tests/agent_server/test_server_agent.py
```diff
@@ -2573,26 +2598,179 @@ def get_resources(version, value_resource_three):
...
+    timeout_time = time.time() + 10
+    while (await client.get_version(environment, version1)).result["model"]["done"] < 1 and time.time() < timeout_time:
+        await asyncio.sleep(0.1)
+
+    # cache has 1 version in flight
+    assert len(myagent_instance._cache.counterforVersion) == 1
+
+    version2 = await clienthelper.get_version()
+    resources_version_2 = get_resources(version2, "value3")
+    await _deploy_resources(client, environment, resources_version_2, version2, False)
```
arnaudsjs, line 2748:
```suggestion
    await _deploy_resources(client, environment, resources_version_2, version2, push=False)
```

tests/agent_server/test_server_agent.py
```diff
@@ -2573,26 +2598,179 @@ def get_resources(version, value_resource_three):
...
+    await myagent_instance.get_latest_version_for_agent(
+        DeployRequest(reason="Second Deploy", is_full_deploy=second_full, is_periodic=second_periodic)
+    )
+
+    versions = [version1, version2]
+    await resource_container.wait_for_done_with_waiters(client, environment, versions[action.wait_for])
+    await asyncio.sleep(1)
```
arnaudsjs, line 2755: Why is this sleep required?

tests/agent_server/test_server_agent.py
```diff
@@ -2573,26 +2598,179 @@ def get_resources(version, value_resource_three):
...
+        [False, True, True, True, terminate, (2, 1, 1, 1)],  # Incremental periodic terminated by full periodic
+        [False, True, True, False, terminate, (2, 1, 1, 1)],  # Incremental periodic terminated by full
+        [False, False, True, True, defer, (2, 1, 2, 2)],  # Incremental defers full periodic
+        [False, False, True, False, defer, (2, 1, 2, 2)],  # Incremental defers full
+        # requires more complex waiting see test_s_repair_interrupted_by_deploy_request
+        # [True, False, False, False, interrupt, (2,1,2,1)],  # full interrupted by increment
+        # [True, True, False, False, interrupt,(2,1,2,1)], # periodic full interrupted by increment
+    ],
+)
+async def test_s_periodic_Vs_full(
```
arnaudsjs, line 2673: Nice test case!

changelogs/unreleased/6202-repair-loop.yml
```diff
@@ -0,0 +1,6 @@
+description: Prevent repairs from restarting indefinitely when a short deploy interval is set
+issue-nr: 6202
+change-type: minor
```
arnaudsjs, line 3: Shouldn't this be a patch?

src/inmanta/agent/agent.py
```diff
@@ -334,6 +336,79 @@ def notify(
...
+    """ interrupt the current deploy: cancel it, start the new one and restart the current deploy when the new one is done"""
+
+
+# This matrix describes what do when a new DeployRequest enters before the old one is done
+# Format is old_is_repair, old_is_periodic, new_is_repair, new_is_periodic
+# The underlying idea is that
+# 1. periodic deploys have no time pressure, they can be delayed
+# 2. non-periodic deploy should run as soon as possible
+# 3. non-periodic incremental deploy take precedence over repairs (as they are smaller)
+# 4. periodic deploys should not interrupt each other to prevent restart loops
```
arnaudsjs, line 381:
```suggestion
# 4. periodic deploys should not interrupt each other to prevent restart loops
# 5. Periodic repairs take precedence over periodic incremental deploys.
```

arnaudsjs (changes requested):

src/inmanta/agent/agent.py
```diff
@@ -398,41 +466,42 @@ def reload(
...
+            # Get correct action
+            response = deploy_response_matrix[
+                (self.running.is_full_deploy, self.running.is_periodic, new_request.is_full_deploy, new_request.is_periodic)
+            ]
+            # Execute action
+            if response == DeployRequestAction.terminate:
+                self.logger.info("Terminating run '%s' for '%s'", self.running.reason, new_request.reason)
+            elif response == DeployRequestAction.defer:
+                self.logger.info("Deferring run '%s' for '%s'", new_request.reason, self.running.reason)
+                self.deferred = new_request
```
sanderr, line 482: What if we already have a request queued? Same for the interrupt action. I think it may be safe because we only ever defer/interrupt full deploys, therefore we can just replace them (possibly excepting the message). But if this is the case, we should probably make it an invariant of the matrix and add a comment here as to why it's safe to replace `self.deferred`.

sanderr (changes requested):

src/inmanta/agent/agent.py
wouterdb: I also don't find the branching code particularly understandable (the table is not very understandable either), so readability is not a strong argument to me.

changelogs/unreleased/6202-repair-loop.yml
wouterdb: I was in doubt myself, it slightly changes the behavior. 

I you think patch is more appropriate, I will change it.

arnaudsjs: I would have chosen patch here, but I also don't have any objections against using minor instead.

src/inmanta/agent/agent.py
wouterdb: added docs and a testcase

tests/agent_server/test_server_agent.py
wouterdb: it is not, thx

src/inmanta/agent/agent.py
```diff
@@ -372,40 +372,58 @@ class DeployRequestAction(str, enum.Enum):
...
 
+# Two letter abbreviations to make table line up
+# N normal
+# P periodic
+# F full
+# I incremental
+NF = (True, False)
+NI = (False, False)
+PF = (True, True)
+PI = (False, True)
```
sanderr, line 383: Slight preference to swap the letters so their order matches the order in the tuple.

sanderr, line 383: I like naming them, though I would prefer non-abbreviated names: `normal_full`, `periodic_increment`, ... I think they are sufficiently concise that it would only help readability. Now I find myself constantly having to translate letters into words into concepts. No strong preference, I find I do get used to it after a few minutes.

src/inmanta/agent/agent.py
```diff
@@ -372,40 +372,58 @@ class DeployRequestAction(str, enum.Enum):
...
+interrupt = DeployRequestAction.interrupt
+
 # This matrix describes what do when a new DeployRequest enters before the old one is done
-# Format is old_is_repair, old_is_periodic, new_is_repair, new_is_periodic
+# Format is (old_is_repair, old_is_periodic), (new_is_repair, new_is_periodic)
 # The underlying idea is that
 # 1. periodic deploys have no time pressure, they can be delayed
 # 2. non-periodic deploy should run as soon as possible
 # 3. non-periodic incremental deploy take precedence over repairs (as they are smaller)
 # 4. periodic deploys should not interrupt each other to prevent restart loops
```
sanderr, line 396: This rule is contradicted by 4. As far as I can see it is the only rule that states a hard condition that is later contradicted. I think it would be good if we make this more clear, e.g. swap 4 and 5 and extend this one with something like "... each other except to satisfy one of the rules above, to prevent ..."

src/inmanta/agent/agent.py
```diff
@@ -372,40 +372,58 @@ class DeployRequestAction(str, enum.Enum):
...
+# shorten table
+ignore = DeployRequestAction.ignore
+terminate = DeployRequestAction.terminate
+defer = DeployRequestAction.defer
+interrupt = DeployRequestAction.interrupt
+
 # This matrix describes what do when a new DeployRequest enters before the old one is done
-# Format is old_is_repair, old_is_periodic, new_is_repair, new_is_periodic
+# Format is (old_is_repair, old_is_periodic), (new_is_repair, new_is_periodic)
 # The underlying idea is that
```
sanderr, line 392: This list defines the logic for deciding which of two deploy requests gets precedence (`pick` in the branching PoC) but it doesn't define under what circumstances the other should be deferred or just skipped (`run_after` in the branching PoC).

src/inmanta/agent/agent.py
```diff
@@ -372,40 +372,58 @@ class DeployRequestAction(str, enum.Enum):
...
 # 1. periodic deploys have no time pressure, they can be delayed
 # 2. non-periodic deploy should run as soon as possible
 # 3. non-periodic incremental deploy take precedence over repairs (as they are smaller)
 # 4. periodic deploys should not interrupt each other to prevent restart loops
+# 5. Periodic repairs take precedence over periodic incremental deploys.
```
sanderr, line 397: I think the fallback rule is still missing: if none of these conditions match, meaning `(NF, NF)` or `(NI, NI)`, we pick the newest. Or do you see this as a consequence of rule 2?

src/inmanta/agent/agent.py
sanderr, line 392: I think the `(NI, PI)` case is not covered. According to rules 1 and 2 I would conclude ignore/defer, but we do in fact terminate (rightfully so I think). Perhaps a rule in the vein of "non-periodic incremental deploy may be terminated for other (periodic or not) incremental deploy)"?

changelogs/unreleased/6202-repair-loop.yml
sanderr: > I would have chosen patch here, but I also don't have any objections against using minor instead.

I side with this comment.

arnaudsjs (approved):

src/inmanta/agent/agent.py
wouterdb: The problem is that I need 
1. short names or black will mess it up
2. names with equal length (or it doesn't line up) 

If you can suggest better names, I'm happy

sanderr: Those are good points, let's leave it as is then.

sanderr (approved):

PR #6203, opened 2023-07-03
issue/6123-Agent incorrectly rewrites requirements files

### LINKED ISSUE #6123: Agent incorrectly rewrites requirements files
When the agent installs the Python dependencies of handler code, it calls into the `_gen_content_requirements_file()` method to transform the requirements from pep440 format to a format that pip understands. However, it does this incorrectly when the same dependency is present more than once with different environment markers.

Example:

```
    content: str = env.ActiveEnv._gen_content_requirements_file(
        [
            "ciscoconfparse==1.6.50; python_version <= '3.6'",
            "ciscoconfparse",
            "lxml",
            "ncclient"
        ]
    )
    print(content)
```

Results in:

```
ciscoconfparse == 1.6.50 ; python_version <= "3.6"
lxml
ncclient
```

This way the dependency `ciscoconfparse` is not installed on servers that use a python version higher than 3.6, while it should.

FloLey:
# Description

When the agent installs the Python dependencies of handler code, it calls into the _gen_content_requirements_file() method to transform the requirements from pep440 format to a format that pip understands. However, it does this incorrectly when the same dependency is present more than once with different environment markers.
This pr fixes this issue

closes #6123 

# Self Check:

- [x] Changelog entry


src/inmanta/env.py
```diff
@@ -837,23 +834,24 @@ def _gen_content_requirements_file(cls, requirements_list: Sequence[str]) -> str
...
-                modules[name]["markers"].append(marker)
-
+                modules[requirement_id]["markers"].append(marker)
             if url is not None:
-                modules[name]["url"] = url
+                modules[requirement_id]["url"] = url
 
             if extras is not None:
-                modules[name]["extras"] = extras
+                modules[requirement_id]["extras"] = extras
```
arnaudsjs, line 850: This can also go wrong, because we set the list instead of appending to it. For example when the requirements.txt file contains:

```
dep[opt]
dep[otheropt]
```

It will results in:

```
dep[otheropt]
```

arnaudsjs (changes requested):

arnaudsjs (approved):

Hugo-Inmanta (approved):

PR #6212, opened 2023-07-04
issue/241-handle-max-clients-limit-in-agents

### LINKED ISSUE #241: Handle max-clients limit in agents
When the agent connects to the server, it can open many connections.

- [x] put heartbeat on a different connection pool than other agent communication
- [x] catch tornado's debug-level warning about exceeding max-client and reraise it as an inmanta warning
- [x] make max-clients setting configurable via config file (http://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.AsyncHTTPClient.configure)

FloLey:
# Description

When the agent connects to the server, it can open many connections.

- put heartbeat on a different connection pool than other agent communication
- catch tornado's debug-level warning about exceeding max-client and reraise it as an inmanta warning
- make max-clients setting configurable via config file

closes #241 

# Self Check:

- [x] Changelog entry


FloLey:
I still have some open questions in the PR and I'm not sure this is the way to handle the logging of the tornado logs.
Also I still have a mypy issue which is annoted in the 'files changed' tab

src/inmanta/protocol/rest/client.py
```diff
@@ -40,14 +40,18 @@ class RESTClient(RESTBase):
...
         super().__init__()
         self.__end_point: "Endpoint" = endpoint
         self.daemon: bool = True
         self.token: Optional[str] = inmanta_config.Config.get(self.id, "token", None)
         self.connection_timout: int = connection_timout
         self.headers: Set[str] = set()
         self.request_timeout: int = inmanta_config.Config.get(self.id, "request_timeout", 120)
+        max_clients: int = inmanta_config.Config.get(
+            "agent_rest_transport", "max_clients", 1000
+        )  # todo what should be the default here?
```
FloLey, line 53: here I have a mypy issue. It is also present for "request_timeout" but I don't know how to fix it.
```

+ src/inmanta/protocol/rest/client.py:51: error: No overload variant of "get" of "Config" matches argument types "str", "str", "int"  [call-overload]
+ src/inmanta/protocol/rest/client.py:51: note: Possible overload variants:
+ src/inmanta/protocol/rest/client.py:51: note:     def get(cls) -> ConfigParser
+ src/inmanta/protocol/rest/client.py:51: note:     def get(cls, section: str, name: str, default_value: Optional[str] = ...) -> Optional[str]
```

tests/test_agent_manager.py
```diff
@@ -1296,3 +1298,92 @@ def log_contains_debug_line():
...
+    await retry_limited(lambda: len(agent_manager.sessions) == 1, 10)
+
+    # Have many connections in flight
+    hangers = asyncio.gather(*(a._client.test_method(i) for i in range(5)))
+    logging.warning("WAITING!")
+    await asyncio.sleep(1)
+    assert not hangers.done()
+
+    def did_exceed_capacity():
+        return "WARNING max_clients limit reached, request queued. 1 active, 4 queued requests." in caplog.text
```
Hugo-Inmanta, line 1375: I would maybe replace this function with `log_contains(caplog, <logger>, <logger_level>, <msg>)` to assert the right logger is being used

src/inmanta/protocol/endpoints.py
```diff
@@ -191,11 +210,13 @@ async def start(self) -> None:
         """
         assert self._env_id is not None
         LOGGER.log(3, "Starting agent for %s", str(self.sessionid))
-        self._client = SessionClient(self.name, self.sessionid, timeout=self.server_timeout)
+        self._client = SessionClient(self.name, self.sessionid, timeout=self.server_timeout, force_instance=True)
+        self._heartbeat_client = SessionClient(self.name, self.sessionid, timeout=self.server_timeout, force_instance=True)
```
Hugo-Inmanta, line 214: I haven't looked too deep into this, but why is the force_instance parameter required ?

Hugo-Inmanta (changes requested):

src/inmanta/protocol/endpoints.py
FloLey: if you don't use the force_instance paramter the session clients use the same pool. the force_instance forces the new session to be use a different pool than the existing one.

Hugo-Inmanta: Oh ok I completely missed that part, I'll have a deeper look with a fresh pair of eyes on monday :+1:

tests/test_agent_manager.py
FloLey: unfortunately log_contains can not be used with retry_limited as it will assert false which will break out of the test :(

Hugo-Inmanta: Ah right, then maybe iterate over `caplog.get_records(...)` as it is done in log_contains to check that the message is logged as an inmanta warning and not as a tornado debug.

changelogs/unreleased/241-handle-max-clients-limit-in-agents.yml
```diff
@@ -0,0 +1,7 @@
+---
+description: Handle max-clients limit in agents better by adding a config parameter to set the limit, raising a warning if the pool gets exhausted and putting the heartbeat in a different pool.
```
Hugo-Inmanta, line 2: I would definitely mention that we added a new configuration parameter as you did, as for the 2 other parts I'm not sure how relevant it is for the end-user, @sanderr wdyt ?

Hugo-Inmanta (approved):
One tiny remark about the changelog, other than that looks good to me since I don't have satisfying answers to your last 2 outstanding questions

src/inmanta/protocol/endpoints.py
```diff
@@ -191,11 +209,13 @@ async def start(self) -> None:
...
```
sanderr, line 213: Shouldn't each session have a unique id?

src/inmanta/protocol/endpoints.py
```diff
@@ -191,11 +209,13 @@ async def start(self) -> None:
...
         await self.start_connected()
         self.add_background_task(self.perform_heartbeat())
 
     async def stop(self) -> None:
+        self._heartbeat_client.close()
```
sanderr, line 218: Why do we close this one and not the other one?

src/inmanta/protocol/endpoints.py
sanderr: Ah no, it's still one session, it just has two clients. Is that right?

src/inmanta/protocol/rest/client.py
sanderr, line 53: Wouldn't it be a lot easier to use `AsyncHTTPClient.configure` when the agent starts? I assume there's a reason for the singleton behavior tornado uses, and we actually want to apply it agent-wide anyway. It would also fix the bug(?) you have now where the `agent_rest_transport.max_clients` setting is used for any rest client, not only the agent's clients.

src/inmanta/protocol/endpoints.py
```diff
@@ -460,3 +481,7 @@ async def _call(
 
         result = await self._transport_instance.call(method_properties, args, kwargs)
         return result
+
+    def close(self) -> None:
+        if self._transport_instance and self._transport_instance.client:
+            self._transport_instance.client.close()
```
sanderr, line 487: The documentation is unclear on the behavior of `close` when the singleton objects are used.

src/inmanta/protocol/endpoints.py
FloLey: it should be closed as well. the docs say you need to close it yourself if you use force_instance. I first hadn't used it for the _client one and forgot to add the close() for that one afterwards.

src/inmanta/protocol/rest/client.py
FloLey: I don't find how to make AsyncHTTPClient.configure work but I fixed the fact that is was set for any rest client

sanderr: I'm going to look into it because I would if we can.

FloLey:
@sanderr except for the AsyncHTTPClient.configure, I addressed you comments. I still haven't manage to make it work using it.

sanderr (commented):
@FloLey PoC for what I think is the best of both worlds: force a new instance (and therefore connection pool and [separate max clients count](https://github.com/tornadoweb/tornado/blob/559f94aa8809fa37cb2015fadc06d8e46430fafe/tornado/simple_httpclient.py#L170)) only for the heartbeat client. Apart from that stick as close to tornado's defaults as possible: 1 client instance per IO loop, configured through the class method's `configure`.

```python
diff --git a/src/inmanta/protocol/endpoints.py b/src/inmanta/protocol/endpoints.py
index 82e9b07a..1b7f98d0 100644
--- a/src/inmanta/protocol/endpoints.py
+++ b/src/inmanta/protocol/endpoints.py
@@ -209,7 +209,7 @@ class SessionEndpoint(Endpoint, CallTarget):
         """
         assert self._env_id is not None
         LOGGER.log(3, "Starting agent for %s", str(self.sessionid))
-        self._client = SessionClient(self.name, self.sessionid, timeout=self.server_timeout, force_instance=True)
+        self._client = SessionClient(self.name, self.sessionid, timeout=self.server_timeout)
         self._heartbeat_client = SessionClient(self.name, self.sessionid, timeout=self.server_timeout, force_instance=True)
         await self.start_connected()
         self.add_background_task(self.perform_heartbeat())
diff --git a/src/inmanta/protocol/rest/client.py b/src/inmanta/protocol/rest/client.py
index 25d937d3..51514729 100644
--- a/src/inmanta/protocol/rest/client.py
+++ b/src/inmanta/protocol/rest/client.py
@@ -50,7 +50,7 @@ class RESTClient(RESTBase):
         self.connection_timout: int = connection_timout
         self.headers: Set[str] = set()
         self.request_timeout: int = inmanta_config.Config.get(self.id, "request_timeout", 120)
-        self.client = AsyncHTTPClient(force_instance=force_instance, max_clients=max_clients)
+        self.client = AsyncHTTPClient(force_instance=force_instance)

     @property
     def endpoint(self) -> "Endpoint":
diff --git a/tests/test_agent_manager.py b/tests/test_agent_manager.py
index e87089de..8c166176 100644
--- a/tests/test_agent_manager.py
+++ b/tests/test_agent_manager.py
@@ -1330,6 +1330,10 @@ async def test_heartbeat_different_session(server_pre_start, async_finalizer, ca

     Config.set("agent_rest_transport", "max_clients", "1")

+    from tornado.httpclient import AsyncHTTPClient
+    max_clients: int = Config.get("agent_rest_transport", "max_clients", "10")
+    AsyncHTTPClient.configure(None, max_clients=max_clients)
+
     server = TestSlice(name="test_slice")

     ibl = InmantaBootloader()
```

src/inmanta/app.py
```diff
@@ -114,9 +115,15 @@ def _handle_startup_done(fut: asyncio.Future) -> None:
 def start_agent(options: argparse.Namespace) -> None:
     from inmanta.agent import agent
 
-    util.ensure_event_loop()
+    # The call to configure() should be done as soon as possible.
+    # If an AsyncHTTPClient is started before this call, the max_client
+    # will not be taken into account.
+    max_clients: int = Config.get("agent_rest_transport", "max_clients", "10")
+    AsyncHTTPClient.configure(None, max_clients=max_clients)
```
sanderr, line 122: On second thought, perhaps we should only set this when the config value is explicitly set? i.e. use `None` as default and only call `configure` if not `None`? Wdyt?

src/inmanta/config.py
```diff
@@ -407,6 +407,13 @@ def __init__(self, name: str, port: int = 8888) -> None:
         self.request_timeout = Option(
             self.prefix, "request_timeout", 120, "The time before a request times out in seconds", is_int
         )
+        self.max_clients = Option(
+            self.prefix,
+            "max_clients",
+            "10",
```
sanderr, line 413: Same here

sanderr (changes requested):

src/inmanta/protocol/endpoints.py
```diff
@@ -357,13 +357,21 @@ def __init__(
...
             self._transport_instance = None
         self._version_match = version_match
         self._exact_version = exact_version
 
+    def close(self):
+        """
+        Closes the RESTclient instance manually. This is only needed when it is started with force_instance set to true
+        """
+        if self.forced_instance:
+            self._transport_instance.close()
```
sanderr, line 373:
```suggestion
        self._transport_instance.close()
```
No need to have the condition on this level

src/inmanta/protocol/endpoints.py
```diff
@@ -357,13 +357,21 @@ def __init__(
         super().__init__(name)
         assert isinstance(timeout, int), "Timeout needs to be an integer value."
         LOGGER.debug("Start transport for client %s", self.name)
+        self.forced_instance = force_instance
```
sanderr, line 360:
```suggestion
```
If we drop the condition we can drop the attribute.

sanderr (changes requested):

sanderr (approved):

PR #6225, opened 2023-07-06
Fix bugs in the inmanta modules list command

arnaudsjs:
# Description

This PR fixes the following bugs in the `inmanta modules list` command:

    * Fix bug where the `inmanta modules list` command incorrectly reports that the installed version matches the relevant version constraints. This issue was caused by an incorrect string comparison in the `merge_specs`. The names of two modules were compared with each other, but in the first part of the comparison the modules name contained underscores, while in the second part it contained dashes.
    * Verify that the version numbers, reported by the `inmanta modules list` command, are correctly formatted. The `Texttable()` class by default tries to infer the types of the column fields. When a version number doesn't have a patch version (example: 3.2), it gets interpreted as a float and `Texttable()` formats it as `3.200` in the table.

# Self Check:

- [x] Changelog entry


src/inmanta/module.py
```diff
@@ -301,7 +301,7 @@ def make_repo(path: str, root: Optional[str] = None) -> Union[LocalFileRepo, Rem
 def merge_specs(mainspec: "Dict[str, List[Requirement]]", new: "List[Requirement]") -> None:
     """Merge two maps str->[T] by concatting their lists."""
     for req in new:
-        key = req.project_name
+        key = req.project_name.replace("-", "_")
```
sanderr, line 304:
```suggestion
        key = req.key
```
I think this may be what you're looking for.

sanderr (changes requested):

FloLey (approved):

src/inmanta/module.py
arnaudsjs: Do you mean `req.key.replace("-", "_")`? Because `req.key` constained the name of the requirement in standard notation, but the standard notation consists of `-` characters and not `_` characters.

sanderr: Right, and we use this dict to do module lookup, you're right.

sanderr (approved):

sanderr (changes requested):
Given what you said, perhaps let's keep it as close to what it was before as possible and keep using `project_name` over `key`. I'm a bit worried about e.g. capital letters in module names.

sanderr (approved):

Issue #6232, opened 2023-07-07
Bump asyncpg 0.27 -> 0.28


Hugo-Inmanta:
Dep-merge is failing because of these version bumps:
- https://github.com/inmanta/inmanta-lsm/pull/1349
- https://github.com/inmanta/inmanta-lsm/pull/1350

As proposed in this slack discussion [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1688723796943219)
- [x] Bump the version constraint on inmanta-core to unstuck the dependabot PR. -> PR [here](https://github.com/inmanta/inmanta-core/pull/6233)
- [x] Create a ticket to investigate how we can get rid of the _copy_in() method call in PGRestore, so that we can get rid of the version constraint in inmanta-core -> Ticket [here](https://github.com/inmanta/inmanta-core/issues/6234)

PR #6233, opened 2023-07-07
[FIX] bump asyncpg's constraint higher bound to 0.28

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/6232

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_namespace_packages
 from os import path
 
 requires = [
-    "asyncpg~=0.25,<0.28",
+    "asyncpg~=0.25,<0.29",
     "click-plugins~=1.0",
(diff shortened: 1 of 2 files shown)


sanderr (approved):

setup.py
```diff
@@ -2,7 +2,7 @@
 from os import path
 
 requires = [
-    "asyncpg~=0.25,<0.28",
+    "asyncpg~=0.25,<0.29",
```
sanderr, line 5:
```suggestion
    "asyncpg>=0.25,<0.29",
```

sanderr (changes requested):

sanderr (approved):

PR #6243, opened 2023-07-10
issue/6224-drop-instance-dump

### LINKED ISSUE #6224: drop Instance.dump method
The `inmanta.execute.runtime.Instance.dump` method doesn't seem to be used and is currently broken because it accesses `ResultVariable.promises` which only exists on some of its child classes.

- [ ] remove the method, run the tests, see if anything breaks.

FloLey:
# Description

The inmanta.execute.runtime.Instance.dump method doesn't seem to be used and is currently broken because it accesses ResultVariable.promises which only exists on some of its child classes. This PR removes the method

closes #6224 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/6224-drop-instance-dump-method.yml
+++ b/changelogs/unreleased/6224-drop-instance-dump-method.yml
@@ -1,0 +1,4 @@
+description: Remove the inmanta.execute.runtime.Instance.dump method as it doesn't seem to be used and is currently broken.
+issue-nr: 6224
+change-type: patch
+destination-branches: [master, iso6, iso5]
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/6224-drop-instance-dump-method.yml
```diff
@@ -0,0 +1,4 @@
+description: Remove the inmanta.execute.runtime.Instance.dump method as it doesn't seem to be used and is currently broken.
+issue-nr: 6224
+change-type: patch
+destination-branches: [master, iso6, iso5]
```
sanderr, line 4:
```suggestion
destination-branches: [master]
```
Let's restrict this to master, just in case.

sanderr (changes requested):

sanderr (approved):

Issue #6025, opened 2023-05-23
Add DiscoveryHandler


sanderr:
Implement `inmanta.agent.handler.DiscoveryHandler` as described in the design.

- [ ] implement handler ABC
- [ ] add documentation including examples, see the design for more details.
- [ ] Verify that the `Id` constructor and `resource_str` method are part of the stable API and documented.

Hugo-Inmanta:
- [ ] implement top-level handler ABC (parent of Handler, CRUDHandler and new DiscoveryHandler), request early and wide review (separate pull request to enable/separate discussion of the generic handler interface from the specicif discovery handler)

PR [here](https://github.com/inmanta/inmanta-core/compare/issue/6025-implement-top-level-handler-ABC?expand=1)

PR #6246, opened 2023-07-10
Issue/6025 implement top level handler abc

Hugo-Inmanta:
# Description

# Self Check:

- [ ] Changelog entry


src/inmanta/agent/handler.py
```diff
@@ -55,7 +55,7 @@ class provider(object):  # noqa: N801
     """
     A decorator that registers a new handler.
 
-    :param resource_type: The type of the resource this handler provides an implementation for.
+    :param resource_type: The type of the resource this handler provides an implementation for. # TODO rephrase 'provides an implementation for'
```
sanderr, line 58: Don't forget about this TODO

sanderr (changes requested):

src/inmanta/agent/handler.py
```diff
@@ -793,7 +823,7 @@ def call() -> typing.Awaitable[Result]:
 
 
 @stable_api
-class CRUDHandler(ResourceHandler):
+class CRUDHandler(ResourceHandler, HandlerABC):
```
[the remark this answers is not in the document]
Hugo-Inmanta: I guess you mean `CRUDHandler` ? This would mean we need to lift some functionality from ResourceHandler to HandlerABC ?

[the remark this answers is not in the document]
sanderr: No, I mean `ResourceHandler`. If it inherits from `HandlerABC`, `CRUDHandler` doesn't need to. Now we have
```python
class HandlerABC: pass
class ResourceHandler: pass
class CRUDHandler(ResourceHandler, HandlerABC): pass
```

and I'm proposing
```python
class HandlerABC: pass
class ResourceHandler(HandlerABC): pass
class CRUDHandler(ResourceHandler): pass
```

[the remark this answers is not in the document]
sanderr: My first snippet is not right. Not sure if you changed it or I saw it wrong. Anyway, if `ResourceHandler(HandlerABC)`, why would we explicitly inherit from it in `CRUDHandler`?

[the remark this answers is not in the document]
Hugo-Inmanta: Right, 100% agree

bartv (commented):
Not sure I can contribute here. I a missing missing the point on what the goal of this refactor is.

src/inmanta/agent/handler.py
```diff
@@ -419,7 +419,52 @@ def _log_msg(self, level: int, msg: str, *args: object, exc_info: bool = False,
...
+        This method is always called by the agent
+        """
+        pass
+
+    @abstractmethod
+    def execute(self, ctx: HandlerContext, resource: resources.Resource, dry_run: Optional[bool] = None) -> None:
+        """
+        Update the given resource. This method is called by the agent.
+        """
+        pass
```
arnaudsjs, line 463: The docstrings on these methods have to be improved. What is the behavioral difference between `deploy()` and `execute()` for example? It's not possible to derive that information for the name itself.

Hugo-Inmanta: I agree, I was planning on hoisting these methods from `ResourceHandler` into `HandlerABC`, but now that I've worked on the actual implementation of the `DiscoveryHandler` I don't think I'll move these up after all.

sanderr: Perhaps `execute` shouldn't exist on the `ABC` then?

src/inmanta/agent/handler.py
```diff
@@ -419,7 +419,52 @@ def _log_msg(self, level: int, msg: str, *args: object, exc_info: bool = False,
...
+    def pre(self, ctx: HandlerContext, resource: resources.Resource) -> None:
+        """
+        Method executed before a handler operation (Facts, dryrun, real deployment, ...) is executed. Override this method
+        to run before an operation.
+
+        :param ctx: Context object to report changes and logs to the agent and server.
+        :param resource: The resource to query facts for.
+        """
+
+    @abstractmethod
```
sanderr, line 437: I don't think this one or `pre` should be marked as `abstractmethod` because that requires concrete classes to override them. We want their default behavior to  be a NOOP, don't we?

Hugo-Inmanta: I agree, that's the route I took on my PR implementing the `DiscoveryHandler`, I'll change it here as well

src/inmanta/agent/handler.py
Hugo-Inmanta: Agreed, and perhaps neither should `deploy`, I'll update this accordingly after feedback on the PR implementing the `DiscoveryHandler`

src/inmanta/agent/handler.py
```diff
@@ -470,19 +490,45 @@ def run() -> None:
...
         """
         if self._client is None:
             self._client = protocol.SessionClient("agent", self._agent.sessionid)
         return self._client
 
+
+@stable_api
+class ResourceHandler(HandlerABC):
+    """
+    A baseclass for classes that handle resources. New handler are registered with the
```
arnaudsjs, line 506:
```suggestion
    A baseclass for classes that handle resources. New handlers are registered with the
```

src/inmanta/agent/handler.py
```diff
@@ -470,19 +490,45 @@ def run() -> None:
...
+    The implementation of a handler should use the ``self._io`` instance to execute io operations. This io objects
+    makes abstraction of local or remote operations. See :class:`~inmanta.agent.io.local.LocalIO` for the available
+    operations.
+
+    :param agent: The agent that is executing this handler.
+    :param io: The io object to use.
+    """
+
+    def __init__(self, agent: "inmanta.agent.agent.AgentInstance", io: Optional["IOBase"] = None) -> None:
+        self._agent = agent
```
arnaudsjs, line 518: Shouldn't we call super here instead of setting `self._agent`, `self._ioloop` and `self._client` directly?

src/inmanta/agent/handler.py
```diff
@@ -419,31 +420,50 @@ def _log_msg(self, level: int, msg: str, *args: object, exc_info: bool = False,
...
+
+    @abstractmethod
+    def deploy(
+        self,
+        ctx: HandlerContext,
+        resource: R,
+        requires: Dict[ResourceIdStr, ResourceState],
+    ) -> None:
+        """
+        Main entrypoint of the handler that will be called by the agent.
```
arnaudsjs, line 463: This docstring doesn't really describe the behavior of the method.

arnaudsjs (changes requested):

src/inmanta/agent/handler.py
sanderr, line 506: The part about handler registration seems relevant for the top level ABC as well.

src/inmanta/agent/handler.py
```diff
@@ -419,31 +420,50 @@ def _log_msg(self, level: int, msg: str, *args: object, exc_info: bool = False,
...
+        :param ctx: Context object to report changes and logs to the agent and server.
+        :param resource: The resource to query facts for.
+        """
+
+    @abstractmethod
+    def deploy(
+        self,
+        ctx: HandlerContext,
+        resource: R,
+        requires: Dict[ResourceIdStr, ResourceState],
```
sanderr, line 460: I think this would be a good time to clean up this type. Either `dict` or `abc.Mapping`. Afaik we don't ever update it so `Mapping` would be best. @arnaudsjs do you know?

src/inmanta/agent/handler.py
```diff
@@ -470,19 +490,45 @@ def run() -> None:
...
-
         :return: A client that is associated with the session of the agent that executes this handler.
         """
         if self._client is None:
             self._client = protocol.SessionClient("agent", self._agent.sessionid)
         return self._client
 
+
+@stable_api
+class ResourceHandler(HandlerABC):
```
sanderr, line 504:
```suggestion
class ResourceHandler(HandlerABC[R]):
```
I think. Then update all type annotations of `resources.Resource` to `R` instead. Same for the `CRUDHandler`.

sanderr (changes requested):

src/inmanta/agent/handler.py
arnaudsjs: I don't think we ever update it.

src/inmanta/agent/handler.py
```diff
@@ -418,12 +418,12 @@ def _log_msg(self, level: int, msg: str, *args: object, exc_info: bool = False,
         self._logs.append(log)
 
 
-@stable_api
-class ResourceHandler(object):
+class HandlerAPI(ABC):
```
arnaudsjs, line 421:
```suggestion
@stable_api
class HandlerAPI(ABC):
```

src/inmanta/agent/handler.py
```diff
@@ -434,16 +434,74 @@ class ResourceHandler(object):
...
+        failed to deploy.
+
+        :param ctx: Context object to report changes and logs to the agent and server.
+        :param resource: The resource to deploy
+        :param requires: A dictionary mapping the resource id of each dependency of the given resource to its resource state.
+        """
+        pass
+
+    @abstractmethod
+    def execute(self, ctx: HandlerContext, resource: resources.Resource, dry_run: bool = False) -> None:
```
arnaudsjs, line 469: What is the difference between the `deploy()` and the `execute()` method in the context of a `DiscoveryHandler`?

arnaudsjs (changes requested):

src/inmanta/agent/handler.py
wouterdb: dryrun still uses execute instead of deploy

wouterdb (approved):

src/inmanta/agent/handler.py
arnaudsjs: Ok. What would be the main entrypoint for the `DiscoveryHandler`? Is it `deploy()` or `execute()`. Because I would find `deploy()` a strange name. It doesn't really deploy anything in the end.

arnaudsjs (approved):

src/inmanta/agent/handler.py
Hugo-Inmanta: I agree, I'm trying to come up with some clearer phrasing

sanderr: I added a suggested change to the `deploy` docstring to try and clear this up. What do you think? Anything missing or still unclear there? After we've reached consensus there I think we need to clear up this docstring in the same vein.

sanderr (commented):
Releasing comments to enable discussion about "deploy" semantics. Review still in progress.

src/inmanta/agent/handler.py
```diff
@@ -419,11 +419,12 @@ def _log_msg(self, level: int, msg: str, *args: object, exc_info: bool = False,
 
 
 @stable_api
-class ResourceHandler(object):
+class HandlerAPI(ABC):
     """
-    A baseclass for classes that handle resources. New handler are registered with the
-    :func:`~inmanta.agent.handler.provider` decorator.
+    Base class describing the interface between the agent and the handler. This class first defines the interface
+    At the end, it also defined a number of utility methods.
```
sanderr, line 425:
```suggestion
    Base class describing the interface between the agent and the handler. This class first defines the interface.
    At the end, it also defines a number of utility methods.
```

src/inmanta/agent/handler.py
```diff
@@ -434,16 +435,74 @@ class ResourceHandler(object):
...
+        :func:`~inmanta.agent.handler.ResourceHandler.do_changes`
+
+        :param ctx: Context object to report changes and logs to the agent and server.
+        :param resource: The resource to check the current state of.
+        :param dry_run: True will only determine the required changes but will not execute them.
+        """
+        pass
+
+    @abstractmethod
+    def check_facts(self, ctx: HandlerContext, resource: resources.Resource) -> Dict[str, object]:
```
sanderr, line 484: Same question: why is this abstract while still declaring how it interacts with utility methods?

src/inmanta/agent/handler.py
```diff
@@ -434,16 +435,74 @@ class ResourceHandler(object):
...
+        :return: A dict with fact names as keys and facts values.
+        """
+        pass
+
+    def available(self, resource: resources.Resource) -> bool:
+        """
+        Returns true if this handler is available for the given resource
+
+        :param resource: Is this handler available for the given resource?
+        :return: Available or not?
```
sanderr, line 501: I think this is a good time to clarify this docstring. I only know what it means because I've seen it in use.

src/inmanta/agent/handler.py
```diff
@@ -434,16 +435,74 @@ class ResourceHandler(object):
...
+        :param resource: The resource to deploy
+        :param requires: A dictionary mapping the resource id of each dependency of the given resource to its resource state.
+        """
+        pass
+
+    @abstractmethod
+    def execute(self, ctx: HandlerContext, resource: resources.Resource, dry_run: bool = False) -> None:
+        """
+        Update the given resource. This method is called by the agent. Most handlers will not override this method
+        and will only override :func:`~inmanta.agent.handler.ResourceHandler.check_resource`, optionally
```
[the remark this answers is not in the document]
sanderr: Also it references `ResourceHandler` methods rather than `HandlerAPI` ones. If you update this, please double check that they are all defined on the ABC.

src/inmanta/agent/handler.py
```diff
@@ -483,6 +542,68 @@ def get_client(self) -> protocol.SessionClient:
...
+            actual_hash_of_file = hash_file(file_contents)
+            if hash_id != actual_hash_of_file:
+                raise Exception(f"File hash verification failed, expected: {hash_id} but got {actual_hash_of_file}")
+            return file_contents
+        else:
+            raise Exception("An error occurred while retrieving file %s" % hash_id)
+
+    def stat_file(self, hash_id: str) -> bool:
+        """
+        Check if a file exists on the server. This method does and async call to the server and blocks on the result.
```
sanderr, line 571: I don't really see the relevance of the second sentence. @arnaudsjs @wouterdb do you?

src/inmanta/agent/handler.py
```diff
@@ -483,6 +542,68 @@ def get_client(self) -> protocol.SessionClient:
             self._client = protocol.SessionClient("agent", self._agent.sessionid)
         return self._client
 
+    def get_file(self, hash_id: str) -> Optional[bytes]:
+        """
+        Retrieve a file from the fileserver identified with the given id. The convention is to use the sha1sum of the
```
sanderr, line 547: I'm not sure if the second part is really relevant for the `get_file` method. It seems like it would be mostly relevant for the upload.

src/inmanta/agent/handler.py
```diff
@@ -434,16 +435,74 @@ class ResourceHandler(object):
...
+        failed to deploy.
+
+        :param ctx: Context object to report changes and logs to the agent and server.
+        :param resource: The resource to deploy
+        :param requires: A dictionary mapping the resource id of each dependency of the given resource to its resource state.
+        """
+        pass
+
+    @abstractmethod
+    def execute(self, ctx: HandlerContext, resource: resources.Resource, dry_run: bool = False) -> None:
```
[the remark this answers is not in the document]
Hugo-Inmanta: What do you mean exactly with lsm ? (I only found this handler https://github.com/inmanta/lsm/blob/5ba3b8fa9e60b358e311d77ce2d4ab4aa37efe4e/plugins/__init__.py#L1567 which does have an execute method through inheritance)

src/inmanta/agent/handler.py
```diff
@@ -434,16 +435,74 @@ class ResourceHandler(object):
...
+    def available(self, resource: resources.Resource) -> bool:
+        """
+        Returns true if this handler is available for the given resource
+
+        :param resource: Is this handler available for the given resource?
+        :return: Available or not?
+        """
+        return True
+
+    # Utility methods
```
sanderr, line 505: `set_cache` needs a docstring.

src/inmanta/agent/handler.py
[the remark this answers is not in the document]
sanderr: That's the one I meant, and luckily you're right, I was missing something here.

sanderr (changes requested):
I'm adding a lot of comments to try and clear up docstrings. I'm not intimately familiar with all of them so others may not agree with my suggestions: if you have doubts about any of them I would suggest asking the team before sinking too much time in it. And either way rerequest everyone's review afterwards because these are quite impactful changes.

src/inmanta/agent/handler.py
arnaudsjs: Me neither.

src/inmanta/agent/handler.py
```diff
@@ -434,16 +435,74 @@ class ResourceHandler(object):
...
+        :param dry_run: True will only determine the required changes but will not execute them.
+        """
+        pass
+
+    @abstractmethod
+    def check_facts(self, ctx: HandlerContext, resource: resources.Resource) -> Dict[str, object]:
+        """
+        This method is called by the agent to query for facts. It runs :func:`~inmanta.agent.handler.ResourceHandler.pre`
+        and :func:`~inmanta.agent.handler.ResourceHandler.post`. This method calls
+        :func:`~inmanta.agent.handler.ResourceHandler.facts` to do the actually querying.
```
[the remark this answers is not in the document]
Hugo-Inmanta: removed those

src/inmanta/agent/handler.py
```diff
@@ -434,16 +435,146 @@ class ResourceHandler(object):
...
+        """
+        return False
+
+    def do_reload(self, ctx: HandlerContext, resource: resources.Resource) -> None:
+        """
+        Perform a reload of this resource.
+
+        :param ctx: Context object to report changes and logs to the agent and server.
+        :param resource: The resource to reload.
+        """
```
Hugo-Inmanta, line 575: Had to lift these two as well since they're called from the `deploy` method. Suggestions for better docstrings are welcome

Hugo-Inmanta:
One mypy error popped up, I think it's tied to genericity, I'm not sure if we can ignore it or not:
```diff
+ src/inmanta/agent/handler.py:1008: error: Argument 2 of "execute" is incompatible with supertype "HandlerAPI"; supertype defines the argument type as "Resource"  [override]
```

src/inmanta/agent/handler.py
wouterdb: Why lift the implementation of deploy? 

My understanding is that this class is purely the definition of the interface between agent and handler, so the deploy logic is not part of that?

sanderr:
> One mypy error popped up, I think it's tied to genericity, I'm not sure if we can ignore it or not:

It has to do with the fact that we try to keep some classes non-generic. Ignore it for now but make sure to run mypy again once we decided where to merge this.

src/inmanta/agent/handler.py
```diff
@@ -446,46 +446,102 @@ def __init__(self, agent: "inmanta.agent.agent.AgentInstance", io: Optional["IOB
     # Interface
 
     def close(self) -> None:
+        """
+        Override this method to implement custom logic called by the agent on handler deactivation. i.e. when the
+        instantiated handler will no longer be used by the handler.
```
sanderr, line 451:
```suggestion
        instantiated handler will no longer be used by the agent.
```
Right?

src/inmanta/agent/handler.py
sanderr: Most of what it does is relevant for the generic concept of a resource, isn't it? We could choose to extract some parts of it to the `Resource` subclass, but any resource has dependencies and its handler should generally be executed when those dependencies have successfully deployed.

Hugo-Inmanta: As discussed: keep it as is, but unmark HandlerABC stable_api annotation

src/inmanta/agent/handler.py
sanderr: Looks good to me, but @wouterdb is more knowledgeable about it.

src/inmanta/agent/handler.py
```diff
@@ -495,13 +551,29 @@ def check_facts(self, ctx: HandlerContext, resource: resources.Resource) -> Dict
 
     def available(self, resource: resources.Resource) -> bool:
         """
-        Returns true if this handler is available for the given resource
+        Override this method to implement custom handler selection logic, e.g. if multiple handlers were defined
+        for the same resource type.
```
sanderr, line 554: I just realized that iirc we want to drop this sometime soon (@wouterdb can you confirm?). Perhaps we should declare that it should generally not be overridden?
```suggestion
        Kept for backwards compatibility, new handler implementations should never override this.
```

sanderr (changes requested):

src/inmanta/agent/handler.py
```diff
@@ -446,46 +446,102 @@ def __init__(self, agent: "inmanta.agent.agent.AgentInstance", io: Optional["IOB
...
         """
         pass
 
     @abstractmethod
     def check_facts(self, ctx: HandlerContext, resource: resources.Resource) -> Dict[str, object]:
         """
-        This method is called by the agent to query for facts. It runs :func:`~inmanta.agent.handler.ResourceHandler.pre`
-        and :func:`~inmanta.agent.handler.ResourceHandler.post`. This method calls
-        :func:`~inmanta.agent.handler.ResourceHandler.facts` to do the actually querying.
+        This method is called by the agent to query for facts.
```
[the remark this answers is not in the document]
Hugo-Inmanta: > pre and post are concepts of the HandlerAPI

I think they got dropped from the ABC when I merged wouter's suggestion, I put them back in, sorry for the mess.

> Same for the facts() reference and the default implementation.

I lifted those up in the ABC as well

src/inmanta/agent/handler.py
```diff
@@ -574,6 +563,61 @@ def do_reload(self, ctx: HandlerContext, resource: resources.Resource) -> None:
...
+
+    def post(self, ctx: HandlerContext, resource: resources.Resource) -> None:
+        """
+        Method executed after an operation. Override this method to run after an operation.
+
+        :param ctx: Context object to report changes and logs to the agent and server.
+        :param resource: The resource to query facts for.
+        """
+
+    def facts(self, ctx: HandlerContext, resource: resources.Resource) -> Dict[str, object]:
```
sanderr, line 583: I think this should be in the utility methods section. Can you check for the other methods you have lifted in the past few days? I think `pre` and `post` could also be considered utility methods if we define the interface methods as the "agent-handler API interface".

wouterdb (approved):

sanderr (commented):
Looks good to me except for the `Generic` discussion which is blocked until Bart is back.

src/inmanta/agent/handler.py
```diff
@@ -647,106 +527,153 @@ def filter_resources_in_unexpected_state(
...
-            ctx.exception(
-                "An error occurred during deployment of %(resource_id)s (exception: %(exception)s",
-                resource_id=resource.id,
-                exception=f"{e.__class__.__name__}('{e}')",
-            )
+    def check_facts(self, ctx: HandlerContext, resource: resources.Resource) -> Dict[str, object]:
+        """
+        This method is called by the agent to query for facts. It runs :func:`~inmanta.agent.handler.HandlerAPI.pre`
+        and :func:`~inmanta.agent.handler.HandlerAPI.post`. This method calls
+        :func:`~inmanta.agent.handler.HandlerAPI.facts` to do the actually querying.
```
arnaudsjs, line 554:
```suggestion
        :func:`~inmanta.agent.handler.HandlerAPI.facts` to do the actual querying.
```

arnaudsjs (changes requested):

arnaudsjs (approved):

Hugo-Inmanta:
Still a couple of outstanding mypy issues:
```
+ src/inmanta/agent/handler.py:651: error: Argument 2 to "chain_future" has incompatible type "concurrent.futures._base.Future[T]"; expected "asyncio.futures.Future[Any]"  [arg-type]
+ src/inmanta/agent/handler.py:920: error: Argument 2 to "read_resource" of "CRUDHandler" has incompatible type "PurgeableResource"; expected "TPurgeableResource"  [arg-type]
+ src/inmanta/agent/handler.py:921: error: Argument 2 to "calculate_diff" of "CRUDHandler" has incompatible type "PurgeableResource"; expected "TPurgeableResource"  [arg-type]

```
Ultimately tied to the typing of Resource https://github.com/inmanta/inmanta-core/blob/cffef5db5d3e21fc54c10a30c1cee56d775eb2db/src/inmanta/resources.py#L425 let me know if you see a solution for those

sanderr (changes requested):
Looks mostly good. A few minor notes/suggestions and a few notes for the follow-up ticket.

> let me know if you see a solution for those

- For the first, I don't think this is actually new, but it does look like an actual type error to me. I think using `asyncio.Future` instead of `concurrent.futures.Future` should resolve that. That is unless @wouterdb knows of any reason we would want to violate this?

- For the second and third: drop the assertions, and update `Resource.clone` and `Resource.deserialize` to return the type of the class rather than `Resource`.

arnaudsjs:
Regarding the typing issue on `chain_future`: This seems to be a typing issue in Tornado. The `chain_future` is annotated to accept a `asyncio.Future` instance, but according to the implementation and the documentation of the method it accepts all types of Futures (asyncio.Future and concurrent.futures.Future). I created [a ticket](https://github.com/tornadoweb/tornado/issues/3314) for this.

arnaudsjs:
@sanderr I still have the following type error left:

```
src/inmanta/resources.py:435: error: Incompatible return value type (got "Resource", expected "T")  [return-value]
```

I don't really see how that can be fixed. `Resource.deserialize()` is a class method. Its return type depends on what `resource.get_class()` returns and that can be any subclass of Resource.

PR #6255, opened 2023-07-11
refactored precedence rules

sanderr:
# Description

Made an attempt to simplify precedence reasoning and to make it more consistent.
- Got rid of most custom precedence levels
- Replaced them with a single `LOW` precedence level that tries to shift tokens eagerly
- Added documentation for the `empty` token to include its two primary use cases
- Use the `empty` rule everywhere custom precedence was currently used. This is the part I'm least certain about. I feel like it helps readability because
  1. the comment for the `empty` token is now clear on its uses
  2. when used for precedence enforcement it always occurs in a place where the preceding token appears in a similar context in another rule, followed by something non-empty
- Added a missing precedence rule for the `MATCHING` terminal token

A lot of this is very subjective so I'd appreciate your input.

@Hugo-Inmanta I added you as a reviewer as well in case you have an opinion. If you feel like it would be too much effort to refresh these concepts for the review, feel free to unassign yourself.

### DIFF
--- a/src/inmanta/parser/plyInmantaParser.py
+++ b/src/inmanta/parser/plyInmantaParser.py
@@ -68,7 +68,8 @@ file = "NOFILE"
 namespace: Optional[Namespace] = None
 
 precedence = (
-    # associativity rule for productions that should eagerly shift new tokens as long as they remain feasible
+    # prededence rule for productions that should eagerly shift new tokens as long as they remain feasible
+    # e.g. prefer one statement `x = 1 not in []` over two statements `x = 1` and `not in []`
     ("nonassoc", "LOW"),
     ("nonassoc", "MATCHING"),
     ("right", ","),
--- a/src/inmanta/parser/plyInmantaParser.py
+++ b/src/inmanta/parser/plyInmantaParser.py
@@ -135,7 +135,7 @@ def p_main(p: YaccProduction) -> None:
 
 
 # low-precedence production that consumes no tokens, has two use cases:
-# - use its low preference compared to other tokens to prefer one production over another (especially relevant for recursion)
+# - use its low precedence compared to other tokens to prefer one production over another (especially relevant for recursion)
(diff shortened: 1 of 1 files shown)


FloLey (approved):
I like it, It's easier than before :)

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -68,16 +68,18 @@
 namespace: Optional[Namespace] = None
 
 precedence = (
+    # prededence rule for productions that should eagerly shift new tokens as long as they remain feasible
```
Hugo-Inmanta, line 71:
```suggestion
    # precedence rule for productions that should eagerly shift new tokens as long as they remain feasible
```

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -132,8 +134,16 @@ def p_main(p: YaccProduction) -> None:
     p[0] = v
 
 
+# low-precedence production that consumes no tokens, has two use cases:
+# - use its low preference compared to other tokens to prefer one production over another (especially relevant for recursion)
```
Hugo-Inmanta, line 138:
```suggestion
# - use its low precedence compared to other tokens to prefer one production over another (especially relevant for recursion)
```

Hugo-Inmanta (changes requested):
I'm not sure I'm fully getting the implications of these changes. This feels like a big change to me, but this might not be the case.
I would expect that moving the precedence rules around this much would have an impact on performance and/or correctness ?
I agree with the move from `left` to `nonassoc` though

sanderr:
@Hugo-Inmanta I think the impact should be limited, as the precedence rules I replaced only occur in places where they are never compared with the tokens I skipped over afaik. But there is indeed a minor risk, though we should catch that in the solutions team nightlies if anything goes wrong.

Hugo-Inmanta (approved):

PR #6264, opened 2023-07-12
Issue/6025 implement discovery handler

Hugo-Inmanta:
# Description

- [ ] implement `DiscoveryHandler`
- [x] Verify that the `Id` constructor and `resource_str` method are part of the stable API and documented.

# Self Check:

- [ ] Changelog entry


src/inmanta/agent/handler.py
```diff
@@ -961,6 +977,109 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+    # and the Unmanaged Resource type (URT) it reports to the server. The second has to be serializable.
+
+    # This class deploys instances of DRT and reports URT to the server.
+
+    def __init__(self, agent: "inmanta.agent.agent.AgentInstance") -> None:
+        self._agent = agent
+        self._ioloop = agent._io_loop
+
+    @abstractmethod
+    def discover_resources(self, ctx: HandlerContext, discovery_resource: DRT) -> abc.Mapping[ResourceIdStr, URT]:
```
sanderr, line 992: Good catch that this should be a `ResourceIdStr`, looks like that was missing from the design.

src/inmanta/agent/handler.py
```diff
@@ -961,6 +977,109 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+        raise NotImplementedError
+
+    def report_discovered_resources(self, ctx: HandlerContext, resource: DRT) -> None:
+        """ """
+
+        def _call_discovered_resource_create_batch() -> typing.Awaitable[Result]:
+            return self.get_client().discovered_resource_create_batch(
+                tid=self._agent.environment, discovered_resources=discovered_resources
```
sanderr, line 1000: Just a style preference but I would always pass `discovered_resources` as a parameter rather than counting on an implicit closure of something that's defined below it.

src/inmanta/agent/handler.py
```diff
@@ -961,6 +977,109 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+        finally:
+            try:
+                self.post(ctx, resource)
+            except Exception as e:
+                ctx.exception(
+                    (
+                        "An error occurred after resource discovery of type %(urt)s, "
+                        "triggered by %(resource_id)s (exception: %(exception)s"
+                    ),
+                    urt=str(URT),
```
sanderr, line 1036: `URT` is just a `TypeVar`, not the bound type. It is also just an intermediate representation type so I'm not sure if it is even relevant to the end user. Same for the other message

src/inmanta/agent/handler.py
```diff
@@ -961,6 +977,109 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+
+    def __init__(self, agent: "inmanta.agent.agent.AgentInstance") -> None:
+        self._agent = agent
+        self._ioloop = agent._io_loop
+
+    @abstractmethod
+    def discover_resources(self, ctx: HandlerContext, discovery_resource: DRT) -> abc.Mapping[ResourceIdStr, URT]:
+        raise NotImplementedError
+
+    def report_discovered_resources(self, ctx: HandlerContext, resource: DRT) -> None:
```
sanderr, line 995: This should be `deploy`. Could optionally be split into discover -> serialize -> report, so deploy just has to chain them together. No strong preference on the last part, unless the serialization turns out to be very verbose.

src/inmanta/agent/handler.py
```diff
@@ -961,6 +977,109 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+
+        def _call_discovered_resource_create_batch() -> typing.Awaitable[Result]:
+            return self.get_client().discovered_resource_create_batch(
+                tid=self._agent.environment, discovered_resources=discovered_resources
+            )
+
+        try:
+            self.pre(ctx, resource)
+            # report to the server
+            discovered_resources: List[DiscoveredResource] = [
```
sanderr, line 1006: Should be typed as lowercase `list` or `abc.Sequence`. You're not modifying it so `abc.Sequence` presents the clearer picture.

src/inmanta/agent/handler.py
```diff
@@ -961,6 +977,109 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+                    resource_id=resource.id,
+                    exception=f"{e.__class__.__name__}('{e}')",
+                )
+
+    async def async_report_discovered_resources(self, ctx: HandlerContext, resource: DRT) -> None:
+        try:
+            self.pre(ctx, resource)
+            # report to the server
+            discovered_resources: List[DiscoveredResource] = [
+                DiscoveredResource(discovered_resource_id=resource_id, values=values)
```
sanderr, line 1046: These types do not match, we should serialize here.

src/inmanta/agent/handler.py
```diff
@@ -961,6 +977,109 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+                self.post(ctx, resource)
+            except Exception as e:
+                ctx.exception(
+                    (
+                        "An error occurred after resource discovery of type %(urt)s, "
+                        "triggered by %(resource_id)s (exception: %(exception)s"
+                    ),
+                    urt=str(URT),
+                    resource_id=resource.id,
+                    exception=f"{e.__class__.__name__}('{e}')",
```
sanderr, line 1038: I think `repr()` does pretty much this:
```python
>>> e = Exception("hellow orld")
>>> str(e)
'hellow orld'
>>> repr(e)
"Exception('hellow orld')"
```

Same for other message

src/inmanta/agent/handler.py
```diff
@@ -961,6 +977,109 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+                resource_id=resource.id,
+                exception=f"{e.__class__.__name__}('{e}')",
+            )
+        finally:
+            try:
+                self.post(ctx, resource)
+            except Exception as e:
+                ctx.exception(
+                    (
+                        "An error occurred after resource discovery of type %(urt)s, "
```
sanderr, line 1033:
```suggestion
                        "An error occurred during post stage for resource discovery of type %(urt)s, "
```

sanderr (changes requested):
Intermediate review: I haven't looked at test cases or `inmanta.resources` yet.

src/inmanta/agent/handler.py
```diff
@@ -47,6 +48,10 @@
 LOGGER = logging.getLogger(__name__)
 
 T = TypeVar("T")
+# Discovery Resource Type
+DRT = TypeVar("DRT", bound=resources.Resource)
+# Unmanaged Resource Type
+URT = TypeVar("URT", bound=pydantic.BaseModel)  # bound=... enough to enforce the fact it needs to be serializable ?
```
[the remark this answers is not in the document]
sanderr: I think we want to drop this bound and just be very clear in the docstring that these objects are expected to be serializable.

src/inmanta/agent/handler.py
```diff
@@ -419,34 +424,45 @@ def _log_msg(self, level: int, msg: str, *args: object, exc_info: bool = False,
...
     """
-    A baseclass for classes that handle resources. New handler are registered with the
-    :func:`~inmanta.agent.handler.provider` decorator.
 
-    The implementation of a handler should use the ``self._io`` instance to execute io operations. This io objects
-    makes abstraction of local or remote operations. See :class:`~inmanta.agent.io.local.LocalIO` for the available
-    operations.
+    # _client: Optional[protocol.SessionClient] = None
+    # _agent: Optional[inmanta.agent.agent.AgentInstance] = None
+    # _io_loop = None
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think this is an artifact of my trials and errors, I commented it out but forgot to remove it

src/inmanta/agent/handler.py
```diff
@@ -961,6 +977,109 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+@stable_api
+class DiscoveryHandler(HandlerABC, Generic[DRT, URT]):
+    # The DiscoveryHandler is generic in both the handler's Discovery Resource type (DRT)
+    # and the Unmanaged Resource type (URT) it reports to the server. The second has to be serializable.
+
+    # This class deploys instances of DRT and reports URT to the server.
+
+    def __init__(self, agent: "inmanta.agent.agent.AgentInstance") -> None:
+        self._agent = agent
+        self._ioloop = agent._io_loop
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think this is the _io_loop that should be used ? https://github.com/inmanta/inmanta-core/blob/9cf390dd8a8fa91f050b3c7ad63b991905e4de34/src/inmanta/agent/agent.py#L1055C1-L1055C40
But it makes me wonder about this one then https://github.com/inmanta/inmanta-core/blob/9cf390dd8a8fa91f050b3c7ad63b991905e4de34/src/inmanta/agent/handler.py#L445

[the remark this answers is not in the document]
sanderr: You're right that that is the loop that should be used, but note that `agent` here is of type `AgentInstance`, so `agent.process` is of the type you linked.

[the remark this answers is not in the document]
Hugo-Inmanta: I see now, I don't get how I got confused

src/inmanta/agent/handler.py
Hugo-Inmanta: Ah turns out it is not the same regarding formatting:

```python
e = Exception("An\nError\tMessage")

print(str(e))
print(repr(e))
print(f"{e.__class__.__name__}('{e}')")
```
```bash
An
Error	Message
Exception('An\nError\tMessage')
Exception('An
Error	Message')
```

And [this test](https://github.com/inmanta/inmanta-core/blob/b423fceb54b73138c0f9203b320bc376230280ff/tests/agent_server/test_resource_handler.py#L122) fails if we change it to repr

sanderr: Interesting. To be honest I personally prefer the `repr` output but if we were already using the other approach (I hadn't realized when I commented) I guess we should stick with that.

Hugo-Inmanta:
Still a couple of mypy errors that I didn't manage to handle on my own:
```sh
+ src/inmanta/agent/handler.py:1018: error: Argument 2 of "execute" is incompatible with supertype "HandlerAPI"; supertype defines the argument type as "Resource"  [override]
+ src/inmanta/agent/handler.py:1036: error: Argument 2 of "deploy" is incompatible with supertype "HandlerAPI"; supertype defines the argument type as "Resource"  [override]
+ src/inmanta/agent/handler.py:1036: note: This violates the Liskov substitution principle
+ src/inmanta/agent/handler.py:1036: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
+ src/inmanta/agent/handler.py:1071: error: Argument 2 of "execute" is incompatible with supertype "HandlerAPI"; supertype defines the argument type as "Resource"  [override]
+ src/inmanta/agent/handler.py:1071: note: This violates the Liskov substitution principle
+ src/inmanta/agent/handler.py:1071: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
+ src/inmanta/agent/handler.py:1089: error: Argument 1 to "json_encode" has incompatible type "D"; expected "Union[Union[BaseModel, Union[UUID, StrictNonIntBool, int, float, datetime, str]], Sequence[Union[BaseModel, Union[UUID, StrictNonIntBool, int, float, datetime, str]]], Mapping[str, Union[BaseModel, Union[UUID, StrictNonIntBool, int, float, datetime, str]]], None]"  [arg-type]

```

src/inmanta/agent/handler.py
```diff
@@ -961,15 +1019,113 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+          discovered and reported to the server. Objects of this type must be serializable.
+    """
+
+    def deploy(
+        self,
+        ctx: HandlerContext,
+        resource: R,
+        requires: abc.Mapping[ResourceIdStr, ResourceState],
+    ) -> None:
+        self.execute(ctx, resource)
```
wouterdb, line 1039: We may want to copy over some of the this code

```
 resources_in_unexpected_state = filter_resources_in_unexpected_state(requires)
        if resources_in_unexpected_state:
            ctx.set_status(const.ResourceState.skipped)
            ctx.warning(
                "Resource %(resource)s skipped because a dependency is in an unexpected state: %(unexpected_states)s",
                resource=resource.id.resource_version_str(),
                unexpected_states=str({rid: state.value for rid, state in resources_in_unexpected_state.items()}),
            )
            return

        failed_dependencies = [req for req, status in requires.items() if status != ResourceState.deployed]
        if not any(failed_dependencies):
            self.execute(ctx, resource)
            if _should_reload():
                self.do_reload(ctx, resource)
        else:
            ctx.set_status(const.ResourceState.skipped)
            ctx.info(
                "Resource %(resource)s skipped due to failed dependencies: %(failed)s",
                resource=resource.id.resource_version_str(),
                failed=str(failed_dependencies),
            )
 ```
 
 To handle skip and failed dependencies

File: src/inmanta/agent/handler.py [hunk 13]
```diff
@@ -961,15 +1019,113 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+            return self.get_client().discovered_resource_create_batch(
+                tid=self._agent.environment, discovered_resources=discovered_resources
+            )
+
+        try:
+            self.pre(ctx, resource)
+            # serialize resources and report to the server
+            discovered_resources: abc.Sequence[DiscoveredResource] = [
+                DiscoveredResource(discovered_resource_id=resource_id, values=json.loads(json_encode(values)))
+                for resource_id, values in self.discover_resources(ctx, resource).items()
```
wouterdb, line 1090: purely a matter of taste, but I would put the  `self.discover_resources(ctx, resource)` on a line of its own, as it is the main thing here. It is kind of hidden now.

src/inmanta/agent/handler.py
```diff
@@ -961,15 +1019,113 @@ def execute(self, ctx: HandlerContext, resource: TPurgeableResource, dry_run: bo
...
+            discovered_resources: abc.Sequence[DiscoveredResource],
+        ) -> typing.Awaitable[Result]:
+            return self.get_client().discovered_resource_create_batch(
+                tid=self._agent.environment, discovered_resources=discovered_resources
+            )
+
+        try:
+            self.pre(ctx, resource)
+            # serialize resources and report to the server
+            discovered_resources: abc.Sequence[DiscoveredResource] = [
```
wouterdb, line 1088: If we run a dryrun, do we want to execute all this?

wouterdb (changes requested):
Looks good overall

src/inmanta/agent/handler.py
Hugo-Inmanta: I lifted this logic in the HandlerAPI's deploy in the other PR

PR #6267, opened 2023-07-13
Issue/5102 docs dot installed

### LINKED ISSUE #5102: Building the documentation while dot is not installed
Our documentation build relies on the `sphinx.ext.graphviz` sphinx extension. This extension in its return relies on the fact that the `dot` command is available in the PATH. When `dot` is not installed, the extension logs a warning instead of raising an exception. As such, the documentation build fails on the absence of an image and it takes some time to find out what the actual issue is. This ticket should make sure that a clear error message is shown when the user tries to build the documentation without `dot` being installed.

Proposal: Check in the `docs/conf.py` whether `dot` is available.

FloLey:
# Description

Check in the docs/conf.py whether dot is available.

closes #5102 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5102-docs-make-sure-dot-in-path.yml
+++ b/changelogs/unreleased/5102-docs-make-sure-dot-in-path.yml
@@ -1,0 +1,5 @@
+description: Ensure that the dot command is present in the path when building the docs.
+issue-nr: 5102
+change-type: patch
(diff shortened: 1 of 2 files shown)


FloLey:
@sanderr Do you have an idea how I can test this?

PR #6270, opened 2023-07-13
Issue/6025 document discovery handler

Hugo-Inmanta:
# Description

- [x] add documentation including examples, see the design for more details.

part of https://github.com/inmanta/inmanta-core/issues/6025

# Self Check:

- [x] Changelog entry


docs/model_developers/unmanaged_resources/basic_example.cf
```diff
@@ -0,0 +1,15 @@
...
+    """
+    A discovery resource that discovers interfaces on a specific host.
+
+    :attr name_filter: If not null, only discover interfaces for which the name matches this regular expression.
+                       Otherwise discover all the interfaces on the host.
+    """
+    string? name_filter = null
+end
+
+InterfaceDiscovery.host [1] -- std::Host
```
sanderr, line 11: Shouldn't this be `ip::Host` to make it a valid model? I think we should add a test case that loads these files and verifies that the example code works (as far as possible).

docs/model_developers/unmanaged_resources/basic_example_handler.py
```diff
@@ -0,0 +1,61 @@
...
+class InterfaceDiscoveryHandler(DiscoveryHandler[InterfaceDiscovery, UnmanagedInterface]):
+    def discover_resources(
+        self, ctx: HandlerContext, discovery_resource: InterfaceDiscovery
+    ) -> abc.Mapping[ResourceIdStr, UnmanagedInterface]:
+        """
+        Entrypoint that is called by the agent when the discovery resource is deployed.
+        """
+        discovered: abc.Iterator[UnmanagedInterface] = (
+            UnmanagedInterface(**attributes) for attributes in self._get_discovered_resources(discovery_resource.host)
+            if discovery_resource.name_filter is not None and re.match(discovery_resource.name_filter, attributes["interface_name"])
```
sanderr, line 44:
```suggestion
            if discovery_resource.name_filter is None or re.match(discovery_resource.name_filter, attributes["interface_name"])
```
I think

docs/model_developers/unmanaged_resources/basic_example_handler.py
```diff
@@ -0,0 +1,61 @@
...
+    host: str
+    interface_name: str
+    ip_address: pydantic.IPvAnyAddress
+
+
+@provider("my_module::InterfaceDiscovery", name="interface_discovery_handler")
+class InterfaceDiscoveryHandler(DiscoveryHandler[InterfaceDiscovery, UnmanagedInterface]):
+    def discover_resources(
+        self, ctx: HandlerContext, discovery_resource: InterfaceDiscovery
+    ) -> abc.Mapping[ResourceIdStr, UnmanagedInterface]:
```
sanderr, line 38:
```suggestion
    ) -> dict[ResourceIdStr, UnmanagedInterface]:
```

Not very important to me, but ideally I think concrete method implementation should have abstract argument types and concrete return types, unless flexibility is required for future changes. In the ABC I like the `abc.Mapping` return type because it allows children some flexibility.

docs/model_developers/unmanaged_resources.rst
```diff
@@ -0,0 +1,88 @@
...
+       discovered resources.
+     * It must implement a method called ``discover_resources`` that contains the logic to discover the resources
+       in the network. This method returns a dictionary. The keys of this dictionary contain the resource ids of the
+       discovered resources and the values the associated attributes.
+
+
+Sharing attributes
+==================
+
+In some situation it might be handy to share a set of attributes between a resource X and the a discovery resource
```
sanderr, line 68: "handy" felt weird to me in this context but googling it didn't reveal any issues so it could be just my non-native English.

sanderr, line 68:
```suggestion
In some situation it might be handy to share a set of attributes between a resource X and the discovery resource
```

docs/model_developers/unmanaged_resources.rst
```diff
@@ -0,0 +1,88 @@
...
+for X. This can be done by making both entities inherit from a shared parent entity. An example is provided below.
+
+.. literalinclude:: unmanaged_resources/shared_attributes_example.cf
+    :language: inmanta
+    :caption: my_project/main.cf
+    :linenos:
+
+The ``Credentials`` entity, in the above-mentioned snippet, contains the shared attributes between the
```
sanderr, line 76: This feels like it's pulled out of thin air. I think it would read better if we introduce it in the previous paragraph:

"... resource for X. For example, both might require credentials to authenticate to their host. This can be done ..."

docs/model_developers/unmanaged_resources.rst
sanderr, line 68: We now only mention attributes, perhaps we should mention behavior (for the handler) as well? e.g. the authentication code.

docs/model_developers/unmanaged_resources.rst
```diff
@@ -0,0 +1,88 @@
...
+PurgeableResource ``Interface`` and the DiscoveryResource ``InterfaceDiscovery``.
+
+The associated handler code is provided below:
+
+.. literalinclude:: unmanaged_resources/shared_attributes_example_handler.py
+    :language: python
+    :caption: my_module/inmanta_plugins/__init__.py
+    :linenos:
+
+In the above-mentioned code snippet the ``Credentials`` class contains the shared attributes between the
```
sanderr, line 86: Did you verify that this works? I believe that it should, but I don't think we've ever done anything like this before.

docs/model_developers/unmanaged_resources/shared_attributes_example_handler.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@provider("my_module::Interface", name="interface_handler")
+class InterfaceHandler(Authenticator, CRUDHandler[Interface]):
+    """
+    Handler for the interfaces managed by the orchestrator.
+    """
+
+    def pre(self, ctx: HandlerContext, resource: Interface) -> None:
+        self.login(resource)
```
sanderr, line 69: Did you run mypy on this? Again, I think it should work, but best to double check.

docs/model_developers/unmanaged_resources/shared_attributes_example_handler.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def delete_resource(self, ctx: HandlerContext, resource: Interface) -> None:
+        raise NotImplementedError()
+
+    def update_resource(self, ctx: HandlerContext, changes: dict, resource: Interface) -> None:
+        raise NotImplementedError()
+
+
+@provider("my_module::InterfaceDiscovery", name="interface_discovery_handler")
+class InterfaceDiscoveryHandler(DiscoveryHandler[InterfaceDiscovery, UnmanagedInterface]):
```
sanderr, line 88:
```suggestion
class InterfaceDiscoveryHandler(Authenticator, DiscoveryHandler[InterfaceDiscovery, UnmanagedInterface]):
```

docs/model_developers/unmanaged_resources/shared_attributes_example_handler.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def discover_resources(
+        self, ctx: HandlerContext, discovery_resource: InterfaceDiscovery
+    ) -> abc.Mapping[ResourceIdStr, UnmanagedInterface]:
+        """
+        Entrypoint that is called by the agent when the discovery resource is deployed.
+        """
+        discovered: abc.Iterator[UnmanagedInterface] = (
+            UnmanagedInterface(**attributes) for attributes in self._get_discovered_resources(discovery_resource.host)
+            if discovery_resource.name_filter is not None and re.match(discovery_resource.name_filter, attributes["interface_name"])
```
sanderr, line 104:
```suggestion
            if discovery_resource.name_filter is None or re.match(discovery_resource.name_filter, attributes["interface_name"])
```

sanderr (changes requested):

FloLey (approved):
I think it is really clear and the examples help a lot

.gitignore
```diff
@@ -89,7 +89,13 @@ build-result
 _build
 logs
 
+# Documentation artifacts
 docs/reference/openapi.json
+docs/lsm/*.pdf
+docs/lsm/*.png
+docs/model_developers/images/*.pdf
+docs/model_developers/images/*.png
```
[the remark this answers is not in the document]
arnaudsjs: These lines are already present in the `.gitignore`. I'll remove them.

src/inmanta/agent/handler.py
```diff
@@ -980,7 +979,11 @@ class DiscoveryHandler(HandlerAPI[TDiscovery], Generic[TDiscovery, TDiscovered])
           conventional resource type expected to be deployed on a network, but rather a way to express
           the intent to discover resources of the second type TDiscovered already present on the network.
         - TDiscovered denotes the handler's Unmanaged Resource type. This is the type of the resources that have been
-          discovered and reported to the server. Objects of this type must be serializable.
+          discovered and reported to the server. Objects of this type must be either:
+             1. A pydantic object
+             2. An object that inherits from `inmanta.util.JSONSerializable`
```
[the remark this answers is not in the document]
arnaudsjs: Right. I didn't pay attention to whether or not these classes are part of the stable API. I dropped the option.

docs/model_developers/unmanaged_resources/basic_example_handler.py
```diff
@@ -28,20 +28,20 @@ class UnmanagedInterface(pydantic.BaseModel):
     """
     host: str
     interface_name: str
-    ip_address: pydantic.IPvAnyAddress
+    ip_address: str
```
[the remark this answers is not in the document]
arnaudsjs: I wanted to keep the type here close to the type in the model. In the model I used `std::ipv4_address` which maps to `ipaddress::IPv4Address`. The latter cannot be serialized by our customer json serializer. That's why I went to str.

sanderr (approved):

PR #6303, opened 2023-07-25
issue/6273-env-creation-500

### LINKED ISSUE #6273: Environment creation results in response code 500
The Inmanta server returns the response code 500 when the user tries to create an environment in a certain project with a name that already exists in that project. The endpoint to update an environment probably suffers from the same bug.

- [x] Fix the bug
- [x] Mention in the API documentation that the names of all environments in a certain project must have unique names.

FloLey:
# Description

The Inmanta server returns the response code 500 when the user tries to create an environment in a certain project with a name that already exists in that project. It will now return a 400

closes #6273 

# Self Check:

- [x] Changelog entry


src/inmanta/server/services/environmentservice.py
```diff
@@ -345,6 +345,11 @@ async def environment_create(
         description: str = "",
         icon: str = "",
     ) -> model.Environment:
+        # check if an environment with this name is already defined in this project
+        envs = await data.Environment.get_list(project=project_id, name=name)
+        if len(envs) > 0 and envs[0].id != environment_id:
```
sanderr, line 350: What if it does have the same id?

sanderr (changes requested):
One follow-up question, otherwise looks good.

src/inmanta/server/services/environmentservice.py
FloLey: Actually this should also raise the same exception I think. I changed it

sanderr (approved):

Issue #6074, opened 2023-05-31
We have a reference to iso7 in the iso6 docs 


wouterdb:
https://docs.inmanta.com/inmanta-service-orchestrator/6/install/2-install-server-with-docker.html#step-2-pull-the-image

- [ ] check if there is already a variable for the version number, if not add it and use it here and for other occurrences.

PR #6310, opened 2023-07-25
Issue/6074 iso6 docs reference iso7

Hugo-Inmanta:
# Description

Use the iso_major variable introduced in https://github.com/inmanta/inmanta-service-orchestrator/pull/406 PR

TODO:
- [ ] add import [Sphinx Substitution Extensions](https://pypi.org/project/Sphinx-Substitution-Extensions/#toc-entry-1)
>Where ? inmanta-dev-dependencies[sphinx] ?

Part of https://github.com/inmanta/inmanta-core/issues/6074

# Self Check:

- [ ] Changelog entry


.gitignore
```diff
@@ -55,6 +55,10 @@ docs/moduleguides
 docs/inmanta.pdf
 docs/_static/*.pdf
 docs/_static/*.png
+docs/lsm/*.png
+docs/lsm/*.pdf
+docs/model_developers/images/*.png
+docs/model_developers/images/*.pdf
```
sanderr, line 61: There is a pdf and a png committed at this location. They should either be excluded here or we should first delete them if they are not intended to be tracked.

The same seems to be true for `docs/_static/` but when I delete them git does pick up on it. Any clue how comes those aren't excluded? I don't find any other gitignore with exceptions.

docs/conf.py
```diff
@@ -11,10 +11,15 @@
 #
 # All configuration values have a default; values that are commented out
 # serve to show the default.
+import logging
 import shutil
 import sys, os, pkg_resources, datetime
+from distutils.version import Version
```
sanderr, line 17: `distutils` has been deprecated. We should use `packaging.version` instead.

docs/conf.py
```diff
@@ -102,6 +107,8 @@ def check_dot_command():
     #    of the inmanta-core documentation. During the ISO docs build,
     #    this value will be overwritten with the ISO product version.
     version = "1.0.0"
+    iso_major = int(version.split(".")[0])
```
sanderr, line 110: If we name this `version_major` we can do this unconditionally (we'd still use it conditionally though) so that it's always here and we can drop the PR on the iso repo.

docs/quickstart.rst
```diff
@@ -209,7 +209,7 @@ The first option, ``inmanta-cli``, will automatically create a ``.inmanta`` file
 If you have chosen the second option, the Web Console, you need to copy the environment ID for later use, either:
 
  - from the URL, e.g. ec05d6d9-25a4-4141-a92f-38e24a12b721 from the http://172.30.0.3:8888/console/desiredstate?env=ec05d6d9-25a4-4141-a92f-38e24a12b721.
- - or by clicking on the gear icon on the top right of the Web Console, then click on Environment, scroll down all the way to the bottom of the page and copy the environment ID.
+ - or by clicking on the `Settings` pane, then in the `Environment` tab, scroll down all the way to the bottom of the page and copy the environment ID.
```
sanderr, line 212: This reads better, nice

sanderr (changes requested):
Could you check if this covers all documentation tasks in the major release process? If so, we can probably drop that task from the list, could you open a PR for that (or just merge it)? If not, you're probably missing a reference in this PR.

sanderr:
I would add the requirement to core directly because it's an extension specific to what we're doing here.

sanderr:
(So `requirements.dev.txt`)

Hugo-Inmanta:
> Could you check if this covers all documentation tasks in the major release process? If so, we can probably drop that task from the list, could you open a PR for that (or just merge it)? If not, you're probably missing a reference in this PR.

TODO: Tasks to remove from the tasks in [the major release process](https://internal.inmanta.com/development/core/tasks/release-product.html#additional-steps-before-a-new-major-iso-release):
- [ ] Update references to the new version in the inmanta-service-orchestrator docs on the master branch:
    - [ ] installation instructions: update version and use newly created cloudsmith repos. Make sure to update all occurrences.
    - [ ] container config documentation (`docs/install/install-server-with-docker/`): update the version

docs/conf.py
Hugo-Inmanta: This is not actually being used, I'll remove it. Should I still change the call to `pkg_resources` to use `packaging` instead ?

.gitignore
Hugo-Inmanta: My intention behind this change is that when you run `make docs` locally, the local index gets polluted with .png and .pdf files.

Maybe there's another option.

>  they are not intended to be tracked.

I'll look into this on monday

>  Any clue how comes those aren't excluded? 

I have no idea !

docs/conf.py
sanderr: That would be good, though I think you're looking for `importlib.metadata.version`.

.gitignore
Hugo-Inmanta: > The same seems to be true for docs/_static/ but when I delete them git does pick up on it

It looks like git ignores deletions of png/pdf files in docs/_static/ as it should on my machine

sanderr: Very strange. I get
```
	deleted:    docs/_static/architecture.png
	deleted:    docs/_static/inmanta-logo.png
```

sanderr: Hmm, it might be just because we committed them at some point. If I delete them, commit, then put them back, git just ignores them. Perhaps we should add an explicit exception?
```diff
diff --git a/.gitignore b/.gitignore
index e5d41e92..7aeb121a 100644
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,8 @@ docs/moduleguides
 docs/inmanta.pdf
 docs/_static/*.pdf
 docs/_static/*.png
+!docs/_static/architecture.png
+!docs/_static/inmanta-logo.png

 # Mypy
 .mypy
 ```

Comment (sanderr) [reply .gitignore:61]: For the ones in `model_developers/` I think we should do the same for the png but we should delete the pdf instead because it is generated from the png. Could you double check that?

File: docs/conf.py (hunk 3 above)
Comment (Hugo-Inmanta) [reply docs/conf.py:109]: I don't fully get what you mean here.

File: .gitignore (hunk 1 above)
Comment (Hugo-Inmanta) [reply .gitignore:61]: I could be missing something but as far as I can tell we only use svg to create both pdf and png (https://github.com/inmanta/inmanta-core/blob/aba37a6560a95eda8c703b5bccfebb8b22fbdf98/docs/Makefile#L193-L197)

Comment (sanderr) [reply .gitignore:61]: Hmm, you're right. And yet when I delete the pdf and run `make docs` the image is still in the generated pdf.

Comment (sanderr) [reply .gitignore:61]: Let's just keep it if we're not sure, but we should add it to the gitignore.

File: docs/conf.py (hunk 3 above)
Comment (Hugo-Inmanta) [reply docs/conf.py]: I added a `version_major` variable that is set when performing an iso build (and used in iso-only parts of the docs).

**Comment by Hugo-Inmanta** [comment]
I added 2 variables in conf.py to set the gpg keys for iso and oss builds, I can add a task in the major release instructions to set these 2 ? e.g
```
- [ ] Set the corresponding gpg keys of these newly created cloudsmith repos in the conf.py file.
```

**Review by sanderr — COMMENTED** [review]
As discussed

File: docs/conf.py [hunk 5]
```diff
@@ -121,6 +122,34 @@ def check_dot_command():
 # The full version, including alpha/beta/rc tags.
 release = version
 
+
+iso_gpg_key: str
+oss_gpg_key: str
```
Hugo-Inmanta, line 127: The next step is to open some PRs on `inmanta-service-orchestrator` (to set `iso_gpg_key`) and `inmanta` (to set `oss_gpg_key`) in conf.py, one for each of iso{4.5.6} branch. 
Where do I find the right gpg key for each version/flavour ?

sanderr: We could choose to add the OSS key in this repo directly because it will never change (there is only 1 OSS version so only 1 repo). For iso I agree with your statement. You can find them on cloudsmith e.g. by opening the repo, then clicking "set me up" and picking Red Hat. Do you have a cloudsmith account?

docs/conf.py
```diff
@@ -121,6 +122,34 @@ def check_dot_command():
...
+    # if product's conf.py injected oss_gpg_key, use that one
+    oss_gpg_key
+except NameError:
+    # else set a dummy value
+    oss_gpg_key = "<gpg_key>"
+
+
+version_major = int(version.split(".")[0])
+rst_prolog = f"""
+    .. |iso_major| replace:: {version_major}
```
sanderr, line 146: Minor preference to keep it generic here as well for consistency.
```suggestion
    .. |version_major| replace:: {version_major}
```

sanderr (changes requested):
Nice, it looks a lot simpler now.

docs/conf.py
Hugo-Inmanta: > Do you have a cloudsmith account?

Yes I do, thanks for the instructions :+1:

Hugo-Inmanta: found it there => https://code.inmanta.com/inmanta-infra/infra/-/blob/master/files/inmanta-and-epel-9-x86_64.cfg

Hugo-Inmanta:
Follow-up PRs to set the gpg keys:
- [master](https://github.com/inmanta/inmanta-service-orchestrator/pull/406/files)
- [iso6](https://github.com/inmanta/inmanta-service-orchestrator/pull/413/files)
- [iso5](https://github.com/inmanta/inmanta-service-orchestrator/pull/414/files)

Hugo-Inmanta:
PR to update the major release task -> https://code.inmanta.com/inmanta/developerdocs/-/merge_requests/126

sanderr (approved):

arnaudsjs (approved):

PR #6311, opened 2023-07-25
issue/5511-doc-multiple-indexes

### LINKED ISSUE #5514: discourage multiple Python package indexes in documentation
Using more than one Python package index in the project config is generally discouraged. It is a security risk if one of the indexes is not fully controlled and internal packages are served from the other. The documentation should warn to use this with extreme care, and only when the user knows the risks of dependency confusion attacks.

FloLey:
# Description

add a warning to the docs about the risk of using multiple python package indexes

closes #5514 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5514-doc-warning-multiple-indexes.yml
+++ b/changelogs/unreleased/5514-doc-warning-multiple-indexes.yml
@@ -1,3 +1,5 @@
 description: add a warning to the docs about the risk of using multiple python package indexes
 change-type: patch
 destination-branches: [master, iso6]
+sections:
+  minor-improvement: "{{description}}"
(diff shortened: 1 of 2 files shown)


sanderr (changes requested):
I think we should link to the [pip documentation](https://pip.pypa.io/en/stable/cli/pip_install/) (which contains the same warning) and the draft [PEP 708](https://peps.python.org/pep-0708/#motivation) for more context

changelogs/unreleased/5514-doc-warning-multiple-indexes.yml
```diff
@@ -0,0 +1,3 @@
+description: add a warning to the docs about the risk of using multiple python package indexes
```
sanderr, line 1: I would also add a minor-improvement note

sanderr (approved):

PR #6321, opened 2023-07-28
Ensure that the id.attribute_value attribute of a Id has type string

arnaudsjs:
# Description

Fix bug where the `id.attribute_value` field of resources emitted by the exporter have a non-string type, when the type in the model is not a string. This issue caused problems in `pytest-inmanta`.

Attention points for reviewer:

* Please review carefully with respect to backwards compatibility issues.
* Can this be included in iso6?

# Self Check:

- [x] Changelog entry


sanderr (approved):
As far as I can see there shouldn't be any backwards compatibility issues (we already to the equivalent of `str()` when building the resource id string) so this can go into iso6.

Hugo-Inmanta (approved):
LGTM but I'm not yet confident to fully assess backwards compatibility

PR #6344, opened 2023-08-01
Added ability to look past incremental deploys

### LINKED ISSUE #5493: get_resource_events may leave events unhandled in case of incrmental deploy
This issue is purely theoretical and never observed. 

1. The idea of get_resource_events is `Get all events that should be processed by this specific resource, for the current deployment`
2. It does this by presenting all events since the last successful deploy.
3. However, an increment is ALSO a successful deploy, but it doesn't process events, so we might get a hole in the event processing coverage,..

I think this answer the question here: 

https://github.com/inmanta/inmanta-core/blob/5e7e7c887916f16497035daefbff218c351532e0/src/inmanta/server/services/resourceservice.py#L345

with yes

# Tasks
- [ ] implement event propagation part of the design: inmanta/designs#121

wouterdb:
# Description

Early review request to agree on the method used. 

closes  #5493

# Self Check:

- [x] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -4256,66 +4256,43 @@ async def get_resource_events(
...
-            if result["begin_status"] != const.ResourceState.deploying:
-                raise BadRequest(
-                    "Fetching resource events only makes sense when the resource is currently deploying. Current deploy state"
-                    f" for resource {resource_version_id_str} is {result['begin_status']}."
-                )
-            current_deploy_start = result["begin_started"]
-            last_deploy_start = result["started"]
-
-            # Step3: Get the resource
+            print("STage 1")
```
wouterdb, line 4268: todo

src/inmanta/data/__init__.py
```diff
@@ -4256,66 +4256,43 @@ async def get_resource_events(
...
 
+            if resource.status != const.ResourceState.deploying:
+                raise BadRequest(
+                    "Fetching resource events only makes sense when the resource is currently deploying. Current deploy state"
+                    f" for resource {resource_version_id_str} is {resource.status}."
+                )
+
+            # steps 2:
+            # find the interval between the current deploy (now) and the previous successful deploy
+            print("Stage 2:", resource.last_success)
```
wouterdb, line 4285: todo

src/inmanta/server/services/resourceservice.py
```diff
@@ -582,6 +588,7 @@ async def resource_action_update(
         change: const.Change,
         send_events: bool,
         keep_increment_cache: bool = False,
+        is_increment_notification: bool = False,
```
wouterdb, line 591: We may want to get rid of this method in iso7

arnaudsjs (changes requested):
The method used looks good to me.

src/inmanta/data/__init__.py
```diff
@@ -4967,6 +4946,24 @@ async def get_resources_in_resource_sets(
...
+        version: int,
+    ) -> None:
+        last_released = await ConfigurationModel.get_latest_version(environment)
+        if not last_released:
+            return
+        previous_version = last_released.version
+        query = (
+            f"UPDATE {cls.table_name()} as new_resource "
+            f"SET last_success = (SELECT last_success from {cls.table_name()} as old_resource where old_resource.model=$3 and old_resource.environment=$2 and old_resource.resource_id=new_resource.resource_id)"
+            f"where new_resource.model=$1 and new_resource.environment=$2 and new_resource.last_success is null"
```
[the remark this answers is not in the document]
wouterdb: I tried to improve it, but I don't know the convention

src/inmanta/data/__init__.py
```diff
@@ -4157,6 +4157,7 @@ async def query_resource_actions(
         first_timestamp: Optional[datetime.datetime] = None,
         last_timestamp: Optional[datetime.datetime] = None,
         action: Optional[const.ResourceAction] = None,
+        resource_id: Optional[ResourceIdStr] = None,
```
wouterdb, line 4160: I don't like this method breaking up the id's so I updated it to be able to handle complete id's

tests/db/migration_tests/test_v202306060_to_v202308010.py
```diff
@@ -0,0 +1,60 @@
...
+            # verify time on last success
+            actions = await ResourceAction.query_resource_actions(
+                environment=env.id, resource_id=resource.resource_id, action=const.ResourceAction.deploy
+            )
+            last_deploy = actions[0]
+            last_deploy.version == 2
+            assert resource.last_success == last_deploy.finished
+            expected += 1
+        if resource.resource_id == "std::File[localhost,path=/tmp/test]":
+            # always fails
```
arnaudsjs, line 54: Can you elaborate on why this one always fails?

src/inmanta/db/versions/v202308010.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    await connection.execute(schema)
+
+    # For each environment
+    # For the latest released version
+    # Set the last success
+    update_query = """
+    UPDATE resource as r
+    SET last_success = (
+        SELECT max(started)
```
arnaudsjs, line 31: Why do we use `started` here instead of finished?

arnaudsjs (changes requested):

src/inmanta/db/versions/v202308010.py
wouterdb:
```suggestion
        SELECT max(finished)
```

arnaudsjs (approved):

changelogs/unreleased/5493-get_events_accross_incremental_deploys.yml
```diff
@@ -0,0 +1,6 @@
+description: Ensure get_resource_events works across incremental compiles
```
sanderr, line 1:
```suggestion
description: Ensure get_resource_events and resource_did_dependency_change work across incremental compiles
```

src/inmanta/data/__init__.py
```diff
@@ -4256,66 +4261,41 @@ async def get_resource_events(
...
 
+            if resource.status != const.ResourceState.deploying:
+                raise BadRequest(
+                    "Fetching resource events only makes sense when the resource is currently deploying. Current deploy state"
+                    f" for resource {resource_version_id_str} is {resource.status}."
+                )
+
+            # steps 2:
+            # find the interval between the current deploy (now) and the previous successful deploy
+            last_deploy_start = resource.last_success
```
sanderr, line 4289: 1. Might `last_successful_execute` be a better name? To make clear that it's not just about success "due to known good state" but explicit success due to handler execution? Completely optional suggestion.
2. As discussed, if we make this the metric to determine the window within which we look at events, it should be a part of these methods' contract/docstring and the handler's reload documentation. Perhaps with an explicit clarification that this means that a handler that uses these methods to monitor or respond to dependencies' events should mark its resource as deployed iff it successfully and completely processed all current events.

src/inmanta/server/services/resourceservice.py
```diff
@@ -539,10 +540,15 @@ async def resource_deploy_done(
                 if not keep_increment_cache:
                     self.clear_env_cache(env)
 
+                extra_fields = {}
+                if status == ResourceState.deployed:
+                    extra_fields["last_success"] = finished
```
sanderr, line 545: If we use `finished` here, don't we risk missing events between when `get_resource_events` was called and when `resource_deploy_done` is called? I'd think we either need to use `started` (meaning we also need to revert the change you made to the filter higher up) or allow the handler to explicitly register the time at which it looked at events (which would also require an update of the filter).

src/inmanta/db/versions/v202308010.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    await connection.execute(schema)
+
+    # For each environment
+    # For the latest released version
+    # Set the last success
+    update_query = """
+    UPDATE resource as r
+    SET last_success = (
+        SELECT max(finished)
```
sanderr, line 31: Related to other comment about using `finished` as marker. Though in this case it may be our best option regardless.

src/inmanta/server/services/resourceservice.py
sanderr, line 591: Could you add this to the method docstring?

sanderr (changes requested):

src/inmanta/data/__init__.py
wouterdb: for point 2, I added some docs to the relevant api endpoint, but it was already quite good

for reload, I'll wait for https://github.com/inmanta/inmanta-core/pull/6246

src/inmanta/db/versions/v202308010.py
wouterdb: modified

tests/db/migration_tests/test_v202306060_to_v202308010.py
```diff
@@ -0,0 +1,61 @@
...
+    for resource in resources:
+        if resource.resource_id == "std::AgentConfig[internal,agentname=localhost]":
+            # always success
+            assert resource.last_success
+            # verify time on last success
+            actions = await ResourceAction.query_resource_actions(
+                environment=env.id, resource_id=resource.resource_id, action=const.ResourceAction.deploy
+            )
+            last_deploy = actions[1]
+            assert last_deploy.version == 2
```
wouterdb, line 50: I had a missing assert here, which allowed me to find a bug

src/inmanta/server/services/resourceservice.py
```diff
@@ -592,6 +592,12 @@ async def resource_action_update(
         *,
         connection: Optional[Connection] = None,
     ) -> Apireturn:
+        """
+        :param is_increment_notification: is this the increment calucation setting the deployed status,
+            instead of an actual deploy? Used to keep track of the last_success field on the resources,
+            which should not be updated for increments.
```
sanderr, line 598: Shouldn't this be documented on `methods.resource_action_update` instead?

src/inmanta/data/__init__.py
sanderr: > I added some docs to the relevant api endpoint

Did you push this?

sanderr (changes requested):

src/inmanta/data/__init__.py
wouterdb: I did not

src/inmanta/server/services/resourceservice.py
wouterdb: no, this argument doens't exist on the rest api, only internally.

sanderr (changes requested):
Still [one](https://github.com/inmanta/inmanta-core/pull/6344#discussion_r1284212187) unresolved comment.

sanderr (approved):

sanderr:
> Still https://github.com/inmanta/inmanta-core/pull/6344#discussion_r1284212187 unresolved comment.

Race condition

PR #6348, opened 2023-08-02
Issue/6074 add iso major variable to conf py iso5

Hugo-Inmanta:
# Description

Follow-up PR to https://github.com/inmanta/inmanta-core/pull/6310 specific to iso5.

Part of https://github.com/inmanta/inmanta-core/issues/6074

# Self Check:

- [ ] Changelog entry


requirements.dev.txt
sanderr, line 1: Is this change required? I explicitly left iso5 out because we don't bump its versions anymore.

Hugo-Inmanta: Ah maybe not. I did it this way because I thought the sphinx extra was removed from `inmanta-dev-dependencies`, but it is still present for the 2.53.0 tag we're using here. Let me try again

sanderr (approved):

PR #6360, opened 2023-08-04
increased max_iterations

wouterdb:
# Description

Increase the default value of INMANTA_MAX_ITERATIONS to 100000

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/execute/scheduler.py
+++ b/src/inmanta/execute/scheduler.py
@@ -55,7 +55,7 @@ DEBUG = True
 LOGGER = logging.getLogger(__name__)
 
 
-MAX_ITERATIONS = 100000
+MAX_ITERATIONS = 100_000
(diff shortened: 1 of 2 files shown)


src/inmanta/execute/scheduler.py
```diff
@@ -55,7 +55,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
-MAX_ITERATIONS = 10000
+MAX_ITERATIONS = 100000
```
sanderr, line 58:
```suggestion
MAX_ITERATIONS = 100_000
```
Slightly easier to count.

sanderr (approved):

Hugo-Inmanta (approved):
I can't recall the reason why we are increasing this default

PR #6362, opened 2023-08-04
Add documentation regarding annotations

arnaudsjs:
# Description

Add documentation regarding annotations that can be set on service entities, embedded entities and LSM exposed attributes.

Part of inmanta/lsm#604

# Self Check:

- [x] Changelog entry


docs/lsm/attributes_metadata/attributes_metadata.rst
```diff
@@ -85,3 +85,90 @@ Supported values
...
+
+    entity Interface extends lsm::ServiceEntity:
+        string interface_name
+        dict interface_name__annotations = {"annotation": "value"}
+    end
+
+Annotations on relational attributes
+####################################
+
+Annotations can be attach to a relational attribute by replacing the ``--`` part of the relationship definition with
```
sanderr, line 147:
```suggestion
Annotations can be attached to a relational attribute by replacing the ``--`` part of the relationship definition with
```

docs/lsm/attributes_metadata/attributes_metadata.rst
```diff
@@ -85,3 +85,90 @@ Supported values
...
+an instance of the ``lsm::RelationAnnotations`` entity. This entity has a dict attribute ``annotations`` that
+represent the annotation that should be set on the relational attribute. The values of this dictionary must
```
sanderr, line 149:
```suggestion
represents the annotation that should be set on the relational attribute. The values of this dictionary must
```

docs/lsm/attributes_metadata/attributes_metadata.rst
```diff
@@ -85,3 +85,90 @@ Supported values
...
+be strings. The annotations are always attached to the attribute that is exposed via the lsm API, i.e. the attribute
```
sanderr, line 150: I don't get what you mean by this sentence. Are you trying to say that you can't / it doesn't make sense to annotate non-API attributes/relations?

sanderr (changes requested):

docs/lsm/attributes_metadata/attributes_metadata.rst
```diff
@@ -85,3 +85,90 @@ Supported values
...
+
+Annotations
+~~~~~~~~~~~
+
+Definition
+##########
+
+Annotations are key-value pairs that can be associated with an entity (service entity or embedded entity) or
+an attribute (simple attribute or relational attribute). These annotations are intended for the web-console to
+improve visualizations and the user-experience.
```
wouterdb, line 100:
```suggestion
Annotations are key-value pairs that can be associated with an entity (service entity or embedded entity) or an attribute (simple attribute or relational attribute). These annotations  don't influence the behavior of LSM or the Inmanta Service Orchestrator itself, but are intended to pass meta data to other components. For example, they can be used to pass on visualization meta-data to the the web-console to improve the user-experience.
```

not sure about this, just a suggestion.

docs/lsm/attributes_metadata/attributes_metadata.rst
wouterdb: me neither

docs/lsm/attributes_metadata/attributes_metadata.rst
arnaudsjs: I wasn't sure how to write this part, because the use cases are open ended at the moment. I like the suggestion,

docs/lsm/attributes_metadata/attributes_metadata.rst
arnaudsjs: I mean: It's a bidirectional relationship, but only the attribute from the parent to the child is part of the API to create/modify/delete the service instance.

wouterdb (approved):

docs/lsm/attributes_metadata/attributes_metadata.rst
sanderr: I'm not entirely sure but I feel like it may sow more confusion than it fixes as it stands now.

sanderr (commented):
See comment in open discussion

docs/lsm/attributes_metadata/attributes_metadata.rst
arnaudsjs: ok. I will remove it.

sanderr (approved):

PR #6366, opened 2023-08-07
Make sure that stop method of FeatureManager is called

arnaudsjs:
# Description

This PR makes the following changes:

* Make sure that calling `BootLoader.stop()` calls the stop() method on the FeatureManager created by the `BootLoader.load_slices()` invocation.
* Make the `stop()` method of the `FeatureManager` class asynchronous. This change is required for https://github.com/inmanta/inmanta-license/pull/528

Part of inmanta/inmanta-license#520

# Self Check:

- [x] Changelog entry


FloLey (approved):

sanderr (approved):
I'm adding @wouterdb to check for any compatibility issues with making this method async. In practice this would mean we tie these versions of inmanta-license and inmanta-core together. I think that's acceptable.

wouterdb (approved):
Most unfortunate, this will wreak havoc on the version constraints in inmanta-license. 

It is in fact a breaking breaking change (major, not patch)

PR #6374, opened 2023-08-09
Issue/6092 repair runs out of specific timerange

### LINKED ISSUE #6092: Make sure repair runs can be kept out of a specific time range
# Goal

We want to be able to have repair runs happen only at night

# Quick fix

1. make a script to make sure this can be ensured on existing orchestrator (e.g. by re-setting it in the night, so it runs then and start the timer then)

# Final fix

1. make sure we can support this better in the future, e.g. by allowing a cron to be set as repair interval .

FloLey:
# Description

allow both a number and a cron expression in the agent_repair_interval so that we can specify a timerange where the repair runs happen. 

closes #6092 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -29,7 +29,7 @@ from asyncio import Lock
 from collections import defaultdict
 from concurrent.futures.thread import ThreadPoolExecutor
 from logging import Logger
-from typing import Any, Dict, Iterable, List, Optional, Sequence, Set, Tuple, cast
+from typing import Any, Dict, Iterable, List, Optional, Sequence, Set, Tuple, Union, cast
 
 from inmanta import const, data, env, protocol
 from inmanta.agent import config as cfg
@@ -634,7 +634,7 @@ class AgentInstance(object):
         self._deploy_splay_value = random.randint(0, deploy_splay_time)
 
         # do regular repair runs
-        self._repair_interval = cfg.agent_repair_interval.get()
+        self._repair_interval: Union[int, str] = cfg.agent_repair_interval.get()
         repair_splay_time = cfg.agent_repair_splay_time.get()
         self._repair_splay_value = random.randint(0, repair_splay_time)
 
@@ -738,7 +738,7 @@ class AgentInstance(object):
             )
             self._enable_time_trigger(deploy_action, interval_schedule_deploy)
         if isinstance(self._repair_interval, int):
-            if not self._repair_interval:
+            if self._repair_interval <= 0:
                 return
             self.logger.info(
                 "Scheduling repair with interval %d and splay %d (first run at %s)",
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -738,7 +738,7 @@ class AgentInstance(object):
             )
             self._enable_time_trigger(deploy_action, interval_schedule_deploy)
         if isinstance(self._repair_interval, int):
-            if self._repair_interval == 0:
+            if not self._repair_interval:
                 return
             self.logger.info(
                 "Scheduling repair with interval %d and splay %d (first run at %s)",
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -44,7 +44,16 @@ from inmanta.loader import CodeLoader, ModuleSource
 from inmanta.protocol import SessionEndpoint, SyncClient, methods, methods_v2
 from inmanta.resources import Id, Resource
 from inmanta.types import Apireturn, JsonType
-from inmanta.util import IntervalSchedule, NamedLock, ScheduledTask, TaskMethod, add_future, join_threadpools
+from inmanta.util import (
+    CronSchedule,
+    IntervalSchedule,
+    NamedLock,
+    ScheduledTask,
+    TaskMethod,
+    TaskSchedule,
+    add_future,
+    join_threadpools,
+)
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


src/inmanta/config.py
```diff
@@ -207,6 +208,17 @@ def is_time(value: str) -> int:
     return int(value)
 
 
+def is_time_or_cron(value: str) -> Union[int, str]:
+    """Time, the number of seconds represented as an integer value"""
+    if isinstance(value, int):
```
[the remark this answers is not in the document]
Hugo-Inmanta: I don't quite understand the issue it looks to me that Config.set("config", "agent-repair-splay-time", "2") is already being set in the test ?
I tried to reproduce but I couldn't, it looks to me value is always a string here? I think we can get rid of the `isinstance` check and update the docstring to reflect that `value` can also be a cron expression.

Hugo-Inmanta (changes requested):
Minor changes, LGTM otherwise :+1:

src/inmanta/config.py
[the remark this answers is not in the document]
FloLey: I made a mistake, I was talking about agent-repair-interval.
you need to remove the line so that this part of the code is used:
```
agent_repair_interval = Option(
    "config",
    "agent-repair-interval",
    600,
    "The number of seconds between two repair runs (full deploy) of the agent. "
    + "Set this to 0 to disable the scheduled repair runs.",
    is_time,
)
```
and here 600 isn't a string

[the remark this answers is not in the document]
sanderr: Yeah, the typing is all wrong for the config/option framework... Since it's not trivial to clean up I would just follow the approach of the `is_bool` validator: accept `Union[int, str]`. Your implementation already matches that, your type signature does not.

src/inmanta/agent/config.py
```diff
@@ -116,9 +116,9 @@
     "config",
     "agent-repair-interval",
     600,
-    "The number of seconds between two repair runs (full deploy) of the agent. "
+    "The number of seconds between two repair runs (full deploy) of the agent or a cron expression of when to run the repair run"
```
sanderr, line 119: I think we need to be a bit more verbose on what this entails. The `auto_full_compile` environment setting already has some more details.

src/inmanta/agent/agent.py
```diff
@@ -724,18 +733,30 @@ async def repair_action() -> None:
                 self._deploy_splay_value,
                 (now + datetime.timedelta(seconds=self._deploy_splay_value)).strftime(const.TIME_LOGFMT),
             )
-            self._enable_time_trigger(deploy_action, self._deploy_interval, self._deploy_splay_value)
-        if self._repair_interval > 0:
+            interval_schedule_deploy: IntervalSchedule = IntervalSchedule(
+                interval=float(self._deploy_interval), initial_delay=float(self._deploy_splay_value)
+            )
+            self._enable_time_trigger(deploy_action, interval_schedule_deploy)
+        if isinstance(self._repair_interval, int):
```
sanderr, line 740: If we accept a dual type here I think we should make it clear in the type definition in the constructor.

src/inmanta/agent/agent.py
```diff
@@ -724,18 +733,30 @@ async def repair_action() -> None:
...
                 (now + datetime.timedelta(seconds=self._repair_splay_value)).strftime(const.TIME_LOGFMT),
             )
-            self._enable_time_trigger(repair_action, self._repair_interval, self._repair_splay_value)
+            interval_schedule_repair: IntervalSchedule = IntervalSchedule(
+                interval=float(self._repair_interval), initial_delay=float(self._repair_splay_value)
+            )
+            self._enable_time_trigger(repair_action, interval_schedule_repair)
+
+        if isinstance(self._repair_interval, str):
+            self.logger.info("Scheduling repair with cron expression %s", self._repair_interval)
```
sanderr, line 755:
```suggestion
            self.logger.info("Scheduling repair with cron expression '%s'", self._repair_interval)
```

sanderr (approved):

sanderr (changes requested):
Seems like I keep missclicking approve.

src/inmanta/agent/config.py
```diff
@@ -116,8 +116,11 @@
     "config",
     "agent-repair-interval",
     600,
-    "The number of seconds between two repair runs (full deploy) of the agent or a cron expression of when to run the repair run"
-    + "Set this to 0 to disable the scheduled repair runs.",
+    "Either the number of seconds between two repair runs (full deploy) of the agent or a cron-like expression"
+    "If a cron-like expression is specified, a repair will be run following a cron-like time-to-run specification, interpreted in UTC"
+    "(e.g. `min hour dom month dow`). A repair will be requested at the scheduled time. Note that if a cron expression is use"
```
sanderr, line 121:
```suggestion
    "(e.g. `min hour dom month dow`). A repair will be requested at the scheduled time. Note that if a cron expression is used"
```

src/inmanta/agent/config.py
```diff
@@ -116,8 +116,11 @@
...
+    "the 'agent_repair_splay_time' setting will be ignored"
+    "Setting this to 0 to disable the scheduled repair runs.",
```
sanderr, line 123: This whole block is missing spaces, I hadn't noticed before.

src/inmanta/config.py
```diff
@@ -208,7 +208,7 @@ def is_time(value: str) -> int:
     return int(value)
 
 
-def is_time_or_cron(value: str) -> Union[int, str]:
+def is_time_or_cron(value: Union[bool, str]) -> Union[int, str]:
```
sanderr, line 211:
```suggestion
def is_time_or_cron(value: Union[int, str]) -> Union[int, str]:
```

sanderr (changes requested):

src/inmanta/agent/config.py
FloLey: indeed, I added them

sanderr (approved):

src/inmanta/config.py
```diff
@@ -207,6 +208,17 @@ def is_time(value: str) -> int:
     return int(value)
 
 
+def is_time_or_cron(value: Union[int, str]) -> Union[int, str]:
+    """Time, the number of seconds represented as an integer value"""
```
Hugo-Inmanta, line 212: This docstring is not accurate

Hugo-Inmanta (changes requested):
One last docstring to rework, looks good otherwise :+1:

Hugo-Inmanta (approved):

PR #6378, opened 2023-08-10
update

FloLey:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = pep8,py39,mypy,docs
+envlist = pep8,py311,mypy,docs
 skip_missing_interpreters=True
 requires = pip
            virtualenv >= 20.2.2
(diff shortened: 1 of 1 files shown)


Hugo-Inmanta:
I checked the output of the build (https://jenkins.inmanta.com/blue/rest/organizations/jenkins/pipelines/core/pipelines/inmanta-core/branches/issue%252F6022-python311/runs/2/log/?start=0) and it looks like there's an error:
```
FileNotFoundError: [Errno 2] No such file or directory: 'junit-py39.xml'
```
---
I'm quite surprised that all the tests succeed :thinking:

Hugo-Inmanta (changes requested):
Looks like there's an error in the build that we should address.

changelogs/unreleased/6022-use-python-311.yml
```diff
@@ -0,0 +1,6 @@
+description: support python3.11
+issue-nr: 6022
+change-type: major
+destination-branches: [master]
+sections:
+  feature: "{{description}}"
```
sanderr, line 6:
```suggestion
```
Let's leave this for when we actually bump the Python version.

sanderr (changes requested):
I agree with @Hugo-Inmanta that we need to address the issue he reported. I think it is a problem with the xml hardcoding in `irt.predictive_test_selection.data_collector` so we'd have to fix it in that repo.

sanderr (approved):

changelogs/unreleased/6022-use-python-311.yml
```diff
@@ -0,0 +1,4 @@
+description: support python3.11
+issue-nr: 6022
+change-type: major
```
sanderr, line 3:
```suggestion
change-type: patch
```

Had missed this before. This is not a major. Bumping the actual minimal version might be but we're not there yet.

sanderr (changes requested):

Hugo-Inmanta (approved):

PR #6379, opened 2023-08-10
Issue/race on iso6 tests

wouterdb:
# Description

Fix for race where agent would not be fully up, applied on all place I could find at first sight 

The file this originates from no longer exists on master, so I made two PR's 
@sanderr will this work with mergetool?

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/race_in_server_tests.yml
+++ b/changelogs/unreleased/race_in_server_tests.yml
@@ -1,0 +1,4 @@
+description: Fix race condition in server tests
+change-type: patch
+destination-branches: [master]
+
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/race_in_server_tests.yml
```diff
@@ -0,0 +1,4 @@
+description: Fix race condition in server tests
+change-type: patch
+destination-branches: [master]
```
sanderr, line 3:
```suggestion
destination-branches: [iso6]
```

sanderr (approved):
Other than the requested change this will work fine.

FloLey (approved):

PR #6384, opened 2023-08-11
added fix for shutdown race

wouterdb:
# Description

Fix race that may cause testcases to fail

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy.py
+++ b/src/inmanta/deploy.py
@@ -482,15 +482,12 @@ host=localhost
         loud_logger = logging.getLogger("tornado")
         loud_logger.propagate = True
 
-        join_list: List[subprocess.Popen] = []
         if hasattr(self, "_server_proc"):
             self._server_proc.terminate()
-            join_list.append(self._server_proc)
 
         if hasattr(self, "_postgresproc"):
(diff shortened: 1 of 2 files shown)


src/inmanta/deploy.py
```diff
@@ -482,8 +482,15 @@ def stop(self) -> None:
...
         loud_logger.propagate = True
 
+        join_list: List[subprocess.Popen] = []
         if hasattr(self, "_server_proc"):
             self._server_proc.terminate()
+            join_list.append(self._server_proc)
 
         if hasattr(self, "_postgresproc"):
             self._postgresproc.stop()
+            join_list.append(self._postgresproc)
```
Hugo-Inmanta, line 492: Not too sure about the typing for the list PostgresProc vs subprocess.Popen

Hugo-Inmanta (changes requested):
One minor typing concern regarding the join_list

src/inmanta/deploy.py
wouterdb: thanks, good catch

Hugo-Inmanta (approved):

PR #6396, opened 2023-08-16
Add support for rounded timestamps on environment metrics

### LINKED ISSUE #6051: upstream metrics query improvements from grafana
- [x] Use nicely rounded timestamps instead of counting back exactly from now. See [grafana dashboard]: open up the "lsm.service_count" panel with "edit". Note how if you hover the graph you only see nicely rounded interval boundaries. Click "query inspector", note how the `$__timeGroup` macro is expanded to a bucket value similar to our own but rounding down to the nearest timestamp that is divisible by the interval width, ensuring more readable boundaries.
- [x] use `AVG` instead of `sum(count)::float)/(count(*)::float`
- [x] Consider using a sliding window query like those on grafana, aggregating not only over the interval itself but over a wider history. This one requires more discussion, may or may not be desired. If desired, open up the "compile rate sliding window" panel with "edit", then click "query inspector". Note especially the `generate_series` subquery and the subsequent join. We may want to delegate this checkbox to some time in the future when we have better insights in how useful the sliding window is.

arnaudsjs:
# Description

This PR adds support to the `GET /metrics` endpoint to round the returned timestamps to the previous hour

closes #6051

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/environment_metrics_service.py
+++ b/src/inmanta/server/services/environment_metrics_service.py
@@ -310,21 +310,54 @@ class EnvironmentMetricsService(protocol.ServerSlice):
                 COLLECTION_INTERVAL_IN_SEC,
             )
 
+    def _round_timestamp_to_hour(self, timestamp: datetime, round_up: bool) -> datetime:
+        """
+        Round the given timestamp to the hour.
+
+        :param timestamp: The timestamp that needs to be rounded.
+        :param round_up: If True, round to the next hour. If false, round to the previous hour.
+        :return: The rounded timestamp. If the timestamp is already rounded, this method returns the provided timestamp
+                 unchanged.
+        """
+        if timestamp.minute != 0 or timestamp.second != 0 or timestamp.microsecond != 0:
+            timestamp = timestamp.replace(minute=0, second=0, microsecond=0)
+            if round_up:
+                timestamp = timestamp.replace(hour=timestamp.hour + 1)
+        return timestamp
+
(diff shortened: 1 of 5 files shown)


FloLey (approved):

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -321,6 +321,8 @@ def _divide_time_interval_in_time_windows(
         total_seconds_in_interval = (end_interval - start_interval).total_seconds()
         seconds_per_window = math.floor(total_seconds_in_interval / nb_time_windows)
         result = [end_interval - timedelta(seconds=seconds_per_window) * i for i in range(nb_time_windows)]
+        if round_timestamps_to_previous_hour:
```
wouterdb, line 324: won't this lead to funny effects? 

e.g.
- if the window <1h we get intervals of size 0
- if the window is not divisible by 1h, we sometimes get additional intervals of size 0


I would instead, if we round:
1. round the interval boundaries to the hour
2. round up the seconds_per_window to hour (shortest interval is one hour)

the number of intervals may then be different, but at least it will behave regularly 
I would document that when round_timestamps_to_previous_hour is set, we can return less datapoints

wouterdb (changes requested):

wouterdb (changes requested):

wouterdb (approved):

FloLey (approved):

PR #6415, opened 2023-08-21
Issue/6025 implement discovery handler (2)

arnaudsjs:
# Description

- [x] implement `DiscoveryHandler`
- [x] Verify that the `Id` constructor and `resource_str` method are part of the stable API and documented.

This PR replaces #6264

# Self Check:

- [ ] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (#6270)

### DIFF
--- a/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
+++ b/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
@@ -1,0 +1,6 @@
+description: "Add handler (DiscoveryHandler) to discover unmanaged resources."
+issue-nr: 6025
+change-type: minor
+destination-branches: [master, iso6]
+sections:
+  feature: "{{description}}"
--- a/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
+++ b/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
@@ -4,4 +4,3 @@ change-type: minor
 destination-branches: [master]
 sections:
   feature: "{{description}}"
-  upgrade-note: "The execution of a handler's do_reload() and execute() now execute between `pre()` and `post()` have been called."
--- a/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
+++ b/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
@@ -1,6 +1,6 @@
 description: "Add handler (DiscoveryHandler) to discover unmanaged resources."
 issue-nr: 6025
 change-type: minor
-destination-branches: [master, iso6]
+destination-branches: [master]
 sections:
   feature: "{{description}}"
--- a/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
+++ b/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
@@ -4,3 +4,4 @@ change-type: minor
 destination-branches: [master]
 sections:
   feature: "{{description}}"
+  upgrade-note: "The execution of do_reload() and execute() now execute between `pre()` and `post()` have been called."
--- a/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
+++ b/changelogs/unreleased/6025-implement-discovery-resource-handler.yml
@@ -4,4 +4,4 @@ change-type: minor
 destination-branches: [master]
 sections:
   feature: "{{description}}"
-  upgrade-note: "The execution of do_reload() and execute() now execute between `pre()` and `post()` have been called."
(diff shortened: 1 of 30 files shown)


src/inmanta/agent/handler.py
```diff
@@ -529,15 +518,53 @@ def filter_resources_in_unexpected_state(
             )
 
     @abstractmethod
+    def _do_execute(self, ctx: HandlerContext, resource: TResource, dry_run: bool = False) -> None:
```
wouterdb, line 521: This is getting very very complicated.

We have 'deploy', 'execute'  and now `_do_execute` and `post_execute`. 

This class was supposed to be a fairly clean interface between agent and handler. I was reluctant against adding an implementation to the deploy method at all (https://github.com/inmanta/inmanta-core/pull/6246#discussion_r1282907025) but I thinks this takes it too far. 

I think, with this structure, there is so much inversion of control, that to know what each sub class does, you have to read up-and-down the inheritance hierarchy and piece it back together.  These methods have no clean function of their own, only in relation to the implementation of other methods. 

I see why you lifted out this error handling code, but I wonder if there is a better way.

For post-execute, I think the easiest way of removing this is to have the subclass override execute and do 
```python
def execute(...):
   super().execute(..._
   if _should_reload():
                self.do_reload(ctx, resource)
```

For the exception handling, I wonder if we can lift it up/duplicate it higher up in the call stack, that would both make it more robust and more out-of-sight?

wouterdb (changes requested):

src/inmanta/agent/handler.py
sanderr: I mostly agree with this. A few additional notes:
1. I think we could consider keeping `_do_execute` iff we more clearly distinguish the implementor interface from the user interface, keeping the footprint for the handler developer small and intuitive.
2. Even though I think we can get away with it, I don't necessairly think we should. Is there a reason we can't lift the error handling and the call to pre and post into `deploy` directly? This may be what Wouter is proposing as well, I'm not sure.
3. For the post execute I wholeheartedly agree.

sanderr: Ah, 2 and 3 are not compatible...

arnaudsjs: > I think we could consider keeping _do_execute iff we more clearly distinguish the implementor interface.

I tried to make it clear with the underscore that this was not part of the public API.

> Is there a reason we can't lift the error handling and the call to pre and post into deploy directly? This may be what Wouter is proposing as well, I'm not sure.

Wouter was proposing to move it out of the hander. That means moving it into the ResourceAction for example.

arnaudsjs: The main question is: Do you feel the same way as Wouter. Namely that this implementation results in very complicates and hard to understand code. And if so, do you see a good way to keep the benefits while making it easier to understand. If we cannot come to a better implementation, I would just roll it back and then we just trade better readability for code duplication.

arnaudsjs: I am fine with moving the error handling logic into `deploy()`. However, this would mean we also have to move the `pre()` and `post()` methods to `deploy()`. So the behavior will change when someone overrides the `execute()` method directly. I can't find a module that does that. If we don't have one, I think we can safely move the error handling logic into `deploy()`.

sanderr: Afaik we don't have one, and this is iso7 so I think we have some freedom here as long as we mention it in a change entry. From where I'm standing, I like this idea best among our options. What about you?

arnaudsjs: I can't find one either.

sanderr (changes requested):
Very minor change request, otherwise as discussed with regards to the error handling etc.

changelogs/unreleased/6025-implement-discovery-resource-handler.yml
```diff
@@ -1,6 +1,7 @@
 description: "Add handler (DiscoveryHandler) to discover unmanaged resources."
 issue-nr: 6025
 change-type: minor
-destination-branches: [master, iso6]
+destination-branches: [master]
 sections:
   feature: "{{description}}"
+  upgrade-note: "The execution of do_reload() and execute() now execute between `pre()` and `post()` have been called."
```
sanderr, line 7:
```suggestion
  upgrade-note: "The execution of a handler's do_reload() and execute() now execute between `pre()` and `post()` have been called."
```

sanderr (changes requested):

wouterdb (commented):
let me know when it is ready again

wouterdb (approved):

PR #6433, opened 2023-08-25
Drop support for the legacy relationship syntax

### LINKED ISSUE #5265: drop support for legacy relation syntax
Only for master. See https://github.com/inmanta/inmanta-core/blob/375eb5a4700123ffb5374bdda3a5d3f923025923/src/inmanta/parser/plyInmantaParser.py#L527.

arnaudsjs:
# Description

Drop support for the legacy relationship syntax from the compiler.

closes #5265

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/compiler/test_docs.py
+++ b/tests/compiler/test_docs.py
@@ -37,26 +37,6 @@ Each file needs to be associated with a host
     assert types["__config__::File"].get_attribute("host").comment.strip() == "Each file needs to be associated with a host"
 
 
-def test_doc_string_on_relation(snippetcompiler):
-    snippetcompiler.setup_for_snippet(
-        """
-entity File:
-end
-
-entity Host:
-end
-
-File file [1] -- [0:] Host host
-\"""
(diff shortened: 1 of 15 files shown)


FloLey (approved):

tests/compiler/test_docs.py
```diff
@@ -37,26 +37,6 @@ def test_doc_string_on_new_relation(snippetcompiler):
     assert types["__config__::File"].get_attribute("host").comment.strip() == "Each file needs to be associated with a host"
 
 
-def test_doc_string_on_relation(snippetcompiler):
```
sanderr, line 40: Isn't this test still relevant with the new syntax? Or do we have another test for it already?

sanderr (changes requested):

tests/compiler/test_docs.py
arnaudsjs: Yes. We should keep it.

Issue #6409, opened 2023-08-18
Improve language reference documentation


arnaudsjs:
The following information is missing in the language reference:

* Mention that tripple-quoted strings are supported.
* Mention that a tripple-quoted string has to be used to define a multi-line string.
* How does escaping work in a string.
* Mention how to append an element to a list.

PR #6457, opened 2023-09-05
Issue/6409 improve language reference documentation

Hugo-Inmanta:
# Description

Improve language reference documentation

closes https://github.com/inmanta/inmanta-core/issues/6409

# Self Check:

- [x] Changelog entry


arnaudsjs (changes requested):

docs/language.rst
```diff
@@ -177,6 +177,24 @@ Lists of primitive types are also primitive types: ``string[]``, ``number[]``, `
     # incorrect, can't assign to dict after construction
     # a["otherkey"] = "othervalue"
 
+Strings
```
arnaudsjs, line 180: Something that our documentation is clearly lacking, is a good overview about which different types of string that exist, what their syntax looks like and how they behave with respect to variable expansion and backslash escaping. At the moment this information is spread over the `Strings`, `string interpolation` and `string formatting` sections. I would like to have one section that gives the entire overview, replacing the existing sections. 

I think we have the following types of string:

* Regular string
* Regular multi-line string
* Raw string
* f-string

I would also explicitly mention that raw string and f-string cannot be used together. Because this is something that is possible in Python, but not in the Inmanta language.

arnaudsjs (changes requested):

docs/language.rst
```diff
@@ -177,6 +177,122 @@ Lists of primitive types are also primitive types: ``string[]``, ``number[]``, `
...
+.. code-block:: inmanta
+
+    regular_string_1 = "This is...\n...a basic string."
+
+    # Expected output e.g. when displayed:
+    # This is...
+    # ...a basic string.
+
+
+    regular_string_2 = 'This one too.'
```
Hugo-Inmanta, line 197: Rendering is weird here: some red boxes around the quotes and `This` is displayed in blue
image

docs/language.rst
```diff
@@ -177,6 +177,122 @@ Lists of primitive types are also primitive types: ``string[]``, ``number[]``, `
...
 
+Strings
++++++++
+
+
+There are four kinds of strings in the Inmanta language:
+
+- regular strings
+
+.. code-block:: inmanta
```
arnaudsjs, line 188: This code block introduces a lexing error in the documentation build:

```
language.rst:188: WARNING: Lexing literal_block 'regular_string_1 = "This is...\\n...a basic string."\n\n# Expected output e.g. when displayed:\n# This is...\n# ...a basic string.\n\n\nregular_string_2 = \'This one too.\'\n\n# Expected output e.g. when displayed:\n# This one too.' as "inmanta" resulted in an error at token: "'". Retrying in relaxed mode.
```

docs/language.rst
```diff
@@ -177,6 +177,122 @@ Lists of primitive types are also primitive types: ``string[]``, ``number[]``, `
...
+
+    regular_string_1 = "This is...\n...a basic string."
+
+    # Expected output e.g. when displayed:
```
arnaudsjs, line 192: Expected doesn't sound very confident.
```suggestion
    # Output when displayed:
```

docs/language.rst
```diff
@@ -177,6 +177,122 @@ Lists of primitive types are also primitive types: ``string[]``, ``number[]``, `
...
+    # This is...
+    # ...a basic string.
+
+
+    regular_string_2 = 'This one too.'
+
+    # Expected output e.g. when displayed:
```
arnaudsjs, line 199:
```suggestion
    # Output when displayed:
```

docs/language.rst
```diff
@@ -177,6 +177,122 @@ Lists of primitive types are also primitive types: ``string[]``, ``number[]``, `
...
+
+.. code-block:: inmanta
+
+    multi_line_string = """This
+    string
+    spans
+    multiple
+    lines"""
+
+    # Expected output e.g. when displayed:
```
arnaudsjs, line 214:
```suggestion
    # Output when displayed:
```

docs/language.rst
```diff
@@ -177,6 +177,122 @@ Lists of primitive types are also primitive types: ``string[]``, ``number[]``, `
...
+
+Raw strings are similar to python's raw strings in that they treat backslashes as regular characters.
+On the other hand, in regular and multi-line strings, escape characters (e.g. ``\n``, ``\t``...) are interpreted and
+therefore backslashes need to be escaped in order to be displayed.
+
+.. code-block:: inmanta
+
+    raw_string = r"This is...\n...a raw string."
+
+    # Expected output e.g. when displayed:
```
arnaudsjs, line 235:
```suggestion
    # Output when displayed:
```

docs/language.rst
```diff
@@ -177,6 +177,122 @@ Lists of primitive types are also primitive types: ``string[]``, ``number[]``, `
...
+
+.. note::
+    Unlike python's multi-line strings, only double quotes are supported to define a multi-line string i.e. ``"""`` is
+    valid, but ``'''`` is not.
+
+- raw strings
+
+Raw strings are similar to python's raw strings in that they treat backslashes as regular characters.
+On the other hand, in regular and multi-line strings, escape characters (e.g. ``\n``, ``\t``...) are interpreted and
+therefore backslashes need to be escaped in order to be displayed.
```
arnaudsjs, line 229: We should also mention here that no variable expansion is done in raw strings.

arnaudsjs (changes requested):

docs/language.rst
arnaudsjs: (Maybe this is related to the red boxes around the quotes)

Hugo-Inmanta: That's weird, this snippet compiles just fine. I noticed another case of a lexing error resulting in red boxes. But it looks like the red boxes only appear in my local dev version of the doc build:

Local dev docs:
image

**VS**

Stable [docs](https://docs.inmanta.com/inmanta-service-orchestrator/latest/lsm/validation_types/validation_types.html#validation-types):
image

The other error:
```sh
inmanta-core/docs/lsm/validation_types/validation_types.rst:67: WARNING: Lexing literal_block 'typedef <attr> as <type> matching self <cmp> <value> [(or | and) <comparison_2> ...]\ntypedef <attr> as <type> matching <value> <cmp> self [(or | and) <comparison_2> ...]' as "inmanta" resulted in an error at token: '|'. Retrying in relaxed mode.
```

docs/language.rst
```diff
@@ -177,6 +177,128 @@ Lists of primitive types are also primitive types: ``string[]``, ``number[]``, `
...
+    # Output when displayed:
+    # This
+    # string
+    # spans
+    # multiple
+    # lines
+
+.. note::
+    Unlike python's multi-line strings, only double quotes are supported to define a multi-line string i.e. ``"""`` is
+    valid, but ``'''`` is not.
```
Hugo-Inmanta, line 223: I was surprised by this

docs/language.rst
arnaudsjs: The lexing here is not done by the compiler itself but by Inmanta sphinx: https://github.com/inmanta/inmanta-sphinx/blob/master/sphinxcontrib/inmanta/pygments_lexer.py.

Hugo-Inmanta: It looks like the syntax highlighter chokes on `'` and `|`.  I found this (discussion)[https://github.com/sphinx-doc/sphinx/issues/4225] about it.
It looks like the prod build just ignores syntax highlighting in case of a lexing error, so even if it pains me to have one block without syntax highlighting here I think this is fine ?

arnaudsjs: This needs to be fixed, but we don't have to do that right now. You can also create a new ticket if it would take too long to fix.

Hugo-Inmanta: I'll open a PR on the lexer to fix this, I didn't know about it until now!

docs/language.rst
arnaudsjs: Yes. Me too. I created [a ticket](https://github.com/inmanta/inmanta-core/issues/6488) to support this. In the meantime we just document well that single quotes are not supported.

arnaudsjs (approved):

PR #6464, opened 2023-09-06
Issue/6349 - release_version

### LINKED ISSUE #6349: Ensure release version is not executed twice
Currently, release_version does not check if the version is already released. 

We should make sure this method gets executed only once.

This may have a few implications:
- it is correctly not in a transaction, so it may take some extra work to make it not run while it is running
- if we change the return code, we may affect the front-end

wouterdb:
# Description

Ensure release_version can only be called once

Moving this into a transaction scares me, I don't know exactly how to protect this from the cascading delete deadlock issue.

closes #6349 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5512,7 +5512,7 @@ class ConfigurationModel(BaseDocument):
         # get versions
         query = f"SELECT version FROM {cls.table_name()} WHERE environment=$1 AND released=true ORDER BY version DESC"
         values = [cls._get_value(environment)]
-        version_records = await cls._fetch_query(query, *values)
+        version_records = await cls._fetch_query(query, *values, connection=connection)
 
         versions = [record["version"] for record in version_records]
 
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5317,12 +5317,32 @@ class ConfigurationModel(BaseDocument):
         version: int,
         *,
         connection: Optional[asyncpg.connection.Connection] = None,
+        lock: Optional[RowLockMode] = None,
     ) -> Optional["ConfigurationModel"]:
         """
         Get a specific version
         """
-        result = await cls.get_one(environment=environment, version=version, connection=connection)
+        result = await cls.get_one(environment=environment, version=version, connection=connection, lock=lock)
         return result
+
+    @classmethod
+    async def get_version_internal(
+        cls,
+        environment: uuid.UUID,
+        version: int,
+        *,
+        connection: Optional[asyncpg.connection.Connection] = None,
+        lock: Optional[RowLockMode] = None,
+    ) -> Optional["ConfigurationModel"]:
+        """Return a version, but don't populate the status and done fields, which are expensive to construct"""
+        query = f"""SELECT *
+                          FROM {ConfigurationModel.table_name()}
+                          WHERE environment=$1 AND version=$2 {lock.value};
+                          """
+        result = await cls.select_query(query, [environment, version], connection=connection)
+        if not result:
+            return None
+        return result[0]
 
     @classmethod
     async def get_latest_version(
@@ -5343,7 +5363,11 @@ class ConfigurationModel(BaseDocument):
         return versions[0]
 
     @classmethod
-    async def get_version_nr_latest_version(cls, environment: uuid.UUID) -> Optional[int]:
+    async def get_version_nr_latest_version(
+        cls,
+        environment: uuid.UUID,
+        connection: Optional[Connection] = None,
+    ) -> Optional[int]:
         """
         Get the version number of the latest released version in the given environment.
         """
@@ -5353,7 +5377,7 @@ class ConfigurationModel(BaseDocument):
                     ORDER BY version DESC
                     LIMIT 1
                     """
-        result = await cls._fetchrow(query, cls._get_value(environment))
+        result = await cls._fetchrow(query, cls._get_value(environment), connection=connection)
         if not result:
             return None
         return int(result["version"])
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4660,7 +4660,7 @@ class Resource(BaseDocument):
 
     @classmethod
     async def get_resources_for_version_raw(
-        cls, environment: uuid.UUID, version: int, projection: Optional[List[str]], connection: Optional[Connection] = None
+        cls, environment: uuid.UUID, version: int, projection: Optional[List[str]], *, connection: Optional[Connection] = None
     ) -> List[Dict[str, Any]]:
         if not projection:
             projection = "*"
@@ -5484,7 +5484,7 @@ class ConfigurationModel(BaseDocument):
 
     @classmethod
     async def get_increment(
-        cls, environment: uuid.UUID, version: int, connection: Optional[Connection] = None
+        cls, environment: uuid.UUID, version: int, *, connection: Optional[Connection] = None
     ) -> tuple[set[m.ResourceIdStr], set[m.ResourceIdStr]]:
         """
         Find resources incremented by this version compared to deployment state transitions per resource
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5434,13 +5434,13 @@ class ConfigurationModel(BaseDocument):
                 self.environment,
             )
 
-    async def get_undeployable(self) -> List[m.ResourceIdStr]:
+    def get_undeployable(self) -> List[m.ResourceIdStr]:
         """
         Returns a list of resource ids (NOT resource version ids) of resources with an undeployable state
         """
         return self.undeployable
 
-    async def get_skipped_for_undeployable(self) -> List[m.ResourceIdStr]:
+    def get_skipped_for_undeployable(self) -> List[m.ResourceIdStr]:
         """
         Returns a list of resource ids (NOT resource version ids)
         of resources which should get a skipped_for_undeployable state
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4660,7 +4660,7 @@ class Resource(BaseDocument):
 
     @classmethod
     async def get_resources_for_version_raw(
-        cls, environment: uuid.UUID, version: int, projection: Optional[List[str]]
+        cls, environment: uuid.UUID, version: int, projection: Optional[List[str]], connection: Optional[Connection] = None
     ) -> List[Dict[str, Any]]:
         if not projection:
             projection = "*"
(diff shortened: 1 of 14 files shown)


src/inmanta/protocol/methods.py
```diff
@@ -535,6 +535,11 @@ def release_version(tid: uuid.UUID, id: int, push: bool = False, agent_trigger_m
     :param push: Notify all agents to deploy the version
     :param agent_trigger_method: Indicates whether the agents should perform a full or an incremental deploy when
                                 push is true.
+
+     :return: Returns the following status codes:
+            200: The version is released
+            404: The requested version does not exist
+            409: The requested version was already release
```
Hugo-Inmanta, line 542:
```suggestion
            409: The requested version was already released
```

Hugo-Inmanta (approved):
Looks good to me as far as I can tell but I don't have a clear view on all the consequences of moving `release_version` inside a transaction.

src/inmanta/deploy.py
```diff
@@ -208,16 +209,16 @@ def _create_environment(self, project_id: str, environment_name: str) -> Optiona
...
             return None
 
-        if "versions" in result.result and len(result.result["versions"]) > 0:
-            versions: List[int] = [x["version"] for x in result.result["versions"]]
-            sorted(versions)
-            return versions[0]
+        if result.result and "versions" in result.result and len(result.result["versions"]) > 0:
+            version = result.result["versions"][0]
+            assert isinstance(version, dict)  # mypy
+            return version
```
sanderr, line 221: You dropped the sorting but `list_versions` doesn't make any guarantees about the ordering (though perhaps it should).

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1032,87 +1032,96 @@ async def release_version(
...
-        increments: tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]] = await self.resource_service.get_increment(
-            env, version_id
-        )
+                if model.released:
+                    raise Conflict(f"The version {version_id} on environment {env} is already released.")
 
-        increment_ids, neg_increment = increments
-        await self.resource_service.mark_deployed(env, neg_increment, now, version_id)
+                # Already mark undeployable resources as deployed to create a better UX (change the version counters)
+                undep = await model.get_undeployable()
```
sanderr, line 1045: Another query on a separate connection

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1032,87 +1032,96 @@ async def release_version(
...
+
+                for agent in agents:
+                    client = self.agentmanager_service.get_agent_client(env.id, agent)
+                    if client is not None:
+                        if not agent_trigger_method:
+                            env_agent_trigger_method = await env.get(ENVIRONMENT_AGENT_TRIGGER_METHOD, connection=connection)
+                            incremental_deploy = env_agent_trigger_method == const.AgentTriggerMethod.push_incremental_deploy
+                        else:
+                            incremental_deploy = agent_trigger_method is const.AgentTriggerMethod.push_incremental_deploy
+                        self.add_background_task(client.trigger(env.id, agent, incremental_deploy))
```
sanderr, line 1118: I think we may need to first commit the transaction before we add any background tasks, e.g. dedent the whole `if push` block to live outside of the transaction.

src/inmanta/server/services/resourceservice.py
```diff
@@ -412,9 +413,12 @@ async def mark_deployed(
             send_events=False,
             keep_increment_cache=True,
             is_increment_notification=True,
+            connection=connection,
         )
 
-    async def get_increment(self, env: data.Environment, version: int) -> tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]]:
+    async def get_increment(
+        self, env: data.Environment, version: int, connection: Optional[Connection] = None
```
sanderr, line 420: This is a tricky one: we move it into a transaction but we keep a cache that lives outside it. I have to give this more thought.

sanderr (changes requested):
I too am scared of moving this in a transaction but I think we can give it a shot. A few notes:
1. I think calling `release_version` twice concurrently would still work (execute twice). Should we cover that scenario?
2. The ticket mentions "it is correctly not in a transaction", what is that about?
3. An alternative solution could be to rely on Postgres advisory locks as we do for `put_partial`. `put_partial` uses the connection the lock is acquired in for its queries, but we could just as well not wire it through, having it act purely as a semantic lock (nothing else in its transaction, just the Python context keeping the PG lock open until the context is exited). I'm not sure whether I prefer this over the transaction approach, depends a bit on the response to 2 and the semantics we want to uphold here.

src/inmanta/data/__init__.py
```diff
@@ -5509,7 +5520,7 @@ async def get_increment(cls, environment: uuid.UUID, version: int) -> tuple[set[
             # todo in next version
             next: list[abc.Mapping[str, object]] = []
 
-            vresources = await Resource.get_resources_for_version_raw(environment, version, projection)
+            vresources = await Resource.get_resources_for_version_raw(environment, version, projection, connection=connection)
```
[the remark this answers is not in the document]
wouterdb: thx!

src/inmanta/deploy.py
wouterdb: the sorting didn't do anything, as 'sorted' doesn't do an in place sort, but returns a sorted list, that was never assigned to anything.

src/inmanta/server/services/orchestrationservice.py
wouterdb: not a query strangely enough, I could drop the async keyword on them, if you think that would clarify it.

src/inmanta/server/services/orchestrationservice.py
wouterdb: it is outside the transaction (but on the same connection)

src/inmanta/server/services/resourceservice.py
wouterdb: A cache we can hopefully get rid off soon and that
1.  we should I believe always miss (and populate) (we do the release, so we are first)
2. I would hope the transaction will give us exclusivity on our own code path for our own environment


So 
1. we can never read it, and we can't get blocked on it. 
2. We should never have to abort the transaction, so we can't make it stale
3. Only after the transaction commits, will the version become available for pull, so only then the cache line will be read. (any stale read will invalidate the cache)

wouterdb:
1. For it to be correct not in a transaction, it needs to be idempotent and side-effect free. (i.e. if can only be considered released if it is fully release and a release after a partial release should also work).  We are adding more and more functionality in this, so I'm afraid it will inadvertently lose (or has lost) one of these properties. 

2 . I go from the assumption a second transaction will be blocked by the first one, but we may have to add some explicit locking to make this cleaner/more efficient?

src/inmanta/deploy.py
sanderr: Ah, I didn't notice the lack of assignment

src/inmanta/server/services/orchestrationservice.py
sanderr: [It is](https://github.com/inmanta/inmanta-core/blob/8f0620df3c673284fcd0a0211b6047900a23d623/src/inmanta/data/__init__.py#L4510), isn't it? Am I missing something?

src/inmanta/server/services/orchestrationservice.py
sanderr: Oh good, sometimes difficult to see.

One potential problem I still see (though it existed before this PR as well I think) is that if the caller passes in a collection object, wrapped in a transaction, we add these background tasks before that outer connection commiits.

sanderr:
2. Could you add a test case? If it is currently blocked, as you assume, then it's fine for me as is. But I expect it won't be, or it will deadlock.

src/inmanta/server/services/orchestrationservice.py
wouterdb: I think it is this one: https://github.com/inmanta/inmanta-core/blob/8f0620df3c673284fcd0a0211b6047900a23d623/src/inmanta/data/__init__.py#L5404

src/inmanta/server/services/orchestrationservice.py
wouterdb: Yes, I think that is part of the responsibility of the caller.

Passing along connections with open transactions is not something we can consider to be modular.

We could assert there is no ongoing transaction, but that is perhaps a bit harsh?

src/inmanta/server/services/orchestrationservice.py
sanderr: Wow, didn't even realize I was looking in the wrong class...

wouterdb: do I remove the async?

src/inmanta/server/services/orchestrationservice.py
wouterdb: added it

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1032,87 +1032,112 @@ async def release_version(
...
+                latest_version = await data.ConfigurationModel.get_version_nr_latest_version(env.id, connection=connection)
 
-        # Set the updated field:
-        await data.Resource.copy_last_success(env.id, version_id)
+                # ensure we are the latest version
+                # this is required for the subsequent increment calculation to make sense
+                # this does introduce a race condition, with any OTHER release running concurrently on this environment
+                # We could lock the get_version_nr_latest_version for update to prevent this
+                if model.version < (latest_version or -1):
+                    raise Conflict(
```
wouterdb, line 1053: This check didn't exist either, it is still a bit of a stink like this, but I'm a bit uncertain on how to test it if I would add another lock here.

src/inmanta/server/services/orchestrationservice.py
sanderr: I see no reason not to (unless we expose this somewhere but I don't think we do)

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1032,87 +1032,113 @@ async def release_version(
...
-                    finished=now,
-                    status=const.ResourceState.skipped_for_undefined,
-                    action=const.ResourceAction.deploy,
-                    changes={},
-                    messages=[],
-                    change=const.Change.nochange,
-                    send_events=False,
-                    connection=connection,
+        async with data.ConfigurationModel.get_connection(connection) as connection:
+            async with connection.transaction():
```
wouterdb, line 1036: Should I retry this?

src/inmanta/server/services/resourceservice.py
sanderr: I don't completely follow this: we invalidate the cache but also write it, persisting the transaction state before it commits (if it does).

wouterdb: if we roll back, 
 - the version number we released will never be pulled, so the cache is invalid for every intent and purpose
 - the release will happen again later, re-populating the cache

src/inmanta/server/services/orchestrationservice.py
wouterdb: yes. This one is safe wrt itself

src/inmanta/server/services/resourceservice.py
sanderr: Ok, I see, thanks.

sanderr (approved):

PR #6478, opened 2023-09-11
Drop equals injection in constraints of typedefs

### LINKED ISSUE #5787: drop Equals injection in DefineTypeConstraint
This used to be required but probably isn't anymore. inmanta/lsm#580 may be required so backwards compatibility should be taken into account.

```python
        if hasattr(expression, "arguments"):
            # some sort of function call
            expression = Equals(expression, Literal(True))
```

arnaudsjs:
# Description

Drop `== True` injection on constraints of a typedef.

closes #5787

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/5787-drop-equals-injection.yml
+++ b/changelogs/unreleased/5787-drop-equals-injection.yml
@@ -5,4 +5,4 @@ issue-repo: inmanta-core
 change-type: patch
 destination-branches: [master]
 sections:
-  deprecation-note: "{{description}}"
+  deprecation-note: "The compiler no longer explicitly injects the implied `== true` for plugin calls in typedef constraints"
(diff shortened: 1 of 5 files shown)


changelogs/unreleased/5787-drop-equals-injection.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: Drop `== True` injection on constraints of a typedef.
+issue-nr: 5787
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master]
+sections:
+  deprecation-note: "{{description}}"
```
sanderr, line 8: I think this message is a bit cryptic. Perhaps this?
```suggestion
  deprecation-note: "The compiler no longer explicitly injects the implied `== true` for plugin calls in typedef constraints"
```

sanderr (changes requested):

sanderr (approved):

Hugo-Inmanta (approved):

sanderr (approved):

PR #6495, opened 2023-09-13
Store files in the database

### LINKED ISSUE #6441: refactor file storage to use database instead of file system
One of the major complications for a high availability orchestrator is that files will be missing on failover. Putting them in the database would be a trivial way to solve that.

The goal of this ticket is to do this refactor for iso7, unless we foresee major complications for the iso6->iso7 migration as a result.

Document clearly that a compile should be triggered after the upgrade and that old files will not be migrated.

arnaudsjs:
# Description

Files handled by the file API are now stored in the database instead of on the file system of the Inmanta server.

closes #6441

# Self Check:

- [x] Changelog entry


src/inmanta/db/versions/v202309120.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Contact: code@inmanta.com
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    """
+    Add the indexes to make the joins done by a cascading delete perform well.
```
Hugo-Inmanta, line 24: I think this is not accurate ?

tests/db/migration_tests/test_v202308100_to_v202309120.py
```diff
@@ -0,0 +1,30 @@
...
+    Contact: code@inmanta.com
+"""
+import os
+from collections import abc
+
+import pytest
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps/v202308100.sql"))
+async def test_add_indexes_for_cascading_delete(
```
Hugo-Inmanta, line 25:
```suggestion
async def test_add_file_table(
```

Hugo-Inmanta (changes requested):
Minor changes requested, looks pretty cool otherwise :+1:

src/inmanta/server/services/fileservice.py
```diff
@@ -42,131 +42,86 @@ def __init__(self) -> None:
...
+
+        async def _get_lines_for_file(content_hash: str) -> list[str]:
+            if content_hash == "" or content_hash == "0":
+                return []
+            else:
+                file = await File.get_one(content_hash=content_hash)
+                if not file:
+                    raise NotFound()
+                file_content = file.content.decode(encoding="utf-8")
+                return _parse_like_readlines(file_content)
```
sanderr, line 121:
```suggestion
                # keepends for backwards compatibility with <file_handle>.readlines()
                return file_content.splitlines(keepends=True)
```
I believe this is equivalent, and less brittle in case of CR line breaks (they should be rare but I'm not sure if we can rule them out).

Or is there a difference I'm not seeing?

sanderr (changes requested):

src/inmanta/server/services/fileservice.py
```diff
@@ -42,131 +42,86 @@ def __init__(self) -> None:
...
-                raise ServerError(
-                    f"File corrupt, expected hash {file_hash} but found {actualhash}. "
-                    "Deleting file, please re-upload the corrupt file."
-                )
-            else:
-                LOGGER.error("File corrupt, expected hash %s but found %s at %s", file_hash, actualhash, file_name)
-                raise ServerError(
-                    f"File corrupt, expected hash {file_hash} but found {actualhash}, please contact the server administrator"
-                )
+        return file.content
```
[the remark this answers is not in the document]
arnaudsjs: When a file is written, we always check whether the hash and the content matches. Since we don't migrate any files on upgrade and the user has to do a full compile to populate the files again, there is no need for a hash check on the read path.

src/inmanta/server/services/fileservice.py
```diff
@@ -42,131 +42,86 @@ def __init__(self) -> None:
...
-        # keys are relevant.
-        response: dict[str, object] = {}
-
-        for f in files:
-            f_path = os.path.join(self.server_slice._server_storage["files"], f)
-            if not os.path.exists(f_path):
-                response[f] = None
-
-        return list(response.keys())
+        return list(await File.get_non_existing_files(files))
```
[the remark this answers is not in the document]
arnaudsjs: As far as I can see this only impacts the upload order.

src/inmanta/server/services/fileservice.py
arnaudsjs: Oh. I forgot this comment.

arnaudsjs: Done

Hugo-Inmanta (approved):

src/inmanta/server/services/fileservice.py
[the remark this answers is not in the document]
sanderr: Indeed, that's what I meant with "the normal operation flow". But the same goes for the previous normal operation flow and yet we had these checks here. But overall I agree that it serves little purpose.

sanderr (approved):

src/inmanta/server/services/fileservice.py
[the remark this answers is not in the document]
arnaudsjs: The difference with the previous normal flow is that we added the hash check after we know that corrupt files were present on the filesystem. That's why the check on the read path was required. The current implementation doesn't allow that anymore.

Issue #6428, opened 2023-08-23
add config option(s) for timezone aware datetimes on the API


sanderr:
The server currently serializes all timestamps to be timezone-naive. We should add an option to have it return aware timestamps instead. The option should be off by default on iso6. For iso7, we try to enable it by default and see what blows up. If the fallout is too big we disable it there as well.

Before merging, notify everyone and ask to report if things catch fire.

Afterwards create a follow-up ticket for iso8 to enable the option by default.

PR #6512, opened 2023-09-15
Issue/6428 add config option for timezone aware datetimes on the api iso6

Hugo-Inmanta:
# Description
Part of https://github.com/inmanta/inmanta-core/issues/6428

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/data/dataview.py
+++ b/src/inmanta/data/dataview.py
@@ -87,7 +87,11 @@ from inmanta.server.validate_filter import (
     LogLevelFilter,
 )
 from inmanta.types import JsonType, SimpleTypes
+<<<<<<< HEAD
 from inmanta.util import datetime_utc_isoformat
+=======
+from inmanta.util import datetime_iso_format
+>>>>>>> 851ca785 ([WIP] add options to make the server return timezone aware timestamps)
 
 T_ORDER = TypeVar("T_ORDER", bound=DatabaseOrderV2)
 T_DTO = TypeVar("T_DTO", bound=BaseModel)
@@ -390,8 +394,14 @@ class DataView(FilterValidator, Generic[T_ORDER, T_DTO], ABC):
             def value_to_string(value: Union[str, int, UUID, datetime]) -> str:
                 if isinstance(value, datetime):
                     # Accross API boundaries, all naive datetime instances are assumed UTC.
+<<<<<<< HEAD
                     # Returns ISO timestamp implicitly in UTC.
                     return datetime_utc_isoformat(value, naive_utc=True)
+=======
+                    # Returns ISO timestamp either implicitly in UTC (if server.tz_aware_timestamps is False)
+                    # or in the server's timezone (set by server.timezone).
+                    return datetime_iso_format(value, naive_utc=True)
+>>>>>>> 851ca785 ([WIP] add options to make the server return timezone aware timestamps)
                 return str(value)
 
             def make_link(**args: Optional[Union[str, int, UUID, datetime]]) -> str:
--- a/src/inmanta/server/agentmanager.py
+++ b/src/inmanta/server/agentmanager.py
@@ -792,7 +792,11 @@ class AgentManager(ServerSlice, SessionListener):
             **query,
         )
 
+<<<<<<< HEAD
         return 200, {"agents": [a.to_dict() for a in ags], "servertime": util.datetime_utc_isoformat(datetime.now())}
+=======
+        return 200, {"agents": [a.to_dict() for a in ags], "servertime": util.datetime_iso_format(datetime.now())}
+>>>>>>> 851ca785 ([WIP] add options to make the server return timezone aware timestamps)
 
     @handle(methods.get_state, env="tid")
     async def get_state(self, env: data.Environment, sid: uuid.UUID, agent: str) -> Apireturn:
--- a/src/inmanta/server/services/paramservice.py
+++ b/src/inmanta/server/services/paramservice.py
@@ -282,7 +282,11 @@ class ParameterService(protocol.ServerSlice):
                 "parameters": params,
                 "expire": self._fact_expire,
                 # Return datetime in UTC without explicit timezone offset
+<<<<<<< HEAD
                 "now": util.datetime_utc_isoformat(datetime.datetime.now()),
+=======
+                "now": util.datetime_iso_format(datetime.datetime.now()),
+>>>>>>> 851ca785 ([WIP] add options to make the server return timezone aware timestamps)
             },
         )
 
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -45,6 +45,7 @@ from tornado import gen
 
 from crontab import CronTab
 from inmanta import COMPILER_VERSION
+from inmanta.server.config import server_timezone, server_tz_aware_timestamps
 from inmanta.stable_api import stable_api
 from inmanta.types import JsonType, PrimitiveTypes, ReturnTypes
 
@@ -406,6 +407,12 @@ def datetime_utc_isoformat(timestamp: datetime.datetime, *, naive_utc: bool = Fa
         if timestamp.tzinfo is None and naive_utc
         else timestamp.astimezone(datetime.timezone.utc).replace(tzinfo=None)
     )
+
+    if server_tz_aware_timestamps:
+        return naive_utc_timestamp.astimezone(datetime.timezone(timedelta(hours=server_timezone))).isoformat(
+            timespec="microseconds"
+        )
+
     return naive_utc_timestamp.isoformat(timespec="microseconds")
 
 
--- a/src/inmanta/util.py
+++ b/src/inmanta/util.py
@@ -36,6 +36,7 @@ from asyncio import CancelledError, Future, Lock, Task, ensure_future, gather
 from collections import abc, defaultdict
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
+from datetime import timedelta
 from logging import Logger
 from types import TracebackType
 from typing import Awaitable, BinaryIO, Callable, Coroutine, Dict, Iterator, List, Optional, Set, Tuple, Type, TypeVar, Union
@@ -48,6 +49,10 @@ from inmanta import COMPILER_VERSION
 from inmanta.server.config import server_timezone, server_tz_aware_timestamps
 from inmanta.stable_api import stable_api
 from inmanta.types import JsonType, PrimitiveTypes, ReturnTypes
+<<<<<<< HEAD
+=======
+from inmanta.server.config import server_tz_aware_timestamps, server_timezone
+>>>>>>> 851ca785 ([WIP] add options to make the server return timezone aware timestamps)
 
 LOGGER = logging.getLogger(__name__)
 SALT_SIZE = 16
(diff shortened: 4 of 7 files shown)


src/inmanta/util.py
```diff
@@ -445,7 +449,7 @@ def api_boundary_json_encoder(o: object) -> Union[ReturnTypes, "JSONSerializable
     """
     if isinstance(o, datetime.datetime):
         # Accross API boundaries, all naive datetime instances are assumed UTC. Returns ISO timestamp implicitly in UTC.
-        return datetime_utc_isoformat(o, naive_utc=True)
+        return datetime_iso_format(o, naive_utc=True)
```
sanderr, line 452: The current implementation still always uses naive times for the json encoder. This should be configurable as well.

src/inmanta/server/services/paramservice.py
```diff
@@ -282,7 +283,7 @@ async def list_params(self, env: data.Environment, query: Dict[str, str]) -> Api
                 "parameters": params,
                 "expire": self._fact_expire,
                 # Return datetime in UTC without explicit timezone offset
-                "now": util.datetime_utc_isoformat(datetime.datetime.now()),
+                "now": util.datetime_iso_format(datetime.datetime.now(), in_local_tz=self._server_tz_aware_timestamps),
```
sanderr, line 286: I wonder why we convert it here in the first place. I would think that the json encoder employed by `RESTHandler` would take care of it. Could you see what happens when you use a plain `datetime` object here rather than a string?

src/inmanta/server/agentmanager.py
```diff
@@ -792,7 +792,10 @@ async def list_agents(
             **query,
         )
 
-        return 200, {"agents": [a.to_dict() for a in ags], "servertime": util.datetime_utc_isoformat(datetime.now())}
+        return 200, {
+            "agents": [a.to_dict() for a in ags],
+            "servertime": util.datetime_iso_format(datetime.now(), in_local_tz=opt.server_tz_aware_timestamps.get()),
```
sanderr, line 797: Same here about explicit serialization not being required.

sanderr (changes requested):

src/inmanta/util.py
```diff
@@ -394,18 +394,22 @@ def get_free_tcp_port() -> str:
...
+    be in local time.
+    :param in_local_tz: Whether to return aware timestamps in the local timezone or naive, implicit UTC timestamp.
     """
     naive_utc_timestamp: datetime.datetime = (
         timestamp
         if timestamp.tzinfo is None and naive_utc
         else timestamp.astimezone(datetime.timezone.utc).replace(tzinfo=None)
     )
+    if in_local_tz:
+        return naive_utc_timestamp.astimezone().isoformat(timespec="microseconds")
```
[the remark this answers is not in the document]
Hugo-Inmanta: Nice catch, I noticed it while working on the iso7 branch, I'll add a unit test

src/inmanta/server/services/paramservice.py
Hugo-Inmanta: Changed to a plain datetime object

Hugo-Inmanta:
Most of the test fixes involve using the `parse_timestamp` that was moved to `resourceservice` to `util`

tests/test_server.py
```diff
@@ -1266,7 +1266,25 @@ async def test_resource_deploy_done(server, client, environment, agent, caplog,
     assert resource_action["action"] == const.ResourceAction.deploy
     assert resource_action["started"] is not None
     assert resource_action["finished"] is not None
-    assert resource_action["messages"] == [
+
+    def check_messages(actual_result, expected_result):
```
sanderr, line 1270: If you do end up using `time-machine`, perhaps you can also get rid of this comparison approach by using the same timezone in the expected list.

changelogs/unreleased/6428-add-config-options-for-timezone-aware-datetimes-on-the-api-iso7.yml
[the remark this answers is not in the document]
Hugo-Inmanta: Oops yes when I tried to cherry pick the tests added on the iso7 branch

src/inmanta/util.py
```diff
@@ -394,19 +394,36 @@ def get_free_tcp_port() -> str:
...
+            return timestamp.isoformat(timespec="microseconds")
+        if naive_utc:
+            return timestamp.replace(tzinfo=datetime.timezone.utc).isoformat(timespec="microseconds")
+        return timestamp.astimezone().isoformat(timespec="microseconds")
+
+    if not timestamp.tzinfo:
+        if naive_utc:
+            return timestamp.replace(tzinfo=None).isoformat(timespec="microseconds")
+
+    return timestamp.astimezone(datetime.timezone.utc).replace(tzinfo=None).isoformat(timespec="microseconds")
```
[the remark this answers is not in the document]
Hugo-Inmanta: Great idea, it does improve readability by a lot. I pondered adding comments but it felt like bloating an already difficult to read method

tests/test_util.py
```diff
@@ -530,3 +530,65 @@ def worker():
...
+
+
+def test_datetime_iso_format():
+    """
+    Unit test for the util.datetime_iso_format function
+    """
+
+    # Check that when using naive timestamps as input only the output format changes (depending on the tz_aware param)
+
+    # Naive Daylight saving time date
```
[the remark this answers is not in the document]
Hugo-Inmanta: I tried using time-machine, it's pretty cool

tests/test_server.py
```diff
@@ -1266,22 +1289,34 @@ async def test_resource_deploy_done(server, client, environment, agent, caplog,
...
+    expected_timestamp: str
+    if opt.server_tz_aware_timestamps.get():
+        if endpoint_to_use == "resource_deploy_done":
+            expected_timestamp = now.astimezone().isoformat(timespec="microseconds")
+        else:
+            expected_timestamp = now.astimezone(timezone.utc).isoformat(timespec="microseconds")
+    else:
+        expected_timestamp = now.astimezone(timezone.utc).replace(tzinfo=None).isoformat()
+
+    parse_timestamp(expected_timestamp)
```
Hugo-Inmanta, line 1302: This feels a bit hackish

src/inmanta/util.py
```diff
@@ -404,21 +404,27 @@ def datetime_iso_format(timestamp: datetime.datetime, *, naive_utc: bool = False
     :param tz_aware: Whether to return timezone aware timestamps or naive, implicit UTC timestamp.
     """
 
-    if tz_aware:
-        if timestamp.tzinfo:
-            return timestamp.isoformat(timespec="microseconds")
-        if naive_utc:
-            return timestamp.replace(tzinfo=datetime.timezone.utc).isoformat(timespec="microseconds")
-        return timestamp.astimezone().isoformat(timespec="microseconds")
+    def convert_timestamp(timestamp: datetime.datetime, naive_utc: bool, tz_aware: bool) -> datetime.datetime:
```
sanderr, line 407: I would keep it simple and drop the parameters since they're exactly the same as the ones passed to the surrounding method.
```suggestion
    def convert_timestamp() -> datetime.datetime:
```

src/inmanta/protocol/common.py
```diff
@@ -959,15 +960,15 @@ def get_operation(self) -> str:
 
 
 # Util functions
-def custom_json_encoder(o: object) -> Union[ReturnTypes, util.JSONSerializable]:
+def custom_json_encoder(o: object, tz_aware: bool = False) -> Union[ReturnTypes, util.JSONSerializable]:
```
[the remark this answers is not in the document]
sanderr: I think I made this comment before I figured out what I wrote [here](https://github.com/inmanta/inmanta-core/pull/6598#discussion_r1355027429) and I forgot to update it. "if we change [the tz-aware option] the attribute hash will change". I think we should probably avoid that, i.e. revert the change I requested here. Sorry for the misdirection.

tests/test_server.py
sanderr: Agreed, I had hoped you could make it simpler but I'd overlooked the fact that the result actually contains the strings. Still, I prefer this over what you had before. I think the next step would be to validate the result with pydantic so you get an actual `datetime.datetime` object, and then you don't need to convert `expected_timestamp` to a specific timezone or to `str`. But that adds complexity in another place so I don't know if it's worth it. No strong preference from me.

sanderr (changes requested):

tests/test_server.py
```diff
@@ -1267,40 +1290,33 @@ async def test_resource_deploy_done(server, client, environment, agent, caplog,
...
-                if key == "timestamp":
-                    assert parse_timestamp(actual_value) == parse_timestamp(expected_value)
-                else:
-                    assert actual_value == expected_value
+    expected_timestamp: str
+    if opt.server_tz_aware_timestamps.get():
+        if endpoint_to_use == "resource_deploy_done":
+            expected_timestamp = now.astimezone().isoformat(timespec="microseconds")
+        else:
+            expected_timestamp = now.astimezone(timezone.utc).isoformat(timespec="microseconds")
```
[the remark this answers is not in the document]
Hugo-Inmanta: Right, this is something I'm still confused about. In the call to `result = await agent._client.resource_action_update(` (~20 lines above) it looks to me as if the timestamps are always transformed to naive utc, no matter what.
Before the call:
now == 2023-10-17 17:42:12.760139+02:00
Inside the call:
messages[0]["timestamp"] == '2023-10-17T15:42:33.923124'
I'm missing something here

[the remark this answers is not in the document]
sanderr: naive utc? The test case seems to assert for explicit UTC.

src/inmanta/util.py
```diff
@@ -404,7 +404,7 @@ def datetime_iso_format(timestamp: datetime.datetime, *, naive_utc: bool = False
     :param tz_aware: Whether to return timezone aware timestamps or naive, implicit UTC timestamp.
     """
 
-    def convert_timestamp(timestamp: datetime.datetime, naive_utc: bool, tz_aware: bool) -> datetime.datetime:
+    def convert_timestamp(timestamp: datetime.datetime) -> datetime.datetime:
```
sanderr, line 407: You can even drop `timestamp`

sanderr (changes requested):
Latest changes look good to me. Requesting changes because of the open discussion about a potential naive timestamp being returned in the test case.

Hugo-Inmanta:
I tried the suggestion to change `resource_action_update` from `method` to `typedmethod` but I couldn't properly type the return value (which is required I think?).
Instead I only changed the `messages` argument type from `list` to `List[model.LogLine]` which I think is enough ?

src/inmanta/data/dataview.py
```diff
@@ -391,7 +392,7 @@ def value_to_string(value: Union[str, int, UUID, datetime]) -> str:
                 if isinstance(value, datetime):
                     # Accross API boundaries, all naive datetime instances are assumed UTC.
                     # Returns ISO timestamp implicitly in UTC.
```
Hugo-Inmanta, line 394:
```suggestion
                    # Returns ISO timestamp.
```

Hugo-Inmanta:
1. all datetime objects are timezone aware -> https://github.com/inmanta/inmanta-core/blob/25805f6b3b22870796bfab88411983b479f20ee5/src/inmanta/data/model.py#L66
2. if we serialize datetime's internally (i.e. towards the DB) (which is rare), we add the timezone to the timestamp explicitly (and it is always UTC) -> https://github.com/inmanta/inmanta-core/blob/c829ede9c3ecd589aba4bd75a2e2adbb993b575d/src/inmanta/util.py#L454
3. if we serialize with the intent of passing the api boundary, we may or may not explicitly mention the timezone (controlled by the option) -> https://github.com/inmanta/inmanta-core/blob/c829ede9c3ecd589aba4bd75a2e2adbb993b575d/src/inmanta/util.py#L465
4. if we parse date-times and they have no explicit timezone, we assume UTC -> https://github.com/inmanta/inmanta-core/blob/c829ede9c3ecd589aba4bd75a2e2adbb993b575d/src/inmanta/util.py#L411

src/inmanta/util.py
```diff
@@ -394,19 +394,36 @@ def get_free_tcp_port() -> str:
         return str(port)
 
 
-def datetime_utc_isoformat(timestamp: datetime.datetime, *, naive_utc: bool = False) -> str:
+def datetime_iso_format(timestamp: datetime.datetime, *, tz_aware: bool = False) -> str:
     """
-    Returns a timestamp ISO string in implicit UTC.
+    Returns a timestamp ISO string. Naive timestamps are assumed to be UTC.
```
Hugo-Inmanta, line 399: comment for cross-reference

PR #6519, opened 2023-09-18
[FIX] fix changelog entry

Hugo-Inmanta:
# Description

I think change entry from stage 2 pr https://github.com/inmanta/inmanta-core/pull/6510/files is invalid.
Original PR for reference (https://github.com/inmanta/inmanta-core/pull/6486/files)

# Self Check:

- [ ] Changelog entry


sanderr (changes requested):
I don't think this is right. I think Wouter linked the wrong PR in the iso6 cherry-pick. Should be #6464 which seems to have the same change entry.

Hugo-Inmanta:
Oh I see now, closing this then.

PR #6530, opened 2023-09-20
Attempt to resolve deadlock

wouterdb:
# Description

Update all 'for update' locks to the more accurate 'for no key update'

# Self Check:

- [x] Changelog entry
- [] Sufficient test cases (reproduces the bug/tests the requested feature)


arnaudsjs (approved):

sanderr (approved):
Very curious what this will do to the nightly deadlocks.

PR #6542, opened 2023-09-25
switch to mono time

wouterdb:
# Description

I observed a negative timeout here, switched to mono time 

# Self Check:

- [x] Changelog entry


sanderr (approved):

Hugo-Inmanta (changes requested):

Hugo-Inmanta (approved):

arnaudsjs:
@wouterdb Don't we also have to update the `time.time()` in the `seen()` method?

PR #6556, opened 2023-09-29
Don't set undeployable resources in deploying state

arnaudsjs:
# Description

Fix bug where undeployable resources are briefly moved into the "deploying" state when their resource action executes. This fix also resolves a race condition in the `test_deploy_no_code` test case.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -211,6 +211,13 @@ class ResourceAction(ResourceActionBase):
             # Explicit cast is required because mypy has issues with * and generics
             results: List[ResourceActionResult] = cast(List[ResourceActionResult], await asyncio.gather(*waiters))
 
+            if self.undeployable:
+                self.status = self.undeployable
+                self.change = const.Change.nochange
+                self.changes = {}
+                self.future.set_result(ResourceActionResult(cancel=False))
+                return
+
             async with self.scheduler.ratelimiter:
                 ctx = handler.HandlerContext(self.resource, logger=self.logger)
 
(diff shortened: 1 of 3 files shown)


Hugo-Inmanta (approved):

src/inmanta/agent/agent.py
```diff
@@ -211,6 +211,13 @@ async def execute(
             # Explicit cast is required because mypy has issues with * and generics
             results: List[ResourceActionResult] = cast(List[ResourceActionResult], await asyncio.gather(*waiters))
 
+            if self.undeployable:
+                self.status = self.undeployable
+                self.change = const.Change.nochange
+                self.changes = {}
+                self.future.set_result(ResourceActionResult(cancel=False))
+                return
```
sanderr, line 219: Don't we need to briefly set `self. running` as well to avoid the race condition? Or is there no risk of that because we have no `await` here, therefore no context switching and we're only using a single thread?

sanderr (changes requested):
Could you confirm that there is no need to call `resource_deploy_done` because we never called the corresponding start method?

src/inmanta/agent/agent.py
arnaudsjs: This is a bug. I implemented the fix.

sanderr (approved):

arnaudsjs:
> Could you confirm that there is no need to call resource_deploy_done because we never called the corresponding start method?

There is indeed an issue with not calling `resource_deploy_done` API endpoint. If we don't call it, the `mark_done_if_done()` method is not called. That way the model version will never go to deployed when all the resources are undeployable. This should be fixed now.

sanderr (approved):

PR #6565, opened 2023-10-02
Fix race in test for incremental deploys

arnaudsjs:
# Description

This PR fixes a race condition in the `test_s_incremental_deploy_interrupts_full_deploy` test case. The old wait condition `wait_for_done_with_waiters(client, environment, version2)` returned as soon as all resources in version2 were deployed successfully. But in fact we should wait for two deployments in version2, one incremental deploy and one repair run that v2 interrupted.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/fix-race-in-incremental-deploy-test.yml
+++ b/changelogs/unreleased/fix-race-in-incremental-deploy-test.yml
@@ -1,0 +1,4 @@
+---
+description: "Fix race condition in the `test_s_incremental_deploy_interrupts_full_deploy` test case where it could happen that a test::Wait resource was not notified."
+change-type: patch
+destination-branches: [master, iso6]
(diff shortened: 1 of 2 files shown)


Hugo-Inmanta (changes requested):
I was looking around a bit and this L2559 looks weird:
https://github.com/inmanta/inmanta-core/blob/e9110e68e9bd1e83e83fb2a8cc772793b236ff08/tests/agent_server/test_server_agent.py#L2559-L2561

looks good otherwise !

sanderr (approved):

arnaudsjs:
> I was looking around a bit and this L2559 looks weird:
> 
> https://github.com/inmanta/inmanta-core/blob/e9110e68e9bd1e83e83fb2a8cc772793b236ff08/tests/agent_server/test_server_agent.py#L2559-L2561
> 
> looks good otherwise !

Nice catch!

Hugo-Inmanta (approved):

Issue #5222, opened 2022-12-08
Improve paging and filtering performance of nested queries


sanderr:
# Description

This ticket is known to apply to the ResourceView query (see #5212). It may apply to others as well. Depending on the scope, the fix may be tailored to this specific query or it may have to be a generic query builder improvement. The rest of this ticket description will describe the issue for this query specifically. The first task is then to investigate whether other queries suffer from the same or a similar issue.

In short, the issue is that we do paging with the equivalent of an `OFFSET`. We don't use it directly, but because the filtering (e.g. for the paging) is only applied on the outer layer of the query, we do fetch a bunch of records that we drop for the final result.

# Tasks
- [ ] Investigate whether other queries suffer from the same issue
- [ ] Make the inner query filtering-aware, either generically if it affects other queries, or specifically for this query if it's a unique issue.
- [ ] If feasible, add some kind of test that could catch a big performance regression. A seemingly minor or semantically identical change could have huge performance consequences, so it would be nice if we had a test that just verified the order of magnitude on this query.

# In detail

The query is nested in the sense that it uses a `WITH` clause to declare a subquery. Concretely, it uses a recursive subquery that adds a single resource in each iteration. This recursive query is not aware of the paging, therefore it starts with the very first (according to `ORDER BY`) resource. The paging that is then applied on the outer query drops all irrelevant leading rows. The resulting query generated by the query builder looks roughly as follows:
```
WITH RECURSIVE cte AS (
    ( {subquery_latest_version_for_single_resource(higher_than=None)} )
    UNION ALL
    SELECT next_r.*
    FROM cte curr_r
    CROSS JOIN LATERAL (
        {subquery_latest_version_for_single_resource(higher_than="curr_r.resource_id")}
    ) next_r
)
SELECT *
FROM cte r
WHERE r.resource_id BETWEEN lower AND upper;
```

Simply making the inner query aware of the paging by replacing the `None` on line 2 with `lower`, greatly increases paging performance. The same applies for any custom filtering: filtering at the source prevents us fetching rows we will drop anyway.

While this is a simple change to the query, it is not necessarily simple to apply it through the query builder. The purpose of this ticket is to figure out a way to do so.

## PoC
[...]

PR #6566, opened 2023-10-02
Recursive query improvement

sanderr:


sanderr:
See https://github.com/inmanta/inmanta-core/issues/5222#issuecomment-1743210238

PR #6572, opened 2023-10-03
Add validation_type method from std module

### LINKED ISSUE #6540: Move type validation in core
Pydantic v2 dropped support for constr of with a regex constraint. This has been replaced by pattern. However, pattern is a linear time rust implementation of regexes. It does not support the same regexes as the python flavor we have been using.

The goal is that both in std and inmanta-lsm we rewrite constr(regex=...) to using a special regex alternative. This alternative is explained in https://docs.pydantic.dev/latest/migration/#patterns-regex-on-strings

The goal of this ticket is to:

- [x] Add a method to core that can be used by both std and inmanta-lsm to validate a type, i.e. move most of `inmanta_plugins.std.validate_type` into this new method. Implement this method in iso6 and iso7 both.
- [x] update std to use this new method.

Once this is done, inmanta-lsm can be updated to use this (not part of this ticket). This needs to be in place, before we can merge pydantic v2 support.

In solutions the following MR also depend on it: https://code.inmanta.com/solutions/modules/config/-/merge_requests/5

arnaudsjs:
# Description

This ticket moves the `validation_type()` method from the std module to inmanta-core.

closes #6540 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/move-validate-type-method-to-core.yml
+++ b/changelogs/unreleased/move-validate-type-method-to-core.yml
@@ -2,7 +2,7 @@
 description: Moved the `validate_type` logic from the std module to inmanta-core.
 issue-nr: 6540
 issue-repo: inmanta-core
-change-type: patch
+change-type: minor
 destination-branches: [master, iso6]
 sections:
   minor-improvement: "{{description}}"
--- a/changelogs/unreleased/move-validate-type-method-to-core.yml
+++ b/changelogs/unreleased/move-validate-type-method-to-core.yml
@@ -1,0 +1,8 @@
+---
+description: Moved the `validate_type` logic from the std module to inmanta-core.
+issue-nr: 6540
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master, iso6]
+sections:
+  minor-improvement: "{{description}}"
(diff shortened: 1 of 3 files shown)


src/inmanta/validation_type.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import Optional
+
+import pydantic
+
+from inmanta.stable_api import stable_api
+from inmanta.types import PrimitiveTypes
+
+
+@stable_api
+def validate_type(fq_type_name: str, value: PrimitiveTypes, validation_parameters: Optional[dict[str, object]] = None) -> None:
```
sanderr, line 28:
```suggestion
def validate_type(fq_type_name: str, value: PrimitiveTypes, validation_parameters: Optional[abc.Mapping[str, object]] = None) -> None:
```

sanderr (changes requested):
One tiny remark

sanderr (approved):

changelogs/unreleased/move-validate-type-method-to-core.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: Moved the `validate_type` logic from the std module to inmanta-core.
+issue-nr: 6540
+issue-repo: inmanta-core
+change-type: minor
+destination-branches: [master]
```
arnaudsjs, line 6: Once this is merged in, I will create a new PR for the ISO6 branch that also bump the `COMPILER_VERSION` on the iso6 branch.

bartv (approved):

PR #6573, opened 2023-10-03
Fix too eager overwrite of state by stale deploys

### LINKED ISSUE #6563: Fix for state overwriting by #6486
# Problem  

The fix introduced in #6486  on forward propagation of states is incorrect.

As it now, if we get a stale skip or fail, we propagate it to the latest released version of this resource, to make sure we overwrite the value that may have been copied in by the increment calculation.However, I can think of two situation this is not entirely correct:

-    the hash of the resource has changed (i.e. the failure is about a different desired state)
-    the state of the resource in the newest version is available (no state was copied over)  (this implies the first)

# Solution

When propagating forward, don't overwrite available

wouterdb:
# Description

Fix too eager overwrite of state by stale deploys

closes  #6563 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/6475-prevent-too-eager-overwrite.yml
+++ b/changelogs/unreleased/6475-prevent-too-eager-overwrite.yml
@@ -1,0 +1,4 @@
+description: Fix for state overwriting introduced by #6486
+issue-nr: 6563
+change-type: patch
+destination-branches: [master, iso6]
--- a/changelogs/unreleased/6475-prevent-too-eager-overwrite.yml
+++ b/changelogs/unreleased/6475-prevent-too-eager-overwrite.yml
@@ -1,4 +1,4 @@
-description: Fix for state overwriting introduced by #6486
+description: "Fix for state overwriting introduced by #6486"
 issue-nr: 6563
 change-type: patch
 destination-branches: [master, iso6]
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/6475-prevent-too-eager-overwrite.yml
```diff
@@ -0,0 +1,4 @@
+description: Fix for state overwriting introduced by #6486
```
sanderr, line 1: I think the ticket reference might be a comment in yaml, not sure.

sanderr (approved):

changelogs/unreleased/6475-prevent-too-eager-overwrite.yml
wouterdb: ooh, good catch

src/inmanta/server/services/resourceservice.py
```diff
@@ -426,6 +426,7 @@ async def _update_deploy_state(
         message: str,
         fail_on_error: bool,
         connection: Optional[Connection] = None,
+        can_overwrite_available: bool = True,
```
arnaudsjs, line 429: Just a question: Why did we parameterize this method. As far as I can see we never use `can_overwrite_available: bool = True` and I don't think we have the intention of ever using it like that.

arnaudsjs (approved):

src/inmanta/server/services/resourceservice.py
wouterdb: True, I kept it at one hand to get this out quickly, at the other hand I use it as documentation for the unexpected behavior that it will not always update the state.

wouterdb:
follow up in https://github.com/inmanta/inmanta-core/pull/6577

Issue #6549, opened 2023-09-27
fix issues with deploy/repair interval cron expressions


sanderr:
#6374 introduced the config option to use a cron expression for the repair interval. There are two things missing here:

- [ ] there should be an environment setting for it
- [ ] the same should be done for the deploy interval

Hugo-Inmanta:
This PR is missing a migration script

PR #6574, opened 2023-10-03
Issue/6549 fix issues with deploy or repair interval cron expressions

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/6549

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/config.py
+++ b/src/inmanta/agent/config.py
@@ -97,7 +97,9 @@ agent_deploy_interval = Option(
     0,
     "Either the number of seconds between two (incremental) deployment runs of the agent or a cron-like expression."
     " If a cron-like expression is specified, a deploy will be run following a cron-like time-to-run specification,"
-    " interpreted in UTC (e.g. `min hour dom month dow`). A deploy will be requested at the scheduled time. Note that if a cron"
+    " interpreted in UTC. The expected format is `[sec] min hour dom month dow [year]` ( If only 6 values are provided, they"
+    " are interpreted as `min hour dom month dow year`)."
+    " A deploy will be requested at the scheduled time. Note that if a cron"
     " expression is used the 'agent_deploy_splay_time' setting will be ignored."
     " Set this to 0 to disable the scheduled deploy runs.",
     is_time_or_cron,
@@ -122,7 +124,9 @@ agent_repair_interval = Option(
     600,
     "Either the number of seconds between two repair runs (full deploy) of the agent or a cron-like expression."
     " If a cron-like expression is specified, a repair will be run following a cron-like time-to-run specification,"
-    " interpreted in UTC (e.g. `min hour dom month dow`). A repair will be requested at the scheduled time. Note that if a cron"
+    " interpreted in UTC. The expected format is `[sec] min hour dom month dow [year]` ( If only 6 values are provided, they"
+    " are interpreted as `min hour dom month dow year`)."
+    " A repair will be requested at the scheduled time. Note that if a cron"
     " expression is used the 'agent_repair_splay_time' setting will be ignored."
     " Setting this to 0 to disable the scheduled repair runs.",
     is_time_or_cron,
--- a/src/inmanta/agent/config.py
+++ b/src/inmanta/agent/config.py
@@ -95,8 +95,12 @@ agent_deploy_interval = Option(
     "config",
     "agent-deploy-interval",
     0,
-    "The number of seconds between two (incremental) deployment runs of the agent. Set this to 0 to disable the scheduled deploy runs.",
-    is_time,
+    "Either the number of seconds between two (incremental) deployment runs of the agent or a cron-like expression."
+    " If a cron-like expression is specified, a deploy will be run following a cron-like time-to-run specification,"
+    " interpreted in UTC (e.g. `min hour dom month dow`). A deploy will be requested at the scheduled time. Note that if a cron"
+    " expression is used the 'agent_deploy_splay_time' setting will be ignored."
+    " Set this to 0 to disable the scheduled deploy runs.",
+    is_time_or_cron,
     predecessor_option=agent_interval,
 )
 agent_deploy_splay_time = Option(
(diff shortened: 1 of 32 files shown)


src/inmanta/data/__init__.py
```diff
@@ -2377,6 +2377,12 @@ def convert_agent_trigger_method(value: object) -> str:
     return value
 
 
+def validate_cron_or_int(value: Union[int, str]) -> str:
+    if isinstance(value, int):
```
arnaudsjs, line 2381: The type defined in the definition of the setting is `str`. This will also be mentioned like this on our documentation pages. But if the user sends an integer value as a string to the API, this condition will not match and the validation will fail.

tests/agent_server/test_server_agent.py
```diff
@@ -303,8 +311,20 @@ async def test_server_restart(
     assert not resource_container.Provider.isset("agent1", "key3")
 
 
+@pytest.mark.parametrize(
+    "agent_deploy_interval",
+    ["2", "*/2 * * * * * *"],
```
arnaudsjs, line 316: According to the documentation the format should be `min hour dom month dow`, but this expression contains 7 elements instead of 5?

tests/agent_server/test_server_agent.py
```diff
@@ -387,20 +407,16 @@ async def test_spontaneous_deploy(
 
 
 @pytest.mark.parametrize(
-    "cron",
-    [False, True],
+    "agent_repair_interval",
+    ["2", "*/2 * * * * * *"],
```
arnaudsjs, line 411: Same here.

tests/agent_server/test_server_agent.py
```diff
@@ -485,6 +501,53 @@ async def verify_deployment_result():
...
 
+async def test_env_setting_wiring_to_autostarted_agent(
+    resource_container, environment, client, clienthelper, no_agent_backoff, async_finalizer, server
+):
+    """
+    Test that the AUTOSTART_AGENT_DEPLOY_INTERVAL and AUTOSTART_AGENT_REPAIR_INTERVAL
+    env settings are properly wired through to auto-started agents.
+    """
+
+    cron_expression = "*/2 * * * * * *"
```
arnaudsjs, line 512: Long cron expression?

tests/agent_server/test_server_agent.py
```diff
@@ -485,6 +501,53 @@ async def verify_deployment_result():
     await resource_action_consistency_check()
 
 
+async def test_env_setting_wiring_to_autostarted_agent(
```
arnaudsjs, line 504: Why does this test case set the `AUTOSTARTED_AGENT_MAP` and why does it start an agent? This is not required to test the behavior of `_make_agent_config()`, is it?

tests/agent_server/test_server_agent.py
```diff
@@ -485,6 +501,53 @@ async def verify_deployment_result():
...
+
+    cron_expression = "*/2 * * * * * *"
+    env_id = UUID(environment)
+
+    result = await client.set_setting(environment, AUTOSTART_AGENT_DEPLOY_INTERVAL, cron_expression)
+    assert result.code == 200
+    result = await client.set_setting(environment, AUTOSTART_AGENT_DEPLOY_SPLAY_TIME, 2)
+    assert result.code == 200
+    result = await client.set_setting(environment, AUTOSTART_AGENT_REPAIR_INTERVAL, cron_expression)
+    assert result.code == 200
```
arnaudsjs, line 520: This PR changes the type of the `AUTOSTART_AGENT_DEPLOY_INTERVAL` and `AUTOSTART_AGENT_REPAIR_INTERVAL` settings from int to string. But it's important to remain backwards compatible. As such, we should also test whether the following still works:

Backwards compatibility check with integer values:

* `await client.set_setting(environment, AUTOSTART_AGENT_DEPLOY_INTERVAL, 2)`  -> 
* `result = await client.set_setting(environment, AUTOSTART_AGENT_REPAIR_INTERVAL, 2)`

Test the new datatype (string):

* `await client.set_setting(environment, AUTOSTART_AGENT_DEPLOY_INTERVAL, "2")`  -> 
* `result = await client.set_setting(environment, AUTOSTART_AGENT_REPAIR_INTERVAL, "2")`

arnaudsjs (changes requested):

tests/agent_server/test_server_agent.py
Hugo-Inmanta: Updated the doc according to https://pypi.org/project/crontab/

tests/agent_server/test_server_agent.py
Hugo-Inmanta: Updated test case to check that int, string representation of int and cron expression pass validation, and that an empty string fails validation

src/inmanta/agent/config.py
```diff
@@ -95,8 +95,14 @@
     "config",
     "agent-deploy-interval",
     0,
-    "The number of seconds between two (incremental) deployment runs of the agent. Set this to 0 to disable the scheduled deploy runs.",
-    is_time,
+    "Either the number of seconds between two (incremental) deployment runs of the agent or a cron-like expression."
+    " If a cron-like expression is specified, a deploy will be run following a cron-like time-to-run specification,"
+    " interpreted in UTC. The expected format is `[sec] min hour dom month dow [year]` ( If only 6 values are provided, they"
+    " are interpreted as `min hour dom month dow year`)."
```
arnaudsjs, line 101: Nice! Very clear explanation.

src/inmanta/data/__init__.py
```diff
@@ -2377,9 +2381,23 @@ def convert_agent_trigger_method(value: object) -> str:
     return value
 
 
-def validate_cron(value: str) -> str:
+def validate_cron_or_int(value: Union[int, str]) -> str:
+    if isinstance(value, int):
+        return str(value)
```
arnaudsjs, line 2386: This if-block is redundant, right?

arnaudsjs (changes requested):

arnaudsjs (approved):

arnaudsjs (approved):

Hugo-Inmanta:
merged into iso6 with https://github.com/inmanta/inmanta-core/pull/6591

PR #6580, opened 2023-10-04
Split deletes into small transactions

### LINKED ISSUE #6427: split environment delete in small transactions
Main motivation: decrease lock contention. Deletes can be performed top-up, without resorting to cascading delete, outside of a transaction so that only one lock is acquired per "transaction". For more details, see [the slides of the planning meeting](https://docs.google.com/presentation/d/1-XcL0-DYmWVHBoacLBLu9UrG8Zs5EHq5npBqFZO9aFM/edit#slide=id.g25dbdc8f3b5_0_0).

- code already exists in the clear environment path that doesn't rely on cascade deletes as much
- this code should be updated to rely on cascading deletes even less, especially when it comes to the resource action - resource join table
- the environment delete should use this code path
- afterwards, check deadlock situation and create follow-up tickets according to planning meeting slides

arnaudsjs:
# Description

Split deletes of projects, environments and configurationmodels into small transactions to prevent deadlocks.

Part of #6427 

# Self Check:

- [x] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -5498,14 +5500,20 @@ async def get_versions(cls, environment: uuid.UUID, start: int = 0, limit: int =
...
     async def delete_cascade(self) -> None:
-        async with self.get_connection() as con:
-            async with con.transaction():
-                # Delete all code associated with this version
-                await Code.delete_all(connection=con, environment=self.environment, version=self.version)
-
-                # Delete ConfigurationModel and cascade delete on connected tables
-                await self.delete(connection=con)
+        await Compile.delete_all(environment=self.environment, version=self.version)  # Triggers cascading delete report table
+        await Code.delete_all(environment=self.environment, version=self.version)
```
Hugo-Inmanta, line 5504: I wonder why don't we have a foreign key reference for `Compile` and `Code` to `ConfigurationModel.version` the same way we have it for `UnknownParameter`  or `DryRun` ? i.e.
```sql
ADD CONSTRAINT unknownparameter_environment_version_fkey FOREIGN KEY (environment, version) REFERENCES public.configurationmodel(environment, version) ON DELETE CASCADE;
```

src/inmanta/data/__init__.py
```diff
@@ -5498,14 +5500,20 @@ async def get_versions(cls, environment: uuid.UUID, start: int = 0, limit: int =
...
+        await self._execute_query(
+            "DELETE FROM public.resourceaction_resource WHERE environment=$1 AND resource_version=$2",
+            self.environment,
+            self.version,
+        )
+        await ResourceAction.delete_all(environment=self.environment, version=self.version)
+        await Resource.delete_all(environment=self.environment, model=self.version)
+        await self.delete()
 
+        async with self.get_connection() as con:
```
Hugo-Inmanta, line 5516: Could we have a nullable foreign key in `Parameter` referencing `Resource.resource_id` and rely on a cascading delete of Resource instead of this ?

Hugo-Inmanta (changes requested):
Looks good as far as I understand it, a couple of (possibly dumb) questions that I'm curious about

src/inmanta/data/__init__.py
arnaudsjs: I was wondering the same. That's why I created the following ticket yesterday: https://github.com/inmanta/inmanta-core/issues/6581

arnaudsjs: I didn't implement it right away, because adding this migration script might be tricky. The database might be in an inconsistent state, because the foreign key constraint was not enforced. I didn't want to pollute this PR with that code and I didn't want to wait merging this PR, because this PR is a blocker for the release. Fixing the foreign key constraint isn't.

src/inmanta/data/__init__.py
arnaudsjs: No, we can't. A foreign key always references one specific record in a certain table. The primary key of the Resource table is (environment, model, resource_id). The same resource_id might occur in multiple version of the configurationmodel and as such there will be multiple records in the resource table for this resource_id. So adding a foreign key is not possible.

Hugo-Inmanta: I see, thanks :+1:

Hugo-Inmanta (approved):

sanderr (changes requested):
- Do I understand correctly that we now use a separate connection for each of these queries? Given that there are so many of them, I think we should use a single connection and wire it through, just don't wrap it in a transaction.

src/inmanta/data/__init__.py
```diff
@@ -2813,26 +2819,22 @@ async def unset(self, key: str) -> None:
...
-                await model.delete_cascade()
-
-            await Parameter.delete_all(environment=self.id)
-            await Resource.delete_all(environment=self.id)
-            await ResourceAction.delete_all(environment=self.id)
-            await Notification.delete_all(environment=self.id)
-        else:
-            # Cascade is done by PostgreSQL
+        await Agent.delete_all(environment=self.id)
+        await AgentProcess.delete_all(environment=self.id)  # This triggers a cascading delete on the agentinstance table
```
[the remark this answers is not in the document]
arnaudsjs: No. But this might be a leftover from the MongoDB implementation.

arnaudsjs:
> * Do I understand correctly that we now use a separate connection for each of these queries? Given that there are so many of them, I think we should use a single connection and wire it through, just don't wrap it in a transaction.

Right. I thought about doing that but somehow I forgot. Thanks!

sanderr (approved):

PR #6596, opened 2023-10-09
Issue/6549 add missing migration script

Hugo-Inmanta:
# Description

Previous PR was missing a migration script

closes https://github.com/inmanta/inmanta-core/issues/6549

# Self Check:

- [ ] Changelog entry


tests/db/migration_tests/test_v202310040_to_v202310090_.py
```diff
@@ -0,0 +1,34 @@
...
+import pytest
+
+from inmanta.data import Environment
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps/v202310040.sql"))
+async def test_type_change(
+    migrate_db_from: abc.Callable[[], abc.Awaitable[None]], get_tables_in_db: abc.Callable[[], abc.Awaitable[list[str]]]
+) -> None:
+    await migrate_db_from()
```
Hugo-Inmanta, line 30: I would have liked to assert that the type is `int` before the migration but I didn't manage to

src/inmanta/db/versions/v202310090.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+
+
+async def update(connection: Connection) -> None:
+    """
+    Update the type of the autostart_agent_repair_interval and autostart_agent_deploy_interval from int to str
+    """
+    async with (connection.transaction()):
+        await connection.execute(convert_setting_to_str("autostart_agent_repair_interval"))
+        await connection.execute(convert_setting_to_str("autostart_agent_deploy_interval"))
```
arnaudsjs, line 35: There is no need to start a transaction. The tool that executes the database migration scripts already starts a transaction to make sure the update can be rolled back if it raises an exception.

```suggestion
    await connection.execute(convert_setting_to_str("autostart_agent_repair_interval"))
    await connection.execute(convert_setting_to_str("autostart_agent_deploy_interval"))
```

src/inmanta/db/versions/v202310090.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+from asyncpg import Connection
+
+
+def convert_setting_to_str(setting: str) -> str:
+    return f"""
+    UPDATE environment
+    SET settings=jsonb_set(settings, '{{{setting}}}'::TEXT[], to_jsonb(settings->>'{setting}'), FALSE)
+    WHERE settings ? '{setting}'
+    """
```
arnaudsjs, line 26: Let's add a comment here saying that the `->>` operator returns something of type TEXT, so that it's clear where in the query the type conversion actually happens.

tests/db/migration_tests/test_v202310040_to_v202310090_.py
```diff
@@ -0,0 +1,34 @@
...
+from collections import abc
+
+import pytest
+
+from inmanta.data import Environment
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), "dumps/v202310040.sql"))
+async def test_type_change(
+    migrate_db_from: abc.Callable[[], abc.Awaitable[None]], get_tables_in_db: abc.Callable[[], abc.Awaitable[list[str]]]
```
arnaudsjs, line 28:
```suggestion
    migrate_db_from: abc.Callable[[], abc.Awaitable[None]]
```

tests/db/migration_tests/test_v202310040_to_v202310090_.py
arnaudsjs: I would expect you just need to same code as after the `migrate_from_db()` call but with `str` replaced by `int`?

arnaudsjs (changes requested):

tests/db/migration_tests/test_v202310040_to_v202310090_.py
Hugo-Inmanta: The old way of checking was raising an exception: calling this before the `migrate_from_db()` call would fail because of db pool exhaustion 
```py
env = await Environment.get_one(name="dev-1")
assert isinstance(env.settings["autostart_agent_deploy_interval"], int)
assert isinstance(env.settings["autostart_agent_repair_interval"], int)
```

src/inmanta/db/versions/v202310090.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+async def update(connection: Connection) -> None:
+    """
+    Update the type of the autostart_agent_repair_interval and autostart_agent_deploy_interval from int to str
+    """
+    await connection.execute(convert_setting_to_str("autostart_agent_repair_interval"))
+    await connection.execute(convert_setting_to_str("autostart_agent_deploy_interval"))
```
arnaudsjs, line 35:
```suggestion
    await connection.execute(convert_setting_to_str(data.AUTOSTART_AGENT_REPAIR_INTERVAL))
    await connection.execute(convert_setting_to_str(data.AUTOSTART_AGENT_DEPLOY_INTERVAL))
```

tests/db/migration_tests/test_v202310040_to_v202310090_.py
```diff
@@ -0,0 +1,48 @@
...
+    postgresql_client: asyncpg.Connection, migrate_db_from: abc.Callable[[], abc.Awaitable[None]]
+) -> None:
+    result = await postgresql_client.fetch(
+        """
+            SELECT * FROM public.environment WHERE name='dev-1';
+        """
+    )
+    settings = json.loads(result[0]["settings"])
+    assert isinstance(settings["autostart_agent_deploy_interval"], int)
+    assert isinstance(settings["autostart_agent_repair_interval"], int)
```
arnaudsjs, line 37:
```suggestion
    assert isinstance(settings[data.AUTOSTART_AGENT_DEPLOY_INTERVAL], int)
    assert isinstance(settings[data.AUTOSTART_AGENT_REPAIR_INTERVAL], int)
```

tests/db/migration_tests/test_v202310040_to_v202310090_.py
```diff
@@ -0,0 +1,48 @@
...
+    await migrate_db_from()
+
+    result = await postgresql_client.fetch(
+        """
+            SELECT * FROM public.environment WHERE name='dev-1';
+        """
+    )
+    settings = json.loads(result[0]["settings"])
+    assert isinstance(settings["autostart_agent_deploy_interval"], str)
+    assert isinstance(settings["autostart_agent_repair_interval"], str)
```
arnaudsjs, line 48:
```suggestion
    assert isinstance(settings[data.AUTOSTART_AGENT_DEPLOY_INTERVAL], str)
    assert isinstance(settings[data.AUTOSTART_AGENT_REPAIR_INTERVAL], str)
```

tests/db/migration_tests/test_v202310040_to_v202310090_.py
arnaudsjs: Do you know why we got the pool exhaustion. This is very strange.

arnaudsjs (changes requested):

tests/db/migration_tests/test_v202310040_to_v202310090_.py
Hugo-Inmanta: I find it strange as well, I have no idea

Hugo-Inmanta:
> Do we also have a test somewhere that verifies that getting these settings via the API returns data in the right type?

Added the `test_api_return_type` test in tests/server/test_env_settings.py

src/inmanta/data/model.py
```diff
@@ -234,7 +235,7 @@ def check_serializable(cls, v: Optional[Any]) -> Optional[Any]:
         return v
 
 
-EnvSettingType = Union[StrictNonIntBool, int, float, str, Dict[str, Union[str, int, StrictNonIntBool]]]
+EnvSettingType = Union[StrictNonIntBool, StrictInt, StrictFloat, StrictStr, Dict[str, Union[str, int, StrictNonIntBool]]]
```
Hugo-Inmanta, line 238: Without strict validation, e.g. `"86400"` would pass validation as an `int` (which comes before `str` in the list) and get type coerced into `86400`.

tests/server/test_env_settings.py
```diff
@@ -38,6 +38,17 @@ def check_only_contains_default_setting(settings_dict: Dict[str, object]) -> Non
         assert setting_value == get_environment_setting_default(setting_name)
 
 
+async def test_api_return_type(client, server, environment_default):
```
arnaudsjs, line 41: Why does this test case use the `environment_default` fixture instead of the `environment` fixture we always use?

arnaudsjs (changes requested):

tests/server/test_env_settings.py
Hugo-Inmanta: Ah, I copy pasted the test definition from `test_environment_settings`. Fixed to use the `environment` fixture.
I wonder, what would the issue be with using `environment_default` instead ? I'm guessing it is the fact that auto_deploy is true in that one ? Could it cause flakiness/slowness ?

arnaudsjs (approved):

PR #6597, opened 2023-10-09
Resultcollector refactor

sanderr:
# Description

- Refactored result collector propagation through the AST: each expression is now responsible for reporting values to the result collector. Composite statements like `CreateList` now pass on the collector to their children rather than to report values directly themselves. This has the advantage that it makes the gradual execution slightly more gradual: composite statements can rely on their children to be gradual as well, and don't need to implement custom logic for eager gradual reporting of elements as they come in. This has become increasingly relevant as new composite statements have been added over the years (conditional expression, list comprehension, ...).
- Fixed bug in list comprehension: if the value expression consisted of a non-gradually executable statement like e.g. a constructor, this would block the gradual execution of the list comprehension itself, causing `ListModifiedAfterFreeze`. The above refactor now enables the constructor (or any other expression) to report values directly to the gradual result collector. This shifting of responsibility from the composite statement to the substatements makes for simpler and in this case more correct gradual execution.

# Self Check:

- [x] Changelog entry


src/inmanta/ast/statements/__init__.py
```diff
@@ -230,6 +221,19 @@ def execute_direct(self, requires: abc.Mapping[str, object]) -> object:
...
+    ) -> dict[object, VariableABC]:
+        """
+        Returns a dict of the result variables required for execution. Behaves like requires_emit, but additionally may attach
+        resultcollector as a listener to result variables.
+        When this method is called, the caller must make sure to eventually call `execute` as well.
+
+        Composite statements (e.g. conditional expression) will pass result collectors to their children rather than
+        report to them themselves.
+        """
+        return {**self.requires_emit(resolver, queue), (self, ResultCollector): WrappedValueVariable(resultcollector)}
```
sanderr, line 235: I ended up passing it through the requires dict after all because it's a well-established method and I wanted to stick to one optimization at a time. I believe the other approach should be made sufficiently generic so that it can be used for e.g. the promises as well, and when I started on it I realized that it would take quite some time. I thought it better to defer it and create a ticket for it.

src/inmanta/execute/runtime.py
```diff
@@ -143,7 +143,7 @@ def get_value(self) -> T_co:
         """
         raise NotImplementedError()
 
-    def listener(self, resultcollector: ResultCollector[T_co], location: Location) -> None:
+    def listener(self, resultcollector: ResultCollector[T_co], location: Location) -> bool:
```
sanderr, line 146: I don't love this but it's the best I could think of.

tests/compiler/test_list_comprehension.py
```diff
@@ -290,49 +290,136 @@ def test_list_comprehension_nested_tail(snippetcompiler) -> None:
...
+                    for o in base.others
+                ]
+            )
+            if w.others is defined:
+                # again bad practice but asserts gradual execution of the `is defined`
+                w.others += A()
+            end
+
+            # verify that a constructor expression does not block gradual execution,
+            # which would result in a list modified after freeze
```
sanderr, line 346: This is actually the only one that was broken on master. The others were inspired by issues / non-trivial edge cases I encountered during the refactor.

wouterdb (approved):
I think this is fine, but my ability to judge this is decreasing.

tests/compiler/test_list_comprehension.py
```diff
@@ -290,49 +290,137 @@ def test_list_comprehension_nested_tail(snippetcompiler) -> None:
     compiler.do_compile()
 
 
+# TODO: tests failures for new tests on master
```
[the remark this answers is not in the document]
sanderr: Thanks. Done but forgot to clean up the todo.

PR #6598, opened 2023-10-09
Add server config option for tz aware timestamps [iso7]

Hugo-Inmanta:
# Description

On iso7 we try to enable the option by default

Part of https://github.com/inmanta/inmanta-core/issues/6428

# Self Check:

- [ ] Changelog entry


src/inmanta/protocol/common.py
```diff
@@ -980,10 +981,10 @@ def attach_warnings(code: int, value: Optional[JsonType], warnings: Optional[Lis
     return code, value
 
 
-def json_encode(value: ReturnTypes) -> str:
+def json_encode(value: ReturnTypes, tz_aware: bool = False) -> str:
```
sanderr, line 984: I think it might be better to make this `True` by default (for all methods). Places where you currently don't explicitly set the option (again, for all methods):
- `inmanta.data` -> doesn't really matter here except that if we change it the attribute hash will change but only once. I think this is acceptable but @wouterdb knows best
- `inmanta.agent.handler` -> doesn't matter too much but I think we can safely change it so I think we should
- `inmanta.data.model` -> doesn't matter at all
- `gzipped_json` method 5 lines down from this comment -> if I understand correctly this is what the client uses to send data to the server. I think we can safely change this one
- `inmanta.export` -> doesn't matter too much but again, no reason not to
- `inmanta.resources` -> doesn't matter at all
- `inmanta.protocol.openapi.converter` -> I think it doesn't matter here

sanderr: - `inmanta.data`: should've been more clear: I'm talking about the `Resource.make_hash` method here.

src/inmanta/server/services/resourceservice.py
```diff
@@ -748,7 +747,7 @@ async def resource_action_update(
         started: datetime.datetime,
         finished: datetime.datetime,
         status: Optional[Union[const.ResourceState, const.DeprecatedResourceState]],
-        messages: List[Dict[str, Any]],
```
wouterdb, line 751: This I believe is not possible, we can't just change this part of the external api?

tests/test_cli.py
```diff
@@ -382,19 +383,19 @@ def assert_nr_records_in_output_table(output: str, nr_records: int) -> None:
         finished=datetime.datetime.now(),
         status=ResourceState.failed,
         messages=[
-            {
```
wouterdb, line 385: this you can not do I think, older agents will still send the data the old way

src/inmanta/protocol/common.py
wouterdb: @sanderr I don't see how this will affect the attribute hash?

sanderr: Concluded on Slack: shouldn't affect the attribute hash.

sanderr (changes requested):
I was trying to rebase this to be able to see the diff with the iso6 PR but then I realized that we'd actually better merge it in two phases as well: first the iso6+master part, then the master-specific part on top of that. Because otherwise the history becomes a mess and we'll never be able to make sense of what changed exactly if we come back to this PR. Could you make that happen, and request review on the revised PR?

src/inmanta/protocol/openapi/converter.py
```diff
@@ -49,10 +49,10 @@
 from inmanta.types import ReturnTypes
 
 
-def openapi_json_encoder(o: object) -> Union[ReturnTypes, util.JSONSerializable]:
+def openapi_json_encoder(o: object, tz_aware: bool = True) -> Union[ReturnTypes, util.JSONSerializable]:
```
sanderr, line 52: I don't really see a need for the parameter here if we never pass it explicitly. I would revert the changes to this file.

src/inmanta/protocol/common.py
```diff
@@ -985,14 +985,14 @@ def attach_warnings(code: int, value: Optional[JsonType], warnings: Optional[Lis
...
-def json_encode(value: ReturnTypes, tz_aware: bool = False) -> str:
+def json_encode(value: ReturnTypes, tz_aware: bool = True) -> str:
     """Our json encode is able to also serialize other types than a dict."""
     # see json_encode in tornado.escape
     return json.dumps(value, default=partial(custom_json_encoder, tz_aware=tz_aware)).replace("</", "<\\/")
 
 
-def gzipped_json(value: JsonType) -> Tuple[bool, Union[bytes, str]]:
-    json_string = json_encode(value)
+def gzipped_json(value: JsonType, tz_aware: bool = True) -> Tuple[bool, Union[bytes, str]]:
```
sanderr, line 994: The changes to this method too seem redundant.

sanderr (changes requested):

sanderr (approved):

PR #6602, opened 2023-10-09
resource_deploy_done called without calling resource_deploy_start

### LINKED ISSUE #6589: fix bug in undeployable resources handling
#6556 introduced a bug where deploy done is called for a resource that was never started. See [Slack discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1696514329431679?thread_ts=1696514183.755329&cid=CKRF0C8R3)

arnaudsjs:
# Description

It seems it was not required to call `resource_deploy_done()` for undeployable resources. The `resource_action_update()` method called by the `release_version()` API endpoint for undeployable resource invokes the `mark_done_if_done()` method. If the model consists of undeployable resources, the `release_version()` method make sure the status of the model is set to deployed.

closes #6589 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -211,6 +211,25 @@ class ResourceAction(ResourceActionBase):
             # Explicit cast is required because mypy has issues with * and generics
             results: List[ResourceActionResult] = cast(List[ResourceActionResult], await asyncio.gather(*waiters))
 
+            if self.undeployable:
+                self.running = True
+                try:
+                    if self.is_done():
+                        # Action is cancelled
+                        self.logger.log(const.LogLevel.TRACE.to_int, "%s %s is no longer active" % (self.gid, self.resource))
+                        return
+                    result = sum(results, ResourceActionResult(cancel=False))
+                    if result.cancel:
+                        # self.running will be set to false when self.cancel is called
+                        # Only happens when global cancel has not cancelled us but our predecessors have already been cancelled
+                        return
+                    self.status = self.undeployable
+                    self.change = const.Change.nochange
+                    self.changes = {}
+                    self.future.set_result(ResourceActionResult(cancel=False))
+                finally:
+                    self.running = False
+
             async with self.scheduler.ratelimiter:
                 ctx = handler.HandlerContext(self.resource, logger=self.logger)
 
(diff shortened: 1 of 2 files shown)


src/inmanta/agent/agent.py
```diff
@@ -211,6 +211,25 @@ async def execute(
...
+                    if result.cancel:
+                        # self.running will be set to false when self.cancel is called
+                        # Only happens when global cancel has not cancelled us but our predecessors have already been cancelled
+                        return
+                    self.status = self.undeployable
+                    self.change = const.Change.nochange
+                    self.changes = {}
+                    self.future.set_result(ResourceActionResult(cancel=False))
+                finally:
+                    self.running = False
```
sanderr, line 231: Shouldn't we return here?

src/inmanta/agent/agent.py
```diff
@@ -211,6 +211,25 @@ async def execute(
...
+                        return
+                    result = sum(results, ResourceActionResult(cancel=False))
+                    if result.cancel:
+                        # self.running will be set to false when self.cancel is called
+                        # Only happens when global cancel has not cancelled us but our predecessors have already been cancelled
+                        return
+                    self.status = self.undeployable
+                    self.change = const.Change.nochange
+                    self.changes = {}
+                    self.future.set_result(ResourceActionResult(cancel=False))
```
sanderr, line 229: I don't really understand why you put this in a separate block, rather than to keep the condition you had before. Then we'd only need these four lines and the return part, rather than having to duplicate the whole block above.

sanderr (changes requested):

src/inmanta/agent/agent.py
arnaudsjs: Indeed. I forgot to add it. Thanks!

src/inmanta/agent/agent.py
arnaudsjs: We need to make sure that we don't set the result of the future twice, as this would result in an exception. Setting `running` to True and checking the value of `is_done()` and `cancel` are required to make sure this doesn't happen. Strictly speaking it's not required to set `running` to True here because the code is entirely synchronous, but I added it here anyway for safety reasons. If we add an asynchronous call to this try-finally block in the future, we will probably not remember to add the try-finally block and set `running` to True...

sanderr: I understand that part, it's just that between this branch of the condition and the other there seems to be only a small difference. The whole complexity surrounding it with the `is_done` and `cancel` is duplicated here and I see no reason for that. Am I missing something?

sanderr: If it's not clear what I mean I can provide a rough diff.

arnaudsjs: I think I understand your question. I mainly did that to bypass the rate limiter of the agent, because there is noreason why we would wait for it.

sanderr: Ok, that's a good point.

sanderr (approved):

Hugo-Inmanta (approved):

PR #6614, opened 2023-10-11
Configure postgresql to stop queries when the connection drops

wouterdb:
# Description

Configure postgresql to stop queries when the connection drops

Should I wait with merging this until after the release?

# Self Check:

- [x] Changelog entry


sanderr (approved):
Do we dare merge this the day before the release?

arnaudsjs (approved):

sanderr:
> Should I wait with merging this until after the release?

I didn't see you were asking the same question. I think we should wait, especially considering that the issue should become less relevant after the release is out. We could choose to add an `other-note` to the changelog that this is a known issue, that it will be fixed in 6.5 and to restart the postgres service when you run into it on 6.4?

wouterdb:
no, this is fairly obscure

PR #6616, opened 2023-10-11
Prevent overload by the Cross Agent Dependencies

wouterdb:
# Description

We have noticed that the cross agent dependency mechanism can flood the agents http client queue

this makes two changes

1. rate limit the amount of in-flight CAD requests
2. eliminate the return value on CAD notification from the RPC protocol 
3. force server-agent calls to batch up (first one goes immediately, next ones batch up). Currently, the agent return so fast we have no batching at all. 

- [ ] Do we need additional testing for this?

# Self Check:

- [x] Changelog entry


src/inmanta/agent/agent.py
```diff
@@ -1103,6 +1109,8 @@ def __init__(
         self.hostname = hostname
         self.poolsize = poolsize
         self.ratelimiter = asyncio.Semaphore(poolsize)
+        # Number of in flight requests for resolving CAD's
+        self.cad_ratelimiter = asyncio.Semaphore(3)
```
wouterdb, line 1113: total connection pool is 10 
I could go higher here

src/inmanta/protocol/endpoints.py
```diff
@@ -293,9 +293,9 @@ async def dispatch_method(self, transport: client.RESTClient, method_call: commo
                 "No such method",
             )
             LOGGER.error(msg)
-            self.add_background_task(
-                self._client.heartbeat_reply(self.sessionid, method_call.reply_id, {"result": msg, "code": 500})
-            )
+            if method_call.reply_id is not None:
+                await self._client.heartbeat_reply(self.sessionid, method_call.reply_id, {"result": msg, "code": 500})
```
wouterdb, line 297: the background task was a bit moot, as this IS a background task

src/inmanta/agent/agent.py
arnaudsjs: I am not sure what is best here. But 3 sounds reasonable to me.

arnaudsjs (changes requested):

src/inmanta/agent/agent.py
```diff
@@ -311,43 +311,49 @@ async def execute(
...
-            status = const.ResourceState[result.result["resource"]["status"]]
-            self.running = True
-            if status in const.TRANSIENT_STATES or self.future.done():
-                # wait for event
-                pass
-            else:
-                if "logs" in result.result and len(result.result["logs"]) > 0:
-                    log = result.result["logs"][0]
+                self.running = True
+                result = await self.scheduler.get_client().get_resource(
```
wouterdb, line 321: I've got some data from production

I'm afraid we do a LOT of these, we may want to batch them up.

The ability to cancel them (we add here), may also help somewhat. 

```json
"rpc.get_resource": {
    "avg": 0.2436818535260758,
    "sum": 19656.353032827377,
    "count": 80664.0,
    "max": 3.0693390369415283,
    "min": 0.0059680938720703125,
    "std_dev": 0.13372318645433742,
    "15m_rate": 1.459812261442336,
    "5m_rate": 0.7380262996161672,
    "1m_rate": 0.007291537878714465,
    "mean_rate": 1.008132603019391,
    "50_percentile": 0.14774608612060547,
    "75_percentile": 0.19217169284820557,
    "95_percentile": 0.3021209955215453,
    "99_percentile": 0.48956055402755744,
    "999_percentile": 1.5138091018199922
  },
```

src/inmanta/agent/agent.py
```diff
@@ -311,43 +311,49 @@ async def execute(
...
 
-            status = const.ResourceState[result.result["resource"]["status"]]
-            self.running = True
-            if status in const.TRANSIENT_STATES or self.future.done():
-                # wait for event
-                pass
-            else:
-                if "logs" in result.result and len(result.result["logs"]) > 0:
-                    log = result.result["logs"][0]
+                self.running = True
```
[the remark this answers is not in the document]
wouterdb: good catch

src/inmanta/protocol/endpoints.py
sanderr: Is it? `Client._call` / `RESTClient.call` wait for the HTTP request to complete, don't they?

sanderr: Also, not specific to this PR but shouldn't this use the `self._heartbeat_client`?

src/inmanta/server/protocol.py
```diff
@@ -726,7 +734,15 @@ async def heartbeat(
         session: Session = await self.get_or_create_session(sid, env.id, set(endpoint_names), nodename)
 
         LOGGER.debug("Let node %s wait for method calls to become available. (long poll)", nodename)
+
+        # keep a minimal timeout between sending out calls to allow them to batch up
+        now = time.monotonic()
+        wait_time = session.dispatch_delay - (now - session.last_dispatched_call)
```
sanderr, line 740: `last_dispatched_call` is never set?

src/inmanta/server/protocol.py
```diff
@@ -426,6 +429,9 @@ def __init__(
         self._callhandle: Optional[asyncio.TimerHandle] = None
         self.expired: bool = False
 
+        self.last_dispatched_call: float = 0
+        self.dispatch_delay = 0.1  # keep at least 100 ms between dispatches
```
sanderr, line 433: Isn't 100ms a lot? For fast calls it can be a big increase in latency.

src/inmanta/agent/agent.py
sanderr: I really don't have a view on what's sensible here.

src/inmanta/agent/agent.py
```diff
@@ -311,43 +311,50 @@ async def execute(
...
-                log_limit=1,
-            )
-            if result.code != 200 or result.result is None:
-                LOGGER.error("Failed to get the status for remote resource %s (%s)", str(self.resource_id), result.result)
-                return
+        async with self.scheduler.agent.process.cad_ratelimiter:
+            try:
+                # got event or cancel first
+                if self.is_done():
+                    return
```
sanderr, line 318: Should we log something here? The other `execute` implementations do.

sanderr (changes requested):

src/inmanta/protocol/endpoints.py
wouterdb: 1. yes it is, the method calling into this is ran as a background task
2. No, the `_heartbeat_client` is reserved for the forward direction, to make sure it is always available. This reverse direction can generate significant traffic volumes, so it should explicitly kept of that pool.

src/inmanta/server/protocol.py
wouterdb: agreed

src/inmanta/agent/agent.py
wouterdb: This one is kind of an invisible internal mechanism.

This is not ideal, as this is becoming slow and slow/invisible is a bad combination.

arnaudsjs (approved):

sanderr (changes requested):
Did you push your latest changes? You commented on some threads I started but I don't see the associated changes.

sanderr (approved):

PR #6630, opened 2023-10-17
Change logger name of compile and export commands

### LINKED ISSUE #6489: Remove first column from compiler output
```
inmanta.compiler         DEBUG   Starting compile
inmanta.module           DEBUG   Parsing took 0.003764 seconds
inmanta.parser.cache     DEBUG   Compiler cache observed 2 hits and 0 misses (100%)
inmanta.module           INFO    verifying project
inmanta.module           DEBUG   Loading module inmanta_plugins.std
inmanta.module           DEBUG   Loading module inmanta_plugins.std.resources
inmanta.module           DEBUG   Parsing took 0.000167 seconds
inmanta.parser.cache     DEBUG   Compiler cache observed 2 hits and 0 misses (100%)
inmanta.module           INFO    The following modules are currently installed:
inmanta.module           INFO    V2 modules:
inmanta.module           INFO      std: 4.2.1.dev20230911043312
```

Should become:

```
DEBUG   Starting compile
DEBUG   Parsing took 0.003764 seconds
DEBUG   Compiler cache observed 2 hits and 0 misses (100%)
INFO    verifying project
DEBUG   Loading module inmanta_plugins.std
DEBUG   Loading module inmanta_plugins.std.resources
DEBUG   Parsing took 0.000167 seconds
DEBUG   Compiler cache observed 2 hits and 0 misses (100%)
INFO    The following modules are currently installed:
INFO    V2 modules:
INFO      std: 4.2.1.dev20230911043312
```

arnaudsjs:
# Description

This PR improves the output of the `inmanta compile` and `inmanta export` commands, by using the logger name `compiler`, `exporter` or `<name-inmanta-module>` for log records produced by respectively the compiler, the exporter or an Inmanta module.

closes #6489

# Self Check:

- [x] Changelog entry


src/inmanta/logging.py
```diff
@@ -99,15 +117,103 @@ def __init__(self, stream: TextIO = sys.stdout) -> None:
...
         logging.root.setLevel(0)
 
+    def _get_module_name_for_source_file(self, path_source_file: str) -> Optional[str]:
+        """
+        Return the module name that the given `path_source_file` belongs to or None if `path_source_file`
+        doesn't belong to a module.
+        """
+        if path_source_file in self._source_file_to_module_cache:
+            return self._source_file_to_module_cache[path_source_file]
+        dirs_containing_modules = []
```
FloLey, line 143: couldn't this one be cached as well? but I think it's overkill

src/inmanta/logging.py
```diff
@@ -99,15 +117,103 @@ def __init__(self, stream: TextIO = sys.stdout) -> None:
...
+        """
+        This log record factory makes sure that the name of the log record is updated
+        in the following way while executing in the "compiler" or "exporter" logger mode:
+
+        * The name of the Inmanta module: When the log record was created by a source file in an Inmanta module.
+        * compiler: When executing in compiler mode and the log record doesn't come from an Inmanta module.
+        * exporter: When executing in exporter mode and the log record doesn't come from an Inmanta module.
+        """
+        new_logger_name: str
+        if self._logger_mode is LoggerMode.COMPILER or self._logger_mode is LoggerMode.EXPORTER:
```
FloLey, line 186:
```suggestion
        if self._logger_mode in [LoggerMode.COMPILER, LoggerMode.EXPORTER]:
```
I just like it more but feel free to ignore

FloLey (approved):

src/inmanta/logging.py
arnaudsjs: Indeed. Much better.

src/inmanta/logging.py
arnaudsjs: Good point. Initially I though we couldn't, but since the compiler doesn't install modules anymore, we can safely assume that the content of `dirs_containing_modules` will not change during execution.

sanderr (changes requested):
I wonder if we should add a config option to toggle this behavior. I know that I will probably want to disable it often when developing, but I can easily hack it. But I can imagine that not all end users even would like their loggers to be converted automatically.

src/inmanta/logging.py
```diff
@@ -99,15 +117,103 @@ def __init__(self, stream: TextIO = sys.stdout) -> None:
...
+        if Project._project:
+            # Directories containing V1 modules
+            dirs_containing_modules += [p for p in Project._project._metadata.modulepath]
+        if (
+            "inmanta_plugins" in sys.modules
+            and sys.modules["inmanta_plugins"].__spec__
+            and sys.modules["inmanta_plugins"].__spec__.submodule_search_locations
+        ):
+            # Directories containing v2 modules
+            dirs_containing_modules += [str(s) for s in sys.modules["inmanta_plugins"].__spec__.submodule_search_locations]
```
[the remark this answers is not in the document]
arnaudsjs: Ok. Let's go for the less accurate, but more robust approach.

src/inmanta/logging.py
```diff
@@ -99,15 +117,103 @@ def __init__(self, stream: TextIO = sys.stdout) -> None:
...
+        in the following way while executing in the "compiler" or "exporter" logger mode:
+
+        * The name of the Inmanta module: When the log record was created by a source file in an Inmanta module.
+        * compiler: When executing in compiler mode and the log record doesn't come from an Inmanta module.
+        * exporter: When executing in exporter mode and the log record doesn't come from an Inmanta module.
+        """
+        new_logger_name: str
+        if self._logger_mode is LoggerMode.COMPILER or self._logger_mode is LoggerMode.EXPORTER:
+            inmanta_module_name: Optional[str] = self._get_module_name_for_source_file(pathname)
+            new_logger_name = inmanta_module_name if inmanta_module_name else self._logger_mode.value
```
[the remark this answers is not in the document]
arnaudsjs: I don't understand the question. The if-statement already verified that `self._logger_mode` is either `LoggerMode.COMPILER` or `LoggerMode.EXPORTER`. So it can't be "other", right?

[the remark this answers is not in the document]
sanderr: You're right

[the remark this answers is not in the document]
arnaudsjs: There has been a discussion about whether or not the log messages have to be prefixed with `inmanta` or not and the conclusion was that we would not do that.

src/inmanta/app.py
```diff
@@ -828,6 +828,13 @@ def cmd_parser() -> argparse.ArgumentParser:
         default=False,
         required=False,
     )
+    parser.add_argument(
+        "--keep-logger-names",
```
sanderr, line 832: Thanks!

src/inmanta/logging.py
```diff
@@ -178,14 +153,18 @@ def custom_log_record_factory(
...
         new_logger_name: str
-        if self._logger_mode is LoggerMode.COMPILER or self._logger_mode is LoggerMode.EXPORTER:
-            inmanta_module_name: Optional[str] = self._get_module_name_for_source_file(pathname)
-            new_logger_name = inmanta_module_name if inmanta_module_name else self._logger_mode.value
+        if not self._keep_logger_names and self._logger_mode in [LoggerMode.COMPILER, LoggerMode.EXPORTER]:
+            match: Optional[re.Match[str]] = self._inmanta_plugin_pkg_regex.match(name)
+            if match:
+                new_logger_name = match.groupdict()["module_name"]
+            else:
+                new_logger_name = self._logger_mode.value
```
sanderr, line 167: There is still the possibility of this being a third party warning, in which case I think we want to keep the original. There is also the case of what is currently the `inmanta.pip` logger that would now become `compiler`, right?

sanderr (changes requested):

src/inmanta/logging.py
arnaudsjs: The question is what we want to achieve by changing the logger names of the log records. I understood the intention was to make it clear to the end user whether we are currently executing the compile or exporter stage. From that perspective it's desired behavior to log everything that comes from the `inmanta.pip` logger as `compiler`. But maybe I misunderstood the goal of this ticket. Do you have something else in mind?

sanderr: I'm not sure to be honest. But I seem to recall we introduced the artificial `inmanta.pip` logger explicitly to make it clear that this came from the pip subprocess. I find it strange to put it back under the `compiler` banner but admittedly I've never been very good at this user-oriented stuff so I also don't want to press my opinion too much. Perhaps @bartv has an opinion on this?

bartv: I think `pip` would be a good idea for that output. Because it also is not compiler.

src/inmanta/logging.py
```diff
@@ -160,11 +160,14 @@ def custom_log_record_factory(
...
-                new_logger_name = match.groupdict()["module_name"]
+            if name == "inmanta.pip":
+                new_logger_name = "pip"
             else:
-                new_logger_name = self._logger_mode.value
+                match: Optional[re.Match[str]] = self._inmanta_plugin_pkg_regex.match(name)
+                if match:
+                    new_logger_name = match.groupdict()["module_name"]
+                else:
+                    new_logger_name = self._logger_mode.value
```
sanderr, line 170: This still captures third party libraries' logs, no? I was thinking more that we could limit ourselves to `inmanta.*`.

sanderr (changes requested):

sanderr (approved):

PR #6637, opened 2023-10-19
set connection options

wouterdb:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -6261,9 +6274,7 @@ async def connect(
         min_size=connection_pool_min_size,
         max_size=connection_pool_max_size,
         timeout=connection_timeout,
-        server_settings={
-            "client_connection_check_interval": "1000",  # make server check the tpc connection every 1 second
-        },
+        setup=set_options,
```
wouterdb, line 6277: We somehow have to do it every time or it gets reset

Issue #6440, opened 2023-08-29
support models that will never run a full compile


sanderr:
Some models will be so large that they can only realistically run partial compiles. However, core currently requires a full compile in some cases. We should figure out when this is required, why, and how we can work around this.

Potential full compile requirements in the current implementation:
- first compile
- any compile that updates handler code
- ...?

PR #6644, opened 2023-10-19
Issue/6518 pip config docs

### LINKED ISSUE #6518: pip config: docs
Add documentation for the new pip config and its behavior according to the [design](https://github.com/inmanta/designs/blob/master/modules/pip_config.md). Make sure to also include any appropriate change entries (implementation tickets will not include any).

Hugo-Inmanta:
# Description

closes #6518

Document the [breaking changes](https://github.com/inmanta/designs/blob/master/modules/pip_config.md#breaking-changes) from the design:

- [x] The module source of type ``package`` will be ignored (with a warning) and removed in a later version. In the current stable
    iso, there is no warning about this yet. We should add a warning that this field will be ignored in the next major and
    recommend that the user already duplicates these indexes in ``pip.index-url``. The warning should not fire if this option is
    already set.
- [x] v1 module dependency installation: unless ``pip.use-system-config=true``, v1 modules' dependencies will now use the index(es)
    defined in ``pip.{,extra-}index-url`` and only those. Projects that are still using v1 modules will need to set the index
    url option accordingly.
- [x] the agent now depends on the pip config of the project, and therefore requires access to the project's configured indexes
- [ ] breaking changes relative to OSS are not covered in this list
- [x] ``--pre`` will no longer be set based on the module install mode, instead it will follow the newly introduced option.
- [x] ``PIP_PRE`` environment variable will now be ignored unless ``use-system-config = true``

--------

- [x]  breaking changes listed above should be documented, including links to docs to configure alternative behavior
- [ ]  additionally include breaking changes relative to latest OSS release but clearly mention that this is not breaking relative to
  iso5
- [ ]  documentation on the master branch should be reviewed to verify that it matches the new behavior
- [x]  netrc remains the recommended method to set up authentication for a pip index url.
- [x]  add common example scenarios and recommended options, namely a single index-url and ``use-system-config = false``.
    ``pre`` should be set to ``true`` during development and toggled for production.
- [x]  track new partial export constraints in [inmanta/inmanta-core#6440](https://github.com/inmanta/inmanta-core/issues/6440)
- [x]  document agent ``pip install`` flow
- [x]  document clearly that pip config files are always ignored if ``use-system-config = false``
- [x]  document clearly that the project options override the following environment variables
    (and respective config options even if ``use-system-config = true``)
    - ``PIP_PRE`` superseded by project config value, unless unset and ``use-system-config = true``
    - ``PIP_INDEX_URL`` superseded by project config value, unless unset and ``use-system-config = true``
    - ``PIP_EXTRA_INDEX_URL`` unless  ``use-system-config = true`` (in which case the union of all configured extras is used)
    - other pip-related environment variables are not overridden/ignored because inmanta is not aware of them.
- [x]  recommend running a full compile after upgrading to export project pip config to the server

# Self Check:

- [ ] Changelog entry


Hugo-Inmanta:
Points of attention:
- The changelog file itself
    - `major` is not technically right since this is only doc but I think it's necessary (I might be wrong)
    - I added some links to the doc itself in there but I didn't test if they render properly or work from within the Changelog. I'd like something similar to https://docs.inmanta.com/inmanta-service-orchestrator/latest/CHANGELOG.html#id110 but I couldn't find the related change entry for that one

- I think I'm missing the part about documenting breaking changes relative to the latest OSS release (the last few unticked boxes) ?

docs/reference/projectyml.rst
```diff
@@ -50,66 +50,122 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
-  [global]
-  timeout = 60
-  index-url = <url of a python package repository>
+.. code-block:: yaml
 
+    pip:
+        index-url: https://devpi.example.com/stable/
+        extra-index-url: []
+        pre: false
+        use-system-config: true
```
wouterdb, line 119: I find this one a bit confusing. 

I would also explain that for production scenario's, we expect all packages to be hosted in an internal, curated python repository (like nexus or devpi) and that the default pip config, which pull packages from `https://pypi.org/` is never used.

(this explanation may also belong elsewhere,  like e.g. https://docs.inmanta.com/inmanta-service-orchestrator/5/administrators/operational_procedures.html#project-release-for-production )

In production scenarios, where the system config is fully managed at the system level and secure, I would expect just `use-system-config: true`  or perhaps also a `extra-index-url:` for a project specific repo that is not managed at the system level.

In production scenarios, where the system config is not managed and secure, I would not expect use-system-config and to set `index-url:` to the secure internal repo.

```
  pip:
        index-url: https://my_repo.secure.company.com/repository/inmanta-production
        extra-index-url: [https://my_repo.secure.company.com/repository/project-alpha-production]
        pre: false
        use-system-config: false
 ```

File: docs/reference/projectyml.rst [hunk 2]
```diff
@@ -50,66 +50,122 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
 
 .. note::
-    The pip config file can also be used in combination with ``index-urls`` specified in the ``pip`` section of the ``project.yml`` file:
 
-    * If the pip config is used (by setting ``use_config_file`` to ``true``), the ``index-url`` specified in the pip config file will take precedence and the ``index-urls`` specified in the ``pip`` section of the ``project.yml`` file will be used as ``extra-index-urls`` when installing with pip.
-    * If the pip config is not used (by setting ``use_config_file`` to ``False``), then the first ``index_url`` specified in the project.yml will be used as an ``index_url`` and all the following ones will be used as ``extra-index-urls`` when installing with pip.
+    Using netrc is the recommended way to set up authentication towards the index. See
+    this :ref:`section<setting_up_pip_index_authentication>` for more information.
+
+.. _migrate_to_project_wide_pip_config:
```
wouterdb, line 154: I would see if we can use this label in the changelog

docs/reference/projectyml.rst
```diff
@@ -50,66 +50,122 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+Previously, there was no centralized way of configuring pip settings for the whole project. This section can be used
+as a migration guide.
 
+Defining a ``repo`` with type ``package`` is deprecated. Make sure you define this index through the ``pip.index-url`` option instead.
 
+Previously, the :class:`~inmanta.module.InstallMode` set at the project level or at a module level was used to determine if the
+installation of pre-release versions was allowed. This behaviour should now be set through the ``pip.pre`` option instead.
 
+A full compile should be run after upgrading, in order to export the project pip config to the server, so that it
+is available for agents.
```
wouterdb, line 168: I would add an example of the before an after

src/inmanta/module.py
```diff
@@ -1537,14 +1537,21 @@ def __str__(self) -> str:
 @stable_api
 class ProjectPipConfig(BaseModel):
```
wouterdb, line 1538: Merge conflict straight ahead!

docs/reference/projectyml.rst
```diff
@@ -50,66 +50,122 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
-    The pip config file can also be used in combination with ``index-urls`` specified in the ``pip`` section of the ``project.yml`` file:
 
-    * If the pip config is used (by setting ``use_config_file`` to ``true``), the ``index-url`` specified in the pip config file will take precedence and the ``index-urls`` specified in the ``pip`` section of the ``project.yml`` file will be used as ``extra-index-urls`` when installing with pip.
-    * If the pip config is not used (by setting ``use_config_file`` to ``False``), then the first ``index_url`` specified in the project.yml will be used as an ``index_url`` and all the following ones will be used as ``extra-index-urls`` when installing with pip.
+    Using netrc is the recommended way to set up authentication towards the index. See
+    this :ref:`section<setting_up_pip_index_authentication>` for more information.
+
+.. _migrate_to_project_wide_pip_config:
+
+Migrate to project-wide pip config
```
sanderr, line 156: Should we also document changes relative to `inmanta-2023.4` (OSS), which already has `use-system-config` but in another location (I think) and somewhat bugged (or at least very unintuitive)?

sanderr (changes requested):
We should also take care to document the moduletool pip config, which wasn't part of the design. i.e. what @wouterdb is currently working on. @Hugo-Inmanta @wouterdb I'll leave it up to you to decide who takes this on.

docs/reference/projectyml.rst
```diff
@@ -50,66 +50,122 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
-This section explains how to configure your project in order to download v2 modules from any python package repository.
-By default, a project created using the :ref:`project-creation-guide` is configured to install packages from ``https://pypi.org/simple/``. There are multiple ways
-to change this behaviour.
+This section explains how to configure a project-wide pip index. This index will be used to download v2 modules and v1
+modules' dependencies.
+By default, a project created using the :ref:`project-creation-guide` is configured to install packages from ``https://pypi.org/simple/``.
+The :class:`~inmanta.module.ProjectPipConfig` section of the project.yml file offers options to configure this behaviour.
 
-Using pip config file
+pip.use-system-config
```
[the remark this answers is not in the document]
Hugo-Inmanta: I figured the class documentation for `ProjectPipConfig` was enough for these. Maybe I can add some explicit sections that point to `ProjectPipConfig` ?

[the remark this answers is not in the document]
sanderr: You make a good point. I don't think explicit sections are required. But perhaps where you now have "The :class:`~inmanta.module.ProjectPipConfig` section of the project.yml file offers options to configure this behaviour.", we follow that up by stating that some of those options are explained in more detail below? This would make it more explicit that what follows is not an exhaustive list of all options. Wdyt?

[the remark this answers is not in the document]
Hugo-Inmanta: Great idea :+1:

docs/install/3-configure-agents.rst
```diff
@@ -155,6 +155,8 @@ configure the behavior of the manually started agent:
 * :inmanta.config:option:`agent_rest_transport.ssl`
 * :inmanta.config:option:`agent_rest_transport.ssl-ca-cert-file`
 
+The agent will follow the pip configuration defined in the :ref:`project_yml`. Make sure it can access the pip
```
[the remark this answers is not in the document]
Hugo-Inmanta: Added to the migration guide section (and not in the changelog itself, maybe it's worth mentioning there as well?)

changelogs/unreleased/6518-pip-config-docs.yml
```diff
@@ -0,0 +1,7 @@
+description: Documentation PR for the new pip config and its behaviour. This PR centralizes the change entries.
+issue-nr: 6518
+change-type: major
+destination-branches: [master]
+sections:
+  feature: "A project-wide pip configuration can be set through the project.yml pip section."
+  upgrade-note: "Reworked the orchestrator's pip configuration: please refer to the [migration guide](migrate_to_project_wide_pip_config) to move to a project-wide pip configuration."
```
wouterdb, line 7:
```suggestion
  upgrade-note: "All projects now require a pip config in the `project.yml`: please refer to the [migration guide](migrate_to_project_wide_pip_config) to move to a project-wide pip configuration."
```

docs/administrators/operational_procedures.rst
```diff
@@ -15,6 +15,10 @@ Project Release for Production
 
 This process describes the steps to prepare an inmanta project for production release.
 
+All packages should be hosted in an internal, curated python repository (like nexus or devpi) and the default pip
+config, which pulls packages from https://pypi.org/ should never be used. See the :ref:`specify_location_pip` section
+for more information on how to set the project's pip configuration.
```
wouterdb, line 20:
```suggestion
All packages should be hosted in an internal, curated python repository (like nexus or devpi). For security reasons, the default pip config, which pulls packages from https://pypi.org/ should never be used. See the :ref:`specify_location_pip` section
for more information on how to set the project's pip configuration.
```

docs/reference/projectyml.rst
```diff
@@ -50,66 +50,122 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
-By setting the ``use_config_file`` option of the pip section to ``True``, the project will use the pip config files.
+This option determines the isolation level of the project's pip config. When false, any pip config set on the system
+(e.g. through environment variables or pip config files) are ignored and pip will only look for packages
+in the index(es) defined in the project.yml, when true, pip will in addition look in the eventual index(es) defined on the system.
+
+Setting this to ``false`` is recommended during development both for portability (by making sure that only the pip
+config defined in the project.yml will be used regardless of the sytem's pip config) and for security (The isolation
+reduces the risk of dependency confusion attacks if the ``index-url`` option is set mindfully).
+
+Setting this to ``true`` will have the following consequences:
```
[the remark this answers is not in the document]
wouterdb: 1. I would not mention not-autostarted agents, as they are rarely used.
2. /etc/sysconfig/inmanta-server also works for the agents (it configures the env vars for the server process, which forks the agents, so they get the same env vars)

docs/reference/projectyml.rst
```diff
@@ -50,66 +50,165 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+option e.g.:
 
+.. code-block:: yaml
 
-Alternatively, a pip config file can be used at a custom location.
-The ``index-url`` can be specified in this file as explained in the previous section.
-To make this work, the ``PIP_CONFIG_FILE`` environment variable needs to be set to the path of the newly created file (See: :ref:`env_vars`).
-For more information see the `Pip documentation <https://pip.pypa.io/en/stable/topics/configuration/>`_.
+    pip:
+        extra-index-url: [https://my_repo.secure.company.com/repository/project-alpha-production]
```
wouterdb, line 135: I would not use the words `alpha` and `production` in the same  url: `alpha` often means not ready for production

docs/reference/projectyml.rst
```diff
@@ -50,66 +50,165 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
-    * If the pip config is not used (by setting ``use_config_file`` to ``False``), then the first ``index_url`` specified in the project.yml will be used as an ``index_url`` and all the following ones will be used as ``extra-index-urls`` when installing with pip.
+    Set up authentication towards the index using netrc. See this :ref:`section<setting_up_pip_index_authentication>`
+    for more information.
+
+.. _migrate_to_project_wide_pip_config:
+
+Migrate to project-wide pip config
+----------------------------------
+
+``inmanta-core 11.0.0`` introduced new options to configure pip settings for the whole project in a centralized way.
```
wouterdb, line 172: I would add a code sample here (or even add a script to the orchestrator to update all configs) 

This should be very copy-pasteable.

docs/administrators/operational_procedures.rst
wouterdb: I would also add an explanation about the security implications (or link to external source), now it is a bit vague

wouterdb (changes requested):

docs/administrators/operational_procedures.rst
Hugo-Inmanta: Added link to PEP 708

docs/reference/projectyml.rst
Hugo-Inmanta: Added a code sample

src/inmanta/module.py
Hugo-Inmanta: I don't mind taking care of this

docs/administrators/operational_procedures.rst
```diff
@@ -15,6 +15,11 @@ Project Release for Production
 
 This process describes the steps to prepare an inmanta project for production release.
 
+All packages should be hosted in an internal, curated python repository (like nexus or devpi). For security reasons,
```
sanderr, line 18: I would add a qualifier that this does not apply for projects that only use public modules and Python dependencies.

docs/model_developers/project_creation.rst
```diff
@@ -47,7 +47,7 @@ An example ``project.yml`` could be:
   install_mode: release
   requires:
   pip:
-      index-url: https://pypi.org/simple/
+      index-url: https://my_repo.secure.company.com/repository/inmanta-production
```
sanderr, line 50: I personally see no reason not to use PyPi here. It is a valid index for simple projects that use the open source modules only. If we really want to avoid it, I think we must use `example.com` because that is a reserved domain while `company.com` is not.

docs/reference/projectyml.rst
```diff
@@ -50,7 +50,7 @@ The code snippet below provides an example of a complete ``project.yml`` file:
     freeze_recursive: true
     freeze_operator: ~=
     pip:
-        index-url: https://pypi.org/simple/
+        index-url: https://my_repo.secure.company.com/repository/inmanta-production
```
sanderr, line 53: Same here: no `company.com`.

sanderr, line 53: Same here: no `company.com`.

changelogs/unreleased/6518-pip-config-docs.yml
```diff
@@ -3,8 +3,5 @@ issue-nr: 6518
 change-type: major
 destination-branches: [master]
 sections:
-  feature: "A project-wide pip configuration can be set through the project.yml pip section"
-  upgrade-note: "Please refer to the `migration guide <https://docs.inmanta.com/inmanta-service-orchestrator/7/reference/projectyml.html#migrate-to-project-wide-pip-config>`_ to move to a project-wide pip configuration"
-  deprecation-note:
-    - "Allowing the installation of pre-release versions for modules through the :class:`~inmanta.module.InstallMode` is now deprecated in favour of the project.yml ``pip.pre`` section"
-    -
+  feature: "A project-wide pip configuration can be set through the project.yml pip section."
+  upgrade-note: "All projects now require a pip config in the `project.yml`: please refer to the [migration guide](migrate_to_project_wide_pip_config) to move to a project-wide pip configuration."
```
sanderr, line 7: This is a different type of link than the one in the change entry we discovered yesterday. If it works I prefer this one. Have you tested that it does?

docs/reference/projectyml.rst
```diff
@@ -65,29 +65,42 @@ This section explains how to configure a project-wide pip index. This index will
...
+security (the isolation reduces the risk of dependency confusion attacks).
 
 Setting this to ``true`` will have the following consequences:
 
 - If no index is set in the project.yml file i.e. both ``index-url`` and ``extra-index-url`` are unset, then Pip's
   default search behaviour will be used: environment variables, pip config files and then PyPi (in that order).
 
-- If ``index-url`` and/or ``extra-index-url`` are set, they will be used and any index defined in the system's environment
-  variables or pip config files will also be used (in that order) and passed to pip as extra indexes.
+- If ``index-url`` is set, this value be used over any index defined in the system's environment
```
sanderr, line 87:
```suggestion
- If ``index-url`` is set, this value will be used over any index defined in the system's environment
```

docs/reference/projectyml.rst
```diff
@@ -65,29 +65,42 @@ This section explains how to configure a project-wide pip index. This index will
...
+  variables or pip config files.
 
-- The ``PIP_PRE`` environment variable (if set) is no longer ignored and will be used to determine whether pre-release
+- If ``extra-index-url`` is set, these indexes will be used in addition to any extra index defined in the system's
+  environment variables or pip config files, and passed to pip as extra indexes.
+
+- If ``PIP_PRE`` is set, pre-release
```
sanderr, line 93: For the other bullet points you focus on the options in our config file and how they override (or don't) the system config. Here you suddenly take a different tack.

docs/reference/projectyml.rst
```diff
@@ -65,29 +65,42 @@ This section explains how to configure a project-wide pip index. This index will
 modules' dependencies.
 By default, a project created using the :ref:`project-creation-guide` is configured to install packages from ``https://pypi.org/simple/``.
 The :class:`~inmanta.module.ProjectPipConfig` section of the project.yml file offers options to configure this behaviour.
+Some of these options are detailed below:
 
 pip.use-system-config
 """""""""""""""""""""
 
-This option determines the isolation level of the project's pip config. When false, any pip config set on the system
+This option determines the isolation level of the project's pip config. When false (the default), any pip config set on the system
```
sanderr, line 73: > any pip config ...

May need to be rephrased depending on the outcome of [this discussion](https://github.com/inmanta/inmanta-core/pull/6655#discussion_r1386279542).

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
 .. code-block:: yaml
 
     pip:
-        index-url: https://devpi.example.com/dev/
+        index-url: https://my_repo.secure.company.com/repository/dev
```
sanderr, line 116: Same

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
+If the pip configuration is fully managed at the system level and secure, use ``use-system-config: true``
+and add any potential project-specific repo that is not managed at the system level through the ``extra-index-url``
+option e.g.:
 
 .. code-block:: yaml
 
     pip:
-        index-url: https://devpi.example.com/stable/
-        extra-index-url: []
+        extra-index-url: [https://my_repo.secure.company.com/repository/project-production]
```
sanderr, line 133: Same

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
-    The options defined in the ``project.yml`` pip section will always take precedence over the corresponding pip options, even
-    when ``use-system-config`` is set to true (other pip-related environment variables are not overridden).
-
-    For example, in the production scenario above, if the following
-    pip environment variables were set by mistake on the server running the compiler: ``PIP_INDEX_URL=https://devpi.example.com/dev/``
-    and ``PIP_PRE=true``, the config used in the end would still be the one defined in the project.yml, namely
-    ``index-url=https://devpi.example.com/stable/`` and ``pre=false``.
 
-An alternative approach would be to configure all pip-related options through the system config. For example:
+On the other hand, if the pip configuration is not fully managed at the system level, make sure to disable ``use-system-config``
```
sanderr, line 138: I think this is still our envisioned approach / recommendation. So I would move this up and rephrase (because now it's relative to the parapraph before it). For the `use-system-config=true` section then I would not advocate it per se, but phrase it more like "if you prefer to manage ...".

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
 
 .. code-block:: yaml
 
     pip:
-        index-url:
-        extra-index-url: []
-        pre:
-        use-system-config: true
+        index-url: https://my_repo.secure.company.com/repository/inmanta-production
+        extra-index-url: [https://my_repo.secure.company.com/repository/project-production]
```
sanderr, line 145: Same, times two

sanderr, line 145: Why have an extra index url in the example here?

wouterdb (approved):

docs/reference/projectyml.rst
sanderr, line 133: I would keep the example to the simple case of "everything is managed at the system", so drop the extra index. There is already a section that explains that you can set it and what it does, so I really see no need for this, and I think it can be confusing for someone just looking for a config for production.

This may need a slight rephrasing of the paragraph above.

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
+        pre: false
+        use-system-config: false
+
+.. note::
+    The ``pre`` and ``index-url`` options defined in the ``project.yml`` pip section will always take precedence over
```
sanderr, line 150: I think this elaborate description of what takes precedence how would fit better higher up where the `use-system-config` option is explained, rather than in the production section.

In this note I would just keep it short and to the point, along the lines of "any pip config set explicitly in the project config will always take precedence over the system config. For more details see ..."

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
+    the corresponding pip options, even when ``use-system-config`` is set to true.
 
-And set the following env variables:
+    The ``extra-index-url`` option will extend the extra indexes already set at the system level (if any).
 
-.. code-block:: bash
+    Other pip-related environment variables are not overridden.
```
sanderr, line 155: "Other" is quite vague. I would qualify it as settings that are not supported by the inmanta project config.

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
 
-    export PIP_INDEX_URL=https://devpi.example.com/stable/
-    export PIP_PRE=false
+    To use the system's pip configuration without altering it, leave the corresponding option unset.
```
sanderr, line 157: "altering it" sounds as if we would make changes to the config file.

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
-as a migration guide.
+``inmanta-core 11.0.0`` introduced new options to configure pip settings for the whole project in a centralized way.
+For detailed information, see :ref:`here<specify_location_pip>`. The following code sample can be used as a baseline
+in the ``project.yml`` file:
+
+
+.. code-block:: yaml
+
+    pip:
+        index-url: https://my_repo.secure.company.com/repository/inmanta-production
```
sanderr, line 178: Same, times two. Also, I would drop the extra-index-url

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
+Make sure the agents have access to the index(es) configured at the project level.
+
+Run a full compile after upgrading in order to export the project pip config to the server, so that it
+is available for agents. This will ensure that the agents follow the pip config defined in the project. For reference,
+prior to ``inmanta-core 11.0.0``, the agents were always using their respective system's pip config.
+
+Deprecation notes:
+""""""""""""""""""
+
+    - Allowing the installation of pre-release versions for v2 modules through the :class:`~inmanta.module.InstallMode`
```
sanderr, line 203: Sounds like a breaking change to me

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
+      is no longer supported. Use the project.yml ``pip.pre`` section instead.
+
+Breaking changes:
```
sanderr, line 206: I would also mention that `PIP_PRE` is now ignored unless `use-system-config` is set.

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
+
+.. code-block:: yaml
+
+    pip:
+        index-url: https://my_repo.secure.company.com/repository/inmanta-production
+        extra-index-url: [https://my_repo.secure.company.com/repository/project-production]
+        pre: false
+        use-system-config: false
+
+When upgrading to ``inmanta-service-orchestrator 7.0.0`` or ``inmanta 2024.0``, this section can be used as
```
sanderr, line 183: This should be part of the section introduction.

docs/reference/projectyml.rst
```diff
@@ -100,72 +113,111 @@ example of a dev config:
...
 .. note::
 
-    Using netrc is the recommended way to set up authentication towards the index. See
-    this :ref:`section<setting_up_pip_index_authentication>` for more information.
+    Set up authentication towards the index using netrc. See this :ref:`section<setting_up_pip_index_authentication>`
+    for more information.
 
 .. _migrate_to_project_wide_pip_config:
 
 Migrate to project-wide pip config
```
sanderr, line 167: This section should mention the alternative of setting `use-system-config=true`.

docs/reference/projectyml.rst
sanderr: Or even the `index-url`.

sanderr (changes requested):

changelogs/unreleased/6518-pip-config-docs.yml
Hugo-Inmanta: It is actually used in the linked [change entry](https://github.com/inmanta/inmanta-lsm/blob/master/changelogs/1.8.0/566-allow-type-updates.yml) (last line)
I checked that the link there works: https://docs.inmanta.com/inmanta-service-orchestrator/latest/CHANGELOG.html#id142

docs/administrators/operational_procedures.rst
Hugo-Inmanta: Added "All **private** packages (...)"

docs/model_developers/project_creation.rst
Hugo-Inmanta: I tend to agree and prefer pypi

docs/reference/projectyml.rst
```diff
@@ -50,66 +50,174 @@ The code snippet below provides an example of a complete ``project.yml`` file:
...
+
+
+All the v2 module sources currently set in a ``repo`` section of the ``project_yml`` with type ``package`` should
+also be duplicated in the ``pip.index-url`` (and ``pip.extra-index-url`` if more than one index is being used).
+
+If you want to allow pre-releases for v2 modules and other Python packages, set ``pip.pre = true`` in the project config
+file. This used to be controlled by the :class:`~inmanta.module.InstallMode` set at the project level or at a module
+level.
+
+Make sure the agents have access to the index(es) configured at the project level.
```
Hugo-Inmanta, line 191: Should I remove this too ? Since we're not mentioning non-autostarted agents ?

docs/administrators/operational_procedures.rst
sanderr: That's not right.
- Projects that use only public modules and Python dependencies may safely use pypi. This is not clear from this paragraph.
- Projects with at least one private depedency should host ALL (not only private) packages in an internal, curated Python repository.

sanderr: > This is not clear from this paragraph.
Reading it again I see that it is there but it reads difficult to me: you say "you should not do x. You may use x if y". Why not just "unless y, you should not do x" or "you should not do x, except if y"?

docs/reference/projectyml.rst
```diff
@@ -123,38 +124,35 @@ example of a dev config:
 Using a single pip index is still the recommended approach, and the use of pre-release versions should be disabled.
 
 
-If the pip configuration is fully managed at the system level and secure, use ``use-system-config: true``
-and add any potential project-specific repo that is not managed at the system level through the ``extra-index-url``
-option e.g.:
+If the pip configuration is not fully managed at the system level, make sure to disable ``use-system-config``
```
sanderr, line 127:
```suggestion
For a portable project (recommended), disable ``use-system-config``
```

docs/reference/projectyml.rst
sanderr: You're right

docs/reference/projectyml.rst
```diff
@@ -123,38 +124,35 @@ example of a dev config:
...
-    The ``pre`` and ``index-url`` options defined in the ``project.yml`` pip section will always take precedence over
-    the corresponding pip options, even when ``use-system-config`` is set to true.
 
-    The ``extra-index-url`` option will extend the extra indexes already set at the system level (if any).
+.. note::
+    Any pip config set explicitly in the project config will always take precedence over the system config. For more
+    details see `pip.use-system-config`_.
 
-    Other pip-related environment variables are not overridden.
+    Other pip-related settings that are not supported by the project config are not overridden.
```
sanderr, line 152:
```suggestion
    Pip-related settings that are not supported by the project config are not overridden.
```

docs/reference/projectyml.rst
```diff
@@ -167,22 +165,21 @@ and set ``index-url`` to the secure internal repo e.g.:
 Migrate to project-wide pip config
 ----------------------------------
 
-``inmanta-core 11.0.0`` introduced new options to configure pip settings for the whole project in a centralized way.
-For detailed information, see :ref:`here<specify_location_pip>`. The following code sample can be used as a baseline
-in the ``project.yml`` file:
+When upgrading to ``inmanta-service-orchestrator 7.0.0`` or ``inmanta 2024.0``, this section can be used as
```
sanderr, line 168: Passive voice alert: "This section is a migration guide for ..."

sanderr (changes requested):

docs/reference/projectyml.rst
```diff
@@ -70,9 +70,10 @@ Some of these options are detailed below:
 pip.use-system-config
 """""""""""""""""""""
 
-This option determines the isolation level of the project's pip config. When false (the default), any pip config set on the system
-(e.g. through environment variables or pip config files) are ignored and pip will only look for packages
-in the index(es) defined in the project.yml, when true, the orchestrator will extend the system's pip configuration rather than replace it.
+This option determines the isolation level of the project's pip config. When false (the default), any pip config set on
+the system through pip config files is ignored, the ``PIP_INDEX_URL``, ``PIP_EXTRA_INDEX_URL`` and ``PIP_PRE``
+environment variables are ignored, and pip will only look for packages in the index(es) defined in the project.yml.
+When true, the orchestrator will extend the system's pip configuration rather than replace it.
```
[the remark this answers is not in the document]
Hugo-Inmanta: Rephrased it a bit but the wording still feels awkward to me

docs/administrators/operational_procedures.rst
sanderr: This is still a bit of an issue. You start with a statement saying "do x". Then you elaborate on how to "do x". And only then do you say that you may do it differently under certain conditions. The overall meaning isn't all that clear I think.

docs/reference/projectyml.rst
```diff
@@ -73,7 +73,8 @@ pip.use-system-config
 This option determines the isolation level of the project's pip config. When false (the default), any pip config set on
 the system through pip config files is ignored, the ``PIP_INDEX_URL``, ``PIP_EXTRA_INDEX_URL`` and ``PIP_PRE``
 environment variables are ignored, and pip will only look for packages in the index(es) defined in the project.yml.
-When true, the orchestrator will extend the system's pip configuration rather than replace it.
+When true, the orchestrator will use the system's pip configuration for the pip-related settings not explicitly
```
sanderr, line 76: > Rephrased it a bit but the wording still feels awkward to me

I think it's better now. To address the final awkwardness, perhaps this?
```suggestion
When true, the orchestrator will use the system's pip configuration for the pip-related settings except when explicitly
```

sanderr (changes requested):

docs/administrators/operational_procedures.rst
Hugo-Inmanta: Rephrased it in a clearer way

docs/reference/projectyml.rst
Hugo-Inmanta: @sanderr wdyt ?

docs/administrators/operational_procedures.rst
sanderr: Big improvement imo.

sanderr (changes requested):
Looks good to me now. Only requesting changes because of the open thread that you started about the agent paragraph.

docs/reference/projectyml.rst
sanderr: @wouterdb do you have an opinion?

wouterdb: It's a detail, correct, obvious but important, I have no strong opinion so I would leave it in.

sanderr (approved):

PR #6650, opened 2023-10-20
Make the typing explicit

bartv:
# Description

Because TDiscovered is fully abstract it is not possible to correct serialize it with pydantic v2. This PR makes it more concrete so we can handle it explicitly.

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/handler.py
+++ b/src/inmanta/agent/handler.py
@@ -34,7 +34,7 @@ from inmanta import const, data, protocol, resources
 from inmanta.agent import io
 from inmanta.agent.cache import AgentCache
 from inmanta.const import ParameterSource, ResourceState
-from inmanta.data.model import AttributeStateChange, DiscoveredResource, ResourceIdStr
+from inmanta.data.model import AttributeStateChange, DiscoveredResource, ResourceIdStr, BaseModel
 from inmanta.protocol import Result, json_encode
 from inmanta.stable_api import stable_api
 from inmanta.types import SimpleTypes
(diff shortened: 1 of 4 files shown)


src/inmanta/agent/handler.py
```diff
@@ -51,7 +51,7 @@
 # A resource present in the model that describes the resources that should be discovered
 TDiscovery = TypeVar("TDiscovery", bound=resources.DiscoveryResource)
 # The type of elements produced by the resource discovery process.
-TDiscovered = TypeVar("TDiscovered")
+TDiscovered = TypeVar("TDiscovered", bound=BaseModel)
```
wouterdb, line 54: Why not allow other serializable types? (We could now, if we wanted, pass on an actual resource instance here)

src/inmanta/agent/handler.py
sanderr, line 1: You'll also have to update the docstring of `DiscoveryHandler`.

src/inmanta/data/model.py
```diff
@@ -725,7 +725,7 @@ class DiscoveredResource(BaseModel):
     """
 
     discovered_resource_id: ResourceIdStr
-    values: JsonType
+    values: dict[str, object]
```
wouterdb, line 728: I would expect JsonType to be more correct no? 

(at least, it should be?)

sanderr (changes requested):

src/inmanta/data/model.py
bartv: See discussion of last week

src/inmanta/agent/handler.py
bartv: See discussion of last week

bartv: It is at least Dict[str, Serializable] 

So why not provide already some structure. Otherwise the handler needs to do the dict() in case pydantic is used. This seemed to be the conclusion last week

sanderr (approved):

src/inmanta/agent/handler.py
wouterdb: No, dict() is done by our serializer. 

But indeed, it would be interesting to have a Serializable type

wouterdb (approved):

src/inmanta/agent/handler.py
bartv: That is what I mean with the implicit stuff. DiscoveredResource is a pydantic object and in strict mode it will just reject the assignment of the value so it never gets to the serializer

PR #6655, opened 2023-10-23
Issue/6516 - pip config

### LINKED ISSUE #6516: pip config: compiler side
Implement compiler side of the [pip config design](https://github.com/inmanta/designs/blob/master/modules/pip_config.md). Do not include changelog entries but if new breaking changes come to light, communicate them at #6518.

This requires meticulously reviewing the install and update path for both v1 and v2 modules

wouterdb:
# Description

Added pip config

closes #6516 

## Questions for the reviewers are attached! 

Please focus on the questions first, if you don't want to plow through. 

## TODO for me:
- [x] ensure we don't break any stable api
- [x] verify the pip install code paths are all cleared
- [x] de-duplicate and centralize pip related code 
- [x] ModuleV2Source should produce clean warnings, not unclear exceptions
- [x] Add warning to the pip type package source
- [x] verify module tool still works outside of projects  (and HOW)
    - I'm quite sure it now follows the existing behavior or always using the system config

# Changes wrt to design

## Every project will break

This makes EVERY project file with no `pip` section unusable as there is no default index and no default use system config! 

This also means pytest-inmanta **must** be updated

# Self Check:

- [x] Changelog entry


src/inmanta/env.py
```diff
@@ -314,11 +316,189 @@ class PipUpgradeStrategy(enum.Enum):
...
+        return self.index_url or self.use_system_config
+
+    def assert_has_source(self, reason: str) -> None:
+        if not self.has_source():
+            # TODO: Is this the correct exception type here?
+            # It is consistent with was there before, but is it Correct
+            # TODO: import cycle
+            from inmanta.module import InvalidModuleException
+
+            raise InvalidModuleException(
```
wouterdb, line 360: What exception should I raise here?

This is surprisingly impactful.

I don't think this is ideal. 

I could go for `env.PackageNotFound`

I could make a new one, two wrap at the appropriate places? (if so, what is its parent?)

src/inmanta/env.py
```diff
@@ -314,11 +316,189 @@ class PipUpgradeStrategy(enum.Enum):
...
+                f"indexes to the project config file. e.g. to add PyPi as a module source, add the following "
+                "to `project.yml`:"
+                "\npip:"
+                "\n  index_url:"
+                "\n    - https://pypi.org/simple"
+                "\nAnother option is to set `pip.use_system_config = true` to use the system's pip config."
+            )
+
+
 class PipCommandBuilder:
```
wouterdb, line 371: No longer the perfect name, but I'm afraid to rename it.
What should I do here? 
- Rename it
- Make a new superclass e.g. `PipHandler`
- Move this code elsewhere (into the config object or the venv object, but I don't feel either is a good choice)

src/inmanta/env.py
```diff
@@ -314,11 +316,189 @@ class PipUpgradeStrategy(enum.Enum):
...
     """
     Class used to compose pip commands.
     """
 
+    @classmethod
+    def run_pip_install_command_from_config(
```
wouterdb, line 377: Copy pasted into place, to be de-duplicated

tests/conftest.py
```diff
@@ -1230,9 +1234,12 @@ def setup_for_snippet_external(
         install_mode: Optional[InstallMode] = None,
         relation_precedence_rules: Optional[List[RelationPrecedenceRule]] = None,
         use_pip_config_file: bool = False,
+        index_url: Optional[str] = None,
```
wouterdb, line 1237: We could make this backward compatible be translating the old options into the new ones, I don't know if we have to?  (I did and then removed it to force a cleanup of all testcases)

src/inmanta/env.py
Hugo-Inmanta: Any specific reason why you are afraid to rename it ?

src/inmanta/module.py
```diff
@@ -1671,7 +1687,16 @@ def get_index_urls(self) -> List[str]:
 
         # This ensures no duplicates are returned and insertion order is preserved.
         # i.e. the left-most index will be passed to pip as --index-url and the others as --extra-index-url
-        return list({value: None for value in itertools.chain(self.pip.index_urls, index_urls_deprecated_option)})
+        return list(
+            {
+                value: None
+                for value in itertools.chain(
+                    [self.pip.index_url] if self.pip.index_url is not None else [],
+                    self.pip.index_url or [],
```
Hugo-Inmanta, line 1695:
```suggestion

```

Hugo-Inmanta (commented):
> => there is not option to disable --pre so if pip.use-system-config=true both False and None mean we use the system config
> Defaults to False. When false and pip.use-system-config=true we follow the system config.

I think this is in line with the design ?

src/inmanta/env.py
wouterdb: It has not stable API annotation, so perhaps I should just rename it

src/inmanta/module.py
wouterdb: good catch!

src/inmanta/env.py
arnaudsjs: `PackageNotFound` sounds good to me.

src/inmanta/env.py
arnaudsjs: Why did we add the `run_*` method to this class. This class had one very clearly responsibility. With this change we move away from that. I am not convinced that's a good idea.

tests/conftest.py
arnaudsjs: I like it like this.

arnaudsjs (changes requested):

src/inmanta/env.py
arnaudsjs: Don't these `run_` method belong in the `VirtualEnv` class? Because they act on a Python environment.

src/inmanta/module.py
wouterdb: method removed

src/inmanta/env.py
arnaudsjs: I see why you did it and I think it makes sense. The only thing that still annoys me at the moment is that the `PythonEnvironment` class has a `_run_pip_install_command` method that does many thing that the new `run_*` methods in the `PipCommandBuilder` class do. Both take care of the logging, etc. So this more or less violates the objective of 1. Are you planning to move this code from the `PythonEnvironment` class to the `PipCommandBuilder` class as well?

src/inmanta/module.py
```diff
@@ -1534,21 +1528,53 @@ def __str__(self) -> str:
...
+        extra = "ignore"
+
+    @pydantic.root_validator(pre=True)
+    def __alert_extra_field__(cls, values: dict[str, object]) -> dict[str, object]:
+        extra_fields = values.keys() - cls.__fields__.keys() - {v.alias for v in cls.__fields__.values()}
+
+        for extra_field in extra_fields:
+            # This is cfr adr 0000
+            warnings.warn(
+                ProjectConfigurationWarning(f"Found unexpected configuration value 'pip.{extra_field} in 'project.yaml'")
```
sanderr, line 1576:
```suggestion
                ProjectConfigurationWarning(f"Found unexpected configuration value 'pip.{extra_field} in 'project.yml'")
```

src/inmanta/moduletool.py
```diff
@@ -1026,9 +1025,14 @@ def install(self, editable: bool = False, path: Optional[str] = None) -> None:
         context of a project.
         """
 
+        # TODO: refine behavior
```
sanderr, line 1028: Perhaps an option to pass a project path to install in the context of a project?

tests/test_env.py
```diff
@@ -74,47 +75,6 @@ def test_venv_pyton_env_empty_string(tmpdir):
     assert e.value.args[0] == "The env_path cannot be an empty string."
 
 
-@pytest.mark.slowtest
-def test_basic_install(tmpdir):
```
sanderr, line 78: Why did this one have to go? Non PEP508 compliant?

tests/test_env.py
```diff
@@ -269,21 +192,32 @@ def test_process_env_install_from_index(
...
+        ),
+    )
     installed: Dict[str, version.Version] = env.process_env.get_installed_packages()
     assert package_name in installed
     if version is not None:
         assert installed[package_name] == version
 
 
 @pytest.mark.slowtest
+# TODO: depends on deciding exception structure
```
sanderr, line 208: TODO alert

tests/test_env.py
```diff
@@ -269,21 +192,32 @@ def test_process_env_install_from_index(
...
 def test_process_env_install_from_index_not_found(tmpvenv_active: Tuple[py.path.local, py.path.local]) -> None:
     """
     Attempt to install a package that does not exist from a pip index. Assert the appropriate error is raised.
     """
     with pytest.raises(env.PackageNotFound):
         # pass empty index list for security reasons (anyone could publish this package to PyPi)
```
sanderr, line 214: This comment used to refer to `index_urls=[]`, which is now gone. It should be moved to `use_system_config=False` (and updated) or removed.

tests/test_env.py
```diff
@@ -314,7 +253,11 @@ def test_process_env_install_from_source(
     package_name: str = "inmanta-module-minimalv2module"
     project_dir: str = os.path.join(modules_v2_dir, "minimalv2module")
     assert package_name not in env.process_env.get_installed_packages()
-    env.process_env.install_from_source([env.LocalPackagePath(path=project_dir, editable=editable)])
+    env.process_env.install_for_config(
+        requirements=[],
+        paths=[env.LocalPackagePath(path=project_dir, editable=editable)],
+        config=PipConfig(use_system_config=True),
```
sanderr, line 259: Is this required? I see that it's consistent with what we did before, but I don't think we need it.

tests/test_env.py
```diff
@@ -341,9 +284,16 @@ def test_active_env_get_module_file(
     if package_name.startswith(module.ModuleV2.PKG_NAME_PREFIX):
         module_name = "inmanta_plugins." + package_name[len(module.ModuleV2.PKG_NAME_PREFIX) :].replace("-", "_")
         index = str(local_module_package_index)
+        pip_config = PipConfig(
+            index_url=index,
+            use_system_config=True,  # we need an upstream for some packages
```
sanderr, line 289: You set both an index and the system config. Is this a mistake?

tests/test_env.py
```diff
@@ -385,7 +335,9 @@ def test_active_env_get_module_file_editable_namespace_package(
 
     assert env.ActiveEnv.get_module_file(module_name) is None
     project_dir: str = os.path.join(modules_v2_dir, "minimalv2module")
-    env.process_env.install_from_source([env.LocalPackagePath(path=project_dir, editable=True)])
+    env.process_env.install_for_config(
+        requirements=[], paths=[env.LocalPackagePath(path=project_dir, editable=True)], config=PipConfig(use_system_config=True)
```
sanderr, line 339: system config required here?

tests/test_env.py
```diff
@@ -429,7 +381,11 @@ def create_install_package(name: str, version: version.Version, requirements: Li
 build-backend = "setuptools.build_meta"
                 """.strip()
             )
-        env.process_env.install_from_source([env.LocalPackagePath(path=str(tmpdir), editable=False)])
+        env.process_env.install_for_config(
+            requirements=[],
+            paths=[env.LocalPackagePath(path=str(tmpdir), editable=False)],
+            config=PipConfig(use_system_config=True),
```
sanderr, line 387: Required? I'd think setuptools and wheel would already be installed.

tests/test_env.py
```diff
@@ -585,9 +546,12 @@ def _assert_install(requirement: str, installed: bool) -> None:
             assert tmpvenv_active_inherit.are_installed(requirements=[r]) == installed
 
     _assert_install("inmanta-module-elaboratev2module==1.2.3", installed=False)
-    tmpvenv_active_inherit.install_from_index(
+    tmpvenv_active_inherit.install_for_config(
         requirements=[Requirement.parse("inmanta-module-elaboratev2module==1.2.3")],
-        index_urls=[local_module_package_index],
+        config=PipConfig(
+            index_url=local_module_package_index,
+            use_system_config=True,  # we need an upstream for some packages
```
sanderr, line 553: Required?

tests/test_env.py
```diff
@@ -659,7 +606,13 @@ def test_are_installed_dependency_cycle_on_extra(tmpdir, tmpvenv_active_inherit:
     )
 
     requirements = [Requirement.parse("pkg[optional-pkg]")]
-    tmpvenv_active_inherit.install_from_index(requirements=requirements, index_urls=[pip_index.url])
+    tmpvenv_active_inherit.install_for_config(
+        requirements=requirements,
+        config=PipConfig(
+            index_url=pip_index.url,
+            use_system_config=True,  # we need an upstream for some packages
```
sanderr, line 613: This one I think we should really try to avoid. As far as I can see it's not required at the moment.

sanderr (changes requested):
I'm being quite nitpicky on using `use_system_config=True` only when it's really required. I hope I didn't flag too many false positives.

Hmm, I now see that in some places you have `use_system_config=True` because we don't allow 0 index urls. Perhaps we should allow an exception in our tests? Or use a dummy index in all these places. I think we should avoid deferring to PyPi at least where we install packages that don't exist on PyPi (e.g. our testing modules).

src/inmanta/env.py
```diff
@@ -589,31 +681,29 @@ def get_installed_packages(self, only_editable: bool = False) -> Dict[str, versi
...
-            allow_pre_releases=allow_pre_releases,
-            constraints_files=[*constraint_files],
             upgrade_strategy=upgrade_strategy,
-            use_pip_config=use_pip_config,
         )
 
     def install_from_source(
         self,
         paths: List[LocalPackagePath],
+        pip_config: PipConfig,
```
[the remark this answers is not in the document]
wouterdb: I made this compatible again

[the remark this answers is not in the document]
sanderr: Looks good. I think that ideally we still want to update that script so that it uses the project's pip config when appropriate (or that it uses whatever we end up with for `inmanta.moduletool`).

tests/test_env.py
```diff
@@ -202,10 +202,14 @@ def test_process_env_install_from_index(
     assert package_name in installed
     if version is not None:
         assert installed[package_name] == version
+    # legacy method
+    env.process_env.install_from_index(
+        [Requirement.parse(package_name + (f"=={version}" if version is not None else ""))],
+        use_pip_config=True,
+    )
```
sanderr, line 209: What are you actually testing here? The package is already installed at this point, isn't it? Same question for the `install_from_source` further down.

src/inmanta/agent/agent.py
```diff
@@ -31,6 +31,8 @@
 from logging import Logger
 from typing import Any, Awaitable, Callable, Dict, Iterable, List, Optional, Sequence, Set, Tuple, Union, cast
 
+import pkg_resources
```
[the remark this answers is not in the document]
wouterdb: This would be a massive undertaking to change it everywhere.

src/inmanta/env.py
```diff
@@ -618,17 +743,41 @@ def install_from_source(
...
+        upgrade_strategy: PipUpgradeStrategy = PipUpgradeStrategy.ONLY_IF_NEEDED,
+        use_pip_config: Optional[bool] = False,
+    ) -> None:
+        """
+        Install requirements from a list of requirement strings. This method uses the Python package repositories
+        configured on the host.
+        :param requirements_list: List of requirement strings to install.
+        :param upgrade: Upgrade requirements to the latest compatible version.
+        :param upgrade_strategy: The upgrade strategy to use for requirements' dependencies.
+        :param use_pip_config: Whether the pip config file specified in the PIP_CONFIG_FILE env var should be used
```
wouterdb, line 772: This option was silently ignored: : https://github.com/inmanta/inmanta-core/blob/0e5eed689d1ba01fc02c2c070ccbd0a16171f248/src/inmanta/env.py#L506-L507

I.e. it was ignored if not index_url was set, which is not possible here.

tests/test_env.py
```diff
@@ -74,73 +75,6 @@ def test_venv_pyton_env_empty_string(tmpdir):
...
-        else:
-            return compose(*args[:3], [], *args[4:], **kwargs)
-
-    monkeypatch.setattr(env.PipCommandBuilder, "compose_install_command", mock_compose)
-
-    with pytest.raises(Exception):
-        venv.install_from_list([package_name])
-
-    log_sequence = LogSequence(caplog)
-    log_sequence.contains("inmanta.env", logging.INFO, f"requirements:\n{package_name}")
```
wouterdb, line 141: This logging no longer happens as we flatten all requirements to to command line now, so this is no longer required I believe

src/inmanta/env.py
wouterdb: I did

src/inmanta/env.py
```diff
@@ -424,94 +389,96 @@ def compose_list_command(
...
+            *(str(requirement) for requirement in requirements),
+            *chain.from_iterable(["-r", f] for f in requirements_files),
+            *chain.from_iterable(["-e", path.path] if path.editable else [path.path] for path in local_paths),
+        ]
+
+        # From where
+        index_args: list[str] = []
+        if config.index_url:
+            index_args.append("--index-url")
+            index_args.append(config.index_url)
```
wouterdb, line 429: If the config doesn't set index url and we are not using pip config, do I need to set `--no-index-url` ?

src/inmanta/env.py
```diff
@@ -900,7 +904,7 @@ def install_for_config(
         upgrade_strategy: PipUpgradeStrategy = PipUpgradeStrategy.ONLY_IF_NEEDED,
         paths: List[LocalPackagePath] = [],
     ) -> None:
-        if not upgrade and self.are_installed(requirements):
+        if (not upgrade and self.are_installed(requirements)) and not paths:
```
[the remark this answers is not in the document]
wouterdb: Good question. We did not take this into account before. 

And we never use them internally either. 

I don't know how much effort we should invest in this?

src/inmanta/env.py
sanderr: Yes

sanderr: Though in practice we always raise an error in that case, don't we?

tests/moduletool/test_update.py
```diff
@@ -213,8 +203,22 @@ def test_module_update_dependencies(
...
+        index_url=index.url,
+    )
+
     # install b-1.0.0 and c-1.0.0
-    process_env.install_from_index([Requirement.parse(req) for req in ("b==1.0.0", "c==1.0.0")], index_urls=[index.url])
+    process_env.install_from_index(
+        [Requirement.parse(req) for req in ("b==1.0.0", "c==1.0.0")],
+        config=PipConfig(
+            index_url=index.url,
+            use_system_config=True,  # we need an upstream for some packages
```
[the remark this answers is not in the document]
wouterdb: turns out to work

tests/server/test_compilerservice.py
```diff
@@ -1500,7 +1500,11 @@ def patch_get_protected_inmanta_packages():
     venv = PythonEnvironment(env_path=venv_path)
     assert name_protected_pkg not in venv.get_installed_packages()
     venv.install_from_index(
-        requirements=[pkg_resources.Requirement.parse(name_protected_pkg)], index_urls=[local_module_package_index]
+        requirements=[pkg_resources.Requirement.parse(name_protected_pkg)],
+        config=PipConfig(
+            index_url=local_module_package_index,
+            use_system_config=True,  # we need an upstream for some packages
```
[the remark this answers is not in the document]
wouterdb: turns out to work

tests/test_env.py
wouterdb: added back

tests/test_env.py
wouterdb: This test is a bit moot, as it will never even attempt to install, as there is not index set?

wouterdb: Updated

tests/test_env.py
wouterdb: reverted to test legacy method

tests/test_env.py
wouterdb: seems to be not required.

wouterdb: So the venv itself is isolated, as intended 

> Activates the venv created by the `tmpvenv` fixture within the currently running process. This venv is completely decoupled  from the active development venv.

But `inmanta-module-minimalv2module` still depends on core (which it somehow can get) but it can't get the dependencies of core.


I don't know how it worked before.

tests/test_env.py
wouterdb: not required indeed

tests/test_env.py
wouterdb: seems ok

tests/test_env.py
wouterdb: seems ok

tests/test_env.py
wouterdb: seems ok

tests/test_module_loader.py
```diff
@@ -559,7 +561,10 @@ def test_module_v2_load_installed_without_required(snippetcompiler_clean, local_
     snippetcompiler_clean.setup_for_snippet("", autostd=False)
     process_env.install_from_index(
         [InmantaModuleRequirement.parse("elaboratev2module").get_python_package_requirement()],
-        index_urls=[local_module_package_index],
+        config=PipConfig(
+            index_url=local_module_package_index,
+            use_system_config=True,  # we need an upstream for some packages
```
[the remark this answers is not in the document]
wouterdb: done

src/inmanta/env.py
wouterdb: not if you also set paths

wouterdb: added

src/inmanta/moduletool.py
wouterdb: This one I may reserve for a second PR, depending on how much pytest-inmanta is broken

arnaudsjs (changes requested):

src/inmanta/env.py
```diff
@@ -424,94 +390,118 @@ def compose_list_command(
...
+        index_args: list[str] = []
+        if config.index_url:
+            index_args.append("--index-url")
+            index_args.append(config.index_url)
+        elif not config.use_system_config:
+            # If the config doesn't set index url
+            # and we are not using system config,
+            # then we need to disable the index.
+            # This can only happen if paths is also set.
+            index_args.append("--no-index-url")
```
Hugo-Inmanta, line 459: I think this is the option:
```suggestion
            index_args.append("--no-index")
```

Hugo-Inmanta (changes requested):

tests/moduletool/test_install.py
```diff
@@ -1053,6 +1054,7 @@ def test_install_with_use_config_but_PIP_CONFIG_FILE_not_set(
 
 
 @pytest.mark.parametrize_any("install_mode", [None, InstallMode.release, InstallMode.prerelease, InstallMode.master])
+@pytest.mark.skip(reason="The install mode is no longer relevant here!")  # TODO REMOVE
```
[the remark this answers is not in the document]
wouterdb: I removed the test, as this assumption no longer holds

tests/test_module_loader.py
```diff
@@ -249,6 +249,7 @@ def test_load_module_v2_already_installed(
 
 
 @pytest.mark.parametrize("install", [True, False])
+# TODO: exception type has changed here
```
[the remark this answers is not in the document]
wouterdb: removed

src/inmanta/env.py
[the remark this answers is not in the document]
wouterdb: done

tests/test_env.py
wouterdb: @sanderr I have no clue how to handle this.

wouterdb: same issue on test_active_env_get_module_file_editable_namespace_package

arnaudsjs (approved):

src/inmanta/env.py
```diff
@@ -534,13 +484,12 @@ def run_pip_install_command_from_config(
             # If we don't use system config, unset env vars
             # setting this env_var to os.devnull disables the loading of all pip configuration files
 
+            # Drop all PIP env vars
+            for env_var in list(sub_env.keys()):
+                if env_var.startswith("PIP_"):
```
sanderr, line 489: I think this is a bit too strict. I thought we concluded we would only explicitly ignore the env vars that we know about?

From the design: "other pip-related environment variables are not overridden/ignored because inmanta is not aware of them."

I am open to discussion on this point, as I think there is something to say for both approaches. Mainly, ignoring all of them eases backwards compatibility when adding support for new options in the future. But on the other hand there is the fact that in this case any option that we don't support would require `use-system-config=true`, which I don't really like.

src/inmanta/env.py
```diff
@@ -687,6 +635,19 @@ def install_for_config(
...
+
+        :param requirements: which requirements to install
+        :param paths: which paths to install
+        :param config: the pip config to use
+        :param constraint_files: pass along the following constraint files
+        :param upgrade: make pip do an upgrade
+        :param upgrade_strategy: what upgrade strategy to use
+
+        limitation:
+         - When upgrade is false, we check if the requirements are installed, but we don't take the constraints into account.
```
sanderr, line 649: I don't think this is entirely accurate / ambiguous. If I understand correctly:
1. It applies only to the pre-install check (which may skip the install), while "don't take the constraints into account" seems to imply we just ignore the constraints when upgrade is false.
2. This is only true for the `ActiveEnv` subclass.

Therefore I think it would be more appropriate to mention that we "don't guarantee" rather than "don't take into account". This would resolve both issues with this statement.

```suggestion
         - When upgrade is false, if requirements are already installed constraints from constraint files may not be verified.
```

src/inmanta/env.py
```diff
@@ -729,7 +690,7 @@ def install_from_index(
                 index_url=index_url,
                 extra_index_url=extra_index_url,
                 pre=allow_pre_releases,
-                use_system_config=use_pip_config or True,
+                use_system_config=use_pip_config if use_pip_config is not None else True,
```
sanderr, line 693: Huh, that was an interestingly redundant (and bugged) statement.

src/inmanta/env.py
```diff
@@ -769,15 +730,16 @@ def install_from_list(
         :param requirements_list: List of requirement strings to install.
         :param upgrade: Upgrade requirements to the latest compatible version.
         :param upgrade_strategy: The upgrade strategy to use for requirements' dependencies.
-        :param use_pip_config: Whether the pip config file specified in the PIP_CONFIG_FILE env var should be used
+        :param use_pip_config: ignored
 
         This method provides backward compatibility with ISO6
+        use_pip_config was ignored on ISO6 and it still is
```
sanderr, line 736: This method is otherwise deprecated, right?

sanderr: As in, its only purpose is to present a backward compatibility layer.

src/inmanta/module.py
```diff
@@ -680,6 +680,12 @@ def _get_module_name(self, module_spec: List[InmantaModuleRequirement]) -> str:
 
 @stable_api
 class ModuleV2Source(ModuleSource["ModuleV2"]):
+    def __init__(self, urls: List[str] = []) -> None:
```
sanderr, line 683: Unsafe default here.

sanderr (changes requested):

src/inmanta/module.py
Hugo-Inmanta: This is fine as long as the parameter is not mutated right ?

Hugo-Inmanta (changes requested):

src/inmanta/module.py
sanderr: Yes, but
1. it should then be typed as `Sequence` so that it would raise a type error if you try
2. I would not rely on it not being mutated. e.g. `pytest-inmanta` does some unorthodox things, it would not surprise me if it toys with this field.
3. I think it's still considered bad practice, though if 1 is met I'm not sure where I personally stand on the issue.

wouterdb: I made it sequence.

src/inmanta/module.py
```diff
@@ -1534,21 +1535,51 @@ def __str__(self) -> str:
...
+    :param extra_index_url:  additional pip index urls for this project. This is generally only
         recommended if all configured indexes are under full control of the end user to protect against dependency
         confusion attacks. See the `pip install documentation <https://pip.pypa.io/en/stable/cli/pip_install/>`_ and
         `PEP 708 (draft) <https://peps.python.org/pep-0708/>`_ for more information.
+    :param pre:  allow pre-releases when installing Python packages, i.e. pip --pre.
+        If null, behaves like false unless pip.use-system-config=true, in which case system config is respected.
+    :param use_system_config: defaults to false.
+        When true, sets the pip's index url, extra index urls and pre according to the respective settings outlined above
+            but otherwise respect any pip environment variables and/or config in the pip config file,
+            including any extra-index-urls.
```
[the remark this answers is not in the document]
wouterdb: done

src/inmanta/module.py
wouterdb: I agree it is not ideal, but this parameter is not in fact used any more. 
It is purely there to maintain api stability.

sanderr: I know, but pytest-inmanta(/-lsm) doesn't use it either then?

src/inmanta/env.py
sanderr: Leaving this unresolved because I linked to it from another PR, and GitHub doesn't like links to resolved conversations.

sanderr (approved):

Hugo-Inmanta (approved):

sanderr (approved):

PR #6656, opened 2023-10-23
update docs dictpath

FloLey:
# Description

-reworked the wrong imports in the docs
-moved the doc about dict-path away from lsm.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/model_developers.rst
+++ b/docs/model_developers.rst
@@ -19,3 +19,4 @@ Model developer documentation
     model_developers/model_design.rst
     model_developers/resource_sets.rst
     model_developers/unmanaged_resources.rst
+    model_developers/dictpath.rst
--- a/docs/model_developers/dictpath.rst
+++ b/docs/model_developers/dictpath.rst
@@ -45,7 +45,7 @@ Example
 
 .. code-block:: python
 
-    from inmanta_lsm import dict_path
+    from inmanta.util import dict_path
 
     container = {
         "a": "b",
--- a/changelogs/unreleased/6631-update-doc-dictpah.yml
+++ b/changelogs/unreleased/6631-update-doc-dictpah.yml
@@ -1,0 +1,6 @@
+---
+description: Update the dict_path docs (imports + location)
+issue-nr: 6631
+issue-repo: inmanta-core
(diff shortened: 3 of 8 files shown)


FloLey:
@arnaudsjs I wasn't completely sure where to put the dictpath doc. wdyt about the new location where I put it now?

docs/model_developers.rst
```diff
@@ -19,3 +19,4 @@ Model developer documentation
     model_developers/model_design.rst
     model_developers/resource_sets.rst
     model_developers/unmanaged_resources.rst
+    model_developers/dictpath.rst
```
arnaudsjs, line 22: Why did you add this line here. This documentation is already referenced from the `Inmanta Lifecycle Service Manager` section. Or did you intend to move it because it's now part of inmanta-core and usable in a broader scope than only LSM?

arnaudsjs (changes requested):

docs/model_developers.rst
FloLey: I moved it( [docs/lsm/allocation/dictpath.rst → docs/model_developers/dictpath.rst](https://github.com/inmanta/inmanta-core/pull/6656/files#diff-0b16210b9d9d3f9a439fefe95e294e3f834d54317e3e161777555e86e380ad79) )

indeed becauuse it now is usable in a broader scope. Isn't that a good idea?

docs/model_developers/dictpath.rst
arnaudsjs, line 1: The `autoclass` and `autofunction` statements still use the old import path.

docs/model_developers.rst
arnaudsjs: No. The move is fine. I just somehow overlooked the fact that the file was moved.

arnaudsjs (changes requested):

docs/model_developers.rst
arnaudsjs: I am not sure where is should come either. It's either here or we need to introduce a new section for this (like `utilities`). I am fine with both options.

arnaudsjs (approved):

PR #6669, opened 2023-10-24
Issue/549 license extension fails to load when lsm extension is disabled

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/inmanta-license/issues/549

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/549-improve-docstrings.yml
+++ b/changelogs/unreleased/549-improve-docstrings.yml
@@ -1,4 +1,3 @@
 description: Fix typos in docstrings.
-issue-nr: 549
 change-type: patch
 destination-branches: [master, iso6]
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/549-improve-docstrings.yml
```diff
@@ -0,0 +1,4 @@
+description: Fix typos in docstrings.
+issue-nr: 549
+change-type: patch
+destination-branches: [master, iso6]
```
arnaudsjs, line 4: I don't really see how these typos are linked to the referenced issue, but the issue-repo was missing.
```suggestion
description: Fix typos in docstrings.
issue-nr: 549
issue-repo: inmanta-license
change-type: patch
destination-branches: [master, iso6]
```

arnaudsjs (changes requested):

changelogs/unreleased/549-improve-docstrings.yml
Hugo-Inmanta: Ah, good point actually. Not linked at all apart from the fact that I fixed them while working on the referenced issue. I think it makes more sense to just remove the `issue-nr`.

arnaudsjs (approved):

PR #6670, opened 2023-10-24
Fix exception message

bartv:
# Description

The exception message used the wrong attribute on the exception. This has never worked before.

I also added the url it tries to make the logs more clear.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/fix-exception-typo.yml
+++ b/changelogs/unreleased/fix-exception-typo.yml
@@ -2,4 +2,4 @@ description: Fix jwt config error message to use the correct attribute and provi
 change-type: patch
 destination-branches: [master, iso6]
 sections:
-  feature: "{{description}}"
+  bugfix: "{{description}}"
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/fix-exception-typo.yml
```diff
@@ -0,0 +1,5 @@
+description: Fix jwt config error message to use the correct attribute and provide more context
+change-type: patch
+destination-branches: [master, iso6]
+sections:
+  feature: "{{description}}"
```
sanderr, line 5:
```suggestion
  bugfix: "{{description}}"
```
Right? Otherwise at least `minor-improvement` rather than `feature`.

sanderr (changes requested):

changelogs/unreleased/fix-exception-typo.yml
bartv: indeed

PR #6677, opened 2023-10-25
accept the password-hashing-algorithm adr

FloLey:
# Description

accept the password-hashing-algorithm adr

# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/adr/0003-password-hashing-algorithm.md
+++ b/docs/adr/0003-password-hashing-algorithm.md
@@ -1,7 +1,7 @@
 # Password hashing algorithm
 
 * Status: accepted
-* Date: 2023-10-25
+* Date: 2023-03-03
 
 
 ## Context and Problem Statement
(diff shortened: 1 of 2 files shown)


docs/adr/0003-password-hashing-algorithm.md
```diff
@@ -1,7 +1,7 @@
 # Password hashing algorithm
 
-* Status: proposed
-* Date: 2022-12-23
+* Status: accepted
+* Date: 2023-10-25
```
sanderr, line 4:
```suggestion
* Date: 2023-03-03
```
If I understand correctly (not 100% certain about that) this was actually decided when the PR was merged, we just forgot to mark it as accepted. If that is the case I would use that date.

sanderr (changes requested):

sanderr (approved):

Issue #6629, opened 2023-10-17
f-strings do not trim whitespace


bartv:
The following code does not work:

```
data = 10
std::print(f"-> { data }")
```

```
Error: variable  data  not found (reported in std::print(Format(-> { data })) (./test.cf:5))
```

The lexer for these strings needs to extract the proper variable names!

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/6629?planning-poker) with Zenhub @arnaudsjs @FloLey @Hugo-Inmanta @sanderr

PR #6679, opened 2023-10-25
Issue/6629 f strings do not trim whitespace

Hugo-Inmanta:
# Description

Fix bug in f-strings where surrounding a variable with spaces would result in a variable not found error 

closes https://github.com/inmanta/inmanta-core/issues/6629

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/6629-f-strings-do-not-trim-whitespaces.yml
+++ b/changelogs/unreleased/6629-f-strings-do-not-trim-whitespaces.yml
@@ -1,0 +1,4 @@
+description: Fix bug in f-strings not working when whitespaces surround the variable.
+issue-nr: 6629
+change-type: patch
+destination-branches: [master, iso6]
--- a/changelogs/unreleased/6629-f-strings-do-not-trim-whitespaces.yml
+++ b/changelogs/unreleased/6629-f-strings-do-not-trim-whitespaces.yml
@@ -2,3 +2,5 @@ description: Fix bug in f-strings not working when whitespaces surround the vari
 issue-nr: 6629
 change-type: patch
 destination-branches: [master, iso6]
+sections:
+  bugfix: "{{description}}"
--- a/src/inmanta/ast/statements/assign.py
+++ b/src/inmanta/ast/statements/assign.py
@@ -608,7 +608,7 @@ class FStringFormatter(Formatter):
         Overrides Formatter.get_field. Composite variable names are expected to be resolved at this point and can be
         retrieved by their full name.
         """
-        return (kwds[key.strip()], key)
+        return (kwds[key], key)
 
 
 class StringFormatV2(FormattedString):
@@ -628,7 +628,7 @@ class StringFormatV2(FormattedString):
         """
         only_refs: abc.Sequence["Reference"] = [k for (k, _) in variables]
         super().__init__(format_string, only_refs)
-        self._variables = only_refs
+        self._variables = {ref:full_name for (ref, full_name) in variables}
 
     def execute(self, requires: typing.Dict[object, object], resolver: Resolver, queue: QueueScheduler) -> object:
         super().execute(requires, resolver, queue)
@@ -637,13 +637,13 @@ class StringFormatV2(FormattedString):
         # We can't cache the formatter because it has no ability to cache the parsed string
 
         kwargs = {}
-        for _var in self._variables:
+        for _var, full_name in self._variables.items():
             value = _var.execute(requires, resolver, queue)
             if isinstance(value, Unknown):
                 return Unknown(self)
             if isinstance(value, float) and (value - int(value)) == 0:
                 value = int(value)
-            kwargs[_var.full_name] = value
+            kwargs[full_name] = value
 
         result_string = formatter.vformat(self._format_string, args=[], kwargs=kwargs)
 
--- a/src/inmanta/ast/statements/assign.py
+++ b/src/inmanta/ast/statements/assign.py
@@ -622,9 +622,9 @@ class StringFormatV2(FormattedString):
     def __init__(self, format_string: str, variables: abc.Sequence[typing.Tuple["Reference", str]]) -> None:
         """
         :param format_string: The string on which to perform substitution
-        :param variables: Sequence of tuples each holding a reference to a variable to substitute in the format_string
-            and the name of this variable. This name is not used in practice since it's already present in the reference's
-            full_name attribute.
+        :param variables: Sequence of tuples each holding a clean reference (i.e. stripped of eventual whitespaces ) to a
+         variable to substitute in the format_string and the raw full name of this variable (i.e. including eventual
+         whitespaces).
         """
         only_refs: abc.Sequence["Reference"] = [k for (k, _) in variables]
         super().__init__(format_string, only_refs)
--- a/src/inmanta/ast/statements/assign.py
+++ b/src/inmanta/ast/statements/assign.py
@@ -560,7 +560,7 @@ class FormattedString(ReferenceStatement):
     This class is an abstraction around a string containing references to variables.
     """
 
-    __slots__ = ("_format_string", "_variables")
+    __slots__ = "_format_string"
 
     def __init__(self, format_string: str, variables: abc.Sequence["Reference"]) -> None:
         super().__init__(variables)
@@ -578,7 +578,7 @@ class StringFormat(FormattedString):
     Create a new string by doing a string interpolation
     """
 
-    __slots__ = ()
+    __slots__ = "_variables"
 
     def __init__(self, format_string: str, variables: abc.Sequence[Tuple["Reference", str]]) -> None:
         super().__init__(format_string, [k for (k, _) in variables])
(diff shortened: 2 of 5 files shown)


src/inmanta/parser/plyInmantaParser.py
```diff
@@ -1027,9 +1035,9 @@ def convert_to_references(variables: List[Tuple[str, LocatableString]]) -> List[
                 ref.location = range_attr
                 ref.namespace = namespace
             # For a composite variable e.g. 'a.b.c', we only add the reference to the innermost attribute (e.g. 'c')
-            _vars.append((ref, match))
+            _vars.append((ref, " " * padding[0] + match + " " * padding[1]))
         else:
-            _vars.append((ref, match))
+            _vars.append((ref, " " * padding[0] + match + " " * padding[1]))
```
wouterdb, line 1040: Why add the padding back?

What if the original padding is not ' '  but e.g. '\t' ?

src/inmanta/ast/statements/assign.py
```diff
@@ -621,7 +621,7 @@ class StringFormatV2(FormattedString):
     def __init__(self, format_string: str, variables: abc.Sequence[typing.Tuple["Reference", str]]) -> None:
```
wouterdb, line 621: this could do with some documentation

wouterdb (changes requested):

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -67,7 +67,7 @@
 
 def t_FSTRING(t: lex.LexToken) -> lex.LexToken:  # noqa: N802
     r"f(\"([^\\\"\n]|\\.)*\")|f(\'([^\\\'\n]|\\.)*\')"
-    t.value = t.value[2:-1]
+    t.value = safe_decode(token=t, warning_message="Invalid escape sequence in f-string.", start=2, end=-1)
```
Hugo-Inmanta, line 70: This fixes what I believe to be another bug namely escape characters not being displayed properly in f-strings:
Given model:
```
data = 10
std::print(f"\n{ data }")
```
output before change:
```
\n10
```

output after change:
```

10
```

changelogs/unreleased/6629-f-strings-do-not-trim-whitespaces.yml
```diff
@@ -0,0 +1,4 @@
+description: Fix bug in f-strings not working when whitespaces surround the variable.
```
sanderr, line 1: Should have a bugfix entry

tests/compiler/test_strings.py
```diff
@@ -246,9 +247,14 @@ def test_fstring_relations(snippetcompiler, capsys):
...
 def test_fstring_numbering_logic():
+    """
+    Check that variable ranges in f-strings are correctly computed
+    """
     statements = parse_code(
         """
-std::print(f"---{s}{mm} - {sub.attr}")
+std::print(f"---{s}{mm} - {sub.attr} - {  padded  } - {  \tpadded.sub.attr   }")
+#                |   |           |           |                          |
+#                [[ [-[        [---[      [-----[                     [---[     <--- expected ranges
```
sanderr, line 257: Cool! Took me a while to interpret it but definitely useful. One potential improvement I think would be to make all ranges exclusive so that it's more visual: the range is whatever is strictly in between the brackets? I would still use `[ ... ]`, even though that clashes with the mathematical notation.

src/inmanta/parser/plyInmantaLex.py
sanderr: I think this is a good change. @wouterdb could you pay special attention to this?

sanderr (changes requested):
The change in the AST looks good to me. The parser change I'm less familiar with but I think it's good.

tests/compiler/test_strings.py
```diff
@@ -235,7 +236,7 @@ def test_fstring_relations(snippetcompiler, capsys):
 b = B(c=c)
 c = C()
 
-std::print(f"{a.b.c.n_c}")
```
wouterdb, line 238: This is still not supported and still produces a bit of a weird error:

```
std::print(f"{  a.b.c. n_c }")
```

src/inmanta/ast/statements/assign.py
```diff
@@ -608,17 +608,24 @@ def get_field(self, key: str, args: abc.Sequence[object], kwds: abc.Mapping[str,
...
     Create a new string by using python build in formatting
+
     """
 
     __slots__ = ()
 
     def __init__(self, format_string: str, variables: abc.Sequence[typing.Tuple["Reference", str]]) -> None:
+        """
+        :param format_string: The string on which to perform substitution
+        :param variables: Sequence of tuples each holding a reference to a variable to substitute in the format_string
```
wouterdb, line 625: This is still not clear to me:

1. what is responsible for removing whitespaces? (i.e. where can we expect them to exist?)
2. I would avoid trying to re-construct the matched strings (which we do here: `full_name` is built up in 'AttributeReference` , so any kind of garbage in there makes this fail with a weird error)


I would actually use the pair we get here to construct a lookup between `literal value with all kinds of garbage` and `clean reference` and I would strip all whitespace of all individual parts of the reference.

wouterdb (changes requested):

wouterdb (approved):

src/inmanta/ast/statements/assign.py
```diff
@@ -622,13 +622,13 @@ class StringFormatV2(FormattedString):
...
-            and the name of this variable. This name is not used in practice since it's already present in the reference's
-            full_name attribute.
+        :param variables: Sequence of tuples each holding a clean reference (i.e. stripped of eventual whitespaces ) to a
+         variable to substitute in the format_string and the raw full name of this variable (i.e. including eventual
+         whitespaces).
         """
         only_refs: abc.Sequence["Reference"] = [k for (k, _) in variables]
         super().__init__(format_string, only_refs)
-        self._variables = only_refs
+        self._variables = {ref: full_name for (ref, full_name) in variables}
```
sanderr, line 631:
```suggestion
        self._variables: abc.Mapping[Reference, str] = {ref: full_name for (ref, full_name) in variables}
```

Could you also move `_variables` to this class' `slots`? I don't think it belongs in the parent anymore.

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -997,10 +999,11 @@ def convert_to_references(variables: List[Tuple[str, LocatableString]]) -> List[
             (ex. LocatableString("a.b", range(a.b), lexpos, namespace))
 
         For f-strings:
-            - The string is the plain variable name without brackets (ex: 'a.b')
+            - The string is the plain variable name without brackets (ex: 'a.b') and including any potential whitespaces.
             - The LocatableString is the same as for regular string interpolation
-    :returns: A tuple where all LocatableString have been converted to Reference. The matching str holding the variable
-        name is left untouched
+    :returns: A tuple where all LocatableString have been converted to Reference. These references are cleaned up of any
```
sanderr, line 1004: Some more eventuals

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -1019,10 +1030,21 @@ def convert_to_references(variables: List[Tuple[str, LocatableString]]) -> List[
...
-                    var.location.file, var.location.lnr, char_offset, var.location.lnr, char_offset + len(attr)
+                    var.location.file,
+                    var.location.lnr,
+                    char_offset + left_spaces,
+                    var.location.lnr,
+                    char_offset + len(attr) - right_spaces,
+                )
+                attr_locatable_string: LocatableString = LocatableString(
+                    field_name_full_trim, range_attr, var.lexpos, var.namespace
                 )
```
sanderr, line 1047: I think this method could become a whole lot easier to read if you extract this part into a function with a clear name.

tests/compiler/test_strings.py
```diff
@@ -245,22 +247,58 @@ def test_fstring_relations(snippetcompiler, capsys):
...
+    Check that variable ranges in f-strings are correctly computed
+    """
     statements = parse_code(
         """
-std::print(f"---{s}{mm} - {sub.attr}")
+#        10        20        30        40        50        60        70        80
+#        |         |         |         |         |         |         |         |
+std::print(f"---{s}{mm} - {sub.attr} - {  padded  } - {  \tpadded.sub.attr   }")
+#                |   |           |           |                          |
+#               [-][--]       [----]     [------]                    [----]    <--- expected ranges
```
sanderr, line 265: Looks good!

sanderr (changes requested):

src/inmanta/ast/statements/assign.py
```diff
@@ -622,13 +622,13 @@ class StringFormatV2(FormattedString):
     def __init__(self, format_string: str, variables: abc.Sequence[typing.Tuple["Reference", str]]) -> None:
         """
         :param format_string: The string on which to perform substitution
-        :param variables: Sequence of tuples each holding a reference to a variable to substitute in the format_string
-            and the name of this variable. This name is not used in practice since it's already present in the reference's
-            full_name attribute.
+        :param variables: Sequence of tuples each holding a clean reference (i.e. stripped of eventual whitespaces ) to a
```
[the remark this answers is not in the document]
Hugo-Inmanta: Oof, nice catch, I was using it wrong in my pip config doc pr as well.

src/inmanta/ast/statements/assign.py
Hugo-Inmanta: Moved it down from `FormattedString` into  `StringFormat` and `StringFormatV2`. I'm not super familiar with slots usage, iirc it saves memory by not duplicating the field's name into each object ?
Any motivation for moving it down from the parent's class ?

sanderr: It's not used/declared in the class. Having it in slots is like declaring it's part of the interface of the class, but it really isn't in this case. So my motivation in this case is really more about readability than anything else.

Hugo-Inmanta: Makes sense :+1:

src/inmanta/ast/statements/assign.py
```diff
@@ -560,7 +560,7 @@ class FormattedString(ReferenceStatement):
     This class is an abstraction around a string containing references to variables.
     """
 
-    __slots__ = ("_format_string", "_variables")
+    __slots__ = "_format_string"
```
sanderr, line 563: I think this should be a tuple. Same for the other occurrences.

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -1002,52 +1001,55 @@ def convert_to_references(variables: List[Tuple[str, LocatableString]]) -> List[
...
-                    char_offset + left_spaces,
-                    var.location.lnr,
-                    char_offset + len(attr) - right_spaces,
-                )
-                attr_locatable_string: LocatableString = LocatableString(
-                    field_name_full_trim, range_attr, var.lexpos, var.namespace
-                )
+            offset = len(var_parts[0]) + 1
+            for attr in var_parts[1:]:
+                attr_locatable_string: LocatableString = normalize(attr, var, offset=offset)
```
sanderr, line 1048: Much better!

sanderr (changes requested):

sanderr (approved):

PR #6684, opened 2023-10-26
fix-issue-ast-cache-modules 

FloLey:
# Description

Resolved a bug (introduced here: https://github.com/inmanta/inmanta-core/commit/cc0d4718740a55b6a76b4a545c53188fde2068cd) where the AST cache was not invalidated after a module installation/update, leading to unrefreshed dependencies for the said module.

# Self Check:

- [x] Changelog entry


tests/moduletool/conftest.py
```diff
@@ -147,6 +162,9 @@ def modules_repo(git_modules_dir) -> str:
     add_file(mod12, "file", "test", "release version 4.0.0.dev0", version="4.0.0.dev0")
     add_file(mod12, "file", "test", "release version 4.0.0", version="4.0.0")
 
+    mod13 = make_module_simple(reporoot, "mod13", version="1.2.3", depends=[("mod11", "<4.2.0")])
+    update_requires(mod13, [["mod11"], ["<4.1.2"]], "add dependency", version="1.2.4")
```
arnaudsjs, line 166: I find the second argument hard to interpret, but maybe that also linked to the fact that `update_requires` doesn't have a docstring.

arnaudsjs (changes requested):

tests/moduletool/test_update.py
```diff
@@ -183,6 +183,45 @@ def assert_version_installed(module_name: str, version: str) -> None:
...
+    tmpdir: py.path.local,
+    modules_dir: str,
+    snippetcompiler_clean,
+    modules_repo: str,
+    install_mode: InstallMode,
+) -> None:
+    """
+    Assert that the ast cache is invalidated after a module is installed. This so that dependencies of
+    installed modules are updated as well.
+
```
Hugo-Inmanta, line 198: I'm not 100% sure this test is actually checking that the cache is invalidated i.e. I would expect this test to fail if we comment out the cache invalidation in `src/inmanta/moduletool.py`

Hugo-Inmanta (changes requested):

tests/moduletool/test_update.py
FloLey: Indeed, it should fail. good catch. i'll continue looking

src/inmanta/moduletool.py
```diff
@@ -489,11 +489,15 @@ def do_update(specs: "Dict[str, List[InmantaModuleRequirement]]", modules: List[
                     allow_pre_releases=my_project.install_mode != InstallMode.release,
                     use_pip_config=my_project.metadata.pip.use_config_file,
                 )
+                # Invalidate ast cache so that dependencies of installed modules can be updated as well
+                my_project.invalidate_state()
```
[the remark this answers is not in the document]
FloLey: I'm not sure I see how this can be done as the pip install command install all the modules at once. I don't see another solution than saving `project.modules` before and after the installation and looking for changes. Would this be ok?

tests/moduletool/test_update.py
FloLey: @arnaudsjs I don't think the cache invalidation is needed for v1 modules. When calling `my_project.collect_requirements()` the new requirements after each module update/install are updated and it manages to install everything

src/inmanta/moduletool.py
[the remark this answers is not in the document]
arnaudsjs: You are right. We should keep it like it was.

tests/moduletool/test_update.py
arnaudsjs: The cache invalidation is required for v1 modules, but the test scenario doesn't trigger the issue. The difference between this test case and the one for v2 modules is that this one adds an import statement for mod11 to mod13 both in the initially installed version of the module and the version installed after the update. The test case for v2 doesn't have the import statement in the initially installed version. The moduletool uses the import statements to determine the full set of modules that are required by the project. The idea behind the test is that only the new module version has the import statement and that the cache causes the moduletool to still see the version without the import statement, causing it not to update the version of the module dependency.

tests/moduletool/conftest.py
```diff
@@ -147,6 +162,9 @@ def modules_repo(git_modules_dir) -> str:
     add_file(mod12, "file", "test", "release version 4.0.0.dev0", version="4.0.0.dev0")
     add_file(mod12, "file", "test", "release version 4.0.0", version="4.0.0")
 
+    mod13 = make_module_simple(reporoot, "mod13", version="1.2.3")
+    add_requires(mod13, [("mod11", "<4.1.2")], commit_msg="add dependency", version="1.2.4")
```
arnaudsjs, line 166:
```suggestion
    add_requires(mod13, deps=[("mod11", "<4.1.2")], commit_msg="add dependency", version="1.2.4")
```

tests/moduletool/common.py
```diff
@@ -100,6 +101,64 @@ def add_file(modpath, file, content, msg, version=None, dev=False, tag=True):
...
+
+    with open(file_path, "r") as file:
+        data = yaml.load(file)
+
+    # Ensure 'requires' field exists and is a list
+    if "requires" not in data:
+        data["requires"] = []
+
+    # Prepare a dictionary to hold the latest version requirement for each module
+    requires_dict = {item.split()[0]: item for item in data["requires"]}
```
arnaudsjs, line 130: This is a bit more robust.
```suggestion
    requires_dict = {item.strip().split()[0]: item for item in data["requires"]}
```

arnaudsjs (changes requested):

arnaudsjs (approved):

tests/moduletool/conftest.py
```diff
@@ -95,11 +96,25 @@ def modules_repo(git_modules_dir) -> str:
...
+    +--------+-------------+----------+
+    |        |             | 4.2.0    |
+    +--------+-------------+----------+
     | mod12  |             | 3.2.1    |
     +--------+-------------+----------+
     |        |             | 4.0.0.dev0 |
     +--------+-------------+----------+
     |        |             | 4.0.0    |
+    +--------+-------------+----------+
+    | mod13  | mod11 <4.2.0| 1.2.3    |
```
Hugo-Inmanta, line 115: this dependency is not present anymore

Hugo-Inmanta (changes requested):

tests/moduletool/conftest.py
FloLey: indeed! I'll remove it!

PR #6685, opened 2023-10-26
Issue/6462 pydantic v2 poc v2 sander

sanderr:


setup.py
```diff
@@ -24,7 +24,6 @@
     # pip>=21.3 required for editable pyproject.toml + setup.cfg based install support
     "pip>=21.3",
     "ply~=3.0",
-    # lower bound because of pydantic/pydantic#5821
     "pydantic~=2.3",
```
bartv, line 27: In my venv I already used the latest version 2.4.2

bartv:
Is this now with strict or lax mode? I do think it is good that we are forced to get rid of these Any's

src/inmanta/data/model.py
```diff
@@ -702,7 +713,7 @@ class DiscoveredResource(BaseModel):
 
     @field_validator("discovered_resource_id")
     @classmethod
-    def discovered_resource_id_is_resource_id(cls, v: str) -> Optional[Any]:
+    def discovered_resource_id_is_resource_id(cls, v: str) -> str:
```
bartv, line 716: Why not ResourceIdStr?

src/inmanta/validation_type.py
```diff
@@ -72,20 +73,36 @@ def validate_type(
...
+    elif not callable(requested_type):
+        raise ValueError(
+            f"validate_type got validation parameters {validation_parameters}"
+            f" but validation type {fq_type_name} is not parametrized"
+        )
     else:
-        model = pydantic.create_model(fq_type_name, value=(t, ...))
+        parametrized_type = requested_type(**validation_parameters)
+
+    validation_type = pydantic.TypeAdapter(
```
bartv, line 102: I like. When I started I was not confident enough yet to do this.

bartv (approved):

sanderr:
> Is this now with strict or lax mode?

Still lax. I'm inclined to do strict (if we even want it) after this has been out in the wild for a few days.

>  I do think it is good that we are forced to get rid of these Any's

We aren't: I get rid of them whenever I encounter them. The pydantic docs, strangely, seem to advocate them: a lot of their examples use `Any` where I feel they should use `object` instead.

src/inmanta/data/model.py
sanderr: Shouldn't matter too much but it may present a clearer picture. Will do. That or what you proposed on Slack: make the validator an annotator on the `ResourceId` type.

src/inmanta/validation_type.py
sanderr: I had the benefit of not being very confident with `create_model` either

PR #6686, opened 2023-10-27
issue/6526-index-number-integer-decimal

### LINKED ISSUE #6526: index on number type differentiates between integer and decimal notation
# Actionables
*ISO7 only!*

- make number behave like float: make an explicit decimal point required for the number type. The `number(x)` cast should be verified to work so it can be used to assign int values to number types.
- create the `float` alias for `number` because it more clearly conveys what it means. Update documentation to use `float` rather than `number`.
- verify that `lsm` and `inmanta-lsm` can work with this new `float` type

# Motivation
An index on an attribute of type `number` only considers a number expressed respectively with and without `.0` to be diffferent, while these are just different representations of the same `number`. As an example, the following model prints `False`.

```
entity A:
    number x
end

implement A using std::none

index A(x)


std::print(A(x=0) == A(x=0.0))
```

FloLey:
# Description

- make number behave like float. The number(x) cast should be verified to work so it can be used to assign int values to number types.

- create the float alias for number because it more clearly conveys what it means. Update documentation to use float rather than number.

- add testcases for following scenario and fix it
```
entity A:
    number x
end

implement A using std::none

index A(x)


std::print(A(x=0) == A(x=0.0))
```
-  index lookup on float with int should work

closes #6526 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -127,6 +127,10 @@ Primitive types
 
 The basic primitive types are ``string``, ``float``, ``int`` or ``bool``. These basic types also support type casts:
 
+.. note::
+    For legacy reason the ``number`` type exists. It is just an alias for ``float``
+
+
 .. code-block:: inmanta
 
     assert = true
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -92,8 +92,8 @@ Literal values can be assigned to variables
 
 .. code-block:: inmanta
 
-    var1 = 1 # assign an integer, var1 contains now a number
-    var2 = 3.14 # assign a float, var2 also contains a number
+    var1 = 1 # assign an integer, var1 contains now an integer
+    var2 = 3.14 # assign a float, var2 now contains a float
     var3 = "This is a string" # var3 contains a string
     var4 = r"This is a raw string" # var4 contains a raw string
 
@@ -125,14 +125,14 @@ Literal values can be assigned to variables
 Primitive types
 ==============================
 
-The basic primitive types are ``string``, ``number``, ``int`` or ``bool``. These basic types also support type casts:
+The basic primitive types are ``string``, ``float``, ``int`` or ``bool``. These basic types also support type casts:
 
 .. code-block:: inmanta
 
     assert = true
     assert = int("1") == 1
-    assert = number("1.2") == 1.2
-    assert = number(true) == 1
+    assert = float("1.2") == 1.2
+    assert = float(true) == 1
     assert = bool(1.2) == true
     assert = bool(0) == false
     assert = bool(null) == false
@@ -164,7 +164,7 @@ For example
     typedef mac_addr as string matching /([0-9a-fA-F]{2})(:[0-9a-fA-F]{2}){5}$/
 
 
-Lists of primitive types are also primitive types: ``string[]``, ``number[]``, ``bool[]`` or ``mac_addr[]``
+Lists of primitive types are also primitive types: ``string[]``, ``float[]``, ``bool[]`` or ``mac_addr[]``
 
 ``dict`` is the primitive type that represents a dictionary, with string keys. Dict values can be accessed using the ``[]`` operator. All members of a dict have to be set when the dict is constructed. e.g.
 
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -706,7 +706,7 @@ For indices on relations (instead of attributes) an alternative syntax can be us
     # a == b
 
 .. note::
-    The use of ``number`` as part of index properties is
+    The use of ``float`` (or ``number``) as part of index properties is
     generally discouraged. This is due to the reliance of index matching on precise equality,
     while floating-point numbers are represented with an inherent imprecision.
     If floating-point attributes are used in an index, it is crucial to handle arithmetic
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -705,6 +705,14 @@ For indices on relations (instead of attributes) an alternative syntax can be us
     b = vm1.files[path="/etc/passwd"]  # selector style index lookup
     # a == b
 
+.. note::
+    The use of ``number`` as part of index properties is
+    generally discouraged. This is due to the reliance of index matching on precise equality,
+    while floating-point numbers are represented with an inherent imprecision.
+    If floating-point attributes are used in an index, it is crucial to handle arithmetic
+    operations with caution to ensure the accuracy of the attribute values for index operations.
+
+
 
 For loop
 =========
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -381,7 +381,20 @@ class Entity(NamedType, WithComment):
         Update indexes based on the instance and the attribute that has
         been set
         """
-        attributes = {k: repr(v.get_value()) for (k, v) in instance.slots.items() if v.is_ready()}
+        # attributes = {k: repr(v.get_value()) for (k, v) in instance.slots.items() if v.is_ready()}
+
+        # Initialize an empty dictionary for attributes
+        attributes = {}
+
+        # Iterate through the items in instance.slots
+        for k, v in instance.slots.items():
+            # Check if the value is ready
+            if v.is_ready():
+                # Get the value and represent it as a string, then assign to the attributes dictionary
+                value_str = repr(v.get_value())
+                print(value_str)
+                attributes[k] = value_str
+
         # check if an index entry can be added
         for index_attributes in self.get_indices():
             index_ok = True
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -381,8 +381,8 @@ class Entity(NamedType, WithComment):
         Update indexes based on the instance and the attribute that has
         been set
         """
-        # check if an index entry can be added
         attributes = {k: repr(v.get_value()) for (k, v) in instance.slots.items() if v.is_ready()}
+        # check if an index entry can be added
         for index_attributes in self.get_indices():
             index_ok = True
             key = []
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -381,6 +381,7 @@ class Entity(NamedType, WithComment):
         Update indexes based on the instance and the attribute that has
         been set
         """
+        # check if an index entry can be added
         attributes = {k: repr(v.get_value()) for (k, v) in instance.slots.items() if v.is_ready()}
         for index_attributes in self.get_indices():
             index_ok = True
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -445,21 +445,6 @@ class Entity(NamedType, WithComment):
             ["%s=%s" % (k, repr(float(v) if isinstance(v, int) else v)) for (k, v) in sorted(params, key=lambda x: x[0])]
         )
 
-        # key = ", ".join(["%s=%s" % (k, repr(v)) for (k, v) in sorted(params, key=lambda x: x[0])])
-
-        # # Initialize an empty list to hold the key-value pairs
-        # key_value_pairs = []
-        #
-        # # Iterate over the sorted parameters
-        # for k, v in sorted(params, key=lambda x: x[0]):
-        #     # Format each key-value pair as a string
-        #     pair = "%s=%s" % (k, repr(v))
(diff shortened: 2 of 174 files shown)


Hugo-Inmanta (changes requested):

src/inmanta/ast/entity.py
```diff
@@ -428,7 +436,10 @@ def lookup_index(
                 stmt, self.get_full_name(), "No index defined on %s for this lookup: " % self.get_full_name() + str(params)
             )
 
-        key = ", ".join(["%s=%s" % (k, repr(v)) for (k, v) in sorted(params, key=lambda x: x[0])])
+        # Convert integers to floats in the key, similar to add_to_index
+        key = ", ".join(
+            ["%s=%s" % (k, str(float(v)) if isinstance(v, int) else repr(v)) for (k, v) in sorted(params, key=lambda x: x[0])]
```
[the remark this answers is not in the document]
FloLey: no reason, it does the same here I think. Why do you prefer `repr`?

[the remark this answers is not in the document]
sanderr: `repr` is definitely appropriate here: `str` is a human representation, `repr` is an unambiguous structured representation.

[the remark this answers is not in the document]
sanderr: Oh, we're talking about the `if` case only... There it matters less, but I'd still go for the semantically appropriate `repr`.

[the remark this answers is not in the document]
Hugo-Inmanta: And the consistency makes it slightly easier to read as well

Hugo-Inmanta (changes requested):
I might have missed it, but I think we are still missing a test for:
```
index lookup on float with int should work
```

FloLey:
> I might have missed it, but I think we are still missing a test for:
> 
> ```
> index lookup on float with int should work
> ```

It was tested for number and number is an alias for float so I considered it done but I changed it in those tests:
`test_int_as_index_for_number` and `test_int_as_index_for_number_2`. They now test for number and float.

Hugo-Inmanta:
This is what I had in mind regarding my previous comment with explicit index lookup (`y = A[x=1]`):
But maybe this is redundant with the existing tests, I'm not 100% sure. @sanderr any opinion on this?
```python
def test_lookup_on_float_with_int(snippetcompiler):
        f"""
entity A:
    float x
end

implement A using std::none

index A(x)

a = A(x=1.0)
y = A[x=1]
        """,
    )
    (_, scopes) = compiler.do_compile()
    root: Namespace = scopes.get_child("__config__")
    a = root.lookup("a").get_value()
    y = root.lookup("y").get_value()
    assert a is y
```

sanderr:
I agree that this should be possible and tested somehow. I haven't looked at this PR in-depth yet so I can't comment on whether it's already covered.

Hugo-Inmanta (approved):

tests/compiler/test_typing.py
```diff
@@ -103,31 +103,162 @@ def test_int_type_invalid(snippetcompiler):
...
+    number i = 0
+end
+
+implement Test using std::none
+
+Test(i = 42)
+Test(i = -42)
+Test()
+Test(i = 42.0)
+Test(i = false)
```
sanderr, line 119: I don't like this at all, and we'll probably change it soon. But I guess it's good to be alerted if we change it by accident at some point.

tests/compiler/test_typing.py
```diff
@@ -103,31 +103,162 @@ def test_int_type_invalid(snippetcompiler):
...
+
+
 def test_cast_to_number(snippetcompiler):
     snippetcompiler.setup_for_snippet(
         """
-x = number("42")
-x = 42
-y = number("42.0")
-y = 42.0
+a = number(21)
```
sanderr, line 128: Huh, this doesn't seem to correspond with my testing, see my other comment. Any clue how comes?

tests/compiler/test_typing.py
```diff
@@ -103,31 +103,162 @@ def test_int_type_invalid(snippetcompiler):
...
     u = root.lookup("u").get_value()
+    assert Number().validate(a)
+    assert Number().validate(b)
     assert Number().validate(x)
     assert Number().validate(y)
     assert Number().validate(z)
     assert Number().validate(u)
 
 
+def test_float_alias_number(snippetcompiler):
```
sanderr, line 158: Could you add some lines verifying that the two types are compatible? i.e. `test.my_float = test.my_int` and vice versa.

Also, `int <- float` and `float <- int` assignments would be useful. The former should raise an error while the latter should work and result in a float.

tests/compiler/test_typing.py
```diff
@@ -103,31 +103,162 @@ def test_int_type_invalid(snippetcompiler):
...
+def test_int_as_index_for_number(snippetcompiler, type):
+    snippetcompiler.setup_for_snippet(
+        f"""
+entity A:
+    {type} x
+end
+
+implement A using std::none
+
+test = A(x=0)
```
sanderr, line 213: What does this test? There are no assertions, no comparisons or no index lookups. There isn't even an index defined.

tests/compiler/test_typing.py
```diff
@@ -103,31 +103,162 @@ def test_int_type_invalid(snippetcompiler):
...
+
+index A(x)
+
+test = (A(x=0) == A(x=0.0))
+        """,
+    )
+    (_, scopes) = compiler.do_compile()
+    root: Namespace = scopes.get_child("__config__")
+    test = root.lookup("test").get_value()
+    assert test
```
sanderr, line 237: Works, but alternatively you could just add `test = true` in the model. I often use `assert` for the variable name. I typically find it a lot easier than to have go through the mess with the namespace lookup.

sanderr (changes requested):

src/inmanta/ast/type.py
```diff
@@ -265,6 +264,19 @@ def type_string_internal(self) -> str:
         return self.type_string()
 
 
+@stable_api
+class Float(Number):
+    """
+    This class is an alias for the Number class and represents a float in
+    the configuration model. On these numbers, the following operations are supported:
+
+    +, -, /, *
```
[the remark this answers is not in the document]
FloLey: I copy pasted the docstring of number :p I removed it at both places

tests/compiler/test_typing.py
FloLey: indeed it seems this doesn't work:
```
    entity Test:
        float i = 0
    end

    entity Test2:
        int i = 0
    end
    
    implement Test2 using std::none
    
    val1 = Test(i = 42.0)
    val2 = Test2(i = 42)
    val1 = val2
```
So i'm a bit surprised, I'll look tomorrow

src/inmanta/ast/type.py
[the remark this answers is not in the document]
sanderr: I noticed, I have no clue how it got there.

tests/compiler/test_typing.py
FloLey: I think this is fixed now

src/inmanta/ast/type.py
```diff
@@ -265,6 +264,19 @@ def type_string_internal(self) -> str:
         return self.type_string()
 
 
+@stable_api
+class Float(Number):
```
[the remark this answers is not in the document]
FloLey: I created a PR here : https://github.com/inmanta/inmanta-core/pull/6722

tests/compiler/test_typing.py
sanderr: Oh, it must be the `.0`: `42 == 42.0` even if `42 is not 42.0`

sanderr: So I'd use `42.1` or something for this test case.

src/inmanta/ast/entity.py
```diff
@@ -381,7 +381,15 @@ def add_to_index(self, instance: Instance) -> None:
...
-        attributes = {k: repr(v.get_value()) for (k, v) in instance.slots.items() if v.is_ready()}
+        attributes = {}
+        for k, v in instance.slots.items():
+            if v.is_ready():
+                value = v.get_value()
+                # Check if the value is an integer,
+                # and convert to float if necessary
+                if isinstance(value, int):
+                    value = float(value)
+                attributes[k] = repr(value)
```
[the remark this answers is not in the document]
FloLey: I was not able to make it work by changing it in Instance.set_attribute : the Integer.validate() failed for the attributes that changed from an int to a float. Am I missing something?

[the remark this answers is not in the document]
sanderr: Could you elaborate on this? I don't understand what doesn't work.

docs/language.rst
```diff
@@ -701,6 +705,14 @@ For indices on relations (instead of attributes) an alternative syntax can be us
     b = vm1.files[path="/etc/passwd"]  # selector style index lookup
     # a == b
 
+.. note::
+    The use of ``number`` as part of index properties is
```
sanderr, line 709:
```suggestion
    The use of ``float`` (or ``number``) as part of index properties is
```

src/inmanta/ast/entity.py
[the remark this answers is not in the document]
sanderr: Note to self: also affects other, similar changes.

tests/compiler/test_typing.py
```diff
@@ -143,18 +145,72 @@ def test_cast_to_number(snippetcompiler):
...
+        int i = 0
+    end
+
+    implement Test using std::none
+    implement Test2 using std::none
+
+    val1 = Test(i = 42.0)
+    val2 = Test2(i = 42)
+    val2.i = val1.i
+    val1.i = val2.i
```
sanderr, line 179: This is not a good test because the value is already set, therefore we fall into an execution flow where the validation is not called. I would expect this to fail.

tests/compiler/test_typing.py
```diff
@@ -143,18 +145,72 @@ def test_cast_to_number(snippetcompiler):
...
+
+
+def test_assign_int_to_float(snippetcompiler):
+    snippetcompiler.setup_for_snippet(
+        """
+    entity Test:
+        float i = 0
+    end
+    implement Test using std::none
+    Test(i = 42)
```
sanderr, line 208: Is this 42 or 42.0? I think that assertion should be part of the test, e.g. as an `isinstance` check.

tests/compiler/test_typing.py
```diff
@@ -143,18 +145,72 @@ def test_cast_to_number(snippetcompiler):
...
+    implement Test using std::none
+    implement Test2 using std::none
+
+    val1 = Test(i = 42.0)
+    val2 = Test2(i = 42)
+    val2.i = val1.i
+    val1.i = val2.i
+    """,
+    )
+    (_, scopes) = compiler.do_compile()
```
sanderr, line 182: I think this test could also use some isinstance checks.

tests/compiler/test_typing.py
sanderr: The problem is that you set up the failure scenario to raise a double set exception. Only, `42 == 42.0`, so `x = 42 x = 42.0` is perfectly valid (though admittedly a bit strange, and execution order would affect the actual type). Therefore even if the cast didn't work correctly (which it didn't) you would still get the correct answer. By making it `42.1`, when the bug is triggered, we get `42.1 == 42`, which is trivially wrong, therefore a more reliable assertion.

tests/compiler/test_typing.py
FloLey: I'll replace it with another test

tests/compiler/test_typing.py
```diff
@@ -159,27 +159,85 @@ def test_cast_to_number(snippetcompiler):
...
-    implement Test2 using std::none
+    implement Int using std::none
+    implement Float using std::none
 
-    val1 = Test(i = 42.0)
-    val2 = Test2(i = 42)
-    val2.i = val1.i
-    val1.i = val2.i
-    """,
+    f = Float(i={val})
```
sanderr, line 204: This part of the test is not really stable. If the `Int` constructor is executed first the compile fails before we ever execute this statement.

sanderr (changes requested):

sanderr (approved):

PR #6690, opened 2023-10-27
Add PreludeBasedFilteringQueryBuilder

arnaudsjs:
# Description

Add PreludeBasedFilteringQueryBuilder.

Part of inmanta/inmanta-lsm#1279

# Self Check:

- [x] Changelog entry


src/inmanta/data/dataview.py
```diff
@@ -1239,3 +1246,91 @@ def construct_dtos(self, records: Sequence[Record]) -> Sequence[dict[str, str]]:
             ).dict()
             for res in records
         ]
+
+
+class PreludeBasedFilteringQueryBuilder(SimpleQueryBuilder):
+    """
+    A query builder that applies any filters and the LIMIT statement to the prelude query rather than the outer query.
```
sanderr, line 1253:
```suggestion
    A query builder that applies any filters and the LIMIT statement to the prelude query rather than the outer query.
    The outer query may use the table name "prelude" to refer to the inner query.
```

sanderr (changes requested):

sanderr (approved):

Issue #6175, opened 2023-06-26
Add __version__ to core


sanderr:
Many projects inspect the version of the `inmanta-core` package. A non-exhaustive list:
- `vscode-inmanta`: `inmantals.server` and `tests/test_smoke.py`
- `pytest-inmanta-lsm`: tests
- `pytest-inmanta`: tests and inmanta/pytest-inmanta#396

Many Python packages use their root `__init__.py` to expose a `__version__` name that contains the current version. Considering how often we need it, I think it would be best to introduce this in `inmanta-core` so we don't have to rely on `pkg_resources.get_distribution` for this.

- [ ] add version
  - [ ] add a comment to both places where the version is to also update the other and/or use bumpversion 
- [ ] make sure it gets bumped correctly by bumpversion, release tool, merge tool, ...
   - [ ]  probably derives from bumpversion 
- [ ]  update uses of `pkg_resources.get_distribution` in core itself 
- [ ] Create a follow up ticket to use this value instead of  `pkg_resources.get_distribution` in other extensions/libraries

sanderr:
Discussed in planning meeting: approved

PR #6694, opened 2023-10-30
issue/6175 add __version__ to core

Hugo-Inmanta:
# Description

Many projects inspect the version of the `inmanta-core` package. A non-exhaustive list:
- `vscode-inmanta`: `inmantals.server` and `tests/test_smoke.py`
- `pytest-inmanta-lsm`: tests
- `pytest-inmanta`: tests and inmanta/pytest-inmanta#396

Many Python packages use their root `__init__.py` to expose a `__version__` name that contains the current version. Considering how often we need it, I think it would be best to introduce this in `inmanta-core` so we don't have to rely on `pkg_resources.get_distribution` for this.

- [x] add version
  - [x] add a comment to both places where the version is to also update the other and/or use bumpversion 
- [x] make sure it gets bumped correctly by bumpversion, release tool, merge tool, ...
   - [x] -> manual bumpversion updates it ok
   - [x] -> release tool and merge tool run bumpversion as well: https://github.com/inmanta/irt/blob/8dc2e7fb5f9df94fb14bcd2edfd6dbe86b86cc38/irt/version.py#L211

- [x]  update uses of `pkg_resources.get_distribution` in core itself 
- [x] Create a follow up ticket to use this value instead of  `pkg_resources.get_distribution` in other extensions/libraries 
    -> https://github.com/inmanta/inmanta-core/issues/6695

closes https://github.com/inmanta/inmanta-core/issues/6175

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/extensions.py
+++ b/src/inmanta/server/extensions.py
@@ -20,10 +20,9 @@ import os
 from collections import defaultdict
 from typing import Any, Dict, Generic, List, Optional, TypeVar
 
-import pkg_resources
 import yaml
 
-from inmanta import data
+from inmanta import __version__, data
 from inmanta.config import feature_file_config
 from inmanta.data.model import ExtensionStatus
 from inmanta.server.protocol import ServerSlice
@@ -134,19 +133,8 @@ class FeatureManager:
             product="Inmanta Service Orchestrator",
             edition="Open Source Edition",
             license="Apache Software License 2",
-            version=self._get_product_version(),
+            version=__version__,
         )
-
-    def _get_product_version(self) -> str:
-        packages = ["inmanta-oss", "inmanta", "inmanta-core"]
-        for package in packages:
-            try:
-                return pkg_resources.get_distribution(package).version
-            except pkg_resources.DistributionNotFound:
-                pass
-
-        LOGGER.warning("Couldn't determine product version. Make sure inmanta is properly installed.")
-        return "0.0.0"
 
     def add_slice(self, slice: ServerSlice) -> None:
         for feature in slice.define_features():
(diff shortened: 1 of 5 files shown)


src/inmanta/server/extensions.py
```diff
@@ -134,20 +133,9 @@ def get_product_metadata(self) -> ProductMetadata:
             product="Inmanta Service Orchestrator",
             edition="Open Source Edition",
             license="Apache Software License 2",
-            version=self._get_product_version(),
+            version=__version__,
```
sanderr, line 136: I'm not entirely sure about this change. I like the simplicity, but it will always report the stable version. I'm not sure if that matters. @wouterdb wdyt?

sanderr (changes requested):

sanderr (changes requested):
Could you also update the title? The merge tool uses that for the commit message.

src/inmanta/server/extensions.py
wouterdb: 1. I like the simplicity
2. I do use this a lot to know the exact version installed (the exact dev version is often relevant on mgmt)
3.  I wonder if it would be worth it, that when building packages, use bumpversion to add the build tag to the version in `__version__`  or to add a `__tag__` as well to be able to know the full version, based purely on source?

sanderr: 3. Could be interesting, but I honestly don't think bumpversion can handle it. Perhaps its replacement can, but I actually think it's considered out of scope. We could look into it.

wouterdb: I would then revert this to use the previous implementation

wouterdb (changes requested):

wouterdb (approved):

sanderr (approved):

Hugo-Inmanta:
Merged into iso6 with https://github.com/inmanta/inmanta-core/commit/17bd5d7b85faf1a042884980301b148dbfbea141

Issue #6372, opened 2023-08-08
move minimal python version to Python 3.11


sanderr:
- [x] Run `pyupgrade` on iso6 and master to use 3.9 conventions, verify results. 
  -> PR with interesting discussion here https://github.com/inmanta/inmanta-core/pull/6696
- [x] Then try the same for 3.11 on master only, don't commit if it causes a lot of divergence. ask on slack if in doubt. 
- [x] bump the minimal python version on master to 3.11
  - [x] https://github.com/inmanta/inmanta/pull/122
  - [x] core -> Already done
  - [x] https://github.com/inmanta/inmanta-license/pull/571
  - [x] https://github.com/inmanta/inmanta-lsm/pull/1518 
  - [x] https://github.com/inmanta/inmanta-service-orchestrator/pull/447
  - [x] https://github.com/inmanta/inmanta-ui/pull/486
- [x] check typing-extensions imports and replace with typing imports where possible. Drop typing-extensions as dependency if appropriate.
- [x] make tickets for the other extensions to do the same
  -> https://github.com/inmanta/inmanta-core/issues/6819
- [x] add internal docs on how to run the tool:
  - https://code.inmanta.com/inmanta/developerdocs/-/merge_requests/137/diffs

sanderr:
Wait for solutions team to give the go-ahead. Expected around end of August.

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/6372?planning-poker) with Zenhub @arnaudsjs @FloLey @Hugo-Inmanta @sanderr

PR #6696, opened 2023-10-30
Issue/6372 run pyupgrade python39

Hugo-Inmanta:
# Description

Run `pyupgrade --py39-plus`

Part of https://github.com/inmanta/inmanta-core/issues/6372

# Self Check:

- [ ] Changelog entry


tests/data/compile_138/libs/lma/plugins/__init__.py
```diff
@@ -16,7 +16,6 @@
     Contact: code@inmanta.com
 """
 
-from inmanta import ast
```
Hugo-Inmanta, line 19: Might not be ok to remove

tests/data/modules_v2/elaboratev2module_extension/inmanta_plugins/elaboratev2module_extension/__init__.py
```diff
@@ -1 +0,0 @@
-import inmanta_plugins.elaboratev2module
```
Hugo-Inmanta, line 1: same

bartv:
Shouldn't we also run this in the linting phase once we are up to date?

bartv:
Based on the failed tests I also think we can only do this once we are using typeadapter instead of our own typing logic in protocol (cc @sanderr )

sanderr (changes requested):
@Hugo-Inmanta this is going to be merge hell with the pydantic PR, that's why I put that one as a blocker. I think it's easier by far to just run this tool again once that PR is in. If you want to go ahead with the rest of the ticket you can split this task off into a separate ticket.

> Shouldn't we also run this in the linting phase once we are up to date?

I like that.

Hugo-Inmanta:
>  that's why I put that one as a blocker

Damn it, I completely missed the blockers.

sanderr (commented):
Wow, this changed way more lines than I would've thought...

bartv:
@sanderr this wil only work if our protocol layer starts supporting dict and list instead of Dict and List. I do not think that is the case and TypeAdapter will solve that for us.

PR #6708, opened 2023-11-03
Issue/6691 catch all arguments 2

### LINKED ISSUE #6691: Add support for "catch-all" positional and key-words arguments in plugins
Currently, plugins can accept args and kwargs, but not the catch-all arguments (`*args` and `**kwargs`).

The idea is to make this possible:
```python
@inmanta.plugins.plugin()
def my_plugin(*args: "any", **kwargs: "any") -> "any":
    print(args)
    print(kwargs)

```

The "only" thing that needs to be done is update the validation logic, located in two methods:
- https://github.com/inmanta/inmanta-core/blob/b184126ee94580036c13c4d035a58d36e85330e9/src/inmanta/plugins.py#L265
- https://github.com/inmanta/inmanta-core/blob/b184126ee94580036c13c4d035a58d36e85330e9/src/inmanta/plugins.py#L391

I tried to use such arguments by "disabling" `check_args` method (making it return `True` no matter what) and it worked like a charm.  Currently the limitation is only the validation:

```
self = <inmanta_plugins.config.get_config_template_value_as_string object at 0x7f15a714b1c0>
args = [{'carrier_ethernet_evc': {'backend': {'default': 'EVPN', 'enabled': ['EVPN']}, 'mtu': 9100, 'multi_homing': {'create_...in', 'enabled': True, 'eth_test': True, ...}, 'juniper': {'netconf_retry_count': 5, 'netconf_retry_interval': 5}, ...}]
kwargs = {'dict_path': 'nokia.customer_id', 'instance': connect::CarrierEthernetEvcEndPoint 7f15a3069dc0}

    def check_args(self, args: List[object], kwargs: Dict[str, object]) -> bool:
        """
        Check if the arguments of the call match the function signature
        """
        max_arg = len(self.arguments)
        if len(args) + len(kwargs) > max_arg:
            raise Exception(
                "Incorrect number of arguments for %s. Expected at most %d, got %d"
                % (self.get_signature(), max_arg, len(args) + len(kwargs))
            )
        # check for missing arguments
        required_args: FrozenSet[str] = frozenset(arg.arg_name for arg in self.arguments if not arg.has_default_value())
        present_positional_args: FrozenSet[str] = frozenset(arg.arg_name for arg in self.arguments[: len(args)])
        present_kwargs: FrozenSet[str] = frozenset(kwargs.keys())
        missing_args = required_args.difference({*present_positional_args, *present_kwargs})
        if missing_args:
            raise RuntimeException(
                None,
                "Missing %d required arguments for %s(): %s"
                % (len(missing_args), self.__class__.__function_name__, ",".join(missing_args)),
            )
        # check for kwargs overlap with positional arguments
        overlapping_args: FrozenSet[str] = present_kwargs.intersection(present_positional_args)
        if overlapping_args:
            raise RuntimeException(
                None,
                "Multiple values for %s in %s()" % (",".join(overlapping_args), self.__class__.__function_name__),
            )
    
        def is_valid(expected_arg: PluginArgument, expected_type: Type[object], arg: object) -> bool:
            if isinstance(arg, Unknown):
                return False
    
            if not self._is_instance(arg, expected_type):
                raise Exception(
                    "Invalid type for argument %s of '%s', it should be %s and %s given."
                    % (expected_arg.arg_name, self.__class__.__function_name__, expected_arg.arg_type, arg.__class__.__name__)
                )
            return True
    
        for i in range(len(args)):
            if not is_valid(self.arguments[i], self.argtypes[i], args[i]):
                return False
        arg_types: Dict[str, Tuple[PluginArgument, Optional[inmanta_type.Type]]] = {
            arg.arg_name: (arg, self.argtypes[i]) for i, arg in enumerate(self.arguments)
        }
        for k, v in kwargs.items():
            try:
>               (expected_arg, expected_type) = arg_types[k]
E               KeyError: 'instance'

/home/guillaume/.virtualenvs/inmanta-module-bics-ce/lib/python3.9/site-packages/inmanta/plugins.py:439: KeyError
```

edvgui:
# Description

* Short description here *

closes #6691 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/plugins.py
+++ b/src/inmanta/plugins.py
@@ -15,6 +15,7 @@
 
     Contact: code@inmanta.com
 """
+import collections
 import asyncio
 import inspect
 import os
@@ -200,11 +201,10 @@ class PluginArgument:
     NO_DEFAULT_VALUE_SET = object()
 
     def __init__(
-        self, arg_name: str, arg_type: object, is_kw_only_argument: bool, default_value: object = NO_DEFAULT_VALUE_SET
+        self, arg_name: str, arg_type: object, default_value: object = NO_DEFAULT_VALUE_SET
     ) -> None:
         self.arg_name = arg_name
         self.arg_type = arg_type
-        self.is_kw_only_argument = is_kw_only_argument
         self._default_value = default_value
 
     @property
@@ -236,11 +236,12 @@ class Plugin(NamedType, WithComment, metaclass=PluginMeta):
         self._context: int = -1
         self._return = None
 
-        self.arguments: List[PluginArgument]
+        self.positional_arguments: dict[int, PluginArgument]
+        self.keyword_arguments: dict[str, PluginArgument]
         if hasattr(self.__class__, "__function__"):
-            self.arguments = self._load_signature(self.__class__.__function__)
+            self.positional_arguments, self.keyword_arguments = self._load_signature(self.__class__.__function__)
         else:
-            self.arguments = []
+            self.positional_arguments, self.keyword_arguments = {}, {}
 
         self.new_statement = None
 
@@ -262,59 +263,76 @@ class Plugin(NamedType, WithComment, metaclass=PluginMeta):
         self.argtypes = [self.to_type(arg.arg_type, self.namespace) for arg in self.arguments]
         self.returntype = self.to_type(self._return, self.namespace)
 
-    def _load_signature(self, function: Callable[..., object]) -> List[PluginArgument]:
+    def _load_signature(
+        self,
+        function: Callable[..., object],
+    ) -> tuple[dict[int, PluginArgument], dict[str, PluginArgument]]:
         """
-        Load the signature from the given python function
+        Load the signature from the given python function, returns a tuple containing:
+        - As first value, a dict with the mapping from each position to the corresponding
+            argument to validate positioned input parameters.
+        - As second value, a dict with the mapping from each kwarg name to the corresponding
+            argument to validate key-word input parameters.
         """
         arg_spec = inspect.getfullargspec(function)
 
-        # Calculate at which index the default values start (for the non-keyword-only attributes)
-        default_start_for_args: Optional[int]
-        if arg_spec.defaults is not None:
-            default_start_for_args = len(arg_spec.args) - len(arg_spec.defaults)
+        positional_args: dict[int, PluginArgument]
+        if arg_spec.varargs is None:
+            positional_args = dict()
         else:
-            default_start_for_args = None
+            # There is a catch all positional arg, define it as default arg
+            # for args
+            catch_all = PluginArgument(
+                arg_name=arg_spec.varargs,
+                arg_type=arg_spec.annotations[arg_spec.varargs],
+            )
+            positional_args = collections.defaultdict(lambda: catch_all)
+        
+        key_word_args: dict[str, PluginArgument]
+        if arg_spec.varkw is None:
+            key_word_args = dict()
+        else:
+            # There is a catch all, define it as default arg for kwargs
+            catch_all = PluginArgument(
+                arg_name=arg_spec.varkw,
+                arg_type=arg_spec.annotations[arg_spec.varkw],
+            )
+            key_word_args = collections.defaultdict(lambda: catch_all)
+        
+        # Save all positional arguments
+        for position, arg in arg_spec.args:
+            positional_args[position] = PluginArgument(
+                arg_name=arg,
+                arg_type=arg_spec.annotations[arg],
+                default_value=(
+                    arg_spec.defaults[-position]
+                    if position < len(arg_spec.defaults)
+                    else PluginArgument.NO_DEFAULT_VALUE_SET,
+                ),
+            )
 
-        arguments: List[PluginArgument] = []
-
-        def process_kw_only_args(argument_name: str, annotation: object) -> PluginArgument:
-            if arg_spec.kwonlydefaults and argument_name in arg_spec.kwonlydefaults:
-                default_value = arg_spec.kwonlydefaults[argument_name]
-                return PluginArgument(argument_name, annotation, is_kw_only_argument=True, default_value=default_value)
-            else:
-                return PluginArgument(argument_name, annotation, is_kw_only_argument=True)
-
-        def process_regular_args(index: int, argument_name: str, annotation: object) -> PluginArgument:
-            if default_start_for_args is not None and default_start_for_args <= index:
-                default_value = arg_spec.defaults[default_start_for_args - index]
-                return PluginArgument(argument_name, annotation, is_kw_only_argument=False, default_value=default_value)
-            else:
-                return PluginArgument(argument_name, annotation, is_kw_only_argument=False)
-
-        def process_arg(index: int, argument_name: str, is_kwonly_arg: bool) -> None:
-            if argument_name not in arg_spec.annotations:
-                raise CompilerException(f"All arguments of plugin '{function.__name__}' should be annotated")
-            annotation = arg_spec.annotations[argument_name]
-            if annotation == Context:
-                self._context = index
-            else:
-                if is_kwonly_arg:
-                    plugin_argument = process_kw_only_args(argument_name, annotation)
-                else:
-                    plugin_argument = process_regular_args(index, argument_name, annotation)
-                arguments.append(plugin_argument)
-
-        # Process regular arguments
-        for i in range(len(arg_spec.args)):
-            process_arg(i, arg_spec.args[i], is_kwonly_arg=False)
-        # Process keyword-only arguments
-        for i in range(len(arg_spec.kwonlyargs)):
-            process_arg(i, arg_spec.kwonlyargs[i], is_kwonly_arg=True)
+            # All positional arguments can also be assigned as kwargs
+            key_word_args[arg] = PluginArgument(
+                arg_name=arg,
+                arg_type=arg_spec.annotations[arg],
+            )
+        
+        # Save all key-word arguments
+        for arg in arg_spec.kwonlyargs:
+            key_word_args[arg] = PluginArgument(
+                arg_name=arg,
+                arg_type=arg_spec.annotations[arg],
+                default_value=(
+                    arg_spec.kwonlydefaults[arg]
+                    if arg in arg_spec.kwonlydefaults
+                    else PluginArgument.NO_DEFAULT_VALUE_SET
+                ),
+            )
 
         if "return" in arg_spec.annotations:
             self._return = arg_spec.annotations["return"]
 
-        return arguments
+        return positional_args, key_word_args
 
     def get_signature(self) -> str:
         """
@@ -392,6 +410,7 @@ class Plugin(NamedType, WithComment, metaclass=PluginMeta):
         """
         Check if the arguments of the call match the function signature
         """
+        return True
         max_arg = len(self.arguments)
         if len(args) + len(kwargs) > max_arg:
             raise Exception(
--- a/src/inmanta/plugins.py
+++ b/src/inmanta/plugins.py
@@ -186,7 +186,7 @@ class PluginMeta(type):
 def resolve_type(locatable_type: LocatableString, resolver: Namespace) -> Optional[inmanta_type.Type]:
     """
     Convert a locatable type string, into a real inmanta type, that can be used for validation.
-    Alternatively, the locatable string defines a type that doesn't have any constraint, return None.
+    Alternatively, if the locatable string defines a type that doesn't have any constraint, return None.
 
     :param locatable_type: An object pointing to the type annotation string.
     :param resolver: The namespace that can be used to resolve the type annotation of this
@@ -271,8 +271,8 @@ class PluginIO:
 
         if not isinstance(self.type_expression, str):
             raise CompilerException(
-                "Bad annotation in plugin %s, expected str but got %s (%s)"
-                % (plugin.get_full_name(), type(self.type_expression).__name__, self.type_expression)
+                "Bad annotation in plugin %s for %s, expected str but got %s (%s)"
+                % (plugin.get_full_name(), self.IO_NAME, type(self.type_expression).__name__, self.type_expression)
             )
 
         plugin_line: Range = Range(plugin.location.file, plugin.location.lnr, 1, plugin.location.lnr + 1, 1)
@@ -545,6 +545,8 @@ class Plugin(NamedType, WithComment, metaclass=PluginMeta):
         """
         Get the argument at a given position, raise a KeyError if it doesn't exists.
         If a catch-all positional argument is defined, it will never raise a KeyError.
+
+        :param position: The position of the argument (excluding any Context argument)
         """
         if self.var_args is not None:
             return self.args.get(position, self.var_args)
@@ -555,6 +557,8 @@ class Plugin(NamedType, WithComment, metaclass=PluginMeta):
         """
         Get the argument with a given name, raise a KeyError if it doesn't exists.
         If a catch-all keyword argument is defined, it will never raise a KeyError.
+
+        :param name: The name of the argument
         """
         if self.var_kwargs is not None:
             return self.kwargs.get(name, self.var_kwargs)
--- a/src/inmanta/plugins.py
+++ b/src/inmanta/plugins.py
@@ -16,7 +16,6 @@
     Contact: code@inmanta.com
 """
 import asyncio
-import collections
 import copy
 import inspect
 import os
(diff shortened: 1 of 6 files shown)


src/inmanta/plugins.py
```diff
@@ -191,21 +192,160 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
...
+        if isinstance(value, Unknown):
+            # Value is not known, it can not be validated
+            return False
+
+        if self.resolved_type is None:
+            # Any value is valid
+            return True
+
+        # Validate the value, use custom validate method of the type if it exists
+        valid = self.resolved_type.validate(value)
```
edvgui, line 313: Code-wise, this is a breaking change, before, it was equivalent to this:
```python
        if hasattr(self.resolved_type, "validate"):
            valid = getattr(self.resolved_type, "validate")(value)
        else:
            valid = isinstance(value, self.resolved_type)
```
But typing-wise, it makes no sense, as the value in `resolved_type` comes from `Namespace.get_type` and that one only returns an inmanta `Type` object.  Is it okay to drop it then? Or are we not confident enough that the typing is correct?

wouterdb (commented):
I'm in favor of this in general terms, I do have some detail comments, but now may not be the time for them?

tests/compiler/test_plugins.py
```diff
@@ -80,7 +86,8 @@ def test_kwargs_in_plugin_call_missing_arg(snippetcompiler):
         """
 std::equals(42, desc="they differ")
         """,
-        "Missing 1 required arguments for equals(): arg2" " (reported in std::equals(42,desc='they differ') ({dir}/main.cf:2))",
+        "std::equals() missing 1 required positional argument: 'arg2' "
```
wouterdb, line 89: nice

wouterdb (changes requested):
Very nice

src/inmanta/plugins.py
```diff
@@ -234,13 +421,15 @@ def __init__(self, namespace: Namespace) -> None:
 
         # The index of the Context attribute.
         self._context: int = -1
-        self._return = None
 
-        self.arguments: List[PluginArgument]
+        # Load the signature and build all the PluginArgument objects corresponding to it
+        self.args: list[PluginArgument] = list()
+        self.var_args: Optional[PluginArgument] = None
+        self.kwargs: dict[str, PluginArgument] = dict()
```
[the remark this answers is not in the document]
edvgui: True, I could actually simply go over all the kwargs, as it includes all the args (and then still do the var_args and var_kwargs)

src/inmanta/plugins.py
```diff
@@ -191,21 +193,191 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
...
+    if locatable_type.value.endswith("[]"):
+        # We don't want to modify the object we received as argument
+        locatable_type = copy.copy(locatable_type)
+        locatable_type.value = locatable_type.value[0:-2]
+        transformation_stack.append(inmanta_type.TypedList)
+
+    return reduce(lambda acc, transform: transform(acc), reversed(transformation_stack), resolver.get_type(locatable_type))
+
+
+class PluginIO:
```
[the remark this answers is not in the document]
edvgui: The thing is, the base class is used for the returned value, which is hardly an argument, is it?  I would also prefer something else than IO though.  Maybe `PluginAnnotation`?

src/inmanta/plugins.py
```diff
@@ -191,21 +193,191 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
             cls.__functions = {}
 
 
-class PluginArgument:
+def report_missing_arguments(
```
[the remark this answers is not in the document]
edvgui: Done

src/inmanta/plugins.py
```diff
@@ -191,21 +193,191 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
...
+    IO_TYPE: str = ""
+    IO_NAME: str = ""
+
+    def __init__(self, type_expression: object) -> None:
+        self.type_expression = type_expression
+
+        # We define the attribute but don't set it yet, this will be done when
+        # the type is resolved.  This allows to differentiate between a type that
+        # has not been resolved and a type that is a match for "any" (None).
+        self._resolved_type: Optional[inmanta_type.Type]
```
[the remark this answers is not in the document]
edvgui: I propose to address it this way: https://github.com/inmanta/inmanta-core/pull/6744
Wdyt?

src/inmanta/plugins.py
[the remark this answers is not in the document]
edvgui: I have something against providing a list that contains all arguments and it is that the catch-all args then would need to have something that differentiate them from the others on the object itself, which I didn't find any elegant solution for atm.  Suggestions are welcome

src/inmanta/plugins.py
```diff
@@ -191,21 +193,191 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
...
+    Alternatively, if the locatable string defines a type that doesn't have any constraint, return None.
+
+    :param locatable_type: An object pointing to the type annotation string.
+    :param resolver: The namespace that can be used to resolve the type annotation of this
+        argument.
+    """
+    if locatable_type.value == "any":
+        return None
+
+    if locatable_type.value == "expression":
```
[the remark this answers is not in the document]
edvgui: https://github.com/inmanta/inmanta-core/issues/6746

src/inmanta/plugins.py
```diff
@@ -191,21 +193,191 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
...
+                ", ".join(repr(arg) for arg in missing_args[:-1]),
+                repr(missing_args[-1]),
+            )
+        )
+        # The exception raised here tries to match as closely as possible what python
+        # would have raised as exception
+        raise RuntimeException(None, f"{func}() missing {len(missing_args)} required {args_sort} arguments: {arg_names}")
+
+
+def resolve_type(locatable_type: LocatableString, resolver: Namespace) -> Optional[inmanta_type.Type]:
```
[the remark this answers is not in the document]
edvgui: I moved it to the type module, and removed anything that was related to plugins

src/inmanta/plugins.py
```diff
@@ -193,76 +191,26 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
             cls.__functions = {}
 
 
-def report_missing_arguments(
-    func: str, missing_args: collections.abc.Sequence[str], args_sort: Literal["positional", "keyword-only"]
-) -> None:
+class Never(inmanta_type.Type):
```
wouterdb, line 194: I would just make a Null type, this is very complicated 

I will also print itself as `Never?` in error messages I think.

src/inmanta/plugins.py
```diff
@@ -490,6 +427,10 @@ def get_annotation(arg: str) -> object:
 
             return arg_spec.annotations[arg]
 
+        # Make sure we have a return annotation even for implicit "null"
+        arg_spec.annotations["return"] = arg_spec.annotations.get("return") or "null"
```
wouterdb, line 431: Why write it back to the arg_spec?

src/inmanta/plugins.py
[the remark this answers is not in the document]
wouterdb: `AbstractPluginArgument` ?

I think a return values is also an argument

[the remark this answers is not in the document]
edvgui: Everywhere I could look, it says something like this: 
> an argument is a value that is accepted by a function

Maybe `PluginAnnotationABC`?

src/inmanta/plugins.py
edvgui: I renamed it to `Null`

src/inmanta/plugins.py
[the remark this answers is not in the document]
edvgui: I renamed it to `PluginValue`

src/inmanta/plugins.py
edvgui: I updated the explanation, is this more clear?

src/inmanta/plugins.py
```diff
@@ -191,41 +191,41 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
             cls.__functions = {}
 
 
-class Never(inmanta_type.Type):
+class Null(inmanta_type.Type):
```
wouterdb, line 194: I would also implement the `type_string`  method, as this is used to format errors

wouterdb (approved):
one more minor remark

src/inmanta/plugins.py
edvgui: Done

src/inmanta/plugins.py
```diff
@@ -259,188 +391,268 @@ def __init__(self, namespace: Namespace) -> None:
...
     def normalize(self) -> None:
         self.resolver = self.namespace
-        self.argtypes = [self.to_type(arg.arg_type, self.namespace) for arg in self.arguments]
-        self.returntype = self.to_type(self._return, self.namespace)
 
-    def _load_signature(self, function: Callable[..., object]) -> List[PluginArgument]:
+        # Resolve all the types that we expect to receive as input of our plugin
+        for arg in self.kwargs.values():
+            # Normalizing the kwargs will also normalize the args as they are all
+            # included in this dict
```
sanderr, line 398: Could you add a comment about this where the attribute is declared? When I first read it I assumed `kwargs` would hold only the keyword-only arguments.

src/inmanta/plugins.py
```diff
@@ -259,188 +391,268 @@ def __init__(self, namespace: Namespace) -> None:
...
+                        arg_spec.defaults[position - defaults_start_at]
+                        if position >= defaults_start_at
+                        else PluginArgument.NO_DEFAULT_VALUE_SET
+                    ),
+                )
+            )
 
-        return arguments
+            # All positional arguments can also be assigned as kwargs
+            self.kwargs[arg] = self.args[position]
```
sanderr, line 482: Slight preference to using a temp var for this object to be a bit more robust.

src/inmanta/plugins.py
```diff
@@ -259,188 +391,268 @@ def __init__(self, namespace: Namespace) -> None:
...
+                    default_value=(
+                        arg_spec.defaults[position - defaults_start_at]
+                        if position >= defaults_start_at
+                        else PluginArgument.NO_DEFAULT_VALUE_SET
+                    ),
+                )
+            )
 
-        return arguments
+            # All positional arguments can also be assigned as kwargs
```
sanderr, line 481: Shame that we can't support positional-only args. I can't seem to find any indication about them in the `arg_spec`.

src/inmanta/plugins.py
```diff
@@ -259,188 +448,234 @@ def __init__(self, namespace: Namespace) -> None:
...
+            return self.kwargs[name]
+        elif self.var_kwargs is not None:
+            return self.var_kwargs
+        else:
+            # Trying to provide a keyword argument which doesn't exist
+            # The exception raised here tries to match as closely as possible what python
+            # would have raised as exception
+            raise RuntimeException(None, f"{self.get_full_name()}() got an unexpected keyword argument: '{name}'")
 
     def check_args(self, args: List[object], kwargs: Dict[str, object]) -> bool:
```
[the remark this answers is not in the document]
sanderr: I'm all for that. I would not try it in this PR.

src/inmanta/plugins.py
```diff
@@ -259,188 +391,268 @@ def __init__(self, namespace: Namespace) -> None:
...
             )
+        ]
+        self.report_missing_arguments(missing_positional_arguments, "positional")
+
+        # (2) Check that all keyword arguments without a default are provided
+        missing_keyword_arguments = [
+            name
+            for name, arg in self.kwargs.items()
+            if (
+                arg.is_kw_only_argument  # The argument was not provided as positional arg
```
sanderr, line 625:
```suggestion
                arg.is_kw_only_argument  # The argument was not yet checked as positional arg
```

sanderr (changes requested):
Nice!

src/inmanta/plugins.py
edvgui: Well, looks like there is a way: https://stackoverflow.com/questions/71106783/how-to-inspect-positional-only-parameters

Should I add it? I never used them, but it is probably not too much work to add it now and it would make the implementation complete

src/inmanta/plugins.py
edvgui: I don't really get what you mean with temp var?  And which object are you referring to?

src/inmanta/plugins.py
```diff
@@ -191,21 +191,136 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
...
+    """
+    This custom type is used for the validation of plugins which only
+    accept null as an argument or return value.  It is meant to be used
+    as the element_type of an instance of the `NullableType` class.
+    """
+
+    def validate(self, value: object | None) -> bool:
+        return isinstance(value, NoneValue)
+
+    def type_string(self) -> str | None:
```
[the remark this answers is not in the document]
edvgui: I copy-pasted the signature from the parent class, but indeed, this one is always returning a string.  Is it considered good practice to always "retrain" the return type of a sub class's method when we can ensure that the returned type will indeed never be one of the parent's offerings?

src/inmanta/plugins.py
sanderr: No strong preference. I also rarely use them and was more stating it for completeness of the implementation. It looks simple enough to identify them like that, so it would be nice to have, but only if it's low-effort.

src/inmanta/plugins.py
```diff
@@ -259,188 +391,268 @@ def __init__(self, namespace: Namespace) -> None:
...
-            if default_start_for_args is not None and default_start_for_args <= index:
-                default_value = arg_spec.defaults[default_start_for_args - index]
-                return PluginArgument(argument_name, annotation, is_kw_only_argument=False, default_value=default_value)
-            else:
-                return PluginArgument(argument_name, annotation, is_kw_only_argument=False)
+        # Make sure we have a return annotation even for implicit "null"
+        # If no return annotation (or a Falsy value) is provided, we write
+        # "null" in the annotations dict so that we can use the get_annotation
+        # helper defined above.
+        arg_spec.annotations["return"] = arg_spec.annotations.get("return") or "null"
```
edvgui, line 437: @sanderr @wouterdb 
I think this is not right, the previous implementation defaulted to `any` when not return annotation was provided (implicitly, it would actually simply not do any validation), this one defaults to none, and will validate that the returned value is None.  It is more inline with python, but also a breaking change, should I keep it consistent with the previous behavior?

sanderr: >  It is more inline with python

I don't think this is true. The following gives a mypy error for passing a string to `x`, but not for `return 1`.
```
def f(x: int):
    return 1


f("test")
```

So I think that indeed we should default to `any`.

src/inmanta/plugins.py
sanderr:
```
some_var = <the_new_arg>
self.args.append(some_var)
self.kwargs[arg] = some_var
```

So that we don't have to rely on the assumption that we inject exactly one object into `self.args` for each iteration.

src/inmanta/plugins.py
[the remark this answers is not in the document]
sanderr: I wouldn't call it "retrain", more "refine" because you still have to be compatible with the parent. i.e. you can return something more specific, but not less so. And indeed I typically try to do this whenever possible, unless there's a chance the signature might change due to a change of implementation. Afaik it is indeed considered good practice.

src/inmanta/plugins.py
edvgui: Oh, something I have to unlearn then.  So there is no discussion, I will fix it

src/inmanta/plugins.py
edvgui: Turned out to be more effort than I had hoped, I will skip it for now

src/inmanta/plugins.py
edvgui: Fixed, and I added a test case for it

src/inmanta/plugins.py
edvgui: I tried to make it more clear with an additional attribute: `all_args`

src/inmanta/plugins.py
```diff
@@ -191,21 +191,136 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
...
 
-class PluginArgument:
+class Null(inmanta_type.Type):
+    """
+    This custom type is used for the validation of plugins which only
+    accept null as an argument or return value.  It is meant to be used
+    as the element_type of an instance of the `NullableType` class.
+    """
+
+    def validate(self, value: object | None) -> bool:
```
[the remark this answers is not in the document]
edvgui: I am aware, yes, this is just coming from the parent class

src/inmanta/plugins.py
```diff
@@ -191,21 +191,136 @@ def clear(cls, inmanta_module: Optional[str] = None) -> None:
             cls.__functions = {}
 
 
-class PluginArgument:
+class Null(inmanta_type.Type):
+    """
+    This custom type is used for the validation of plugins which only
+    accept null as an argument or return value.  It is meant to be used
+    as the element_type of an instance of the `NullableType` class.
```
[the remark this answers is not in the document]
edvgui: Indeed, this was outdated, I removed the string

sanderr (approved):

PR #6716, opened 2023-11-07
issue6595/incorrect anchor for constructor with inferred namespace

### LINKED ISSUE #6595: incorrect anchor for constructor with inferred namespace
When a constructor is called without an explicit namespace, relying on the namespace inference, the `TypeReferenceAnchor` it registers can not be resolved.

At construction of a `inmanta.ast.statements.generator.Constructor` object, it [registers a `TypeReferenceAnchor`](https://github.com/inmanta/inmanta-core/blob/17484701e1bf85722a59881444d4669907052749/src/inmanta/ast/statements/generator.py#L823C18-L823C18). This anchor contains the explicit namespace for the constructor call, but it is not aware of the [inference that happens at normalization time](https://github.com/inmanta/inmanta-core/blob/17484701e1bf85722a59881444d4669907052749/src/inmanta/ast/statements/generator.py#L865). Therefore the anchor will never successfully [resolve](https://github.com/inmanta/inmanta-core/blob/17484701e1bf85722a59881444d4669907052749/src/inmanta/ast/__init__.py#L253). This causes e.g. vscode to be incompatible with namespace inference.

FloLey:
# Description

When a constructor is called without an explicit namespace, relying on the namespace inference, the TypeReferenceAnchor it registers can not be resolved.

closes #6595

# Self Check:

- [x] Changelog entry


src/inmanta/ast/__init__.py
```diff
@@ -251,9 +251,21 @@ def __init__(self, namespace: "Namespace", type: LocatableString) -> None:
         self.type = type
 
     def resolve(self) -> Optional[AnchorTarget]:
-        t = self.namespace.get_type(self.type)
-        location = t.get_location()
-        docstring = t.comment if isinstance(t, WithComment) else None
+        tp = None
+        try:
+            tp = self.namespace.get_type(self.type)
+        except TypeNotFoundException as e:
```
sanderr, line 257: I don't think this logic is entirely correct, I need to think on a better approach. The issue with this one is that it returns the first name it finds that matches, regardless of the type. Suppose you have the modules below. Then it would result in `b::C` rather than `c::C` (assuming that `b` would come before `c` in `visible_namespaces`, which I think is the case). The logic in `Constructor` would disregard `b::C` because it is not compatible with the type declared in the relation definition.

`a`:
```
entity A: end
A.c [1] -- c::C

A(c=C())
```

`b`:
```
entity B: end
entity C: end
```

`c`:
```
entity C: end
```

tests/data/modules/tests/model/_init.cf
```diff
@@ -0,0 +1,7 @@
+entity Test:
+end
+
+implementation test for Test:
+end
```
sanderr, line 5: Why even have an implementation if it's a NOOP?

tests/test_compiler_entrypoints.py
```diff
@@ -309,3 +310,59 @@ def verify_anchor(flnr, s, e, docs):
...
+    snippetcompiler.setup_for_snippet(
+        """
+    import tests
+    entity A:
+    end
+
+    A.mytest [1] -- tests::Test [1]
+
+    implementation a for A:
+    end
```
sanderr, line 328: Same question here.

sanderr (changes requested):

tests/data/modules/tests/model/_init.cf
FloLey: No reason, I copy/pasted it from an existing test above and didn't remove it.

src/inmanta/ast/__init__.py
FloLey: indeed, I didn't think of that :( I'll try to find something better to

sanderr: How about simply move the anchor registration to the normalization call? Anchors are never fetched before normalization as far as I can see. This is not part of any docstring, so we would have to document it in e.g. the `get_anchors` method or something. @wouterdb CC.

```diff
diff --git a/src/inmanta/ast/statements/generator.py b/src/inmanta/ast/statements/generator.py
index a585269b..df03f3cc 100644
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -820,7 +820,6 @@ class Constructor(ExpressionStatement):
         self.__wrapped_kwarg_attributes: List[WrappedKwargs] = wrapped_kwargs
         self.location = location
         self.namespace = namespace
-        self.anchors.append(TypeReferenceAnchor(namespace, class_type))
         for a in attributes:
             self.add_attribute(a[0], a[1])
         self.type: Optional["Entity"] = None
@@ -863,6 +862,7 @@ class Constructor(ExpressionStatement):

     def normalize(self, *, lhs_attribute: Optional[AttributeAssignmentLHS] = None) -> None:
         self.type = self._resolve_type(lhs_attribute)
+        self.anchors.append(TypeReferenceAnchor(self.namespace, self.type))

         inindex: abc.MutableSet[str] = set()
```

sanderr: My patch is not correct, but you get the point. It should be something more like `...(self.type.namespace, self.class_type)` I think.

wouterdb: I agree that:
1. this is making a guess, which (up to this point), we never do in the compiler
2. I would not duplicate the resolution logic
3. I agree that we should make sure normalization is called before we get the anchors out and use the existing normalization code to set the anchors

src/inmanta/ast/statements/generator.py
```diff
@@ -888,7 +891,7 @@ def normalize(self, *, lhs_attribute: Optional[AttributeAssignmentLHS] = None) -
             if not self.wrapped_kwargs and (self._lhs_attribute is None or len(self._required_dynamic_args) > 1):
                 raise IndexAttributeMissingInConstructorException(self, self.type, self._required_dynamic_args)
 
-        self._normalize_rhs(inindex)
+        self.anchors.extend(self._normalize_rhs(inindex))
```
FloLey, line 894: @sanderr for my test to pass (and the anchormap to contain the anchor to the constructor) I had make change this. Do you think this is an issue or this could be done in a better way? 

(I still need to move the anchor logic to the normalizers for the other cases)

sanderr: Why not just `append` to the anchors from that method?

FloLey: this is indeed way better, I don't know why I didn't do it like that immediately

Hugo-Inmanta (changes requested):

tests/test_compiler_entrypoints.py
```diff
@@ -309,3 +310,55 @@ def verify_anchor(flnr, s, e, docs):
...
+                and range1.start_char == start_char1
+                and range1.end_char == end_char1
+                and range2.file == file2
+                and range2.lnr == lnr2
+                and range2.start_char == start_char2
+                and range2.end_char == end_char2
+            ):
+                return True
+
+        return False
```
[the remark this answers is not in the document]
FloLey: I'm not sure I understand what you mean: We iterate over all the tuples present in the anchormap looking for the one that corresponds to the constructor we are interested in. If we find it we are happy, otherwise it is missing and we are sad. This test doesn't care about the other anchors present in the anchormap. I could verify the length of the anchormap to make sure he have the right number of anchors. I'll add that.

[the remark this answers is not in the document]
Hugo-Inmanta: Oh I see, I completely missed the point !
Then I think you could simplify this by creating 2 `Range´ objects instead of tuples and asserting:
```python
range_source = Range(...)
range_target = Range(...)
assert (range_source, range_target) in anchormap
```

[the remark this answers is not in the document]
FloLey: indeed, way better, thanks!

tests/test_compiler_entrypoints.py
```diff
@@ -309,3 +310,40 @@ def verify_anchor(flnr, s, e, docs):
...
     verify_anchor(24, 29, 30, None)
     verify_anchor(24, 11, 22, None)
+
+
+def test_constructor_with_inferred_namespace(snippetcompiler):
+    """
+    Test that the anchor for a constructor for an entity with an inferred namespace is correctly added to the anchormap
+    The test checks if the anchormap correctly reflects the relationship between the
+    source range (where the entity is instantiated: line 9 in main.cf) and the target range (where the
+    entity is defined: line 1 in the _init.cf file of the tests mmodule)
```
Hugo-Inmanta, line 320:
```suggestion
    entity is defined: line 1 in the _init.cf file of the tests module)
```

Hugo-Inmanta (approved):
LGTM !

sanderr (changes requested):
Looks good but I'm still missing a few things:
- docstrings in `Statement.get_anchors` and `BasicBlock.get_anchors` that state when it's safe to call (see e.g. `BasicBlock.get_eager_promises()` and `DynamicStatement.get_all_eager_promises()`
- You seem to have missed some statements (some of these depend on the outcome of the question I asked @wouterdb in another comment): `DefineRelation`, `DefineImplement` (`select`is normalized by `Implement`), `DefineIndex`, `DefineEntity`, `DefineImplementation` (implementation block is normalized), `DefineTypeConstraint`. Many of these are `DefinitionStatement`, which have `evaluate` rather than `normalize`

src/inmanta/ast/statements/assign.py
```diff
@@ -438,13 +438,13 @@ def __init__(
...
         ReferenceStatement.__init__(self, list(chain((v for (_, v) in query), wrapped_query)))
         self.index_type = index_type
-        self.anchors.append(TypeReferenceAnchor(index_type.namespace, index_type))
         self.query = [(str(n), e) for n, e in query]
         self.wrapped_query: typing.List["WrappedKwargs"] = wrapped_query
 
     def normalize(self, *, lhs_attribute: Optional[AttributeAssignmentLHS] = None) -> None:
         ReferenceStatement.normalize(self)
         self.type = self.namespace.get_type(self.index_type)
+        self.anchors.append(TypeReferenceAnchor(self.index_type.namespace, self.index_type))
```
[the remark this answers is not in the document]
FloLey: could you explain how? I don't see the issue. I'll add a testcase if needed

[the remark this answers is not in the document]
sanderr: I was wrong about this one. `ShortIndexLookup` doesn't call its parent constructor so it never got the anchors in the first place.

src/inmanta/ast/statements/define.py
```diff
@@ -134,7 +134,6 @@ def __init__(
             raise HyphenException(lname)
 
         self.anchors = [TypeReferenceAnchor(namespace, x) for x in parents]
```
FloLey, line 136: @sanderr this one I don't find how to move into the evaluate function without breaking existing tests. Except for this I think I did all the asked changes.

src/inmanta/ast/statements/assign.py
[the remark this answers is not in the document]
sanderr: Also applies to `DefinitionStatement.evaluate`.

src/inmanta/ast/statements/__init__.py
```diff
@@ -78,6 +78,7 @@ def get_location(self) -> Location:
         return self.location
 
     def get_anchors(self) -> List[Anchor]:
+        """Should only be called after normalization."""
```
sanderr, line 81:
```suggestion
        """Should only be called after normalization (DynamicStatement) or evaluation (DefinitionStatement)."""
```

src/inmanta/ast/statements/assign.py
```diff
@@ -228,6 +229,7 @@ def __init__(self, instance: "Reference", attribute_name: str, value: Expression
     def normalize(self, *, lhs_attribute: Optional[AttributeAssignmentLHS] = None) -> None:
         # register this assignment as left hand side to the value on the right hand side
         self.rhs.normalize(lhs_attribute=AttributeAssignmentLHS(self.instance, self.attribute_name))
+        self.anchors.extend(self.rhs.get_anchors())
```
sanderr, line 232: What about lhs?

src/inmanta/ast/statements/define.py
```diff
@@ -290,12 +287,14 @@ def evaluate(self) -> None:
...
         try:
             cls = self.namespace.get_type(self.entity)
+            self.anchors = [TypeReferenceAnchor(self.namespace, self.entity)]
             if not isinstance(cls, Entity):
                 raise TypingException(
                     self, "Implementation can only be define for an Entity, but %s is a %s" % (self.entity, cls)
                 )
             self.type.set_type(cls)
             self.copy_location(self.type)
+            self.anchors.extend(self.block.get_anchors())
```
sanderr, line 297: Oh damn, this does not work, does it? Afaik we only normalize the types (e.g. `Implementation`, which normalizes the block) after we've evaluated all definition statements. Perhaps we have to add a `normalize` stage to the `DefinitionStatement`?

Let me know if I'm missing something.

src/inmanta/ast/statements/define.py
```diff
@@ -367,6 +363,7 @@ def evaluate(self) -> None:
             implement.comment = self.comment
             implement.constraint = self.select
             implement.location = self.entity_location
+            self.anchors.extend(implement.constraint.get_anchors())
```
sanderr, line 366: I don't think the statement is normalized yet at this point. It's normalized in `Entity.normalize` (a type).

Damn, this is becoming a bit of a mess now that the definition statements come into it.

src/inmanta/ast/statements/define.py
```diff
@@ -564,6 +558,10 @@ def normalize(self) -> None:
         for _, exp in self.annotation_expression:
             exp.normalize()
 
+        self.anchors.extend((anchor for _, exp in self.annotation_expression for anchor in exp.get_anchors()))
```
sanderr, line 561: Could move this into the loop above

src/inmanta/ast/statements/define.py
```diff
@@ -573,11 +571,7 @@ class DefineIndex(DefinitionStatement):
     def __init__(self, entity_type: LocatableString, attributes: List[LocatableString]):
         DefinitionStatement.__init__(self)
         self.type = entity_type
-        self.attributes = [str(a) for a in attributes]
-        self.anchors.append(TypeReferenceAnchor(entity_type.namespace, entity_type))
-        self.anchors.extend(
-            [AttributeReferenceAnchor(x.get_location(), entity_type.namespace, entity_type, str(x)) for x in attributes]
-        )
+        self.attributes = attributes
```
sanderr, line 574:
```suggestion
        self.attributes: Sequence[LocatableString] = attributes
```

sanderr: Could you make sure to run a mypy-diff for this?

src/inmanta/ast/statements/define.py
sanderr: What's the issue? This one doesn't depend on normalization of anything else so we could leave it here if we need to but I'd prefer to move it if we can.

sanderr (changes requested):

src/inmanta/ast/statements/assign.py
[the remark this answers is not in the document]
sanderr: Related thread: https://github.com/inmanta/inmanta-core/pull/6716#discussion_r1404199869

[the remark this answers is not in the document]
wouterdb: I prefer consistency as well

src/inmanta/ast/statements/define.py
FloLey: Whenever I move it to the evaluate function I have extra anchors in the anchormap. I don't where they come from but the source location is always __internal__:1:1.

sanderr: See 6 lines down

FloLey: So can I leave it there right? I don't think I can move this code block to the evaluate function

PR #6722, opened 2023-11-09
Fix bug number cast to int

FloLey:
# Description

while working on https://github.com/inmanta/inmanta-core/pull/6686 we noticed a bug [here](https://github.com/inmanta/inmanta-core/pull/6686#discussion_r1386642720) . This PR fixes this bug

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/fix-cast-number.yml
+++ b/changelogs/unreleased/fix-cast-number.yml
@@ -1,6 +1,5 @@
 ---
 description: Fix a bug where numbers where cast to int instead of float
 change-type: patch
-destination-branches: [master, iso6]
-sections:
-  bugfix: "{{description}}"
+destination-branches: [iso6]
+
--- a/changelogs/unreleased/fix-cast-number.yml
+++ b/changelogs/unreleased/fix-cast-number.yml
@@ -1,0 +1,6 @@
+---
+description: Fix a bug where numbers where cast to int instead of float
+change-type: patch
+destination-branches: [master, iso6]
+sections:
+  bugfix: "{{description}}"
--- a/changelogs/unreleased/fix-cast-number.yml
+++ b/changelogs/unreleased/fix-cast-number.yml
@@ -2,4 +2,6 @@
 description: Fix a bug where numbers where cast to int instead of float
 change-type: patch
 destination-branches: [iso6]
+sections:
+  bugfix: "{{description}}"
 
--- a/src/inmanta/ast/type.py
+++ b/src/inmanta/ast/type.py
@@ -229,10 +229,7 @@ class Primitive(Type):
 @stable_api
 class Number(Primitive):
     """
-    This class represents an integer or float in the configuration model. On
-    these numbers the following operations are supported:
-
-    +, -, /, *
+    This class represents a float in the configuration model.
     """
 
     def __init__(self) -> None:
--- a/src/inmanta/ast/type.py
+++ b/src/inmanta/ast/type.py
@@ -234,27 +234,15 @@ class Number(Primitive):
 
     def __init__(self) -> None:
         Primitive.__init__(self)
-        self.try_cast_functions: Sequence[Callable[[Optional[object]], numbers.Number]] = [int, float]
+        self.try_cast_functions: Sequence[Callable[[Optional[object]], numbers.Number]] = [float]
 
     def cast(self, value: Optional[object]) -> object:
         """
         Attempts to cast a given value to an int or a float.
         """
-        exception: RuntimeException = RuntimeException(None, "Failed to cast '%s' to %s" % (value, self))
-
-        if isinstance(value, Unknown):
-            # propagate unknowns
-            return value
-
-        elif isinstance(value, int):
-            # Value is already an int, return it as-is
-            return value
-        else:
-            try:
-                # Attempt to cast the value to a float
-                return float(value)
-            except (ValueError, TypeError):
-                raise exception
+        if isinstance(value, int):
+            return int(value)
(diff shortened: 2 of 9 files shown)


src/inmanta/ast/type.py
```diff
@@ -237,7 +237,7 @@ class Number(Primitive):
 
     def __init__(self) -> None:
```
wouterdb, line 238: the docstring on the class is not entirely correct

FloLey: In my other [PR](https://github.com/inmanta/inmanta-core/pull/6686/files#diff-0a38a04c81c27002daab3fea8b2c4eac57e25c020dcca78a124df914b5de6ab3R232) I changed it like this so I've just made the same change here. I think it is indeed just a float now. That's the plan right @sanderr ?

sanderr: This is a good point. For iso6 we probably want `number(1)` to be stable and produce an int. So for iso7 I like this as is, for iso6 perhaps we have to override the cast method so it does an `int` cast if the value is already an int and a float cast otherwise.

I hadn't realized we'd want different behavior between iso6 and iso7 here. In this case it's probably easiest to just make this an iso6 PR and to put the iso7 part + bugfix entry in your in progress one.

sanderr (changes requested):

tests/compiler/test_typing.py
```diff
@@ -274,15 +279,6 @@ def test_cast_exception_value_error(snippetcompiler):
...
 
-def test_cast_exception_int_value_error(snippetcompiler):
-    snippetcompiler.setup_for_error(
-        """
-int("0.0")
-        """,
-        "Failed to cast '0.0' to int (reported in int('0.0') ({dir}/main.cf:2))",
-    )
-
-
```
FloLey, line 285: I'm not sure I understand why this test failed now but I think it is ok?

src/inmanta/ast/type.py
```diff
@@ -229,15 +229,20 @@ def __eq__(self, other: object) -> bool:
...
         Primitive.__init__(self)
-        self.try_cast_functions: Sequence[Callable[[Optional[object]], numbers.Number]] = [int, float]
+        self.try_cast_functions: Sequence[Callable[[Optional[object]], numbers.Number]] = [float]
+
+    def cast(self, value: Optional[object]) -> object:
+        """
+        Attempts to cast a given value to an int or a float.
+        """
+        if isinstance(value, int):
```
sanderr, line 243:
```suggestion
        # Keep precision: cast to an int only if it already is an int
        if isinstance(value, int):
```

tests/compiler/test_typing.py
```diff
@@ -380,3 +380,14 @@ def test_2243_override_optional(snippetcompiler) -> None:
 Child()
         """,
     )
+
+
+def test_print_number(snippetcompiler, capsys):
```
sanderr, line 385: Can you also add a test case for `int(1)` that it does not become `1.0`? Or does that already exist somewhere?

sanderr (changes requested):

tests/compiler/test_typing.py
FloLey: I think this test is enough:
https://github.com/inmanta/inmanta-core/blob/fix-bug-number-cast/tests/compiler/test_typing.py#L131

src/inmanta/ast/type.py
```diff
@@ -229,15 +229,21 @@ def __eq__(self, other: object) -> bool:
...
-        self.try_cast_functions: Sequence[Callable[[Optional[object]], numbers.Number]] = [int, float]
+        self.try_cast_functions: Sequence[Callable[[Optional[object]], numbers.Number]] = [float]
+
+    def cast(self, value: Optional[object]) -> object:
+        """
+        Attempts to cast a given value to an int or a float.
+        """
+        # Keep precision: cast to an int only if it already is an int
+        if isinstance(value, int):
+            return int(value)
```
wouterdb, line 245: This case is a bit unnecessary

wouterdb (approved):

changelogs/unreleased/fix-cast-number.yml
```diff
@@ -0,0 +1,5 @@
+---
+description: Fix a bug where numbers where cast to int instead of float
```
sanderr, line 2: I think this deserves a bugfix entry

src/inmanta/ast/type.py
sanderr: The cast you mean? I recommended it because `isinstance(True, int)`, yet `int(True) is 1`. In other words: when you cast you really expect to get the numeric value, not a "number-compatible" one.

tests/compiler/test_typing.py
sanderr: I think so. But perhaps we could extend it with an explicit `not any(isinstance(v, (bool, float)) for v in (x, y, z))`?

sanderr (changes requested):

sanderr (approved):

src/inmanta/ast/type.py
wouterdb: Do we also allow boolean to be cast to int? Never realized.

wouterdb (approved):

src/inmanta/ast/type.py
FloLey: yes we do: https://github.com/inmanta/inmanta-core/blob/master/tests/compiler/test_typing.py#L113

PR #6727, opened 2023-11-10
Issue/build master investigate fedora pytest docker issue

Hugo-Inmanta:
# Description

The use of ` ` or `"` in the venv paths was actually not supported.
In this PR:
- remove faulty `test_pip_binary_when_venv_path_contains_double_quote` (test is failing on fedora-latest and was previously failing silently)
- add validation for the env's path on virtualenv creation (error is raised for `&`, `"` or whitespace)
- add test `test_invalid_chars_in_venv_path` for this behaviour

# Self Check:

- [ ] Changelog entry


src/inmanta/env.py
```diff
@@ -566,24 +566,38 @@ class PythonEnvironment:
...
+
+        :param path: Path to validate.
+        :param path_name: Used to display the type of the path being validated.
+        """
+        if not path:
+            raise ValueError(f"The {path_name} cannot be an empty string.")
+
+        match = PythonEnvironment._invalid_chars_in_path_re.search(path)
+        if match:
+            raise ValueError(f"Invalid character `{match.group()}` in {path_name} {path}.")
```
arnaudsjs, line 599: Why perform this check at the level of the `PythonEnvironment` class? The custom pip binary is only created by the `VirtualEnv` subclass. Wouldn't it be too strict to enforce this character constraint at the level of the `PythonEnvironment` class.

src/inmanta/env.py
```diff
@@ -566,24 +566,38 @@ class PythonEnvironment:
     Inmanta product packages don't change.
     """
 
+    _invalid_chars_in_path_re = re.compile(r"[ \"&]")
```
arnaudsjs, line 569: I can see that having a double quote can cause issues. But I am failing to see why we would also reject the whitespace and ampersand characters? Could you explain?

arnaudsjs (changes requested):

src/inmanta/env.py
Hugo-Inmanta: Moved this down to VirtualEnv

src/inmanta/env.py
Hugo-Inmanta: Whitespace and ampersand are not causing issues after all. Added `$` and backtick char + some docstring
Full slack discussion [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1699604351677279) for reference.

src/inmanta/env.py
```diff
@@ -1094,14 +1094,41 @@ class VirtualEnv(ActiveEnv):
     Creates and uses a virtual environment for this process. This virtualenv inherits from the previously active one.
     """
 
+    _invalid_chars_in_path_re = re.compile(r"[\"$`]")
```
arnaudsjs, line 1097: This makes the set on invalid characters more clear.
```suggestion
    _invalid_chars_in_path_re = re.compile(r'["$`]')
```

src/inmanta/env.py
```diff
@@ -1094,14 +1094,41 @@ class VirtualEnv(ActiveEnv):
...
+        """
+        Check for the presence of the following invalid characters in the given path.
+            - '$' because of a limitation in python venv causing prompt display issues:
+                - '$$' will be replaced by the pid of the process
+                - '${<var>}' will be expanded (possibly to empty string if var is undefined)
+
+            - '`' to prevent command substitution e.g. `whoami` would be interpreted in the prompt.
+
+            - '"' because of a limitation in python venv causing venv activation to fail silently for python<3.12 and
+              a crash for python>=3.12
```
arnaudsjs, line 1118: I think this docstring is missing a bit of context. The essential part of the problem is that this path is injected in the `activate` file of the created venv, which is a bash script. So we need to exclude all special characters that might be interpreted incorrectly in a bash string.

```
The given path is used in the `./bin/activate` file of the created venv without escaping any special characters. As such, we refuse all special characters here that might cause the given path to be interpreted incorrectly:

* $: Character used for variable expansion in bash strings.
* `: Character used to perform command substitution in bash strings.
* ": Character that will be interpreted incorrectly as the end of the string. 
```

arnaudsjs (changes requested):

src/inmanta/env.py
Hugo-Inmanta: Oh, nice !

src/inmanta/env.py
```diff
@@ -1094,14 +1094,38 @@ class VirtualEnv(ActiveEnv):
...
+        :param path: Path to validate.
+        """
+        if not path:
+            raise ValueError("Cannot create virtual environment because the provided path is an empty string.")
+
+        match = VirtualEnv._invalid_chars_in_path_re.search(path)
+        if match:
+            raise ValueError(
+                f"Cannot create virtual environment because the provided path `{path}` contains an"
+                f"invalid character (`{match.group()}`)."
```
arnaudsjs, line 1126:
```suggestion
                f" invalid character (`{match.group()}`)."
```

arnaudsjs (changes requested):

tests/test_env.py
```diff
@@ -555,26 +554,20 @@ def test_override_inmanta_package(tmpvenv_active_inherit: env.VirtualEnv) -> Non
...
-    # Ensure inheritance works correctly
-    assert "inmanta-core" in [elem["name"] for elem in parsed_output]
+    venv_name = f"test{invalid_char}test"
+    venv_dir = os.path.join(tmpdir, venv_name)
+
+    with pytest.raises(ValueError) as excinfo:
+        env.VirtualEnv(venv_dir)
+    assert (
+        f"Cannot create virtual environment because the provided path `{venv_dir}` contains an"
+        f"invalid character (`{invalid_char}`)."
```
Hugo-Inmanta, line 569:
```suggestion
        f" invalid character (`{invalid_char}`)."
```

arnaudsjs (approved):

PR #6730, opened 2023-11-14
Fix bug number cast iso6

FloLey:
# Description

while working on https://github.com/inmanta/inmanta-core/pull/6686 we noticed a bug https://github.com/inmanta/inmanta-core/pull/6686#discussion_r1386642720 . This PR fixes this bug

# Self Check:

- [ ] Changelog entry


FloLey:
@sanderr I couldn't merge [this one](https://github.com/inmanta/inmanta-core/pull/6722) as it was on the wrong base.

PR #6734, opened 2023-11-15
issue/6731:dict path: allow matching float with corresponding int

### LINKED ISSUE #6731: dict path: allow matching float with corresponding int
`1.0` is a float. When it appears in a dict, the dict path library does not support matching it as `key=1`, only `key=1.0`. Since `1 == 1.0` this should be supported. I expect changing `NormalValue.matches` should fix this.

FloLey:
# Description

1.0 is a float. When it appears in a dict, the dict path library does not support matching it as key=1, only key=1.0. Since 1 == 1.0 this should be supported. I expect changing NormalValue.matches should fix this.

closes #6731 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/6731-fix-dict-path-matching-float-int.yml
+++ b/changelogs/unreleased/6731-fix-dict-path-matching-float-int.yml
@@ -1,0 +1,4 @@
+description: Allow matching floats with corresponding ints in dict path
+issue-nr: 6734
+change-type: patch
+destination-branches: [master, iso6]
--- a/changelogs/unreleased/6731-fix-dict-path-matching-float-int.yml
+++ b/changelogs/unreleased/6731-fix-dict-path-matching-float-int.yml
@@ -2,3 +2,6 @@ description: Allow matching floats with corresponding ints in dict path
 issue-nr: 6734
 change-type: patch
 destination-branches: [master, iso6]
+sections:
+  bugfix: "Fix the handling of numeric keys in dict paths:
+  floating-point numbers and their integer equivalents are treated as the same key."
--- a/src/inmanta/util/dict_path.py
+++ b/src/inmanta/util/dict_path.py
@@ -115,6 +115,7 @@ class NormalValue(DictPathValue):
     def __init__(self, value: str) -> None:
         super(DictPathValue, self).__init__()
         self._value: str = value
+        self._numeric_value: Optional[float] = self._try_parse_numeric(value)
 
     def escape(self) -> str:
         return WildDictPath.PATTERN_SPECIAL_CHARACTER.sub(r"\\\1", self._value)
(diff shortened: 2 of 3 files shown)


src/inmanta/util/dict_path.py
```diff
@@ -122,6 +122,15 @@ def escape(self) -> str:
...
             return False
+
+        # Attempt numeric comparison if both values can be converted to floats
+        try:
+            if float(self._value) == float(value):
+                return True
+        except ValueError:
+            pass
+
+        # Fallback to string comparison
```
arnaudsjs, line 133: This implementation will always attempt a numeric comparison, also when the provided value has the type `str`. I don't think that makes much sense. Wouldn't it be better to only attempt a numeric comparison when the provided value is numeric?

For example, consider the following dict: `{"1": "val"}`. If we apply the dict path expression `1.0` on this dict, do we want it to match? I personally think not. But I do think the dict path expression should match `{1: "val"}`. Do you agree?

arnaudsjs (changes requested):

src/inmanta/util/dict_path.py
```diff
@@ -122,6 +122,15 @@ def escape(self) -> str:
...
         return self._value == str(value)
```
Hugo-Inmanta, line 134: I think this is slightly more efficient when both values are float and different + it reads nicer
```suggestion
        # Attempt numeric comparison if both values can be converted to floats
        try:
            return float(self._value) == float(value):
        except ValueError:
            # Fallback to string comparison
            return self._value == str(value)

        
```

Hugo-Inmanta (changes requested):

changelogs/unreleased/6731-fix-dict-path-matching-float-int.yml
```diff
@@ -0,0 +1,4 @@
+description: Allow matching floats with corresponding ints in dict path
+issue-nr: 6734
+change-type: patch
+destination-branches: [master, iso6]
```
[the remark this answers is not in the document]
arnaudsjs: Yes. I would add a changelog entry for this.

src/inmanta/util/dict_path.py
FloLey: indeed but I changed it :see_no_evil:

src/inmanta/util/dict_path.py
FloLey: The values in the dictpaths are always provided as strings aren't they? I changed the code and added some test. I think it is better now.

arnaudsjs: Yes. In a dichtpath expression we don't have a way to indicate the difference between a string and an integer/float key.

src/inmanta/util/dict_path.py
```diff
@@ -115,14 +115,29 @@ class NormalValue(DictPathValue):
...
+        # Perform a numeric comparison only if the value is an int/float and
+        # The key in the dictpath can be interpreted as an int/float
+        if self._numeric_value is not None and isinstance(value, (int, float)):
+            return float(self._value) == value
+
+        # Fallback to string comparison for other types
+        return str(self._value) == str(value)
+
+    @staticmethod
+    def _try_parse_numeric(value: object) -> Optional[float]:
```
arnaudsjs, line 136:
```suggestion
    def _try_parse_numeric(value: str) -> Optional[float]:
```

arnaudsjs (changes requested):

Hugo-Inmanta (changes requested):

arnaudsjs (approved):

Hugo-Inmanta (approved):
LGTM :+1:

PR #6463, opened 2023-09-06
Issue/6462 pydantic v2

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/inmanta-core/issues/6462

# Self Check:

Strike through any lines that are not applicable (`~~line~~`) then check the box

- [ ] Attached issue to pull request
- [ ] Changelog entry
- [ ] Type annotations are present
- [ ] Code is clear and sufficiently documented
- [ ] No (preventable) type errors (check using make mypy or make mypy-diff)
- [ ] Sufficient test cases (reproduces the bug/tests the requested feature)
- [ ] Correct, in line with design
- [ ] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )
- [ ] If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)

Issue #6462, opened 2023-09-05
pydantic v2


sanderr:
# Support pydantic v2

Bart already did the bulk of this. Someone needs to take over the lead for the final sprint: review and discuss with other reviewers, then address own comments. Bart is available for questions but otherwise the person who picks this up has ownership/authority.

## Original description

Use the conversion tool to convert core and lsm to pydantic v2, then get the basic functionality to work:
- run `lsm` module test cases or set it up on an orchestrator. We don't care too much about edge cases, just the general functionality
- run some `connect` moudule tests that use the config module (ask solutions team)
- investigate if we can make the lsm/config module be compatible with both v1 and v2
- Fixes are PoC grade, does not need to be clean. Keep in mind the target: we want to find a migration path: modules first / core first / ...
- keep notes

The goal of this ticket is to get a concrete idea of the pain points in the conversion and how much of an effort it will be to do it properly. 

Timebox this ticket to 2 days, then report results.

Hugo-Inmanta:
Migration notes:
https://github.com/inmanta/inmanta-core/blob/58778a30d4a7bdf1be4f3546fdb21108eebd7f00

Tentative PRs:
- https://github.com/inmanta/inmanta-core/pull/6463
- https://github.com/inmanta/inmanta-lsm/pull/1425
- https://github.com/inmanta/lsm/pull/640
/pydanticv2_migration_notes.md

PR #6739, opened 2023-11-16
Pydantic v2

sanderr:
# Description

Migrate to Pydantic v2. This is the final PR, which is a follow-up from the PoC PRs #6463 and #6685.

Current status: I still need to refine a few details (marked with TODO), and review the full mypy diff (it's big). But I think any changes from here on out should be relatively minor so it makes sense to open it for review now. inmanta-lsm PR will follow soon.

part of #6462

# Self Check:

- [ ] Changelog entry


src/inmanta/util/__init__.py
```diff
@@ -476,7 +477,7 @@ def _custom_json_encoder(o: object) -> Union[ReturnTypes, "JSONSerializable"]:
     if isinstance(o, JSONSerializable):
         return o.json_serialization_step()
 
-    if isinstance(o, uuid.UUID):
+    if isinstance(o, (uuid.UUID, Url)):
```
Hugo-Inmanta, line 480: Why do we need this?

src/inmanta/protocol/openapi/converter.py
```diff
@@ -150,62 +149,109 @@ def get_swagger_html(self, openapi_spec: str) -> str:
...
         return schema
 
-    def _handle_pydantic_model(self, type_annotation: Type, by_alias: bool = True) -> Schema:
-        # JsonSchema stores the model (and sub-model) definitions at #/definitions,
-        # but OpenAPI requires them to be placed at "#/components/schemas/"
-        # The ref_prefix changes the references, but the actual schemas are still at #/definitions
-        schema = model_schema(type_annotation, by_alias=by_alias, ref_prefix=self.ref_prefix)
-        if "definitions" in schema.keys():
-            definitions: Dict[str, Dict[str, object]] = schema.pop("definitions")
+    def _handle_pydantic_model(self, type_annotation: BaseModel, by_alias: bool = True) -> Schema:
```
Hugo-Inmanta, line 216: It seems the `by_alias` parameter is not used, is this intentional ?

src/inmanta/module.py
```diff
@@ -263,7 +264,7 @@ def __init__(self, msg: str, validation_error: Optional[ValidationError] = None)
     def _extend_msg_with_validation_information(cls, msg: str, validation_error: ValidationError) -> str:
         errors = validation_error.errors()
         if errors:
-            msg += "\n" + textwrap.indent(display_errors(errors), " " * 2)
+            msg += "\n" + textwrap.indent(str(validation_error), " " * 2)
```
sanderr, line 267: This change has an effect on how the error message looks. I think it's acceptable but if you disagree I can fix it. It would require more manual formatting than before.

Previous output:
```
Metadata defined in ./project.yml is invalid:
  pip -> use-system-config
    Input should be a valid boolean, unable to interpret input (type=bool_parsing)
```

New output:
```
Metadata defined in ./project.yml is invalid:
  1 validation error for ProjectMetadata
  pip.use-system-config
    Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='hello', input_type=str]
      For further information visit https://errors.pydantic.dev/2.5/v/bool_parsing
```

CC @arnaudsjs @Hugo-Inmanta

src/inmanta/validation_type.py
```diff
@@ -16,13 +16,58 @@
...
+        is a type generator function.
+    :param type_name: Optionally, the human-readable name of the type, for error messages.
+
+    :raises ValueError: When invalid parameters are passed.
+    :raises TypeError: When the given validation parameters are not valid with respect to the requested type.
+    """
+    type_name = type_name if type_name is not None else repr(base_type)
+
+    custom_annotations: list[object] = []
+    validation_parameters = dict(validation_parameters) if validation_parameters is not None else {}
```
Hugo-Inmanta, line 51: I'm curious, why cast to `dict` here ?

tests/test_pydantic.py
```diff
@@ -0,0 +1,52 @@
...
+    Contact: code@inmanta.com
+"""
+import datetime
+import uuid
+from typing import Any, Dict, Union
+
+import pydantic
+
+
+def test_method_validation() -> None:
```
Hugo-Inmanta, line 25: Are these tests still relevant?

src/inmanta/module.py
Hugo-Inmanta: This is fine by me :+1:

tests/moduletool/test_install.py
```diff
@@ -1204,7 +1204,7 @@ def test_module_install_logging(local_module_package_index: str, snippetcompiler
         python_requires=v2_requirements,
         install_project=False,
         project_requires=[
-            module.InmantaModuleRequirement.parse("std==4.1.7"),
+            module.InmantaModuleRequirement.parse("std==4.2.1"),
```
Hugo-Inmanta, line 1207: I haven't looked too much into this, but why not `4.3.3` ? This looks inconsistent with the previous behaviour ?

Hugo-Inmanta (changes requested):

setup.py
```diff
@@ -24,8 +24,7 @@
     # pip>=21.3 required for editable pyproject.toml + setup.cfg based install support
     "pip>=21.3",
     "ply~=3.0",
-    # lower bound because of pydantic/pydantic#5821
-    "pydantic>=1.10.8,<2",
+    "pydantic~=2.4",
```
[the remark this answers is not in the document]
sanderr: Tests have been run against both versions so I see no reason to require 2.5.1 specifically.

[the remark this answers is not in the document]
Hugo-Inmanta: The purpose of `requirements.txt` vs `setup.py` wasn't clear in my mind. From what I gathered:
- `setup.py` is used for distribution when packaging the core package so we want the lowest compatible bound to have the widest compatibility possible
- `requirements.txt` is used during development when working on the core package so we want the latest available version. (Dependabot makes sure of this)

----
Is this correct ? And related question, is it possible that having different version in both files lead to inconsistencies ?

src/inmanta/protocol/openapi/model.py
```diff
@@ -65,11 +67,30 @@ class Schema(BaseModel):
...
+        Convert null in anyOf to the `nullable` property.
+        """
+        if self.anyOf is not None:
+            without_null: abc.Sequence["Schema"] = [e for e in self.anyOf if e.type != "null"]
+            if len(without_null) != len(self.anyOf):
+                if len(without_null) == 1:
+                    # promote single child, which has already been validated at this point
+                    child: "Schema" = without_null[0]
+                    for field in child.model_fields_set:
+                        setattr(self, field, getattr(child, field))
```
sanderr, line 87: I'm not a huge fan of this but it's the best/cleanest approach I could think of. Suggestions welcome.

arnaudsjs (changes requested):

sanderr:
All TODO's I had marked in this PR are now done.

setup.py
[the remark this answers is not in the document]
sanderr: This is indeed correct. `requirements.txt` is what we call a lock file. It locks dependencies to a specific version, but only when explicitly requested with `-r`. It is used for repetitive testing. `setup.py`, `setup.cfg` or `pyproject.toml` specify the project's dependencies. These are also referred to (at least by dependabot) as the manifest files. They represent the constraints that have to be satisfied to install the package, be it from source or from PyPi. No manual `-r` is required for them to take effect.

> is it possible that having different version in both files lead to inconsistencies

Depends how you install. If you do it with one command (`pip install -e . -r requirements.txt`), this would fail, so it would never be merged. I think tox does it like that.

src/inmanta/data/__init__.py
```diff
@@ -1475,7 +1475,7 @@ def get_valid_field_names(cls) -> List[str]:
     def _get_names_of_primary_key_fields(cls) -> List[str]:
         return [name for name, value in cls.get_field_metadata().items() if value.is_part_of_primary_key()]
 
-    def _get_filter_on_primary_key_fields(self, offset: int = 1) -> Tuple[str, List[Any]]:
+    def _get_filter_on_primary_key_fields(self, offset: int = 1) -> Tuple[str, List[object]]:
```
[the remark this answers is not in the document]
sanderr: I consider `Any` to be bad practice. It boils down to "don't bother checking this value or anything I do with it", while `object` means "this value can be anything so don't allow me to do anything with it unless it works for any object". I see both used often in a context where a more concrete type can not be specified (even in the pydantic docs) but `Any` then means "don't bother" while `object` represents "we don't know the type".

I believe `Any` exists to allow gradual migration to static typing. You could have `Any` types on the frontier of the typed interface: anything on the other side will not be type checked. For this use case, if used carefully, I think it's really neat. But in my experience it's all too often confused and used inappropriately.

I did a presentation about it in one of the team weeks. You can find the slides [here](https://docs.google.com/presentation/d/1nkGrRMn9aiOODmzEDL30DoFu1_Cfkc0XMiNPStuZPUw/edit#slide=id.p3) if you're interested.

src/inmanta/util/__init__.py
sanderr: Good question actually, @bartv do you know? I think it can't hurt either way, but I'm not sure what changed to require it now.

bartv: I think because before these Urls were strings and not they are their own class.

src/inmanta/protocol/openapi/converter.py
sanderr: No, good catch, will fix

src/inmanta/validation_type.py
sanderr: To make a copy because we modify it later on

tests/test_pydantic.py
sanderr: I think not. Thanks.

tests/moduletool/test_install.py
sanderr: How so, inconsistent? I mainly didn't want to pick the latest version because then we wouldn't really be testing anything.

src/inmanta/types.py
```diff
@@ -18,46 +18,62 @@
...
 
 
-class StrictNonIntBool(object):
+@dataclass
+class PythonRegex:
     """
-    StrictNonIntBool to allow for bools which are not type-coerced and that are not a subclass of int
-    Based on StrictBool from pydantic
+    A pydantic regex type for constrained strings that use the old "regex" parameter instead of the new "pattern".
+    Added for compatibility with
```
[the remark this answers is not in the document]
sanderr: Right, I was writing this when I realized it didn't really have a purpose anymore. Forgot to update.

tests/test_pydantic.py
```diff
@@ -0,0 +1,52 @@
...
+
+def test_float_str() -> None:
+    """Test parsing version numbers from yaml files"""
+
+    class Value(pydantic.BaseModel):
+        version: str
+
+        @pydantic.field_validator("version", mode="before")
+        @classmethod
+        def is_pep440_version(cls, v: Union[str, float]) -> str:
```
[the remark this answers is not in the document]
sanderr: Deleted the file

src/inmanta/types.py
```diff
@@ -18,46 +18,62 @@
...
     """
 
-    @classmethod
-    def __get_validators__(cls) -> "types.CallableGenerator":
-        yield cls.validate
+    regex: str
+
+    def __get_pydantic_core_schema__(self, source_type: object, handler: GetCoreSchemaHandler) -> CoreSchema:
```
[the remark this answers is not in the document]
sanderr: This was copied from the docs. It seemed more complicated than required to me but we decided to stick with the recommended approach. Based on your comment I now wanted to add a link to the docs, and it looks like they introduced a way cleaner approach in last week's release, so I'll just migrate to that instead.

tests/moduletool/test_install.py
Hugo-Inmanta: Inconsistent is maybe not the right word here, I mean the version previously used was `4.1.7` everywhere.
The version in some tests have been bumped from `4.1.7` to `4.2.1`  and to `4.3.3` in some others (e.g. `test_moduletool_list`).

> then we wouldn't really be testing anything.

How's so ?

src/inmanta/types.py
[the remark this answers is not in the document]
sanderr: Turns out the new mechanism has some limitations after all. I reworked it and I think I prefer it now, but let me know if you'd like me to revert and add some documentation.

src/inmanta/protocol/openapi/model.py
Hugo-Inmanta: I don't fully understand what we're trying to achieve here, I'm not familiar with openapi

Hugo-Inmanta (approved):

src/inmanta/protocol/openapi/model.py
arnaudsjs: The same for me. This needs more documentation.

arnaudsjs (approved):

sanderr:
@arnaudsjs @Hugo-Inmanta all done.

src/inmanta/util/__init__.py
```diff
@@ -498,7 +498,7 @@ def _custom_json_encoder(o: object) -> Union[ReturnTypes, "JSONSerializable"]:
     if isinstance(o, BaseModel):
         return o.model_dump(by_alias=True)
 
-    if dataclasses.is_dataclass(o):
+    if dataclasses.is_dataclass(o) and not isinstance(o, type):
```
arnaudsjs, line 501: Why is the second condition required?

arnaudsjs (approved):

Hugo-Inmanta (approved):

src/inmanta/util/__init__.py
sanderr: See [here](https://github.com/inmanta/inmanta-lsm/pull/1491/#discussion_r1399327453)

PR #6744, opened 2023-11-17
Issue/6691 catch all arguments 4

edvgui:
# Description

Variation of https://github.com/inmanta/inmanta-core/pull/6708 where the type validation uses some custom types, which are only defined for plugins (it was already the case implicitly, this is now more explicit)

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/plugins.py
+++ b/src/inmanta/plugins.py
@@ -295,7 +295,8 @@ class PluginIO:
             )
 
         if self.type_expression in PLUGIN_TYPES:
-            return PLUGIN_TYPES[self.type_expression]
+            self._resolved_type = PLUGIN_TYPES[self.type_expression]
+            return self._resolved_type
 
         plugin_line: Range = Range(plugin.location.file, plugin.location.lnr, 1, plugin.location.lnr + 1, 1)
         locatable_type: LocatableString = LocatableString(self.type_expression, plugin_line, 0, resolver)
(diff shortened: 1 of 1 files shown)


src/inmanta/plugins.py
```diff
@@ -234,6 +228,25 @@ def resolve_type(locatable_type: LocatableString, resolver: Namespace) -> Option
...
+
+
+# Define some types which are only used in the context of plugins.
+PLUGIN_TYPES = {
+    "any": inmanta_type.Type(),  # Any value will pass validation
+    "expression": inmanta_type.Type(),  # Any value will pass validation
+    "null": inmanta_type.NullableType(Nothing()),  # Only NoneValue will pass validation
+}
+
+
```
edvgui, line 249: @sanderr this is what I meant with "Can I define custom types for the plugin validation".  Wdyt?

sanderr: I think error messages may be messed up if you do it like this because you don't override the string representation methods (don't recall their name off the top of my head).

edvgui: The validation of arguments happens before wrapping them and the validation of the returned value happens after wrapping it, so I think it would work.  I will take the other comments though

sanderr: I'm not sure what you mean by "wrapping them". As to my other comments, I'd say have a look, but it is entirely possible that I'm wrong, it depens on how and where this is used exactly.

PR #6751, opened 2023-11-21
Added regex_string method

sanderr:
# Description

Added `regex_string` method

replaces inmanta/std#481, see [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1700554456479609)

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/validation_type.py
+++ b/src/inmanta/validation_type.py
@@ -26,6 +26,34 @@ from inmanta.stable_api import stable_api
 from inmanta.types import PrimitiveTypes
 
 
+def _regex_validator(regex: str) -> pydantic.AfterValidator:
+    """
+    Returns an AfterValidator for regex validation.
+    """
+    # python-re engine can only be selected on model/TypeAdapter level
+    # => add custom validator that delegates to TypeAdapter
+    return pydantic.AfterValidator(
+        pydantic.TypeAdapter(
+            Annotated[
+                str,
+                pydantic.StringConstraints(pattern=regex),
+            ],
+            config=pydantic.ConfigDict(regex_engine="python-re"),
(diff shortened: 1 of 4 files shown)


src/inmanta/validation_type.py
```diff
@@ -26,6 +26,34 @@
...
+    """
+    Returns an AfterValidator for regex validation.
+    """
+    # python-re engine can only be selected on model/TypeAdapter level
+    # => add custom validator that delegates to TypeAdapter
+    return pydantic.AfterValidator(
+        pydantic.TypeAdapter(
+            Annotated[
+                str,
+                pydantic.StringConstraints(pattern=regex),
```
sanderr, line 39: Made it use `StringConstraints` instead of `constr`

bartv (approved):

src/inmanta/validation_type.py
```diff
@@ -26,6 +26,34 @@
...
+            ],
+            config=pydantic.ConfigDict(regex_engine="python-re"),
+        ).validate_python
+    )
+
+
+@stable_api
+def regex_string(regex: str) -> object:
```
edvgui, line 47: Why not this?
```suggestion
def regex_string(regex: str) -> type:
```
Or even this?
```suggestion
def regex_string(regex: str) -> Type[str]:
```

edvgui (approved):

src/inmanta/validation_type.py
edvgui: Too bad

arnaudsjs (changes requested):
See comment

arnaudsjs (approved):

PR #6737, opened 2023-11-16
Update behavior of `inmanta module install`

FloLey:
# Description

Update the pip config used by inmanta module install to behave as follows:
1. raise an error stating that the command is no longer supported.
2. Offer the suggestion to run pip install -e . to install a module in development mode
3. Offer the suggestion to run inmanta module build followed by pip install if the user really does not want an editable install.
4. The inmanta module build command should use use-system-config=true. It is assumed safe because it only installs build dependencies.

closes #6717 

# Self Check:

Strike through any lines that are not applicable (`~~line~~`) then check the box

- [x] Attached issue to pull request
- [x] Changelog entry
- [x] Type annotations are present
- [x] Code is clear and sufficiently documented
- [x] No (preventable) type errors (check using make mypy or make mypy-diff)
- [x] Sufficient test cases (reproduces the bug/tests the requested feature)
- [x] Correct, in line with design
- [x] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )
- [ ] If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)

PR #6764, opened 2023-11-22
Fix race condition where dist directory is not cleaned up

arnaudsjs:
# Description

#6737 introduced a race condition where the `test_module_install` test case creates a dist directory in the `./data/modules_v2/minimalv2module` directory, which makes the `test_build_v2_module_set_output_directory` test case fail. The latter test case has an assertion to make sure no dist directory exists.

# Self Check:

- [x] Changelog entry


FloLey (approved):
:+1:

Issue #6473, opened 2023-09-08
Document all API arguments


arnaudsjs:
This ticket should make sure that all the arguments of all API endpoints are documented in the `methods.py` and `methods_v2.py`.

Initial timebox: 1 day

PR #6771, opened 2023-11-24
update

FloLey:
# Description

document all the missing api arguments

closes https://github.com/inmanta/inmanta-core/issues/6473

# Self Check:

- [ ] Changelog entry


Hugo-Inmanta:
Only `fileservice.py` was changed compared to master, right ?

FloLey:
@Hugo-Inmanta Yes, the implementation of the diff function is not the same on master and iso6

Hugo-Inmanta (approved):

PR #6772, opened 2023-11-24
issue/6525-index-number-float-v2 

FloLey:
# Description

1. create the float type which is like a float in python. Update documentation to mention number will be deprecated

2. add testcases for following scenario and fix it

```
entity A:
    float x
end

implement A using std::none

index A(x)


std::print(A(x=0) == A(x=0.0))
```
This will still not work with the number type

3. index lookup on float with int should work

closes https://github.com/inmanta/inmanta-core/issues/6526

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/6526-index-float-int.yml
+++ b/changelogs/unreleased/6526-index-float-int.yml
@@ -4,6 +4,6 @@ issue-repo: inmanta-core
 change-type: minor
 destination-branches: [master]
 sections:
-  feature: "Introduce the 'float' type which is like a float in python"
-  deprecation-note: "Deprecate the 'number' type. Instead the 'int' or 'float' type should be used"
+  feature: "Introduce the 'float' type for floating point numbers"
+  deprecation-note: "Deprecate the 'number' type. Use the 'int' or 'float' type instead"
 
--- a/changelogs/unreleased/6526-index-float-int.yml
+++ b/changelogs/unreleased/6526-index-float-int.yml
@@ -4,4 +4,6 @@ issue-repo: inmanta-core
 change-type: minor
 destination-branches: [master]
 sections:
-  feature: "Introduce an alias 'float' for the 'number' type in the inmanta language"
+  feature: "Introduce the 'float' type which is like a float in python"
+  deprecation-note: "Deprecate the 'number' type. Instead the 'int' or 'float' type should be used"
+
--- a/changelogs/unreleased/6526-index-float-int.yml
+++ b/changelogs/unreleased/6526-index-float-int.yml
@@ -6,5 +6,4 @@ destination-branches: [master]
 sections:
   feature: "Introduce the 'float' type for floating point numbers"
   deprecation-note: "Deprecate the 'number' type. Use the 'int' or 'float' type instead"
-  bugfix: "Fix a bug where numbers where cast to int instead of float"
 
--- a/changelogs/unreleased/6526-index-float-int.yml
+++ b/changelogs/unreleased/6526-index-float-int.yml
@@ -1,0 +1,7 @@
+description: number type should no longer differentiate between integer and decimal notation when used in indexes.
+issue-nr: 6526
+issue-repo: inmanta-core
+change-type: minor
+destination-branches: [master]
+sections:
+  feature: "Introduce an alias 'float' for the 'number' type in the inmanta language"
--- a/changelogs/unreleased/6526-index-float-int.yml
+++ b/changelogs/unreleased/6526-index-float-int.yml
@@ -6,4 +6,5 @@ destination-branches: [master]
 sections:
   feature: "Introduce the 'float' type for floating point numbers"
   deprecation-note: "Deprecate the 'number' type. Use the 'int' or 'float' type instead"
+  bugfix: "Fix a bug where numbers where cast to int instead of float"
 
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -127,6 +127,10 @@ Primitive types
 
 The basic primitive types are ``string``, ``float``, ``int`` or ``bool``. These basic types also support type casts:
 
+.. note::
+    To initialize a variable as a float, the value should either include a decimal point or be explicitly converted to a float type.
+
+
 .. code-block:: inmanta
 
     assert = true
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -125,18 +125,14 @@ Literal values can be assigned to variables
 Primitive types
 ==============================
 
-The basic primitive types are ``string``, ``number``, ``float``, ``int`` or ``bool``. These basic types also support type casts:
-
-.. note::
-    ``number`` will be deprecated in the future
-
+The basic primitive types are ``string``, ``float``, ``int`` or ``bool``. These basic types also support type casts:
 
 .. code-block:: inmanta
 
     assert = true
     assert = int("1") == 1
     assert = float("1.2") == 1.2
-    assert = float(true) == 1
+    assert = int(true) == 1
     assert = bool(1.2) == true
     assert = bool(0) == false
     assert = bool(null) == false
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -128,7 +128,7 @@ Primitive types
 The basic primitive types are ``string``, ``float``, ``int`` or ``bool``. These basic types also support type casts:
 
 .. note::
-    To initialize a variable as a float, the value should either include a decimal point or be explicitly converted to a float type.
+    To initialize or assign a float, the value should either include a decimal point or be explicitly converted to a float type.
 
 
 .. code-block:: inmanta
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -125,14 +125,18 @@ Literal values can be assigned to variables
 Primitive types
 ==============================
 
-The basic primitive types are ``string``, ``number``, ``int`` or ``bool``. These basic types also support type casts:
+The basic primitive types are ``string``, ``number``, ``float``, ``int`` or ``bool``. These basic types also support type casts:
+
+.. note::
+    ``number`` will be deprecated in the future
+
 
 .. code-block:: inmanta
 
     assert = true
     assert = int("1") == 1
-    assert = number("1.2") == 1.2
-    assert = number(true) == 1
+    assert = float("1.2") == 1.2
+    assert = float(true) == 1
     assert = bool(1.2) == true
     assert = bool(0) == false
     assert = bool(null) == false
@@ -164,7 +168,7 @@ For example
     typedef mac_addr as string matching /([0-9a-fA-F]{2})(:[0-9a-fA-F]{2}){5}$/
 
 
-Lists of primitive types are also primitive types: ``string[]``, ``number[]``, ``bool[]`` or ``mac_addr[]``
+Lists of primitive types are also primitive types: ``string[]``, ``float[]``, ``number[]``, ``bool[]`` or ``mac_addr[]``
 
 ``dict`` is the primitive type that represents a dictionary, with string keys. Dict values can be accessed using the ``[]`` operator. All members of a dict have to be set when the dict is constructed. e.g.
 
@@ -700,6 +704,15 @@ For indices on relations (instead of attributes) an alternative syntax can be us
     a = File[host=vm1, path="/etc/passwd"]  # normal index lookup
     b = vm1.files[path="/etc/passwd"]  # selector style index lookup
     # a == b
+
+.. note::
+    The use of ``float`` (or ``number``) as part of index properties is
+    generally discouraged. This is due to the reliance of index matching on precise equality,
+    while floating-point numbers are represented with an inherent imprecision.
+    If floating-point attributes are used in an index, it is crucial to handle arithmetic
+    operations with caution to ensure the accuracy of the attribute values for index operations.
+
+
 
 
 For loop
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -92,8 +92,8 @@ Literal values can be assigned to variables
 
 .. code-block:: inmanta
 
-    var1 = 1 # assign an integer, var1 contains now an integer
-    var2 = 3.14 # assign a float, var2 now contains a float
+    var1 = 1 # assign an integer, var1 contains now a number
+    var2 = 3.14 # assign a float, var2 also contains a number
     var3 = "This is a string" # var3 contains a string
     var4 = r"This is a raw string" # var4 contains a raw string
 
@@ -125,18 +125,14 @@ Literal values can be assigned to variables
 Primitive types
 ==============================
 
-The basic primitive types are ``string``, ``float``, ``int`` or ``bool``. These basic types also support type casts:
-
-.. note::
-    For legacy reason the ``number`` type exists. It is just an alias for ``float``
-
+The basic primitive types are ``string``, ``number``, ``int`` or ``bool``. These basic types also support type casts:
 
 .. code-block:: inmanta
 
     assert = true
     assert = int("1") == 1
-    assert = float("1.2") == 1.2
-    assert = float(true) == 1
+    assert = number("1.2") == 1.2
+    assert = number(true) == 1
     assert = bool(1.2) == true
     assert = bool(0) == false
     assert = bool(null) == false
@@ -168,7 +164,7 @@ For example
     typedef mac_addr as string matching /([0-9a-fA-F]{2})(:[0-9a-fA-F]{2}){5}$/
 
 
-Lists of primitive types are also primitive types: ``string[]``, ``float[]``, ``bool[]`` or ``mac_addr[]``
+Lists of primitive types are also primitive types: ``string[]``, ``number[]``, ``bool[]`` or ``mac_addr[]``
 
 ``dict`` is the primitive type that represents a dictionary, with string keys. Dict values can be accessed using the ``[]`` operator. All members of a dict have to be set when the dict is constructed. e.g.
 
@@ -704,14 +700,6 @@ For indices on relations (instead of attributes) an alternative syntax can be us
     a = File[host=vm1, path="/etc/passwd"]  # normal index lookup
     b = vm1.files[path="/etc/passwd"]  # selector style index lookup
     # a == b
-
-.. note::
-    The use of ``float`` (or ``number``) as part of index properties is
-    generally discouraged. This is due to the reliance of index matching on precise equality,
-    while floating-point numbers are represented with an inherent imprecision.
-    If floating-point attributes are used in an index, it is crucial to handle arithmetic
-    operations with caution to ensure the accuracy of the attribute values for index operations.
-
 
 
 For loop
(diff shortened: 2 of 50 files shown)


changelogs/unreleased/6526-index-float-int.yml
```diff
@@ -0,0 +1,9 @@
+description: number type should no longer differentiate between integer and decimal notation when used in indexes.
+issue-nr: 6526
+issue-repo: inmanta-core
+change-type: minor
+destination-branches: [master]
+sections:
+  feature: "Introduce the 'float' type which is like a float in python"
+  deprecation-note: "Deprecate the 'number' type. Instead the 'int' or 'float' type should be used"
```
sanderr, line 8: I think we should try to avoid passive voice
```suggestion
  deprecation-note: "Deprecate the 'number' type. Use the 'int' or 'float' type instead"
```

changelogs/unreleased/6526-index-float-int.yml
```diff
@@ -0,0 +1,9 @@
+description: number type should no longer differentiate between integer and decimal notation when used in indexes.
+issue-nr: 6526
+issue-repo: inmanta-core
+change-type: minor
+destination-branches: [master]
+sections:
+  feature: "Introduce the 'float' type which is like a float in python"
```
sanderr, line 7: I would not make a comparison with Python here, just say what it is.
```suggestion
  feature: "Introduce the 'float' type for floating point numbers"
```

docs/language.rst
```diff
@@ -125,14 +125,18 @@ Literal values can be assigned to variables
 Primitive types
 ==============================
 
-The basic primitive types are ``string``, ``number``, ``int`` or ``bool``. These basic types also support type casts:
+The basic primitive types are ``string``, ``number``, ``float``, ``int`` or ``bool``. These basic types also support type casts:
+
+.. note::
+    ``number`` will be deprecated in the future
```
sanderr, line 131: It is already deprecated in this PR, isn't it? I would not mention deprecated types in the list above.

docs/language.rst
```diff
@@ -125,14 +125,18 @@ Literal values can be assigned to variables
...
+
 
 .. code-block:: inmanta
 
     assert = true
     assert = int("1") == 1
-    assert = number("1.2") == 1.2
-    assert = number(true) == 1
+    assert = float("1.2") == 1.2
+    assert = float(true) == 1
```
sanderr, line 139:
```suggestion
    assert = int(true) == 1
```
I would not use `float` for this one:
- it makes little sense, I can't think of any use case for converting a bool to float
- the comparison to `== 1` in a casting section might make the reader believe that you would get exactly that, while you would actually get 1.0.

docs/language.rst
```diff
@@ -164,7 +168,7 @@ For example
     typedef mac_addr as string matching /([0-9a-fA-F]{2})(:[0-9a-fA-F]{2}){5}$/
 
 
-Lists of primitive types are also primitive types: ``string[]``, ``number[]``, ``bool[]`` or ``mac_addr[]``
+Lists of primitive types are also primitive types: ``string[]``, ``float[]``, ``number[]``, ``bool[]`` or ``mac_addr[]``
```
sanderr, line 171:
```suggestion
Lists of primitive types are also primitive types: ``string[]``, ``float[]``, ``bool[]`` or ``mac_addr[]``
```

src/inmanta/ast/__init__.py
```diff
@@ -382,6 +387,9 @@ def get_type(self, typ: LocatableString) -> "Type":
             else:
                 raise TypeNotFoundException(typ, self)
         elif name in self.primitives:
+            if name == "number":
+                msg = "Type 'number' is deprecated, use 'float' or 'int' instead (%s)" % str(typ.location)
+                warnings.warn(TypeDeprecationWarning(msg))
```
sanderr, line 392: Did you verify that you only get the warning once, no matter how many number types you use?

src/inmanta/ast/entity.py
```diff
@@ -381,7 +381,16 @@ def add_to_index(self, instance: Instance) -> None:
...
         """
-        attributes = {k: repr(v.get_value()) for (k, v) in instance.slots.items() if v.is_ready()}
+        attributes = {}
+        for k, v in instance.slots.items():
+            if v.is_ready():
+                value = v.get_value()
+                # Check if the value is an integer,
+                # and convert to float if necessary
+                if isinstance(value, int):
+                    value = float(value)
```
sanderr, line 391: We should not convert all ints to floats, only if the attribute type is float.

src/inmanta/ast/entity.py
```diff
@@ -427,8 +436,10 @@ def lookup_index(
             raise NotFoundException(
                 stmt, self.get_full_name(), "No index defined on %s for this lookup: " % self.get_full_name() + str(params)
             )
-
-        key = ", ".join(["%s=%s" % (k, repr(v)) for (k, v) in sorted(params, key=lambda x: x[0])])
+        # Convert integers to floats in the key, similar to add_to_index
+        key = ", ".join(
+            ["%s=%s" % (k, repr(float(v) if isinstance(v, int) else v)) for (k, v) in sorted(params, key=lambda x: x[0])]
```
sanderr, line 441: Same here

src/inmanta/ast/type.py
```diff
@@ -276,11 +301,17 @@ def __init__(self) -> None:
...
-            return False
-        if not isinstance(value, numbers.Integral):
+        """
+        Validate the given value to check if it satisfies the constraints
+        associated with this type
+        """
+        if isinstance(value, AnyType):
+            return True
+
+        if not isinstance(value, int):
```
sanderr, line 311: What does the change from `numbers.Ingegral` -> `int` entail? Is this breaking?

src/inmanta/ast/type.py
```diff
@@ -265,6 +262,34 @@ def type_string_internal(self) -> str:
...
+
+    def validate(self, value: Optional[object]) -> bool:
+        """
+        Validate the given value to check if it satisfies the constraints
+        associated with this type
+        """
+        if isinstance(value, AnyType):
+            return True
+
+        if not isinstance(value, numbers.Number):
```
sanderr, line 284: Shouldn't this be `float`?

src/inmanta/ast/type.py
```diff
@@ -265,6 +262,34 @@ def type_string_internal(self) -> str:
...
+            raise RuntimeException(None, "Invalid value '%s', expected Float" % value)
```
sanderr, line 285:
```suggestion
            raise RuntimeException(None, "Invalid value '%s', expected float" % value)
```
Or even better: use `self.type_string()`

src/inmanta/ast/entity.py
sanderr: And use `Float.cast` to handle `Unknown`

changelogs/unreleased/6526-index-float-int.yml
sanderr, line 1: You should delete the old change entry.

src/inmanta/ast/type.py
```diff
@@ -229,10 +229,7 @@ def __eq__(self, other: object) -> bool:
 @stable_api
 class Number(Primitive):
```
sanderr, line 230: You're missing the bugfix for `number()` casts and associated change entry. I understand the "Do not apply any fixes for it" part may have been misleading here. We can decide to keep the bug, but then we must also revert the fix in iso6 to keep behavior consistent. I think I prefer to just apply the fix. See [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1700836622599319).

src/inmanta/ast/entity.py
sanderr: Actually, we don't need the index construction patch, do we? I don't really understand why you changed the index approach. The goal was to "keep `float` as in the PR".

src/inmanta/execute/runtime.py
```diff
@@ -267,6 +267,8 @@ def set_value(self, value: T, location: Location, recur: bool = True) -> None:
                 return
         if not isinstance(value, Unknown) and self.type is not None:
             self.type.validate(value)
+        if isinstance(self.type, (Float)):
```
sanderr, line 270:
```suggestion
        if isinstance(self.type, Float):
```

tests/compiler/test_typing.py
```diff
@@ -380,3 +380,241 @@ def test_2243_override_optional(snippetcompiler) -> None:
...
+        match=re.escape("Type 'number' is deprecated, use 'float' or 'int' instead"),
+    ):
+        (_, scopes) = compiler.do_compile()
+
+
+def test_number_type(snippetcompiler):
+    snippetcompiler.setup_for_snippet(
+        """
+entity Test:
+    number i = 0
```
sanderr, line 404: No need to have explicit tests for number I think. Was this meant to be a test for `float`?

tests/compiler/test_typing.py
```diff
@@ -380,3 +380,241 @@ def test_2243_override_optional(snippetcompiler) -> None:
...
+    compiler.do_compile()
+
+
+def test_same_value_float_int(snippetcompiler):
+    snippetcompiler.setup_for_snippet(
+        """
+    i = 42.0
+    j = 42
+    i = j
+    j = i
```
sanderr, line 423: Can you also add a similar test verifying the `==` operator? i.e. `(42 == 42.0) == true`.

sanderr (changes requested):

src/inmanta/ast/__init__.py
FloLey: No you get it for each use of the number type. But isn't that better? I would like to see all the usages of the number type with where they are used at once so that I can make all the changes without needing to run the code multiple times.

sanderr: I think the solutions team will disagree very strongly. They will not be able to update immediately after this change, and they may well have hundreds of usages of this type. This would pollute their logs incredibly.

FloLey: But so then no need to have an indication of the line where it is used? Just one warning saying that the type is still present?

sanderr: Ah, I think `warnings.warn` shows each warning only once, but the message differs here for each of them because of the location information. Perhaps we can ask the solutions team how they would prefer to be warned about this?

changelogs/unreleased/6526-index-float-int.yml
FloLey: but the PR was reverted? it was deleted at the same time wasn't it?

src/inmanta/ast/__init__.py
sanderr: > But so then no need to have an indication of the line where it is used? Just one warning saying that the type is still present?

I'm honestly not sure. I agree that pointing to the line is best, but I don't immediately see how to achieve both goals.

changelogs/unreleased/6526-index-float-int.yml
sanderr: You're absolutely right, my bad.

src/inmanta/ast/type.py
FloLey: Ok i'll apply the fix. I prefer it with the fix to

src/inmanta/ast/type.py
FloLey: I must say I have no idea, I didn't think of it when changing this. I'll revert to numbers.Integral, it makes me more comfortable

tests/compiler/test_typing.py
FloLey: no there is already one for float, i'll remove this one.

src/inmanta/ast/__init__.py
FloLey: It's seems even when removing the location a lot of warnings are produced. I'll need to look how to fix this

src/inmanta/ast/type.py
FloLey: I don't think so as `isinstance(1, float) == false` and then this doesn't work anymore:

```
entity A:
    float x
end
index A(x)
a = A(x=1)
```

src/inmanta/ast/entity.py
FloLey: @sanderr I'm not sure I understand this comment. Do you mean that noting needs to be changed about the indexes? I thought this was the original goal?

FloLey: because If I can't change the number type and I don't change the way indexes are added, I don't see how I can make this test pass:
```

def test_int_as_index_for_number(snippetcompiler):
        f"""
entity A:
    number x
end
implement A using std::none
index A(x)
test = (A(x=0) == A(x=0.0))
test = true
        """,
    )
    compiler.do_compile()

```
This will always add 2 different indexes

FloLey: I removed this test. It passes with floats but not with the number type

sanderr: > It passes with floats but not with the number type

Perfect. We wanted to "keep float as in the PR" and "keep number as in the current stable but deprecate it. Do not apply any fixes for it, like the index matching."

tests/compiler/test_typing.py
```diff
@@ -573,12 +558,11 @@ def test_float_type(snippetcompiler):
     assert Number().validate(u)
 
 
-@pytest.mark.parametrize_any("type", ["number", "float"])
-def test_int_as_index_for_number(snippetcompiler, type):
+def test_int_as_index_for_number(snippetcompiler):
```
sanderr, line 561:
```suggestion
def test_int_as_index_for_float(snippetcompiler):
```

tests/compiler/test_typing.py
```diff
@@ -619,3 +603,25 @@ def test_lookup_on_int_with_float(snippetcompiler):
         """,
     )
     compiler.do_compile()
+
+
+def test_print_float(snippetcompiler, capsys):
+    snippetcompiler.setup_for_snippet(
+        """
+std::print(float(1.234))
```
sanderr, line 611: Perhaps also add a `1.0`?

sanderr (changes requested):

sanderr:
I think we also need to consider plugins, both with `float` parameters and return types.

Hugo-Inmanta (changes requested):

src/inmanta/ast/type.py
```diff
@@ -590,7 +632,7 @@ class Literal(Union):
     """
 
     def __init__(self) -> None:
-        Union.__init__(self, [NullableType(Number()), Bool(), String(), TypedList(self), TypedDict(self)])
+        Union.__init__(self, [NullableType(Float()), Number(), Bool(), String(), TypedList(self), TypedDict(self)])
```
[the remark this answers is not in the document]
FloLey: This is the Union of all the types, By making one of them Nullable, we basically do "Float or None Or Bool or String..."
if we made them all Nullable it would be "Float or None Or Bool or None or String or None ..." So it comes down to the same. I moved it to Float as number will be gone one day.

Hugo-Inmanta:
I've been playing around a bit with the `number` type and complex numbers:
Plugin file:

```py
from inmanta.plugins import plugin

@plugin
def cplx(re:"int", im: "int") -> "number":
    return re + im*1j
```
And model:
```
import mv2

entity A:
    number x
end
implement A using std::none

index A(x)
a = A(x=mv2::cplx(0,0))
b = A(x=mv2::cplx(2,3))

std::print(A[x=mv2::cplx(0,0)])
std::print(A[x=mv2::cplx(2,3)])
std::print(b.x)
```

Outputs:
```
@__config__::A 7f435bf269e0
@__config__::A 7f435bf26a70
@(2+3j)

```
I don't quite get why it seems to work even though direct cast to number fails
e.g.
```
c = number(mv2::cplx(2,3))

```
Fails:
```
Error: Failed to cast '(2+3j)' to number (reported in number(mv2::cplx(2,3)) (./main.cf:17))
```

tests/compiler/test_typing.py
```diff
@@ -625,3 +572,59 @@ def test_print_number(snippetcompiler, capsys):
...
+    x = root.lookup("test").get_value()
+    assert Integer().validate(x)
+
+
+def test_float_type_argument_plugin_error(snippetcompiler, caplog):
+    snippetcompiler.setup_for_error(
+        """
+import test_674
+
+test = test_674::test_float_to_int("test")
```
sanderr, line 596: Can you parametrize this to try passing an int as well?

tests/data/modules/test_674/plugins/__init__.py
```diff
@@ -24,3 +24,18 @@ def test_not_nullable_list(param: "int[]") -> "bool":
...
+
+
+@plugin
+def test_int_to_float(val1: "int") -> "float":
+    return float(val1)
+
+
+@plugin
+def test_error_float() -> "float":
+    return "hello"
```
sanderr, line 41: Here too I would rather opt for the `int` than `str`. No need to try both I think, just the least trivial one should suffice.

src/inmanta/execute/runtime.py
```diff
@@ -267,6 +267,8 @@ def set_value(self, value: T, location: Location, recur: bool = True) -> None:
                 return
         if not isinstance(value, Unknown) and self.type is not None:
             self.type.validate(value)
+        if isinstance(self.type, Float):
+            value = self.type.cast(value)
```
sanderr, line 271: I think we should revert this change now.

sanderr (changes requested):
Looks good. Final comments.

@Hugo-Inmanta I think you should find your experiments with complex numbers to be more intuitive/consistent now.

tests/compiler/test_typing.py
```diff
@@ -643,18 +588,23 @@ def test_float_type_argument_plugin(snippetcompiler, caplog):
...
-        """
+@pytest.mark.parametrize("val", [42, "test"])
+def test_float_type_argument_plugin_error(snippetcompiler, val):
+    snippet = f"""
 import test_674
 
-test = test_674::test_float_to_int("test")
-        """,
-        ("Invalid value 'test', expected float (reported in " "test_674::test_float_to_int('test') ({dir}/main.cf:4))"),
+test = test_674::test_float_to_int({val if val == 42 else '"'+val+'"'})
```
Hugo-Inmanta, line 596: I think it reads nicer by calling `repr` which adds the single quotes for strings: e.g.

```py
    snippet = f"""
import test_674

test = test_674::test_float_to_int({repr(val)})
        """
    msg = f"Invalid value '{val}', expected float (reported in test_674::test_float_to_int({repr(val)}) ({{dir}}/main.cf:4))"
```

Hugo-Inmanta (changes requested):

tests/compiler/test_typing.py
FloLey: Ho I didn't know that, I'll try it

FloLey:
the compiler version are already bumped for iso6 and master

Hugo-Inmanta (changes requested):
One last thing, I think we should mention in the doc somewhere that our float requires a point to be valid

Hugo-Inmanta (approved):

sanderr (approved):

PR #6773, opened 2023-11-24
Fix a bug where numbers where cast to int instead of float (PR #6730)

FloLey:
# Description

while working on https://github.com/inmanta/inmanta-core/pull/6686 we noticed a bug [here](https://github.com/inmanta/inmanta-core/pull/6686#discussion_r1386642720) . This PR fixes this bug

# Self Check:

- [x] Changelog entry


FloLey:
@sanderr I created a separated pr as my other one will now also be merged in iso6 and I was afraid of conflicts

sanderr (approved):
Looks ok now but it's very much tied to the other PR so it'll depend on how you handle it there.

PR #6795, opened 2023-11-30
Issue/6726 compact database migration scripts

### LINKED ISSUE #6726: compact database migration scripts for iso<=5
Migration scripts can take a long time to execute. We can improve this by getting rid of the older ones, for versions the noone is running anymore. Concretely, we would apply all scripts earlier than the first iso6 release, then dump the schema in v1.py.

Only for iso7

- [ ] apply for core
- [ ] apply for inmanta-lsm
- [ ] add change entry for both, warning that to upgrade to this version you have to be running at least version <latest_component_version_in_iso5>
- [ ] document procedure in internal docs, including which branches to merge it in and which scripts to include
- [ ] after merging, notify solutions and frontend teams

FloLey:
# Description

Migration scripts can take a long time to execute. We can improve this by getting rid of the older ones, for versions that no one is running anymore. Concretely, we would apply all scripts earlier than the first iso6 release, then dump the schema in v1.py.

- [x]  apply for core
- [x]   add change entry for both, warning that to upgrade to this version you have to be running at least version <latest_component_version_in_iso5>
- [x]   document procedure in internal docs, including which branches to merge it in and which scripts to include (https://code.inmanta.com/inmanta/developerdocs/-/merge_requests/140)
- [ ]   after merging, notify solutions and frontend teams

closes #6726 

# Self Check:

- [x] Changelog entry


src/inmanta/db/versions/v1.py
```diff
@@ -23,319 +23,851 @@
...
--- => Prevent sequential scan through all resource actions
-CREATE INDEX resourceaction_started_index ON resourceaction (started);
-
--- Table: public.resourceversionid
--- TODO: FK CONSTRAINT???
-CREATE TABLE IF NOT EXISTS public.resourceversionid (
+
+SET default_tablespace = '';
+
+SET default_table_access_method = heap;
```
arnaudsjs, line 117: These have not been part of any database migration script up to now. Why is this required?

tests/db/compact_db_migrations.py
```diff
@@ -0,0 +1,99 @@
...
+import importlib
+import os
+import re
+
+import asyncpg
+import pytest
+
+UP_TO_VERSION = "202206290"  # Specify the version up to which migrations should be applied
+
+MIGRATIONS_DIR = "../../src/inmanta/db/versions"
```
arnaudsjs, line 31: Like this the script will fail/succeed depending on the current working directory you use to execute the script.

```suggestion
MIGRATIONS_DIR = f"{os.path.dirname(os.path.dirname(__file__))}/src/inmanta/db/versions"
```

tests/db/compact_db_migrations.py
```diff
@@ -0,0 +1,99 @@
...
+"""
+import asyncio
+import importlib
+import os
+import re
+
+import asyncpg
+import pytest
+
+UP_TO_VERSION = "202206290"  # Specify the version up to which migrations should be applied
```
arnaudsjs, line 29: How did you get to this version? The latest ISO5 version is `5.4.2` and it has `v202209160.py` as it's latest database schema version for inmanta-core. Am I missing something?

tests/db/compact_db_migrations.py
```diff
@@ -0,0 +1,99 @@
...
+
+MIGRATIONS_DIR = "../../src/inmanta/db/versions"
+# Regular expression to match migration filenames
+MIGRATION_FILE_PATTERN = re.compile(r"^v(\d+)\.py$")
+
+
+# Extract version number from the filename
+def extract_version(filename):
+    match = MIGRATION_FILE_PATTERN.match(filename)
+    return int(match.group(1)) if match else None
```
arnaudsjs, line 39: Shouldn't we raise an exception when `match` is `None`.

tests/db/compact_db_migrations.py
```diff
@@ -0,0 +1,99 @@
...
+
+UP_TO_VERSION = "202206290"  # Specify the version up to which migrations should be applied
+
+MIGRATIONS_DIR = "../../src/inmanta/db/versions"
+# Regular expression to match migration filenames
+MIGRATION_FILE_PATTERN = re.compile(r"^v(\d+)\.py$")
+
+
+# Extract version number from the filename
+def extract_version(filename):
```
arnaudsjs, line 37:
```suggestion
def extract_version(filename: str) -> int:
```

tests/db/compact_db_migrations.py
```diff
@@ -0,0 +1,99 @@
...
+# Apply all migrations up to a specified version
+async def apply_migrations_up_to_version(connection, up_to_version):
+    migration_files = [f for f in os.listdir(MIGRATIONS_DIR) if MIGRATION_FILE_PATTERN.match(f)]
+    sorted_files = sorted(migration_files, key=extract_version)
+
+    for filename in sorted_files:
+        if extract_version(filename) < int(up_to_version):
+            print(f"Applying migration: {filename}")
+            migration_file_path = os.path.join(MIGRATIONS_DIR, filename)
+            await apply_migration(connection, migration_file_path)
```
arnaudsjs, line 60: We should rely on our existing database migration tool `DBSchema` to apply the database migration scripts. Otherwise we risk that the existing database migration tool and this tool behave slightly differently, causing nasty bugs.

tests/db/compact_db_migrations.py
FloLey: I went into the changelogs here: https://docs.inmanta.com/inmanta-service-orchestrator/latest/CHANGELOG.html
and I saw the latest release before ISO6 was 5.3.0. I think I went wrong there.

Hugo-Inmanta (approved):

tests/db/compact_db_migrations.py
```diff
@@ -0,0 +1,68 @@
...
+
+PACKAGE_NAME = "versions_to_compact"
+
+
+@pytest.mark.asyncio
+async def test_compact_and_dump(postgres_db, database_name):
+    """
+    Compact, apply database migrations using DBSchema, and dump the database schema with modifications.
+    """
+    outfile = "./compacted_dump.sql"
```
arnaudsjs, line 36:
```suggestion
    outfile = f"{os.path.dirname(__file__)}/compacted_dump.sql"
```

tests/db/compact_db_migrations.py
```diff
@@ -0,0 +1,67 @@
...
+
+import asyncpg
+import pytest
+
+from inmanta.data.schema import CORE_SCHEMA_NAME, DBSchema
+
+PACKAGE_NAME = "versions_to_compact"
+
+
+async def test_compact_and_dump(postgres_db, database_name):
```
arnaudsjs, line 32: This approach has the downside that you have to move the migration scripts you want to compact to a separate directory. As such, there is the risk of accidentally forgetting to one version. I would add a check here to verify that no such mistake was made, namely that the version present in the versions_to_compact package start at V1 up to a certain version vX. And that all versions left in the original versions directory are all strictly higher than vX.

arnaudsjs (changes requested):

tests/db/migration_tests/test_v000000001_to_v202211230.py
arnaudsjs, line 1: I don't think it makes sense to keep this test case. The idea behind this test case was to test the migration from version X to version Y, but we no longer have that because we did the compaction. The v1 script will only be executed on an empty database, so it will never migrate data.

arnaudsjs (commented):
It think it would be better to collocate all the files related to the db version compaction under the same directory. Right now we have the directory structure:

```
tests/db/compact_db_migrations.py
tests/db/versions_to_compact
tests/db/versions_to_compact/__init__.py
```
I would change this to:
```
tests/db/compact_migration_scripts
tests/db/compact_migration_scripts/compact_db_migrations.py
tests/db/compact_migration_scripts/versions_to_compact
tests/db/compact_migration_scripts/versions_to_compact/__init__.py
```

tests/db/common.py
```diff
@@ -18,7 +18,103 @@
...
+
+
+async def compact_and_dump(
+    compact_dir: str, original_dir: str, schema_name: str, database_name: str, postgres_db: asyncpg.Connection
+) -> None:
+    """
+    Compact, apply database migrations using DBSchema, and dump the database schema with modifications.
+    """
+
+    def extract_version(filename):
```
arnaudsjs, line 45:
```suggestion
    def extract_version(filename: str) -> Optional[int]:
```

tests/db/common.py
```diff
@@ -18,7 +18,103 @@
...
 
 # The content of this file was moved to inmanta.db.util to allow it to be used from other extensions.
 # This import statement is present to ensure backwards compatibility.
 from inmanta.db.util import MODE_READ_COMMAND, MODE_READ_INPUT, AsyncSingleton, PGRestore  # noqa: F401
+
+PACKAGE_NAME = "versions_to_compact"
+
+
+async def compact_and_dump(
+    compact_dir: str, original_dir: str, schema_name: str, database_name: str, postgres_db: asyncpg.Connection
```
arnaudsjs, line 39: The `postgres_db` argument cannot be of type `asyncpg.Connection`. This method creates a connection using this object.

tests/db/common.py
```diff
@@ -18,7 +18,103 @@
...
+from inmanta.data.schema import DBSchema
 
 # The content of this file was moved to inmanta.db.util to allow it to be used from other extensions.
 # This import statement is present to ensure backwards compatibility.
 from inmanta.db.util import MODE_READ_COMMAND, MODE_READ_INPUT, AsyncSingleton, PGRestore  # noqa: F401
+
+PACKAGE_NAME = "versions_to_compact"
+
+
+async def compact_and_dump(
```
arnaudsjs, line 38: Let's also add documentation regarding the arguments of this method. Some of them are not intuitively clear.

tests/db/common.py
```diff
@@ -18,7 +18,103 @@
...
+        """
+        Get a sorted list of version numbers from all migration scripts in a directory.
+        """
+        versions = []
+        for filename in os.listdir(directory):
+            if filename.endswith(".py"):
+                version = extract_version(filename)
+                if version is not None:  # only select the .py files with a version number and ignore __init__.py
+                    versions.append(version)
+        return sorted(versions)
```
arnaudsjs, line 62:
```suggestion
        match = re.match(r"^v(\d+)\.py", filename)
        return int(match[1]) if match else None

    def get_sorted_versions(directory: str) -> List[int]:
        """
        Get a sorted list of version numbers from all migration scripts in a directory.
        """
        versions = []
        for filename in os.listdir(directory):
            version = extract_version(filename)
            if version is not None:  # only select the .py files with a version number and ignore __init__.py
                versions.append(version)
        return sorted(versions)
```

tests/db/common.py
```diff
@@ -18,7 +18,103 @@
...
+        str(postgres_db.port),
+        "-U",
+        postgres_db.user,
+        "-f",
+        outfile,
+        "-O",
+        "-s",  # This option tells pg_dump to dump only the schema, no data
+        database_name,
+    )
+    await proc.wait()
```
arnaudsjs, line 120: We should also check the return code of the process, because the `wait()` method doesn't raise an exception if the process fails.

arnaudsjs (changes requested):

tests/db/compacted_dump.sql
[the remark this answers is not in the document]
FloLey: indeed I'll remove it

arnaudsjs (approved):

Issue #6056, opened 2023-05-26
document Unknown semantics


sanderr:
The current docs only have three very brief references to the `Unknown` concept. The glossary more or less explains what it is and why we need it but it is not really explained where it may be encountered and what it means for the model or resources.

- [ ] model part: document where `Unknown` comes from in the model, where it might occur in the model and how it affects the data flow (i.e. how is it propagated). Include documentation about `Unknown` in a list: it represents zero or more unknown elements in the list, so if a list contains an unknown its size is also not known.
- [ ] resources/deploy part: the current docs already briefly mention what happens with `Unknown` during the export (translation from model part to deploy part) but what happens with it after that isn't really elaborated on.

Loosely related to inmanta/std#433 and #6033

PR #6799, opened 2023-12-01
Resultcollector refactor fixed

### LINKED ISSUE #6033: most operators don't propagate unknowns
e.g. `And` and `Or` don't special-case `Unknown` at all. They just check for its truthiness, thereby considering it to be true. The same goes for many of the other operators.

sanderr:
# Description

- Restored #6597, with fixes for the bug it introduced
- Addressed many cases of Unknown handling and added tests

closes #6033

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: #6056)


sanderr:
@wouterdb I think it's done now. I will merge it today but I would appreciate an after-the fact review. Could you pay special attention to the change entries? They represent all changes of behavior that I'm aware of. I think that except for the for loop (which we already discussed) the others can all be classified as bugfixes and can therefore go into iso6.

If you want the full diff with master, see #6806.

PR #6802, opened 2023-12-04
Issue/6372 drop typing extensions

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/inmanta-core/issues/6372

# Self Check:

- [x] Changelog entry

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,6 @@ requires = [
     "pyyaml~=6.0",
     "texttable~=1.0",
     "tornado~=6.0",
-    "typing-extensions~=4.8.0",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -32,7 +32,6 @@
     "pyyaml~=6.0",
     "texttable~=1.0",
     "tornado~=6.0",
-    "typing-extensions~=4.8.0",
```
sanderr, line 35: I think we should also drop it from `requirements.txt`.

sanderr (changes requested):

sanderr (approved):

PR #6807, opened 2023-12-04
Add documentation on how to perform upgrade in-place

arnaudsjs:
# Description

Add documentation on how to perform an upgrade in-place.

closes inmanta/inmanta-service-orchestrator#393

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation~~

### DIFF
--- a/docs/administrators/upgrading_the_orchestrator.rst
+++ b/docs/administrators/upgrading_the_orchestrator.rst
@@ -4,6 +4,53 @@
 Upgrading the orchestrator
 --------------------------
 
+Upgrading the orchestrator can be done either in-place or by setting up a new orchestrator next to the old one
+and migrating the state from the old to the new instance. The sections below describe the upgrade procedure
+for each of both situations.
+
+Upgrading the orchestrator in-place
+###################################
+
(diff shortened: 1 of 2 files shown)


Hugo-Inmanta (approved):

docs/administrators/upgrading_the_orchestrator.rst
```diff
@@ -4,6 +4,53 @@
...
+
+Upgrading the orchestrator in-place
+###################################
+
+This section describes how to upgrade an orchestrator in-place.
+
+.. note::
+    **Pre-requisite**
+
+    - Before upgrading the orchestrator to a new major version, make sure the old orchestrator is at the latest version available within its major.
```
edvgui, line 19: Is the guide about upgrading the orchestrator to a new version or a new major version? In the case of the latter, where is the other update procedure documented?

sanderr (changes requested):
> One thing I think we should add is a note to read the new version's changelog for any version-specific upgrade notes.

I think this is still important. Other than that I'm fine with it as is, the stopping of the server is less important imo.

sanderr (approved):

PR #6814, opened 2023-12-05
Change entry for net module deprecation

Hugo-Inmanta:
# Description

Change entry for `net` module deprecation

Part of https://github.com/inmanta/net/issues/209

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/209-deprecate-the-net-module.yml
+++ b/changelogs/unreleased/209-deprecate-the-net-module.yml
@@ -1,0 +1,7 @@
+description: Deprecate the net module.
+issue-nr: 209
+issue-repo: net
+change-type: minor
+destination-branches: [master]
+sections:
+  deprecation-note: "The ``net`` module has been deprecated."


changelogs/unreleased/209-deprecate-the-net-module.yml
```diff
@@ -0,0 +1,7 @@
+description: Deprecate the net module.
+issue-nr: 209
+issue-repo: net
+change-type: minor
+destination-branches: [master]
```
arnaudsjs, line 5: The deprecation log message will also show up on iso6, so we need to make sure it's also mentioned in the changelog.
```suggestion
destination-branches: [master, iso6]
```

changelogs/unreleased/209-deprecate-the-net-module.yml
```diff
@@ -0,0 +1,7 @@
+description: Deprecate the net module.
+issue-nr: 209
+issue-repo: net
+change-type: minor
```
arnaudsjs, line 4:
```suggestion
change-type: patch
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #6815, opened 2023-12-05
Issue/fix 6691

edvgui:
# Description

Fix bug introduced in https://github.com/inmanta/inmanta-core/pull/6808

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/6691-catch-all-arguments-fix.yml
+++ b/changelogs/unreleased/6691-catch-all-arguments-fix.yml
@@ -2,4 +2,5 @@ change-type: patch
 description: Fix bugs introduced in https://github.com/inmanta/inmanta-core/pull/6808
 destination-branches: [master]
 sections:
-  minor-improvement: "Stricter type checking of values returned by plugins.  Specifying None as returned type requires None to be returned."
+  minor-improvement: "Stricter type checking of values returned by plugins."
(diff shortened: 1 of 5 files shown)


changelogs/unreleased/6691-catch-all-arguments-fix.yml
```diff
@@ -0,0 +1,5 @@
+change-type: patch
+description: Fix bugs introduced in https://github.com/inmanta/inmanta-core/pull/6808
+destination-branches: [master]
+sections:
+  minor-improvement: "Stricter type checking of values returned by plugins.  Specifying None as returned type requires None to be returned."
```
edvgui, line 5: Once this is merged in master, I will create another PR for iso6, and:
1. Merge the fix in this PR
2. Allow any returned value for plugins that use None as returned value annotation

sanderr, line 5: I would make sure we also have an upgrade-note. Perhaps like this?
```suggestion
  minor-improvement: "Stricter type checking of values returned by plugins."
  upgrade-note: "Stricter type checking of values returned by plugins. Specifying None as returned type requires None to be returned."
```

FloLey (approved):

sanderr (changes requested):

sanderr (approved):

sanderr:
I'll retrigger the container builds once this is merged

PR #6817, opened 2023-12-05
Issue/iso6/6691 fix

edvgui:
# Description

Follow up to https://github.com/inmanta/inmanta-core/pull/6815#discussion_r1415276205

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/plugins.py
+++ b/src/inmanta/plugins.py
@@ -216,7 +216,7 @@ PLUGIN_TYPES = {
     "any": inmanta_type.Type(),  # Any value will pass validation
     "expression": inmanta_type.Type(),  # Any value will pass validation
     "null": Null(),  # Only NoneValue will pass validation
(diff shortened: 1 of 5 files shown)


src/inmanta/plugins.py
```diff
@@ -216,7 +216,7 @@ def type_string_internal(self) -> str:
     "any": inmanta_type.Type(),  # Any value will pass validation
     "expression": inmanta_type.Type(),  # Any value will pass validation
     "null": Null(),  # Only NoneValue will pass validation
-    None: Null(),  # Only NoneValue will pass validation
+    None: inmanta_type.Type(),  # In iso6, a None return type accepts any value, from iso7+ it expects a None value.
```
edvgui, line 219: This is the change to make iso6 backward compatible with regard to returned types validation

PR #6824, opened 2023-12-05
Issue/race in increment

wouterdb:
# Description

Fix for race condition, 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/resourceservice.py
+++ b/src/inmanta/server/services/resourceservice.py
@@ -328,6 +328,13 @@ class ResourceService(protocol.ServerSlice):
 
         now = datetime.datetime.now().astimezone()
 
+        def on_agent(res: ResourceIdStr) -> bool:
+            idr = Id.parse_id(res)
+            return idr.get_agent_name() == agent
+
+        # set already done to deployed
+        await self.mark_deployed(env, neg_increment, now, version, filter=on_agent)
+
         resources = await data.Resource.get_resources_for_version(env.id, version, agent)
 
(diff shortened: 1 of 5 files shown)


src/inmanta/server/services/resourceservice.py
```diff
@@ -328,6 +328,25 @@ async def get_resource_increment_for_agent(self, env: data.Environment, agent: s
 
         now = datetime.datetime.now().astimezone()
 
+        def on_agent(res: ResourceIdStr) -> bool:
+            idr = Id.parse_id(res)
+            return idr.get_agent_name() == agent
+
+        # This is a bit subtle
+        # Any resource we consider deployed has to be marked as such
+        # Otherwise the agent will fail the deployment
```
wouterdb, line 337: We could fix this by considering available as deployed, but I do like this second layer of defense in the agent

src/inmanta/server/services/resourceservice.py
```diff
@@ -328,6 +328,25 @@ async def get_resource_increment_for_agent(self, env: data.Environment, agent: s
...
+        # Stale successful deployments can cause resource that were available to be now considered deployed
+        # We don't do this back propagation on deploy,
```
wouterdb, line 339: there were other reasons not to, but I forgot again @sanderr may recall

arnaudsjs (changes requested):

arnaudsjs (approved):

sanderr (approved):
I have some questions I'd like to ask in person, but nothing blocking this.

PR #6833, opened 2023-12-06
Issue/deadlock on clear

wouterdb:
# Description

Race condition detected on systemtenant

Do we merge this prior to release? 

# Self Check:

- [x] Changelog entry


sanderr (approved):
I would prefer to merge after the release. What's your opinion?

wouterdb:
that might be the safest way

PR #6837, opened 2023-12-07
Add a filter to the resource action endpoint to only select resource actions that have a change

### LINKED ISSUE #6733: Add a filter to the resource action endpoint to only select resource actions that have a change
Allow filtering on change != nochange to be able to get a log of all things the orchestrator did

FloLey:
# Description

Allow filtering on change != nochange to be able to get a log of all things the orchestrator did

closes #6733 

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (changes requested):

changelogs/unreleased/6733-add-filter-resource-action-nochange.yml
```diff
@@ -0,0 +1,7 @@
+description: Add a filter to the resource action endpoint to only select resource actions that have a change
+issue-nr: 6733
+change-type: minor
+destination-branches: [master, iso6]
```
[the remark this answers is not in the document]
FloLey: probably, i'll change it

tests/test_server.py
```diff
@@ -864,9 +864,33 @@ async def test_get_resource_actions(postgresql_client, client, clienthelper, ser
...
+    resource_action = data.ResourceAction(
+        environment=uuid.UUID(environment),
+        version=1,
+        resource_version_ids=[rvid_r1_v1],
+        action_id=later_action_id,
+        action=const.ResourceAction.deploy,
+        started=start_time,
+        change=const.Change.created,
+    )
+    await resource_action.insert()
```
arnaudsjs, line 888: Why directly query the database here instead of using the API. The API is a lot more stable over time than the database schema. A small change in the schema of the ResourceAction table might break this test case.

tests/test_server.py
```diff
@@ -864,9 +864,33 @@ async def test_get_resource_actions(postgresql_client, client, clienthelper, ser
...
 
+    #  adding a resource action with its change field set to "created" to test the get_resource_actions
+    #  filtering on resources with changes
+    rvid_r1_v1 = "std::File[agent1,path=/etc/file1],v=1"
+    await data.Resource.new(
+        environment=uuid.UUID(environment),
+        status=const.ResourceState.available,
+        resource_version_id=rvid_r1_v1,
+        attributes={"purge_on_delete": False, "purged": True, "requires": []},
+    ).insert()
```
arnaudsjs, line 875: Why not add this resource using the `put_version()` endpoint like all the other resources? Doing this breaks the fundamental invariant of the code base that no resources can be added to an existing configuration model. This might cause all sorts of instabilities on the test.

arnaudsjs (changes requested):

arnaudsjs (changes requested):

src/inmanta/protocol/methods_v2.py
```diff
@@ -521,6 +521,7 @@ def get_resource_actions(
     action_id: Optional[uuid.UUID] = None,
     first_timestamp: Optional[datetime.datetime] = None,
     last_timestamp: Optional[datetime.datetime] = None,
+    exclude_changes: Optional[list[str]] = None,
```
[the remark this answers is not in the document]
FloLey: I had another failure in the `process` function of `rest/__init__.py:`

`isinstance(self._message[arg], typing_inspect.get_args(arg_type)[0])` is false whenever an enum is used as it becomes:

`isinstance('nochange', Change)` for example.

I thought it was ok to do it like this

changelogs/unreleased/6733-add-filter-resource-action-nochange.yml
```diff
@@ -0,0 +1,7 @@
+description: Add a filter to the resource action endpoint to only select resource actions that have a change
+issue-nr: 6733
+change-type: minor
+destination-branches: [master, iso7, iso6]
```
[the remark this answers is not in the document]
FloLey: If I recall correctly it was nice to have it for iso6 but not to bad if it wasn't there. I removed iso6 for now. I'll add it back and change the change-type if I misremembered

src/inmanta/protocol/methods_v2.py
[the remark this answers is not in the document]
arnaudsjs: There are two main reasons why we shouldn't use type `str`:

* First of all we need to add the code that converts the `str` type into the `Change` type to the `get_resource_actions()` method and over time this code will be copied to more and more API endpoints. So having this type conversion/validation once in the protocol layer is much better. This also leads to error messages that are more consistent across API endpoints.
* Secondly, the signature of these methods is used to generate the swagger documentation. If we use type `str` it will be unclear for the user which values are allowed. We could solve that by adding the possible values to the docstring, but experience learns us that the docstring easily get out-of-sync with the code. If we use the enum type in the signature, the possible values are kept up-to-date automatically based on the implementation of the enum.

src/inmanta/protocol/rest/__init__.py
```diff
@@ -220,6 +221,9 @@ async def process(self) -> None:
             arg_type: Optional[type[object]] = self._argspec.annotations.get(arg)
             if arg in self._message:
                 # Argument is parameter in body of path of HTTP request
+                if typing_inspect.is_optional_type(arg_type) and self._properties.operation == "GET":
+                    # Convert Optional[SomeType] to SomeType for get requests
+                    arg_type = more_itertools.one(arg for arg in typing_inspect.get_args(arg_type) if arg is not type(None))
```
FloLey, line 226: @arnaudsjs this fails for the `get_agents` call for those argument:

```
    start: Optional[Union[datetime.datetime, bool, str]] = None,
    end: Optional[Union[datetime.datetime, bool, str]] = None,
```
it results in more than one item

src/inmanta/protocol/methods_v2.py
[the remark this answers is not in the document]
FloLey: I will try to add to fix this but I don't know how yet. I'll look tomorrow.

src/inmanta/protocol/rest/__init__.py
FloLey: I reworked that part, I think it is better now.

src/inmanta/server/services/resourceservice.py
```diff
@@ -1069,7 +1069,18 @@ async def get_resource_actions(
...
         first_timestamp: Optional[datetime.datetime] = None,
         last_timestamp: Optional[datetime.datetime] = None,
+        exclude_changes: Optional[list[Change]] = None,
     ) -> ReturnValue[list[ResourceAction]]:
+        if exclude_changes is None:
+            exclude_changes = []
+        else:
+            try:
+                # Convert strings to Change enum
+                exclude_changes = [Change(change) for change in exclude_changes]
```
Hugo-Inmanta, line 1079: Why do we need this conversion ? isn't `exclude_changes` already a `list[Change]` ?

src/inmanta/protocol/rest/__init__.py
```diff
@@ -220,21 +220,35 @@ async def process(self) -> None:
...
+                        and typing_inspect.is_generic_type(arg_type)
+                        and issubclass(typing_inspect.get_origin(arg_type), list)
+                    )
+                    is_single_element_list = len(typing_inspect.get_args(arg_type, evaluate=True)) == 1 and isinstance(
+                        self._message[arg], typing_inspect.get_args(arg_type)[0]
+                    )
+                    is_not_list = not isinstance(self._message[arg], list)
+
+                    if is_generic_list and is_not_list and is_single_element_list:
+                        # Handle edge case where a single element is not parsed as a list
```
Hugo-Inmanta, line 245: >                 # If a GET endpoint has a parameter of type list that is encoded as a URL query parameter and the
>                 # specific request provides a list with one element, urllib doesn't parse it as a list.
>                 # Map it here explicitly to a list.
I think the previous comment could be added back here, wdyt ?

src/inmanta/protocol/rest/__init__.py
```diff
@@ -220,21 +220,35 @@ async def process(self) -> None:
...
-                    # Map it here explicitly to a list.
-                    value = [self._message[arg]]
+                if self._properties.operation == "GET":
+                    # Check if the argument type is an Optional type (e.g., Optional[List[SomeType]])
+                    if typing_inspect.is_optional_type(arg_type):
+                        # Extract non-None types from the Optional type
+                        non_none_arg_types = [arg for arg in typing_inspect.get_args(arg_type) if arg is not type(None)]
+
+                        # If there's only one type and it's a list, simplify the type to just the list type
+                        if len(non_none_arg_types) == 1 and typing_inspect.get_origin(non_none_arg_types[0]) is list:
```
Hugo-Inmanta, line 230: Why not `issubclass(typing_inspect.get_origin(arg_type), list)` ?

Hugo-Inmanta (changes requested):

src/inmanta/server/services/resourceservice.py
FloLey: the typing in the arguments is wrong, currently those are strings. ([see arnauds comment](https://github.com/inmanta/inmanta-core/pull/6837#discussion_r1426564804))

src/inmanta/protocol/rest/__init__.py
FloLey: This:
`if len(non_none_arg_types) == 1 and issubclass(typing_inspect.get_origin(arg_type), list):`
doesn't work if that is what you mean?

FloLey: I don't know if this comment is still relevant as I reworked it quit a bit :p

tests/test_server.py
```diff
@@ -841,6 +841,25 @@ async def test_get_resource_actions(postgresql_client, client, clienthelper, ser
             }
         )
 
+    #  adding a resource action with its change field set to "created" to test the get_resource_actions
+    #  filtering on resources with changes
+
+    rvid_r1_v1 = f"std::File[agent1,path=/etc/file1],v={version}"
```
arnaudsjs, line 847:
```suggestion
    rvid_r1_v1 = f"std::File[agent1,path=/tmp/file200],v={version}"
```

src/inmanta/protocol/rest/__init__.py
```diff
@@ -219,23 +298,25 @@ async def process(self) -> None:
...
+                        arg_type_not_optional = self.transform_optional_list_to_list(arg_type)
+                        # Update the argument type for optional lists.
+                        arg_type_actual = get_args(arg_type_not_optional)[0]
+                        # Check if the argument type is an enum and get the enum value if applicable.
+                        is_enum = self._is_enum_type(arg_type_actual)
+                        enum_value = self._get_enum_value(arg_type_actual, message_arg) if is_enum else None
+
+                        value = self._handle_generic_list(message_arg, enum_value, is_enum)
+                    else:
+                        value = message_arg
```
arnaudsjs, line 314: Why is all the `enum` related logic required here? The conversion from str to enum value is normally done using pydantic in the `validate_arguments()` method below. We explicitly introduced Pydantic in our code base to not have to write this type conversion logic. Could you elaborate?

arnaudsjs (changes requested):

src/inmanta/protocol/rest/__init__.py
```diff
@@ -189,6 +191,83 @@ async def _run_getters(self, arg: str, value: Optional[object]) -> Optional[obje
...
+    def transform_optional_list_to_list(arg_type: Optional[type[object]]) -> Optional[type[object]]:
+        """
+        Transforms an optional list type into a list type. If the provided type is not an optional list,
+        it returns the original type.
+        """
+        if typing_inspect.is_optional_type(arg_type):
+            # Extract types within Optional
+            inner_types = get_args(arg_type)
+            # Find and return the list type within the inner types
+            for inner_type in inner_types:
```
Hugo-Inmanta, line 220: Isn't it always the first element of the tuple returned by `get_args` ?

src/inmanta/protocol/rest/__init__.py
```diff
@@ -219,23 +298,25 @@ async def process(self) -> None:
...
-                    # If a GET endpoint has a parameter of type list that is encoded as a URL query parameter and the
-                    # specific request provides a list with one element, urllib doesn't parse it as a list.
-                    # Map it here explicitly to a list.
-                    value = [self._message[arg]]
+                message_arg = self._message[arg]
+
+                if self._properties.operation == "GET":
+                    if self.is_list_or_optional_list(arg_type):
+                        arg_type_not_optional = self.transform_optional_list_to_list(arg_type)
+                        # Update the argument type for optional lists.
```
Hugo-Inmanta, line 306: I would move this  comment one line higher, and maybe change it to 'Strip the `Optional[...]` annotation if present'

src/inmanta/protocol/rest/__init__.py
```diff
@@ -219,23 +298,25 @@ async def process(self) -> None:
...
+                        arg_type_actual = get_args(arg_type_not_optional)[0]
```
Hugo-Inmanta, line 307: If I understand correctly if `arg_type` is `list[str]`, then `arg_type_actual` would be `str` at this point ?
I would rename this variable to list_type_parameter, wdyt ?

FloLey: @Hugo-Inmanta sorry Hugo I removed all this part as it isn't needed.

Hugo-Inmanta (approved):

src/inmanta/protocol/rest/__init__.py
```diff
@@ -219,23 +219,39 @@ async def process(self) -> None:
...
-                    # If a GET endpoint has a parameter of type list that is encoded as a URL query parameter and the
-                    # specific request provides a list with one element, urllib doesn't parse it as a list.
-                    # Map it here explicitly to a list.
-                    value = [self._message[arg]]
+                message_arg = self._message[arg]
+
+                if self._properties.operation == "GET":
+                    # Check if the argument type is an Optional type (e.g., Optional[List[SomeType]])
+                    if typing_inspect.is_optional_type(arg_type):
+                        # Extract non-None types from the Optional type
```
arnaudsjs, line 227:
```suggestion
                        # typing_inspect.get_args(Optiona[<type>]) return a tuple (Type, NoneType). Extract the non-None type.
```

src/inmanta/protocol/rest/__init__.py
```diff
@@ -219,23 +219,39 @@ async def process(self) -> None:
         for i, arg in enumerate(args):
             arg_type: Optional[type[object]] = self._argspec.annotations.get(arg)
             if arg in self._message:
-                # Argument is parameter in body of path of HTTP request
```
arnaudsjs, line 222: Why was this comment removed?

src/inmanta/protocol/rest/__init__.py
```diff
@@ -219,23 +219,39 @@ async def process(self) -> None:
...
-                if (
-                    arg_type
```
arnaudsjs, line 224: This condition is not checked anymore.

src/inmanta/protocol/rest/__init__.py
```diff
@@ -219,23 +219,39 @@ async def process(self) -> None:
...
+                    # Check if the list type has exactly one element type
+                    is_single_element_list = len(typing_inspect.get_args(arg_type, evaluate=True)) == 1
+                    # Check if the argument value itself is not a list
+                    is_not_list = not isinstance(self._message[arg], list)
+
+                    if is_generic_list and is_not_list and is_single_element_list:
+                        # If a GET endpoint has a parameter of type list that is encoded as a URL query parameter and the
+                        # specific request provides a list with one element, urllib doesn't parse it as a list.
+                        # Map it here explicitly to a list.
+                        value = [self._message[arg]]
```
arnaudsjs, line 247: The problem with this new implementation is that the goal is rather vague. In the past we had this comment that clarified a lot. But with the new implementation this comment has sunk quite deep into the if-the-else nesting. I think readability would be improved quite a lot if we extract this logic into a separate method. Then we can put this comment in the docstring for example.

src/inmanta/protocol/rest/__init__.py
```diff
@@ -219,23 +219,39 @@ async def process(self) -> None:
...
-                    and issubclass(typing_inspect.get_origin(arg_type), list)
-                    and not isinstance(self._message[arg], list)
-                    and len(typing_inspect.get_args(arg_type, evaluate=True)) == 1
-                    and isinstance(self._message[arg], typing_inspect.get_args(arg_type)[0])
-                ):
-                    # If a GET endpoint has a parameter of type list that is encoded as a URL query parameter and the
-                    # specific request provides a list with one element, urllib doesn't parse it as a list.
-                    # Map it here explicitly to a list.
-                    value = [self._message[arg]]
+                message_arg = self._message[arg]
```
arnaudsjs, line 222: `message_arg` is a rather vague name. It actually represent the value that belongs to the argument, so why not call it value? This would also make many of the else-blocks redundant.
```suggestion
                value = self._message[arg]
```

arnaudsjs (changes requested):

src/inmanta/protocol/rest/__init__.py
```diff
@@ -189,6 +189,38 @@ async def _run_getters(self, arg: str, value: Optional[object]) -> Optional[obje
...
+        """
+        Validates the consistency of arguments, ensuring they are not passed both as a header and a non-header value.
+        """
+        for arg in args:
+            if arg in self._message and self._is_header_param(arg) and self._is_header_param_provided(arg):
+                message_value = self._message[arg]
+                header_value = self._get_header_value_for(arg)
+                if message_value != header_value:
+                    raise exceptions.BadRequest(
+                        f"Argument {arg} has inconsistent values (header={header_value}; non-header={message_value})."
```
arnaudsjs, line 221: Why was this error message changed? The previous message is more clear in my opinion?

src/inmanta/protocol/rest/__init__.py
```diff
@@ -189,6 +189,38 @@ async def _run_getters(self, arg: str, value: Optional[object]) -> Optional[obje
             LOGGER.exception("Failed to use getter for arg %s", arg)
             raise e
 
+    @staticmethod
+    def _handle_get_request_arg(arg_type: Optional[Type[object]], message_arg: object) -> object:
```
arnaudsjs, line 193:
```suggestion
    def _handle_get_request_arg(arg_type: Optional[Type[object]], arg_value: object) -> object:
```

src/inmanta/protocol/rest/__init__.py
```diff
@@ -189,6 +189,38 @@ async def _run_getters(self, arg: str, value: Optional[object]) -> Optional[obje
...
+        Maps single element lists provided as query parameters to actual lists.
+        """
+        if typing_inspect.is_optional_type(arg_type):
+            non_none_arg_types = [arg for arg in typing_inspect.get_args(arg_type) if arg is not type(None)]
+            if len(non_none_arg_types) == 1 and typing_inspect.get_origin(non_none_arg_types[0]) is list:
+                arg_type = non_none_arg_types[0]
+
+        is_generic_list = arg_type and typing_inspect.is_generic_type(arg_type) and typing_inspect.get_origin(arg_type) is list
+        is_single_element_list = len(typing_inspect.get_args(arg_type, evaluate=True)) == 1
+        is_not_list = not isinstance(message_arg, list)
```
arnaudsjs, line 205:
```suggestion
        arg_value_is_not_list = not isinstance(message_arg, list)
```

src/inmanta/protocol/rest/__init__.py
```diff
@@ -189,6 +189,38 @@ async def _run_getters(self, arg: str, value: Optional[object]) -> Optional[obje
...
+        Handles processing of arguments for GET requests, especially for list types encoded as URL query parameters.
+        Maps single element lists provided as query parameters to actual lists.
+        """
+        if typing_inspect.is_optional_type(arg_type):
+            non_none_arg_types = [arg for arg in typing_inspect.get_args(arg_type) if arg is not type(None)]
+            if len(non_none_arg_types) == 1 and typing_inspect.get_origin(non_none_arg_types[0]) is list:
+                arg_type = non_none_arg_types[0]
+
+        is_generic_list = arg_type and typing_inspect.is_generic_type(arg_type) and typing_inspect.get_origin(arg_type) is list
+        is_single_element_list = len(typing_inspect.get_args(arg_type, evaluate=True)) == 1
```
arnaudsjs, line 204:
```suggestion
        is_single_type_list = len(typing_inspect.get_args(arg_type, evaluate=True)) == 1
```

src/inmanta/protocol/rest/__init__.py
arnaudsjs, line 193:
```suggestion
    def _ensure_list_if_list_type(arg_type: Optional[Type[object]], message_arg: object) -> object:
```

src/inmanta/protocol/rest/__init__.py
```diff
@@ -189,6 +189,38 @@ async def _run_getters(self, arg: str, value: Optional[object]) -> Optional[obje
             LOGGER.exception("Failed to use getter for arg %s", arg)
             raise e
 
+    @staticmethod
+    def _handle_get_request_arg(arg_type: Optional[Type[object]], message_arg: object) -> object:
+        """
+        Handles processing of arguments for GET requests, especially for list types encoded as URL query parameters.
+        Maps single element lists provided as query parameters to actual lists.
```
arnaudsjs, line 196:
```suggestion
        Handles processing of arguments for GET requests, especially for list types encoded as URL query parameters.
        If a GET endpoint has a parameter of type list that is encoded as a URL query parameter and the specific request provides a list with one element, urllib doesn't parse it as a list. Map it here explicitly to a list.
```

arnaudsjs (changes requested):

src/inmanta/protocol/rest/__init__.py
```diff
@@ -189,21 +189,30 @@ async def _run_getters(self, arg: str, value: Optional[object]) -> Optional[obje
...
-        Process the message
+        Handles processing of arguments for GET requests, especially for list types encoded as URL query parameters.
+        If a GET endpoint has a parameter of type list that is encoded as a URL query parameter and the specific request
+        provides a list with one element, urllib doesn't parse it as a list. Map it here explicitly to a list.
         """
-        args: list[str] = list(self._argspec.args)
+        if typing_inspect.is_optional_type(arg_type):
+            non_none_arg_types = [arg for arg in typing_inspect.get_args(arg_type) if arg is not type(None)]
+            if len(non_none_arg_types) == 1 and typing_inspect.get_origin(non_none_arg_types[0]) is list:
+                arg_type = non_none_arg_types[0]
```
arnaudsjs, line 202: Do you know why the second condition is required? This if-block only removes the optional part of the type. The code that follow check whether it's a list type or not, right? So I guess this condition is redundant here.
```suggestion
            if len(non_none_arg_types) == 1:
                arg_type = non_none_arg_types[0]
```

arnaudsjs (changes requested):

src/inmanta/protocol/rest/__init__.py
FloLey: indeed you are right

arnaudsjs (approved):

PR #6838, opened 2023-12-07
Fix the performance degradation on get_resources_for_agent()

arnaudsjs:
# Description

Fix the performance degradation on the `get_resources_for_agent()` call.

Follow up of #6824

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/resourceservice.py
+++ b/src/inmanta/server/services/resourceservice.py
@@ -332,20 +332,23 @@ class ResourceService(protocol.ServerSlice):
             idr = Id.parse_id(res)
             return idr.get_agent_name() == agent
 
-        # This is a bit subtle
-        # Any resource we consider deployed has to be marked as such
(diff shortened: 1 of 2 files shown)


sanderr (approved):
Looks good as far as I can tell. I have to confess I'm still not clean on some of the details here and I didn't get the chance yet to discuss with Wouter. But the parts I do understand look good.

src/inmanta/server/services/resourceservice.py
```diff
@@ -936,7 +942,8 @@ async def update_fields_resource(
                     # transient resource update
                     if not is_resource_action_finished:
```
wouterdb, line 943: the list of resource ids for the resource action is not filtered 

I find this hard to review, I would simply filter the list of resources and resource_ids higher up in the code, so the rest of the code is more uniform

wouterdb (approved):

PR #6839, opened 2023-12-07
added unknown docs

### LINKED ISSUE #6056: document Unknown semantics

sanderr:
# Description

Added documentation about unknowns.

closes #6056

# Self Check:

- [ ] Changelog entry


docs/language.rst
```diff
@@ -864,3 +864,76 @@ Plug-ins
...
 accepts parameters and returns a value that it computed out of the variables. Each module that is included can also provide plug-ins. These plug-ins are accessible within the namespace of the
 module. The :ref:`module-plugins` section of the module guide provides more details about how to write a plugin.
+
+
+.. _language_unknowns:
+Unknowns
+========
+
+Wherever the configuration model interacts with the outside world (e.g. to fetch external values) :term:`unknown` values
+may be present. These unknowns are propagated through the model to finally end up in the resources that require these unknown
```
wouterdb, line 874: I would add a sentence to say what they are. 

```suggestion
may be present. These unknowns represent value we don't know yet, such as the IP address of a machine that hasn't been created yet. Because the compiler can handle unknowns, developer can writes models as if all information is present up front, even when this is not the case. These unknowns are propagated through the model to finally end up in the resources that require these unknown
```

docs/language.rst
```diff
@@ -864,3 +864,76 @@ Plug-ins
...
+values. This section describes how unknown values flow through the model, and perhaps equally importantly, where they do not
+flow at all.
+
+.. note::
+    Unknowns are a subtle concept. Luckily, for the majority of model development you don't really need to take them into
+    account. However, for some advanced scenarios it may be important to know how and where they may occur.
+
+For the most part, unknowns are simply propagated along the data flow: statements like assignment statements, constructors
```
wouterdb, line 882: I would first give the simple explanation

```suggestion
For the most part, unknowns are simply propagated along the data flow: any value derived from an unkown in any way becomes an unknown as well. More specifically: statements like assignment statements, constructors
```

docs/language.rst
```diff
@@ -864,3 +864,76 @@ Plug-ins
...
+
+.. code-block:: inmanta
+
+    # std::env returns an unknown if the environment variable is not (yet) set
+    my_unknown = std::get_env("THIS_ENV_VAR_DOES_NOT_EXIST")
+
+    a = my_unknown  # a is unknown
+    b = [1, 2, my_unknown, 3]  # b is a list with 1 unknown element
+    c = my_unknown is defined  # we can not know if c is null, so c is also unknown
+    d = true or my_unknown  # trivial, value of my_unknown is irrelevant -> d is true
```
wouterdb, line 898:
```suggestion
    d = true or my_unknown  # value of my_unknown is irrelevant -> d is true
```

docs/language.rst
```diff
@@ -864,3 +864,76 @@ Plug-ins
...
+    else:
+        # neither is this one
+    end
+
+    for x in my_unknown:
+        # neither is this one
+    end
+
+    for x in [1, 2, my_unknown]:
+        # this block is executed twice: x=1 and x=2
```
wouterdb, line 913: my opinion on this is known. I would like to still have this power, but I leave it to you.

docs/language.rst
```diff
@@ -864,3 +864,76 @@ Plug-ins
...
+    end
+
+    g = my_unknown ? true : false  # condition is unknown -> neither branch is executed, result is unknown
+    h = [1 for x in [1, 2, my_unknown]]  # the expression `1` is executed once with x=1 and once with x=2. Unknown is propagated as is -> h = [1, 1, unknown]
```
wouterdb, line 917: I find these two to be hard to reconcile with the for loop. For me, a list comprehension is 

```
result = [a for item in c if d]
```
with a and d expressions referring to item and c an expression
```
collector = ...
for item in c:
  if d:
    collector.collection += a
result = collector.collection
``` 

but again, I raised this before.

docs/language.rst
```diff
@@ -864,3 +864,76 @@ Plug-ins
...
+    my_unknown = std::get_env("THIS_ENV_VAR_DOES_NOT_EXIST")
+    my_unknown2 = std::get_env("THIS_ENV_VAR_DOES_NOT_EXIST2")
+
+    l = [1, my_unknown, 3, my_unknown2, 5]
+    a = [x for x in l if x > 2]  # l = [unknown, 3, unknown, 5]
+
+    # an unknown can even represent more than one unknown value
+    b = my_unknown == 0 ? [1, 2] : [3, 4]  # b = unknown -> when it becomes known it will be either [1, 2] or [3, 4]
+
+    c = std::len(l)  # c = unknown (l contains unknowns, so its length is also unknown)
```
wouterdb, line 939: I would also remark the risk of having unknowns in conditions

```
implementation things_are_on for A:
   self.b.is_on = self
end

implementation things_are_off for  A:
     self.b.is_off = self
end

implement A using things_are_on if self.on
implement A using things_are_off if self.off


collector = B()
all = [
 A(b=b, on=True)
  A(b=b, on=False)
   A(b=b, on=unknown())
 ]

std::len(b.is_on) # 1  
std::len(b.is_off) # 1
# the fact that the length of b.is_on and b.is_off are unknown is lost
# the intuitive notion that all things are the things that are off + the things tat are on no longer holds 
````

**Comment by wouterdb** [comment]
Nice!

File: docs/language.rst [hunk 7]
```diff
@@ -864,3 +864,76 @@ Plug-ins
...
+
+.. code-block:: inmanta
+
+    my_unknown = std::get_env("THIS_ENV_VAR_DOES_NOT_EXIST")
+    my_unknown2 = std::get_env("THIS_ENV_VAR_DOES_NOT_EXIST2")
+
+    l = [1, my_unknown, 3, my_unknown2, 5]
+    a = [x for x in l if x > 2]  # l = [unknown, 3, unknown, 5]
+
+    # an unknown can even represent more than one unknown value
```
edvgui, line 936: I would also say that they can represent less than one unknown value (empty list in a list)

edvgui (approved):
Very nice!

docs/language.rst
sanderr: Updated your suggestion to fix some typos, then committed it.

docs/language.rst
sanderr: Nice, but I'm going to move it one sentence later, because it really applies to the "any expression that cannot produce ..." part. e.g. `[1, unknown, 3]` does not become `unknown`.

docs/language.rst
sanderr: By now the behavior is no longer under discussion, it's simply about documenting it. I'll double check this now.

docs/language.rst
sanderr: I'll update this to construct a simple entity.

docs/language.rst
sanderr: Made an attempt for a middle ground, to give some more context upfront, but in a way that applies for both statements that follow.

docs/language.rst
sanderr: I realize it's been a very long time. But do you remember what you were getting at here? Two notes:

1. the model had some gaps. I reworked it in the following format that does compile
```
entity A:
    bool on
end

implementation things_are_on for A:
    self.b.is_on = self
end

implementation things_are_off for A:
    self.b.is_off = self
end

implement A using things_are_on when self.on
implement A using things_are_off when not self.on

entity B:
end
implement B using std::none
A.b [1] -- B
B.is_on [0:] -- A
B.is_off [0:] -- A


b = B()
all = [
    A(b=b, on=true),
    A(b=b, on=false),
    A(b=b, on=std::get_env("THIS_ENV_VAR_DOES_NOT_EXIST")),
]


on_len = std::len(b.is_on)
off_len = std::len(b.is_off)
on_len_txt = std::is_unknown(on_len) ? "unknown" : on_len
off_len_txt = std::is_unknown(off_len) ? "unknown" : off_len
std::print(f"b.is_on = {b.is_on}")
std::print(f"b.is_off = {b.is_off}")
std::print(f"std::len(b.is_on) = {on_len_txt}")  # 2
std::print(f"std::len(b.is_off) = {off_len_txt}")  # 2

# THE BELOW APPEARS TO BE INCORRECT
std::len(b.is_on) # 1
std::len(b.is_off) # 1
# the fact that the length of b.is_on and b.is_off are unknown is lost
# the intuitive notion that all things are the things that are off + the things tat are on no longer holds
```

2. This behavior suprises me a lot. It looks like `when` treats unknowns differently than `if` does. I'm inclined to conclude that that's a bug for `when`, rather than a cautionary note for the model developer.

sanderr:
It's been a while, but I finally found the time to address review comments.

edvgui (approved):
Lgtm.  I was wondering if it wouldn't add value to the example models to augment them with `std::assert(std::is_unknown(...), "")` because at the moment, the model will be valid (it will compile) whether the unknowns actually behave as advertised or not, and the only thing illustrating that something is unknown is its name.  I could hear a different take on this though, this is not a test case after all.

sanderr:
I considered it, but I found it too verbose for the docs. That said, I should probably add a test case that loads the snippets.

wouterdb (approved):

edvgui (approved):

Issue #6500, opened 2023-09-14
Improve error reporting in string formatters


bartv:
Consider the following code:

```
entity Test:
end

implementation test for Test:
    std::print("This is test {{n}}")
    std::print(f"This is test {n}")
end

implement Test using test

Test()
```

It will report the error on the line of the implement statement and not the line where the formatter is located:

```
variable n not found (reported in implement __config__::Test using test when true (./test.cf:9:11))
```

sanderr:
Does not seem to be specific to string formatters: any name lookup error in an implementation results in this behavior.

PR #6844, opened 2023-12-08
Issue/6500 improve error reporting in string formatters

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/6500

# Self Check:

- [ ] Changelog entry


src/inmanta/execute/runtime.py
```diff
@@ -1014,6 +1014,9 @@ def _unsafe_execute(self) -> None:
     def execute(self) -> None:
         try:
             self._unsafe_execute()
+        except NotFoundException as e:
+            # In case of a NotFoundException, keep all the info of the exception as is
+            raise e
         except RuntimeException as e:
             e.set_statement(self.owner)
```
wouterdb, line 1021: I wonder, it is ever appropriate to overwrite the statement if one is already set? 
-> If it is not appropriate, we can simplify all of this very significantly but not overwriting. 

Second, how do we know that all `NotFoundExceptions` that pass by here do have a statement set?

Hugo-Inmanta:
Closed in favor of https://github.com/inmanta/inmanta-core/pull/6867

PR #6867, opened 2023-12-14
proposal

wouterdb:
# Description

closes https://github.com/inmanta/inmanta-core/issues/6500

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/ast/__init__.py
+++ b/src/inmanta/ast/__init__.py
@@ -798,7 +798,7 @@ class OptionalValueException(RuntimeException):
 
     def __init__(self, instance: "Instance", attribute: "Attribute") -> None:
         RuntimeException.__init__(
-            self, instance, f"Optional variable accessed that has no value (attribute `{attribute}` of `{instance}`)"
+            self, None, f"Optional variable accessed that has no value (attribute `{attribute}` of `{instance}`)"
         )
         self.instance = instance
         self.attribute = attribute
--- a/src/inmanta/ast/__init__.py
+++ b/src/inmanta/ast/__init__.py
@@ -554,8 +554,8 @@ class CompilerException(Exception, export.Exportable):
         # store root namespace so error reporters can inspect the compiler state
         self.root_ns: Optional[Namespace] = None
 
-    def set_location(self, location: Location, replace_location: bool = False) -> None:
-        if self.location is None or replace_location:
+    def set_location(self, location: Location) -> None:
+        if self.location is None:
             self.location = location
 
     def get_message(self) -> str:
@@ -623,13 +623,13 @@ class RuntimeException(CompilerException):
             self.set_location(stmt.get_location())
             self.stmt = stmt
 
-    def set_statement(self, stmt: "Locatable", replace: bool = True, replace_location: bool = False) -> None:
+    def set_statement(self, stmt: "Locatable", replace: bool = True) -> None:
         for cause in self.get_causes():
             if isinstance(cause, RuntimeException):
                 cause.set_statement(stmt, replace)
 
         if replace or self.stmt is None:
-            self.set_location(stmt.get_location(), replace_location)
+            self.set_location(stmt.get_location())
             self.stmt = stmt
 
     def format(self) -> str:
--- a/src/inmanta/ast/statements/__init__.py
+++ b/src/inmanta/ast/statements/__init__.py
@@ -529,9 +529,13 @@ class ReferenceStatement(ExpressionStatement):
         return [req for v in self.children for req in v.requires()]
 
     def requires_emit(self, resolver: Resolver, queue: QueueScheduler) -> dict[object, VariableABC]:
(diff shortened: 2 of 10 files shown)


src/inmanta/ast/statements/__init__.py
```diff
@@ -529,9 +529,13 @@ def requires(self) -> list[str]:
...
     def requires_emit(self, resolver: Resolver, queue: QueueScheduler) -> dict[object, VariableABC]:
-        requires: dict[object, VariableABC] = super().requires_emit(resolver, queue)
-        requires.update({rk: rv for i in self.children for (rk, rv) in i.requires_emit(resolver, queue).items()})
-        return requires
+        try:
+            requires: dict[object, VariableABC] = super().requires_emit(resolver, queue)
+            requires.update({rk: rv for i in self.children for (rk, rv) in i.requires_emit(resolver, queue).items()})
+            return requires
+        except RuntimeException as e:
+            e.set_statement(self, False)
```
wouterdb, line 537: this does the trick

tests/compiler/test_conditional.py
```diff
@@ -464,7 +464,7 @@ def test_1808_non_boolean_not(snippetcompiler):
 if not str:
 end
         """,
-        "Invalid value `some_string`: `not` expects a boolean (reported in If ({dir}/main.cf:3))",
+        "Invalid value `some_string`: `not` expects a boolean (reported in Not(str) ({dir}/main.cf:3))",
```
Hugo-Inmanta, line 467: This is an improvement

src/inmanta/execute/runtime.py
```diff
@@ -1015,8 +1015,8 @@ def execute(self) -> None:
         try:
             self._unsafe_execute()
         except RuntimeException as e:
-            e.set_statement(self.owner)
-            e.location = self.owner.location
+            replace: bool = isinstance(e, OptionalValueException)
```
Hugo-Inmanta, line 1018: Special case for [test_optional_unset](https://github.com/inmanta/inmanta-core/blob/5c955d1278fe8d90440e598aa075526225adcce7/tests/compiler/test_relations.py#L711) 
Without this, the constructor would be reported.

Hugo-Inmanta (changes requested):
I tried to see what it took to fix the failing tests, I added some comments directly in the code showing what we lose/gain wrt the original. I still have a hard time pondering what we gain vs what we lose as a whole

src/inmanta/execute/runtime.py
wouterdb: I fixed this differently

src/inmanta/ast/__init__.py
```diff
@@ -798,7 +798,7 @@ class OptionalValueException(RuntimeException):
 
     def __init__(self, instance: "Instance", attribute: "Attribute") -> None:
         RuntimeException.__init__(
-            self, instance, f"Optional variable accessed that has no value (attribute `{attribute}` of `{instance}`)"
+            self, None, f"Optional variable accessed that has no value (attribute `{attribute}` of `{instance}`)"
```
wouterdb, line 801: This is I believe a more sensible fix

Hugo-Inmanta (approved):

src/inmanta/ast/statements/generator.py
```diff
@@ -143,7 +143,7 @@ def execute(self, requires: dict[object, object], instance: Instance, queue: Que
         try:
             inmanta_type.Bool().validate(condition)
         except RuntimeException as e:
-            e.set_statement(self.implements)
+            e.set_statement(self)
```
wouterdb, line 146: Why is this?

Hugo-Inmanta: This was because of this test: https://github.com/inmanta/inmanta-core/blob/1be072d3347839804f8c598520089966c5cb73b5/tests/compiler/test_conditional.py#L483

```diff
+ ('Invalid value `some_string`: the condition for a conditional implementation '
+ 'can only be a boolean expression (reported in <inmanta.ast.entity.Implement '
+ 'object at 0x7f8be8da0a90> (/tmp/tmpq26mn7x6/main.cf:10:11))')
- ('Invalid value `some_string`: the condition for a conditional implementation '
- 'can only be a boolean expression (reported in implement __config__::A using '
- 'i when str (/tmp/tmpq26mn7x6/main.cf:10:11))')
```

Doing it this way might have some implications that I'm missing, the big picture is still not super clear to me.

Hugo-Inmanta: Fixed in a different way: don't set the statement in here, and instead set it one layer above through https://github.com/inmanta/inmanta-core/blob/b830762ba7e217e149bcac261765d45c4ee7787b/src/inmanta/execute/runtime.py#L1018
wdyt ?

Hugo-Inmanta (changes requested):
I set the statement one layer up for `SubConstructor`s wdyt ?

Hugo-Inmanta (approved):

PR #1451, opened 2019-10-02
Add is defined to language reference.

arnaudsjs:
Fixes #1419 #1454

- [x] add testcase for using is defined in implements (with implicit and explicit self)
- [x] add testcase for using is defined in if statement in an implementation refering to 
   - [x] block scope variable
   - [x] variable on implicit self
   - [x] variable on explicit self
   - [x] global in containing module

PR #6868, opened 2023-12-14
Issue/1451 improve performances of lsm_service_resources_list

FloLey:
# Description

add optional argument to some of the functions to be able to pass an existing connection to querries.
needed by:  https://github.com/inmanta/inmanta-lsm/pull/1534
closes [#1451 ](https://github.com/inmanta/inmanta-lsm/issues/1451)

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/1451-performance-lsm-service-resources-list.yml
+++ b/changelogs/unreleased/1451-performance-lsm-service-resources-list.yml
@@ -1,0 +1,5 @@
+description: Improve the performance of the lsm_services_resources_list by reusing the same connection
+issue-nr: 1451
+issue-repo: inmanta-lsm
+change-type: patch
+destination-branches: [master, iso7, iso6]
--- a/changelogs/unreleased/1451-performance-lsm-service-resources-list.yml
+++ b/changelogs/unreleased/1451-performance-lsm-service-resources-list.yml
@@ -1,4 +1,4 @@
-description: Improve the performance of the lsm_services_resources_list by reusing the same connection
+description: Add a connection attribute to some methods in the database layer.
 issue-nr: 1451
 issue-repo: inmanta-lsm
 change-type: patch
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/1451-performance-lsm-service-resources-list.yml
```diff
@@ -0,0 +1,5 @@
+description: Improve the performance of the lsm_services_resources_list by reusing the same connection
```
arnaudsjs, line 1:
```suggestion
description: Add a connection attribute to some methods in the database layer.
```

src/inmanta/server/services/resourceservice.py
```diff
@@ -243,7 +244,7 @@ async def get_resource(
                 action_name = log_action.name
 
             actions = await data.ResourceAction.get_log(
-                environment=env.id, resource_version_id=resource_id, action=action_name, limit=log_limit
+                environment=env.id, resource_version_id=resource_id, action=action_name, limit=log_limit, connection=connection
```
arnaudsjs, line 247: The call to `data.Resource.get()` in this method, is also a call to the database.

arnaudsjs (changes requested):

src/inmanta/server/services/resourceservice.py
```diff
@@ -222,14 +222,15 @@ async def get_resource(
...
+        connection: Optional[Connection] = None,
     ) -> Apireturn:
         # Validate resource version id
         try:
             Id.parse_resource_version_id(resource_id)
         except ValueError:
             return 400, {"message": f"{resource_id} is not a valid resource version id"}
 
-        resv = await data.Resource.get(env.id, resource_id)
+        resv = await data.Resource.get(env.id, resource_id, connection)
```
arnaudsjs, line 233: If `connection is None`, this method will acquire a connection twice. Once inside the `data.Resource.get()` method and once inside the `data.ResourceAction.get_log()`. We should make sure that a connection is only acquired once and reused for both queries.

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #6870, opened 2023-12-18
Issue/6501 wrong error message int

### LINKED ISSUE #6501: Wrong error message provided for integer type
Consider the following code:

```
entity Test:
    int test
end

x = Test()
x.test = "1234"
```

This will result in 
```
Could not set attribute `test` on instance `__config__::Test (instantiated at ./test.cf:8)` (reported in x.test = '1234' (./test.cf:9))
caused by:
  Invalid value '1234', expected Number (reported in x.test = '1234' (./test.cf:9))
```

I would expect here to see `int`. Similar when the type is `number` it should say `number` not `Number`

Even better would be:

` Invalid value '1234' of type string, expected int (reported in x.test = '1234' (./test.cf:9))`

FloLey:
# Description

Make sure the right type is shown in the error message when a value with a wrong type is used.
This was fixed [here](https://github.com/inmanta/inmanta-core/pull/6772/files#diff-0a38a04c81c27002daab3fea8b2c4eac57e25c020dcca78a124df914b5de6ab3R257). This PR adds a test
and improves the way the error message is created.

closes #6501 

# Self Check:

- [x] Changelog entry


changelogs/unreleased/6501-fix-error-message-wrong-type.yml
```diff
@@ -0,0 +1,4 @@
+description: Fix a wrong error message when using a wrong type in the inmanta language.
```
arnaudsjs, line 1: I don't get this description. As far as I can see this PR only makes sure that the types are represented using a small leading character instead of a capital letter (Bool becomes bool for example). Am I missing something?

arnaudsjs (changes requested):

changelogs/unreleased/6501-fix-error-message-wrong-type.yml
FloLey: indeed I created this description before the fix was already done. I'll change it;

arnaudsjs (approved):

PR #6887, opened 2023-12-20
update the pip config in the docs of lsm quickstart

FloLey:
1. update the pip config in the docs of lsm quickstart
2. fix trace in stable api doc (https://github.com/inmanta/inmanta-core/commit/f8c7fa371b85e8b6563e64cddf2e557c21e48064#diff-d13056303bf7d5951ef0b02236d5b40efdd6d557e566775f1716bebf0bbe4977L80-R69 BaseModel.config was removed )

### DIFF
--- a/docs/lsm/quickstart/quickstart.rst
+++ b/docs/lsm/quickstart/quickstart.rst
@@ -82,13 +82,18 @@ This folder contains a project.yml, which looks like this:
     copyright: 2022 Inmanta
     modulepath: libs
     downloadpath: libs
-    repo:
-    - type: package
-        url: https://packages.inmanta.com/public/quickstart/python/simple/
-    - type: package
-        url: https://packages.inmanta.com/<token>/inmanta-service-orchestrator-|version_major|-stable/python/simple/
     install_mode: release
-    requires:
+    # The repo block is for ISO6, inmanta-core<=10.0
+    # This example requires licensed modules,
+    # replace <token> with inmanta access token you received with your license
+    repo:
+      - type: package
+        url: https://packages.inmanta.com/<token>/inmanta-service-orchestrator-6-stable/python/simple/
+    # The pip block is for ISO7 and up, inmanta-core>10.0
+    # This example requires licensed modules,
+    # replace <token> with inmanta access token you received with your license
+    pip:
+      index_url: https://packages.inmanta.com/<token>/inmanta-service-orchestrator-|version_major|-stable/python/simple/
 
 .. note::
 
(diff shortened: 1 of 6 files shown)


arnaudsjs (changes requested):
The regular quickstart has the same issue. We should also update that one.

docs/lsm/quickstart/quickstart.rst
```diff
@@ -82,13 +82,18 @@ This folder contains a project.yml, which looks like this:
...
+    # This example requires licensed modules,
+    # replace <token> with inmanta access token you received with your license
+    repo:
+      - type: package
+        url: https://packages.inmanta.com/<token>/inmanta-service-orchestrator-6-stable/python/simple/
+    # The pip block is for ISO7 and up, inmanta-core>10.0
+    # This example requires licensed modules,
+    # replace <token> with inmanta access token you received with your license
+    pip:
+      index_url: https://packages.inmanta.com/<token>/inmanta-service-orchestrator-|version_major|-stable/python/simple/
```
arnaudsjs, line 96: This PR goes into master and iso7, so there is no need to mention the config for iso6 here.

docs/quickstart.rst
```diff
@@ -60,11 +60,16 @@ This folder contains a **project.yml**, which looks like this:
     copyright: 2022 Inmanta
     modulepath: libs
     downloadpath: libs
+    # The repo block is for ISO6, inmanta-core<=10.0
```
arnaudsjs, line 63: Same here.

arnaudsjs (changes requested):

docs/lsm/quickstart/quickstart.rst
FloLey: I though we wanted to stay as close as possible to the content of this file in the example repo which contains this part. I removed it.

docs/lsm/quickstart/quickstart.rst
```diff
@@ -82,13 +82,11 @@ This folder contains a project.yml, which looks like this:
...
     modulepath: libs
     downloadpath: libs
-    repo:
-    - type: package
-        url: https://packages.inmanta.com/public/quickstart/python/simple/
-    - type: package
-        url: https://packages.inmanta.com/<token>/inmanta-service-orchestrator-|version_major|-stable/python/simple/
     install_mode: release
-    requires:
+    # This example requires licensed modules,
```
arnaudsjs, line 86: The `install_mode` option is deprecated in favor of `pip.pre`, so we should not set it here anymore.

arnaudsjs (changes requested):

docs/lsm/quickstart/quickstart.rst
FloLey: I removed it everywhere

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #6889, opened 2023-12-20
First step towards removing the state copying

wouterdb:
# Description

work still todo:

- [x] make migration work
- [x] cleanup of code
- [x] make cleanup mechanism for the new table
- [x] update get_resource_details 
- [x] get_resources_report
- [x] remove last_deploy field 
- [x] update ResourceView

## Type errors added / moved 

```
src/inmanta/data/__init__.py:4929: error: Argument 1 to "loads" has incompatible type "object"; expected "str | bytes | bytearray"  [arg-type]

```

##  Open opportunities
1.  track last and first version for simpler get_resources_report and  &get_resource_details*
2. track last_released for easier ResourceView
3. get rid of  mark deployed for known good state (biggest issue is tracking when a version is deployed and the dashboard)
4. add new state fields 
      - Deploy result -> derived from last_non_deploying_state
      - Resource status -> last deployed hash Vs current hash + non-recursive orphans 	
5. we may be able to get rid of mark_deployed

# Self Check:

- [x] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -4402,6 +4410,29 @@ def to_dto(self) -> m.ResourceAction:
         )
 
 
+class ResourcePersistentState(BaseDocument):
+    @classmethod
+    def table_name(cls) -> str:
+        # TODO is this a good name?
+        return "resource_persistent_state"
```
wouterdb, line 4417: Is this  good name?

src/inmanta/data/__init__.py
```diff
@@ -5681,10 +5734,19 @@ async def get_increment(
...
                 increment.append(resource)
             else:
-                next.append(resource)
+                # If no events, shortcut on known good
+                if (
+                    status == const.ResourceState.deployed.name
+                    and resource["attribute_hash"] == resource["last_deployed_attribute_hash"]
+                ):
+                    # Shortcut on easy good ones
+                    not_increment.append(resource)
```
sanderr, line 5747: I'm not convinced this is actually the same. What if another version is deploying at the moment? The previous implementation would inspect that version before ever getting to the last deployed one, and decide there that it should be in the increment if the hash does not match.

Here, the hash matches, but the intermediate deploying version might have a different hash.

Overall, I see how these changes are good opportunities, but unless they are really trivial, I would prefer to split them off into a separate PR to keep the scope contained.

src/inmanta/server/services/resourceservice.py
```diff
@@ -412,6 +413,11 @@ async def mark_deployed(
         :param version: Version of the resources to consider.
         :param filter: Filter function that takes a resource id as an argument and returns True if it should be kept.
         """
+        # TODO: can we get rid of mark deployed now????
+        # Looks feasible on the path for local dependencies, with start deploy
+        # what about CAD's?
+        # what about the dashboard?
```
sanderr, line 419: If you mean the web-console with "dashboard", I *think* it doesn't rely on `mark_deployed` at all.

src/inmanta/server/services/resourceservice.py
```diff
@@ -722,49 +731,6 @@ async def resource_deploy_done(
...
-        propagate_deploy_state: bool,
-    ) -> None:
-        if propagate_deploy_state or propagate_last_produced_events:
-            # lock out release version
-            await env.acquire_release_version_lock(connection=connection)
-            latest_version = await data.ConfigurationModel.get_version_nr_latest_version(env.id, connection=connection)
-
-            for resource_id in resource_ids:
-                if latest_version is not None and latest_version > resource_id.version:
-                    # we are stale, forward propagate our status
```
sanderr, line 742: Can you elaborate on why this can go now? The resource table still has a state field, which is still the authority on a resource's state until we move those checks to the read path, isn't it?

sanderr: I have to give this comment more thought. That'll be for Monday.

sanderr (commented):
The overall idea looks good to me, but it's difficult to reason on.

src/inmanta/data/__init__.py
wouterdb: I haven't quite figured that one out thb

src/inmanta/data/__init__.py
```diff
@@ -5029,66 +5070,31 @@ async def get_resources_in_resource_sets(
...
-               )
-           WHERE new_resource.model=$1
-           AND new_resource.environment=$2
-           AND new_resource.last_produced_events is null"""
-        await cls._execute_query(query, to_version, environment, from_version, connection=connection)
-
     async def insert(self, connection: Optional[asyncpg.connection.Connection] = None) -> None:
         self.make_hash()
         await super().insert(connection=connection)
+        # TODO: On conflict or is not exists or just make every update an upsert?
```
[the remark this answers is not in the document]
wouterdb: These are three options, which one do I take?

tests/agent_server/test_increment_interactions.py
```diff
@@ -92,7 +92,7 @@ async def make_version() -> int:
 
     # add new version
     # hammer it with new versions!
-    new_versions_to_add = 20
+    new_versions_to_add = 2
```
wouterdb, line 95: increase again?

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
sanderr: Not sure how I missed that, I found it a weirdly phrased question the way I interpreted it. I think the `ON CONFLICT DO NOTHING` approach is good. I think it's probably the least restrictive of the three, locking-wise.

src/inmanta/data/__init__.py
```diff
@@ -4755,6 +4766,34 @@ async def get_resources_for_version_raw(
...
+        query = f"""
+        SELECT {collect_projection(projection, 'r')}, {collect_projection(projection_presistent, 'ps')}
+            FROM {cls.table_name()} r JOIN resource_persistent_state ps ON r.resource_id = ps.resource_id
+            WHERE r.environment=$1 AND ps.environment = $1 and r.model = $2;"""
+
+        resource_records = await cls._fetch_query(query, environment, version, connection=connection)
+        resources = [dict(record) for record in resource_records]
+        for res in resources:
+            if "attributes" in res:
+                res["attributes"] = json.loads(res["attributes"])
```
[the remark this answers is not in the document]
wouterdb: I added a comment, but it is not used on any code path that returns actual Resources

src/inmanta/data/__init__.py
```diff
@@ -5038,66 +5086,31 @@ async def get_resources_in_resource_sets(
...
+            connection=connection,
+        )
 
     @classmethod
     async def insert_many(
         cls, documents: Sequence["Resource"], *, connection: Optional[asyncpg.connection.Connection] = None
     ) -> None:
         for doc in documents:
             doc.make_hash()
+        # TODO performance?
```
wouterdb, line 5106: is this good enough

src/inmanta/server/services/resourceservice.py
```diff
@@ -224,6 +224,7 @@ async def get_resource(
         log_limit: int,
         connection: Optional[Connection] = None,
     ) -> Apireturn:
+        # TODO: this output will slightly change
```
wouterdb, line 227: Field have been added to the DB and they leaked out here, now we remove them again, ....

tests/server/test_resource_history.py
```diff
@@ -92,234 +125,184 @@ async def env_with_resources(server, client):
...
     # A resource with multiple resources in its requires list, and multiple versions where it was released,
     # and is also present in versions that were not released
-    resources["std::File[internal,path=/tmp/dir1/file1]"].append(
-        await create_resource(
-            "/tmp/dir1/file1",
-            ResourceState.undefined,
-            1,
-            {"key1": "val1", "requires": ["std::Directory[internal,path=/tmp/dir1]"]},
-        )
+    await resource_factory.create_resource(
```
wouterdb, line 132: This is a plain refactor: moved out some duplication

tests/server/test_resource_history.py
```diff
@@ -339,23 +322,23 @@ async def test_resource_history(client, server, env_with_resources):
     expected = [
         {
             "date": cm_times[1].astimezone(datetime.timezone.utc),
-            "attributes": resources[resource_with_long_history][1].attributes,
+            "attributes": resources[resource_with_long_history][2].attributes,
```
wouterdb, line 325: I index on version now instead of array (i.e. always +1 )

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
arnaudsjs: I agree with the `DO NOTHING` approach

arnaudsjs (changes requested):

tests/conftest.py
```diff
@@ -263,7 +265,7 @@ def postgres_db(request: pytest.FixtureRequest):
             sublogger = logging.getLogger("pytest.postgresql.deadlock")
             for line in fh:
                 sublogger.warning("%s", line)
-        os.remove(pg_logfile)
+        # os.remove(pg_logfile)
```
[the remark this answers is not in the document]
wouterdb:
```suggestion
        os.remove(pg_logfile)
```

src/inmanta/db/util.py
```diff
@@ -277,3 +278,57 @@ async def replace_enum_type(new_type: EnumUpdateDefinition, *, connection: Conne
             if default:
                 await connection.execute(f"ALTER TABLE {table} ALTER COLUMN {column} SET DEFAULT '{default}'")
     await connection.execute(f"DROP TYPE {temp_name}")
+
+
+class ConnectionMaybeInTransaction(abc.ABC):
```
wouterdb, line 283: Cleanup of unrelated mess: I made util.db, but db.util already existed

arnaudsjs (approved):

PR #6892, opened 2023-12-22
Fixed type resolution for anchormaps

wouterdb:
# Description

Fixed anchormap
# Self Check:

- [x] Changelog entry
- [ x] No (preventable) type errors (check using make mypy or make mypy-diff)

### DIFF
--- a/src/inmanta/ast/statements/call.py
+++ b/src/inmanta/ast/statements/call.py
@@ -83,8 +83,8 @@ class FunctionCall(ReferenceStatement):
 
     def normalize(self, *, lhs_attribute: Optional[AttributeAssignmentLHS] = None) -> None:
         ReferenceStatement.normalize(self)
-        self.anchors = [TypeReferenceAnchor(self.namespace, self.name)]
         func = self.namespace.get_type(self.name)
+        self.anchors = [TypeReferenceAnchor(self.namespace, self.name)]
         if isinstance(func, InmantaType.Primitive):
             self.function = Cast(self, func)
         elif isinstance(func, plugins.Plugin):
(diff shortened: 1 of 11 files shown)


src/inmanta/ast/statements/call.py
```diff
@@ -83,8 +83,8 @@ def __init__(
 
     def normalize(self, *, lhs_attribute: Optional[AttributeAssignmentLHS] = None) -> None:
         ReferenceStatement.normalize(self)
-        self.anchors = [TypeReferenceAnchor(self.namespace, self.name)]
         func = self.namespace.get_type(self.name)
+        self.anchors = [TypeReferenceAnchor(self.namespace, self.name)]
```
wouterdb, line 87: Typeanchor acted up here,so I left is as it is

FloLey: Do you have an idea why?

tests/test_compiler_entrypoints.py
```diff
@@ -346,3 +346,37 @@ def test_constructor_with_inferred_namespace(snippetcompiler):
     range_target = Range(target_path, 1, 8, 1, 12)
 
     assert (range_source, range_target) in anchormap
+
+
+def test_constructor_renamed_namespace(snippetcompiler):
+    """
+    Test that the anchor for a constructor with `import a as b` work
```
FloLey, line 353:
```suggestion
    Test that the anchor for a constructor with `import a as b` works
```

FloLey (approved):

src/inmanta/ast/statements/call.py
wouterdb: yes, this is not quite a type, it might work is I move it a few lines down, but I didn't want to take risks

PR #6911, opened 2023-12-27
 issue/6053-unclear-error-list-validation

### LINKED ISSUE #6053: unclear error message when list validation fails
The model `std::print(std::count("hello"))` fails with
```
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/app.py", line 769, in app
    options.func(options)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/app.py", line 368, in compile_project
    do_compile()
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/compiler/__init__.py", line 81, in do_compile
    compiler.handle_exception(e)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/compiler/__init__.py", line 264, in handle_exception
    self._handle_exception_export(e)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/compiler/__init__.py", line 270, in _handle_exception_export
    raise exception
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/compiler/__init__.py", line 262, in handle_exception
    self._handle_exception_datatrace(exception)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/compiler/__init__.py", line 274, in _handle_exception_datatrace
    raise exception
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/compiler/__init__.py", line 76, in do_compile
    success = sched.run(compiler, statements, blocks)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/execute/scheduler.py", line 417, in run
    next.execute()
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/ast/statements/call.py", line 289, in execute
    raise e
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/ast/statements/call.py", line 285, in execute
    self.function.resume(requires, self.resolver, self.queue, self.result)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/ast/statements/call.py", line 129, in resume
    self.function.call_in_context(arguments, kwargs, resolver, queue, result)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/ast/statements/call.py", line 229, in call_in_context
    no_unknows = self.plugin.check_args(args, kwargs)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/plugins.py", line 432, in check_args
    if not is_valid(self.arguments[i], self.argtypes[i], args[i]):
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/plugins.py", line 424, in is_valid
    if not self._is_instance(arg, expected_type):
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/plugins.py", line 387, in _is_instance
    return arg_type.validate(value)
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/ast/type.py", line 411, in validate
    if not List.validate(self, value):
  File "/home/sander/documents/projects/inmanta/inmanta-core/src/inmanta/ast/type.py", line 385, in validate
    raise RuntimeException(None, "Invalid value '%s', expected %s" % (value, self.type_string()))
inmanta.ast.RuntimeException: Invalid value 'hwllo', expected None (reported in std::count('hwllo') (/home/sander/documents/projects/inmanta/model-test/main.cf:53))
```

It looks like the `type_string` method of `inmanta.ast.type.List` isn't complete, or it is used when it should not (read its docstring!).

FloLey:
# Description

fix an unclear message regarding list validation failures.

closes #6053 
# Self Check:

- [x] Changelog entry


src/inmanta/ast/type.py
```diff
@@ -436,6 +436,9 @@ def validate(self, value: Optional[object]) -> bool:
 
         return True
 
+    def type_string(self) -> str:
+        return "list"
```
sanderr, line 440: This slightly changes the semantics of this class. I don't like that we now have two distinct classes returning `"list"` here, but that's more a property of the language (`list` in DSL vs `list` in plugin annotation) than anything else, so I think you took the right approach. That said, could you extend the docstring with something like "e.g. the `list` type in plugin annotations."? And perhaps refer to `LiteralList` for the DSL `list` type?

Also, an additional comment here could help future confusion.

```suggestion
        # this is not a type in the model, but it is used in plugin annotations, which are also part of the DSL.
        return "list"
```

src/inmanta/ast/type.py
```diff
@@ -762,7 +765,7 @@ def resolve_type(locatable_type: LocatableString, resolver: Namespace) -> Type:
     # quickfix issue #1774
     allowed_element_type: Type = Type()
     if locatable_type.value == "list":
-        return TypedList(allowed_element_type)
+        return List()
     if locatable_type.value == "dict":
```
sanderr, line 769: Looking at this code I expect the same issue for dict. Could you check that as well?

tests/compiler/test_list.py
```diff
@@ -556,3 +556,14 @@ def test_relation_list_duplicate_assignment(snippetcompiler):
...
+
+def test_error_list_validation(snippetcompiler):
+    snippetcompiler.setup_for_snippet(
+        """
+        std::print(std::count("hello"))
+        """
+    )
+    with pytest.raises(RuntimeException) as exc_info:
+        (_, scopes) = compiler.do_compile()
+    assert "Invalid value 'hello', expected list" in exc_info.value.msg
```
sanderr, line 569: `pytest.raises` has a `match` kwarg for this. Just an FYI, not a change request.

sanderr (changes requested):

src/inmanta/ast/type.py
FloLey: It seems we dont have the issue with dicts due to the usage of `type_string_internal` and `type_string` in TypedDict and Dict respectively

FloLey: I've added a test

sanderr (approved):

PR #6916, opened 2023-12-28
6861: Use 401 and 403 correctly

### LINKED ISSUE #6861: Use 401 and 403 correctly
Currently, we raise the same exception of a not-authenticated and not-authorized user.
This causes issues for the front-end 

- [ ] Make sure we use 401 and 403 correctly.

FloLey:
# Description

Currently, we raise the same exception for a not-authenticated and not-authorized user.
This causes issues for the front-end

closes #6861

# Self Check:

- [x] Changelog entry


FloLey:
I think those are the only one that are wrongly used an need a change.

Hugo-Inmanta (approved):

PR #6918, opened 2024-01-02
Issue/6866 log version startup

### LINKED ISSUE #6866: Log the version of the inmanta-server at startup

FloLey:
# Description

log server version at startup

closes #6866

# Self Check:

- [x] Changelog entry


src/inmanta/app.py
```diff
@@ -83,6 +83,7 @@
 
 @command("server", help_msg="Start the inmanta server")
 def start_server(options: argparse.Namespace) -> None:
+    LOGGER.info(f"Starting inmanta-server version {__version__}")
```
arnaudsjs, line 86: This will log the version of the inmanta-core component. The idea was to log the version of the product.

arnaudsjs (changes requested):

src/inmanta/app.py
```diff
@@ -83,6 +83,7 @@
 
 @command("server", help_msg="Start the inmanta server")
 def start_server(options: argparse.Namespace) -> None:
+    LOGGER.info(f"Starting inmanta-server version {COMPILER_VERSION}")
```
arnaudsjs, line 86: The compiler version and the product version are not the same thing. The log line should contains the product version. This is the version of the product python package. For ISO the product package is called `inmanta-service-orchestrator`. For OSS it's called `inmanta`.

arnaudsjs, line 86: Make sure to use the `%s` notation.
```suggestion
    LOGGER.info(f"Starting inmanta-server version %s", COMPILER_VERSION)
```

arnaudsjs (changes requested):

src/inmanta/app.py
FloLey: Ow now I see, thanks

src/inmanta/app.py
```diff
@@ -81,8 +82,30 @@
...
+    """
+    try:
+        # Try to get the version of 'inmanta-service-orchestrator' (ISO)
+        return importlib.metadata.version("inmanta-service-orchestrator")
+    except importlib.metadata.PackageNotFoundError:
+        # If not found, try 'inmanta' (OSS)
+        try:
+            return importlib.metadata.version("inmanta")
+        except importlib.metadata.PackageNotFoundError:
+            return
```
arnaudsjs, line 100: * I notice that this code is more or less a duplication of the code in the `FeatureManager` class (`FeatureManager.get_product_metadata().version`). I think we should rely on that logic instead.
* I would also move this log line into the `InmantaBootloader.start()` method. That way the log line is also present when the server is started in a different way than using the `inmanta` command.

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #6928, opened 2024-01-04
Simplify isinstance and issubclass checks in data layer

### LINKED ISSUE #6747: simplify isinstance and issubclass checks in data layer
Both core and inmanta-lsm have checks like `value.__class__ is some_type or isinstance(value, some_type)`, and similar for `issubclass`. We should see if we can drop the redundant first statement here. Just search for `__class__` in both these files.

First noted [here](https://github.com/inmanta/inmanta-lsm/pull/1491#discussion_r1397205633) but decided to keep the change out of that PR to isolate anything that may break on it.

arnaudsjs:
# Description

Simplify isinstance and issubclass checks in data layer

Part of #6747

# Self Check:

- [x] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -1987,7 +1987,7 @@ def _get_value(cls, value: object) -> object:
         if isinstance(value, dict):
             return json_encode(value)
 
-        if isinstance(value, DataDocument) or issubclass(value.__class__, DataDocument) or isinstance(value, BaseModel):
+        if isinstance(value, DataDocument) or isinstance(value, BaseModel):
```
sanderr, line 1990:
```suggestion
        if isinstance(value, (DataDocument, BaseModel)):
```

sanderr (approved):

FloLey (approved):

PR #6936, opened 2024-01-05
 issue/1535-multiple-prelude

FloLey:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/data/dataview.py
+++ b/src/inmanta/data/dataview.py
@@ -779,7 +779,7 @@ class ResourceHistoryView(DataView[ResourceHistoryOrder, ResourceHistory]):
         return {}
 
     def get_base_url(self) -> str:
-        return f"/api/v2/resource/{quote(self.rid,safe='')}/history"
+        return f"/api/v2/resource/{quote(self.rid, safe='')}/history"
 
     def get_base_query(self) -> SimpleQueryBuilder:
         query_builder = SimpleQueryBuilder(
@@ -1250,6 +1250,9 @@ class PreludeBasedFilteringQueryBuilder(SimpleQueryBuilder):
     """
     A query builder that applies any filters and the LIMIT statement to the prelude query rather than the outer query.
     The outer query may use the table name "prelude" to refer to the inner query.
+    Other preludes can be added to the query using the prelude_query_builder_extra argument.
+    This is a list of tuples where each tuple consist of the name that will be given to the CTE and
+    a SimpleQueryBuilder that will create the CTE. The filtering will only be done on the first prelude.
     """
 
     def __init__(
@@ -1292,19 +1295,23 @@ class PreludeBasedFilteringQueryBuilder(SimpleQueryBuilder):
             extra_prelude_queries.append(f"{prelude_name} AS ({extra_query})")
             extra_prelude_values.extend(extra_values)
 
+        # The filter_statements and values are not aligned. this is because we don't want to re-apply the filters of
+        # the prelude to the main query. We still need all the values and the offset to be able to apply additional
+        # filters.
         delegate = SimpleQueryBuilder(
             select_clause=self.select_clause,
             from_clause=self._from_clause,
             filter_statements=self.filter_statements,
-            values=self.values,
+            values=self._prelude_query_builder.values + self.values,
             db_order=self.db_order,
             limit=None,
             backward_paging=self.backward_paging,
             prelude=prelude_section,
             prelude_extra=extra_prelude_queries,
+            base_offset=len(self._prelude_query_builder.values) + 1,
         )
         full_query, values_full = delegate.build()
-        return full_query, prelude_values
+        return full_query, prelude_values + extra_prelude_values
 
     def select(self, select_clause: str) -> "PreludeBasedFilteringQueryBuilder":
         """Set the select clause of the query"""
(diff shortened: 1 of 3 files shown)


src/inmanta/data/dataview.py
```diff
@@ -1323,6 +1338,7 @@ def order_and_limit(
             db_order=db_order,
             backward_paging=backward_paging,
             prelude_query_builder=self._prelude_query_builder.order_and_limit(db_order, limit, backward_paging),
+            prelude_query_builder_extra=self._prelude_query_builder_extra,
```
FloLey, line 1341: Here I think I need to do something (as for the prelude_query_builder) but I don't understand why we need to apply a limit/order on the subqueries, I would have expected doing it only on ht main query was enough

src/inmanta/data/dataview.py
```diff
@@ -1332,4 +1348,5 @@ def filter(self, filter_statements: list[str], values: list[object]) -> "Prelude
             db_order=self.db_order,
             backward_paging=self.backward_paging,
             prelude_query_builder=self._prelude_query_builder.filter(filter_statements, values),
+            prelude_query_builder_extra=self._prelude_query_builder_extra,
```
FloLey, line 1351: Here I think I need to do something (as for the prelude_query_builder) but I don't understand why we need to apply a limit/order on the subqueries, I would have expected doing it only on ht main query was enough

tests/test_data.py
```diff
@@ -3228,3 +3229,101 @@ async def test_retrieve_optional_field_no_default(init_dataclasses_and_load_sche
...
+    )
+
+
+def test_prelude_extra_in_simple_query_builder(simple_query_builder):
+    query, _ = simple_query_builder.build()
+    assert (
+        """WITH prelude_query AS (SELECT * FROM prelude_table),
+extra_prelude_1 AS (SELECT * FROM extra_table1),
+extra_prelude_2 AS (SELECT * FROM extra_table2)
+SELECT *"""
```
arnaudsjs, line 3267: This query looks very strange. It doesn't have a FROM statement. Maybe you need to add a comment to describe the purpose of this query.

tests/test_data.py
```diff
@@ -3228,3 +3229,101 @@ async def test_retrieve_optional_field_no_default(init_dataclasses_and_load_sche
...
+SELECT mt.*
+
+        FROM main_table mt
+        JOIN prelude p ON mt.some_column = p.some_related_column
+        JOIN extra_prelude_1 ep ON mt.another_column = ep.another_related_column
+"""
+        in query
+    )
+    assert "prelude_value" == values[0]
+    assert "extra_prelude_value" == values[1]
```
arnaudsjs, line 3309: Why not assert the content of values exhaustively?

tests/test_data.py
```diff
@@ -3228,3 +3229,101 @@ async def test_retrieve_optional_field_no_default(init_dataclasses_and_load_sche
...
+SELECT *
+FROM extra_prelude_table
+WHERE extra_prelude_column = $2)
+SELECT mt.*
+
+        FROM main_table mt
+        JOIN prelude p ON mt.some_column = p.some_related_column
+        JOIN extra_prelude_1 ep ON mt.another_column = ep.another_related_column
+"""
+        in query
```
arnaudsjs, line 3306: Why use `in` here? This way we cannot be sure whether query contains more than the given string.

tests/test_data.py
```diff
@@ -3228,3 +3229,101 @@ async def test_retrieve_optional_field_no_default(init_dataclasses_and_load_sche
...
+    return SimpleQueryBuilder(
+        select_clause="SELECT *",
+        from_clause="FROM prelude_table",
+        filter_statements=["prelude_column = $1"],
+        values=["prelude_value"],
+    )
+
+
+def test_prelude_extra_in_simple_query_builder(simple_query_builder):
+    query, _ = simple_query_builder.build()
```
arnaudsjs, line 3262: Why not assert values?

src/inmanta/data/dataview.py
```diff
@@ -1271,27 +1272,39 @@ def __init__(
...
     def build(self) -> tuple[str, list[object]]:
         prelude_query, prelude_values = self._prelude_query_builder.build()
-        prelude_query_in_with_block = f"WITH prelude AS ({prelude_query})"
-        delegate: SimpleQueryBuilder = SimpleQueryBuilder(
+        prelude_section = f"prelude AS ({prelude_query})"
+        extra_prelude_values = []
+        extra_prelude_queries = []
+
+        for idx, extra_builder in enumerate(self._prelude_query_builder_extra):
+            extra_query, extra_values = extra_builder.build()
```
arnaudsjs, line 1290: How do you make sure that the parameter index of this query doesn't overlap with the previous prelude?

arnaudsjs (changes requested):

src/inmanta/data/dataview.py
FloLey: Indeed this is an issue, I don't know yet how to fix it. I encountered it on my branch on lsm: both my prelude use the same parameter but the query is now expecting 2 different parameters. I'll have to think about this.

src/inmanta/data/dataview.py
```diff
@@ -1271,27 +1272,39 @@ def __init__(
...
+            prelude_name = f"extra_prelude_{idx + 1}"
```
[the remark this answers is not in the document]
FloLey: Indeed, I'll change it: instead of a list of preludes the argument will become a list of tuples where the first element of each tuple is the name of the table and the second argument is the prelude

tests/test_data.py
FloLey: You mean why not use real tables and execute queries on those?

Issue #6743, opened 2023-11-17
Improve error message when installing module in editable mode


hlloreda:
I got this error when I tried to install a new module in editable mode.

![image](https://github.com/inmanta/inmanta-core/assets/32882487/6b749930-4021-47ba-99ac-c2b87575ce6e)

Adding more flags helps us find out what's causing the issue:

![image](https://github.com/inmanta/inmanta-core/assets/32882487/fc657ef3-4a7f-48c4-babf-34d372fbf03d)

But I think we should be able to figure out the source of the problem without adding these flags.

sanderr:
Now that we've disabled/deprecated editable mode for this command, first task is to check if this is still an issue without editable mode.

PR #6937, opened 2024-01-05
Issue/6743 improve error message when installing module in editable mode

Hugo-Inmanta:
# Description

Tried running `inmanta module add` for some v1 and v2 modules.
I noticed the following issue with v1 module with no "requires" field in their module.yml:

```
(.env) hugo@hugo-Latitude-5421:~/tmp/tmp_project/module_install_test$ inmanta module add --v1 net
inmanta.warnings         WARNING ProjectConfigurationWarning: Setting a pip index through the `repo.url` option with type `package` in the project.yml file is no longer supported and will be ignored. Please set the pip index url through the `pip.index_url` option instead.
inmanta.warnings         WARNING ModuleDeprecationWarning: Module net has been deprecated
'NoneType' object is not iterable

```

closes https://github.com/inmanta/inmanta-core/issues/6743

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/6743-improve-error-message-installing-module-in-editable-mode.yml
+++ b/changelogs/unreleased/6743-improve-error-message-installing-module-in-editable-mode.yml
@@ -1,0 +1,4 @@
+description: Fix error when adding a v1 module that doesn't require other modules.
+issue-nr: 6743
+change-type: patch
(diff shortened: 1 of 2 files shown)


sanderr (approved):
I'm not sure if this will ever occur in practice for recent modules, but best be safe.

PR #6940, opened 2024-01-08
Small performance optimization

wouterdb:
# Description

Ensure the parameter refresh call is done without waiting for it

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/protocol/methods.py
+++ b/src/inmanta/protocol/methods.py
@@ -828,6 +828,10 @@ def get_parameter(tid: uuid.UUID, agent: str, resource: dict):
     """
     Get all parameters/facts known by the agents for the given resource
 
+    This method will not actually return them.
+    This call wil register the request with the agent and return,
+    The agent will push the parameters back to the server when they are available.
+
     :param tid: The environment
     :param agent: The agent to get the parameters from
     :param resource: The resource to query the parameters from
--- a/src/inmanta/protocol/methods.py
+++ b/src/inmanta/protocol/methods.py
@@ -815,7 +815,15 @@ def set_parameters(tid: uuid.UUID, parameters: list):
 # Get parameters from the agent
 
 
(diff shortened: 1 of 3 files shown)


arnaudsjs (approved):

src/inmanta/protocol/methods.py
```diff
@@ -815,7 +815,15 @@ def set_parameters(tid: uuid.UUID, parameters: list):
...
 
-@method(path="/agent_parameter", operation="POST", server_agent=True, timeout=5, arg_options=AGENT_ENV_OPTS, client_types=[])
+@method(
+    path="/agent_parameter",
+    operation="POST",
+    server_agent=True,
+    timeout=5,
+    arg_options=AGENT_ENV_OPTS,
+    client_types=[],
+    reply=False,
```
sanderr, line 825: I'm quite confused by this. The method name and docstring heavily imply that this returns relevant information, and its handler does in fact return something. Yet by my understanding of `reply=False` we will now no longer return anything. What am I missing?

sanderr (changes requested):

src/inmanta/protocol/methods.py
sanderr: I see now that it returns nothing more than a status code. I think we should update the docstring, because it presents a false picture.

arnaudsjs: Iirc, the `reply` option only has a meaning in the context of our client. If set to True, our client will not wait for a reply from the server. Which does not imply that the server doesn't return anything.

sanderr: If I understand correctly it's used by the server as well. Not that the handler can not return a reply, but that it [is not expected to](https://github.com/inmanta/inmanta-core/blob/25e7811826d2c9e1e1c1f9f51779b28acacb1b3c/src/inmanta/protocol/rest/__init__.py#L475C37-L475C37).

arnaudsjs: Indeed. I didn't know that.

sanderr (approved):
Thanks

PR #6957, opened 2024-01-10
Add arithmetic support to the language

### LINKED ISSUE #1799: Add arithmetic to the language
Add support for + - * / for numbers and integers.

arnaudsjs:
# Description

**Draft PR that add the support for the + operator. Once approved, I will add the other operators.**

Add arithmetic support to the language.

closes #1799

# Self Check:

- [ ] Changelog entry


tests/compiler/test_arithemetic.py
```diff
@@ -0,0 +1,91 @@
...
+import textwrap
+
+from inmanta import compiler
+
+
+def test_plus(snippetcompiler) -> None:
+    snippetcompiler.setup_for_snippet(
+        textwrap.dedent(
+            """
+            var1 = 1 + 2
```
bartv, line 27: does it also work in expression? so the where clause of implementation, if, ...

Does is work in the body of a list comprehension?

arnaudsjs: I'll extend the test case.

arnaudsjs: Fix had to update the precedence rules to make it work in expressions. That's fixed now.

tests/compiler/test_arithemetic.py
```diff
@@ -0,0 +1,122 @@
...
+               fail = 1
+               fail = 2
+            end
+
+            ######################
+            # List comprehension #
+            ######################
+
+            seq = std::sequence(2)
+            filtered = [elem for elem in seq if elem + 1 < 2]
```
bartv, line 114: and in the body for the comprehension? Something like:

```suggestion
            filtered = [elem + 10 for elem in seq if elem + 1 < 2]
```

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -60,9 +60,22 @@
...
+    "SEP",
+    "STRING",
+    "MLS",
+    "CMP_OP",
+    "REGEX",
+    "REL",
+    "PEQ",
+    "RSTRING",
+    "FSTRING",
+    "ARITHMETIC_OP",
```
sanderr, line 77: I think a generic arithmetic token will not work. The arithmetic operators differ from e.g. the comparison operators in that they have unequal precedence (which honestly might have been a good idea for the comparison operators as well). Since precedence is specified by token, I think we need different tokens. At least one per precedence level, but honestly I don't see a strong reason to not just create one token per operator. You can always bundle them in a single parser production (rather than a lexer token).

It's been a while since I've been involved in the parser, so let me know if you think I'm missing something.

tests/compiler/test_arithemetic.py
```diff
@@ -0,0 +1,122 @@
...
+            # List comprehension #
+            ######################
+
+            seq = std::sequence(2)
+            filtered = [elem + 10 for elem in seq if elem + 1 < 2]
+            std::assert(10 in filtered)
+            std::assert(11 not in filtered)
+            """.strip(
+                "\n"
+            )
```
sanderr, line 119: FYI, I recently learned that Python allows you to escape the leading newline with a backslash:

```python
mls = """\
    this is a multiline string
    without leading backslash
"""
```

tests/compiler/test_arithemetic.py
```diff
@@ -0,0 +1,122 @@
...
+
+            t_int_and_float = TestIntAndFloat(val1=1, val2=2.5)
+            t_int_and_float.result = 3.5
+
+            #####################
+            # Test if-statement #
+            #####################
+
+            if 1 + 1 == 2:
+               val = 55
```
sanderr, line 103: This variable shadows the one in the outer scope (unlike in Python). Unintuitive, I know.
```
x = 1
if true:
    x = 2
end
```

sanderr (changes requested):
Nice. Looks good to me overall.

tests/compiler/test_arithemetic.py
```diff
@@ -0,0 +1,595 @@
...
+                float result
+            end
+            implement TestFloat using testFloat
+
+            implementation testFloat for TestFloat:
+                self.result = val1 ** val2
+            end
+
+            t_float = TestFloat(val1=0.2, val2=2.0)
+            std::assert(not (t_float.result > 0.4 and t_float.result < 0.41))
```
Hugo-Inmanta, line 527:
```suggestion
            std::assert( t_float.result >= 0.04 and t_float.result < 0.041)
```

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -1080,8 +1107,8 @@ def p_constants_collect(p: YaccProduction) -> None:
 
 
 def p_wrapped_kwargs(p: YaccProduction) -> None:
-    "wrapped_kwargs : '*' '*' operand"
-    p[0] = WrappedKwargs(p[3])
+    "wrapped_kwargs : DOUBLE_STAR operand"
+    p[0] = WrappedKwargs(p[2])
```
Hugo-Inmanta, line 1111: This is a (imo good) breaking change, i.e. we don't support unpacking stuff with spaces anymore e.g. `B(   *  *a)` should we bother mentioning this ?

Hugo-Inmanta (changes requested):

src/inmanta/ast/constraint/expression.py
```diff
@@ -510,6 +510,94 @@ def _bin_op(self, arg1: object, arg2: object) -> object:
...
+
+    def _bin_op(self, arg1: object, arg2: object) -> object:
+        if not isinstance(arg1, numbers.Number):
+            raise TypingException(self, f"The {self.get_name()} operator in not supported for {arg1} of type {type(arg1)}.")
+        if not isinstance(arg2, numbers.Number):
+            raise TypingException(self, f"The {self.get_name()} operator in not supported for {arg2} of type {type(arg2)}.")
+        return self._arithmetic_op(arg1, arg2)
+
+    @abstractmethod
+    def _arithmetic_op(self, arg1: object, arg2: object) -> object:
```
sanderr, line 524:
```suggestion
    def _arithmetic_op(self, arg1: numbers.Number, arg2: numbers.Number) -> object:
```
Right?

src/inmanta/ast/constraint/expression.py
```diff
@@ -510,6 +510,94 @@ def _bin_op(self, arg1: object, arg2: object) -> object:
...
+
+class Modulo(ArithmeticOperator):
+    __slots__ = ()
+    __op = "%"
+
+    def __init__(self, op1: ExpressionStatement, op2: ExpressionStatement) -> None:
+        ArithmeticOperator.__init__(self, "modulo", op1, op2)
+
+    def _arithmetic_op(self, arg1: object, arg2: object) -> object:
+        return arg1 % arg2
```
sanderr, line 584: I was going to remark that this doesn't work for floats but apparently it does (with limited precision). I'd never considered it.

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -56,13 +55,56 @@
...
+    "FSTRING",
+    "PLUS_OP",
+    "MINUS_OP",
+    "DIVISION_OP",
+    "DOUBLE_STAR",
+] + sorted(list(reserved.values()))
+
+
+def t_REGEX(t: lex.LexToken) -> lex.LexToken:  # noqa: N802
+    r"matching[\s]+/([^/\\\n]|\\.)+/"
```
sanderr, line 85: Could you add a comment as to why we include the matching here?

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -56,13 +55,56 @@
...
+    value = Reference("self")  # anonymous value
+    try:
+        expr = Regex(value, regex_as_str)
+        t.value = expr
+        return t
+    except RegexError as error:
+        t.value = regex_with_slashes_as_str
+        end = t.lexer.lexpos - t.lexer.linestart + 1
+        (s, e) = t.lexer.lexmatch.span()
+        start = end - (e - s) + index_first_slash_char
```
sanderr, line 98: I'm not sure this is accurate if there's a newline betweeen `matching` and the start of the regex. Same for the lineno below.

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -56,13 +55,56 @@
...
+
+        r: Range = Range(t.lexer.inmfile, t.lexer.lineno, start, t.lexer.lineno, end)
+        raise ParserException(r, t.value, f"Regex error in {t.value}: '{error}'")
+
+
+def t_MATCHING(t: lex.LexToken) -> lex.LexToken:  # noqa: N802
```
sanderr, line 104: Same note about the whitespace. Though I have to admit it's more difficult here. Something like the following is allowed
```
typedef x as string matching(
    self == "hello"
)
```

But of course something like `typedef x as string matchingself == "hello"` isn't because then `matchintself` is a single token.

I'm a bit too far removed from the lexer at the moment to know if this is easy to address, or whether we should consider just accepting the breaking change on the edge case.

src/inmanta/parser/plyInmantaParser.py
sanderr: Good point. Another change that I think is intrinsically good, yet technically breaking.

sanderr (changes requested):

src/inmanta/parser/plyInmantaLex.py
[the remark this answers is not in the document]
arnaudsjs: This is a leftover from a previous implementation. I will remove it.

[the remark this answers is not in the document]
arnaudsjs: It should be `*` right, instead of `?`. Because more than one whitespace is also allowed.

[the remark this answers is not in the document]
bartv: This is newline inside the regex?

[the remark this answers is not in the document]
wouterdb: Removing the \n may cause line numbering to go wrong, to  be tested !

src/inmanta/parser/plyInmantaLex.py
bartv: not too concerned. Might make sense to ask @edvgui if we generate something like this

src/inmanta/parser/plyInmantaLex.py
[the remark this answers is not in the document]
arnaudsjs: > This is newline inside the regex?

yes. The previous implementation allowed a newline in a regex. The new one doesn't.

src/inmanta/parser/plyInmantaLex.py
edvgui: Everything that we generate using `inmanta-module-factory` (everything that I am aware off) uses this template for typedef expressions, so there is always a space after `matching`: https://code.inmanta.com/solutions/tools/inmanta-module-factory/-/blob/2ebb08e917bab274cb7276a76305f2929f37960b/src/inmanta_module_factory/inmanta/typedef.py#L74
```
stmt = f"typedef {self.name} as {base_type_reference} matching {self.constraint}"
```

src/inmanta/ast/constraint/expression.py
```diff
@@ -521,7 +521,7 @@ def _bin_op(self, arg1: object, arg2: object) -> object:
         return self._arithmetic_op(arg1, arg2)
 
     @abstractmethod
-    def _arithmetic_op(self, arg1: object, arg2: object) -> object:
+    def _arithmetic_op(self, arg1: numbers.Number, arg2: numbers.Number) -> object:
```
Hugo-Inmanta, line 524: Why not using this same signature for the Subclasses as well ?

tests/compiler/test_arithemetic.py
```diff
@@ -524,7 +524,7 @@ def test_exponentiation(snippetcompiler) -> None:
             end
 
             t_float = TestFloat(val1=0.2, val2=2.0)
-            std::assert(not (t_float.result > 0.4 and t_float.result < 0.41))
+            std::assert( t_float.result >= 0.04 and t_float.result < 0.041)
```
Hugo-Inmanta, line 527: Just noticed a typo in the filename `tests/compiler/test_arithemetic.py` -> `tests/compiler/test_arithmetic.py`

src/inmanta/ast/constraint/expression.py
arnaudsjs: I changed all types to `int|float`, because mypy was complaining. The type `number.Number` is too wide.

src/inmanta/parser/plyInmantaLex.py
wouterdb: I think the lexer is set to ignore whitespace between tokens, so it is a moot point?

wouterdb (approved):

src/inmanta/parser/plyInmantaLex.py
sanderr: I don't think so because we now make `matching /regex/` a single token, which is the change I'm talking about: the lexer used to ignore whitespace, we require one.

Hugo-Inmanta (approved):

sanderr (changes requested):
- I'm not sure if the breaking changes discussion has fully settled yet but I'll leave that to you
- I believe [this thread](https://github.com/inmanta/inmanta-core/pull/6957#discussion_r1455859548) is still relevant.

arnaudsjs:
> I'm not sure if the breaking changes discussion has fully settled yet but I'll leave that to you

Which part didn't settle yet?

src/inmanta/parser/plyInmantaLex.py
arnaudsjs: The lexer doesn't allow a newline between the matching and the start of the regex, right?

arnaudsjs: I made sure the line numbers are reported correctly when there is a newline between the matching keyword and the regex.

sanderr:
> Which part didn't settle yet?

I don't know, that was the "I'm not sure" part in my sentence.

Issue #6560, opened 2023-10-02
Add fact that don't expire


wouterdb:
Add the ability to make fact that don't expire

This should be added to

- to the Agent-handler push API (optional attribute)
- to the Agent-handler pull API (optional attribute)
- the rest API (optional attribute)
- fact expiry logic

PR #6995, opened 2024-01-17
Issue/6560 add facts that dont expire

Hugo-Inmanta:
# Description

Add facts that don't expire

closes https://github.com/inmanta/inmanta-core/issues/6560

# Self Check:

- [x] Changelog entry


changelogs/unreleased/6560-add-facts-that-dont-expire.yml
```diff
@@ -0,0 +1,7 @@
+description: Add facts that don't expire.
+issue-nr: 6560
+change-type: minor
+destination-branches: [master, iso7, iso6]
```
wouterdb, line 4: I don't think this is for iso6

tests/db/migration_tests/test_v202311170_to_v202401160.py
```diff
@@ -0,0 +1,31 @@
...
+import pytest
+
+file_name_regex = re.compile("test_v([0-9]{9})_to_v[0-9]{9}")
+part = file_name_regex.match(__name__)[1]
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), f"dumps/v{part}.sql"))
+async def test_add_non_expiring_facts(migrate_db_from: abc.Callable[[], abc.Awaitable[None]]) -> None:
+    # This migration script adds a column. Just verify that the script doesn't fail.
+    await migrate_db_from()
```
wouterdb, line 31: Also assert that all facts returned do expire

wouterdb (changes requested):
We also need documentation

(could you go over the self-check-list?)

tests/db/migration_tests/test_v202311170_to_v202401160.py
Hugo-Inmanta: I'm not sure how to test for this, should I populate the `parameter` table in `test_dump_db` and check that the default is correctly set ?

docs/model_developers/handlers.rst
```diff
@@ -95,6 +95,17 @@ Each Handler should override 4 methods of the CRUDHandler:
 The context (See :class:`~inmanta.agent.handler.HandlerContext`) passed to most methods is used to
 report results, changes and logs to the handler and the server.
 
+The :meth:`~inmanta.agent.handler.HandlerContext.set_fact` method can be used to report facts to the server.
```
wouterdb, line 98: I would also link to std::get_fact

and link from the glossary to here.

And I would explain what facts are in a few words.

docs/model_developers/handlers.rst
```diff
@@ -95,6 +95,17 @@ Each Handler should override 4 methods of the CRUDHandler:
...
+By default, facts are considered stale after not being refreshed or updated for a certain time. This time span is
```
wouterdb, line 99: I would first explain here there there are two models (fact push and fact pull), and add some example code on both. 

I would also explain: if you have a fact that never changes, do it like so. If it may change, do it like so. 

and explain it should not be used for things that change rapidly (e.g. cpu usage), as they are not intended to refresh very quickly.

tests/agent_server/conftest.py
```diff
@@ -413,6 +421,28 @@ def facts(self, ctx: HandlerContext, resource: Resource) -> dict:
...
+
+        def delete_resource(self, ctx: HandlerContext, resource: SetNonExpiringFactResource) -> None:
+            pass
+
+        def update_resource(self, ctx: HandlerContext, changes: dict, resource: SetNonExpiringFactResource) -> None:
+            pass
+
+        def facts(self, ctx: HandlerContext, resource: Resource) -> dict:
+            self._do_set_fact(ctx, resource)
+            return {fact["id"]: fact for fact in ctx.facts}
```
wouterdb, line 440:
```suggestion
            return {}
```

tests/agent_server/test_server_agent.py
```diff
@@ -3672,6 +3672,109 @@ async def _wait_until_facts_are_available():
...
+    assert result.code == 503
+    result = await client.get_param(tid=environment, id="expiring", resource_id="test::SetNonExpiringFact[agent1,key=expiring]")
+    assert result.code == 503
+
+    async def _wait_until_facts_are_available():
+        params = await data.Parameter.get_list()
+        return len(params) == 2
+
+    await retry_limited(_wait_until_facts_are_available, 10)
+    await asyncio.sleep(0.5)
```
wouterdb, line 3761: why? (add a comment)

tests/db/migration_tests/test_v202311170_to_v202401160.py
wouterdb: yes, and run that setup on the previous version, as described in the adr in the docs folder (iirc)

wouterdb (changes requested):

docs/model_developers/handlers.rst
```diff
@@ -95,6 +95,62 @@ Each Handler should override 4 methods of the CRUDHandler:
...
+                ctx.set_fact(fact_id=key, value=value, expires=True)
+
+By default, facts expire when they are not being refreshed or updated for a certain time, controlled by the
+:inmanta.config:option:`server.fact-expire` config option. Querying for an expired fact will force the
+agent to refresh it first.
+Expired facts are also refreshed periodically. This time interval is controlled by the
+:inmanta.config:option:`server.fact-renew` config option.
+
+
+When reporting a fact, setting the ``expires`` parameter to ``False`` will ensure that this fact never expires. This
```
wouterdb, line 144: I would also explain the `get_fact` method and explain there are two modes (push and pull) and that if the fact expires, it must also be returned by `get_fact`

tests/db/migration_tests/test_v202311170_to_v202401160.py
```diff
@@ -0,0 +1,51 @@
...
+        - Update "pre" dump v202311170.sql: add a fact to the parameter table using "old" codebase (master)
+        - Ensure the migration correctly populates the newly added ``expires`` column with default ``True`` value.
+    """
+
+    columns_before_migration = await get_columns_in_db_table("parameter")
+    assert "expires" not in columns_before_migration
+
+    await migrate_db_from()
+
+    columns_after_migration = await get_columns_in_db_table("parameter")
```
wouterdb, line 46: If possible, I would prefer to the the API instead of the DB directly (as this test will break when the DB schema is evolved further).

wouterdb (changes requested):
Starting to look nice

docs/model_developers/handlers.rst
```diff
@@ -134,18 +139,44 @@ method. e.g.:
...
-
 
 When reporting a fact, setting the ``expires`` parameter to ``False`` will ensure that this fact never expires. This
 is useful to take some load off the agent when working with facts whose values never change. On the other hand, when
 working with facts whose values are subject to change, setting the ``expires`` parameter to ``True`` will ensure
 they are periodically refreshed.
 
+---------
+
+Facts are automatically pulled periodically (this time interval is controlled by the
```
wouterdb, line 154: They are only pulled if they are about to expire or if a model requested them and they where not present.

docs/model_developers/handlers.rst
```diff
@@ -134,18 +139,44 @@ method. e.g.:
...
+    :linenos:
+
+    @provider("openstack::FloatingIP", name="openstack")
+    class FloatingIPHandler(OpenStackHandler):
+        ...
+
+        def facts(self, ctx, resource) -> dict[str, object]:
+            port_id = self.get_port_id(resource.port)
+            fip = self._neutron.list_floatingips(port_id=port_id)["floatingips"]
+            if len(fip) == 0:
```
wouterdb, line 169: I would prefer switching to a style where we also use `ctx.set_fact` here, it is cleaner.

tests/db/migration_tests/test_v202311170_to_v202401160.py
```diff
@@ -22,30 +22,30 @@
...
-    assert "expires" in columns_after_migration
-
-    expires = await postgresql_client.fetchval("SELECT expires FROM public.parameter where name='test_default_expires';")
-
-    assert expires is True
+    client = Client("client")
+    result = await client.list_params(tid="5fe05a29-1ffd-4073-84ca-ce523fc30aa0")
+    assert result.code == 200
+    for param in result.result["parameters"]:
+        if param["id"] == "36df4879-e609-4456-afcc-04cf2fe9b518":
```
wouterdb, line 50: is it not simply all of them that have `expires == true` ?

wouterdb (approved):
I have some minor comments, but I don't think I need to review again

tests/db/migration_tests/test_v202311170_to_v202401160.py
Hugo-Inmanta: Currently, yes. The rationale was if we ever add one that doesn't expire in the future it would break this test if we only check all(expires == true)

PR #6997, opened 2024-01-17
issue/5558-Logging-improvement-for-nullable-list

### LINKED ISSUE #5558: Logging improvement for nullable list
Currently when we loop over a nullable list attribute which is not set, the current message is the following : ` A for loop can only be applied to lists and relations`.  It would be nice to have an additional hint stating that the list is actually null.
An example to reproduce the issue: 
```
entity A:
    string[]? elements=null
end
implement A using std::none

a = A()
for element in a.elements:
    std::print(element)
end
```

FloLey:
# Description

Currently when we loop over a nullable list attribute which is not set, the current message is the following : "A for loop can only be applied to lists and relations". Now there will also be a hint telling that the attribute is null

closes #5558 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -253,8 +253,8 @@ class For(RequiresEmitStatement):
 
         if not isinstance(var, (list, Unknown)):
             msg = "A for loop can only be applied to lists and relations."
-            if isinstance(var, NoneValue):
-                msg += " Hint: '%s' is null." % self.base
+            if not isinstance(self.base, Literal):
+                msg += " Hint: '%s' is '%s'." % (self.base, str(var))
             raise TypingException(self, msg)
 
         # we're done here: base's execute has reported results to helper
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -74,7 +74,7 @@ from inmanta.execute.runtime import (
     WrappedValueVariable,
 )
 from inmanta.execute.tracking import ImplementsTracker
-from inmanta.execute.util import NoneValue, Unknown
+from inmanta.execute.util import Unknown
 
 try:
     from typing import TYPE_CHECKING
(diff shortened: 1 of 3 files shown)


src/inmanta/ast/statements/generator.py
```diff
@@ -252,7 +252,10 @@ def execute(self, requires: dict[object, object], resolver: Resolver, queue: Que
         var = self.base.execute(requires, resolver, queue)
 
         if not isinstance(var, (list, Unknown)):
-            raise TypingException(self, "A for loop can only be applied to lists and relations")
+            msg = "A for loop can only be applied to lists and relations."
+            if isinstance(var, NoneValue):
```
sanderr, line 256: I think it might be better to address this generically. i.e. no tailor-made message when it's `null`, just always tell the user what the value is.

sanderr (changes requested):

tests/compiler/test_loops.py
```diff
@@ -57,7 +59,25 @@ def test_for_error_2(snippetcompiler):
         for i in "foo":
         end
     """,
-        "A for loop can only be applied to lists and relations (reported in For(i) ({dir}/main.cf:2))",
+        "A for loop can only be applied to lists and relations. (reported in For(i) ({dir}/main.cf:2))",
```
FloLey, line 62: I added the 
`if not isinstance(self.base, Literal):` for the hint because I found it weird here to see : Hint: "foo" is "foo".

sanderr: I like it. One suggestion for further improvement:
- Instead of `not isinstance(self.base, Literal)`, go for `isinstance(self.base, Reference)`. Otherwise I think you'd get the same weird behavior for a literal dict, which is of type `CreateDict` iirc. I think it makes sense to restrict ourselves to variable/attribute references here to keep it concise.
- Even if it is a literal I'd mention the value. Not twice as in your example above, but just once.

Some examples of how I think the output could then look:
- `Hint: 'a.b.c' resolves to 'null'`
- `Hint: `'"foo"' is not a list`
- `Hint: `'{"test": 1}' is not a list`

Just suggestions, wdyt?

FloLey: l like it :)

FloLey: @sanderr I can't rerequest review but it is ready

sanderr (approved):
Nice!

PR #6998, opened 2024-01-17
Make agent more resilient to resource loading issues

wouterdb:
# Description

Ensure exception produced by resource construction are properly handled 

Previously, exceptions could escape, making deployments hang

```
  File "/opt/inmanta/lib64/python3.11/site-packages/inmanta/agent/agent.py", line 737, in deploy_action
    await self.get_latest_version_for_agent(
  File "/opt/inmanta/lib64/python3.11/site-packages/inmanta/agent/agent.py", line 860, in get_latest_version_for_agent
    undeployable, resources = await self.load_resources(
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/inmanta/lib64/python3.11/site-packages/inmanta/agent/agent.py", line 1053, in load_resources
    resource: Resource = Resource.deserialize(res["attributes"])
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/inmanta/lib64/python3.11/site-packages/inmanta/resources.py", line 353, in deserialize
    obj.populate(obj_map, force_fields)
  File "/opt/inmanta/lib64/python3.11/site-packages/inmanta/resources.py", line 386, in populate
    raise Exception("Resource with id {} does not have field {}".format(fields["id"], field))
```

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -114,7 +114,7 @@ class ResourceActionBase(abc.ABC):
 
     def cancel(self) -> None:
         if not self.is_running() and not self.is_done():
-            LOGGER.info("Cancelled deploy of %s %s", self.gid, self.resource_id)
+            self.logger.info("Cancelled deploy of %s %s %s", self.__class__.__name__, self.gid, self.resource_id)
             self.future.set_result(ResourceActionResult(cancel=True))
 
     def long_string(self) -> str:
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -1062,7 +1062,7 @@ class AgentInstance:
                     resource = Resource.deserialize(res["attributes"], use_generic=True)
                     loaded_resources.append(resource)
 
-            except TypeError:
+            except Exception:
                 failed_resources.append(res["id"])
                 undeployable[res["id"]] = const.ResourceState.unavailable
                 resource = Resource.deserialize(res["attributes"], use_generic=True)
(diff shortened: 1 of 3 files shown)


src/inmanta/protocol/endpoints.py
```diff
@@ -208,7 +208,7 @@ async def start(self) -> None:
         Connect to the server and use a heartbeat and long-poll for two-way communication
         """
         assert self._env_id is not None
-        LOGGER.log(3, "Starting agent for %s", str(self.sessionid))
+        LOGGER.info("Starting agent for %s", str(self.sessionid))
```
wouterdb, line 211: This is to make it clear in the log when the agent restarts

src/inmanta/agent/agent.py
```diff
@@ -114,7 +114,7 @@ def is_done(self) -> bool:
 
     def cancel(self) -> None:
         if not self.is_running() and not self.is_done():
-            LOGGER.info("Cancelled deploy of %s %s", self.gid, self.resource_id)
+            self.logger.info("Cancelled deploy of %s %s %s", self.__class__.__name__, self.gid, self.resource_id)
```
arnaudsjs, line 117: Just for my information. Why is it important to rely on the specific logger from the agent?

arnaudsjs (approved):

src/inmanta/agent/agent.py
wouterdb: currently, we see the cancel being logged, but we don't know which agent instance produced it. 
So to analyze the agent getting stuck, it is near impossible to say if it goes into the cancel code path at all. 

(cross-agent-dependencies are also getting cancelled here, meaning that the agent name in the resource does not necessarily correspond to the agent it is on, so that means you have to make  assumptions, which is not very useful when diagnosing a seemingly impossible scenario, where you know some of your assumptions don't hold).

FloLey (approved):

PR #7025, opened 2024-01-23
Issue/document podman deployment

edvgui:
# Description

Document how to deploy the inmanta orchestrator with podman and systemd (rootless).

# Self Check:

- [ ] Changelog entry


docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,328 @@
...
+        ExecStart=/usr/bin/podman run \
+                --cidfile=%t/%n.ctr-id \
+                --cgroups=no-conmon \
+                --sdnotify=conmon \
+                -d \
+                --replace \
+                --publish=127.0.0.1:8888:8888 \
+                --uidmap=993:@%U \
+                --uidmap=0:1:993 \
+                --gidmap=993:@%G \
```
arnaudsjs, line 143: The `@%U` and `@%G` syntax is not supported on the podman version available on Rocky Linux 8 and 9 as discussed.

arnaudsjs (changes requested):

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,328 @@
...
+    .. code-block:: sh
+
+        systemctl --user daemon-reload
+        systemctl --user enable inmanta-orchestrator-db.service
+        systemctl --user start inmanta-orchestrator-db.service
+
+6.  In the unit file of the orchestrator (as described :ref:`here<setup-systemd-unit>`), make sure to attach the orchestrator
+    container to the network the database is a part of, using the ``--network`` option of the ``podman run`` command.
+
+7.  Don't forget to update the ip address of the database in the inmanta server configuration file (``~/.config/inmanta/inmanta.cfg``)!
```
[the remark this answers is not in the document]
edvgui: Well, it will not be running yet, as it didn't have a database (I assumed that this setup would be done before the orchestrator setup)

docs/install/2-install-server-with-podman-and-systemd.rst
edvgui: Fixed in https://github.com/inmanta/inmanta-core/pull/7025/commits/d4acb57193b6d0a51866e7ca6defe92276bfbaa1

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,328 @@
...
+
+    Use ``podman pull`` to get the desired image:
+
+    .. code-block:: sh
+        :substitutions:
+
+        podman pull containers.inmanta.com/containers/service-orchestrator:|version_major|
+
+
+    This command will pull the latest version of the Inmanta Service Orchestrator image.
```
arnaudsjs, line 62: For the ISO docs build, the tag is a major version number. So this statement is only true within the major version this documentation build is for.

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,328 @@
...
+            mkdir -p ~/.config/inmanta
+            podman run --rm -ti containers.inmanta.com/containers/service-orchestrator:|version_major| cat /etc/inmanta/inmanta.cfg > ~/.config/inmanta/inmanta.cfg
+
+2.  Update database settings:
+    It is very unlikely that your database setup will match the one described in the default config we just got.  Update the configuration in the ``[database]`` section
+    to reflect the setup you have.
+
+.. only:: iso
+
+    3.  Get the license files
```
arnaudsjs, line 95:
```suggestion
    3.  Get the license files:
```

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,328 @@
...
+        Together with the access to the inmanta container repo, you should also have received a license and an entitlement files.  The orchestrator will need them
```
arnaudsjs, line 96:
```suggestion
        Together with the access to the inmanta container repo, you should also have received a license and an entitlement file. The orchestrator will need them
```

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,328 @@
...
+                --gidmap=993:0:1 \
+                --gidmap=0:1:993 \
+                --name=inmanta-orchestrator-server \
+                --volume=%E/inmanta/inmanta.cfg:/etc/inmanta/inmanta.cfg:z \
+                --volume=%E/inmanta/license/com.inmanta.license:/etc/inmanta/license/com.inmanta.license:z \
+                --volume=%E/inmanta/license/com.inmanta.jwe:/etc/inmanta/license/com.inmanta.jwe:z \
+                --entrypoint=/usr/bin/inmanta \
+                --user=993:993 \
+                containers.inmanta.com/containers/service-orchestrator:|version_major| \
+                -vvv --timed-logs server
```
arnaudsjs, line 198: Like this the server logs will be sent to stdout and not to a file. Is this something we want? The RPM executes the following command to send the logs file: https://github.com/inmanta/inmanta-core/blob/ae61aa04b4b53447585a8ef055f5bb874dc6436e/misc/inmanta-server.service#L10

The same is true for the OSS service file of course.

edvgui: Yes, but that only works if the user didn't change the log folder configuration.  Or rather, it would always work, but be confusing if this log folder is not meant to be used.
I am not sure what is the best, I am open to both.

docs/install/2-install-server-with-podman-and-systemd.rst
edvgui: Is it better now?

docs/install/2-install-server-with-podman-and-systemd.rst
arnaudsjs: As discussed we will go for the approach were:

* We set ` --log-file /var/log/inmanta/server.log`. That way the support archive also captures the server logs.
* And, we map the `/var/log/inmanta` directory by default to the hosts to make the logs accessible for the server in case the server fails to start.

arnaudsjs (changes requested):

arnaudsjs (approved):

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,341 @@
+.. _install-server-with-podman:
+
+Install Inmanta with Podman and Systemd
+***************************************
+
+This page explains how to setup an Inmanta orchestration server using Podman and Systemd.
+This guide assumes you already have `Podman <http://podman.io/>`_ installed on your machine and that you are running a Linux distribution with Systemd.
```
bartv, line 7:
```suggestion
This guide assumes you already have `Podman <http://podman.io/>`_ installed on your machine and that you are running a Linux distribution using Systemd.
```

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,341 @@
...
+
+.. note::
+    The full setup should be doable without any root privilege (rootless) on the host, running the orchestrator with an unprivileged user.
```
bartv, line 10:
```suggestion
    The full setup can be done root privileges (rootless) on the host and running the orchestrator with an unprivileged user.  
```

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,341 @@
...
+
+
+    This command will pull the latest release of the Inmanta Service Orchestrator image within this major version.
+
+
+Prepare the orchestrator configuration
+######################################
+
+1.  Get the default configuration file:
+    As of now, the container can not be configured with environment variables, we should use a configuration file, mounted inside of the container.
```
bartv, line 69:
```suggestion
    As of now, the container cannot be configured with environment variables, we should use a configuration file, mounted inside the container.
```

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,341 @@
...
+3.  Make sure that there is a folder on your host that can persist all the logs of the server and that it is owned by the user running the orchestrator service.  
+    In this setup, the folder on the host will be ``~/.local/share/inmanta-orchestrator-server/logs``.
+
+    .. code-block:: sh
+
+        mkdir -p ~/.local/share/inmanta-orchestrator-server/logs
+
+    .. warning:: 
+        Inside of the container, this folder will be mounted at ``/var/log/inmanta`` as it is the default location where the orchestrator saves its logs.  This
+        location is configurable in the orchestrator configuration file.  If you were, for any reason, to change this location, make sure to update any usage
```
bartv, line 102:
```suggestion
        location is configurable in the orchestrator configuration file.  If you for any reason would change this location in the mount, make sure to update any usage
```

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,341 @@
...
+        Restart=on-failure
+        TimeoutStopSec=70
+        ExecStart=/usr/bin/podman run \
+                --cidfile=%t/%n.ctr-id \
+                --cgroups=no-conmon \
+                --sdnotify=conmon \
+                -d \
+                --replace \
+                --publish=127.0.0.1:8888:8888 \
+                --uidmap=993:0:1 \
```
bartv, line 153: I would like a note that explains what these numbers are. Something like:

993 is the uid and gid of the inmanta user and group inside the container. This option instructs podman to map this user to our local non-root user.

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,341 @@
...
+    .. code-block:: 
+
+        [Unit]
+        Description=Podman 
+        Documentation=https://docs.inmanta.com
+        Wants=network-online.target
+        After=network-online.target
+        RequiresMountsFor=%t/containers
+
+        [Service]
```
bartv, line 142: I would also like to have the option where the user that is setting this up has root permissions (either working as root or working as the system user using sudo). This would require a user and group setting here as well.

docs/install/2-install-server-with-podman-and-systemd.rst
edvgui: I think a word is missing there, I am not sure of the point you are making

bartv: Like this. What I changed was the your version is very conditional: "should be doable" sounds like: we have not tried it but we so no reason why it would not work

docs/install/2-install-server-with-podman-and-systemd.rst
edvgui: I added a tab-set with rootful and rootless versions, I think it is fairly complete

docs/install/2-install-server-with-podman-and-systemd.rst
edvgui: I added an explanation, is it clear enough or maybe too detailed?

docs/install/2-install-server-with-podman-and-systemd.rst
edvgui: I formulated it completely differently, can you check again?

docs/install/2-install-server-with-podman-and-systemd.rst
```diff
@@ -0,0 +1,585 @@
...
+        .. only:: oss
+
+            .. code-block:: 
+
+                [Unit]
+                Description=Podman 
+                Documentation=https://docs.inmanta.com
+                Wants=network-online.target
+                After=network-online.target
+                RequiresMountsFor=%t/containers
```
edvgui, line 332: @bartv this will not work as the user of the unit is root and not inmanta.  I am not sure how to work around this, should we have two unit files? The root one starting the inmanta one? And the inmanta one starting the container?

edvgui: Same issue for `--cidfile=%t/%n.ctr-id`

edvgui: My question is rather: what is the sys-admin way of doing this?

bartv (approved):

Issue #6959, opened 2024-01-11
update project release documentation


sanderr:
The [Project Release for Production](https://docs.inmanta.com/inmanta-service-orchestrator/7/administrators/operational_procedures.html) section in the docs is outdated and does not apply to modules v2. It should be updated.

# Task
- Follow the steps outlined in the docs + export to orchestrator to verify that it installs the pinned versions rather than the latest one. Use both v1 and v2 modules.
- Report anything that doesn't work -> either we fix the tooling or we update the docs
- Fix any explanation that may be outdated, i.e. mentioning an old file, ...

PR #7027, opened 2024-01-23
Issue/6959 update project release documentation

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/6959

>Follow the steps outlined in the docs + export to orchestrator to verify that it installs the pinned versions rather than the latest 
> one. Use both v1 and v2 modules.

Checked against an iso7-dev orchestrator for both v1 and v2

# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/administrators/operational_procedures.rst
+++ b/docs/administrators/operational_procedures.rst
@@ -25,7 +25,7 @@ section for more information on how to set the project's pip configuration.
 Context
 ++++++++
 * The project development and testing is complete
-* All modules have been properly released.
+* All modules have been properly released (See `here <Releasing and distributing modules>`_ for the procedure).
 * The project is in a git repo, with a specific branch dedicated to production releases
 * The project is checked out on disk.
 * All modules are checked out on the correct, tagged commit.
--- a/docs/administrators/operational_procedures.rst
+++ b/docs/administrators/operational_procedures.rst
@@ -33,12 +33,15 @@ Context
 Procedure
 ++++++++++
 
-1. Verify in ``project.yml`` that ``install_mode`` is set to ``release``.
+1. Check modules install mode:
+    - For v1 modules (if any): ensure that ``install_mode`` in ``project.yml`` is set to ``release``.
+    - For v2 modules (if any): ensure that ``pip.pre`` is not set.
+
 2. Freeze all modules with:
 
 .. code-block:: bash
 
-    inmanta -vv project freeze --recursive --operator "=="
+    inmanta -vv -X project freeze --recursive --operator "=="
 
 This will cause the ``project.yml`` file to be updated with constraints that only allow this project to work with
 this exact set of module versions. This ensures that no unwanted updates can 'leak' into the production environment.
--- a/docs/administrators/operational_procedures.rst
+++ b/docs/administrators/operational_procedures.rst
@@ -25,7 +25,7 @@ section for more information on how to set the project's pip configuration.
 Context
 ++++++++
 * The project development and testing is complete
-* All modules have been properly released (See `here <Releasing and distributing modules>`_ for the procedure).
+* All modules have been properly released (See :ref:`Releasing and distributing modules` for the procedure).
 * The project is in a git repo, with a specific branch dedicated to production releases
 * The project is checked out on disk.
 * All modules are checked out on the correct, tagged commit.
--- a/docs/administrators/operational_procedures.rst
+++ b/docs/administrators/operational_procedures.rst
@@ -35,7 +35,7 @@ Procedure
 
 1. Check modules install mode:
     - For v1 modules (if any): ensure that ``install_mode`` in ``project.yml`` is set to ``release``.
-    - For v2 modules (if any): ensure that ``pip.pre`` is not set.
+    - For v2 modules (if any): ensure that ``pip.pre`` is not set in ``project.yml``.
 
 2. Freeze all modules with:
 
--- a/docs/administrators/operational_procedures.rst
+++ b/docs/administrators/operational_procedures.rst
@@ -33,20 +33,19 @@ Context
 Procedure
 ++++++++++
 
-1. Verify in `project.yml` that `install_mode` is set to `release`.
+1. Verify in ``project.yml`` that ``install_mode`` is set to ``release``
 2. Freeze all modules with
 
 .. code-block:: bash
 
-    inmanta -vv -X project freeze --recursive --operator "=="
+    inmanta -vv project freeze --recursive --operator "=="
 
+This will cause the `project.yml` file to be updated with constraints that only allow this project to work with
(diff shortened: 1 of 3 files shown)


docs/administrators/operational_procedures.rst
```diff
@@ -33,23 +33,31 @@ Context
...
 
-1. Verify in `project.yml` that `install_mode` is set to `release`.
-2. Freeze all modules with
+1. Verify in ``project.yml`` that ``install_mode`` is set to ``release``.
+2. Freeze all modules with:
 
 .. code-block:: bash
 
-    inmanta -vv -X project freeze --recursive --operator "=="
+    inmanta -vv project freeze --recursive --operator "=="
```
sanderr, line 41: Considering how bad we sometimes are at error messages, perhaps we should keep the `-X`.

docs/administrators/operational_procedures.rst
```diff
@@ -33,23 +33,31 @@ Context
 Procedure
 ++++++++++
 
-1. Verify in `project.yml` that `install_mode` is set to `release`.
-2. Freeze all modules with
+1. Verify in ``project.yml`` that ``install_mode`` is set to ``release``.
```
sanderr, line 36: I would split this up: this part is relevant if you're using v1 modules. If you're using v2 you want to ensure that `pip.pre` is not set. Both if you're using both.

docs/administrators/operational_procedures.rst
sanderr, line 41: Have you verified how this behaves for v2 modules? Does it add the constraints to `project.yml`? Are they taken into account by `project install`?

docs/model_developers/release.rst
```diff
@@ -36,42 +36,59 @@ When done, first use git to add files:
...
-To commit, use the module tool. This will create a new dev release.
+Create a new dev release:
 
 .. code-block:: bash
 
-    inmanta module commit --patch -m "Fixed small bug"
+    inmanta module release --dev --patch -m "Fixed small bug"
 
-This command will set the version to the next dev version (`+0.0.1dev`) and add a timestamp.
+This command will set the version to the next dev version, e.g. ``+0.0.1dev`` here for a patch increment.
```
sanderr, line 45:
```suggestion
This command will set the version to the next dev version, e.g. ``+0.0.1dev`` for a patch increment.
```

docs/model_developers/release.rst
```diff
@@ -36,42 +36,59 @@ When done, first use git to add files:
 
     git add *
 
-To commit, use the module tool. This will create a new dev release.
+Create a new dev release:
```
sanderr, line 39: I think we can improve this further by not using the word "release" here. Perhaps "version"? Otherwise, further down, the line "Once the dev version is ready for release" becomes quite ambiguous.

docs/model_developers/release.rst
```diff
@@ -36,42 +36,59 @@ When done, first use git to add files:
...
 
 For the dev releases, no tags are created.
+Once the dev version is ready for release, perform a proper release by following
+the steps in the `Release Versions`_ section below.
 
 Release Versions
 ----------------
 
-To make an actual release (without `.dev` at the end):
+To perform an actual non-dev release:
```
sanderr, line 57:
```suggestion
To perform an actual stable release:
```

docs/model_developers/release.rst
```diff
@@ -36,42 +36,59 @@ When done, first use git to add files:
...
+
+First, merge your changes from your feature branch (e.g. ``mywork`` in the previous example) onto the branch dedicated to
+releases (e.g. master) and then do the actual release:
```
sanderr, line 60: I have to say I don't really see the relevance of this addition, and the code block below. We're not in the context of a pull request here, we're releasing a dev version that has presumably lived on master for a while already.

docs/model_developers/release.rst
```diff
@@ -36,42 +36,59 @@ When done, first use git to add files:
...
-This will remove the `.dev` version and automatically set the right tags on the module.
+    inmanta module freeze --recursive --operator "=="
 
-To automatically freeze all dependencies of this module to the currently checked out versions
+For v2 modules:
 
 .. code-block:: bash
 
-	inmanta module freeze --recursive --operator ==
+    pip freeze > out_file.txt
```
sanderr, line 84: Do I understand correctly that `inmanta module freeze` does not work for v2?

Either way, I don't think this is a helpful substitute because it will not be used by the orchestrator. I'm not entirely certain what the motivation is for freezing a module's dependencies. If it's just for repititive testing this is not an issue, but I would guess it's to freeze it on the orchestrator. Then again, with the project freeze available, I don't see the value. Perhaps this section is just no longer relevant?

@wouterdb could you shed some light on this?

sanderr (changes requested):

docs/administrators/operational_procedures.rst
Hugo-Inmanta: Seems to work fine, tested with the `inmanta-module-dummy`

Hugo-Inmanta: I missed the fact that the `-X` option was part of the top command, adding it back

docs/model_developers/release.rst
Hugo-Inmanta: > Do I understand correctly that inmanta module freeze does not work for v2?

Yes I think it's not intended to work for v2 :
https://github.com/inmanta/inmanta-core/blob/5c955d1278fe8d90440e598aa075526225adcce7/src/inmanta/moduletool.py#L664
and
https://github.com/inmanta/inmanta-core/blob/5c955d1278fe8d90440e598aa075526225adcce7/src/inmanta/moduletool.py#L1104

> Perhaps this section is just no longer relevant?

I think so, I'll update the relevant parts of `inmanta module freeze` for v2 modules depending on @wouterdb 's answer

Hugo-Inmanta: As discussed in standup:
- phase out `inmanta module freeze` in iso8: https://github.com/inmanta/inmanta-core/issues/7076

docs/administrators/operational_procedures.rst
```diff
@@ -25,20 +25,23 @@ section for more information on how to set the project's pip configuration.
...
 * The project is checked out on disk.
 * All modules are checked out on the correct, tagged commit.
 
 Procedure
 ++++++++++
 
-1. Verify in ``project.yml`` that ``install_mode`` is set to ``release``.
+1. Check modules install mode:
+    - For v1 modules (if any): ensure that ``install_mode`` in ``project.yml`` is set to ``release``.
+    - For v2 modules (if any): ensure that ``pip.pre`` is not set.
```
sanderr, line 38:
```suggestion
    - For v2 modules (if any): ensure that ``pip.pre`` is not set in ``project.yml``.
```

sanderr (approved):

PR #7044, opened 2024-01-25
Issue/fix broken link in docs

Hugo-Inmanta:
# Description

- Fix broken links 
- Fix parameter order breaking a test (newly added default params should be appended at the end)
- Closes https://github.com/inmanta/inmanta-core/issues/7045

# Self Check:

- [x] Changelog entry


src/inmanta/server/services/paramservice.py
```diff
@@ -129,12 +129,15 @@ async def get_param(self, env: data.Environment, param_id: str, resource_id: Opt
 
         param = params[0]
 
-        # check if it was expired
+        # check if it was expired:
+        #   - If it's not associated with a resource, this is a regular parameter that never expires
+        #   - Else, it's a fact: check if it can expire and if it has expired.
+
```
Hugo-Inmanta, line 135: comment to link to

Hugo-Inmanta:
I'm still checking that the links work as intended

src/inmanta/server/services/paramservice.py
```diff
@@ -181,9 +184,12 @@ async def _update_param(
         else:
             param = params[0]
             value_updated = param.value != value
-            await param.update(
-                source=source, value=value, updated=datetime.datetime.now().astimezone(), metadata=metadata, expires=expires
-            )
+            if expires is not None:
```
arnaudsjs, line 187: The if-block that belongs to this else-block contains a type violations. The constructor `data.Parameter` doesn't allow the `expires` field to be `None`. Did you run mypy on this? Do you see why we didn't catch this in a test case. Are we missing test coverage here?

arnaudsjs (changes requested):

src/inmanta/server/services/paramservice.py
Hugo-Inmanta: > Did you run mypy on this?
Yes, mypy seems ok with that, weirdly enough
I'm looking into this

Hugo-Inmanta: It seems the constructor is fine with the `expires` field being `None` e.g.:

```python
from uuid import UUID

from inmanta import data

param = data.Parameter(
    environment=UUID("0d54b1ac-6ddf-4e18-8c8f-9ac30d4d58ac"),
    name="name",
    source="source",
    # resource_id=resource_id,
    # value=value,
    # updated=datetime.datetime.now().astimezone(),
    # metadata=metadata,
    expires=None,
)

assert param.expires is None
```

arnaudsjs: Strange. But it's still a typing violation. Now I also see that the data migration script is also missing a `NOT NULL` constraint. That needs to be fixed as well. If you would not do that in this PR, you should create a ticket for this.

Hugo-Inmanta: Indeed. Also I think the migration is incomplete. Parameters (not associated with resource id) should get a default value expires=False and Facts (associated with a resource id) should get a default value of True ?

Hugo-Inmanta: Created a ticket here: https://github.com/inmanta/inmanta-core/issues/7045

Hugo-Inmanta: I think this is another type violation: https://github.com/inmanta/inmanta-core/blob/636f3f12569fb2be6296c831eb3318a00bbe5b02/src/inmanta/server/services/paramservice.py#L250
resource_id is None for facts and _update_param should allow None for this parameter.

I'll try to clean this up by adding a proper set_fact endpoint

Hugo-Inmanta:
I didn't manage to check that the links are actually fixed. If that's ok I'll wait for this to be merged and then check that the links work once the docs build has run ?
(I tried locally with the `irt build-docs  --in-place` but it didn't work since it's linking to a module)

tests/db/migration_tests/test_v202401160_to_v202401260.py
```diff
@@ -0,0 +1,70 @@
...
+        "dc7b5b59-33f4-417b-9afc-3c91b24b4008": False,  # False
+        "379e582a-007e-46eb-ac76-479c23da3d14": True,  # None
+        # Parameters:
+        "7428c4f7-d03d-47f9-8cf0-bce6688b61b3": False,  # False
+        "bb53a43c-c74f-43a9-999d-dc04c54324a1": False,  # None
+    }
+
+    client = Client("client")
+    result = await client.list_params(tid="630a163d-d010-4f70-b158-3434f7477aa9")
+    assert result.code == 200
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think adding this would break this test if we add more facts or parameters later on since the `migrate_db_from` updates to the latest migration available at the time it is called.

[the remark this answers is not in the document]
arnaudsjs: Yes, it always updates to the latest schema version. But the version of the schema doesn't impact the amount of records present in the database. The test case always starts from the same dump, so the content of the database should remain the same.

tests/db/migration_tests/test_v202401160_to_v202401260.py
```diff
@@ -0,0 +1,70 @@
...
+    This migration script adds the non-null constraint to the ``expires`` column of the parameter table.
+
+    The v202401160.sql dump is initialized with 2 parameters and 2 facts:
+
+    COPY public.parameter (id, name, value, environment, resource_id, source, updated, metadata, expires) FROM stdin;
+    dc7b5b59-33f4-417b-9afc-3c91b24b4008	fact1	value1	630a163d-d010-4f70-b158-3434f7477aa9	std::File[localhost,path=/tmp/test1]	fact	\N	\N	f       # NOQA E501
+    379e582a-007e-46eb-ac76-479c23da3d14	fact2	value2	630a163d-d010-4f70-b158-3434f7477aa9	std::File[localhost,path=/tmp/test2]	fact	\N	\N	\N      # NOQA E501
+    7428c4f7-d03d-47f9-8cf0-bce6688b61b3	parameter1	value1	630a163d-d010-4f70-b158-3434f7477aa9		fact	\N	\N	f
+    bb53a43c-c74f-43a9-999d-dc04c54324a1	parameter2	value2	630a163d-d010-4f70-b158-3434f7477aa9		fact	\N	\N	\N
+    \.
```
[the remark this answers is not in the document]
Hugo-Inmanta: The combination `(resource_id IS NULL, expires=True)` is an invalid state since parameters never expire. Should we detect it and raise a warning somehow ?

tests/db/migration_tests/test_v202401160_to_v202401260.py
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    client = Client("client")
+    result = await client.list_params(tid="630a163d-d010-4f70-b158-3434f7477aa9")
+    assert result.code == 200
+    for param in result.result["parameters"]:
+        try:
+            assert expected_expire_values[param["id"]] == param["expires"], param["name"]
+        except KeyError:
+            pass
```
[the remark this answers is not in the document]
Hugo-Inmanta: Not actually needed since:
> The test case always starts from the same dump, so the content of the database should remain the same.

tests/db/migration_tests/test_v202401160_to_v202401260.py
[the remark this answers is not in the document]
arnaudsjs: This is indeed an invalid state. But with the previous (broken) implementation it was possible to add records to the database like this. This PR should make sure that we cleanup all in the invalid data that sneaked into the database. That means we should also be able to cleanup this specific combination and the test suite should verify that the cleanup works correctly.

src/inmanta/server/services/paramservice.py
Hugo-Inmanta: Ticket to add the endpoints here: https://github.com/inmanta/inmanta-core/issues/7068

tests/db/migration_tests/test_v202401160_to_v202401260.py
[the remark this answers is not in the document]
Hugo-Inmanta: I see :+1:

src/inmanta/server/services/paramservice.py
```diff
@@ -200,54 +217,69 @@ async def _update_param(
...
+
+            result = await self._update_param(env, name, value, source, resource_id, metadata, expires=expires)
             if result:
                 recompile = True
                 params.append((name, resource_id))
 
+        if updating_facts:
+            parameters_and_or_facts = "facts"
+        if updating_parameters and updating_facts:
+            parameters_and_or_facts = "parameters and facts"
```
arnaudsjs, line 279: Aren't we missing the third option here where only parameters were updated?

src/inmanta/server/services/paramservice.py
```diff
@@ -200,54 +217,69 @@ async def _update_param(
...
         env: data.Environment,
-        param_id: str,
+        name: str,
         source: ParameterSource,
         value: str,
-        resource_id: str,
+        resource_id: Optional[str],
         metadata: JsonType,
         recompile: bool,
+        expires: bool = False,
```
arnaudsjs, line 230: I don't get this.

* This expires field is only present on the `@handle` method and not on the `@method`. Why?
* Why did we default the `expires` field to `False`? As far as I can see this method allow a user to set parameters and fact. This default doesn't make sense for acts, right?

arnaudsjs (changes requested):

src/inmanta/server/services/paramservice.py
Hugo-Inmanta: If no facts were set then the default value for the `parameters_and_or_facts` variable will be used ( "parameters" ).
If this is not clear I can try to make this more explicit

arnaudsjs: Correct. I missed that.

src/inmanta/protocol/methods.py
```diff
@@ -747,6 +747,7 @@ def set_param(
     resource_id: Optional[str] = None,
     metadata: dict = {},
     recompile: bool = False,
+    expires: Optional[bool] = None,
```
arnaudsjs, line 750: Do you also add documentation about what it means when `expires=None`?

src/inmanta/data/__init__.py
```diff
@@ -3000,13 +3000,18 @@ class Parameter(BaseDocument):
     resource_id: m.ResourceIdStr = ""
     updated: Optional[datetime.datetime] = None
     metadata: Optional[JsonType] = None
-    expires: bool = True
+    expires: bool
 
     @classmethod
-    async def get_updated_before_active_env(cls, updated_before: datetime.datetime) -> list["Parameter"]:
+    async def get_updated_before_active_env(
+        cls, updated_before: datetime.datetime, only_expiring: bool = False
```
arnaudsjs, line 3007: The old implementation would only return parameters that are expiring. This new implementation adds the `only_expiring` argument with the default value set to False. This is a breaking change. Every location where this method is used that doesn't provide the `only_expiring` will not behave differently than. Let's change the default value to True to keep things backwards compatible.

src/inmanta/server/services/paramservice.py
```diff
@@ -200,54 +227,69 @@ async def _update_param(
...
-        params = await data.Parameter.get_list(environment=env.id, name=param_id, resource_id=resource_id)
+        params = await data.Parameter.get_list(environment=env.id, name=name, resource_id=resource_id)
 
         return attach_warnings(200, {"parameter": params[0]}, warnings)
 
     @handle(methods.set_parameters, env="tid")
     async def set_parameters(self, env: data.Environment, parameters: list[dict[str, Any]]) -> Apireturn:
         recompile = False
 
+        updating_facts: bool = False
```
arnaudsjs, line 263: This method also needs to raise an exception is the provided expires parameter is invalid.

arnaudsjs (changes requested):

src/inmanta/server/services/paramservice.py
Hugo-Inmanta: I'm not confident about this part. I used a transaction so that we only insert the params iff they all are ok but I'm not sure if that's the way to go. Adding some tests for this would be nice but I don't quite see how to do this because of the `agent_server=True,` on the `set_parameters` endpoint

src/inmanta/data/__init__.py
```diff
@@ -3000,13 +3000,18 @@ class Parameter(BaseDocument):
     resource_id: m.ResourceIdStr = ""
     updated: Optional[datetime.datetime] = None
     metadata: Optional[JsonType] = None
-    expires: bool = True
+    expires: bool
 
     @classmethod
-    async def get_updated_before_active_env(cls, updated_before: datetime.datetime) -> list["Parameter"]:
+    async def get_updated_before_active_env(
+        cls, updated_before: datetime.datetime, only_expiring: bool = True
```
arnaudsjs, line 3007: Now I am wondering why you added the `only_expiring` flag here. This query filters based on the `updated` flag. This flag has no meaning when the fact is not expiring, right?

src/inmanta/server/services/paramservice.py
```diff
@@ -200,54 +230,74 @@ async def _update_param(
...
-            metadata = param["metadata"] if "metadata" in param else None
-            expires = param["expires"] if "expires" in param else True
-
-            result = await self._update_param(env, name, value, source, resource_id, metadata, expires)
-            if result:
-                recompile = True
-                params.append((name, resource_id))
+
+        async with data.Parameter.get_connection() as connection:
+            async with connection.transaction():
```
arnaudsjs, line 273: Why was this transaction added. I don't think this is required here. It will only increase the likelihood of deadlocks.

arnaudsjs (changes requested):

src/inmanta/server/services/paramservice.py
```diff
@@ -200,54 +221,85 @@ async def _update_param(
...
+        self,
+        name: str,
+        resource_id: Optional[str],
+        expires: Optional[bool],
+    ) -> None:
+        if not resource_id and expires:
+            # Parameters cannot expire
+            raise BadRequest(
+                "Cannot update or set parameter %s: `expire` set to True but parameters cannot expire."
+                "Consider using a fact instead by providing a resource_id." % name,
```
arnaudsjs, line 234:
```suggestion
                " Consider using a fact instead by providing a resource_id." % name,
```

tests/test_data.py
```diff
@@ -2254,7 +2254,13 @@ async def test_parameter(init_dataclasses_and_load_schema, halted):
     for current_time in [time1, time2, time3]:
         t = current_time.strftime("%Y-%m-%dT%H:%M:%S.%f")
         parameter = data.Parameter(
-            name="param_" + t, value="test_val_" + t, environment=env.id, source="test", updated=current_time
+            name="param_" + t,
+            value="test_val_" + t,
+            environment=env.id,
+            source="test",
+            updated=current_time,
+            resource_id="test::SetExpiringFact",
```
arnaudsjs, line 2262: This is not a valid resource_id.

src/inmanta/server/services/paramservice.py
```diff
@@ -200,54 +221,85 @@ async def _update_param(
...
             name: str = param["id"]
             source = param["source"]
             value = param["value"] if "value" in param else None
             resource_id: ResourceIdStr = param["resource_id"] if "resource_id" in param else None
             metadata = param["metadata"] if "metadata" in param else None
-            expires = param["expires"] if "expires" in param else True
+            expires = param["expires"] if "expires" in param else None
 
-            result = await self._update_param(env, name, value, source, resource_id, metadata, expires)
+            self._validate_parameter(name, resource_id, expires)
```
arnaudsjs, line 285: We should move this validation up, so that we refuse the entire request if there is one invalid parameter in the list. Before applying any changes.

arnaudsjs (changes requested):
I would also add a check constraint to the database to enforce that the resource_id and the expires field are consistent.

```
ALTER TABLE <table>
ADD CONSTRAINT <constraint_name>
CHECK (resource_id!='' OR NOT expires);
```

arnaudsjs (approved):

Issue #7020, opened 2024-01-22
Add volume for pg to docker compose in documentation (and quickstarts)


bartv:
Currently all the docker compose files we use do not configure a volume for postgres. Although we highly discourage to use this for production (especially with the db in there). It does make mistakes less harmful if we configure a volume for the postgres data.

PR #7057, opened 2024-01-26
Add volume PostgreSQL server in docker-compose setup

arnaudsjs:
# Description

This PR updates the documentation of the docker-compose setup. It adds a named volume to the PostgreSQL server to not lose any data when the container restarts.

Part of inmanta/inmanta-core#7020

# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/install/2-install-server-with-docker.rst
+++ b/docs/install/2-install-server-with-docker.rst
@@ -72,10 +72,14 @@ Here is a minimalistic docker-compose file content that can be used to deploy th
                 environment:
                     POSTGRES_USER: inmanta
                     POSTGRES_PASSWORD: inmanta
+                    PGDATA: /var/lib/postgresql/data/pgdata
                 networks:
                     inm_net:
                         ipv4_address: 172.30.0.2
-
+                volumes:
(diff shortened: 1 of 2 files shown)


FloLey (approved):

docs/install/2-install-server-with-docker.rst
```diff
@@ -147,6 +158,10 @@ Here is a minimalistic docker-compose file content that can be used to deploy th
 
 You should be able to reach the orchestrator to this address: `http://172.30.0.3:8888 <http://172.30.0.3:8888>`_.
 
+The PostgreSQL server started by the above-mentioned docker-compose file has a named volume ``pgdata`` attached. This
+means that no data will go lost when the PostgreSQL container restarts. Pass the ``-v`` option to the
```
Hugo-Inmanta, line 162: This reads slightly better to me:
```suggestion
means that no data will be lost when the PostgreSQL container restarts. Pass the ``-v`` option to the
```

Hugo-Inmanta (approved):
Small wording suggestion, lgtm otherwise

PR #7058, opened 2024-01-26
Issue/5778 option delegate to config

### LINKED ISSUE #5778: inconsistency between Config and Option
`inmanta.config.Option.get` seems to go straight to `ConfigParser.get`, skipping over `inmanta.config.Config.get`. This means that we may get a different value for `Config.get("section", "name")` than for `corresponding_option.get()`. Specifically, only one of the two considers env vars.

# Task
- [ ] investigate current behavior
- [ ] make sure `Option` has the same behavior as `Config`. This *probably* means that `Option.get` should call into `Config.get` instead of `ConfigParser.get`

sanderr:
# Description

- Made config.Option behave like config.Config
- improved typing for Config and Option. There are still some errors in there but it's a step in the right direction
- minor refactor of Config and Option to decrease coupling

closes #5778 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/config.py
+++ b/src/inmanta/config.py
@@ -118,7 +118,7 @@ class Config:
 
     @overload
     @classmethod
-    def get(cls, section: str, name: str, default_value: Optional[str] = None) -> Optional[str]:
+    def get(cls, section: str, name: str, default_value: Optional[str] = None) -> Optional[object]:
         ...
 
     # noinspection PyNoneFunctionAssignment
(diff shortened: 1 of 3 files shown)


src/inmanta/config.py
```diff
@@ -115,35 +118,41 @@ def get(cls) -> ConfigParser:
...
+    @classmethod
+    def get_for_option(cls, option: "Option[T]") -> T:
+        raw_value: Optional[str | T] = cls._get_value(option.section, option.name, option.get_default_value())
+        return option.validate(raw_value)
 
-        val = _get_from_env(section, name)
+    @classmethod
+    def _get_value(cls, section: str, name: str, default_value: Optional[T] = None) -> Optional[str | T]:
+        cfg: ConfigParser = cls._get_instance()
+        val: Optional[str] = _get_from_env(section, name)
```
Hugo-Inmanta, line 149: I'm curious, can the else block below return anything else than `default_value` ? If I understand correctly, this block is reached when the option is not found in the config or in the environment
```
val = cfg.get(section, name, fallback=default_value)
```

Hugo-Inmanta (approved):
One small question, lgtm otherwise :+1:

wouterdb (approved):

src/inmanta/config.py
sanderr: This is the line that searches in the config file. It is reached when there's nothing in the environment.

PR #7059, opened 2024-01-26
References

bartv:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry


src/inmanta/references.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def resolve_reference(self, values: list["ValueReferenceModel"]) -> object:
+        """Resolve the reference.
+
+        :param values: A list of all value reference
+        """
+        print(self.model_fields)
+
+
+class ValueReference(str):
```
bartv, line 65: This is a string so that in the model/compiler you do not have a python object.  https://github.com/inmanta/std/pull/498/files#diff-5e188e9a2f257f5c800131fb03d2c32f7aae6816bfe2ea0ac9812db5a148cffbR646

However, we have to consider to introduce something in the language to safely allow this. For example something like `opaque` which you can also specialize, like `opaque<string>`. This can then both be used for things like allocators and the reference themselves (opaque). These values cannot be assigned to any attribute unless those are opaque as well.

opaque<string> can be assigned to a field of type string. But, the compiler can still guard you from using them where you expect a proper string.

sanderr: > opaque can be assigned to a field of type string. But, the compiler can still guard you from using them where you expect a proper string.

Alternatively, they can not be assigned to a field of type string, but we provide a built-in dereference/unwrap, so you can e.g. `std::print(unwrap(my_opaque))` so you can access it if you really want to but at least it's explicit.

bartv: for the pure opaque ones yes, but this could also be used for the references like thing where you know that they will produce a string but you cannot use it in the model

sanderr: You should not, but you can, can't you? If we need it for allocation, we need to be able to dereference it in the compiler. Therefore we have to be able to dereference them at compile-time, no matter if it's the model or the plugin domain.

sanderr: But I have to think about it some more. Another thing that popped in my mind would be something like `agent<string>` to indicate that this is a reference that will only be resolvable on the agent. Again, I need to think about it, it might not prove useful.

bartv: so, that is what I used opaque for:

- opaque: is what you describe and cannot be assigned or accessed. For example for this thing that made you shiver some time ago: https://code.inmanta.com/solutions/athonet/athonet_mpn/-/blob/master/model/services/mpn.cf?ref_type=heads#L35
- opaque<string>: the secret reference attribute that you can assign to password. 

I get the agent<string> idea, but you might also need to resolve it during compilation.

src/inmanta/ast/type.py
```diff
@@ -248,7 +248,8 @@ def validate(self, value: Optional[object]) -> bool:
         """
         if isinstance(value, AnyType):
             return True
-
+        if references.is_reference_of(value, float):
```
wouterdb, line 251: wicked!

src/inmanta/references.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+class ReferenceArgument(Argument):
+    """Use the value of another reference as an argument for the reference"""
+
+    type: typing.Literal["reference"] = "reference"
+    id: uuid.UUID
+
+    def get_arg_value(self, resource: "resources.Resource", references: dict[uuid.UUID, "Reference"]) -> object:
+        return ReferencePlaceHolder(self.id, references)
```
wouterdb, line 75: I would expect this to return the actual value (not a placeholder)

src/inmanta/resources.py
```diff
@@ -302,10 +414,15 @@ def map_field(
...
             else:
                 value = getattr(model_object, field_name)
-            # serialize to weed out all unknowns
-            # not very efficient, but the tree has to be traversed anyways
-            # passing along the serialized version would break the resource apis
-            json.dumps(value, default=inmanta.util.api_boundary_json_encoder)
+
+            # walk the entire model to find any value references and Unknowns (instead of json dump)
+            if reference_collector:
+                value = collect_references(reference_collector, value, field_name)
```
wouterdb, line 420: does this cover all cases of the json collector?

wouterdb (commented):
Looks very nice, I didn't go in full detail.

src/inmanta/util/dict_path.py
```diff
@@ -271,8 +271,23 @@ def get_path_sections(self) -> Sequence["WildDictPath"]:
...
-    def _validate_container(self, container: object) -> TypeGuard[dict[object, object]]:
-        return isinstance(container, dict)
+    def _validate_container(
+        self, container: object, set: bool = False, remove: bool = False
+    ) -> TypeGuard[dict[object, object]]:
+        """Validate that the container supports the required mapping protocol operations
+
+        TODO: fix the typeguard. Probably by providing multiple methods for validation!
+        """
+        if "__getitem__" not in dir(container):
```
wouterdb, line 281: I would expect `hasttr` here?

src/inmanta/references.py
bartv: The problem is that we do not know if the reference is already deserialized.

src/inmanta/resources.py
bartv: At least according to the comment. We can add more or even do an explicit serialization using our custom encoder

src/inmanta/util/dict_path.py
bartv: For some reason (that I no longer remember) this did not work on our resources.

src/inmanta/export.py
```diff
@@ -507,12 +508,23 @@ def deploy_code(self, conn: protocol.SyncClient, tid: uuid.UUID, version: Option
...
         # Load both resource definition and handlers
         for type_name, resource_definition in resource.get_resources():
             code_manager.register_code(type_name, resource_definition)
+            types.add(type_name)
 
         for type_name, handler_definition in Commander.get_providers():
             code_manager.register_code(type_name, handler_definition)
+            types.add(type_name)
+
+        # Register all reference and mutator code to all resources. This is very coarse grained and can be optimized once
```
bartv, line 522: We could optmize here but in reality I am not sure if that would make things more efficient.

PR #7064, opened 2024-01-29
issue/6096-log-which-pip-index

### LINKED ISSUE #6096: Log which pip index is used to install Python package
The `env.PythonEnvironment._run_pip_install_command()` method logs a message when it's not able to find a certain Python package. However, it doesn't log which indexes were used to do the installation. This issue should make sure this information is available in the log message.

FloLey:
# Description

Add the used pip indexes to the PackageNotFound exception in run_pip

closes #6096 

# Self Check:

- [x] Changelog entry


tests/test_env.py
```diff
@@ -237,7 +237,11 @@ def test_process_env_install_from_index_not_found(
     """
     Attempt to install a package that does not exist from a pip index. Assert the appropriate error is raised.
     """
-    with pytest.raises(env.PackageNotFound, match="Packages this-package-does-not-exist were not found in the given indexes."):
+    expected: str = (
+        "Packages this-package-does-not-exist were not found in the given indexes. "
+        "(Looking in indexes: %s)" % local_module_package_index
+    )
+    with pytest.raises(env.PackageNotFound, match=re.escape(expected)):
```
arnaudsjs, line 244: The indexes are obtained by parsing the output of pip. This is a quite fragile method, so we need to test is properly. We should add tests for the cases where:
* There is more than one index configure
* Indexes are configured via environment variables (`PIP_INDEX_URL` and `PIP_EXTRA_INDEX_URL`).
* Also check if we don't miss a combination here.

tests/test_env.py
```diff
@@ -231,19 +231,68 @@ def test_process_env_install_from_index(
...
         # pass use_system_config=False for security reasons (anyone could publish this package to PyPi)
         env.process_env.install_for_config(
             [Requirement.parse("this-package-does-not-exist")],
-            config=PipConfig(
-                index_url=local_module_package_index,
-            ),
+            config=PipConfig(index_url=local_module_package_index, extra_index_url=extra_indexes if extra_indexes else []),
+        )
```
arnaudsjs, line 255: `extra_indexes` is always a list, right?
```suggestion
            config=PipConfig(index_url=local_module_package_index, extra_index_url=extra_indexes),
        )
```

tests/test_env.py
```diff
@@ -231,19 +231,68 @@ def test_process_env_install_from_index(
...
+) -> None:
+    """
+    Attempt to install a package that does not exist from a pip index defined in the env vars.
+    Assert the appropriate error is raised.
+    """
+    monkeypatch.setenv("PIP_INDEX_URL", "http://example.com/env_1")  # not used as index_url is set in PipConfig
+    monkeypatch.setenv("PIP_EXTRA_INDEX_URL", extra_indexes_env)
+
+    index_url = local_module_package_index
+    extra_index_url = ["http://example.com/extra_index_1", "http://example.com/extra_index_2"]
```
arnaudsjs, line 278:
```suggestion
    extra_index_url = extra_indexes_env.split(" ")
```

tests/test_env.py
```diff
@@ -231,19 +231,68 @@ def test_process_env_install_from_index(
...
+
+    all_indexes = all_indexes + ", " + ", ".join(extra_index_url)
+
+    expected: str = (
+        "Packages this-package-does-not-exist were not found in the given indexes. " "(Looking in indexes: %s)" % all_indexes
+    )
+    with pytest.raises(env.PackageNotFound, match=re.escape(expected)):
+        env.process_env.install_for_config(
+            [Requirement.parse("this-package-does-not-exist")],
+            config=PipConfig(index_url=index_url, extra_index_url=extra_index_url, use_system_config=use_system_config),
```
arnaudsjs, line 295: We should not pass in the `index_url` and the `extra_index_url` here if these parameters should come from the environment variables.

tests/test_env.py
```diff
@@ -231,19 +231,68 @@ def test_process_env_install_from_index(
...
-            ),
+            config=PipConfig(index_url=local_module_package_index, extra_index_url=extra_indexes if extra_indexes else []),
+        )
+
+
+@pytest.mark.slowtest
+@pytest.mark.parametrize_any(
+    "extra_indexes_env", ["", "http://example.com/extra_env_index_1 http://example.com/extra_env_index_2"]
+)
+@pytest.mark.parametrize_any("use_system_config", [True, False])
```
arnaudsjs, line 262: Why do we also parameterize here on `use_system_config=False`? This is already tested by the previous test case, right?

arnaudsjs (changes requested):

tests/test_env.py
```diff
@@ -231,19 +231,68 @@ def test_process_env_install_from_index(
 
 
 @pytest.mark.slowtest
+@pytest.mark.parametrize_any("extra_indexes", [[], ["http://example.com/extra_index_1", "http://example.com/extra_index_2"]])
```
arnaudsjs, line 234: I don't think it's a good idea to use the `examples.com` domain here. It will cause the test case to connect to the internal. This will make the test case more unstable. Let's just create the other indexes in the same way as the `local_module_package_index` fixture does it.

tests/conftest.py
```diff
@@ -1675,54 +1675,65 @@ def tmpvenv_active_inherit(deactive_venv, tmpdir: py.path.local) -> Iterator[env
...
-        os.makedirs(build_dir)
-        # Build modules
-        for module_dir in os.listdir(modules_v2_dir):
-            path: str = os.path.join(modules_v2_dir, module_dir)
-            ModuleTool().build(path=path, output_dir=build_dir)
-        # Download bare necessities
-        CommandRunner(logging.getLogger(__name__)).run_command_and_log_output(
-            ["pip", "download", "setuptools", "wheel"], cwd=build_dir
-        )
+    def _create_local_package_index(modules_v2_dir: str, index_name: str = "simple"):
```
arnaudsjs, line 1684: This method uses a cache directory that is created at the same level in the filesystem as the `modules_v2_dir`. This cache is useful for the index that is built for the `modules_v2_dir`, because like that we don't need to build the index over and over again when re-running the test suite. But this cache is not useful for the `test_process_env_install_from_index_not_found_env_var` test case. That test case passes in a `tmpdir` as `modules_v2_dir`. This means that the cache directory is created outside of the tmpdir. This cache directory is never re-used and never cleaned up.

That said, I think you are making things way too complicated for yourself. All you need is a fixture that returns an empty pip index. So it would be sufficient to create a new fixture that executes `dir2pi(argv=["dir2pi", build_dir])` for an empty `build_dir` and yield the index directory `os.path.join(build_dir, "simple")`. Like that you can keep the `local_module_package_index` fixture as it was.

tests/conftest.py
```diff
@@ -1675,54 +1675,65 @@ def tmpvenv_active_inherit(deactive_venv, tmpdir: py.path.local) -> Iterator[env
...
+            )
 
-        # Build python package repository
-        dir2pi(argv=["dir2pi", build_dir])
-        # Update timestamp file
-        open(timestamp_file, "w").close()
-    else:
-        logger.info(f"Using cache {cache_dir}")
+        if _should_rebuild_cache():
+            logging.info(f"Cache {cache_dir} is dirty. Rebuilding cache.")
```
arnaudsjs, line 1710:
```suggestion
            logging.info(f"Cache %s is dirty. Rebuilding cache.", cache_dir)
```

tests/conftest.py
```diff
@@ -1675,54 +1675,65 @@ def tmpvenv_active_inherit(deactive_venv, tmpdir: py.path.local) -> Iterator[env
...
+            for module_dir in os.listdir(modules_v2_dir):
+                path = os.path.join(modules_v2_dir, module_dir)
+                ModuleTool().build(path=path, output_dir=build_dir)
+            CommandRunner(logging.getLogger(__name__)).run_command_and_log_output(
+                ["pip", "download", "setuptools", "wheel"], cwd=build_dir
+            )
+            dir2pi(argv=["dir2pi", build_dir])
+            open(timestamp_file, "w").close()
+        else:
+            logging.info(f"Using cache {cache_dir}")
```
arnaudsjs, line 1723:
```suggestion
            logging.info(f"Using cache %s", cache_dir)
```

tests/test_env.py
```diff
@@ -232,20 +232,57 @@ def test_process_env_install_from_index(
...
+@pytest.mark.parametrize_any("use_extra_indexes_env", [True, False])
+@pytest.mark.parametrize_any("use_extra_indexes", [True, False])
+@pytest.mark.parametrize_any("use_system_config", [True, False])
+def test_process_env_install_from_index_not_found_env_var(
+    tmpvenv_active: tuple[py.path.local, py.path.local],
+    monkeypatch,
+    create_local_package_index,
+    use_extra_indexes,
+    use_extra_indexes_env,
+    use_system_config,
```
arnaudsjs, line 244:
```suggestion
    use_extra_indexes: bool,
    use_extra_indexes_env: bool,
    use_system_config: bool,
```

tests/test_env.py
```diff
@@ -232,20 +232,57 @@ def test_process_env_install_from_index(
...
 ) -> None:
     """
-    Attempt to install a package that does not exist from a pip index. Assert the appropriate error is raised.
+    Attempt to install a package that does not exist from a pip indexes defined in the env vars, in the pip config or in both.
```
arnaudsjs, line 247:
```suggestion
    Attempt to install a package that does not exist from the pip indexes defined in the env vars, in the pip config or in both.
```

tests/test_env.py
```diff
@@ -232,20 +232,57 @@ def test_process_env_install_from_index(
...
+        expected_indexes = expected_indexes + ", " + ", ".join(extra_index_urls) if extra_index_urls else expected_indexes
+
+        expected: str = (
+            "Packages this-package-does-not-exist were not "
+            "found in the given indexes. (Looking in indexes: %s)" % expected_indexes
         )
+        with pytest.raises(env.PackageNotFound, match=re.escape(expected)):
+            env.process_env.install_for_config(
+                [Requirement.parse("this-package-does-not-exist")],
+                config=PipConfig(index_url=index_url, extra_index_url=extra_index_urls, use_system_config=use_system_config),
```
arnaudsjs, line 284: This test case never sets `index_url` to `None`. We should also test that scenario.

arnaudsjs (changes requested):

tests/conftest.py
```diff
@@ -1674,12 +1674,40 @@ def tmpvenv_active_inherit(deactive_venv, tmpdir: py.path.local) -> Iterator[env
...
+    def _create_local_package_index(index_name: str = "simple"):
+        """
+        Creates an empty pip index with the given index_name.
+        """
+        tmpdir = tempfile.mkdtemp()
+        created_directories.append(tmpdir)  # Keep track of the tempdir for cleanup
+        index_dir = os.path.join(tmpdir, index_name)
+        os.makedirs(index_dir)
+        dir2pi(argv=["dir2pi", str(tmpdir)])
+        return index_dir
```
arnaudsjs, line 1695: I don't understand the purpose of the `index_name` argument. As far as I can see, the `dir2pi` function always creates a sub-directory `simple` that contains the index, so matter what the value of the `index_name` argument is. I think we can remove it and hard-code the value `simple`, like in the original fixture.

tests/conftest.py
```diff
@@ -1674,12 +1674,40 @@ def tmpvenv_active_inherit(deactive_venv, tmpdir: py.path.local) -> Iterator[env
...
+@pytest.fixture(scope="session")
+def create_local_package_index_factory():
+    """
+    A fixture that acts as a factory to create empty local pip package indexes
+    during a test session. Each call creates a new index in a temporary directory.
+    """
+
+    created_directories = []
+
+    def _create_local_package_index(index_name: str = "simple"):
```
arnaudsjs, line 1686:
```suggestion
    def _create_local_package_index(index_name: str = "simple") -> str:
```

tests/conftest.py
```diff
@@ -1674,12 +1674,40 @@ def tmpvenv_active_inherit(deactive_venv, tmpdir: py.path.local) -> Iterator[env
     loader.unload_modules_for_path(venv.site_packages_dir)
 
 
+@pytest.fixture(scope="session")
+def create_local_package_index_factory():
```
arnaudsjs, line 1678:
```suggestion
def create_local_package_index_factory() -> Callable[[], str]:
```

tests/conftest.py
```diff
@@ -1674,12 +1674,40 @@ def tmpvenv_active_inherit(deactive_venv, tmpdir: py.path.local) -> Iterator[env
     loader.unload_modules_for_path(venv.site_packages_dir)
 
 
+@pytest.fixture(scope="session")
```
arnaudsjs, line 1677: Why is this a session scoped fixture? This way we will only cleanup the indexes at the end of the run of the entire test suite. By making it function scoped we cleanup after ever test. That way we are more efficient on disk-space and memory usage.

tests/conftest.py
arnaudsjs, line 1678: Let's put the word "empty" explicitly in the name. Otherwise we get into a naming conflict if we ever want to have a factory for non-empty indexes.
```suggestion
def create_empty_local_package_index_factory():
```

tests/test_env.py
```diff
@@ -232,18 +232,110 @@ def test_process_env_install_from_index(
...
 @pytest.mark.slowtest
-def test_process_env_install_from_index_not_found(
-    tmpvenv_active: tuple[py.path.local, py.path.local], local_module_package_index: str
+@pytest.mark.parametrize_any("use_extra_indexes_env", [True, False])
+@pytest.mark.parametrize_any("use_extra_indexes", [True, False])
+@pytest.mark.parametrize_any("use_system_config", [True, False])
+def test_process_env_install_from_index_not_found_env_var(
+    tmpvenv_active: tuple[py.path.local, py.path.local],
+    monkeypatch,
+    create_local_package_index_factory,
```
arnaudsjs, line 241:
```suggestion
    create_local_package_index_factory: Callable[[], str],
```

tests/test_env.py
```diff
@@ -232,18 +232,110 @@ def test_process_env_install_from_index(
...
+
+    expected: str = (
+        "Packages this-package-does-not-exist were not "
+        "found in the given indexes. (Looking in indexes: %s)" % expected_indexes
+    )
+    with pytest.raises(env.PackageNotFound, match=re.escape(expected)):
         env.process_env.install_for_config(
             [Requirement.parse("this-package-does-not-exist")],
+            config=PipConfig(index_url=index_url, extra_index_url=extra_index_urls, use_system_config=use_system_config),
+        )
```
arnaudsjs, line 284: This code does a lot of string manipulations. Code that manipulates string is often hard to read by nature. In general it's a good idea to store your data in a structured data structure (e.g. list or dict) and only convert it to a string when you need a string. For example, in this code a string are only required at two points: once for setting the environment variable and once to build the string assigned to the `expected` variable. All other operations are just appends to a list but then done using a string manipulation

tests/test_env.py
```diff
@@ -232,18 +232,110 @@ def test_process_env_install_from_index(
...
+
+
+@pytest.mark.slowtest
+def test_process_env_install_with_no_index_url(
```
arnaudsjs, line 288: I now realize that this is a dangerous test, because it opens the door for a dependency confusion attack. We should remove it.

tests/test_env.py
```diff
@@ -232,18 +232,110 @@ def test_process_env_install_from_index(
...
+
+    with pytest.raises(env.PackageNotFound, match=re.escape(expected_error_message)):
+        env.process_env.install_for_config(
+            [Requirement.parse("this-package-does-not-exist")],
+            config=PipConfig(index_url=index_url, extra_index_url=extra_index_urls, use_system_config=True),
+        )
+
+
+@pytest.mark.slowtest
+def test_install_package_not_found_in_default_index(
```
arnaudsjs, line 319: The same for this test.

arnaudsjs (changes requested):

tests/conftest.py
wouterdb: But we do become faster because we don't have to rebuild the index every time?

wouterdb: (ignore that, it is a non-caching factory)

tests/conftest.py
FloLey: Indeed, but I still liked the fact that in the test, by reading the call you would directly see that those are different indexes for sure, without the need to look at the fixture. I changed the description of the fixture to emphasize that each call creates a different temp dir.

arnaudsjs: Fine by me to keep the argument if you feel like it makes the code more readable. But we should make sure that the returned `index_dir` refers to the index created by the `dir2pi()` method (i.e. the "simple" directory).

FloLey: Also to debug the test it helps: while rewriting a test, I got this error: 
```Packages this-package-does-not-exist were not found in the given indexes. (Looking in indexes: /tmp/tmpw83luvxx/simple, /tmp/tmpvcvwlq50/simple, /tmp/tmp1cpkys1u/simple, /tmp/tmp4sbqgugm/simple, /tmp/tmppxit2vyy/simple, /tmp/tmpvcvwlq50/simple, /tmp/tmp1cpkys1u/simple)``` and it is really hard to know what index is missing/wrongly placed. 
I'll put it back and make sure to return the index_dir referes to the right index

arnaudsjs: The `tempfile.mkdtemp()` method accepts a `prefix` argument. You could incorporate the `index_name` in the directory name like that, while still returning the correct path.

arnaudsjs (approved):

PR #7077, opened 2024-01-30
JWT Claim constraints

bartv:
# Description

This PR implements custom claim support.

# Self Check:

- [ ] Changelog entry


src/inmanta/config.py
```diff
@@ -445,6 +447,80 @@ def __init__(self, name: str, port: int = 8888) -> None:
 # auth
 #############################
 AUTH_JWT_PREFIX = "auth_jwt_"
+AUTH_JWT_CLAIM_RE = r"^\s*([^\s].+)\s+(in|is)\s+([^\s].+)\s*$"
+
+claim_type = dict[str, str | list[str]]
+
+
+@dataclasses.dataclass
```
wouterdb, line 455: All of the auth config could be moved into a separate module or in the server config?

Config is loaded very early, so we do have to be aware for import loops.

wouterdb (commented):
It would be interesting if we could test it some more completely. Perhaps generating a token ( we have that in testing already) and make a client call would be interesting (with e.g. valid, invalid and garbage claims).

src/inmanta/config.py
```diff
@@ -445,6 +447,80 @@ def __init__(self, name: str, port: int = 8888) -> None:
...
+    :return: The result of the check
+    """
+    try:
+        return all(constraint.match_claim(claims) for constraint in claim_constraints)
+    except Exception as e:
+        LOGGER.info(f"The configured claim constraints failed to evaluate against the provided claims: {e}")
+        return False
 
 
 class AuthJWTConfig:
```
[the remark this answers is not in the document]
bartv: But is not a problem I want to solve in this ticket ;-)

src/inmanta/config.py
bartv: seems to work

bartv:
> Looks nice
> 
> It would be interesting if we could test it some more completely. Perhaps generating a token ( we have that in testing already) and make a client call would be interesting (with e.g. valid, invalid and garbage claims).

done

src/inmanta/config.py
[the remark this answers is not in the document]
wouterdb: Fair.

src/inmanta/config.py
bartv: Not sure how I will be able to move this into inmanta.protocol without import problems. See CI

wouterdb (approved):

wouterdb (approved):

PR #7084, opened 2024-01-30
issue/7019-remove-netiface-from-reporting

### LINKED ISSUE #7019: Install from source misses requirements
The install from source section in the documentation is missing that requirements such as gcc and python-devel are required. This is for example required for netifaces dependency.

FloLey:
# Description

 drop the functionality relying on netifaces

closes #7019 

# Self Check:

- [x] Changelog entry


arnaudsjs (approved):
Looks good to me, but wait for the impact analysis of Lukas before merging this.

PR #7101, opened 2024-02-01
[POC] agent refactor

Hugo-Inmanta:
# Description

Agent refactor poc:

- bring together the code loading and code install in a single class `InProcessExecutor`
- to allow the new behavior to be easily plugged in the `ExternalExecutor` class

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -48,7 +48,7 @@ from inmanta.data.model import LEGACY_PIP_DEFAULT, AttributeStateChange, PipConf
 from inmanta.loader import CodeLoader, ModuleSource
 from inmanta.protocol import SessionEndpoint, SyncClient, methods, methods_v2
 from inmanta.resources import Id, Resource
-from inmanta.types import Apireturn, JsonType
+from inmanta.types import Apireturn, JsonType, FailedResourcesSet
 from inmanta.util import (
     CronSchedule,
     IntervalSchedule,
@@ -264,8 +264,8 @@ class InProcessExecutor(Executor):
         self.agent: AgentInstance = agent
 
     @classmethod
-    async def get_executor(cls, agent: "AgentInstance", code: Sequence[ResourceInstallSpec]) -> tuple[Self, Set[str]]:
-        failed_resource_types: Set[str] = await agent.process.ensure_code(code)
+    async def get_executor(cls, agent: "AgentInstance", code: Sequence[ResourceInstallSpec]) -> tuple[Self, FailedResourcesSet]:
+        failed_resource_types: FailedResourcesSet = await agent.process.ensure_code(code)
         return cls(agent), failed_resource_types
 
     async def send_in_progress(
@@ -1298,7 +1298,7 @@ class AgentInstance:
 
             return await executor.get_facts(resource_refs[0])
 
-    async def get_executor(self, code: Sequence[ResourceInstallSpec]) -> tuple[Executor, Set[str]]:
+    async def get_executor(self, code: Sequence[ResourceInstallSpec]) -> tuple[Executor, FailedResourcesSet]:
         return await InProcessExecutor.get_executor(self, code)
 
     async def setup_executor(
@@ -1321,12 +1321,12 @@ class AgentInstance:
         code: Sequence[ResourceInstallSpec]
         # Resource types for which no handler code exist for the given version
         # or for which the pip config couldn't be retrieved
-        invalid_resource_types: Set[str]
+        invalid_resource_types: FailedResourcesSet
         code, invalid_resource_types = await self.process.get_code(
             self._env_id, version, [res["resource_type"] for res in resources]
         )
         # Resource types for which an error occurred during handler code installation
-        failed_resource_types: Set[str]
+        failed_resource_types: FailedResourcesSet
         executor, failed_resource_types = await self.get_executor(code)
 
         loaded_resources: list[ResourceDetails] = []
@@ -1610,7 +1610,7 @@ class Agent(SessionEndpoint):
 
     async def get_code(
         self, environment: uuid.UUID, version: int, resource_types: Sequence[str]
-    ) -> tuple[Collection[ResourceInstallSpec], set[str]]:
+    ) -> tuple[Collection[ResourceInstallSpec], FailedResourcesSet]:
         """
         Get the collection of installation specifications (i.e. pip config, python package dependencies, Inmanta modules sources)
         required to deploy a given version for the provided resource types.
@@ -1626,7 +1626,7 @@ class Agent(SessionEndpoint):
         pip_config: Optional[PipConfig] = None
 
         resource_install_specs: list[ResourceInstallSpec] = []
-        invalid_resource_types: set[str] = set()
+        invalid_resource_types: FailedResourcesSet = set()
         for resource_type in set(resource_types):
             LOGGER.debug("Building ResourceInstallSpec for resource_type=%s version=%d", resource_type, version)
 
@@ -1667,9 +1667,9 @@ class Agent(SessionEndpoint):
 
         return resource_install_specs, invalid_resource_types
 
-    async def ensure_code(self, code: Sequence[ResourceInstallSpec]) -> Set[str]:
+    async def ensure_code(self, code: Sequence[ResourceInstallSpec]) -> FailedResourcesSet:
         """Ensure that the code for the given environment and version is loaded"""
-        failed_to_load: set[str] = set()
+        failed_to_load: FailedResourcesSet = set()
         for resource_install_spec in code:
             # only one logical thread can load a particular resource type at any time
             async with self._resource_loader_lock.get(resource_install_spec.resource_type):
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -223,12 +223,24 @@ class Executor(ABC):
     @abc.abstractmethod
     async def dry_run(
         self,
-        resource_ref: Sequence[ResourceDetails],
+        resources: Sequence[ResourceDetails],
+        dry_run_id: uuid.UUID,
+        undeployable: Mapping[ResourceVersionIdStr, ResourceState],
     ) -> None:
         """
         Perform a dryrun for the given resources
 
-        :param resource_ref: Sequence of resources for which to perform a dryrun.
+        :param resources: Sequence of resources for which to perform a dryrun.
+        :param dry_run_id: id for this dryrun
+        :param undeployable: maps undeployable resources to their current state
+        """
+        pass
+
+    @abc.abstractmethod
+    async def get_facts(self, resource: ResourceDetails) -> Apireturn:
+        """
+        Get facts for a given resource
+        :param resource: The resource for which to get facts.
         """
         pass
 
@@ -425,11 +437,18 @@ class InProcessExecutor(Executor):
         if response.code != 200:
             LOGGER.error("Resource status update failed %s", response.result)
 
-    async def dry_run(self, resources: Sequence[ResourceDetails], dry_run_id: uuid.UUID) -> None:
+    async def dry_run(
+        self,
+        resources: Sequence[ResourceDetails],
+        dry_run_id: uuid.UUID,
+        undeployable: dict[ResourceVersionIdStr, ResourceState],
+    ) -> None:
         """
         Perform a dryrun for the given resources
 
         :param resources: Sequence of resources for which to perform a dryrun.
+        :param dry_run_id: id for this dryrun
+        :param undeployable: maps undeployable resources to their current state
         """
         model_version: int = resources[0].model_version
         env_id: uuid.UUID = resources[0].env_id
@@ -452,22 +471,22 @@ class InProcessExecutor(Executor):
                         resource_id=resource_id,
                         status=undeployable[resource_id],
                     )
-                    await self.get_client().dryrun_update(tid=self._env_id, id=dry_run_id, resource=resource_id, changes={})
+                    await self.agent.get_client().dryrun_update(tid=env_id, id=dry_run_id, resource=resource_id, changes={})
                     continue
 
                 try:
-                    self.logger.debug("Running dryrun for %s", resource_id)
+                    self.agent.logger.debug("Running dryrun for %s", resource_id)
 
                     try:
-                        provider = await self.get_provider(resource)
+                        provider = await self.agent.get_provider(resource_obj)
                     except Exception as e:
                         ctx.exception(
                             "Unable to find a handler for %(resource_id)s (exception: %(exception)s",
                             resource_id=resource_id,
                             exception=str(e),
                         )
-                        await self.get_client().dryrun_update(
-                            tid=self._env_id,
+                        await self.agent.get_client().dryrun_update(
+                            tid=env_id,
                             id=dry_run_id,
                             resource=resource_id,
                             changes={"handler": {"current": "FAILED", "desired": "Unable to find a handler"}},
@@ -475,7 +494,7 @@ class InProcessExecutor(Executor):
                     else:
                         try:
                             await asyncio.get_running_loop().run_in_executor(
-                                self.thread_pool, provider.execute, ctx, resource, True
+                                self.agent.thread_pool, provider.execute, ctx, resource, True
                             )
 
                             changes = ctx.changes
@@ -483,37 +502,37 @@ class InProcessExecutor(Executor):
                                 changes = {}
                             if ctx.status == const.ResourceState.failed:
                                 changes["handler"] = AttributeStateChange(current="FAILED", desired="Handler failed")
-                            await self.get_client().dryrun_update(
-                                tid=self._env_id, id=dry_run_id, resource=resource_id, changes=changes
+                            await self.agent.get_client().dryrun_update(
+                                tid=env_id, id=dry_run_id, resource=resource_id, changes=changes
                             )
                         except Exception as e:
                             ctx.exception(
                                 "Exception during dryrun for %(resource_id)s (exception: %(exception)s",
-                                resource_id=str(resource.id),
+                                resource_id=resource.rvid,
                                 exception=str(e),
                             )
                             changes = ctx.changes
                             if changes is None:
                                 changes = {}
                             changes["handler"] = AttributeStateChange(current="FAILED", desired="Handler failed")
-                            await self.get_client().dryrun_update(
-                                tid=self._env_id, id=dry_run_id, resource=resource_id, changes=changes
+                            await self.agent.get_client().dryrun_update(
+                                tid=env_id, id=dry_run_id, resource=resource_id, changes=changes
                             )
 
                 except Exception:
                     ctx.exception("Unable to process resource for dryrun.")
                     changes = {}
                     changes["handler"] = AttributeStateChange(current="FAILED", desired="Resource Deserialization Failed")
-                    await self.get_client().dryrun_update(
-                        tid=self._env_id, id=dry_run_id, resource=resource_id, changes=changes
+                    await self.agent.get_client().dryrun_update(
+                        tid=env_id, id=dry_run_id, resource=resource_id, changes=changes
                     )
                 finally:
                     if provider is not None:
                         provider.close()
 
                     finished = datetime.datetime.now().astimezone()
-                    await self.get_client().resource_action_update(
-                        tid=self._env_id,
+                    await self.agent.get_client().resource_action_update(
+                        tid=env_id,
                         resource_ids=[resource_id],
                         action_id=ctx.action_id,
                         action=const.ResourceAction.dryrun,
@@ -522,6 +541,65 @@ class InProcessExecutor(Executor):
                         messages=ctx.logs,
                         status=const.ResourceState.dry,
                     )
+
+    async def get_facts(self, resource: ResourceDetails) -> Apireturn:
+        """
+        Get facts for a given resource
+        :param resource: The resource for which to get facts.
+        """
+        model_version: int = resource.model_version
+        env_id: uuid.UUID = resource.env_id
+
+        provider = None
+        try:
+            try:
+                resource_obj: Resource = await self.deserialize(resource, const.ResourceAction.getfact)
+            except Exception:
+                return 500
+            ctx = handler.HandlerContext(resource_obj)
+
+            with self.cache(model_version):
+                try:
+                    started = datetime.datetime.now().astimezone()
+                    provider = await self.agent.get_provider(resource_obj)
+                    result = await asyncio.get_running_loop().run_in_executor(
+                        self.agent.thread_pool, provider.check_facts, ctx, resource_obj
+                    )
+
+                    parameters = [
+                        {
+                            "id": name,
+                            "value": value,
+                            "resource_id": resource.rid,
+                            "source": ParameterSource.fact.value,
+                        }
+                        for name, value in result.items()
+                    ]
+                    # Add facts set via the set_fact() method of the HandlerContext
+                    parameters.extend(ctx.facts)
+
+                    await self.agent.get_client().set_parameters(tid=env_id, parameters=parameters)
+                    finished = datetime.datetime.now().astimezone()
+                    await self.agent.get_client().resource_action_update(
+                        tid=env_id,
+                        resource_ids=[resource.rvid],
+                        action_id=ctx.action_id,
+                        action=const.ResourceAction.getfact,
+                        started=started,
+                        finished=finished,
+                        messages=ctx.logs,
+                    )
+
+                except Exception:
+                    self.agent.logger.exception("Unable to retrieve fact")
+
+        except Exception:
+            self.agent.logger.exception("Unable to find a handler for %s", resource.id)
+            return 500
+        finally:
+            if provider is not None:
+                provider.close()
+        return 200
 
     def cache(self, model_version: int) -> CacheVersionContext:
         return self.agent._cache.manager(model_version)
@@ -1186,11 +1264,11 @@ class AgentInstance:
                 if len(resources) > 0:
                     self._nq.reload(resources, undeployable, deploy_request, executor)
 
-    async def dryrun(self, dry_run_id: uuid.UUID, env_id: uuid.UUID, version: int) -> Apireturn:
-        self.process.add_background_task(self.do_run_dryrun(version, dry_run_id, env_id))
+    async def dryrun(self, dry_run_id: uuid.UUID, version: int) -> Apireturn:
+        self.process.add_background_task(self.do_run_dryrun(version, dry_run_id))
         return 200
 
-    async def do_run_dryrun(self, version: int, dry_run_id: uuid.UUID, env_id: uuid.UUID) -> None:
+    async def do_run_dryrun(self, version: int, dry_run_id: uuid.UUID) -> None:
 
         async with self.dryrunlock:
             async with self.ratelimiter:
@@ -1208,107 +1286,10 @@ class AgentInstance:
                 )
 
                 await executor.dry_run(resource_refs, dry_run_id, undeployable)
-                self._cache.open_version(version)
-                for resource_ref in resource_refs:
-                    try:
-                        resource: Resource = await executor.deserialize(resource_ref, const.ResourceAction.dryrun)
-                    except Exception:
-                        await self.get_client().dryrun_update(
-                            tid=self._env_id, id=dry_run_id, resource=resource_ref.rvid, changes={}
-                        )
-                        continue
 
-                    await executor.dry_run(resource)
-                    ctx = handler.HandlerContext(resource, True)
-                    started = datetime.datetime.now().astimezone()
-                    provider = None
-
-                    if resource_ref.rvid in undeployable:
-                        ctx.error(
-                            "Skipping dryrun %(resource_id)s because in undeployable state %(status)s",
-                            resource_id=resource_ref.rvid,
-                            status=undeployable[resource_ref.rvid],
-                        )
-                        await self.get_client().dryrun_update(
-                            tid=self._env_id, id=dry_run_id, resource=resource_ref.rvid, changes={}
-                        )
-                        continue
-
-                    try:
-                        self.logger.debug("Running dryrun for %s", resource_ref.rvid)
-
-                        try:
-                            provider = await self.get_provider(resource)
-                        except Exception as e:
-                            ctx.exception(
-                                "Unable to find a handler for %(resource_id)s (exception: %(exception)s",
-                                resource_id=resource_ref.rvid,
-                                exception=str(e),
-                            )
-                            await self.get_client().dryrun_update(
-                                tid=self._env_id,
-                                id=dry_run_id,
-                                resource=resource_ref.rvid,
-                                changes={"handler": {"current": "FAILED", "desired": "Unable to find a handler"}},
-                            )
-                        else:
-                            try:
-                                await asyncio.get_running_loop().run_in_executor(
-                                    self.thread_pool, provider.execute, ctx, resource, True
-                                )
-
-                                changes = ctx.changes
-                                if changes is None:
-                                    changes = {}
-                                if ctx.status == const.ResourceState.failed:
-                                    changes["handler"] = AttributeStateChange(current="FAILED", desired="Handler failed")
-                                await self.get_client().dryrun_update(
-                                    tid=self._env_id, id=dry_run_id, resource=resource_ref.rvid, changes=changes
-                                )
-                            except Exception as e:
-                                ctx.exception(
-                                    "Exception during dryrun for %(resource_id)s (exception: %(exception)s",
-                                    resource_id=resource_ref.rvid,
-                                    exception=str(e),
-                                )
-                                changes = ctx.changes
-                                if changes is None:
-                                    changes = {}
-                                changes["handler"] = AttributeStateChange(current="FAILED", desired="Handler failed")
-                                await self.get_client().dryrun_update(
-                                    tid=self._env_id, id=dry_run_id, resource=resource_ref.rvid, changes=changes
-                                )
-
-                    except Exception:
-                        ctx.exception("Unable to process resource for dryrun.")
-                        changes = {}
-                        changes["handler"] = AttributeStateChange(current="FAILED", desired="Resource Deserialization Failed")
-                        await self.get_client().dryrun_update(
-                            tid=self._env_id, id=dry_run_id, resource=resource_ref.rvid, changes=changes
-                        )
-                    finally:
-                        if provider is not None:
-                            provider.close()
-
-                        finished = datetime.datetime.now().astimezone()
-                        await self.get_client().resource_action_update(
-                            tid=self._env_id,
-                            resource_ids=[resource_ref.rvid],
-                            action_id=ctx.action_id,
-                            action=const.ResourceAction.dryrun,
-                            started=started,
-                            finished=finished,
-                            messages=ctx.logs,
-                            status=const.ResourceState.dry,
-                        )
-
-                self._cache.close_version(version)
-
-    async def get_facts(self, resource: JsonType, env_id: uuid.UUID) -> Apireturn:
+    async def get_facts(self, resource: JsonType) -> Apireturn:
         async with self.ratelimiter:
-            # this only works with InProcessExecutor
-            # TODO: adapt for ExternalExecutor
-            undeployable, resource_refs, _ = await self.setup_executor(
+            undeployable, resource_refs, executor = await self.setup_executor(
                 resource["model"], const.ResourceAction.getfact, [resource]
             )
 
@@ -1318,76 +1299,7 @@ class AgentInstance:
                 )
                 return 500
 
-            started = datetime.datetime.now().astimezone()
-            provider = None
-            try:
-                resource_ref: ResourceDetails = resource_refs[0]
-                try:
-                    resource_obj: Resource = Resource.deserialize(resource_ref.attributes)
-                except Exception:
-                    msg = data.LogLine.log(
-                        level=const.LogLevel.ERROR,
-                        msg="Unable to deserialize %(resource_id)s",
-                        resource_id=resource_ref.rvid,
-                        timestamp=datetime.datetime.now().astimezone(),
-                    )
-
-                    await self.get_client().resource_action_update(
-                        tid=env_id,
-                        resource_ids=[resource_ref.rvid],
-                        action_id=uuid.uuid4(),
-                        action=const.ResourceAction.getfact,
-                        started=started,
-                        finished=datetime.datetime.now().astimezone(),
-                        status=const.ResourceState.unavailable,
-                        messages=[msg],
-                    )
-                    return 500
-                ctx = handler.HandlerContext(resource_obj)
-
-                try:
-                    self._cache.open_version(resource_ref.model_version)
-                    provider = await self.get_provider(resource_obj)
-                    result = await asyncio.get_running_loop().run_in_executor(
-                        self.thread_pool, provider.check_facts, ctx, resource_obj
-                    )
-
-                    parameters = [
-                        {
-                            "id": name,
-                            "value": value,
-                            "resource_id": resource_obj.id.resource_str(),
-                            "source": ParameterSource.fact.value,
-                        }
-                        for name, value in result.items()
-                    ]
-                    # Add facts set via the set_fact() method of the HandlerContext
-                    parameters.extend(ctx.facts)
-
-                    await self.get_client().set_parameters(tid=self._env_id, parameters=parameters)
-                    finished = datetime.datetime.now().astimezone()
-                    await self.get_client().resource_action_update(
-                        tid=self._env_id,
-                        resource_ids=[resource_ref.rvid],
-                        action_id=ctx.action_id,
-                        action=const.ResourceAction.getfact,
-                        started=started,
-                        finished=finished,
-                        messages=ctx.logs,
-                    )
-
-                except Exception:
-                    self.logger.exception("Unable to retrieve fact")
-                finally:
-                    self._cache.close_version(resource_ref.model_version)
-
-            except Exception:
-                self.logger.exception("Unable to find a handler for %s", resource["id"])
-                return 500
-            finally:
-                if provider is not None:
-                    provider.close()
-            return 200
+            return await executor.get_facts(resource_refs[0])
 
     async def get_executor(self, code: Sequence[ResourceInstallSpec]) -> tuple[Executor, Set[str]]:
         return await InProcessExecutor.get_executor(self, code)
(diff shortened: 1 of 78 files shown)


plan.md
```diff
@@ -0,0 +1,20 @@
+1 - refactor to work with dict repr
```
sanderr, line 1: Reminder to drop this file at some point.

sanderr (commented):
Things I'll still get to tomorrow:
- review `get_code` and `ensure_code`
- give the `HandlerLogger` hack some thought.

src/inmanta/agent/agent.py
```diff
@@ -236,17 +455,18 @@ async def execute(
                     self.running = False
 
             async with self.scheduler.ratelimiter:
-                ctx = handler.HandlerContext(self.resource, logger=self.logger)
+                self.running = True
```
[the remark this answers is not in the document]
Hugo-Inmanta: I don't recall moving this on purpose, I'll move it back

src/inmanta/agent/agent.py
```diff
@@ -1033,41 +1232,52 @@ async def get_facts(self, resource: JsonType) -> Apireturn:
...
+                loaded_resources.append(res["attributes"])
 
-            except Exception:
+                state = const.ResourceState[res["status"]]
+                if state in const.UNDEPLOYABLE_STATES:
+                    undeployable[res["id"]] = state
+            else:
                 failed_resources.append(res["id"])
                 undeployable[res["id"]] = const.ResourceState.unavailable
-                resource = Resource.deserialize(res["attributes"], use_generic=True)
```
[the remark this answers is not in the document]
sanderr: If I understand correctly we only ever did this when we also added it to `undeployable`, meaning that the scheduler would never even try to deploy it, therefore we don't need the specific handler type. We did it just to present a `Resource`-like object, because that is what the surrounding code expected. Now that we use dicts here, there's no need for that anymore, and at the point where we deserialize now, there is no point to keeping the object at all anymore, we can just report to the server directly and be done.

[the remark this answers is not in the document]
Hugo-Inmanta: Makes sense :+1:

src/inmanta/agent/agent.py
```diff
@@ -137,43 +140,259 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
         super().__init__(scheduler, dummy_id, gid, reason)
 
 
+class Executor(ABC):
+    @classmethod
+    @abc.abstractmethod
+    async def get_executor(
+        cls, agent: "AgentInstance", code: Sequence[InstallBlueprint], client: protocol.Client
+    ) -> tuple[Self, Set[str]]:
```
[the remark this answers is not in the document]
Hugo-Inmanta: Something that is still blurry to me is how this is going to integrate with the next step. i.e. we currently have only one in-process executor but we will have to adapt this to interface with the `AgentSupervisor` and multiple executors ? Maybe this is just out of scope for now?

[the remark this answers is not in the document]
sanderr: `AgentSupervisor` is for a future stage, remember? We will have to adapt it with external executors though.

src/inmanta/agent/agent.py
```diff
@@ -137,43 +140,259 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+                messages=ctx.logs,
+                changes=changes,
+                change=resource_change,
+            )
+        else:
+            response = await self.client.resource_deploy_done(
+                tid=env_id,
+                rvid=resource_id.resource_version_str(),
+                action_id=uuid.uuid4(),
+                status=resource_status,
```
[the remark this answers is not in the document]
Hugo-Inmanta: removed the method as per your other [comment](https://github.com/inmanta/inmanta-core/pull/7101/files#r1491178181)

src/inmanta/agent/agent.py
```diff
@@ -192,24 +252,136 @@ async def _execute(self, ctx: handler.HandlerContext, requires: dict[ResourceIdS
...
+    async def execute(
+        self,
+        ctx: handler.HandlerLogger,
+        gid: uuid.UUID,
+        resource_id: Id,
+        resource: dict[str, object],
+        env_id: uuid.UUID,
+        agent: "AgentInstance",
+    ) -> tuple[Optional[const.ResourceState], const.Change, dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]]:
+        ctx.debug("Start deploy %(deploy_id)s of resource %(resource_id)s", deploy_id=gid, resource_id=resource_id)
```
sanderr, line 309: I would move this to after deserialization so we can use the proper handler context.

src/inmanta/agent/agent.py
```diff
@@ -137,43 +140,259 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+        try:
+            resource: Resource = Resource.deserialize(resource)
+        except Exception:
+
+            ctx.exception("Unable to deserialize %(resource_id)s", resource_id=resource_id.resource_version_str())
+            return await self._report_resource_status(
+                env_id=env_id,
+                resource_id=resource_id,
+                resource_status=const.ResourceState.unavailable,
+                resource_change=const.Change.nochange,
```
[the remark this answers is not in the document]
sanderr: This would also allow us to attach the log message (like the old implementation did) instead of using the handler context.

src/inmanta/agent/agent.py
```diff
@@ -236,17 +408,18 @@ async def execute(
...
+                ctx = handler.HandlerLogger(self.resource_id, logger=self.logger)
 
                 ctx.debug(
                     "Start run for resource %(resource)s because %(reason)s",
-                    resource=str(self.resource.id),
+                    resource=str(self.resource_id),
                     deploy_id=self.gid,
                     agent=self.scheduler.agent.name,
                     reason=self.reason,
                 )
```
sanderr, line 420: Just had a quick chat with Wouter and we believe this can be safely moved to the executor as well. If I understand correctly this would then completely remove the need for the `HandlerLogger` hack, so I think we can revert the changes we made there.

src/inmanta/agent/agent.py
```diff
@@ -192,24 +252,136 @@ async def _execute(self, ctx: handler.HandlerContext, requires: dict[ResourceIdS
...
+        agent: "AgentInstance",
+    ) -> tuple[Optional[const.ResourceState], const.Change, dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]]:
+        ctx.debug("Start deploy %(deploy_id)s of resource %(resource_id)s", deploy_id=gid, resource_id=resource_id)
+        # TODO: double check: is this required? Is failure handled properly?
+
+        try:
+            resource: Resource = Resource.deserialize(resource)
+        except Exception:
+
+            ctx.exception("Unable to deserialize %(resource_id)s", resource_id=resource_id.resource_version_str())
```
Hugo-Inmanta, line 316: here

src/inmanta/agent/agent.py
```diff
@@ -1332,52 +1490,81 @@ async def on_disconnect(self) -> None:
...
+                    pass
+
+        return sorted(resource_install_specs, key=lambda resource_install_spec: resource_install_spec.resource_type)
+
+    async def ensure_code(self, code: Sequence[ResourceInstallSpec]) -> Set[str]:
+        """Ensure that the code for the given environment and version is loaded"""
+        failed_to_load: set[str] = []
+        for resource_install_spec in code:
+            # only one logical thread can load a particular resource type at any time
+            async with self._resource_loader_lock.get(resource_install_spec.resource_type):
```
sanderr, line 1545: Open question (forgot about this one for a bit): this lock used to be around both the fetching of the code and the install+load. Now it's only in the install + load part.
1. do we still need this lock here? I think yes
2. do we need this lock where we fetch the code? I think not.

src/inmanta/agent/agent.py
```diff
@@ -1332,52 +1490,81 @@ async def on_disconnect(self) -> None:
...
-                    continue
-                # clear cache, for retry on failure
-                self._last_loaded[rt] = -1
+        resource_install_specs: list[ResourceInstallSpec] = []
+        for resource_type in set(resource_types):
 
-                result: protocol.Result = await self._client.get_source_code(environment, version, rt)
+            if self._last_loaded[resource_type] == version:
+                LOGGER.debug("Code already present for %s version=%d", resource_type, version)
+                continue
```
sanderr, line 1506: I'm not entirely certain that this is correct now. We're only fetching the code here. The `ensure_code` method that now populates this cache is in-process while it might be that we're fetching code for an external executor.

src/inmanta/agent/agent.py
```diff
@@ -1332,52 +1490,81 @@ async def on_disconnect(self) -> None:
...
+        """Ensure that the code for the given environment and version is loaded"""
+        failed_to_load: set[str] = []
+        for resource_install_spec in code:
+            # only one logical thread can load a particular resource type at any time
+            async with self._resource_loader_lock.get(resource_install_spec.resource_type):
+                # stop if the last successful load was this one
+                # The combination of the lock and this check causes the reloads to naturally 'batch up'
+
+                # clear cache, for retry on failure
+                self._last_loaded[resource_install_spec.resource_type] = -1
```
sanderr, line 1550: With the new implementation we no longer skip if it is in the cache, while I think that would still be a valid approach, right?

changelogs/9.3.0/update-documentation-keycloak.yml
```diff
@@ -2,4 +2,4 @@ description: "Update the documentation about setting up authentication, to use K
 change-type: patch
 destination-branches: [master, iso6]
```
wouterdb, line 3:
```suggestion
destination-branches: [master]
```

src/inmanta/agent/agent.py
```diff
@@ -137,52 +142,124 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+    resource_type: str
+    model_version: int
+    pip_config: PipConfig
+    requirements: Sequence[str]
+    sources: Sequence["ModuleSource"]
+
+
+class Executor(ABC):
+    """
+    This class encapsulates abstraction around a process able to deploy a resource.
```
wouterdb, line 167:
```suggestion
    This class represents a process able to deploy a resource.
```

src/inmanta/agent/agent.py
```diff
@@ -137,52 +142,124 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+        cls, agent: "AgentInstance", code: Sequence[ResourceInstallSpec], client: protocol.Client
+    ) -> tuple[Self, Set[str]]:
+        pass
+
+    @abc.abstractmethod
+    async def execute(
+        self,
+        gid: uuid.UUID,
+        resource_id: Id,
+        resource: dict[str, object],
```
wouterdb, line 186: I would like the resources to be a class as well (e.g. `ResourceReference`)

src/inmanta/agent/agent.py
```diff
@@ -192,24 +269,160 @@ async def _execute(self, ctx: handler.HandlerContext, requires: dict[ResourceIdS
...
+        env_id: uuid.UUID,
+        reason: str,
+    ) -> tuple[Optional[const.ResourceState], const.Change, dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]]:
+        started: datetime.datetime = datetime.datetime.now().astimezone()
+        # TODO: double check: is this required? Is failure handled properly?
+
+        try:
+            resource: Resource = Resource.deserialize(resource)
+        except Exception:
+            msg = data.LogLine.log(level=const.LogLevel.ERROR, msg="Unable to deserialize %(resource_id)s", resource_id=resource_id.resource_version_str(), timestamp=datetime.datetime.now().astimezone()),
```
wouterdb, line 321: review note: this is new, how was it handled before?

src/inmanta/agent/agent.py
```diff
@@ -192,24 +269,160 @@ async def _execute(self, ctx: handler.HandlerContext, requires: dict[ResourceIdS
...
+            await self.client.resource_action_update(
+                tid=env_id,
+                resource_ids=[resource_id.resource_version_str()],
+                action_id=uuid.uuid4(),
+                action=const.ResourceAction.deploy,
+                started=started,
+                finished=datetime.datetime.now().astimezone(),
+                status=const.ResourceState.unavailable,
+                messages=[msg]
+            )
```
wouterdb, line 332: I would expect a return statement here

src/inmanta/agent/agent.py
```diff
@@ -192,24 +269,160 @@ async def _execute(self, ctx: handler.HandlerContext, requires: dict[ResourceIdS
...
+
+        try:
+            requires: dict[ResourceIdStr, const.ResourceState] = await self.send_in_progress(
+                ctx.action_id, env_id, resource_id.resource_version_str()
+            )
+        except Exception:
+            ctx.set_status(const.ResourceState.failed)
+            ctx.exception("Failed to report the start of the deployment to the server")
+        else:
+            await self._execute(resource, ctx=ctx, requires=requires)
```
wouterdb, line 351: I find it more readable if you add this to the end of the try-except block.

(for me, the main flow of the code should be visually apparent, not hidden behind an exception handling block)

src/inmanta/agent/agent.py
```diff
@@ -192,24 +269,160 @@ async def _execute(self, ctx: handler.HandlerContext, requires: dict[ResourceIdS
...
+        self.executor: Executor = executor
+        self.env_id: uuid.UUID = env_id
+
     async def execute(
         self, dummy: "ResourceActionBase", generation: "Dict[ResourceIdStr, ResourceActionBase]", cache: AgentCache
     ) -> None:
         self.logger.log(const.LogLevel.TRACE.to_int, "Entering %s %s", self.gid, self.resource)
-        with cache.manager(self.resource.id.get_version()):
-            self.dependencies = [generation[x.resource_str()] for x in self.resource.requires]
+        with cache.manager(self.resource_id.get_version()):
```
wouterdb, line 424: the cache will have to live on the executor. 

This will somewhat complicate the executor, as it will have state (the cache state)

src/inmanta/agent/agent.py
```diff
@@ -882,7 +1064,7 @@ async def do_run_dryrun(self, version: int, dry_run_id: uuid.UUID) -> None:
                     self.logger.warning("Got an error while pulling resources and version %s", version)
                     return
 
-                undeployable, resources = await self.load_resources(
+                undeployable, resources, _ = await self.setup_executor(
```
wouterdb, line 1067: This can use a very large TODO, as it will only work with the local one

src/inmanta/agent/agent.py
```diff
@@ -974,7 +1156,7 @@ async def do_run_dryrun(self, version: int, dry_run_id: uuid.UUID) -> None:
 
     async def get_facts(self, resource: JsonType) -> Apireturn:
         async with self.ratelimiter:
-            undeployable, resources = await self.load_resources(resource["model"], const.ResourceAction.getfact, [resource])
+            undeployable, resources, _ = await self.setup_executor(resource["model"], const.ResourceAction.getfact, [resource])
```
wouterdb, line 1159: This can use a very large TODO, as it will only work with the local one

src/inmanta/agent/agent.py
Hugo-Inmanta: I don't see how this would look like (assuming we want to keep the current behavior of only running `self._execute` if no exceptions were raised during `send_in_progress`)

src/inmanta/agent/agent.py
```diff
@@ -137,43 +140,259 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+    @abc.abstractmethod
+    async def execute(
+        self,
+        ctx: handler.HandlerLogger,
+        gid: uuid.UUID,
+        resource_id: Id,
+        resource: dict[str, object],
+        env_id: uuid.UUID,
+        agent: "AgentInstance",
+    ) -> tuple[Optional[const.ResourceState], const.Change, dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]]:
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think they're not even used it tests as far as I can see, do you know if they're still used @wouterdb ?

src/inmanta/agent/agent.py
Hugo-Inmanta: brain fart: just added a return in the except block

Hugo-Inmanta:
Mypy complains still:
```diff
(inmanta-core311) hugo@hugo-Latitude-5421:~/work/inmanta/github-repos/inmanta-core$ make mypy-diff 
+ src/inmanta/agent/handler.py:243: error: Argument 3 to "log" of "LogLine" has incompatible type "**dict[str, object]"; expected "datetime | None"  [arg-type]
- src/inmanta/agent/handler.py:425: error: Argument 3 to "log" of "LogLine" has incompatible type "**dict[str, object]"; expected "datetime | None"  [arg-type]
+ src/inmanta/agent/agent.py:397: error: Returning Any from function declared to return "CacheVersionContext"  [no-any-return]
+ src/inmanta/agent/agent.py:397: error: Cannot determine type of "_cache"  [has-type]
+ src/inmanta/agent/agent.py:442: error: Argument 1 to "parse_id" of "Id" has incompatible type "object"; expected "ResourceVersionIdStr | ResourceIdStr"  [arg-type]
+ src/inmanta/agent/agent.py:450: error: "object" has no attribute "__iter__"; maybe "__dir__" or "__str__"? (not iterable)  [attr-defined]
+ src/inmanta/agent/agent.py:753: error: Argument 1 to "parse_id" of "Id" has incompatible type "object"; expected "ResourceVersionIdStr | ResourceIdStr"  [arg-type]
+ src/inmanta/agent/agent.py:758: error: Attribute "generation" already defined on line 676  [no-redef]
+ src/inmanta/agent/agent.py:759: error: Argument 1 to "parse_id" of "Id" has incompatible type "object"; expected "ResourceVersionIdStr | ResourceIdStr"  [arg-type]
+ src/inmanta/agent/agent.py:772: error: "object" has no attribute "__iter__"; maybe "__dir__" or "__str__"? (not iterable)  [attr-defined]
+ src/inmanta/agent/agent.py:1107: error: "object" has no attribute "resource_version_str"  [attr-defined]
+ src/inmanta/agent/agent.py:1114: error: "object" has no attribute "resource_version_str"  [attr-defined]

```

src/inmanta/agent/agent.py
```diff
@@ -137,52 +139,136 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+    """
+
+    resource_type: str
+    model_version: int
+    pip_config: PipConfig
+    requirements: Sequence[str]
+    sources: Sequence["ModuleSource"]
+
+
+class ResourceReference(dict[str, object]):
```
wouterdb, line 162: this has to be a proper object, not just a dict. I would have an attributes field, an ID field an a requires field

src/inmanta/agent/agent.py
```diff
@@ -79,11 +81,13 @@ def __str__(self) -> str:
...
     """Base class for Local and Remote resource actions"""
 
+    # Model dependencies between resources + inform each other when they're done deploying
+
     resource_id: Id
     future: ResourceActionResultFuture
     dependencies: list["ResourceActionBase"]
     # resourceid -> attribute -> {current: , desired:}
-    changes: dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]
+    # changes: dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]
```
wouterdb, line 90: These commented lines have to be cleaned up

Hugo-Inmanta: We have a discussion over [there](https://github.com/inmanta/inmanta-core/pull/7101#discussion_r1494916852) about these fields. They aren't used anymore are they ?

Hugo-Inmanta:
Latest mypy diff
```diff
(inmanta-core311) hugo@hugo-Latitude-5421:~/work/inmanta/github-repos/inmanta-core$ make mypy-diff 
+ src/inmanta/agent/handler.py:243: error: Argument 3 to "log" of "LogLine" has incompatible type "**dict[str, object]"; expected "datetime | None"  [arg-type]
- src/inmanta/agent/handler.py:425: error: Argument 3 to "log" of "LogLine" has incompatible type "**dict[str, object]"; expected "datetime | None"  [arg-type]
+ src/inmanta/agent/agent.py:399: error: Returning Any from function declared to return "CacheVersionContext"  [no-any-return]
+ src/inmanta/agent/agent.py:399: error: Cannot determine type of "_cache"  [has-type]
```

src/inmanta/agent/agent.py
```diff
@@ -159,11 +157,15 @@ class ResourceInstallSpec:
     sources: Sequence["ModuleSource"]
 
 
-class ResourceReference(dict[str, object]):
+class ResourceReference(BaseModel):
```
wouterdb, line 160: Why get pydantic involved I wonder?

src/inmanta/agent/agent.py
```diff
@@ -325,7 +327,7 @@ async def execute(
         # TODO: double check: is this required? Is failure handled properly?
 
         try:
-            resource: Resource = Resource.deserialize(resource_ref)
+            resource: Resource = Resource.deserialize(resource_ref.model_dump()["attributes"])
```
wouterdb, line 330:
```suggestion
            resource: Resource = Resource.deserialize(resource_ref.attributes)
```

src/inmanta/agent/agent.py
wouterdb, line 160: I would add a class method that constructs this from the dict

src/inmanta/agent/agent.py
```diff
@@ -439,15 +441,15 @@ def __init__(
         """
         :param gid: A unique identifier to identify a deploy. This is local to this agent.
         """
-        super().__init__(scheduler, Id.parse_id(resource["id"]), gid, reason)
+        super().__init__(scheduler, Id.parse_id(ResourceVersionIdStr(str(resource.attributes["id"]))), gid, reason)
```
wouterdb, line 444: ????

```suggestion
        super().__init__(scheduler, resource.id, gid, reason)
```

src/inmanta/agent/agent.py
```diff
@@ -159,11 +157,15 @@ class ResourceInstallSpec:
...
 
 
-class ResourceReference(dict[str, object]):
+class ResourceReference(BaseModel):
     """
-    In memory dict representation of the desired state of a resource
+    In memory representation of the desired state of a resource
     """
 
+    id: ResourceIdStr
```
wouterdb, line 165: Isn't this a ResourceVersionIdStr ?

src/inmanta/agent/agent.py
```diff
@@ -1097,29 +1090,33 @@ async def do_run_dryrun(self, version: int, dry_run_id: uuid.UUID, env_id: uuid.
...
                 self._cache.open_version(version)
                 for resource_ref in resource_refs:
+                    started = datetime.datetime.now().astimezone()
+                    id = str(resource_refs[0].attributes["id"])
+
+                    resource_id = Id.parse_id(ResourceVersionIdStr(id)).resource_version_str()
+
                     try:
-                        resource: Resource = Resource.deserialize(resource_ref)
+                        resource: Resource = Resource.deserialize(resource_ref.model_dump()["attributes"])
```
wouterdb, line 1099:
```suggestion
                        resource: Resource = Resource.deserialize(resource_ref.attributes)
```

wouterdb (changes requested):
first impressions

tests/agent_server/test_facts.py
```diff
@@ -431,4 +433,4 @@ async def test_get_fact_no_code(resource_container, client, clienthelper, enviro
     log_entry = result["logs"][0]
     assert log_entry["action"] == "getfact"
     assert log_entry["status"] == "unavailable"
-    assert "Failed to load" in log_entry["messages"][0]["msg"]
+    assert "Unable to deserialize" in log_entry["messages"][0]["msg"]
```
Hugo-Inmanta, line 436: For this change and the one in test_server_agent, I don't fully understand what we're testing.
These changes make the tests pass but I'm not sure if we're still testing the same thing as before

src/inmanta/agent/agent.py
```diff
@@ -137,52 +136,153 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+        :param env_id: the environment
+        :param reason: textual reason for this deploy
         """
-        super().__init__(scheduler, resource.id, gid, reason)
-        self.resource: Resource = resource
+        pass
+
+    @abc.abstractmethod
+    def cache(self, model_version: int) -> CacheVersionContext:
+        # Todo: properly type return value -> create sibling class ExecutorCache from AgentCache?
```
wouterdb, line 220: it will move in its entirety

src/inmanta/agent/agent.py
```diff
@@ -137,52 +136,153 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+    raw: JsonType
+
+    def __init__(self, resource_dict: JsonType) -> None:
+        self.raw = resource_dict
+        self.attributes = resource_dict["attributes"]
+        self.attributes["id"] = resource_dict["id"]
+        self.id = Id.parse_id(resource_dict["id"])
+        self.rid = self.id.resource_str()
+        self.rvid = self.id.resource_version_str()
+        self.requires = resource_dict["attributes"]["requires"]
```
wouterdb, line 179: We may want to parse them already, but that is a very very minor detail

src/inmanta/agent/agent.py
wouterdb: As it is now, I would simply cache the set of `ResourceInstallSpec`  for a given `(resourcetype, version)`

src/inmanta/agent/agent.py
```diff
@@ -1332,63 +1597,93 @@ async def on_disconnect(self) -> None:
...
+        resource_install_specs: list[ResourceInstallSpec] = []
+        for resource_type in set(resource_types):
+
+            if self._last_loaded[resource_type] == version:
+                LOGGER.debug("Code already present for %s version=%d", resource_type, version)
+                continue
             # only one logical thread can load a particular resource type at any time
-            async with self._resource_loader_lock.get(rt):
+            # TODO deal with this lock  -> not sure how
+            async with self._resource_loader_lock.get(resource_type):
```
wouterdb, line 1619: This has no function here, ensure_code does it already

src/inmanta/agent/agent.py
```diff
@@ -1332,63 +1597,93 @@ async def on_disconnect(self) -> None:
...
+                    resource_install_specs.append(
+                        ResourceInstallSpec(resource_type, version, pip_config, list(requirements), sources)
+                    )
+                else:
+                    # TODO what if server call fails ?
+                    pass
+
+        return resource_install_specs
+
+    async def ensure_code(self, code: Sequence[ResourceInstallSpec]) -> Set[str]:
```
wouterdb, line 1651: All of the source code handling is not quite in the right place yet, but I think we can gloss over that for now.

src/inmanta/agent/agent.py
```diff
@@ -1033,41 +1087,69 @@ async def do_run_dryrun(self, version: int, dry_run_id: uuid.UUID) -> None:
...
+                undeployable, resource_refs, _ = await self.setup_executor(
                     version, const.ResourceAction.dryrun, response.result["resources"]
                 )
 
                 self._cache.open_version(version)
-                for resource in resources:
+                for resource_ref in resource_refs:
+                    started = datetime.datetime.now().astimezone()
+                    try:
+                        resource: Resource = Resource.deserialize(resource_ref.attributes)
```
sanderr, line 1100: Shouldn't this be part of the executor?

src/inmanta/agent/agent.py
```diff
@@ -1136,12 +1222,34 @@ async def get_facts(self, resource: JsonType) -> Apireturn:
             started = datetime.datetime.now().astimezone()
             provider = None
             try:
-                resource_obj = resources[0]
+                # id = str(resource_refs[0].attributes["id"])
+                # resource_id = Id.parse_id(ResourceVersionIdStr(id)).resource_version_str()
+                resource_ref: ResourceReference = resource_refs[0]
+                try:
+                    resource_obj: Resource = Resource.deserialize(resource_ref.attributes)
```
sanderr, line 1227: Same here

src/inmanta/agent/agent.py
```diff
@@ -1491,84 +1598,91 @@ async def on_disconnect(self) -> None:
...
                         )
-                        requirements.update(source["requirements"])
+                    )
+                    requirements.update(source["requirements"])
 
-                    if pip_config is None:
-                        # TODO: what if this fails?
-                        pip_config = await self._get_pip_config(environment, version)
+                if pip_config is None:
+                    # TODO: what if this fails?
```
sanderr, line 1636: TODO alert

src/inmanta/agent/agent.py
```diff
@@ -1491,84 +1598,91 @@ async def on_disconnect(self) -> None:
...
-                        ResourceInstallSpec(resource_type, version, pip_config, list(requirements), sources)
-                    )
-                else:
-                    # TODO what if server call fails ?
-                    pass
+                resource_install_specs.append(
+                    ResourceInstallSpec(resource_type, version, pip_config, list(requirements), sources)
+                )
+            else:
+                # TODO what if server call fails ?
```
sanderr, line 1643: TODO alert

src/inmanta/agent/agent.py
```diff
@@ -161,9 +155,37 @@ class ResourceInstallSpec:
...
+    In memory representation of the desired state of a resource
+    """
+
+    id: Id
+    rid: ResourceIdStr
+    rvid: ResourceVersionIdStr
+    model_version: int
+    requires: Sequence[Id]
+    attributes: dict[str, Any]
+    raw: JsonType
```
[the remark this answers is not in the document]
Hugo-Inmanta: removed

src/inmanta/agent/agent.py
```diff
@@ -270,72 +309,78 @@ async def _report_resource_status(
...
-        resource: dict[str, object],
+        resource_ref: ResourceReference,
         env_id: uuid.UUID,
-        agent: "AgentInstance",
-    ) -> tuple[Optional[const.ResourceState], const.Change, dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]]:
-        ctx.debug("Start deploy %(deploy_id)s of resource %(resource_id)s", deploy_id=gid, resource_id=resource_id)
+        reason: str,
+    ) -> None:
+        started: datetime.datetime = datetime.datetime.now().astimezone()
         # TODO: double check: is this required? Is failure handled properly?
```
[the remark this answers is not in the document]
Hugo-Inmanta: removed, I think this is now handled

src/inmanta/agent/agent.py
Hugo-Inmanta: I replicated the initial behavior: mark resources for which we can't get the model source or the pip config as failed

src/inmanta/agent/agent.py
Hugo-Inmanta: moved to the executor

src/inmanta/agent/agent.py
Hugo-Inmanta: moved to the executor

src/inmanta/agent/agent.py
```diff
@@ -137,52 +135,184 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+    rid: ResourceIdStr
+    rvid: ResourceVersionIdStr
+    env_id: uuid.UUID
+    model_version: int
+    requires: Sequence[Id]
+    attributes: dict[str, object]
+
+    def __init__(self, resource_dict: JsonType) -> None:
+        self.attributes = resource_dict["attributes"]
+        self.attributes["id"] = resource_dict["id"]
```
wouterdb, line 173: two questions here:
1. why don't we keep the resource_dict itself? I would expect it to be required?
2. why modify the attributes, seems risky?

src/inmanta/agent/agent.py
```diff
@@ -137,52 +135,184 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+    @abc.abstractmethod
+    async def get_facts(self, resource: ResourceDetails) -> Apireturn:
+        """
+        Get facts for a given resource
+        :param resource: The resource for which to get facts.
+        """
+        pass
+
+    @abc.abstractmethod
+    def cache(self, model_version: int) -> CacheVersionContext:
```
wouterdb, line 248: this can use a docstring to indicate what it is for.

src/inmanta/agent/agent.py
```diff
@@ -1033,41 +1232,52 @@ async def get_facts(self, resource: JsonType) -> Apireturn:
                     provider.close()
             return 200
 
-    async def load_resources(
+    async def get_executor(self, code: Sequence[InstallBlueprint], client: protocol.Client) -> tuple[Executor, Set[str]]:
+        # ultimately:
+        # process_manager.get_executor_for_resouce(version, )
+        return await OldExecutor.get_executor(self, code, client)
+
+    async def setup_executor(
```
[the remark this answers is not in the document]
wouterdb: agreed, I think we can take this on later

wouterdb (approved):

src/inmanta/agent/agent.py
Hugo-Inmanta: @sanderr please correct me if I'm wrong there,

> 1. why don't we keep the resource_dict itself? I would expect it to be required?

Do you have something specific in mind for which it is required ? As far as I can tell current behavior is in line with the original implementation where we use:
- the raw JsonType resource's `resource_type` attribute to get the required code (now in `get_code`, previously in `ensure_code`
- the raw JsonType resource's `attributes` attribute during resource deserialization (now happening later in the flow, in the executor)

> 2. why modify the attributes, seems risky?

The end goal is that we need an `id` field to be present in the dict passed to `Resource.deserialize(<dict>)`. Would it be better if add the `id` field to `resource_dict["attributes"]` prior to the `ResourceDetails` constructor call ? That's what the old implementation did but I don't think it is any better.

An alternative would be to to something like `Resource.deserialize({**attributes, 'id': rvid})` ?

> seems risky 

Why is it risky I'm curious ?

sanderr: I think both 1 and 2 have remained essientially unchanged versus master. Can you confirm, @Hugo-Inmanta?

Hugo-Inmanta: yes

src/inmanta/agent/agent.py
```diff
@@ -1332,63 +1617,98 @@ async def on_disconnect(self) -> None:
...
-            return failed_to_load
+            return [], set()
 
         # store it outside the loop, but only load when required
         pip_config: Optional[PipConfig] = None
 
-        for rt in set(resource_types):
+        resource_install_specs: list[ResourceInstallSpec] = []
+        invalid_resource_types: set[str] = set()
+        for resource_type in set(resource_types):
```
sanderr, line 1635: This makes the iteration order nondeterministic. Not sure if this is an issue, but if the collection has no meaningful order we should probably return `Collection[ResourceInstallSpec]` instead of `list`. I think it's good practice in general but it's especially important here to signal that the caller can not rely on order, because some sort of order is really important eventually.

src/inmanta/agent/agent.py
```diff
@@ -1332,63 +1617,98 @@ async def on_disconnect(self) -> None:
...
+                )
+            else:
+                invalid_resource_types.add(resource_type)
+
+        return resource_install_specs, invalid_resource_types
+
+    async def ensure_code(self, code: Sequence[ResourceInstallSpec]) -> Set[str]:
+        """Ensure that the code for the given environment and version is loaded"""
+        failed_to_load: set[str] = set()
+        # breakpoint()
```
sanderr, line 1678: breakpoint alert

src/inmanta/agent/agent.py
```diff
@@ -1332,63 +1617,98 @@ async def on_disconnect(self) -> None:
...
+        for resource_install_spec in code:
             # only one logical thread can load a particular resource type at any time
-            async with self._resource_loader_lock.get(rt):
+            async with self._resource_loader_lock.get(resource_install_spec.resource_type):
                 # stop if the last successful load was this one
                 # The combination of the lock and this check causes the reloads to naturally 'batch up'
-                if self._last_loaded[rt] == version:
-                    LOGGER.debug("Code already present for %s version=%d", rt, version)
+                if self._last_loaded.get((resource_install_spec.resource_type, resource_install_spec.model_version)):
```
sanderr, line 1684: I don't think this holds anymore: `_last_loaded` doesn't really represent "last loaded" anymore. Just "loaded at some point. Either we have to change it back to mean "last loaded" or we have to be smarter about it.

src/inmanta/agent/handler.py
```diff
@@ -205,8 +205,48 @@ def _log_msg(
         raise NotImplementedError
 
 
+# TODO: extract logging behavior from HandlerContext
+class HandlerLogger(LoggerABC):
```
sanderr, line 209: I vaguely recall that we didn't need this anymore? i.e. I think all changes in this file can be reverted, unless you see something we want to keep?

sanderr (changes requested):
Did a final end-to-end review.

src/inmanta/agent/agent.py
```diff
@@ -137,43 +140,259 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+        resource_id: Id,
+        resource: dict[str, object],
+        env_id: uuid.UUID,
+        agent: "AgentInstance",
+    ) -> tuple[Optional[const.ResourceState], const.Change, dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]]:
+        pass
+
+
+class OldExecutor(Executor):
+    client: protocol.Client
```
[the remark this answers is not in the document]
sanderr: This is still the case in some places, but not a huge issue.

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
sanderr: @Hugo-Inmanta see this thread if there's any confusion about my comment on the docstring that mentions `AgentSupervisor`.

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
sanderr: Can you link the ticket if/when you create(d) it?

sanderr:
Nice work! Looks like it came together nicely!

src/inmanta/agent/agent.py
```diff
@@ -192,24 +326,338 @@ async def _execute(self, ctx: handler.HandlerContext, requires: dict[ResourceIdS
...
+                except Exception:
+                    await self.agent.get_client().dryrun_update(tid=env_id, id=dry_run_id, resource=resource.rvid, changes={})
+                    continue
+                assert resource_obj is not None
+                ctx = handler.HandlerContext(resource_obj, True)
+                started = datetime.datetime.now().astimezone()
+                provider = None
+
+                resource_id: ResourceVersionIdStr = resource.rvid
+                if resource_id in undeployable:
```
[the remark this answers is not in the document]
Hugo-Inmanta: > Any reason for the behavioral difference

Not that I know of, this is in line with the original implementation

>  we report it in the scheduler iirc?

Yes, and for get_facts the agent instance's logger is used : https://github.com/inmanta/inmanta-core/blob/cda078e2e80aef775767f0c5277adeba86e58798/src/inmanta/agent/agent.py#L984

Should we make an effort to make this more uniform ?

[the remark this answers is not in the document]
sanderr: > Not that I know of, this is in line with the original implementation

How do you mean? The original implementation did not have an executor.

> Yes, and for get_facts the agent instance's logger is used

Which is the same logger that we use to construct the `HandlerContext` iirc?

FloLey (approved):

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
Hugo-Inmanta: > How do you mean? The original implementation did not have an executor.

I meant we are still reporting through the handler context, as it was the case before. But maybe I misunderstood what you meant with `Here we report undeployable in the executor` ?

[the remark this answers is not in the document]
Hugo-Inmanta: > Which is the same logger that we use to construct the HandlerContext iirc?

ah, yes indeed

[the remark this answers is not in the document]
sanderr: I mean that in one case it's the responsibility of the executor while in the other case it's the responsibility of the scheduler. That makes little sense to me: the scheduler and executor should have a clear responsibility, independent of which flow it's executing. Unless there's a reason to implement it like this?

src/inmanta/agent/agent.py
```diff
@@ -137,52 +135,188 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+
+
+class Executor(ABC):
+    """
+    An executor for resources. Delegates to the appropriate handlers.
+    """
+
+    @classmethod
+    @abc.abstractmethod
+    async def get_executor(cls, agent: "AgentInstance", code: Sequence[ResourceInstallSpec]) -> tuple[Self, Set[str]]:
```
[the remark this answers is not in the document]
wouterdb: I wouldn't invest too much time in this, will be affected quite heavily by the merge with the code of @FloLey

src/inmanta/agent/agent.py
Hugo-Inmanta: Added back a dict to keep track of the actual last loaded version + fixed a bug where the try..catch surrounding the installation below was executing outside of the `_resource_loader_lock`

src/inmanta/agent/agent.py
```diff
@@ -1672,39 +1672,54 @@ async def get_code(
...
+                    LOGGER.debug(
+                        "Installed handler %s version=%d",
+                        resource_install_spec.resource_type,
+                        resource_install_spec.model_version,
+                    )
+                    # Update the ``_previously_loaded`` cache to indicate that the given resource type's code
+                    # was loaded successfully at the specified version.
+                    self._previously_loaded[(resource_install_spec.resource_type, resource_install_spec.model_version)] = (
+                        resource_install_spec
+                    )
```
sanderr, line 1714: I would think that this one is the responsibility of `get_code` rather than `ensure_code`? i.e. `get_code` fetches the code and manages its cache, `ensure_code` loads code and keeps track of what is currently loaded.

src/inmanta/agent/agent.py
```diff
@@ -1672,39 +1672,54 @@ async def get_code(
...
-                )
-            except Exception:
-                LOGGER.exception(
-                    "Failed to install handler %s version=%d",
-                    resource_install_spec.resource_type,
-                    resource_install_spec.model_version,
-                )
-                failed_to_load.add(resource_install_spec.resource_type)
+
+                self._last_loaded_version[resource_install_spec.resource_type] = -1
```
sanderr, line 1691: Might make more sense in the except block?

sanderr (changes requested):

sanderr (approved):
I'm rerequesting review from @wouterdb to double check the changes I requested since his review.

PR #7102, opened 2024-02-01
POC. light ipc

wouterdb:
# Description

A basic IPC protocol, that is async and integrates with multiprocessing

# TODO

- [x] exception handling
- [x] exception propagation
- [x] process monitoring 
- [x] cleanup of child process? 
- [x] stop procedure
- [x] document shutdown procedure
- [x] monitoring of child process -> not required: pipe breaks immediately
- [x] proper disconnect on server side (kill futures on disconnect)

All advanced features are missing:
- rate limiting
- timeouts
- keep-alive messages / connection shutdown on silence -> less relevant, as the pipe seems to close very fast and clean


arnaudsjs (commented):
Looks good to me. But I have a feeling that the changes to the `clean_reset` fixture might cause trouble. Let's see what the CI thinks about it.

arnaudsjs (approved):

src/inmanta/protocol/ipc_light.py
```diff
@@ -0,0 +1,218 @@
...
+
+class ConnectionLost(IPCException):
+    pass
+
+
+@dataclass
+class IPCRequestFrame:
+    id: uuid.UUID
+    method: str
+    arguments: list[object]
```
sanderr, line 42: I wonder if we might improve on this by, instead of sending a method name and a list of arguments, to send a proper method request object, e.g.

```python
def f(x: int, y: str) -> None:
    pass

@dataclass
class IPCMethodCall: pass

@dataclass
class IPCRequestFrame:
    id: uuid.UUID
    method: IPCMethodCall

@dataclass
class IPCMethodCallF(IPCMethodCall):
    x: int
    y: str


def process_incoming_ipc_frame(frame: IPCRequestFrame) -> None:
    match frame.method:
        case IPCMethodCallF(x, y): f(x, y)
        case _: raise Exception()
```

> I wonder if we might improve on this

It's an open question. I haven't given it sufficient thought to be certain what benefit we get out of it. But I get the impression that it would give us a very low-effort typed interface.

sanderr (commented):
Looks good to me. I added one potential easy win improvement point. I only went over it very high-level.

src/inmanta/protocol/ipc_light.py
wouterdb: Even more interesting
```python
def f(x: int, y: str) -> None:
    pass


class IPCMethodCall(abc.ABC): 

     @abc.abstractmethod
     def call(self):
          ...

@dataclass
class IPCRequestFrame:
    id: uuid.UUID
    method: IPCMethodCall

@dataclass
class IPCMethodCallF(IPCMethodCall):
    x: int
    y: str
    
    def call():
        f(x,y)


def process_incoming_ipc_frame(frame: IPCRequestFrame) -> None:
   frame.method.call()
   
```

The big different is that the caller get's full control over what the server offers: if it can load it, it can execute it. 
Which is cool, but may be considered overpowered.

wouterdb: We will keep this in mind for a next iteration, when we integrate

PR #7106, opened 2024-02-01
Improve typing

bartv:
# Description

Some more type fixes I made after the claims PR (+ some old stuff I still had)

# Self Check:

- [ ] Changelog entry


src/inmanta/module.py
```diff
@@ -1997,7 +2007,7 @@ def __init__(
                 self.virtualenv = env.VirtualEnv(venv_path)
 
         self.loaded = False
-        self.modules: dict[str, Module] = {}
+        self.modules: dict[str, Module[ModuleMetadata]] = {}
```
sanderr, line 2010: Don't these changes raise as many errors as they solve? `TModuleMetadata` is invariant, so I would expect this to raise errors elsewhere. Iirc that's the reason I've left this one alone up till now.

src/inmanta/moduletool.py
```diff
@@ -181,7 +183,7 @@ def determine_new_version(self, old_version, version, major, minor, patch, dev):
             if not dev:
                 outversion = baseversion
             else:
-                outversion = "%s.dev%d" % (baseversion, time.time())
+                outversion = cast(Version, parse_version("%s.dev%d" % (baseversion, time.time())))
```
sanderr, line 186: Actual change request: I don't think we should cast this. This is not the same class (although they may have the same implementation). We should get rid of `pkg_resources` for iso8 anyway so this will be resolved then.

src/inmanta/moduletool.py
```diff
@@ -832,10 +835,10 @@ def v1tov2(self, module: str) -> None:
         """
         Convert a V1 module to a V2 module in place
         """
-        module = self.get_module(module)
```
sanderr, line 835: Just curious, why do you prefer `mod` over `module`? I tend to write it full out.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -58,7 +58,7 @@
 from inmanta.server.agentmanager import AgentManager, AutostartedAgentManager
 from inmanta.server.services.resourceservice import ResourceService
 from inmanta.server.validate_filter import InvalidFilter
-from inmanta.types import Apireturn, JsonType, PrimitiveTypes
+from inmanta.types import Apireturn, JsonType, PrimitiveTypes, ReturnTupple
```
sanderr, line 61: Oh, there's a typo in this one... First time I notice. Little to be done about it though.

src/inmanta/server/services/compilerservice.py
```diff
@@ -32,7 +32,7 @@
 from itertools import chain
 from logging import Logger
 from tempfile import NamedTemporaryFile
-from typing import Optional, cast
+from typing import Optional, Type, cast
```
sanderr, line 35: Actual change request: pretty sure this is deprecated in favor of built-in `type`.

src/inmanta/server/services/compilerservice.py
```diff
@@ -500,7 +504,7 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
 
     def __init__(self) -> None:
         super().__init__(SLICE_COMPILER)
-        self._recompiles: dict[uuid.UUID, Task] = {}
+        self._recompiles: dict[uuid.UUID, Task[Type[None]]] = {}
```
sanderr, line 507: Why `type[None]` instead of just `None`?

src/inmanta/util/__init__.py
```diff
@@ -119,15 +119,15 @@ def ensure_future_and_handle_exception(
...
         try:
-            exc = future.exception()
+            exc = task.exception()
             if exc is not None:
                 logger.exception(msg, exc_info=exc)
         except CancelledError:
             pass
         finally:
-            notify_done_callback(future)
+            notify_done_callback(task)
```
sanderr, line 130: This is tough to review. The typing in `asyncio.tasks` and `asyncio.futures` looks incomplete and even somewhat inconsistent to me, but as far as I can tell this is accurate. Just for the sake of clarity, could you add `future: Task[T]` where we call `ensure_future`?

src/inmanta/util/__init__.py
```diff
@@ -584,7 +584,7 @@ def is_stopped(self) -> bool:
     def is_running(self) -> bool:
         return not self._stopped
 
-    def add_background_task(self, future: Union[Future, Coroutine], cancel_on_stop: bool = True) -> Task:
+    def add_background_task(self, future: Union[Future[T], Coroutine[Any, Any, T]], cancel_on_stop: bool = True) -> Task[T]:
```
sanderr, line 587: Same here
```suggestion
    def add_background_task(self, future: Awaitable[T]], cancel_on_stop: bool = True) -> Task[T]:
```

src/inmanta/util/__init__.py
sanderr: And perhaps on the other lines where we call it as well?

stubs/typing_inspect.pyi
```diff
@@ -15,3 +15,5 @@ def is_typevar(tp: Any) -> bool: ...
 def is_classvar(tp: Any) -> bool: ...
 def is_new_type(tp: Any) -> bool: ...
 def is_forward_ref(tp: Any) -> bool: ...
+def get_args(tp: object, evaluate: bool = False) -> List[Type[object]]: ...
```
sanderr, line 18:
```suggestion
def get_args(tp: object, evaluate: bool = False) -> list[type[object]]: ...
```

sanderr (changes requested):
I'm adding some suggestions. You can consider most of them optional (they add a bit more typing still, but this PR is already a strict improvement as is). I'll make sure to mark anything I consider non-optional explicitly as such.

src/inmanta/config.py
```diff
@@ -50,10 +51,10 @@ def optionxform(self, name: str) -> str:
 class Config:
     __instance: Optional[ConfigParser] = None
     _config_dir: Optional[str] = None  # The directory this config was loaded from
-    __config_definition: dict[str, dict[str, "Option"]] = defaultdict(dict)
+    __config_definition: dict[str, dict[str, "Option[object]"]] = defaultdict(dict)
```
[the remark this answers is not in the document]
bartv: Indeed. Having less warnings without making the typing worse should be the first goal. Later we can make it better.

src/inmanta/deploy.py
```diff
@@ -386,13 +386,14 @@ def deploy(self, dry_run: bool, report: bool = True) -> None:
 
         else:
             result = self._check_result(self._client.dryrun_request(tid=self._environment_id, id=version))
+            assert result.result
```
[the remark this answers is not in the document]
bartv: I think we need to do what we discussed some time ago and add an equivalent of this method: https://requests.readthedocs.io/en/latest/api/#requests.Response.raise_for_status

Check for non 200 and that we get json from the server.

src/inmanta/module.py
bartv: It removed a huge amount of warnings from the output. I only see now that I should have used TModuleMetadata instead of  ModuleMetadata. However, it did reduce the amount of errors a lot

src/inmanta/moduletool.py
bartv: is the vendored version an actual string?

src/inmanta/moduletool.py
bartv: because I needed something else than module which is a method argument of type string. I can use module_object if you prefer?

The parameter should have been module_name but that is a potential breaking change

bartv: image

src/inmanta/server/services/compilerservice.py
bartv: I None valid? because that is the instance and not the type. But maybe that is no longer required.

src/inmanta/server/services/orchestrationservice.py
bartv: damn, and this time the author is no who you suspect first ;-)

src/inmanta/util/__init__.py
bartv: good idea, this is something that changed after we wrote the first version of this. Now this returns a task

bartv:
> I'm adding some suggestions. You can consider most of them optional (they add a bit more typing still, but this PR is already a strict improvement as is). I'll make sure to mark anything I consider non-optional explicitly as such.

Feel free to make changes, anything that improves typing and reduces warnings/errors I welcome

src/inmanta/server/services/compilerservice.py
sanderr: You write it all the time in function return types. Afaik `type[None]` is the type for `type(None)` (the type), while these tasks just return `None` (the instance).

src/inmanta/deploy.py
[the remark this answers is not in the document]
sanderr: Perhaps, but until we do this is strange, isn't it? We call a method, then assert an invariant that is never checked by the method?

src/inmanta/module.py
sanderr: Ok, then we can let it slide. I'm pretty sure it's not entirely correct, but perhaps it's a step in the right direction. To be honest, I'd really like it if `mypy` would no longer assume `Any` for missing generic parameters. Sometimes you just don't want to be as verbose.

src/inmanta/moduletool.py
sanderr: It's a `Version`-like object, just not of the type `packaging.Version`.

bartv: but then we need to at least change something, for example use parse_version everywhere. because now we are mixing strings and version objects

sanderr: My change request was a bit ambiguous. I did not mean to say we should not make the return type uniform. We just shouldn't pretend like `parse_version` returns a `Version` and all is well, so I don't like the `cast` here.

When I wrote this I didn't realize this lived in relatively new code. In this instance I think it's actually safe to just use `packaging.version.Version` directly. Then we get rid of the deprecated method and the type errors both.

bartv: you kind of lost me, can you make the changes you propose?

src/inmanta/util/__init__.py
sanderr: > Now this returns a task

It's not entirely clear to me to be honest:

```python
_FT = TypeVar("_FT", bound=Future[Any])

@overload
def ensure_future(coro_or_future: _FT, *, loop: AbstractEventLoop | None = None) -> _FT: ...  # type: ignore[overload-overlap]
@overload
def ensure_future(coro_or_future: Awaitable[_T], *, loop: AbstractEventLoop | None = None) -> Task[_T]: ...

class Task(Future[_T_co]):
```

Technically I think that if you pass it a non-task `Future`, you get the same type of `Future` back. Any other awaitable (including `Task`) will result in a `Task`. So it may actually be a non-task future, but only if `action` is a non-task future, which can't really be expressed in the typing.

All this said, I think typing it as `Task` is a definite step forward, even if there are some edge cases where it might not (and I'm not even sure about this) be completely accurate.

src/inmanta/moduletool.py
sanderr: Done

sanderr:
```diff
diff --git a/src/inmanta/moduletool.py b/src/inmanta/moduletool.py
index ed831e5c..2fb3bac2 100644
--- a/src/inmanta/moduletool.py
+++ b/src/inmanta/moduletool.py
@@ -183,7 +183,7 @@ class ModuleLikeTool:
             if not dev:
                 outversion = baseversion
             else:
-                outversion = cast(Version, parse_version("%s.dev%d" % (baseversion, time.time())))
+                outversion = Version("%s.dev%d" % (baseversion, time.time()))
         else:
             opts = [x for x in [major, minor, patch] if x]
             if version is not None:
@@ -207,9 +207,9 @@ class ModuleLikeTool:
                     outversion = str(VersionOperation.set_version_tag(old_version, version_tag=""))

             if dev:
-                outversion = cast(Version, parse_version("%s.dev%d" % (outversion, time.time())))
+                outversion = Version("%s.dev%d" % (outversion, time.time()))

-        outversion = cast(Version, parse_version(outversion))
+        outversion = Version(outversion)
         if outversion <= old_version:
             LOGGER.error(f"new versions ({outversion}) is not larger then old version ({old_version}), aborting")
             return None
```

src/inmanta/util/__init__.py
bartv: I based myself on the documentation:

https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future

sanderr: That's the same, isn't it? The first bullet point might result in a non-task `Future`.

src/inmanta/deploy.py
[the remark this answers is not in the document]
bartv: done

src/inmanta/server/services/compilerservice.py
bartv: fixed

src/inmanta/util/__init__.py
bartv: indeed. What do we do here? Do we type it as `Task | Future`? We can also try to type the method as `Task[T]` so that we never pass a future into it?

bartv: I did a quick check and according to mypy we only pass a coroutine. So a solution might be to type it as: `Coroutine[Any, Any, T]` instead of `Awaitable[T]`

I think there are a number of options:

1. We type it correct that the return type if `Task[T] | Future[T]` when the argument is `Awaitable[T]`. I am not a big fan of this because we never use a future (as far as I can tell).
2. We keep the typing like it is now
3. We keep the typing like it is but add an assert `assert not asyncio.isfuture(future)` (with a comment because the naming sucks). This is risky because we create a runtime failure where we shouldn't (we can also wrap it in an if typingchecking)
4. We change the type of the method to `Coroutine[Any, Any, T]` instead of `Awaitable[T]` In that case ensure_future will always return a Task

I seem to lean towards 4 because that means we can also start moving to https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task and use task groups for cancellation.

For 4 I would move it to another PR though.

sanderr: 4 sounds ok to me. Though I'd think that we'd want to use `object` instead of `Any`?

bartv: I used any because I was not sure if object is correct here. The types are not relevant for us. I can try and see what it does, I am just a bit concerned that if we pass a coroutine from an external library. But everything is a subclass of object so it should work.

sanderr: Perhaps `Coroutine[object, Never, T]`? (`Couritine` is `Generic[YieldType, SendType, ReturnType]`) so I suspect the first is covariant while the second is contravariant.

sanderr: Confirmed: https://github.com/python/typeshed/blob/994d22bfbb542956be24580323d5b4aeef822c77/stdlib/typing.pyi#L406

bartv: Why would you use Never? It might be that the code that wants to background the coroutine does need a send type

sanderr: How so? We used plain `Awaitable` before so we're clearly not sending anything.

bartv: Coroutine is a subclass of Awaitable. Say we use a coroutine in a service to which we send something and then we background it. In utils we are not sending something, but the type would not be Coroutine[object, Never, T].

sanderr: Except that the inferred type for `async`-based coroutines seems to use `Any`, so we really have no way to know... The pains of gradual typing...

bartv: I now have a change that uses Coroutine[object, object, T], would that work for you?

sanderr: `asyncio` does send things: https://github.com/python/cpython/blob/2afc7182e66635b3ec7efb59d2a6c18a7ad1f215/Lib/asyncio/tasks.py#L301

But as far as I can tell it only sends `None`, so we could make it `Couroutine[object, None, T]`?

bartv: ok, I will try that. We can always change it later if it does not work out

sanderr: > would that work for you?

I'm afraid that's always too restrictive. When I proposed it I didn't take the contra-variance into account. It would require all coroutines to be able to handle any kind of object you send them. I have no clue if they do.

bartv: so that means any again?

src/inmanta/moduletool.py
sanderr: I would then keep the part where you make this method consistent by not returning strings, but I'd not use the `cast` so that we don't cover up the lie in the return type.

bartv: I think I fixed it by making sure we never use the pkg_resource parse_version anymore

src/inmanta/config.py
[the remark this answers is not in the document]
bartv: I had to remove it because it introduced 1 error

src/inmanta/server/services/resourceservice.py
```diff
@@ -1135,29 +1135,28 @@ def _get_query_params(
...
+            return {param_key: param_value for param_key, param_value in query_params.items() if param_value is not None}
 
         if limit and resource_action_dtos:
             base_url = "/api/v2/resource_actions"
             common_query_params = _get_query_params(resource_type, agent, attribute, attribute_value, log_severity, limit)
             # Next is always earlier with regards to 'started' time
             next_params = common_query_params.copy()
-            next_params["last_timestamp"] = resource_action_dtos[-1].started
-            next_params["action_id"] = resource_action_dtos[-1].action_id
+            next_params["last_timestamp"] = resource_action_dtos[-1].started.isoformat(timespec="microseconds")
```
bartv, line 1154: @sanderr these gave warnings which I think were proper errors. I tried to fix them but it would be good if you double check this

Before this was an implicit str() which I think gives the wrong timestamp

sanderr: I think so, but I'm not very familiar with this flow. But from the timestamp perspective, yes. It might be better to call `inmanta.util.datetime_iso_format` though.

sanderr (approved):
Added one last comment but nothing blocking for me.

src/inmanta/server/services/resourceservice.py
bartv: done

PR #7107, opened 2024-02-01
poc: process and venv manager

FloLey:
# Description

Add the executor manager and venv manager.

# Self Check:

- [x] Changelog entry


src/inmanta/agent/agent.py
```diff
@@ -1526,3 +1527,133 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+        await loop.run_in_executor(self.thread_pool, install_for_config)
+
+
+class VirtualEnvironmentManager:
+    def __init__(self):
+        self._environment_map: dict[EnvBlueprint, ExecutorVirtualEnvironment] = {}
+
+    def create_env_storage(self, storage: str) -> str:
+        envs_dir = storage
+        new_env_dir = os.path.join(envs_dir, str(uuid.uuid4()))
```
wouterdb, line 1563: This will make it impossible to continue using the venv after restart?

src/inmanta/agent/agent.py
```diff
@@ -1526,3 +1527,133 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+
+
+class ExecutorManager:
+    def __init__(self, agent: Agent, environment_manager: VirtualEnvironmentManager):
+        self.executor_map: dict[ExecutorId, Executor] = {}
+        self.environment_manager = environment_manager
+        self.agent = agent  # for now the executorManager lives in an agent
+        self.storage = self.create_storage()
+
+    async def create_executor(self, agent_name: str, config_model_version: int, blueprint: EnvBlueprint) -> Executor:
```
wouterdb, line 1611: I don't think we want to restart the executor for every version, only if the actual code changes.

src/inmanta/agent/agent.py
```diff
@@ -1526,3 +1527,133 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+        executor_id = ExecutorId(agent_name, config_model_version)
+        venv = await self.environment_manager.get_environment(blueprint, self.storage["envs"], self.agent.thread_pool)
+        executor = Executor(executor_id, venv, self.storage["codes"])
+        self.executor_map[executor_id] = executor
+        return executor
+
+    async def get_executor(self, agent_name: str, config_model_version: int, blueprint: EnvBlueprint) -> Executor:
+        executor_id = ExecutorId(agent_name, config_model_version)
```
wouterdb, line 1619: the blueprint has to be part of the executor id as well

src/inmanta/agent/agent.py
```diff
@@ -1526,3 +1527,133 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+    async def execute(
+        self,
+        agent_name: str,
+        config_model_version: int,
+        blueprint: EnvBlueprint,
+        sources: Sequence["ModuleSource"],
+        resources: list[Resource],
+    ):
+        executor = await self.get_executor(agent_name, config_model_version, blueprint)
+        executor.load_code(sources, self.agent.thread_pool)
```
wouterdb, line 1633: I think code loading should only happen on creation of the executor

src/inmanta/agent/agent.py
```diff
@@ -1526,3 +1527,149 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+
+@dataclasses.dataclass(frozen=True)
+class ExecutorId:
+    agent_name: str
+    blueprint: ExecutorBlueprint
+
+
+class ExecutorVirtualEnvironment:
+    def __init__(self, storage: str, threadpool: ThreadPoolExecutor):
+        self.env = env.VirtualEnv(storage)
```
wouterdb, line 1551: I think you should use 'PythonEnvironment' 

This one is a child of `ActiveEnv` which it clearly is not. 

I would see how difficult it is to lift the required functionality from `VirtualEnv` to `ActiveEnv`

src/inmanta/env.py
```diff
@@ -1270,8 +1270,9 @@ def install_for_config(
         constraint_files: Optional[list[str]] = None,
         upgrade_strategy: PipUpgradeStrategy = PipUpgradeStrategy.ONLY_IF_NEEDED,
         paths: list[LocalPackagePath] = [],
+        should_be_active: bool = True,
```
wouterdb, line 1273: I would try to avoid this, as this is clearly a child of ActiveEnv

src/inmanta/agent/agent.py
```diff
@@ -1533,3 +1535,240 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+    async def create_executor(self, agent_name: str, blueprint: ExecutorBlueprint) -> Executor:
+        """
+        Creates an Executor based with the specified agent name and blueprint.
+        It ensures the required virtual environment is prepared and source code is loaded.
+
+        :param agent_name: The name of the agent for which the Executor is being created.
+        :param blueprint: The ExecutorBlueprint defining the configuration for the Executor.
+        :return: An Executor instance
+        """
+        executor_id = ExecutorId(agent_name, blueprint)
```
Hugo-Inmanta, line 1722: I think we can avoid re-creating this `executor_id` if we pass in the one created in `get_executor` to this method

Hugo-Inmanta (approved):

arnaudsjs (changes requested):

src/inmanta/agent/agent.py
```diff
@@ -1533,3 +1535,240 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+        Creates a new virtual environment based on the provided blueprint or reuses an existing one if suitable.
+        This involves setting up the virtual environment and installing any required packages as specified in the blueprint.
+
+        :param blueprint: The blueprint specifying the configuration for the new virtual environment.
+        :param threadpool: A ThreadPoolExecutor
+        :return: An instance of ExecutorVirtualEnvironment representing the created or reused environment.
+
+        TODO: Improve handling of bad venv scenarios, such as when the folder exists but is empty or corrupted.
+        """
+        env_storage, is_new = self.create_env_storage(blueprint)
```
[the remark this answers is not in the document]
arnaudsjs: (Feel free to ignore if you don't agree)

src/inmanta/agent/agent.py
```diff
@@ -1533,3 +1535,240 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+    async def create_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPoolExecutor) -> ExecutorVirtualEnvironment:
+        """
+        Creates a new virtual environment based on the provided blueprint or reuses an existing one if suitable.
+        This involves setting up the virtual environment and installing any required packages as specified in the blueprint.
+
+        :param blueprint: The blueprint specifying the configuration for the new virtual environment.
+        :param threadpool: A ThreadPoolExecutor
+        :return: An instance of ExecutorVirtualEnvironment representing the created or reused environment.
+
+        TODO: Improve handling of bad venv scenarios, such as when the folder exists but is empty or corrupted.
```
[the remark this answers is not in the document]
FloLey: https://github.com/inmanta/inmanta-core/pull/7107#discussion_r1495692235

src/inmanta/agent/agent.py
```diff
@@ -1526,3 +1527,149 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+        else:
+            LOGGER.info("Found existing venv for blueprint")
+            return new_env_dir, False  # Returning the path and False for existing directory
+
+    async def create_environment(
+        self, blueprint: EnvBlueprint, storage: str, threadpool: ThreadPoolExecutor
+    ) -> ExecutorVirtualEnvironment:
+        # create a new storage location for the new environment or reuse an existing one
+        env_storage, is_new = self.create_env_storage(storage, blueprint)
+        process_environment = ExecutorVirtualEnvironment(env_storage, threadpool)
```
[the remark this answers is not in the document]
arnaudsjs: What do you mean with "later" here? If later means after this is merged, then I would create a ticket instead.

src/inmanta/agent/agent.py
```diff
@@ -1533,3 +1535,240 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+
+    def create_env_storage(self, blueprint: EnvBlueprint) -> tuple[str, bool]:
+        """
+        Determines the storage path for a virtual environment based on its blueprint.
+        It either identifies an existing directory for the environment or creates a new one.
+
+        :param blueprint: The blueprint of the environment for which the storage is being determined.
+        :return: A tuple containing the path to the directory and a boolean indicating whether the directory was newly created.
+        """
+        hashed_blueprint_name: int = hash(blueprint)
```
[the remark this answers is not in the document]
FloLey: I've added a test for the first part but I don't know how to test the second scenario. It should be ok now as i'm using the md5 hash which doens't use a seed.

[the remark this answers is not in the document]
arnaudsjs: If you need to test the behavior across interpreter session, then there is not other alternative than spawning a new Python subprocess I think. The behavior of this hash is very crucial for the correct behavior of this code, so I don't think we can afford not to test this.

src/inmanta/agent/agent.py
```diff
@@ -1533,3 +1535,240 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+        Retrieves an existing virtual environment that matches the given blueprint or creates a new one if no match is found.
+
+        :param blueprint: The blueprint for which a matching environment is sought.
+        :param threadpool: A ThreadPoolExecutor
+        :return: An ExecutorVirtualEnvironment instance matching the blueprint.
+        """
+        assert type(blueprint) is EnvBlueprint, "Only EnvBlueprint instances are accepted, subclasses are not allowed."
+        if blueprint in self._environment_map:
+            return self._environment_map[blueprint]
+        return await self.create_environment(blueprint, threadpool)
```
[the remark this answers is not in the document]
FloLey: Indeed, I don't know how and when this should be handled
@wouterdb @sanderr

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
wouterdb: Oh, I hadn't considered the instability of `hash` good catch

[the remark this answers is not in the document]
sanderr: Afaik `hash` also is not guaranteed to be unique so we can not rely on it alone.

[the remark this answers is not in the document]
wouterdb: It is not guaranteed, but the odds of collision are 1/2**64 which I think is a chance we can take

src/inmanta/agent/agent.py
```diff
@@ -1533,3 +1537,257 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+    requirements: Sequence[str]
+
+    def generate_env_blueprint_hash(self) -> str:
+        """
+        Generate a stable hash for an EnvBlueprint instance by serializing its pip_config
+        and requirements in a sorted, consistent manner. This ensures that the hash value is
+        independent of the order of requirements and consistent across interpreter sessions.
+        """
+        # Convert the blueprint to a dictionary, ensuring a consistent ordering of keys
+        blueprint_dict = dataclasses.asdict(self)
```
wouterdb, line 1556: Why do this? you override both fields?

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
FloLey: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/7281

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
FloLey: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/7281

src/inmanta/agent/agent.py
FloLey: I changed it

src/inmanta/agent/agent.py
```diff
@@ -1533,3 +1537,253 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+        executor.execute(resources)
+
+    def create_storage(self) -> str:
+        """
+        Prepares and returns the path to the storage directory used by Executors for their source code.
+
+        :return: The path to the storage directory.
+        """
+        state_dir = cfg.state_dir.get()
+        code_dir = os.path.join(state_dir, "codes")
```
arnaudsjs, line 1787: Do we have tickets to cleanup the old venv/code directories?

src/inmanta/agent/agent.py
```diff
@@ -1533,3 +1537,253 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+        If an Executor does not exist for the given configuration, a new one is created.
+
+        :param agent_name: The name of the agent for which an Executor is being retrieved or created.
+        :param blueprint: The ExecutorBlueprint defining the configuration for the Executor.
+        :return: An Executor instance
+        """
+        executor_id = ExecutorId(agent_name, blueprint)
+        if executor_id in self.executor_map:
+            return self.executor_map[executor_id]
+        return await self.create_executor(executor_id)
```
arnaudsjs, line 1762: This method has the same race condition, right?

tests/test_agent.py
```diff
@@ -217,3 +224,286 @@ async def test_update_agent_map(server, environment, agent_factory):
...
+    # Generate hash in the current session for comparison
+    pip_config = PipConfig(**pip_config_dict)
+    current_session_blueprint = EnvBlueprint(pip_config=pip_config, requirements=requirements)
+    current_hash = current_session_blueprint.generate_env_blueprint_hash()
+
+    # Generate hash in a new interpreter session
+    result = subprocess.run(["python", "-c", python_code], capture_output=True, text=True)
+    new_session_hash = result.stdout.strip()
+
+    assert current_hash == new_session_hash, "Hash values should be consistent across interpreter sessions"
```
arnaudsjs, line 463: * We should check the exitcode here to verify whether the command executed successfully or not.
* We have to make sure that we print stderr on failure, because that's the stream that holds the stack trace.

tests/test_agent.py
```diff
@@ -217,3 +224,286 @@ async def test_update_agent_map(server, environment, agent_factory):
...
+    task_diff_1 = asyncio.create_task(manager.get_environment(blueprint2, None))
+
+    # Allow all get_environment tasks to proceed
+    creation_event.set()
+
+    # Wait for all tasks to complete
+    env_same_1, env_same_2, env_diff_1 = await asyncio.gather(task_same_1, task_same_2, task_diff_1)
+
+    assert env_same_1 is env_same_2, "Expected the same instance for the same blueprint"
+    assert env_same_1 is not env_diff_1, "Expected different instances for different blueprints"
```
arnaudsjs, line 509: To which extend is this code different from the following code?
```suggestion

    # Wait for all tasks to complete
    env_same_1, env_same_2, env_diff_1 = await asyncio.gather(
        manager.get_environment(blueprint1, None),
        manager.get_environment(blueprint1, None),
        manager.get_environment(blueprint2, None),
    )

    assert env_same_1 is env_same_2, "Expected the same instance for the same blueprint"
    assert env_same_1 is not env_diff_1, "Expected different instances for different blueprints"
```

arnaudsjs (changes requested):

src/inmanta/agent/agent.py
FloLey: https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/7284

tests/test_agent.py
FloLey: I find my approach makes it more clear that the 3 calls are started at the same time.

arnaudsjs (approved):

Issue #6732, opened 2023-11-15
Introduce env vars collection


edvgui:
Create a constant/class/anything in the stable api, which exposes the environment variables that the server uses when communicating with:
- The compiler
- The agent?

The idea is to make is safe, easy and sytable to load env vars in error reporting, making sure that we don't log anything too noisy (inmanta user env vars) or too sensitive (credentials).

The request originates from [here](https://github.com/inmanta/lsm/pull/665#discussion_r1392165174), where we try to load all lsm-related env vars for debug in case of some helper function mis-usage.

FloLey:
@edvgui this has been merged :)

PR #7137, opened 2024-02-07
Add inmanta_lsm.const.LSM_ENV_VARS to the Programmatic API reference

FloLey:
# Description

Add inmanta_lsm.const.LSM_ENV_VARS to the Programmatic API reference
closes https://github.com/inmanta/inmanta-core/issues/6732
# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -20,6 +20,12 @@ Constants
     :members:
     :undoc-members:
 
+.. data:: inmanta_lsm.const.LSM_ENV_VARS
+    :annotation: : Sequence[str]
+
+    This sequence contains a list of all environment variables passed to the compiler by inmanta-lsm
+
+
 
 .. _compiler-exceptions:
 
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -23,7 +23,7 @@ Constants
 .. data:: inmanta_lsm.const.LSM_ENV_VARS
     :annotation: : Sequence[str]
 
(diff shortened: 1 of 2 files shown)


arnaudsjs (changes requested):

docs/reference/api.rst
```diff
@@ -20,6 +20,12 @@ Constants
     :members:
     :undoc-members:
 
+.. data:: inmanta_lsm.const.LSM_ENV_VARS
```
arnaudsjs, line 23: I think you can automatically extract the docstring from `inmanta_lsm.const.LSM_ENV_VARS` using the `.. automodule::`. But I am not sure. Same for the type annotation.

FloLey: I tried some automodule and it didn't work, I already tried autodata and it didn't work neither. But shouldn't inmanta_lsm be imported somewhere for this to work?

docs/reference/api.rst
```diff
@@ -20,6 +20,14 @@ Constants
     :members:
     :undoc-members:
 
+.. only:: iso
```
FloLey, line 23: I tested this by changing it to `.. only:: oss ` to see if it was displayed when running `make docs html` locally (and it was). and then changing it to `.. only:: iso` (and then it was not). I don't know how to build the iso docs locally and I didn't find a jenkins job to which I could give my branch as parameter. I also didn't find anything about this in the internal docs.

arnaudsjs (approved):

PR #7138, opened 2024-02-07
issue/6938-fix-regression-optimized-python-file-collection 

FloLey:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/module.py
+++ b/src/inmanta/module.py
@@ -2958,20 +2958,17 @@ class Module(ModuleLike[TModuleMetadata], ABC):
         raise NotImplementedError()
 
     def _list_python_files(self, plugin_dir: str) -> list[str]:
-        """
-        Generate a list of all Python files in the given plugin directory.
-        This method prioritizes .pyc files over .py files, uses caching to avoid duplicate directory walks,
-        and only considers directories that are Python packages.
-        """
+        """Generate a list of all python files, including namespace packages and excluding the model directory."""
         # Return cached results if this directory has been processed before
+        model_dir_path = os.path.join(plugin_dir, "inmanta_plugins", self.name, "model")
         if plugin_dir in self._dir_cache:
             return self._dir_cache[plugin_dir]
 
         files: dict[str, str] = {}
 
         for dirpath, dirnames, filenames in os.walk(plugin_dir):
-            # Skip non-package directories (those without an __init__.py or __init__.pyc file)
-            if not any(fname for fname in filenames if fname in ["__init__.py", "__init__.pyc"]):
+            # Explicitly skip the model directory
+            if dirpath.startswith(model_dir_path):
                 continue
 
             # Skip this directory if it's already in the cache


src/inmanta/module.py
```diff
@@ -2958,20 +2958,17 @@ def get_plugin_dir(self) -> Optional[str]:
...
-        """
+        """Generate a list of all python files, including namespace packages and excluding the model directory."""
         # Return cached results if this directory has been processed before
+        model_dir_path: str = os.path.join(plugin_dir, "inmanta_plugins", self.name, "model")
         if plugin_dir in self._dir_cache:
             return self._dir_cache[plugin_dir]
 
         files: dict[str, str] = {}
 
         for dirpath, dirnames, filenames in os.walk(plugin_dir):
```
wouterdb, line 2969: I don't think we can afford to use `os.walk` here, as it doesn't allow us to stop scanning below the model dir. 

(there can be >10k files below it). 

I would look at the implementation of os.walk, steal the relevant parts and make is so that you can simply stop going down below the model folder.

src/inmanta/moduletool.py
```diff
@@ -1741,7 +1741,14 @@ def _move_data_files_into_namespace_package_dir(self, build_path: str) -> None:
         """
         Copy all files that have to be packaged into the Python package of the module
         """
-        python_pkg_dir = os.path.join(build_path, "inmanta_plugins", self._module.name)
+        python_pkg_dir: str = os.path.join(build_path, "inmanta_plugins", self._module.name)
+        model_dir: str = os.path.join(python_pkg_dir, "model")
+        if os.path.exists(model_dir):
+            raise ModuleBuildFailedError(
+                msg="There is already a model directory in %s"
```
wouterdb, line 1748: I would also mention that this is not supported.

wouterdb (changes requested):

src/inmanta/module.py
```diff
@@ -2958,22 +2957,30 @@ def get_plugin_dir(self) -> Optional[str]:
...
+            An os.walk that can stop descending into specific directories.
+
+            :param top: The root directory from which to start walking.
+            :param exclude_dirs: A list of paths of directories to stop descending into.
+            """
+            for dirpath, dirnames, filenames in os.walk(top, topdown=True):
+                # Check if the current directory matches any path in exclude_dirs
+                if any(dirpath.startswith(stop_dir) for stop_dir in exclude_dirs):
+                    # Modify dirnames in-place to stop os.walk from descending into any more subdirectories
+                    dirnames[:] = []
```
wouterdb, line 2979: I didn't know this. 

Looks nasty, but it is a documented feature

```
When topdown is true, the caller can modify the dirnames list in-place
    (e.g., via del or slice assignment), and walk will only recurse into the
    subdirectories whose names remain in dirnames; this can be used to prune the
    search, or to impose a specific order of visiting.  Modifying dirnames when
    topdown is false has no effect on the behavior of os.walk(), since the
    directories in dirnames have already been generated by the time dirnames
    itself is generated. No matter the value of topdown, the list of
    subdirectories is retrieved before the tuples for the directory and its
    subdirectories are generated.
    ```

File: src/inmanta/module.py [hunk 4]
```diff
@@ -2958,22 +2957,30 @@ def get_plugin_dir(self) -> Optional[str]:
...
+                    continue
+                yield dirpath, dirnames, filenames
```
wouterdb, line 2981: As this is on the critical path, I would avoid the double yield  and not wrap this in an extra method.

wouterdb (approved):
I still have one minor comment, but re-review is not required for me

src/inmanta/module.py
```diff
@@ -2958,20 +2957,18 @@ def get_plugin_dir(self) -> Optional[str]:
         raise NotImplementedError()
 
     def _list_python_files(self, plugin_dir: str) -> list[str]:
-        """
-        Generate a list of all Python files in the given plugin directory.
-        This method prioritizes .pyc files over .py files, uses caching to avoid duplicate directory walks,
```
sanderr, line 2963: Isn't this part still relevant?

src/inmanta/module.py
```diff
@@ -2958,20 +2957,18 @@ def get_plugin_dir(self) -> Optional[str]:
...
         files: dict[str, str] = {}
+        model_dir_path: str = os.path.join(plugin_dir, "inmanta_plugins", self.name, "model")
 
-        for dirpath, dirnames, filenames in os.walk(plugin_dir):
-            # Skip non-package directories (those without an __init__.py or __init__.pyc file)
-            if not any(fname for fname in filenames if fname in ["__init__.py", "__init__.pyc"]):
+        for dirpath, dirnames, filenames in os.walk(plugin_dir, topdown=True):
+            # Modify dirnames in-place to stop os.walk from descending into any more subdirectories of the model directory
+            if dirpath.startswith(model_dir_path):
+                dirnames[:] = []
```
sanderr, line 2971: Is this something you found in the `os` docs somewhere? It seems to rely on some implementation details of `os.walk`, doesn't it?

src/inmanta/moduletool.py
```diff
@@ -1741,7 +1741,15 @@ def _move_data_files_into_namespace_package_dir(self, build_path: str) -> None:
         """
         Copy all files that have to be packaged into the Python package of the module
         """
-        python_pkg_dir = os.path.join(build_path, "inmanta_plugins", self._module.name)
+        python_pkg_dir: str = os.path.join(build_path, "inmanta_plugins", self._module.name)
+        model_dir: str = os.path.join(python_pkg_dir, "model")
+        if os.path.exists(model_dir):
+            raise ModuleBuildFailedError(
+                msg="There is already a model directory in %s, which is not supported. "
+                "Please remove the model directory from this location."
```
sanderr, line 1749: Perhaps a bit more solution-oriented:
```suggestion
                msg="There is already a model directory in %s. The `inmanta_plugins.TODO.model` package is reserved for bundling the inmanta model files."
                "Please use a different name for this Python package."
```

Wdyt? Note the TODO

sanderr (changes requested):

src/inmanta/module.py
FloLey: yes:
```
When topdown is true, the caller can modify the dirnames list in-place
    (e.g., via del or slice assignment), and walk will only recurse into the
    subdirectories whose names remain in dirnames; this can be used to prune the
    search, or to impose a specific order of visiting.  Modifying dirnames when
    topdown is false has no effect on the behavior of os.walk(), since the
    directories in dirnames have already been generated by the time dirnames
    itself is generated. No matter the value of topdown, the list of
    subdirectories is retrieved before the tuples for the directory and its
    subdirectories are generated.
```

sanderr: Ok, I don't love it but that's on their side so nothing to do about it.

sanderr (approved):

PR #7147, opened 2024-02-08
Remove legacy table migration

### LINKED ISSUE #7117: drop legacy version handling from db schema
Pre-iso4, we used a different db schema version scheme. `inmanta.data.schema` still has methods to interpret and convert this legacy scheme. I believe these methods are now redundant, since all versions `< iso6` are no longer supported, therefore there are no more upgrade scenarios from the legacy version to the new ones (except perhaps in some of the older db schema migration test cases).

iso7 + iso8

wouterdb:
# Description

remove legacy schema update 

closes #7117 

note for meeting: tiny

# Self Check:

- [X] Attached issue to pull request
- [X] Changelog entry
- [X] Type annotations are present
- [] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )


arnaudsjs (changes requested):
Shouldn't we also going to remove the `legacy_version` column from the `schemamanager` table? It serves no purpose anymore, does it?

Hugo-Inmanta (approved):

PR #7148, opened 2024-02-08
issue/6096-log-which-pip-index-no-index

FloLey:
# Description

[Add the used pip indexes to the PackageNotFound exception in run_pip](https://github.com/inmanta/inmanta-core/pull/7064) didn't take --no-index into account. This Pr fixes this.

part of https://github.com/inmanta/inmanta-core/issues/6096

# Self Check:

- [x] Changelog entry


FloLey:
the merge with iso6 will fail again. I'll open a pr for that one once this is merged.

tests/test_env.py
```diff
@@ -288,6 +288,36 @@ def test_process_env_install_from_index_not_found_env_var(
...
+    we should not be using the system config and a path needs to be specified.
+    Assert the appropriate error is raised.
+    """
+
+    # Setup a temporary directory as project_dir
+    with tempfile.TemporaryDirectory() as temp_dir:
+        project_dir = temp_dir
+        setup_py_content = """
+from setuptools import setup
+setup()
```
arnaudsjs, line 304: A Python package without a name feels a bit weird to me.
```suggestion
setup(name="test")
```

tests/test_env.py
```diff
@@ -288,6 +288,36 @@ def test_process_env_install_from_index_not_found_env_var(
...
+def test_process_env_install_no_index(modules_v2_dir) -> None:
+    """
+    Attempt to install a package that does not exist with --no-index.
+    To have --no-index set in the pip cmd, the config should not contain an index_url,
+    we should not be using the system config and a path needs to be specified.
+    Assert the appropriate error is raised.
+    """
+
+    # Setup a temporary directory as project_dir
+    with tempfile.TemporaryDirectory() as temp_dir:
```
arnaudsjs, line 300: Why not rely on the tmpdir fixture?

tests/test_env.py
```diff
@@ -288,6 +288,36 @@ def test_process_env_install_from_index_not_found_env_var(
         )
 
 
+def test_process_env_install_no_index(modules_v2_dir) -> None:
```
arnaudsjs, line 291: Why use the `modules_v2_dir` fixture here?

src/inmanta/env.py
```diff
@@ -508,6 +508,7 @@ def create_log_content_files(title: str, files: list[str]) -> list[str]:
             not_found: list[str] = []
             conflicts: list[str] = []
             indexes: str = ""
+            no_index: bool = "--no-index" in cmd
```
arnaudsjs, line 511: It's also possible to set `--no-index` through an environment variable.

tests/test_env.py
FloLey: I forgot to remove it, I first tried using this instead of creating the package myself. I removed it now, thanks

src/inmanta/env.py
FloLey: Do you know where all the pip env vars are documented? I see in is mentioned at some places and when I try it it does work but I don't find any doc about it.

src/inmanta/util/__init__.py
```diff
@@ -78,6 +78,22 @@ def is_sub_dict(subdct: dict[PrimitiveTypes, PrimitiveTypes], dct: dict[Primitiv
...
+    """
+    val = val.lower()
+    if val in ("y", "yes", "t", "true", "on", "1"):
+        return True
+    elif val in ("n", "no", "f", "false", "off", "0"):
+        return False
+    else:
+        raise ValueError("invalid truth value %r" % (val,))
+
+
```
FloLey, line 96: this is the same implementation as in distutils  but I didn't like that it returned 1/0 instead of True/False

tests/test_env.py
```diff
@@ -288,6 +288,37 @@ def test_process_env_install_from_index_not_found_env_var(
         )
 
 
+@pytest.mark.parametrize_any("use_system_config", [True])
```
arnaudsjs, line 291:
```suggestion
@pytest.mark.parametrize_any("use_system_config", [True, False])
```

src/inmanta/env.py
```diff
@@ -508,6 +511,7 @@ def create_log_content_files(title: str, files: list[str]) -> list[str]:
             not_found: list[str] = []
             conflicts: list[str] = []
             indexes: str = ""
+            no_index: bool = "--no-index" in cmd or strtobool(env.get("PIP_NO_INDEX", "false").lower())
```
arnaudsjs, line 514: The `strtobool()` method already does the conversion to `lower()`.
```suggestion
            no_index: bool = "--no-index" in cmd or strtobool(env.get("PIP_NO_INDEX", "false"))
```

arnaudsjs, line 514: We have to move this code into the `if not_found` block. Otherwise this code can raise a `ValueError` if the value of the environment variable is invalid. Once we enter the `if non_found` block, we know that the input was not rejected by pip and we know that there will be no `ValueError`. Let's also add a comment about this in the code.

src/inmanta/util/__init__.py
```diff
@@ -78,6 +78,22 @@ def is_sub_dict(subdct: dict[PrimitiveTypes, PrimitiveTypes], dct: dict[Primitiv
...
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
+    val = val.lower()
+    if val in ("y", "yes", "t", "true", "on", "1"):
+        return True
+    elif val in ("n", "no", "f", "false", "off", "0"):
+        return False
+    else:
+        raise ValueError("invalid truth value %r" % (val,))
```
arnaudsjs, line 94: I am not sure to which extend we should consider this code to be copied and we should make sure we are in-line with the copy-right obligations of distutils. @wouterdb do you know?

src/inmanta/env.py
```diff
@@ -455,6 +456,8 @@ def run_pip_install_command_from_config(
                 del sub_env["PIP_INDEX_URL"]
             if "PIP_PRE" in sub_env:
                 del sub_env["PIP_PRE"]
+            if "PIP_NO_INDEX" in sub_env:
+                del sub_env["PIP_PRE"]
```
arnaudsjs, line 460: Nice catch! As this is a bugfix, I think we need to add a bugfix section for this to the changelog.

arnaudsjs (changes requested):

src/inmanta/env.py
arnaudsjs: I think every cli option can be set using an environment variable by prefixing the option with `PIP_`, capitalizing all the characters and replacing all the `-` chars with `_`.

arnaudsjs: https://pip.pypa.io/en/stable/topics/configuration/#environment-variables

src/inmanta/env.py
arnaudsjs: I think this was actually intentional. We only ignore the environment variables for things that can be set via the PipConfig. `--no-index` cannot be set using the `PipConfig` so we don't remove the environment variable when `use_system_config` is False.

FloLey: As discussed this will stay and I'll add a bugfix section

src/inmanta/util/__init__.py
FloLey: I don't think it is really needed by I added the license just in case.

src/inmanta/env.py
wouterdb:
```suggestion
            if "PIP_NO_INDEX" in sub_env:
                del sub_env["PIP_NO_INDEX"]
```

src/inmanta/env.py
```diff
@@ -520,10 +523,13 @@ def create_log_content_files(title: str, files: list[str]) -> list[str]:
                 if "Looking in indexes:" in line:
```
wouterdb, line 523:
```suggestion
                # This is not printed when not using any index or when only using PyPi
                if "Looking in indexes:" in line:
```

src/inmanta/util/__init__.py
```diff
@@ -78,6 +78,43 @@ def is_sub_dict(subdct: dict[PrimitiveTypes, PrimitiveTypes], dct: dict[Primitiv
     return not any(True for k, v in subdct.items() if k not in dct or dct[k] != v)
 
 
+def strtobool(val: str) -> bool:
```
wouterdb, line 81: 1. I don't want to even think about the implication of including a piece of licensed code for this little gain
2. This is not correct wrt the behavior of pip https://pip.pypa.io/en/stable/topics/configuration/#environment-variables

wouterdb (changes requested):

wouterdb (approved):

arnaudsjs (approved):

PR #7151, opened 2024-02-08
Removed get siblings

### LINKED ISSUE #7040: improve performance of SourceInfo.get_siblings
When running on larger modules, it seems that `SourceInfo.get_siblings ` in the code upload service causes significant load

image

We should investigate if this code can be made more efficient (I suspect we now search the same siblings over and over again)

wouterdb:
# Description

Remove getsiblings 

closes #7040 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/loader.py
+++ b/src/inmanta/loader.py
@@ -49,6 +49,17 @@ PLUGIN_DIR = "plugins"
 LOGGER = logging.getLogger(__name__)
 
 
+def get_inmanta_module_name(python_module_name: str) -> str:
+    """Small utility to convert python module into inmanta module"""
+    module_parts = python_module_name.split(".")
+    if module_parts[0] != const.PLUGINS_PACKAGE:
+        raise Exception(
+            "All instances from which the source is loaded, should be defined in the inmanta plugins package. "
+            "%s does not match" % python_module_name
+        )
+    return module_parts[1]
+
+
 class SourceNotFoundException(Exception):
     """This exception is raised when the source of the provided type is not found"""
 
@@ -87,14 +98,7 @@ class SourceInfo:
 
     def _get_module_name(self) -> str:
         """Get the name of the inmanta module, derived from the python module name"""
-        module_parts = self.module_name.split(".")
-        if module_parts[0] != const.PLUGINS_PACKAGE:
-            raise Exception(
-                "All instances from which the source is loaded, should be defined in the inmanta plugins package. "
-                "%s does not match" % self.module_name
-            )
-
-        return module_parts[1]
+        return get_inmanta_module_name(self.module_name)
 
     @property
     def requires(self) -> list[str]:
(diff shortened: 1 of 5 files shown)


src/inmanta/module.py
```diff
@@ -2778,6 +2778,7 @@ def __init__(self, project: Optional[Project], path: str) -> None:
         self._ast_cache: dict[str, tuple[list[Statement], BasicBlock]] = {}  # Cache for expensive method calls
         self._import_cache: dict[str, list[DefineImport]] = {}  # Cache for expensive method calls
         self._dir_cache: Dict[str, list[str]] = {}  # Cache containing all the filepaths present in a dir
+        self._plugin_file_cache: Optional[list[tuple[Path, ModuleName]]] = None
```
wouterdb, line 2781: this may not be strictly required

src/inmanta/module.py
```diff
@@ -3022,13 +3026,15 @@ def get_plugin_files(self) -> Iterator[tuple[Path, ModuleName]]:
...
+        self._plugin_file_cache = [
             (
                 Path(file_name),
                 ModuleName(self._get_fq_mod_name_for_py_file(file_name, plugin_dir, self.name)),
             )
             for file_name in self._list_python_files(plugin_dir)
-        )
+        ]
+
+        return iter(self._plugin_file_cache)
```
wouterdb, line 3037: I would love to change the signature of this method, but it is stable api, ...

src/inmanta/loader.py
```diff
@@ -140,16 +140,39 @@ def register_code(self, type_name: str, instance: object) -> None:
...
+        # get the module
+        def get_inmanta_module_name(instance: object) -> str:
+            module_name = instance.__module__
+            module_parts = module_name.split(".")
+            if module_parts[0] != const.PLUGINS_PACKAGE:
+                raise Exception(
+                    "All instances from which the source is loaded, should be defined in the inmanta plugins package. "
+                    "%s does not match" % module_name
+                )
+            return module_parts[1]
```
FloLey, line 152: couldn't we add this to utils so that we can start reusing it? this logic seems to be present in a lot of different places.

src/inmanta/module.py
FloLey: because we don't really need it to return the module name and it to be an iterator?

FloLey (approved):
questions but nothing that would block the merge

src/inmanta/module.py
Hugo-Inmanta: I'm failing to understand the lifecycle for this cache:
- At project creation all modules will get an empty cache
- after a compile this cache will be set
- during an export it will be read

But once it is set, it's never updated ? Won't we miss some things if for example we 
1 compile to set the cache
2 add some plugins code with dependencies
3 call project freeze to read the cache -> these newly added deps will be missing ?

Hugo-Inmanta (changes requested):

src/inmanta/module.py
wouterdb: no code can be added during the compile (i.e. the compiler starts fresh every time, so during the compile, is literally the time it takes to run the compile command)

src/inmanta/module.py
wouterdb: we don't need it to be an iterator mostly

src/inmanta/loader.py
wouterdb: I factored out the two I could find

Hugo-Inmanta (approved):

PR #7156, opened 2024-02-09
Covariant returntype

bartv:
# Description

This makes the typeparam in ReturnValue covariant so that the typing for handle is correct.

# Self Check:

- [ ] Changelog entry


wouterdb (approved):

src/inmanta/protocol/common.py
```diff
@@ -150,7 +150,7 @@ def from_dict(cls, value: JsonType) -> "Request":
         return req
 
 
-T = TypeVar("T", bound=Union[None, ArgumentTypes])
+T = TypeVar("T", bound=Union[None, ArgumentTypes], covariant=True)
```
sanderr, line 153: Not a strict change request, but I like the practice they use in the `typing` module to include the covariance in the name so you don't accidentally reuse this where you shouldn't.
```suggestion
T_co = TypeVar("T_co", bound=Union[None, ArgumentTypes], covariant=True)
```

reference:
```python
>>> import typing
>>> typing.T, typing.T_co, typing.T_contra
(~T, +T_co, -T_contra)
>>>
```

bartv: I like the idea.

PR #7162, opened 2024-02-12
issue/6096-log-which-pip-index-no-index-iso6

FloLey:
# Description
**Different than for iso7/master as the pipconfig did not exist in iso6**
[Add the used pip indexes to the PackageNotFound exception in run_pip](https://github.com/inmanta/inmanta-core/pull/7064) didn't take --no-index into account. This Pr fixes this.

part of https://github.com/inmanta/inmanta-core/issues/6096

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -47,6 +47,7 @@ from inmanta import const
 from inmanta.ast import CompilerException
 from inmanta.server.bootloader import InmantaBootloader
 from inmanta.stable_api import stable_api
+from inmanta.util import strtobool
 from packaging import version
 
 try:
@@ -497,6 +498,8 @@ class PythonEnvironment:
             del sub_env["PIP_EXTRA_INDEX_URL"]
         if index_urls is not None and "PIP_INDEX_URL" in sub_env:
             del sub_env["PIP_INDEX_URL"]
+        if index_urls is not None and "PIP_NO_INDEX" in sub_env:
+            del sub_env["PIP_NO_INDEX"]
 
         def create_log_content_files(title: str, files: list[str]) -> list[str]:
             """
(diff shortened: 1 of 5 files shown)


src/inmanta/env.py
```diff
@@ -547,10 +550,14 @@ def create_log_content_files(title: str, files: list[str]) -> list[str]:
                 if "versions have conflicting dependencies" in line:
                     conflicts.append(line)
                 # Get the indexes line from full_output
+                # This is not printed when not using any index or when only using PyPi
                 if "Looking in indexes:" in line:
                     indexes = line
             if not_found:
-                if indexes:
+                no_index: bool = "--no-index" in cmd or strtobool(sub_env.get("PIP_NO_INDEX", "false"))
```
arnaudsjs, line 557: 1. I now realize that this is actually incorrect. We are not taking into account how the `no-index` setting is configured in the configuration files. If the `--no-index` cli option was not provided and the `PIP_NO_INDEX` environment setting was not set, but the `no-index` config option is set to `true` in the configuraiton file, then error message will say `Packages %s were not found at PyPI.`, while it should say `Packages %s were not found. No indexes were used.`
2. Next to that I think we should make the error message `Packages %s were not found. No indexes were used.` more clear. Currently we only report that pip did not use any indexes, but we leave it entirely to the end-user to figure out why. The error message should hint towards the `no-index` option. Otherwise the end-user will have a hard time figuring out what is going on.

@wouterdb Do you have an opinion on these two points? Because you reviewed the previous PR as well.

arnaudsjs (changes requested):

src/inmanta/env.py
FloLey: Indeed but isn't this starting to become a little to "fine-grained"? I mean I liked to idea of helping the user by giving the info about the pip index that is used but I feel there are so much corner cases that we are trying to cover that are starting to add a lot of if conditions which I don't like. 
Can't we just do:
```
if indexes:
    msg = "Packages %s were not found in the given indexes. (%s)" % (", ".join(not_found), indexes)
else:
    msg = "Packages %s were not found. This could be because '--no-index' is set somewhere (config, env_vars) or because only the Pypi Index is used." % (", ".join(not_found))
```
?

arnaudsjs (approved):

Issue #7065, opened 2024-01-29
"NoneType object is not iterable" error on dryrun_request call


jptrindade:
When executing a `dryrun_request` with existing tid and id I got this error. When executing a dryrun using the orchestrator directly it seems to work.

```
remote_orchestrator.client.dryrun_request(tid=lsm_project.environment, id=111)
<inmanta.protocol.common.Result object at 0x7f04cd428950>
special variables
function variables
code:
500
result:
{'message': 'An unexpected error ...terable",)'}
special variables
function variables
'message':
'An unexpected error occurred in the server while processing the request: ("\'NoneType\' object is not iterable",)'
len():
1
set_result:
{'message': 'An unexpected error ...terable",)'}
_callback:
None
_result:
{'message': 'An unexpected error ...terable",)'}
```

```
2024-01-29 12:40:59,715 inmanta.protocol.rest    DEBUG   Calling method <bound method DyrunService.dryrun_request of <inmanta.server.services.dryrunservice.DyrunService object at 0x7fdc47661290>>(version_id='111', env='<inmant...')
2024-01-29 12:40:59,869 inmanta.protocol.rest.serverDEBUG   HTTP version of request: HTTP/1.1
2024-01-29 12:40:59,878 inmanta.protocol.rest    DEBUG   Calling method <bound method DyrunService.dryrun_update of <inmanta.server.services.dryrunservice.DyrunService object at 0x7fdc47661290>>(resource='netbox:...', changes='{}', dryrun_id='50b7b59...', env='<inmant...')
2024-01-29 12:40:59,915 tornado.access           INFO    200 PUT /api/v1/dryrun/50b7b591-47fb-47e0-a992-87170f2db500 (127.0.0.1) 53.25ms
2024-01-29 12:40:59,921 inmanta.protocol.rest    ERROR   An exception occured during the request.
                                                         Traceback (most recent call last):
                                                           File "/opt/inmanta/lib64/python3.11/site-packages/inmanta/protocol/rest/__init__.py", line 554, in _execute_call
                                                             result = await config.handler(**arguments.call_args)
                                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                           File "/opt/inmanta/lib64/python3.11/site-packages/inmanta/server/services/dryrunservice.py", line 70, in dryrun_request
                                                             dryrun = await self.create_dryrun(env, version_id, model)
[...]

PR #7164, opened 2024-02-12
[FIX] set undeployable to empty array instead of null

Hugo-Inmanta:
# Description

The put_partial endpoint was setting null values for undeployable and skipped_for_undeployable if both the previous' version set and the current partial compile's set were empty instead of empty arrays.

This was causing internal server errors when performing a dryrun after a put_partial call.

closes  https://github.com/inmanta/inmanta-core/issues/7065

# Self Check:

- [x] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -5290,7 +5290,7 @@ async def create_for_partial_compile(
                 $4,
                 $5,
                 (
-                    SELECT array_agg(rid)
+                    SELECT coalesce(array_agg(rid), ARRAY[]::character varying[])
```
arnaudsjs, line 5293: This makes sure that no new NULL values will be added to the database, but it doesn't fix the NULL values that already exist in the database. Is this intentional?

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
Hugo-Inmanta: No, nice catch. Should I add a migration to transform NULL values into empty arrays ?
Should the migration also add a not null constraint ?

arnaudsjs: > Should I add a migration to transform NULL values into empty arrays ?

Yes.

> Should the migration also add a not null constraint ?

Yes.

src/inmanta/db/versions/v202402130.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+async def update(connection: Connection) -> None:
+    schema = """
+    UPDATE public.configurationmodel
+    SET undeployable=(
+        CASE
+            WHEN undeployable IS NULL
+            -- Set default to empty array
+            THEN ARRAY[]::character varying[]
```
arnaudsjs, line 29: If I read this correctly, this contains first a constructor call, then a type cast and them a type definition. Do you know why this is required? The type of the `undeployable` is already known by PostgreSQL, no?

arnaudsjs (changes requested):

src/inmanta/db/versions/v202402130.py
Hugo-Inmanta: Right, I'll simplify this

Hugo-Inmanta: I don't think I see what you mean here:
> and them a type definition 

To me it is only a constructor `ARRAY[]` into a typecast of type array of `character varying`

And we cannot use only the constructor or else we get :
```
cannot determine type of empty array
               HINT:  Explicitly cast to the desired type, for example ARRAY[]::integer[].
```

Hugo-Inmanta: I'll go back to my inital implementation of `THEN '{}'`

arnaudsjs: I would expect the following:
```suggestion
            THEN ARRAY[]::varying[]
```
So that you have:

* `ARRAY[]` that creates a new array
* `::varying[]` which type casts the array to the correct type.

But the construct `ARRAY[]::character varying[]` I am unable to interpret. That doesn't mean it's incorrect of course. So that's why I am asking why it was done like this and how I should read it. It would also help if you could point me to the relevant PostgreSQL documentation for example.

I see you now changed it into `'{}'`. That's fine by me as well.

Hugo-Inmanta: I think I see,  `character varying` is an actual type itself (https://www.postgresql.org/docs/current/datatype-character.html#DATATYPE-CHARACTER) so these are equivalent :

```sql
THEN ARRAY[]::varchar[]
```

And
```sql
THEN ARRAY[]::character varying[]
 ```

arnaudsjs (approved):

PR #7179, opened 2024-02-13
Added cache to validate type

### LINKED ISSUE #7041: Cache validation_type in validate_type  method
The validation_type method is use to construct typedefs, as such, it is often called with the same validation_type.
We should cache the validation_type, as the current approach of re-constructing the same type adaptor is not efficient 

image

wouterdb:
# Description

closes #7041

# Self Check:

- [x] Changelog entry
```
src/inmanta/validation_type.py:106: error: Incompatible types in assignment (expression has type "object", variable has type "Mapping[str, object] | None")  [assignment]
```

overloaded typedefs causes issues 

```python
    @overload
    @classmethod
    def unwrap(cls, item: "DictProxy") -> Mapping[str, object]: ...

    @overload
    @classmethod
    def unwrap(cls, item: "SequenceProxy") -> Sequence[object]: ...

    @overload
    @classmethod
    def unwrap(cls, item: "None") -> NoneValue: ...

    # This one is not correct
    # @overload
    # @classmethod
    # def unwrap(cls, item: "DynamicProxy") -> "Instance": ...

    @overload
    @classmethod
    def unwrap(cls, item: object) -> object: ...

```


tests/compiler/test_loops.py
```diff
@@ -173,6 +173,14 @@ def test_resultcollector_receive_result_flatten(snippetcompiler) -> None:
                     assert.success = false
                 end
             end
+
```
wouterdb, line 176: @sanderr this testcase alerted me to some existing bugs in the execute_direct and as_constant

src/inmanta/ast/statements/assign.py
```diff
@@ -145,15 +145,22 @@ def _resolve(self, requires: dict[object, object], resolver: Resolver, queue: Qu
 
     def execute_direct(self, requires: abc.Mapping[str, object]) -> object:
         qlist = []
-
         for i in range(len(self.items)):
             value = self.items[i]
-            qlist.append(value.execute_direct(requires))
+            item = value.execute_direct(requires)
```
wouterdb, line 150: the execute_direct and as_constant did not do the proper flattening 

I also feel like we should merge the as_constant and direct_execute methods if at all possible, because both are so rarely used and similar it is hell to test them both

Hugo-Inmanta (approved):

src/inmanta/ast/statements/__init__.py
```diff
@@ -217,12 +217,28 @@ def requires(self) -> list[str]:
 
     def execute_direct(self, requires: abc.Mapping[str, object]) -> object:
```
wouterdb, line 218: @sanderr I'm more confused then every here, can you check this?

src/inmanta/ast/statements/__init__.py
```diff
@@ -217,12 +217,28 @@ def requires(self) -> list[str]:
...
         """
         raise DirectExecuteException(self, f"The statement {str(self)} can not be executed in this context")
 
+    def as_constant(self) -> object:
+        """
+        Returns this expression as a constant value, if possible. Otherwise, raise a RuntimeException.
+
+        It is very similar to self.execute_direct({}) with the difference that:
+         - This method can be called prior to normalization
+         - It is expected to not call plugins
```
sanderr, line 234:
```suggestion
         - It is expected to not have side effects (e.g. call plugins, construct entities, ....)
```
I'm a bit on the fence about the second example. I think it's important to keep in mind if we ever decide to extend the functionality. However, it might be interpreted to mean that `execute_direct` does allow constructing entities, which it doesn't (it's surface is a bit less well defined).

Interestingly, the fact that it executes without a context in effect prohibits it from doing any of these things, even by accident. But that's by happenstance more than by design.

PR #7181, opened 2024-02-14
[WIP] increase timeout in test_s_incremental_deploy_interrupts_full_d…

Hugo-Inmanta:
…eploy

This job: https://jenkins.inmanta.com/job/core/job/inmanta-core-postgresql-versions/job/iso7/
failed on the `test_s_incremental_deploy_interrupts_full_deploy` test a couple of times by taking slightly more than the timeout limit of 10s

# Self Check:

- [ ] Changelog entry


sanderr (approved):
I'm adding @wouterdb because I want to get some extra confidence that this is not a regression.

wouterdb (commented):
I don't feel comfortable increasing this timeout.

15 seconds for this is abnormally long. I would like to know where that time is spent. 

(1 second is already long imho)

PR #7194, opened 2024-02-15
Fix incompatiblity pytest asyncpg

arnaudsjs:
# Description

Due to a bugfix in pytest8, the `postgresql_proc_with_log` fixture now returns None instead of an empty string. This PR ensure that None values are correctly converted into an empty string.

Part of inmanta/inmanta-dev-dependencies#372

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/fix-incompatiblity-pytest-postgresql.yml
+++ b/changelogs/unreleased/fix-incompatiblity-pytest-postgresql.yml
@@ -1,0 +1,6 @@
+---
+description: "Fix incompatibilty with pytest-postgresql when using pytest8. Due to a bugfix in pytest8, the postgresql_proc_with_log fixture now returns None instead of an empty string."
+issue-nr: 372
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/fix-incompatiblity-pytest-postgresql.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: "Fix incompatibilty with pytest-postgresql when using pytest8. Due to a bugfix in pytest8, the postgresql_proc_with_log fixture now returns None instead of an empty string."
```
arnaudsjs, line 2:
```suggestion
description: "Fix incompatibility with pytest-postgresql when using pytest8. Due to a bugfix in pytest8, the postgresql_proc_with_log fixture now returns None instead of an empty string."
```

sanderr (approved):

PR #7199, opened 2024-02-15
add comments to code loading

FloLey:
# Description

add some comments to the code of the code loading

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -1334,16 +1334,8 @@ class Agent(SessionEndpoint):
 
     async def ensure_code(self, environment: uuid.UUID, version: int, resource_types: Sequence[str]) -> set[str]:
         """
-        Ensure that the code for the given environment and version is loaded by:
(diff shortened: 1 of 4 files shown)


src/inmanta/agent/agent.py
```diff
@@ -1333,7 +1333,18 @@ async def on_disconnect(self) -> None:
             agent_instance.pause("Connection to server lost")
 
     async def ensure_code(self, environment: uuid.UUID, version: int, resource_types: Sequence[str]) -> set[str]:
-        """Ensure that the code for the given environment and version is loaded"""
+        """
+        Ensure that the code for the given environment and version is loaded by:
+         - Storing the pipconfig once and reusing it for all the handlers that will be installed.
```
wouterdb, line 1338: I would add all of this (except for the first line), in between the statements as that makes it easier to read (a lot of it is already there)

wouterdb (changes requested):

wouterdb (approved):

PR #7200, opened 2024-02-15
Issue/prep parser

wouterdb:
Trying to rework the parser, I got it more or less in, but it required a lot of changes. 

I basically moved a lot of selectivity from the parser to the AST 

Some thing need to be re-enabled

### DIFF
--- a/src/inmanta/ast/variables.py
+++ b/src/inmanta/ast/variables.py
@@ -145,7 +145,6 @@ class Reference(ExpressionStatement):
             ref = Reference("self")
             self.copy_location(ref)
             attr_ref = AttributeReference(ref, self.locatable_name)
-            self.copy_location(attr_ref)
             return attr_ref
 
     def get_dataflow_node(self, graph: DataflowGraph) -> dataflow.AssignableNodeReference:
@@ -278,6 +277,7 @@ class AttributeReference(Reference):
 
         # a reference to the instance
         self.instance = instance
+        self.location = attribute.get_location()
 
     def requires(self) -> list[str]:
         return self.instance.requires()
--- a/src/inmanta/parser/plyInmantaParser.py
+++ b/src/inmanta/parser/plyInmantaParser.py
@@ -116,13 +116,16 @@ def expand_range(start: Range, end: Range) -> Range:
     return Range(start.file, start.lnr, start.start_char, end.end_lnr, end.end_char)
 
 
-def expand_range_or_location(start: Range|Location, end: Range|Location) -> Range|Location:
+def expand_range_or_location(start: Range | Location, end: Range | Location) -> Range | Location:
     """
     Returns a new range from the start of `start` to the end of `end`. Assumes both ranges are on the same file.
 
     If end is a location, we just return start
     """
     if not isinstance(end, Range):
+        return start
+
+    if not isinstance(start, Range):
         return start
     return Range(start.file, start.lnr, start.start_char, end.end_lnr, end.end_char)
 
@@ -217,7 +220,16 @@ def p_stmt(p: YaccProduction) -> None:
     """statement : assign
     | for
     | if
-    | expression empty"""
+    | boolean_expression
+    | constant
+    | function_call
+    | var_ref empty
+    | constructor
+    | map_def
+    | map_lookup empty
+    | index_lookup
+    | conditional_expression
+    | arithmetic_expression"""  # Can't do anything starting with `[` as it breaks dict lookup
     p[0] = p[1]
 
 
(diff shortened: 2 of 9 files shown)


src/inmanta/ast/variables.py
```diff
@@ -145,7 +145,6 @@ def root_in_self(self) -> "Reference":
             ref = Reference("self")
             self.copy_location(ref)
             attr_ref = AttributeReference(ref, self.locatable_name)
-            self.copy_location(attr_ref)
```
wouterdb, line 148: this is moved into the AST node itself

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -205,7 +220,16 @@ def p_stmt(p: YaccProduction) -> None:
...
+    | boolean_expression
+    | constant
+    | function_call
+    | var_ref empty
+    | constructor
+    | map_def
+    | map_lookup empty
+    | index_lookup
+    | conditional_expression
+    | arithmetic_expression"""  # Can't do anything starting with `[` as it breaks dict lookup
```
wouterdb, line 232: This is the nasty one: can't have expression starting with `[` 

because

```
a = b
['a']
```
can either be 

```
Assign('a', Reference('b'))
CreateLis([Literal('a')])
```
or
```
Assign('a', MapLookup(Reference('b'),'a'))
```

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -231,19 +255,19 @@ def p_stmt_list_empty(p: YaccProduction) -> None:
 
 
 def p_assign(p: YaccProduction) -> None:
-    "assign : var_ref '=' operand"
+    "assign : var_ref '=' expression"
```
wouterdb, line 258: I reduce the number of distinct types, to make the parser simpler

tests/compiler/test_strings.py
```diff
@@ -231,8 +231,8 @@ def test_fstring_relations(snippetcompiler, capsys):
 implement B using std::none
 implement C using std::none
 
-A.b [1] -- B [1]
-B.c [1] -- C [1]
+A.b [1] -- B
```
wouterdb, line 234: existing error, didn't mean what it looked like

src/inmanta/parser/plyInmantaParser.py
sanderr: Wasn't this also the case before? i.e. on master:
```
d = {"a": 42}

x = d
["a"]


std::print(x)
```

produces 42. It's just as ambiguous.

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -871,8 +884,14 @@ def p_index_lookup(p: YaccProduction) -> None:
...
 
 def p_short_index_lookup(p: YaccProduction) -> None:
-    "index_lookup : attr_ref '[' param_list ']'"
+    "index_lookup : expression '[' param_list ']'"
     attref = p[1]
+    if not isinstance(attref, AttributeReference):
+        raise ParserException(
+            attref.location,
+            attref,
+            f"Short index lookup must start with an attribute reference (`<expression>.<id>`), but got `{repr(attref)}`",
```
sanderr, line 893: I think you would get this when you try to do a list index operation? Perhaps this error message would just be more confusing to a user that thinks we support list indexing?

sanderr: Ah no, `param_list` means kwargs...

wouterdb: True, what would you say instead?

sanderr: I think my second comment invalidates the first, do you agree?

src/inmanta/parser/plyInmantaParser.py
```diff
@@ -1328,6 +1339,20 @@ def p_error(p: YaccProduction) -> None:
...
+    is_bad_relation = p.type == "." and parser.symstack[-3].type == "multi"
+    if is_bad_class_reference:
+        var: ExpressionStatement = parser.symstack[-2].value
+        location = var.get_location()
+        cid: LocatableString = p.value
+        raise InvalidNamespaceAccess(f"{repr(var)}.{cid}", expand_range_or_location(location, cid.location))
+    if is_bad_relation:
+        var = parser.symstack[-1].value
+        location = var.get_location()
+        raise InvalidNamespaceAccess(f"{repr(var)}", location)
```
arnaudsjs, line 1354: Could you elaborate which case this error handler covers?

arnaudsjs (changes requested):

PR #7201, opened 2024-02-15
Remove visible todos from lsm doc

### LINKED ISSUE #7134: LSM documentation landing page contains todos
remove these (either by filling them in or dropping these parts of the page)

https://docs.inmanta.com/inmanta-service-orchestrator/7/lsm/index.html

any todos that are simply dropped -> create follow-up tickets

FloLey:
# Description

add docs/ remove todos
follow-up ticket: 
1. https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/7202
2. https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/std/504
3. https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/apt/281

closes #7134 

# Self Check:

- [ ] Changelog entry


docs/lsm/index.rst
```diff
@@ -89,12 +123,19 @@
 
     The state machine attached to the lifecycle will determine whether the API call is successful or not.
 
-    TODO: Three set of attributes
+    Each service instance defines 3 sets of its attributes that are used in different situations:
+
+    - ``candidate_attributes``: The set of attributes used when the service instance model state is candidate or designed.
+    - ``active_attributes``: The set of attributes when the model is evaluated with only active instances.
```
arnaudsjs, line 129: This is not True. These attributes are also used when a validation compile is done for another service instance for example.

docs/lsm/index.rst
```diff
@@ -89,12 +123,19 @@
...
+    - ``rollback_attributes``: The set of attributes that can be used to rollback active attributes that result in an error
+                                state.
+
```
arnaudsjs, line 132:
```suggestion
    - ``rollback_attributes``: The set of attributes that can be used to rollback active attributes that result in an error state.

```

docs/lsm/index.rst
```diff
@@ -47,27 +47,61 @@
...
+            std::ipv_any_address router_ip
+            string router_name
+            string interface_name
+
+            std::ipv_any_interface address
+            lsm::attribute_modifier address__modifier="rw+"
+        end
+
+        implement InterfaceIPAssignment using parents
```
arnaudsjs, line 85: When we include a model in the docs, we usually add the model to a separate file. Then we can use that file to add a test case that verifies whether the model actually compiles and we can use the same file to include the model (or part of it) in the docs. We should do the same here. Otherwise the code included in the docs might no longer compile due to future changes to the lsm module or compiler.

arnaudsjs (changes requested):
* I see there are still some TODO's left in the following places:
    * https://docs.inmanta.com/inmanta-service-orchestrator/7/reference/modules/std.html#std.resources.DirectoryHandler
    * https://docs.inmanta.com/inmanta-service-orchestrator/7/reference/modules/apt.html#apt.AptPackage
* Regarding the follow up ticket: We should make sure to include a reference in that ticket to this PR. That way it's easier to determine what the different bullets in the ticket mean.

PR #7220, opened 2024-02-19
Added transaction finalizer

### LINKED ISSUE #6955: resource service triggers background tasks before transaction is committed
The orchestration service calls the resource service's `mark_deployed` method [in a transaction context](https://github.com/inmanta/inmanta-core/blob/7cc942b19550f8ab20206752e6516bc0a23f01de/src/inmanta/server/services/orchestrationservice.py#L1151C49-L1151C49). The latter passes it on to `resource_action_update`, which adds a background task [for `ConfigurationModel.mark_done_if_done`](https://github.com/inmanta/inmanta-core/blob/7cc942b19550f8ab20206752e6516bc0a23f01de/src/inmanta/server/services/resourceservice.py#L1000) and [for `agent.resource_event`](https://github.com/inmanta/inmanta-core/blob/7cc942b19550f8ab20206752e6516bc0a23f01de/src/inmanta/server/services/resourceservice.py#L1011). These background tasks are added before the transaction is committed.

Consequences (hypothesis):
1. `mark_done_if_done` inspects resource state. I believe it now races (and typically loses) with the transaction commit so that it never sees all resources as deployed. The configuration model is therefore never marked as deployed. This impacts e.g. `inmanta-cli version list`. An easy reproduction is to simply run the quickstart. Export a new version without changes to the resources, then run `inmanta-cli  --host 172.30.0.3  version list -e SR_Linux`. It will remain in `deploying` until a repair is triggered. The web console does show all resources as deployed.
2. Does `agent.resource_event` inspect any database state (through an API call)? Is it safe to be used in this transaction context? We may be sending an event for something that hasn't committed yet. When the agent responds to the event by querying the server for resource state, it may get the wrong state.
3. There may or may not be other flows that pass a connection object into this method.

I did not go more in-depth at this time, because it is not entirely clear to me what is the best way forward. The `mark_done_if_done` approach might be no longer relevant, in which case it might be better to drop it and use something else for `inmanta-cli version list` rather than to address the concurrency problem. More input is needed from the team.

wouterdb:
# Description

Added transaction finalizer 
fixes #6955 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/orchestrationservice.py
+++ b/src/inmanta/server/services/orchestrationservice.py
@@ -59,6 +59,7 @@ from inmanta.server.agentmanager import AgentManager, AutostartedAgentManager
 from inmanta.server.services.resourceservice import ResourceService
 from inmanta.server.validate_filter import InvalidFilter
 from inmanta.types import Apireturn, JsonType, PrimitiveTypes, ReturnTupple
+from inmanta.util.db import ConnectionInTransaction
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1067,93 +1068,97 @@ async def release_version(
...
-
-                # ensure we are the latest version
-                # this is required for the subsequent increment calculation to make sense
-                # this does introduce a race condition, with any OTHER release running concurrently on this environment
-                # We could lock the get_version_nr_latest_version for update to prevent this
-                if model.version < (latest_version or -1):
-                    raise Conflict(
-                        f"The version {version_id} on environment {env.id} "
-                        f"is older then the latest released version {latest_version}."
+                with ConnectionInTransaction(connection) as connection_holder:
```
wouterdb, line 1071: indented one level

src/inmanta/server/services/resourceservice.py
```diff
@@ -1005,23 +1006,33 @@ async def update_fields_resource(
...
-                aclient = self.agentmanager_service.get_agent_client(env.id, agent)
-                if aclient is not None:
-                    if change is None:
-                        change = const.Change.nochange
-                    self.add_background_task(
-                        aclient.resource_event(env.id, agent, resource_id, send_events, status, change, changes)
-                    )
+            def post_deploy_update() -> None:
+                assert model_version is not None  # mypy can't figure this out
+                # Make sure tasks are scheduled AFTER the tx is done
```
arnaudsjs, line 1016:
```suggestion
                assert model_version is not None  # mypy can't figure this out
                # Make sure tasks are scheduled AFTER the tx is done.
                # This method is only called if the transaction commits successfully.
```

arnaudsjs (changes requested):

src/inmanta/util/db.py
```diff
@@ -0,0 +1,85 @@
...
+        finalizer()
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
+        return None
+
```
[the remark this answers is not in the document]
wouterdb:
```suggestion

```

src/inmanta/util/db.py
```diff
@@ -0,0 +1,85 @@
...
+    def call_after_tx(self, finalizer: Callable[[], object]) -> None:
+        finalizer()
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
+        return None
```
wouterdb, line 61:
```suggestion
```

src/inmanta/util/db.py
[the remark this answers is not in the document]
wouterdb: removed

arnaudsjs (approved):

PR #7223, opened 2024-02-20
Issue/1641 rename module changelog

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/irt/pull/1917

# Self Check:

- [x] Changelog entry


src/inmanta/moduletool.py
```diff
@@ -1328,9 +1326,9 @@ def release(
             self.release(dev=True, message="Bump version to next development version", patch=True)
 
 
-class ModuleChangelog:
```
wouterdb, line 1331: For backward compatibility I would add `ModuleChangelog = Changelog` below this class

wouterdb (approved):

PR #7233, opened 2024-02-20
POC import cleanup 

wouterdb:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry


src/inmanta/plugins.py
```diff
@@ -21,40 +21,35 @@
 import inspect
 import os
 import subprocess
+import typing
 import warnings
-from collections import abc
-from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Type, TypeVar
 
+import inmanta.ast as ast
```
wouterdb, line 27: I like this to keep them short (i.e. snip off as much as possible)

src/inmanta/plugins.py
```diff
@@ -21,40 +21,35 @@
...
-from inmanta import const, protocol, util
-from inmanta.ast import LocatableString, Location, Namespace, Range, RuntimeException, TypeNotFoundException, WithComment
-from inmanta.ast.type import NamedType
-from inmanta.config import Config
-from inmanta.execute.proxy import DynamicProxy
-from inmanta.execute.runtime import QueueScheduler, Resolver, ResultVariable
-from inmanta.execute.util import NoneValue, Unknown
-from inmanta.stable_api import stable_api
-from inmanta.warnings import InmantaWarning
+import inmanta.warnings
```
wouterdb, line 29: This one can't be shortened, as we also import warnings: keep it like this or make it into inmanta_warnings?

src/inmanta/plugins.py
```diff
@@ -21,40 +21,35 @@
...
+from inmanta import config, const, protocol, stable_api, util
+from inmanta.execute import proxy, runtime
+from inmanta.execute import util as execute_util
```
wouterdb, line 32: this one shadows `inmanta.util` how to best handle this?

src/inmanta/plugins.py
```diff
@@ -21,40 +21,35 @@
 import inspect
 import os
 import subprocess
+import typing
 import warnings
-from collections import abc
```
wouterdb, line 25: Do we want just `abc` or `collection.abc`

src/inmanta/plugins.py
arnaudsjs: I like this too

src/inmanta/plugins.py
arnaudsjs: By making it into `inmanta_warnings` we don't win anything, right? In that case I would keep it like this.

src/inmanta/plugins.py
arnaudsjs: I like it like this were we alias to `execute_util`.

src/inmanta/plugins.py
arnaudsjs: I don't have a strong opinion on this.

PR #7236, opened 2024-02-21
Poc stage 2

wouterdb:
1. make sure we load config
2. make sure we configure logging
3. move to typed IPC (still a bit weird)

@bartv the IPC feels a bit weird to me like this. WDYT

# Description

* Short description here *

# Self Check:

- [x] Changelog entry
Mypy can't figure these out
```
src/inmanta/protocol/ipc_light.py:212: error: Invalid index type "UUID | None" for "dict[UUID, Future[object]]"; expected type "UUID"  [index]
src/inmanta/server/services/resourceservice.py:986: error: Argument 6 to "propagate_resource_state_if_stale" of "ResourceService" has incompatible type "ResourceState | None"; expected "ResourceState"  [arg-type]
```

### DIFF
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -51,11 +51,13 @@ Pipe break: go to 3a and 3b
 
 
 class ExecutorContext:
+    """The context object used by the executor to expose state to the incoming calls"""
 
     def __init__(self, server: "ExecutorServer") -> None:
         self.server = server
 
     async def stop(self) -> None:
+        """Request the executor to stop"""
         await self.server.stop()
 
 
@@ -89,7 +91,7 @@ class ExecutorServer(IPCServer[ExecutorContext]):
         self.stopped.set()
 
 
-class StopCommand(inmanta.protocol.ipc_light.IPCMethod[None, ExecutorContext]):
+class StopCommand(inmanta.protocol.ipc_light.IPCMethod[ExecutorContext, None]):
 
     async def call(self, context: ExecutorContext) -> None:
         await context.stop()
@@ -100,11 +102,14 @@ def mp_worker_entrypoint(
     name: str,
     logfile: str,
     inmanta_log_level: str,
+    cli_log: bool,
     config: typing.Mapping[str, typing.Mapping[str, typing.Any]],
 ) -> None:
     """Entry point for child processes"""
     log_config = InmantaLoggerConfig.get_instance()
     log_config.configure_file_logger(logfile, inmanta_log_level)
+    if cli_log:
+        log_config.add_cli_logger(inmanta_log_level)
     logging.captureWarnings(True)
     logger = logging.getLogger(f"agent.executor.{name}")
     logger.info(f"Started with PID: {os.getpid()}")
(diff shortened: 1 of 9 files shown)


src/inmanta/agent/executor.py
```diff
@@ -73,24 +88,30 @@ def connection_lost(self, exc: Exception | None) -> None:
...
+    name: str,
+    logfile: str,
+    inmanta_log_level: str,
+    config: typing.Mapping[str, typing.Mapping[str, typing.Any]],
+) -> None:
     """Entry point for child processes"""
-    # TODO: set up logging
-    InmantaLoggerConfig.get_instance()
+    log_config = InmantaLoggerConfig.get_instance()
+    log_config.configure_file_logger(logfile, inmanta_log_level)
```
wouterdb, line 107: This sucks for testing, Perhaps I also have to write to std out

src/inmanta/protocol/ipc_light.py
```diff
@@ -35,11 +36,22 @@ class ConnectionLost(IPCException):
     pass
 
 
+ServerContext = typing.TypeVar("ServerContext")
+R = typing.TypeVar("R")
```
arnaudsjs, line 40: It was not immediately clear to me this `R` represents the return value of the method. I think it would be good to give this TypeVar a better name. I think we should in general prevent single character identifiers, unless it's a very locally scoped loop variable.

src/inmanta/protocol/ipc_light.py
```diff
@@ -35,11 +36,22 @@ class ConnectionLost(IPCException):
...
+
+
+class IPCMethod(abc.ABC, typing.Generic[R, ServerContext]):
```
arnaudsjs, line 43: I would sway the two generic parameters, so that it's in-line with the signature of the method. So first the input arguments and then the return value.
```suggestion
class IPCMethod(abc.ABC, typing.Generic[ServerContext, R]):
```

src/inmanta/agent/executor.py
```diff
@@ -48,22 +50,35 @@
 """
 
 
-class ExecutorServer(IPCServer):
+class ExecutorContext:
```
arnaudsjs, line 53: I am missing a docstring here.

src/inmanta/agent/executor.py
```diff
@@ -73,24 +88,30 @@ def connection_lost(self, exc: Exception | None) -> None:
         self._sync_stop()
         self.stopped.set()
 
-    def get_method(self, name: str) -> typing.Callable[[...], typing.Coroutine[typing.Any, typing.Any, object]]:
-        if name == "stop":
-            return self.stop
 
-        async def echo(*args):
-            return list(args)
+class StopCommand(inmanta.protocol.ipc_light.IPCMethod[None, ExecutorContext]):
```
arnaudsjs, line 92: Just a suggestion: Would it be better to move all the `IPCMethod`s into a separate module?

arnaudsjs (changes requested):

src/inmanta/agent/executor.py
wouterdb: Probably, this one is a bit special, as it is feature of the server itself.

src/inmanta/agent/executor.py
wouterdb: added

arnaudsjs (approved):

PR #7239, opened 2024-02-22
Fix race in test

wouterdb:
# Description

Add no agent backoff 

was done on iso6 already: https://github.com/inmanta/inmanta-core/pull/6561

# Self Check:

- [x] Changelog entry


arnaudsjs (approved):
I think it would be useful to add the `no_agent_backoff` fixture to the `agent`, `agent_multi` and `agent_factory` fixtures. Because in general that's what you want if you use one of those fixtures. But we should investigate if this would introduce undesired side-effects on some tests.

PR #7246, opened 2024-02-23
issue/7068:Add dedicated set_fact and set_parameter endpoint

FloLey:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/protocol/methods_v2.py
+++ b/src/inmanta/protocol/methods_v2.py
@@ -22,7 +22,7 @@ import datetime
 import uuid
 from typing import Literal, Optional, Union
 
-from inmanta.const import AgentAction, ApiDocsFormat, Change, ClientType, ResourceState
+from inmanta.const import AgentAction, ApiDocsFormat, Change, ClientType, ParameterSource, ResourceState
 from inmanta.data import model
 from inmanta.data.model import DiscoveredResource, PipConfig, ResourceIdStr
 from inmanta.protocol import methods
@@ -1123,6 +1123,34 @@ def get_parameters(
 
 
 @typedmethod(
+    path="/parameters/<id>",
+    operation="PUT",
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.api, ClientType.compiler, ClientType.agent],
+    api_version=2,
+)
+def set_parameter(
+    tid: uuid.UUID,
+    id: str,
+    source: ParameterSource,
+    value: str,
+    metadata: dict[str, str] = {},
+    recompile: bool = False,
+) -> model.Parameter:
+    """
+    Set a parameter on the server. If the parameter is an tracked unknown, it will trigger a recompile on the server.
+    Otherwise, if the value is changed and recompile is true, a recompile is also triggered.
+
+    :param tid: The id of the environment
+    :param id: The name of the parameter
+    :param source: The source of the parameter.
+    :param value: The value of the parameter
+    :param metadata: Optional. Metadata about the parameter
+    :param recompile: Optional. Whether to trigger a recompile
+    """
+
+
+@typedmethod(
     path="/facts",
     operation="GET",
     arg_options=methods.ENV_OPTS,
@@ -1163,6 +1191,37 @@ def get_all_facts(
     :return: A list of all matching facts
     :raise NotFound: This exception is raised when the referenced environment is not found
     :raise BadRequest: When the parameters used for filtering, sorting or paging are not valid
+    """
+
+
+@typedmethod(
+    path="/facts/<name>",
+    operation="PUT",
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.api, ClientType.compiler, ClientType.agent],
+    api_version=2,
+)
+def set_fact(
+    tid: uuid.UUID,
+    name: str,
+    source: ParameterSource,
+    value: str,
+    resource_id: str,
+    metadata: dict[str, str] = {},
+    recompile: bool = False,
+    expires: Optional[bool] = True,
+) -> ReturnValue[model.Fact]:
+    """
+    Set a fact on the server
+
+    :param tid: The id of the environment
+    :param name: The name of the parameter
+    :param source: The source of the parameter.
+    :param value: The value of the parameter
+    :param resource_id: Optional. Scope the parameter to resource (fact)
+    :param metadata: Optional. Metadata about the parameter
+    :param recompile: Optional. Whether to trigger a recompile
+    :param expires: Optional. If the fact should expire or not. By default, facts expire.
     """
 
 
--- a/src/inmanta/server/services/paramservice.py
+++ b/src/inmanta/server/services/paramservice.py
@@ -364,6 +364,38 @@ class ParameterService(protocol.ServerSlice):
             raise NotFound(f"Fact with id {id} does not exist")
         return param.as_fact()
 
+    @handle(methods_v2.set_fact, env="tid")
+    async def set_fact(
+        self,
+        env: data.Environment,
+        name: str,
+        source: ParameterSource,
+        value: str,
+        resource_id: str,
+        metadata: dict = {},
+        recompile: bool = False,
+        expires: Optional[bool] = True,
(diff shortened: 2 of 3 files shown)


tests/test_server.py
```diff
@@ -1893,3 +1893,140 @@ async def put_version(version: int) -> int:
...
+    )
+    assert result.code == 200
+    assert len(result.result["data"]) == 1
+    assert result.result["data"][0] == fact
+
+
+async def test_set_param_v2(server, client, environment):
+    """
+    Test the set_parameter endpoint. Create a parameters and verify that expires is set to false.
+    also test we can modify it and create a second one.
```
arnaudsjs, line 1984:
```suggestion
    Also test we can modify it and create a second one.
```

tests/test_server.py
```diff
@@ -1893,3 +1893,140 @@ async def put_version(version: int) -> int:
...
+        {
+            "key": "key1",
+            "id": resource_version_id,
+            "send_event": True,
+            "purged": False,
+            "requires": [],
+        }
+    ]
+
+    result = await client.put_version(
```
arnaudsjs, line 1930:
```suggestion

    # Put a new version containing a resource with id=resource_id, to make sure the fact is not cleaned up.
    result = await client.put_version(
```

tests/test_server.py
```diff
@@ -1893,3 +1893,140 @@ async def put_version(version: int) -> int:
...
+):
+    """
+    Test the set_fact endpoint. First create a fact with expires set to true.
+    Then set expires to false for the same fact.
+    """
+    await agent_factory(
+        environment=environment,
+        agent_map={"discovery_agent": "localhost"},
+        hostname="discovery_agent",
+    )
```
arnaudsjs, line 1914: Is there a reason why this agent needs to be started. If not, I wouldn't do it, because it only makes the test case slower.

tests/test_server.py
```diff
@@ -1893,3 +1893,140 @@ async def put_version(version: int) -> int:
     # give it time to attempt to be release
     await asyncio.sleep(0.1)
     assert not await clienthelper.is_released(v1)
+
+
+async def test_set_fact(
```
arnaudsjs, line 1898:
```suggestion
async def test_set_fact_v2(
```

src/inmanta/protocol/methods_v2.py
arnaudsjs, line 1: I think we also need to add a note on the old endpoint to set facts/parameter saying that the new endpoint should be used instead.

src/inmanta/server/services/paramservice.py
```diff
@@ -364,6 +364,79 @@ async def get_fact(self, env: data.Environment, rid: ResourceIdStr, id: uuid.UUI
...
+        # Validate parameter or fact
+        self._validate_parameter(name, resource_id, expires)
+
+        # Update parameter with new value and metadata
+        result = await self._update_param(env, name, value, source, resource_id, metadata, recompile, expires)
+        warnings = None
+        if result:
+            compile_metadata = {
+                "message": "Recompile model because one or more parameters were updated",
+                "type": "param",
```
arnaudsjs, line 390: Here we always talk about parameters, but it could be facts as well.

src/inmanta/server/services/paramservice.py
```diff
@@ -364,6 +364,79 @@ async def get_fact(self, env: data.Environment, rid: ResourceIdStr, id: uuid.UUI
...
+        resource_id: Optional[str] = None,
+        expires: Optional[bool] = None,
+    ) -> tuple[data.Parameter, Optional[list[str]]]:
+        """
+        Update a parameter or fact and optionally trigger recompilation.
+        """
+        # Validate parameter or fact
+        self._validate_parameter(name, resource_id, expires)
+
+        # Update parameter with new value and metadata
```
arnaudsjs, line 384:
```suggestion
        # Update parameter/fact with new value and metadata
```

src/inmanta/server/services/paramservice.py
```diff
@@ -364,6 +364,79 @@ async def get_fact(self, env: data.Environment, rid: ResourceIdStr, id: uuid.UUI
...
+        result = await self._update_param(env, name, value, source, resource_id, metadata, recompile, expires)
+        warnings = None
+        if result:
```
arnaudsjs, line 387:
```suggestion
        recompile_required: bool = await self._update_param(env, name, value, source, resource_id, metadata, recompile, expires)
        warnings = None
        if recompile_required:
```

src/inmanta/server/services/paramservice.py
```diff
@@ -364,6 +364,79 @@ async def get_fact(self, env: data.Environment, rid: ResourceIdStr, id: uuid.UUI
...
+            compile_metadata = {
+                "message": "Recompile model because one or more parameters were updated",
+                "type": "param",
+                "params": [(name, resource_id)],
+            }
+            warnings = await self.server_slice._async_recompile(env, False, metadata=compile_metadata)
```
arnaudsjs, line 393:
```suggestion
            warnings = await self.server_slice._async_recompile(env, update_repo=False, metadata=compile_metadata)
```

src/inmanta/server/services/paramservice.py
```diff
@@ -364,6 +364,79 @@ async def get_fact(self, env: data.Environment, rid: ResourceIdStr, id: uuid.UUI
...
+
+        # Retrieve the updated parameter/fact
+        param = await data.Parameter.get_list(environment=env.id, name=name)
+        if not param:
+            raise NotFound(f"Parameter with id {name} does not exist")
```
arnaudsjs, line 398:
```suggestion
            raise NotFound(f"Parameter with id {name} does not exist in environment {env.id}")
```

src/inmanta/server/services/paramservice.py
```diff
@@ -364,6 +364,79 @@ async def get_fact(self, env: data.Environment, rid: ResourceIdStr, id: uuid.UUI
...
+        if result:
+            compile_metadata = {
+                "message": "Recompile model because one or more parameters were updated",
+                "type": "param",
+                "params": [(name, resource_id)],
+            }
+            warnings = await self.server_slice._async_recompile(env, False, metadata=compile_metadata)
+
+        # Retrieve the updated parameter/fact
+        param = await data.Parameter.get_list(environment=env.id, name=name)
```
arnaudsjs, line 396: * The primary key is (environment, name, resource_id). So you can have an environment with several parameters with the same each belonging to a different resource. So you have to specify the resource_id to make sure we return the correct parameter.
* You also need to make sure that `resource_id=None` is converted into `resource_id=""`, because that's how we store them into the database.
* The `get_list()` can also become a `get_one()` like this.

```suggestion
        param = await data.Parameter.get_list(environment=env.id, name=name, resource_id=resource_id)
```

arnaudsjs (changes requested):

src/inmanta/protocol/methods_v2.py
FloLey: should it be deprecated? should I raise a deprecation warning if it is used?

arnaudsjs: I would mark it as deprecated in the docstring. I am not sure whether we also want the deprecation warning, because at the moment we are still using the old API endpoints and I don't know how long we will keep on using them. @wouterdb Do you know?

wouterdb: There is no concrete plan to phase them out

src/inmanta/protocol/methods_v2.py
```diff
@@ -1122,6 +1122,34 @@ def get_parameters(
     """
 
 
+@typedmethod(
+    path="/parameters/<name>",
+    operation="PUT",
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.api, ClientType.compiler, ClientType.agent],
```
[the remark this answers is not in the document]
wouterdb: difficult. I think all is a safe default here

arnaudsjs (changes requested):

arnaudsjs (approved):

Issue #6993, opened 2024-01-16
inmanta deploy command ignore -f option


edvgui:
When using the `inmanta deploy` command, I expect to be able to pass the "main" file of my choice, just like the command helper suggests and similarly to the compile command, but it seems that this option is ignored.
```
(tmp-a44f955e1f6f00e) guillaume@fedora:/tmp/project$ inmanta -vvv deploy
inmanta.warnings         WARNING ProjectConfigurationWarning: Setting a pip index through the `repo.url` option with type `package` in the project.yml file is no longer supported and will be ignored. Please set the pip index url through the `pip.index_url` option instead.
inmanta.deploy           DEBUG   Storing state data in ./data/deploy
inmanta.deploy           DEBUG   Starting postgresql on port 41545
inmanta.deploy           DEBUG   Started server on port 37837
asyncio                  DEBUG   Using selector: EpollSelector
inmanta.deploy           INFO    Export of model complete
inmanta.deploy           INFO    Latest version for created environment is 4
std::AgentConfig[internal,agentname=localhost],v=4 - deployed
std::File[localhost,path=/tmp/test.txt],v=4 - deployed
[2 / 2]
Deploy ready
(tmp-a44f955e1f6f00e) guillaume@fedora:/tmp/project$ mv main.cf test.cf
(tmp-a44f955e1f6f00e) guillaume@fedora:/tmp/project$ inmanta -vvv deploy -f test.cf 
inmanta.warnings         WARNING ProjectConfigurationWarning: Setting a pip index through the `repo.url` option with type `package` in the project.yml file is no longer supported and will be ignored. Please set the pip index url through the `pip.index_url` option instead.
inmanta.deploy           DEBUG   Storing state data in ./data/deploy
inmanta.deploy           DEBUG   Starting postgresql on port 37261
inmanta.deploy           DEBUG   Started server on port 50497
asyncio                  DEBUG   Using selector: EpollSelector
An error occurred while compiling the model:
compiler       WARNING ProjectConfigurationWarning: Setting a pip index through the `repo.url` option with type `package` in the project.yml file is no longer supported and will be ignored. Please set the pip index url through the `pip.index_url` option instead.
compiler       DEBUG   Starting compile
compiler       WARNING Compile cache loading failure, ignoring cache entry for ./main.cf
                       Traceback (most recent call last):
[...]

PR #7254, opened 2024-02-26
Fix: deploy command no longer ignores `-f` option

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/6993

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/6993-inmanta-deploy-command-ignores-f-option.yml
+++ b/changelogs/unreleased/6993-inmanta-deploy-command-ignores-f-option.yml
@@ -1,0 +1,7 @@
+description: Deploy command no longer ignores ``-f`` option.
+issue-nr: 6993
+change-type: patch
+destination-branches: [master, iso7, iso6]
+sections:
+  bugfix: "{{description}}"
+
--- a/src/inmanta/deploy.py
+++ b/src/inmanta/deploy.py
@@ -319,7 +319,7 @@ host=localhost
 
         return True
 
-    def export(self) -> bool:
+    def export(self, main_file: str) -> bool:
         """
         Export a version to the embedded server
         """
@@ -336,6 +336,8 @@ host=localhost
             "localhost",
             "--server_port",
             str(self._server_port),
+            "-f",
+            main_file,
         ]
 
         sub_process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
@@ -485,7 +487,7 @@ host=localhost
         raise FinishedException()
 
     def run(self) -> None:
-        self.export()
+        self.export(main_file=self._options.main_file)
         self.deploy(dry_run=self._options.dryrun)
 
     def stop(self) -> None:


arnaudsjs (changes requested):
This PR is missing a test case.

tests/test_deploy.py
```diff
@@ -27,21 +27,31 @@
 from inmanta import deploy
 
 
-def test_deploy(snippetcompiler, tmpdir, postgres_db):
+@pytest.mark.parametrize("default_main_file", [True, False])
+def test_deploy(snippetcompiler, tmpdir, postgres_db, default_main_file):
```
arnaudsjs, line 31:
```suggestion
def test_deploy(snippetcompiler, tmpdir, postgres_db, default_main_file: bool):
```

tests/test_deploy.py
```diff
@@ -27,21 +27,31 @@
...
-        % file_name
-    )
+
+    main_file = "main.cf"
+
+    if default_main_file:
+        project = snippetcompiler.setup_for_snippet(code)
+    else:
+        main_file = "other.cf"
+        project = snippetcompiler.setup_for_snippet(code, main_file=main_file)
```
arnaudsjs, line 50: Just a simplification:
```suggestion
    main_file = "main.cf" if default_main_file else "other.cf"
     project = snippetcompiler.setup_for_snippet(code, main_file=main_file)
```

tests/conftest.py
```diff
@@ -1154,6 +1154,7 @@ def setup_for_snippet(
         use_pip_config_file: bool = False,
         index_url: Optional[str] = None,
         extra_index_url: list[str] = [],
+        main_file: str = "main.cf",
```
arnaudsjs, line 1157: This parameter needs documentation, because it can be supplied in two different ways:

* Either it's a relative path and then it's relative to the root of the project
* Or, it's an absolute path.

arnaudsjs (changes requested):

tests/test_deploy.py
Hugo-Inmanta: I wanted to test that the default was being used (i.e. not explicitly providing 'main.cf') but maybe it's not that relevant

tests/conftest.py
```diff
@@ -1174,6 +1175,7 @@ def setup_for_snippet(
         :param strict_deps_check: True iff the returned project should have strict dependency checking enabled.
         :param use_pip_config_file: True iff the pip config file should be used and no source is required for v2 to work
                                     False if a package source is needed for v2 modules to work
+        :param main_file: Relative path from the project root to the .cf file to use as main entry point.
```
arnaudsjs, line 1178: This can also be an absolute path, no?
```suggestion
        :param main_file: Path to the .cf file to use as main entry point. A relative or an absolute path can be provided. If a relative path is used, it's interpreted relative to the root of the project directory.
```

arnaudsjs (changes requested):

tests/conftest.py
Hugo-Inmanta: Hmm I don't think so. Unless I'm missing a usage it's currently used as a relative path:
- https://github.com/inmanta/inmanta-core/blob/71c9e88c6f9621dd31f3cd50d6b8f627d232febf/tests/conftest.py#L1304
- https://github.com/inmanta/inmanta-core/blob/71c9e88c6f9621dd31f3cd50d6b8f627d232febf/tests/conftest.py#L1207

arnaudsjs: I don't see what would fail. It's possible to join a two absolute paths together:

```
>>> import os.path
>>> 
>>> a = "/tmp/test"
>>> b = "/root/test2"
>>> os.path.join(a,b)
'/root/test2'
>>> 
>>> a = "/tmp/test"
>>> b = "root/test2"
>>> os.path.join(a,b)
'/tmp/test/root/test2'
>>> 

```

Hugo-Inmanta: Wow I didn't know that!

> If a path component represents an absolute path, then all previous components joined are discarded, and joining continues from the absolute path component.

arnaudsjs (approved):

Issue #7042, opened 2024-01-24
Postgresql installation documentation for RHEL 9 is not working


arnaudsjs:
Documentation: https://docs.inmanta.com/inmanta-service-orchestrator/7/install/1-install-server.html#step-3-install-postgresql-13

When I execute the command `sudo systemctl enable --now postgresql`, I get an error.

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/7042?planning-poker) with Zenhub @arnaudsjs @FloLey @Hugo-Inmanta @sanderr

PR #7268, opened 2024-02-27
Update docs for postgresql installation

Hugo-Inmanta:
# Description

Update docs

closes https://github.com/inmanta/inmanta-core/issues/7042

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/docs/install/1-install-server.rst
+++ b/docs/install/1-install-server.rst
@@ -4,7 +4,7 @@ Install Inmanta
 ***************
 
 This page explains how to install the Inmanta orchestrator software and setup an orchestration server. Regardless what platform
-you installed it on, Inmanta requires at least the latest Python 3.9 and git to be installed.
+you installed it on, Inmanta requires git to be installed.
 
 
 Install the software
@@ -189,8 +189,9 @@ Install the software
     Create a repositories file to point yum to the inmanta service orchestrator release repository. Create a file
     ``/etc/yum.repos.d/inmanta.repo`` with the following content:
 
+
     .. code-block:: sh
-       :substitutions:
+        :substitutions:
 
         [inmanta-service-orchestrator-|version_major|-stable]
         name=inmanta-service-orchestrator-|version_major|-stable
@@ -228,8 +229,8 @@ Install the software
     - The file with the .license extension is the license file
     - The file with the .jwe extension is the entitlement file
 
-    Copy the license file to the server and store them for example in ``/etc/inmanta/license``. If this directory does not exist, create it. Then create a
-    configuration file to point the orchestrator to the license file. Create a file ``/etc/inmanta/inmanta.d/license.cfg`` with the following content:
+    Copy both files to the server and store them for example in ``/etc/inmanta/license``. If this directory does not exist, create it. Then create a
+    configuration file to point the orchestrator to the license files. Create a file ``/etc/inmanta/inmanta.d/license.cfg`` with the following content:
 
     .. code-block::
 
@@ -265,7 +266,7 @@ Step 3: Install PostgreSQL 13
 
 .. only:: iso
 
-    Install the PostgreSQL 13 package included in RHEL.
+    Install the PostgreSQL 13 package included in RHEL. More info in the `postgresql documentation <https://www.postgresql.org/download/linux/redhat/>`_
 
     .. tab-set::
 
@@ -280,7 +281,7 @@ Step 3: Install PostgreSQL 13
 
             .. code-block:: sh
 
-                sudo dnf module install postgresql-server
+                sudo dnf install postgresql-server
                 sudo systemctl enable postgresql
 
 
@@ -300,6 +301,10 @@ Initialize the PostgreSQL server:
 .. only:: iso
 
     .. code-block:: sh
+
+        # TODO: the following step was failing because of missing langpacks, I had to install them with
+        # `sudo dnf install langpacks-en glibc-all-langpacks -y`
+        # I'm not sure whether this is specific to my setup of more general
 
         sudo su - postgres -c "postgresql-setup --initdb"
 
@@ -437,7 +442,7 @@ ssh and if the remote agent is used.
 
     sudo chown inmanta:inmanta /var/lib/inmanta/.ssh/config
 
-3. Add the public key to any git repositories and save if to include in configuration models that require remote agents.
+3. Add the public key to any git repositories and save it to include in configuration models that require remote agents.
 4. Test if you can login into a machine that has the public key and make sure ssh does not show you any prompts to store
    the host key.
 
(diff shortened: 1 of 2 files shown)


docs/install/1-install-server.rst
```diff
@@ -4,7 +4,7 @@ Install Inmanta
 ***************
 
 This page explains how to install the Inmanta orchestrator software and setup an orchestration server. Regardless what platform
-you installed it on, Inmanta requires at least the latest Python 3.9 and git to be installed.
+you installed it on, Inmanta requires git to be installed.
```
Hugo-Inmanta, line 7: I think this can go since inmanta service orchestrator server installs python 3.11

docs/install/1-install-server.rst
```diff
@@ -189,8 +189,9 @@ Install the software
     Create a repositories file to point yum to the inmanta service orchestrator release repository. Create a file
     ``/etc/yum.repos.d/inmanta.repo`` with the following content:
 
+
     .. code-block:: sh
-       :substitutions:
+        :substitutions:
```
Hugo-Inmanta, line 194: This was causing the whole block to be rendered with a leading space. That lead to some error when copy pasting this block into a config file

docs/install/1-install-server.rst
```diff
@@ -301,6 +302,10 @@ Initialize the PostgreSQL server:
 
     .. code-block:: sh
 
+        # TODO: the following step was failing because of missing langpacks, I had to install them with
+        # `sudo dnf install langpacks-en glibc-all-langpacks -y`
+        # I'm not sure whether this is specific to my setup of more general
+
         sudo su - postgres -c "postgresql-setup --initdb"
```
Hugo-Inmanta, line 309: This step was failing with following error:
`initdb: error: invalid locale settings; check LANG and LC_* environment variables`
Which comes from the fact that the system is using a locale that is not installed:

current status of the system locales settings:
```sh
[rocky@issue-7042 ~]$ localectl status
   System Locale: LANG=en_US.UTF-8     <---- Not installed
       VC Keymap: us
      X11 Layout: us
```

Installed locales:

```sh
[rocky@issue-7042 ~]$ locale -a

locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_MESSAGES to default locale: No such file or directory
locale: Cannot set LC_COLLATE to default locale: No such file or directory
C
C.utf8
POSIX
```


Managed to make the command work by changing to a locale that is installed :
```sh
[rocky@issue-7042 ~]$ sudo localectl set-locale LANG=C.utf8
```

An alternative fix involved installing the missing locale through 
```
[rocky@issue-7042 ~]$ sudo dnf install langpacks-en glibc-all-langpacks -y
```

Only lead I have on why this is happening is that it comes from my ssh configuration that used to have:
```txt
# file : /etc/ssh/ssh_config
#   SendEnv LANG LC_*    <-- this wasn't commented out
```

Hugo-Inmanta:
The `sudo su - postgres -c "postgresql-setup --initdb"` step was failing with following error:
`initdb: error: invalid locale settings; check LANG and LC_* environment variables`
Which comes from the fact that the system is using a locale that is not installed:

current status of the system locales settings:
```sh
[rocky@issue-7042 ~]$ localectl status
   System Locale: LANG=en_US.UTF-8     <---- Not installed
       VC Keymap: us
      X11 Layout: us
```

Installed locales:

```sh
[rocky@issue-7042 ~]$ locale -a

locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_MESSAGES to default locale: No such file or directory
locale: Cannot set LC_COLLATE to default locale: No such file or directory
C
C.utf8
POSIX
```


Managed to make the command work by changing to a locale that is installed :
```sh
[rocky@issue-7042 ~]$ sudo localectl set-locale LANG=C.utf8
```

An alternative fix involved installing the missing locale through 
```
[rocky@issue-7042 ~]$ sudo dnf install langpacks-en glibc-all-langpacks -y
```

Only lead I have on why this is happening is that it comes from my ssh configuration that used to have:
```txt
# file : /etc/ssh/ssh_config
#   SendEnv LANG LC_*    <-- this wasn't commented out
```

arnaudsjs:
> The `sudo su - postgres -c "postgresql-setup --initdb"` step was failing with following error: `initdb: error: invalid locale settings; check LANG and LC_* environment variables` Which comes from the fact that the system is using a locale that is not installed:
> 
> current status of the system locales settings:
> 
> ```shell
> [rocky@issue-7042 ~]$ localectl status
>    System Locale: LANG=en_US.UTF-8     <---- Not installed
>        VC Keymap: us
>       X11 Layout: us
> ```
> 
> Installed locales:
> 
> ```shell
> [rocky@issue-7042 ~]$ locale -a
> 
> locale: Cannot set LC_CTYPE to default locale: No such file or directory
> locale: Cannot set LC_MESSAGES to default locale: No such file or directory
> locale: Cannot set LC_COLLATE to default locale: No such file or directory
> C
> C.utf8
> POSIX
> ```
> 
> Managed to make the command work by changing to a locale that is installed :
> 
> ```shell
> [rocky@issue-7042 ~]$ sudo localectl set-locale LANG=C.utf8
> ```
> 
> An alternative fix involved installing the missing locale through
> 
> ```
> [rocky@issue-7042 ~]$ sudo dnf install langpacks-en glibc-all-langpacks -y
> ```

I would not change the locale, because that might have other undesired side-effect on the system. So I would do for the approach that installs the langpacks. I noticed that doing

```
sudo dnf install langpacks-en -y
```

is enough to make it work on a default RHEL 9. I would not mention the `glibc-all-langpacks` package for that reason in the install docs. But I would mention in the documentation why the `langpacks-en` package is required and that a different langpack might be required based on the locale used on the users system.

arnaudsjs (changes requested):

docs/install/1-install-server.rst
```diff
@@ -265,7 +266,7 @@ Step 3: Install PostgreSQL 13
 
 .. only:: iso
 
-    Install the PostgreSQL 13 package included in RHEL.
+    Install the PostgreSQL 13 package included in RHEL. More info in the `postgresql documentation <https://www.postgresql.org/download/linux/redhat/>`_
```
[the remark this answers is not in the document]
Hugo-Inmanta: I meant to link to the `Included in Distribution` part of the page but there's no anchor to link to :/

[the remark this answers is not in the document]
arnaudsjs: I see. Then I would explicitly mention the relevant section.

Hugo-Inmanta:
locale warning part renders as such: 
image

arnaudsjs (approved):

arnaudsjs:
> locale warning part renders as such: image

With the orange warning box it looks more dramatic than what I had in mind, but the message is definitely clear.

PR #7278, opened 2024-02-28
Performance investigation

### LINKED ISSUE #7262: Investigate performance of very large deployment
https://code.inmanta.com/solutions/athonet/athonet_mpn/-/merge_requests/237

wouterdb:
# For Reviewers

I changed some things, I would like to have reviewer input on what is acceptable:
1. more fine grained triggering of agents on put_partial: only notify agents that are in the increment
2. `put_version` and `put_partial` no longer wait for auto deploy to be completed **this may break wait conditions in tests everywhere**
3. increment cache is pre sorted per agent (slower on release (done once), faster for every agent (done often))
4. increment cache now refuses to move back to older versions
5. micro optimizations to use the DB more efficiently

** Do I need more tests anywhere? **

close #7262 

# Self Check:

- [x] Changelog entry
```

src/inmanta/config.py:219: error: Argument 1 to "CronTab" has incompatible type "str | int"; expected "str"  [arg-type]
src/inmanta/config.py:311: error: Incompatible default for argument "validator" (default has type "Callable[[str], str]", argument has type "Callable[[str | T], T]")  [assignment]
src/inmanta/data/__init__.py:4982: error: Argument 1 to "loads" has incompatible type "object"; expected "str | bytes | bytearray"  [arg-type]
src/inmanta/data/__init__.py:5441: error: Signature of "get_list" incompatible with supertype "BaseDocument"  [override]
src/inmanta/data/__init__.py:5441: note:      Superclass:
src/inmanta/data/__init__.py:5441: note:          @classmethod
src/inmanta/data/__init__.py:5441: note:          def get_list(cls, *, order_by_column: str | None = ..., order: str | None = ..., limit: int | None = ..., offset: int | None = ..., no_obj: bool | None = ..., lock: RowLockMode | None = ..., connection: Connection | None = ..., **query: object) -> Coroutine[Any, Any, list[ConfigurationModel]]
src/inmanta/data/__init__.py:5441: note:      Subclass:
src/inmanta/data/__init__.py:5441: note:          @classmethod
src/inmanta/data/__init__.py:5441: note:          def get_list(cls, *, order_by_column: str | None = ..., order: str | None = ..., limit: int | None = ..., offset: int | None = ..., no_obj: bool | None = ..., lock: RowLockMode | None = ..., connection: Connection | None = ..., no_status: bool = ..., **query: object) -> Coroutine[Any, Any, list[ConfigurationModel]]
src/inmanta/data/__init__.py:5509: error: Argument 2 to "_get_status_field" of "ConfigurationModel" has incompatible type "object"; expected "str"  [arg-type]
src/inmanta/data/__init__.py:5513: error: Argument 1 to "append" of "list" has incompatible type "ConfigurationModel"; expected "dict[str, object]"  [arg-type]
src/inmanta/data/__init__.py:5514: error: Incompatible return value type (got "list[dict[str, object]]", expected "list[ConfigurationModel]")  [return-value]
src/inmanta/data/__init__.py:5796: error: "object" has no attribute "__iter__"; maybe "__dir__" or "__str__"? (not iterable)  [attr-defined]
src/inmanta/data/__init__.py:5886: error: "object" has no attribute "__iter__"; maybe "__dir__" or "__str__"? (not iterable)  [attr-defined]
src/inmanta/data/__init__.py:5962: error: Incompatible types in assignment (expression has type "object", variable has type "int")  [assignment]
src/inmanta/server/services/orchestrationservice.py:849: error: Argument 1 to "add_background_task" of "TaskHandler" has incompatible type "Coroutine[Any, Any, tuple[int, dict[str, Any] | None]]"; expected "Coroutine[object, None, Result | None]"  [arg-type]
src/inmanta/server/services/compilerservice.py:795: error: Incompatible types in assignment (expression has type "bool | int | float | str | dict[str, str | int | bool]", variable has type "int")  [assignment]
Generated HTML report (via XSLT): /home/wouter/projects/inmanta/mypy/index.html
``` 

# Preliminary results
(on 15k resources)

1. we cause a storm of agent pulls (each put_partial makes every agent pull, we recompile fast)
   - [x] make it smarter
3. increment calculation is both on agent pull path and the release version
   - very performance sensitive
   - [x] it pulled in all attributes, so large config, slow increment
   - [ ] it interferes with itself somehow or mucks up the cache (to be investigated)
       - [x] make cache invalidation monotonic (only ever allow newer versions)
   - [x] auto deploy triggering is done in-line: the compile has to wait for it, we could change that
4. pyinstrument works somewhat as an async profiler, but not to the point where the numbers add up
6. slow query log, lock timing log and all query log on postgresql help somehwat
7. we still have some slow queries, basically everything related to the whole version or the part that remains the same
 - ` UPDATE resource SET status='deployed' WHERE environment=$1 AND model=$2 AND resource_id =ANY($3) ` 
 - 
 ```
 INSERT INTO resource(
	                environment,
	                model,
	                resource_id,
	                resource_type,
	                resource_id_value,
	                agent,
	                status,
	                attributes,
	                attribute_hash,
	                resource_set,
	                provides
	            )(
	                SELECT
	                    r.environment,
	                    $3,
	                    r.resource_id,
	                    r.resource_type,
	                    r.resource_id_value,
	                    r.agent,
	                    (
	                        CASE WHEN r.status='undefined'::resourcestate
	                        THEN 'undefined'::resourcestate
	                        ELSE 'available'::resourcestate
	                        END
	                    ) AS status,
	                    r.attributes AS attributes,
	                    r.attribute_hash,
	                    r.resource_set,
	                    r.provides
	                FROM resource AS r
	                WHERE r.environment=$1 AND r.model=$2 AND r.resource_set IS NOT NULL AND NOT r.resource_set=ANY($4)
	            )
	            RETURNING resource_id, resource_set
```
```
  SUM(CASE WHEN r.status NOT IN($1,$2) THEN 1 ELSE 0 END) AS done,
	                           to_json(array(SELECT jsonb_build_object('status', r2.status, 'id', r2.resource_id)
	                                         FROM resource AS r2
	                                         WHERE c.environment=r2.environment AND c.version=r2.model
	                                        )
	                           ) AS status
	                    FROM configurationmodel AS c LEFT OUTER JOIN resource AS r
	                    ON c.environment = r.environment AND c.version = r.model
	                    WHERE c.environment=$3 
	                    GROUP BY c.environment, c.version

```
```
INSERT INTO public.resourceaction_resource (resource_id, resource_version, environment, resource_action_id) SELECT unnest($1::text[]), unnest($2::int[]), $3, $4
```

Current time taken over parts of put_partial

```
2024-02-28 17:41:57,308 performance              WARNING STARTING PUT PARTIAL
2024-02-28 17:41:57,312 performance              WARNING INPUT VALIDATION: 0.0035941700043622404
2024-02-28 17:41:57,441 performance              WARNING LOAD STAGE: 0.1291558850207366
2024-02-28 17:41:57,802 performance              WARNING MERGE STAGE: 0.3613146049901843
2024-02-28 17:41:59,651 performance              WARNING PUT STAGE: 1.849367157992674
2024-02-28 17:42:01,870 performance              WARNING AUTO DEPLOY STAGE: 2.218535807012813
```

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4837,6 +4837,8 @@ class Resource(BaseDocument):
         """This method performs none of the mangling required to produce valid resources!
 
         project_attributes performs a projection on the json attributes of the resources table
+
+        all projections must be disjoint, as they become named fields in the output record
         """
 
         def collect_projection(projection: Optional[list[str]], prefix: str) -> str:
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4626,12 +4626,10 @@ class Resource(BaseDocument):
         cls,
         environment: uuid.UUID,
         resource_ids: list[m.ResourceIdStr],
-        version:int,
+        version: int,
         connection: Optional[asyncpg.connection.Connection] = None,
     ) -> list[m.ResourceIdStr]:
-        query = (
-            "UPDATE resource SET status='deployed' WHERE environment=$1 AND model=$2 AND resource_id =ANY($3) "
-        )
+        query = "UPDATE resource SET status='deployed' WHERE environment=$1 AND model=$2 AND resource_id =ANY($3) "
         async with cls.get_connection(connection) as connection:
             await connection.execute(query, environment, version, resource_ids)
 
@@ -4645,14 +4643,13 @@ class Resource(BaseDocument):
         lock: Optional[RowLockMode] = None,
         connection: Optional[asyncpg.connection.Connection] = None,
     ) -> list[m.ResourceIdStr]:
-        query = (
-            "SELECT resource_id as resource_id FROM resource WHERE environment=$1 AND model=$2 AND status = ANY($3) and resource_id =ANY($4) "
-        )
+        query = "SELECT resource_id as resource_id FROM resource WHERE environment=$1 AND model=$2 AND status = ANY($3) and resource_id =ANY($4) "
         if lock:
             query += lock.value
         async with cls.get_connection(connection) as connection:
-            return [r["resource_id"] for r in
-                    await connection.fetch(query, environment, version, statuses, resource_version_ids)]
+            return [
+                r["resource_id"] for r in await connection.fetch(query, environment, version, statuses, resource_version_ids)
+            ]
 
     @classmethod
     async def get_undeployable(cls, environment: uuid.UUID, version: int) -> list["Resource"]:
@@ -4845,7 +4842,7 @@ class Resource(BaseDocument):
                 return ",".join(f"{prefix}.{field}" for field in projection)
 
         if project_attributes:
-            json_projection = ',' + ",".join(f"r.attributes->'{v}' as {v}" for v in project_attributes)
+            json_projection = "," + ",".join(f"r.attributes->'{v}' as {v}" for v in project_attributes)
         else:
             json_projection = ""
 
@@ -4859,7 +4856,8 @@ class Resource(BaseDocument):
         for res in resources:
             if project_attributes:
                 for k in project_attributes:
-                    res[k] = json.loads(res[k])
+                    if res[k]:
+                        res[k] = json.loads(res[k])
         return resources
 
     @classmethod
@@ -5765,10 +5763,7 @@ class ConfigurationModel(BaseDocument):
             "last_deployed_attribute_hash",
             "last_non_deploying_status",
         ]
-        projection_a_attributes = [
-            "requires",
-            "send_event"
-        ]
+        projection_a_attributes = ["requires", "send_event"]
         projection = ["resource_id", "status", "attribute_hash"]
 
         # get resources for agent
@@ -5798,11 +5793,7 @@ class ConfigurationModel(BaseDocument):
                 req_res = id_to_resource[req]
                 assert req_res is not None  # todo
                 last_produced_events = req_res["last_produced_events"]
-                if (
-                    last_produced_events is not None
-                    and last_produced_events > last_success
-                    and req_res["send_event"]
-                ):
+                if last_produced_events is not None and last_produced_events > last_success and req_res["send_event"]:
                     in_increment = True
                     break
 
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4625,10 +4625,10 @@ class Resource(BaseDocument):
     async def set_deployed_multi(
         cls,
         environment: uuid.UUID,
-        resource_ids: list[m.ResourceIdStr],
+        resource_ids: Sequence[m.ResourceIdStr],
         version: int,
         connection: Optional[asyncpg.connection.Connection] = None,
-    ) -> list[m.ResourceIdStr]:
+    ) -> None:
         query = "UPDATE resource SET status='deployed' WHERE environment=$1 AND model=$2 AND resource_id =ANY($3) "
         async with cls.get_connection(connection) as connection:
             await connection.execute(query, environment, version, resource_ids)
@@ -4639,7 +4639,7 @@ class Resource(BaseDocument):
         environment: uuid.UUID,
         resource_version_ids: list[m.ResourceIdStr],
         version: int,
-        statuses: list[const.ResourceState],
+        statuses: Sequence[const.ResourceState],
         lock: Optional[RowLockMode] = None,
         connection: Optional[asyncpg.connection.Connection] = None,
     ) -> list[m.ResourceIdStr]:
@@ -4651,7 +4651,7 @@ class Resource(BaseDocument):
             query += lock.value
         async with cls.get_connection(connection) as connection:
             return [
-                r["resource_id"] for r in await connection.fetch(query, environment, version, statuses, resource_version_ids)
+                m.ResourceIdStr(cast(str, r["resource_id"])) for r in await connection.fetch(query, environment, version, statuses, resource_version_ids)
             ]
 
     @classmethod
@@ -5491,8 +5491,8 @@ class ConfigurationModel(BaseDocument):
                     {lock_statement}"""
         query_result = await cls._fetch_query(query_string, *values, connection=connection)
         result = []
-        for record in query_result:
-            record = dict(record)
+        for in_record in query_result:
+            record = dict(in_record)
             if no_obj:
                 if no_status:
                     record["status"] = {}
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4622,6 +4622,39 @@ class Resource(BaseDocument):
         return out
 
     @classmethod
+    async def set_deployed_multi(
+        cls,
+        environment: uuid.UUID,
+        resource_ids: list[m.ResourceIdStr],
+        version:int,
+        connection: Optional[asyncpg.connection.Connection] = None,
+    ) -> list[m.ResourceIdStr]:
+        query = (
+            "UPDATE resource SET status='deployed' WHERE environment=$1 AND model=$2 AND resource_id =ANY($3) "
+        )
+        async with cls.get_connection(connection) as connection:
+            await connection.execute(query, environment, version, resource_ids)
+
+    @classmethod
+    async def get_resource_ids_with_status(
+        cls,
+        environment: uuid.UUID,
+        resource_version_ids: list[m.ResourceIdStr],
+        version: int,
+        statuses: list[const.ResourceState],
+        lock: Optional[RowLockMode] = None,
+        connection: Optional[asyncpg.connection.Connection] = None,
+    ) -> list[m.ResourceIdStr]:
+        query = (
+            "SELECT resource_id as resource_id FROM resource WHERE environment=$1 AND model=$2 AND status = ANY($3) and resource_id =ANY($4) "
+        )
+        if lock:
+            query += lock.value
+        async with cls.get_connection(connection) as connection:
+            return [r["resource_id"] for r in
+                    await connection.fetch(query, environment, version, statuses, resource_version_ids)]
+
+    @classmethod
     async def get_undeployable(cls, environment: uuid.UUID, version: int) -> list["Resource"]:
         """
         Returns a list of resources with an undeployable state
@@ -4796,10 +4829,14 @@ class Resource(BaseDocument):
         version: int,
         projection: Optional[list[str]],
         projection_presistent: Optional[list[str]],
+        project_attributes: Optional[list[str]] = None,
         *,
         connection: Optional[Connection] = None,
     ) -> list[dict[str, object]]:
-        """This method performs none of the mangling required to produce valid resources!"""
+        """This method performs none of the mangling required to produce valid resources!
+
+        project_attributes performs a projection on the json attributes of the resources table
+        """
 
         def collect_projection(projection: Optional[list[str]], prefix: str) -> str:
             if not projection:
@@ -4807,16 +4844,22 @@ class Resource(BaseDocument):
             else:
                 return ",".join(f"{prefix}.{field}" for field in projection)
 
+        if project_attributes:
+            json_projection = ',' + ",".join(f"r.attributes->'{v}' as {v}" for v in project_attributes)
+        else:
+            json_projection = ""
+
         query = f"""
-        SELECT {collect_projection(projection, 'r')}, {collect_projection(projection_presistent, 'ps')}
+        SELECT {collect_projection(projection, 'r')}, {collect_projection(projection_presistent, 'ps')} {json_projection}
             FROM {cls.table_name()} r JOIN resource_persistent_state ps ON r.resource_id = ps.resource_id
             WHERE r.environment=$1 AND ps.environment = $1 and r.model = $2;"""
 
         resource_records = await cls._fetch_query(query, environment, version, connection=connection)
         resources = [dict(record) for record in resource_records]
         for res in resources:
-            if "attributes" in res:
-                res["attributes"] = json.loads(res["attributes"])
+            if project_attributes:
+                for k in project_attributes:
+                    res[k] = json.loads(res[k])
         return resources
 
     @classmethod
@@ -5403,6 +5446,7 @@ class ConfigurationModel(BaseDocument):
         no_obj: Optional[bool] = None,
         lock: Optional[RowLockMode] = None,
         connection: Optional[asyncpg.connection.Connection] = None,
+        no_status: bool = False,  # don't load the status field
         **query: object,
     ) -> list["ConfigurationModel"]:
         # sanitize and validate order parameters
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: src/inmanta/data/__init__.py [hunk 1]
```diff
@@ -4796,27 +4826,38 @@ async def get_resources_for_version_raw_with_persistent_state(
         version: int,
         projection: Optional[list[str]],
         projection_presistent: Optional[list[str]],
+        project_attributes: Optional[list[str]] = None,
```
wouterdb, line 4829: I added this so we don't have to pull in all attributes for the resource when not needed, this can be a  LOT of data

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -838,7 +847,7 @@ async def _trigger_auto_deploy(
             agent_trigger_method_on_autodeploy = cast(str, await env.get(data.AGENT_TRIGGER_METHOD_ON_AUTO_DEPLOY))
             agent_trigger_method_on_autodeploy = const.AgentTriggerMethod[agent_trigger_method_on_autodeploy]
             await self.release_version(
```
wouterdb, line 849: Only push agent that are affected

src/inmanta/server/services/resourceservice.py
```diff
@@ -335,9 +337,10 @@ async def get_resource_increment_for_agent(self, env: data.Environment, agent: s
 
         now = datetime.datetime.now().astimezone()
 
+        ON_AGENT_REGEX = re.compile(rf"^[a-zA-Z0-9_:]+\[{re.escape(agent)},")
+
         def on_agent(res: ResourceIdStr) -> bool:
-            idr = Id.parse_id(res)
-            return idr.get_agent_name() == agent
+            return ON_AGENT_REGEX.match(res) is not None
```
wouterdb, line 343: optimization because this is called a on every resource for every agent ( about `o(n**2)` )

src/inmanta/server/services/resourceservice.py
```diff
@@ -414,34 +417,74 @@ async def mark_deployed(
...
+        if not resources_id:
+            return
+
+        action_id = uuid.uuid4()
+
+        async with data.Resource.get_connection(connection.connection) as inner_connection:
+            async with inner_connection.transaction():
+                # validate resources
+                if only_update_from_states is not None:
+                    resources = await data.Resource.get_resource_ids_with_status(
```
wouterdb, line 433: avoid having full resources

src/inmanta/server/services/resourceservice.py
```diff
@@ -414,34 +417,74 @@ async def mark_deployed(
...
+                        return None
+
+                resources_version_ids: list[ResourceVersionIdStr] = [
+                    ResourceVersionIdStr(f"{res_id},v={version}") for res_id in resources_id
+                ]
+
+                resource_action = data.ResourceAction(
+                    environment=env.id,
+                    version=version,
+                    resource_version_ids=resources_version_ids,
```
wouterdb, line 453: this remains very heavy on the DB, as this blows up into a join table

src/inmanta/server/services/resourceservice.py
```diff
@@ -414,34 +417,74 @@ async def mark_deployed(
...
+                await resource_action.insert(connection=inner_connection)
+                self.log_resource_action(
+                    env.id,
+                    resources_version_ids,
+                    const.LogLevel.INFO.to_int,
+                    timestamp,
+                    "Setting deployed due to known good status",
+                )
+
+                await data.Resource.set_deployed_multi(env.id, resources_id, version, connection=inner_connection)
```
wouterdb, line 479: One query instead of a loop

src/inmanta/server/services/resourceservice.py
```diff
@@ -414,34 +417,74 @@ async def mark_deployed(
...
+                # Resource persistent state should not be affected
+
+        def post_deploy_update() -> None:
+            # Make sure tasks are scheduled AFTER the tx is done.
+            # This method is only called if the transaction commits successfully.
+            self.add_background_task(data.ConfigurationModel.mark_done_if_done(env.id, version))
```
wouterdb, line 485: Even this may not be required?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -69,6 +72,8 @@
 from inmanta.types import Apireturn, JsonType, PrimitiveTypes, ReturnTupple
 
 LOGGER = logging.getLogger(__name__)
+PLOGGER = logging.getLogger("performance")
```
wouterdb, line 75: todo: remove

tests/utils.py
```diff
@@ -347,6 +347,14 @@ async def put_version_simple(self, resources: dict[str, Any], version: int) -> N
             compiler_version=get_compiler_version(),
         )
         assert res.code == 200, res.result
+        if wait_for_released:
+            await retry_limited(functools.partial(self.is_released, version), 0.2, interval=0.05)
```
arnaudsjs, line 351:
```suggestion
            await retry_limited(functools.partial(self.is_released, version), timeout=0.2, interval=0.05)
```

tests/test_server.py
```diff
@@ -1882,17 +1883,13 @@ async def put_version(version):
...
-        f"Could not perform auto deploy on version 2 in environment {environment}, "
-        f"because Request conflicts with the current state of the resource: "
-        f"The version 2 on environment {environment} is older then the latest released version",
-    )
+            return version
+
+    v0 = await put_version(v0)
+    await retry_limited(functools.partial(clienthelper.is_released, v0), 1, 0.05)
+    v2 = await put_version(v2)
+    await retry_limited(functools.partial(clienthelper.is_released, v2), 1, 0.05)
```
arnaudsjs, line 1891:
```suggestion
    await retry_limited(functools.partial(clienthelper.is_released, v0), timeout=1, interval=0.05)
    v2 = await put_version(v2)
    await retry_limited(functools.partial(clienthelper.is_released, v2), timeout=1, interval=0.05)
```

arnaudsjs (changes requested):

arnaudsjs (approved):

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -916,13 +924,7 @@ async def put_version(
...
                     connection=con,
                 )
-            try:
-                await self._trigger_auto_deploy(env, version, connection=con)
-            except Conflict as e:
-                # this should be an api warning, but this is not supported here
-                LOGGER.warning(
-                    "Could not perform auto deploy on version %d in environment %s, because %s", version, env.id, e.log_message
-                )
+            await self._trigger_auto_deploy(env, version)
```
sanderr, line 927: I think we need to add to a comment a few lines up where we call `get_connection` that it must never be changed to use an existing, encapsulating connection object because we trigger a background task that requires this method's changes to be committed. i.e. avoid races like the ones we introduced in #6955.

src/inmanta/server/services/resourceservice.py
```diff
@@ -414,34 +419,74 @@ async def mark_deployed(
...
-            changes={},
-            messages=[logline],
-            change=const.Change.nochange,
-            send_events=False,
-            keep_increment_cache=True,
-            is_increment_notification=True,
-            only_update_from_states=only_update_from_states,
-            connection=connection,
-        )
+        resources_id_int = [res_id for res_id in resources_id if filter(res_id)]
```
sanderr, line 425: Why name it int? These are strings, aren't they?

src/inmanta/server/services/resourceservice.py
```diff
@@ -414,34 +419,74 @@ async def mark_deployed(
...
+        if not resources_id_int:
+            return
+
+        action_id = uuid.uuid4()
+
+        async with data.Resource.get_connection(connection.connection) as inner_connection:
```
sanderr, line 431: Isn't this a noop?

src/inmanta/server/services/resourceservice.py
sanderr, line 425:
```suggestion
        # performance-critical path: avoid parsing cost if we can
        resources_id_int = [res_id for res_id in resources_id if filter(res_id)]
```

src/inmanta/server/services/resourceservice.py
```diff
@@ -414,34 +417,74 @@ async def mark_deployed(
...
+                self.log_resource_action(
+                    env.id,
+                    resources_version_ids,
+                    const.LogLevel.INFO.to_int,
+                    timestamp,
+                    "Setting deployed due to known good status",
+                )
+
+                await data.Resource.set_deployed_multi(env.id, resources_id, version, connection=inner_connection)
+                # Resource persistent state should not be affected
```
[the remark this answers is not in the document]
sanderr: 1. Why don't we need to update it? Because we assume they're already in that state?

src/inmanta/server/services/resourceservice.py
```diff
@@ -563,31 +612,46 @@ async def get_increment(
...
             if cache_entry is None:
                 # No cache entry found
                 return None
-            (version_cache_entry, incr, neg_incr) = cache_entry
-            if version_cache_entry != version:
+            (version_cache_entry, incr, neg_incr, neg_incr_per_agent, run_ahead_lock) = cache_entry
+            if version_cache_entry > version:
+                # Cache is ahead,
+                if run_ahead_lock is not None:
+                    await run_ahead_lock.wait()
```
sanderr, line 634: Took me a while to realize this `run_ahead_lock` shadows the one passed in to the method. Perhaps we can rename it `cache_run_ahead_lock`?
Also added a comment.

```suggestion
            (version_cache_entry, incr, neg_incr, neg_incr_per_agent, cache_run_ahead_lock) = cache_entry
            if version_cache_entry > version:
                # Cache is ahead,
                if cache_run_ahead_lock is not None:
                    # wait for cache entry to be complete
                    await cache_run_ahead_lock.wait()
```

src/inmanta/server/services/resourceservice.py
```diff
@@ -563,31 +612,46 @@ async def get_increment(
...
             or None if no such cache entry exists.
             """
             cache_entry = self._increment_cache.get(env.id, None)
             if cache_entry is None:
                 # No cache entry found
                 return None
-            (version_cache_entry, incr, neg_incr) = cache_entry
-            if version_cache_entry != version:
+            (version_cache_entry, incr, neg_incr, neg_incr_per_agent, run_ahead_lock) = cache_entry
+            if version_cache_entry > version:
```
sanderr, line 631: Why don't we need to wait if `version_cache_entry == version`?

src/inmanta/server/services/resourceservice.py
```diff
@@ -563,31 +612,46 @@ async def get_increment(
...
-            return incr, neg_incr
+            return version_cache_entry, incr, neg_incr, neg_incr_per_agent
 
-        increment: Optional[tuple[abc.Set[ResourceIdStr], abc.Set[ResourceIdStr]]] = _get_cache_entry()
+        increment: Optional[
+            tuple[int, abc.Set[ResourceIdStr], abc.Set[ResourceIdStr], abc.Mapping[str, abc.Set[ResourceIdStr]]]
+        ] = await _get_cache_entry()
         if increment is None or (connection is not None and connection.is_in_transaction()):
             lock = self._increment_cache_locks[env.id]
             async with lock:
```
sanderr, line 646: Can't this lock create a deadlock with the `run_ahead_lock`? In the release flow the `run_ahead lock` is acquired first but if a fast agent calls `get_increment` before the release flow does it acquires this lock first. Then it starts waiting for the `run_ahead_lock` which will in turn start waiting for this lock.

sanderr (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -4796,27 +4830,38 @@ async def get_resources_for_version_raw_with_persistent_state(
...
```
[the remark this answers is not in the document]
wouterdb: https://github.com/python/mypy/issues/12554

src/inmanta/server/config.py
```diff
@@ -244,18 +244,18 @@ def validate_fact_renew(value: object) -> int:
...
     "File prefix in log-dir, containing the resource-action logs. The after the prefix the environment uuid and .log is added",
     is_str,
 )
 
-server_enabled_extensions = Option(
+server_enabled_extensions: Option[list[str]] = Option(
     "server",
     "enabled_extensions",
-    "",
+    [],
```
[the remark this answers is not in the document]
wouterdb: changed

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -818,13 +825,14 @@ async def _put_version(
...
         LOGGER.debug("Successfully stored version %d", version)
+        return list(all_agents)
 
     async def _trigger_auto_deploy(
         self,
         env: data.Environment,
         version: int,
         *,
-        connection: Optional[Connection],
+        agents: Optional[abc.Sequence[str]] = None,
```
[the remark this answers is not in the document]
wouterdb: added

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -916,13 +924,7 @@ async def put_version(
                     pip_config=pip_config,
                     connection=con,
                 )
-            try:
-                await self._trigger_auto_deploy(env, version, connection=con)
-            except Conflict as e:
-                # this should be an api warning, but this is not supported here
-                LOGGER.warning(
-                    "Could not perform auto deploy on version %d in environment %s, because %s", version, env.id, e.log_message
-                )
```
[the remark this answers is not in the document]
wouterdb: Very difficult and not very valuable I think. 

It is not an error per-se

src/inmanta/server/services/resourceservice.py
wouterdb: if the connection holder is not empty, yes, if it is empty, no

src/inmanta/server/services/resourceservice.py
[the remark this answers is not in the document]
wouterdb: Correct, this is a copy

src/inmanta/server/services/resourceservice.py
wouterdb: right

src/inmanta/server/services/resourceservice.py
wouterdb: the release flow never acquires the `run_ahead_lock` , it creates it. 
- the release flow is always head, so it passes in the lock for others, but doesn't every block on it
- the run_ahead_lock is placed just prior to leaving the critical section here, so it will be resolved.

src/inmanta/data/__init__.py
```diff
@@ -4796,27 +4830,38 @@ async def get_resources_for_version_raw_with_persistent_state(
...
         *,
         connection: Optional[Connection] = None,
     ) -> list[dict[str, object]]:
-        """This method performs none of the mangling required to produce valid resources!"""
+        """This method performs none of the mangling required to produce valid resources!
+
+        project_attributes performs a projection on the json attributes of the resources table
```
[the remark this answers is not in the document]
wouterdb: This method is so purpose specific, that I would expect this to be obvious

src/inmanta/server/services/resourceservice.py
sanderr: Doesn't it acquire it in order to populate the associated cache? i.e. [here](https://github.com/inmanta/inmanta-core/pull/7278/files#diff-e3723a8f8dde22b64bfd475f55960927014a3896679d1ca244b250dd8b1e5987R1154). So it creates the lock, and then eventually populates the cache by calling into the same flow that acquires both this lock and the other one.

sanderr: Ah no, I see. The context manager does not place the lock, it's the `get_increment` call itself that places it. And indeed it is only placed after populating the cache (even if it is created before).

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
sanderr: Oh, shame

sanderr (approved):

Issue #7026, opened 2024-01-23
enable agent install of dependency modules by default


sanderr:
The `agent_install_dependency_modules` project option has to be set in all practical models. Therefore it would make sense to phase it out and enable its behavior by default. This requires a few extra steps to make it safe:

- [x] add documentation about inter-module dependencies: you can use them (no option required) but the module you depend on should not have any resources (technically it can iff they are always there but that's more difficult to explain so only mention that if the solutions team sees a use case for it). Otherwise you could get the following timeline:
1. `a` depends on `b`
2. exporter exports `a-1.0` and `b-1.0`, both of which have resources for this agent. Agent puts code in its code directory.
3. exporter exports `a-1.0` and `b-2.0`, `b` no longer has resources for this agent. Agent puts code for `a`, not `b` in its code directory but the old code for `b` is still there.
4. agent is restarted (also triggers without restart but unlike other code loading issues a restart doesn't solve it here): agent's module finder points to its code directory -> stale version of `b-1.0` is loaded instead of `b-2.0` installed by pip in the venv.

- [x] on the agent, before loading the code, detect the above scenario and add a warning log message and notification for it.

All of this is iso7+?

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/7026?planning-poker) with Zenhub @arnaudsjs @FloLey @Hugo-Inmanta @sanderr

PR #7285, opened 2024-03-01
Enable agent_install_dependency_modules config option by default

arnaudsjs:
# Description

This PR enables the project configuration option `agent_install_dependency_modules` by default.

closes inmanta/inmanta-core#7026

# Self Check:

- [x] Changelog entry


FloLey (approved):

docs/model_developers/modules.rst
```diff
@@ -347,10 +347,26 @@ Add the following content to the file:
...
 You will also need to specify the url of the repository in the ``project.yml`` file of your project (See: :ref:`specify_location_pip`).
 
 By following the previous steps, the Inmanta server will be able to install modules from a private Python package repository.
 
 
+Inter-module dependencies
+#########################
+
+The plugins code of a module A can have a dependency on the plugins code of another V2 module B. When doing this,
+care should be taken that the Python module(s) you depend on, do not define any resources or providers. Otherwise the
```
sanderr, line 361: Have you checked this with the solutions team? From the ticket:

> technically it can iff they are always there but that's more difficult to explain so only mention that if the solutions team sees a use case for it

Another, similar, technicality that we probably want to skip: the module can contain resources as long as those resources are not present in the model.

sanderr, line 361:
```suggestion
care should be taken that the module(s) you depend on, do not define any resources or providers. Otherwise the
```

I think the "Python" qualifier could be more confusing than helpful. It may seem to imply that it's the specific python module (file) that you depend on, while in practice none of the Python modules in that inmanta module may have resources.

docs/model_developers/modules.rst
```diff
@@ -347,10 +347,26 @@ Add the following content to the file:
...
+python environment of the agent can get corrupt in the following way:
+
+1. Module A-1.0 depends on X.py of module B-1.0
```
sanderr, line 364: Any reason to use capital letters for module names and Python file names? Those are conventionally lowercase.

docs/model_developers/modules.rst
```diff
@@ -347,10 +347,26 @@ Add the following content to the file:
...
+2. A-1.0 and B-1.0 are exported: The exporter exports the X.py file to the server and the agent puts the X.py file in
+   its code directory.
+3. In a later version of module B (B-2.0) the X.py file doesn't have any resources anymore.
+4. A-1.0 and B-2.0 are exported: The exporter doesn't export X.py anymore, because it doesn't have any resources. The
+   agent doesn't modify the X.py present in its code directory, but the old version of X.py is still present. The
+   results is that the stale version of X.py is loadedinstead of the version present in the agents python environment.
```
sanderr, line 370:
```suggestion
   result is that the stale version of X.py is loaded instead of the version present in the agent's python environment.
```

docs/model_developers/modules.rst
```diff
@@ -347,10 +347,26 @@ Add the following content to the file:
...
+#########################
+
+The plugins code of a module A can have a dependency on the plugins code of another V2 module B. When doing this,
+care should be taken that the Python module(s) you depend on, do not define any resources or providers. Otherwise the
+python environment of the agent can get corrupt in the following way:
+
+1. Module A-1.0 depends on X.py of module B-1.0
+2. A-1.0 and B-1.0 are exported: The exporter exports the X.py file to the server and the agent puts the X.py file in
+   its code directory.
+3. In a later version of module B (B-2.0) the X.py file doesn't have any resources anymore.
```
sanderr, line 367: I think this one should be split up:
- the resources being present or not are a property of the model, which will in most (if not all) cases be on the side of either the project or module `A`.
- Then, sometime after that a new version of module `B` is released and included in the project.

docs/model_developers/modules.rst
```diff
@@ -347,10 +347,26 @@ Add the following content to the file:
...
+4. A-1.0 and B-2.0 are exported: The exporter doesn't export X.py anymore, because it doesn't have any resources. The
```
sanderr, line 368: > B-2.0 are exported: The exporter doesn't export X.py anymore

So `B` is not exported at all I think.

src/inmanta/data/__init__.py
```diff
@@ -6104,7 +6104,7 @@ class Notification(BaseDocument):
     title: str
     message: str
     severity: const.NotificationSeverity = const.NotificationSeverity.message
-    uri: str
+    uri: Optional[str] = None
```
sanderr, line 6107: Can the frontend handle this?

src/inmanta/module.py
```diff
@@ -1657,7 +1657,7 @@ class ProjectMetadata(Metadata, MetadataFieldRequires):
         Annotated[str, StringConstraints(strip_whitespace=True, pattern=_re_relation_precedence_rule, min_length=1)]
     ] = []
     strict_deps_check: bool = True
-    agent_install_dependency_modules: bool = False
+    agent_install_dependency_modules: bool = True
```
sanderr, line 1660: Do we even still need the option? I think we can just drop it completely? As long as we don't break models that set it, but I think extra options are always just ignored.

changelogs/unreleased/7026-agent-install-dependency-modules.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: The project configuration option `agent_install_dependency_modules` is now enabled by default.
```
sanderr, line 2: The wording may or may not have to change a bit depending on how my comment on the option resolves.

sanderr (changes requested):

docs/model_developers/modules.rst
arnaudsjs: > Have you checked this with the solutions team?

Yes.

src/inmanta/data/__init__.py
arnaudsjs: Yes.

docs/model_developers/modules.rst
arnaudsjs: I don't understand this. Could you elaborate?

docs/model_developers/modules.rst
arnaudsjs: Correct! I will change that.

docs/model_developers/modules.rst
arnaudsjs: I see that usually there is also a change required on module mod-a, but I still don't see clearly which changes you want to see applied. Could you make a concrete proposal?

sanderr: 1. <current>. The configuration model (in the project or one of the modules) constructs resources from both modules A and B.
2. <unchanged>
3. The configuration model is changed to only construct resources from module A.
4. A-1.0 is exported again, B-1.0 is no longer exported because it doesn't have any resources. The X.py file still exists in the agent code directory.
5. A new version of module B (B-2.0) is released and included in the project. The project is re-exported: A-1.0 is exported again, B-2.0 is not (again because it doesn't have any resources). The old X.py file still exists in the agent code directory. It is loaded by the agent instead of the one from B-2.0.

It still cuts some corners but I think it is an OK compromise.

PR #7287, opened 2024-03-01
parent 0ba57ee05f46b154084218def8b108ed04ae2b72

FloLey:
* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -21,9 +21,14 @@ import asyncio
 import dataclasses
 import datetime
 import enum
+import functools
+import hashlib
+import json
 import logging
 import os
 import random
+import shutil
+import subprocess
 import time
 import uuid
 from asyncio import Lock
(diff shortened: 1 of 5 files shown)


src/inmanta/agent/agent.py
```diff
@@ -1691,30 +1693,91 @@ async def create_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPo
...
+            return False
+        return True
+
+    def get_installed_packages(self, env_storage: str) -> dict:
+        """
+        Gets a dictionary of installed packages and their versions
+        :param env_storage: The file system path to the virtual environment.
+        :return: A dictionary with package names as keys and their versions as values.
+        """
+        pip_path = os.path.join(env_storage, "bin", "pip")
```
wouterdb, line 1741: I would prefer not to fork pip.
We are going to rather extreme lengths to make this process efficient, forking pip up to 3 times in a single `get_environment` operation is not an option (not even once, tbh)

wouterdb (changes requested):

src/inmanta/agent/agent.py
FloLey: @wouterdb than I'll close this PR/Branch and put the ticket back

PR #7288, opened 2024-03-01
Fix: unfreeze-pip-config

FloLey:
# Description
Unfreeze the PipConfig class as it is no longer needed

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -1564,6 +1564,14 @@ class EnvBlueprint:
         hash_obj = hashlib.md5(serialized_blueprint.encode("utf-8"))
         return hash_obj.hexdigest()
 
+    def __eq__(self, other):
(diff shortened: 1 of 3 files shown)


src/inmanta/agent/agent.py
```diff
@@ -1597,6 +1605,14 @@ def to_env_blueprint(self) -> EnvBlueprint:
         """
         return EnvBlueprint(pip_config=self.pip_config, requirements=self.requirements)
 
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, ExecutorBlueprint):
+            return False
+        return self.generate_blueprint_hash() == other.generate_blueprint_hash()
```
wouterdb, line 1611: I would cache `generate_blueprint_hash` is you are going to use it so often

wouterdb (changes requested):
One comment

wouterdb (approved):

PR #7290, opened 2024-03-01
Refactor for integration with IPC

wouterdb:
# Description

1. move base infrastructure to `inmanta.agent.executor`, this package doesn't depend on the agent itself 
2. create LocalExecutor  to contain the local code 
3. Move tests to own file 

# Self Check:

- [x] Changelog entry


src/inmanta/agent/local_executor.py
```diff
@@ -0,0 +1,66 @@
...
+import os
+import typing
+from concurrent.futures import ThreadPoolExecutor
+
+import inmanta.agent.executor as executor
+from inmanta.agent import config as cfg
+from inmanta.loader import ModuleSource
+
+
+class LocalExecutor(executor.Executor):
```
arnaudsjs, line 24: I do not entirely understand where term `Local` comes from here?

wouterdb: It is the one that runs in the agent process itself.

src/inmanta/agent/local_executor.py
```diff
@@ -0,0 +1,66 @@
...
+    ) -> LocalExecutor:
+        """
+        Creates an Executor based with the specified agent name and blueprint.
+        It ensures the required virtual environment is prepared and source code is loaded.
+
+        :param executor_id: executor identifier containing an agent name and a blueprint configuration.
+        :return: An Executor instance
+        """
+
+        executor = LocalExecutor(executor_id, venv, self.storage)
```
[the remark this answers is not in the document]
wouterdb: the current agent has only one storage pool `self._storage["code"]` 

It could/should/will even share the entire `CodeLoader`

src/inmanta/agent/local_executor.py
arnaudsjs: Shouldn't we call it `InCurrentProcessExecutor` instead? It's quite a long name, but it's clear. `Local` seems to imply that the other executors run on a remote machine, which is not really the case.

src/inmanta/agent/local_executor.py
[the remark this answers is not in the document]
arnaudsjs: Ok. I see.

arnaudsjs (approved):

FloLey (approved):

PR #7300, opened 2024-03-04
Fix race shutdown logic MPExecutor

arnaudsjs:
# Description

This PR fixes a race condition that causes `Process.close()` to be called twice, which results in a ValueError the second time. The regular shutdown logic races with invocation of the `connection_lost()` finalizer of the `FinalizingIPCClient`.

This made the following test case fail:

```
_____________________________ test_executor_server _____________________________

tmp_path = PosixPath('/tmp/pytest-of-jenkins/pytest-37620/test_executor_server0')

    async def test_executor_server(tmp_path):
        manager = MPManager(log_folder=str(tmp_path), cli_log=True)
        manager.init_once()
    
        inmanta.config.Config.set("test", "aaa", "bbbb")
    
        child1 = await manager.make_child_and_connect("Testchild")
    
        result = await child1.connection.call(Echo(["aaaa"]))
    
        assert ["aaaa"] == result
    
        result = await child1.connection.call(GetConfig("test", "aaa"))
        assert "bbbb" == result
    
        print("stopping")
        await manager.stop()
        print("joining")
>       await manager.join(10)

tests/forking_agent/test_executor.py:68: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.tox/tests/lib/python3.11/site-packages/inmanta/agent/forking_executor.py:225: in join
    await asyncio.gather(*(child.join(timeout) for child in self.children))
.tox/tests/lib/python3.11/site-packages/inmanta/agent/forking_executor.py:159: in join
    self.process.close()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Process name='agent.executor.Testchild' parent=717276 closed>

    def close(self):
        '''
        Close the Process object.
    
        This method releases resources held by the Process object.  It is
        an error to call this method if the child process is still running.
        '''
        if self._popen is not None:
            if self._popen.poll() is None:
                raise ValueError("Cannot close a process while it is still running. "
                                 "You should first call join() or terminate().")
            self._popen.close()
            self._popen = None
>           del self._sentinel
E           AttributeError: 'Process' object has no attribute '_sentinel'

/usr/lib64/python3.11/multiprocessing/process.py:185: AttributeError
```

# Self Check:

- [x] Changelog entry
- [ ] ~~If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes]~~(https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)


wouterdb (approved):
I found out as well: https://github.com/inmanta/inmanta-core/pull/7299

arnaudsjs:
Will be fixed in https://github.com/inmanta/inmanta-core/pull/7299#pullrequestreview-1914355275

PR #7309, opened 2024-03-05
Issue/6999 cad performance

### LINKED ISSUE #6999: Improve performance of cross agent depdency handling
# Problem 
We are observing scaling problems with CAD's. 

The current mechanism is as follows:
- when a agent starts to deploy, it find all requires that go to another agent
- for those, it constructs a RemoteResourceAction that will 
   - ask the server for the last know state for this resource
   - if that is deployed available or deploying, it indicates it ready
   - otherwise it waits for an event from the server 

When we have many resources with many CAD's (a few thousand) this 'ask the server step' overwhelms the agent max session limit. For this, we added a rate limiter (3 at a time), but now it is just very slow as each call takes up to 50ms (and DB connection pool contention adds to this)

# Solution

We may want to resolve these dependencies in batch. 
Luck would have it that the code on the server that produces the resource for the agent already has to pull in all CAD's to the server. So we can just add this extra info to the return value, at little cost. If does however make all of this a bit more complicated,

wouterdb:
# Description

Attempt to improve CAD performance by batching 

closes #6999 

# Self Check:

- [x] Changelog entry

```
src/inmanta/protocol/methods_v2.py:1522: error: Missing return statement  [empty-body]
src/inmanta/data/__init__.py:4639: error: Argument 1 to "ResourceIdStr" has incompatible type "object"; expected "str"  [arg-type]
src/inmanta/data/__init__.py:5724: error: Enum index should be a string (actual index type "object")  [misc]
src/inmanta/server/services/resourceservice.py:1245: error: Incompatible return value type (got "dict[ResourceIdStr, ResourceState]", expected "dict[str, ResourceState]")  [return-value]

``` 

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -562,7 +562,7 @@ class ResourceScheduler:
                 self.deferred = None
 
     async def resolve_cads(self, version: int, cads: dict[ResourceIdStr, RemoteResourceAction]) -> None:
-        """ Batch resolve all CAD's """
+        """Batch resolve all CAD's"""
         async with self.agent.process.cad_ratelimiter:
             result = await self.get_client().resources_status(
                 tid=self.agent._env_id,
@@ -575,10 +575,9 @@ class ResourceScheduler:
 
             for rid_raw, status_raw in result.result["data"].items():
                 status = const.ResourceState[status_raw]
-                rra = cads.get(rid_raw, None)
-                if rra is not None:
-                    if status not in const.TRANSIENT_STATES:
-                        rra.notify(status)
+                rra = cads[rid_raw]
+                if status not in const.TRANSIENT_STATES:
+                    rra.notify(status)
 
     def notify_ready(
         self,
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -562,9 +562,8 @@ class ResourceScheduler:
                 self.deferred = None
 
     async def resolve_cads(self, version: int, cads: dict[ResourceIdStr, RemoteResourceAction]) -> None:
-
+        """ Batch resolve all CAD's """
         async with self.agent.process.cad_ratelimiter:
-
             result = await self.get_client().resources_status(
                 tid=self.agent._env_id,
                 model_version=version,
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -842,6 +842,7 @@ class ResourceScheduler:
             ra.cancel()
         if self.cad_resolver is not None:
             self.cad_resolver.cancel()
+            self.cad_resolver = None
         self.generation = {}
         self.cad = {}
 
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -82,8 +82,6 @@ class ResourceActionBase(abc.ABC):
     resource_id: Id
     future: ResourceActionResultFuture
     dependencies: list["ResourceActionBase"]
-    # resourceid -> attribute -> {current: , desired:}
-    changes: dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]
 
     def __init__(self, scheduler: "ResourceScheduler", resource_id: Id, gid: uuid.UUID, reason: str) -> None:
         """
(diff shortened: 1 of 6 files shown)


src/inmanta/agent/agent.py
```diff
@@ -138,12 +135,16 @@ def __init__(self, scheduler: "ResourceScheduler", gid: uuid.UUID, reason: str)
...
+    # resourceid -> attribute -> {current: , desired:}
+    changes: dict[ResourceVersionIdStr, dict[str, AttributeStateChange]]
+
     def __init__(self, scheduler: "ResourceScheduler", resource: Resource, gid: uuid.UUID, reason: str) -> None:
         """
         :param gid: A unique identifier to identify a deploy. This is local to this agent.
         """
         super().__init__(scheduler, resource.id, gid, reason)
         self.resource: Resource = resource
+        self.change: Optional[const.Change] = const.Change.nochange
```
wouterdb, line 147: Moved up from baseclass, not needed there

src/inmanta/server/services/resourceservice.py
```diff
@@ -1231,6 +1231,19 @@ async def resource_list(
 
         # TODO: optimize for no orphans
 
+    @handle(methods_v2.resources_status, env="tid")
+    async def resources_status(
+        self,
+        env: data.Environment,
+        model_version: int,
+        rids: list[ResourceIdStr],
+    ) -> dict[str, ResourceState]:
```
Hugo-Inmanta, line 1240:
```suggestion
    ) -> dict[ResourceIdStr, ResourceState]:
```

src/inmanta/agent/agent.py
```diff
@@ -599,6 +561,25 @@ async def mark_deployment_as_finished(self, resource_actions: Iterable[ResourceA
...
+                model_version=version,
+                rids=list(cads.keys()),
+            )
+            if result.code != 200 or result.result is None:
+                LOGGER.error("Failed to get the status for remote resources %s (%s)", list(cads.keys()), result.result)
+                return
+
+            for rid_raw, status_raw in result.result["data"].items():
+                status = const.ResourceState[status_raw]
+                rra = cads.get(rid_raw, None)
```
Hugo-Inmanta, line 578: Is there a specific scenario where a resource id would be returned by `resources_status` but not be present in the `cads` dict ?

Hugo-Inmanta (changes requested):

src/inmanta/server/services/resourceservice.py
wouterdb: that produces a type error on the @handle annotation, so I had to choose

src/inmanta/agent/agent.py
wouterdb: mhh, right. That shouldn't happen (the other way around can happen)

Hugo-Inmanta (approved):

src/inmanta/data/__init__.py
```diff
@@ -4621,6 +4621,22 @@ def convert_or_ignore(rvid):
...
+        cls,
+        env: uuid.UUID,
+        model_version: int,
+        rids: list[ResourceIdStr],
+    ) -> dict[ResourceIdStr, ResourceState]:
+        query = (
+            "SELECT r.resource_id, r.status FROM resource r"
+            " WHERE r.environment=$1"
+            " AND r.model=$2"
+            " AND r.resource_id = ANY($3);"
```
arnaudsjs, line 4635:
```suggestion
            """
            SELECT r.resource_id, r.status
            FROM resource r
            WHERE r.environment=$1
                AND r.model=$2
                AND r.resource_id = ANY($3);
            """
```

src/inmanta/data/__init__.py
```diff
@@ -4621,6 +4621,22 @@ def convert_or_ignore(rvid):
...
         return out
 
+    @classmethod
+    async def get_status_for(
+        cls,
+        env: uuid.UUID,
+        model_version: int,
+        rids: list[ResourceIdStr],
+    ) -> dict[ResourceIdStr, ResourceState]:
+        query = (
```
arnaudsjs, line 4631:
```suggestion
    ) -> dict[ResourceIdStr, ResourceState]:
        if not rids:
            return {}
        query = (
```

src/inmanta/agent/agent.py
```diff
@@ -888,6 +840,8 @@ def cancel(self) -> None:
         """
         for ra in self.generation.values():
             ra.cancel()
+        if self.cad_resolver is not None:
+            self.cad_resolver.cancel()
```
arnaudsjs, line 844:
```suggestion
        if self.cad_resolver is not None:
            self.cad_resolver.cancel()
            self.cad_resolver = None
```

arnaudsjs (changes requested):

arnaudsjs (approved):

wouterdb:
take over by  #7315

PR #7311, opened 2024-03-06
issue/7289-include-orphans

### LINKED ISSUE #7289: Include orphans in db migration script for resource_persistent_state table
[This commits](https://github.com/inmanta/inmanta-core/commit/6c8d45438e67c46e3d6c6b42eb98ed24b4b9cf27#diff-4f3f01b61c14c23e4d081d8d85f4d632a0bda6e770111e15900ffa89723c17f6) added the `resource_persistent_state` database table. It contains a database migration script to populate this table, but the script doesn't include any orphans. This ticket should make sure that the orphans are migrated as well.

FloLey:
# Description

[This commits](https://github.com/inmanta/inmanta-core/commit/6c8d45438e67c46e3d6c6b42eb98ed24b4b9cf27#diff-4f3f01b61c14c23e4d081d8d85f4d632a0bda6e770111e15900ffa89723c17f6) added the resource_persistent_state database table. It contains a database migration script to populate this table, but the script doesn't include any orphans. fix the migration script to include them.

closes #7289

# Self Check:

- [x] Changelog entry


tests/db/migration_tests/dump_tool.py
```diff
@@ -200,6 +201,15 @@ async def test_dump_db(server, client, postgres_db, database_name):
     )
     assert result.code == 200
 
+    # create an orphan resource
+    version = 2
+    path = "/tmp/test" + str(2)
+    key = "std::File[localhost,path=" + path + "]"
+    res2_v1 = data.Resource.new(
```
wouterdb, line 208: this setup should always use the API, never the DB directly. In the next version, the single resource is not valid (it always would have a corresponding presistent state) . To ensure consistent data, the actual code should make it.

I think making the model respond to an env var and calling `request_recompile` on the `CompilerService` directly is the easiest way.

src/inmanta/db/versions/v202312190.py
```diff
@@ -35,13 +35,50 @@ async def update(connection: Connection) -> None:
...
 
-INSERT INTO public.resource_persistent_state
- (environment, resource_id, last_deploy, last_success, last_non_deploying_status, last_produced_events)
- SELECT environment, resource_id, last_deploy, last_success, last_non_deploying_status, last_produced_events
- FROM public.resource
- WHERE (environment, model) IN (
-    SELECT environment, max(version) FROM public.configurationmodel WHERE released=true GROUP BY environment
- );
+
+-- The INSERT query uses a subquery to find the maximum model version
```
wouterdb, line 39: I can't judge the performance of this, @arnaudsjs

wouterdb (changes requested):

arnaudsjs (changes requested):

src/inmanta/db/versions/v202312190.py
arnaudsjs: I ran the SELECT part of the query on mgmt. It executes in about 2.3 seconds. The query plan isn't great but I guess it's just in the nature of the query. I don't really see a way to make it perform better. 2.3s is also not bad for a migration script.

```
                                                                             QUERY PLAN                                                                              
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Gather  (cost=53382.12..117907.04 rows=32 width=154) (actual time=576.725..675.391 rows=4508 loops=1)
   Workers Planned: 2
   Workers Launched: 2
   ->  Hash Join  (cost=52382.12..116903.84 rows=13 width=154) (actual time=579.598..663.079 rows=1503 loops=3)
         Hash Cond: ((r.environment = resource.environment) AND ((r.resource_id)::text = (resource.resource_id)::text) AND (r.model = (max(resource.model))))
         ->  Parallel Seq Scan on resource r  (cost=0.00..63085.62 rows=182362 width=154) (actual time=0.012..57.042 rows=146231 loops=3)
         ->  Hash  (cost=52148.81..52148.81 rows=13332 width=93) (actual time=575.214..575.219 rows=4508 loops=3)
               Buckets: 16384  Batches: 1  Memory Usage: 689kB
               ->  HashAggregate  (cost=51882.17..52015.49 rows=13332 width=93) (actual time=572.881..573.859 rows=4508 loops=3)
                     Group Key: resource.environment, resource.resource_id
                     Batches: 1  Memory Usage: 1425kB
                     Worker 0:  Batches: 1  Memory Usage: 1425kB
                     Worker 1:  Batches: 1  Memory Usage: 1425kB
                     ->  Nested Loop  (cost=0.55..48599.66 rows=437668 width=93) (actual time=0.046..461.675 rows=438694 loops=3)
                           ->  Seq Scan on configurationmodel  (cost=0.00..34.11 rows=211 width=20) (actual time=0.013..0.512 rows=207 loops=3)
                                 Filter: released
                           ->  Index Only Scan using resource_pkey on resource  (cost=0.55..223.02 rows=715 width=93) (actual time=0.016..2.055 rows=2119 loops=621)
                                 Index Cond: ((environment = configurationmodel.environment) AND (model = configurationmodel.version))
                                 Heap Fetches: 394218
 Planning Time: 1.729 ms
 Execution Time: 676.022 ms
```

tests/db/migration_tests/dump_tool.py
FloLey: I will need more information about this as I don't understand what I'm doing.
I tried to add this at the start of the test:
```
    await compilerslice.request_recompile(
        env=env1, force_update=False, do_export=False, remote_id=remote_id1, env_vars={"my_unique_var": "1"}
    )
```
and 
to change to model:
```
agent = std::AgentConfig(agentname=host.name, autostart=true)
host = std::Host(name="localhost", os=std::linux, requires=agent)
file = std::ConfigFile(host=host, path="/tmp/test", content="1234", requires=agent)
val = std::get_env(name="my_unique_var",default_value="0")
if val == "1":
    file2 = std::ConfigFile(host=host, path="/tmp/test_orphan", content="test", requires=agent)
end

```
I was expecting to see an additional File resource when the the env var is set to 1 but I see no new resource in the DB.

arnaudsjs: Your compile request has `do_export=False` set. This means don't export the resources to the server.

tests/data/simple_project/main.cf
```diff
@@ -1,3 +1,7 @@
 agent = std::AgentConfig(agentname=host.name, autostart=true)
 host = std::Host(name="localhost", os=std::linux, requires=agent)
 file = std::ConfigFile(host=host, path="/tmp/test", content="1234", requires=agent)
+val = std::get_env(name="my_unique_var",default_value="0")
```
wouterdb, line 4: I would name the env var `add_one_resource` as this is not very descriptive naming 
And use 'true' as a value

wouterdb (approved):
One comment left

arnaudsjs (changes requested):
Do you also make sure to check the return codes of the API calls done in the dump_tool.py?

tests/db/migration_tests/dump_tool.py
```diff
@@ -80,18 +80,19 @@ async def populate_facts_and_parameters(client, env_id: str):
             "id": "parameter3",
             "source": const.ParameterSource.fact,
             "value": "value3",
-            "expires": True,
+            "expires": False,
```
FloLey, line 83: This is a mistake, parameters cannot expire

tests/db/migration_tests/dump_tool.py
```diff
@@ -157,33 +159,41 @@ async def test_dump_db(server, client, postgres_db, database_name):
...
+        )
     )
 
     await populate_facts_and_parameters(client, env_id_1)
 
     await wait_for_version(client, env_id_1, env_1_version)
 
-    await client.release_version(
-        env_id_1, env_1_version, push=False, agent_trigger_method=const.AgentTriggerMethod.push_full_deploy
+    env_1_version += 1
```
FloLey, line 186: otherwise the next call fails as the version is already released.

arnaudsjs: I think we have to move the `env_1_version += 1` up until before the `await wait_for_version(client, env_id_1, env_1_version)` call, no?

arnaudsjs (changes requested):

tests/db/migration_tests/dump_tool.py
FloLey: When I do that it fails. But I don't think this wait_for_version is required

arnaudsjs: Why does it fail?

arnaudsjs (approved):

PR #7313, opened 2024-03-07
Issue/1888 add dedicated doc page for config requirements

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/irt/issues/1888

- Add the `configuration_requirements.rst` file that reads json data from the `compatibility.json` generated during irt build-docs.

# Self Check:

- [x] Changelog entry


docs/configuration_requirements.rst
```diff
@@ -0,0 +1,19 @@
+Configuration requirements
```
sanderr, line 1: I find the term "configuration requirements" not very intuitive (same for the variable name in irt and the occurence in the text below). I do have to confess I'm having trouble coming up with a good title. "Compatible versions" or "Version constraints" are the best I can think of right now, but I don't really like either of them very much. Will give this some more thought.

docs/configuration_requirements.rst
```diff
@@ -0,0 +1,19 @@
...
+    These constraints are defined for the whole lifetime of this major version i.e.
+    they are not pinned to a specific three-digit version.
+
+
+    .. datatemplate:json:: required_configuration.json
+       :template: iso_config_requirements.tmpl
+
+.. only:: oss
+
+    This page shows the configuration requirements for version |version_major| of Inmanta.
```
sanderr, line 16: For OSS the major digit doesn't really have a special semantic meaning. I think we should use the full version here.

docs/configuration_requirements.rst
```diff
@@ -0,0 +1,19 @@
...
+
+    .. datatemplate:json:: required_configuration.json
+       :template: oss_config_requirements.tmpl
```
sanderr, line 19: Why not split it in two templates? One for the "system requirements" (Python, RHEL, ...) and one for the components? Then we can put the latter in a `.. only:: iso` block but the former would be identical.

docs/configuration_requirements.rst
```diff
@@ -0,0 +1,19 @@
...
+**************************
+
+.. only:: iso
+
+    This page shows the configuration requirements for the Inmanta Service Orchestrator |version_major| version.
+    These constraints are defined for the whole lifetime of this major version i.e.
+    they are not pinned to a specific three-digit version.
+
+
+    .. datatemplate:json:: required_configuration.json
```
sanderr, line 11: What if this file does not exist? We probably need an empty one in this repo so we can still build the docs stand-alone?

docs/configuration_requirements.rst
```diff
@@ -0,0 +1,19 @@
+Configuration requirements
+**************************
+
+.. only:: iso
+
+    This page shows the configuration requirements for the Inmanta Service Orchestrator |version_major| version.
+    These constraints are defined for the whole lifetime of this major version i.e.
+    they are not pinned to a specific three-digit version.
```
sanderr, line 8: We should mention somewhere that there is a machine-consumable version of these tables at \<insert link>. I think we may even want to link to both the json and the plain requirements file.

sanderr (changes requested):

docs/configuration_requirements.rst
Hugo-Inmanta: What about:
Title : "Compatibility"
In text: "Compatible versions"

sanderr: Sounds OK to me. Might be a good idea to quickly demo this next week. If anyone doesn't like the section name we'll hear it then.

arnaudsjs (changes requested):

docs/conf.py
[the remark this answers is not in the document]
Hugo-Inmanta: Woops, yes we do. I'll add them back

docs/_templates/system_requirements.tmpl
```diff
@@ -0,0 +1,25 @@
+.. -*- mode: rst -*-
+
+
+System requirements
+~~~~~~~~~~~~~~~~~~~
+
+
+{% if data['system_requirements']%}
+.. list-table:: System requirements
+   :widths: 25 25
```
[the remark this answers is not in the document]
Hugo-Inmanta: > What happens if you leave sphinx to do this based on the content of the table?

Works fine and looks better, I'll remove these hard coded values :+1:

docs/configuration_requirements.rst
```diff
@@ -0,0 +1,23 @@
...
+    they are not pinned to a specific three-digit version. Upper bounds may be added when a new major version is released.
+
+
+.. only:: oss
+
+    This page shows the compatibility of Inmanta version |release| with other components.
+
+
+.. datatemplate:json:: compatibility.json
+   :template: system_requirements.tmpl
```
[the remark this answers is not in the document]
arnaudsjs: I see.

docs/_templates/components_requirements.tmpl
```diff
@@ -0,0 +1,29 @@
...
+
+   {% for key, value in data["component_requirements"] | dictsort %}
+   * - {{ key|e }}
+     - {{ value|e }}
+   {% endfor %}
+
+.. note::
+
+    This information is also available as a constraints file in `requirements.components.txt <./requirements.components.txt>`_.
+    and in a machine-consumable format in the `compatibility.json <./compatibility.json>`_ file.
```
arnaudsjs, line 23:
```suggestion
    This information is also available as a constraints file in `requirements.components.txt <./requirements.components.txt>`_
    and in a machine-consumable format in the `compatibility.json <./compatibility.json>`_ file.
```

arnaudsjs (changes requested):

docs/_templates/components_requirements.tmpl
arnaudsjs, line 23: A requirements file is also a machine-consumable format. But just for a specific tool, namely pip. 
```suggestion
    and in json format in the `compatibility.json <./compatibility.json>`_ file.
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #7316, opened 2024-03-07
Improve executor test

FloLey:
# Description

make sure lorem is not installed at the start of the test.

# Self Check:

- [x] Changelog entry


FloLey:
@wouterdb I had a look at the existing tests that use lorem but I don't see which one could install it. I propose to add this to pinpoint the issue.

PR #7317, opened 2024-03-07
prevent cross environment deletes

wouterdb:
# Description

Fix cleanup query

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -210,6 +210,8 @@ async def test_create_too_many_versions(client, server, n_versions_to_keep, n_ve
     result = await client.set_setting(tid=env_1_id, id=data.AVAILABLE_VERSIONS_TO_KEEP, value=n_versions_to_keep)
     assert result.code == 200
 
+    # make a second environment to be sure we don't do cross env deletes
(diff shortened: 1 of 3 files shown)


tests/test_server.py
```diff
@@ -210,6 +210,8 @@ async def test_create_too_many_versions(client, server, n_versions_to_keep, n_ve
     result = await client.set_setting(tid=env_1_id, id=data.AVAILABLE_VERSIONS_TO_KEEP, value=n_versions_to_keep)
     assert result.code == 200
 
+    await client.create_environment(project_id=project_id, name="env_2")
```
arnaudsjs, line 213: This doesn't add any value to the test case if it doesn't contain any resources, right?

arnaudsjs (changes requested):

tests/test_server.py
wouterdb: it does, this is how I reproduced it.

bartv: then add a comment!

arnaudsjs (approved):

PR #7359, opened 2024-03-13
Issue/1638 race condition in compiles on a state with set state transition

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/inmanta-lsm/issues/1638

# Self Check:

- [x] Changelog entry


src/inmanta/protocol/methods_v2.py
```diff
@@ -71,6 +72,9 @@ def put_partial(
               * resources: a list of resource objects. Since the version is not known yet resource versions should be set to 0.
               * version_info: Model version information
     :param pip_config: Pip config used by this version
+    :param soft_delete: When True, ignore deletion of resources in removed_resource_sets if they are also being exported
+        (i.e. also present in resource_sets). When False, attempting to delete a resource being exported will raise an
+        exception.
```
sanderr, line 77: I think we can save some complexity by making the soft delete purely the responsibility of the exporter, and leave the `put_partial` endpoint untouched. But it might be good to ask the opinion of the rest of the team.

src/inmanta/export.py
```diff
@@ -505,6 +512,7 @@ def commit_resources(
         partial_compile: bool,
         resource_sets_to_remove: list[str],
         pip_config: PipConfig,
+        soft_delete: bool,
```
sanderr, line 515: *If* we decide to make this the responsibility of the exporter rather than `put_partial`, as suggested in another comment, I think we can even lift the logic to the `run` method.

changelogs/unreleased/1638-add-soft-delete-option.yml
```diff
@@ -0,0 +1,8 @@
+description: >
+    Add ``soft_delete`` option to allow compiles to silently prevent resource
+    set deletion if the model is exporting resources included in this set.
+issue-nr: 1638
+issue-repo: inmanta-lsm
+change-type: minor
+destination-branches: [master, iso7]
```
sanderr, line 7: I think this bug also exists in iso6. I'm not 100% sure, did you check?

src/inmanta/app.py
```diff
@@ -455,6 +455,13 @@ def export_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
         "be used together with the --partial option.",
         action="append",
     )
+    parser.add_argument(
+        "--soft-delete",
```
sanderr, line 459: I'm still thinking if I can come up with a better name. This one was always meant as a placeholder, but perhaps it's not too bad.

tests/agent_server/test_resource_sets.py
sanderr, line 1: Haven't reviewed these yet, depends a tiny bit on whether we consider this the responsibility of exporter or server.

sanderr (changes requested):

src/inmanta/app.py
Hugo-Inmanta: I couldn't come up with something better and it kinda grew on me.

changelogs/unreleased/1638-add-soft-delete-option.yml
```diff
@@ -3,6 +3,6 @@ description: >
     set deletion if the model is exporting resources included in this set.
 issue-nr: 1638
 issue-repo: inmanta-lsm
-change-type: minor
-destination-branches: [master, iso7]
+change-type: patch
```
sanderr, line 6: Right, hadn't considered that it would have to be a patch to go in iso6. Still, it's more of a bugfix addition than a real feature.

src/inmanta/export.py
```diff
@@ -469,14 +478,26 @@ def add_resource(self, resource: Resource) -> None:
 
         self._resources[resource.id] = resource
 
-    def resources_to_list(self) -> list[dict[str, Any]]:
-        """Convert the resource list to a json representation"""
+    def resources_to_list(self) -> tuple[list[dict[str, Any]], set[str]]:
+        """
+        Convert the resource list to a json representation
+        If the soft_delete option is set, this will also compute the set of resource sets containing these resources
```
sanderr, line 484: I think this could be a lot simpler if you put it in `_load_resource_sets`, or am I missing something. We could even go so far as to add an attribute `self._resource_sets: set[str]` because that method already works by populating an attribute rather than returning its result.

src/inmanta/protocol/methods_v2.py
```diff
@@ -71,6 +71,9 @@ def put_partial(
...
```
sanderr, line 76: This should go now.

sanderr (changes requested):

sanderr (approved):
Can you still request a second review?

changelogs/unreleased/1638-add-soft-delete-option.yml
```diff
@@ -0,0 +1,8 @@
+description: >
+    Add ``soft_delete`` option to the exporter, to allow compiles to silently prevent resource
+    set deletion if the model is exporting resources included in this set.
```
arnaudsjs, line 3: I don't think this sentence will make a lot of sense to anyone who is no part of the core team.
```suggestion
    Add `--soft-delete` option to the exporter. This option makes sure that resource sets, specified using the `--delete-resource-set` option, are only deleted if there are no resources exported for this set.
```

tests/db/migration_tests/test_v202403010_to_v202403120.py
```diff
@@ -0,0 +1,36 @@
...
+part = file_name_regex.match(__name__)[1]
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), f"dumps/v{part}.sql"))
+async def test_drop_not_null_constraint(
+    postgresql_client: asyncpg.Connection,
+    migrate_db_from: abc.Callable[[], abc.Awaitable[None]],
+) -> None:
+    # This migration script adds a column. Just verify that the script doesn't fail.
+    await migrate_db_from()
```
arnaudsjs, line 36: Shouldn't we test that `soft_delete` is false on a compiles after the migration?

arnaudsjs (changes requested):

arnaudsjs (changes requested):

docs/model_developers/resource_sets.rst
```diff
@@ -98,6 +98,10 @@ Two arguments can be passed to the ``inmanta export`` command in order to export
 - ``--partial`` To specify that the model being compiled only contains the resources that need to be updated in relation to the previous version of the model.
 - ``--delete-resource-set <resource-set-name>`` This option, which may be used more than once, instructs the model to remove the resource set with the specified name. Only in conjunction with the preceding choice may this option be utilized. Note that utilizing a ``std::ResourceSet`` that includes no resources allows resource sets to be implicitly deleted during a partial compilation.
 
+.. note::
```
arnaudsjs, line 101: I don't see why we need to the note box here. It used to attract the attention of the user to something special. This bit is not something special is it?

arnaudsjs (changes requested):

PR #7365, opened 2024-03-14
issue/1630-extend-instances-with-deployment-progress

FloLey:
# Description

add get_aggregated_deployment_status for https://github.com/inmanta/inmanta-lsm/pull/1655

closes https://github.com/inmanta/inmanta-lsm/issues/1630

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/1484-in_line_compile_listener.yml
+++ b/changelogs/unreleased/1484-in_line_compile_listener.yml
@@ -1,0 +1,5 @@
+description: Added inline compile listeners
+issue-nr: 1484
+issue-repo: inmanta-lsm
+change-type: minor
+destination-branches: [master, iso7]
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -64,14 +64,21 @@ COMPILER_LOGGER: Logger = LOGGER.getChild("report")
 class CompileStateListener:
     @abc.abstractmethod
     async def compile_done(self, compile: data.Compile) -> None:
-        """Receive notification of all completed compiles
+        """
+        Coroutine to execute when this listener is notified of a completed compile
 
         1- Notifications are delivered at least once (retry until all listeners have returned or raise an exception other
         than CancelledError)
-        2- Notification are delivered out-of-band (i.e. the next compile can already start, multiple notifications can be in
-        flight at any given time, out-of-order delivery is possible but highly unlikely)
-        3- Notification are cancelled upon shutdown
+        2- If this listener is blocking, the next compile can only start after this coroutine is done
+        3- Multiple notifications can be in flight at any given time, out-of-order delivery is possible but highly unlikely.
+        4- This coroutine is cancelled upon shutdown
         """
+
+    def is_blocking(self) -> bool:
+        """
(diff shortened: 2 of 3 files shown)


arnaudsjs (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -4483,6 +4483,33 @@ async def trim(cls, environment: UUID, connection: Optional[Connection] = None)
...
+        SELECT last_non_deploying_status, COUNT(*) AS status_count
+        FROM {cls.table_name()}
+        WHERE environment = $1 AND resource_id = ANY($2)
+        GROUP BY last_non_deploying_status
+        """
+        values = [environment_id, resource_ids]
+
+        status_counts = {}
+        async with cls.get_connection() as con:
+            async with con.transaction():
```
wouterdb, line 4505: for this one, it may be more interesting to use [fetch](https://magicstack.github.io/asyncpg/current/api/index.html#asyncpg.connection.Connection.fetch) instead of a cursor? (there will never be many records in the result)

wouterdb (approved):

src/inmanta/data/__init__.py
FloLey: It used to be like this:
```
        query = f"""
        SELECT last_non_deploying_status, COUNT(*) AS status_count
        FROM {cls.table_name()}
        WHERE environment = $1 AND resource_id = ANY($2)
        GROUP BY last_non_deploying_status
        """
        results = await cls._fetch_query(query, environment_id, resource_ids)

        status_counts = {
            cast(str, record["last_non_deploying_status"]): cast(int, record["status_count"]) for record in results
        }
        return status_counts
```
 @arnaudsjs thought a cursor would be better. I have no idea how many records there will be and what we consider to be many.

src/inmanta/data/__init__.py
```diff
@@ -4483,6 +4483,38 @@ async def trim(cls, environment: UUID, connection: Optional[Connection] = None)
             connection=connection,
         )
 
+    @classmethod
+    async def get_aggregated_deployment_status(
+        cls, environment_id: UUID, resource_ids: list[ResourceIdStr]
+    ) -> dict[str, dict[const.NonDeployingResourceState, int]]:
```
arnaudsjs, line 4489:
```suggestion
    ) -> dict[ResourceIdStr, dict[const.NonDeployingResourceState, int]]:
```

arnaudsjs, line 4489: We should really prevent these kinds of datastructures with deeply nested dicts. This makes the code so much harder to read. Why not use an object instead?

src/inmanta/data/__init__.py
arnaudsjs: The default prefetch size of a cursor is 50 elements. So as far as I can see we will fetch everything in a single batch if the number of records is low. But I have no strong objections against using fetch(), so if Wouter prefers that one, it's fine for me as well.

wouterdb: The method is now different, so I don't think the fetch is still a benefit (it used to return < 10 resources, now it is a lot more)

PR #7380, opened 2024-03-18
Issue/667 document allocation v3

Hugo-Inmanta:
# Description

Documentation PR for allocation v3

Closes https://github.com/inmanta/lsm/issues/667

# Self Check:

- [x] Changelog entry


docs/lsm/allocation/allocation_sources/allocation_v3/complex_example/v3_plugin.py
```diff
@@ -0,0 +1,25 @@
+@allocation_helpers.allocator()
+def get_free_value_in_range(
```
edvgui, line 2: The lsm module already has an equivalent allocator, it would be good to illustrate this one instead.

docs/lsm/allocation/allocation_sources/allocation_v3/complex_example/v3_main.cf
```diff
@@ -0,0 +1,26 @@
...
+    self.vlan_id = get_free_value_in_range(self, "vlan_id", lower=50000, upper=70000)
+end
+
+implement VlanAssignment using parents, get_available_vlan
+
+vlan_binding = lsm::ServiceEntityBindingV2(
+    service_entity="VlanAssignment",
+    lifecycle=lsm::fsm::simple,
+    service_entity_name="vlan-assignment",
+    allocation_spec="allocate_vlan",
```
edvgui, line 17: For allocation v3 you don't need an allocation_spec anymore, so it is better not to use it

docs/lsm/allocation/allocation_sources/allocation_v3/basic_example/v3_plugin.py
```diff
@@ -0,0 +1,15 @@
...
+    attribute_path: "string",
+) -> "int":
+    return 1
+
+@allocation_helpers.allocator()
+def get_embedded_value(
+    service: "lsm::ServiceEntity",
+    attribute_path: "string",
+) -> "int":
+    return 3
```
edvgui, line 13: It would be nice to illustrate more what these allocators do, and when they run.  Something like
```python
@allocation_helpers.allocator()
def get_value(
    service: "lsm::ServiceEntity",
    attribute_path: "string",
    *,
    value: "any",
) -> "any":
    """
    Store a given value in the attributes of a service.

    :param value: The value to store on the service.
    """
    print("This will be called in the first validation compile only!", str(service), str(attribute_path), str(value))
    return value
```

And then you can call this for both the top level value and the embedded value

docs/lsm/allocation/allocation_v3.rst
```diff
@@ -0,0 +1,175 @@
+**************
+Allocation V3
+**************
+
+
+Allocation V3 is a new framework that changes significantly compared to Allocation V2. It doesn't have
+its own lifecycle stage anymore, but instead allocation happens during validation compile. The purpose is
```
edvgui, line 7: > It doesn't have its own lifecycle stage anymore, but instead allocation happens during validation compile.

That is not true, nothing changed lifecycle-wise.  For any type of allocation, there needs to be a validating state, in which a validation compile is run.

docs/lsm/allocation/allocation_v3.rst
```diff
@@ -0,0 +1,175 @@
...
+        service: "lsm::ServiceEntity",
+        attribute_path: "string",
+    ) -> "int":
+        return 5
+
+
+An allocator must accept exactly two positional arguments:
+    1. ``service``, the service instance for which the value is being allocated (usually ``self`` in the model).
+    2. ``attribute_path``, the attribute of the service instance in which the allocated
+    value should be saved. The decorated function can define a default value.
```
edvgui, line 38:
```suggestion
    2. ``attribute_path``, the attribute of the service instance in which the allocated
    value should be saved, as a dict_path expression. The decorated function can define a default value.
```

edvgui (changes requested):
I think it gives a clear picture overall.  I have one general comments: it would be good to document (with docstrings and inline comments) all the examples we provide, for two reasons:
1. They might look self-explanatory for us, but they will not be for everyone.
2. This code will sit in the documentation, it should be exemplary, and follow the best guidelines we have, to write the best model we can.  That includes documenting the code.

docs/lsm/allocation/allocation_sources/allocation_v3/basic_example/v3_main.cf
```diff
@@ -0,0 +1,81 @@
...
+entity TopLevelService extends lsm::ServiceEntity:
+    """
+    This service entity demonstrates how a single allocator
+    can be used for both a service entity and its embedded
+    entities.
+    """
+
+    string                      name
+    lsm::attribute_modifier     name__modifier="rw"
+    int?                        top_level_value
```
edvgui, line 20: I thought we always needed to have a default, null, value for the allocated attributes, isn't it the case? Which value will the api assign when the service is created by the user otherwise?

docs/lsm/allocation/allocation_sources/allocation_v3/basic_example/v3_main.cf
```diff
@@ -0,0 +1,81 @@
...
+        entity_binding=top_level_service_binding,
+        name=attributes["name"],
+    )
+
+    for embedded_service in attributes["embedded_services"]:
+        service.embedded_services += EmbeddedService(
+            id=embedded_service["id"],
+        )
+    end
+end
```
edvgui, line 81:
```suggestion


top_level_service_binding = lsm::ServiceEntityBindingV2(
    service_entity="allocatorv3_demo::TopLevelService",
    lifecycle=lsm::fsm::simple,
    service_entity_name="top-level-service",
    service_identity="name",
    service_identity_display_name="Name",
)

for assignment in lsm::all(top_level_service_binding):
    attributes = assignment["attributes"]

    service = TopLevelService(
        instance_id=assignment["id"],
        entity_binding=top_level_service_binding,
        name=attributes["name"],
        top_level_value=get_value(self, "top_level_value", value=1),
        embedded_services=[
            EmbeddedService(
                id=embedded_service["id"],
                embedded_value=get_value(
                    self,
                    lsm::format(
                        "embedded_services[id={embedded_service_id}].embedded_value",
                        [],
                        {"embedded_service_id": embedded_service["id"]},
                    ),
                    value=3,
                ),
            )
            for embedded_service in attributes["embedded_services"]
        ],
    )
end
```

This is more inline with how we would do it, and might actually be the only way (depending on the response to my comment on default values)

edvgui (changes requested):
Still a lot of docstrings to add/complete, can you document the attributes, relations, implementation, etc?

docs/lsm/allocation/allocation_sources/allocation_v3/basic_example/v3_main.cf
sanderr: The API will always default to `None` for nullable attributes, if no explicit default is given. I'm not sure how I feel about that but it is the current behavior as far as I can tell.

edvgui: I partially agree and partially don't, here is some more food for thoughts:
1. I agree that "nullable" and "optional" are different concepts.  I associate "optional" to "has a default value".  Assuming any defined attribute should have a value, if you allow (or force in case of `ro` attribute) the user to not specify a value, making the attribute optional in the input, then that attribute should define a default, which you can then use when the entity is constructed (in the api).  In some cases (talking about javascript here), a defined attribute might not have a value (`undefined`), but this is not a thing that exists in our DSL, and I think we should not make it a thing in our API as it is defined in the DSL.
2. I agree that defaults complicates things in the DSL in general, but I would say it is not the case for lsm entities, as whenever we construct them, we will always have all their attributes at our disposal (as we read them from the api, or from an allocator).
3. Regarding the `nullable` aspect of allocated values, I would say it should indeed not be required, but I believe that `null` is a generic enough default for a value that is not provided (yet).  And it is actually a convention we expect users to follow when using the `allocator` decorator, as this is the "no value" we check for, before calling the decorated plugin.  In the future, we will introduce the `reallocator` decorator, which will be called in any validating compile, providing the previous value that was set in the api.  With this mechanism, the user could define another default (i.e. `"unknown"`) that the decorated plugin can recognize and replace with the real allocated value.

sanderr: 3. Doesn't this ever give trouble in the model? e.g. suppose you have a service that writes files, but the file name is allocated. It would be an invariant of the model that the file name can never be `null`: either it's provided through the API (`lsm::all`) or it is allocated. Yet you can not express this invariant in the model because you have to set the type to `string?`.

sanderr: 1. I was thinking more of a value being absent from the dict, than a special API value.

sanderr: 1. Then again, perhaps a better practice would be to model the decoupling that I describe directly in the model: to have an API entity layer, and a model entity layer right below that. Like you might have a Pydantic layer for boundary data validation, but use a more stricly typed object for internal representation.

edvgui: > either it's provided through the API (lsm::all) or it is allocated  
If an attribute is writable by the user, than it is not read-only, and can therefore not be allocated.  cf. a conversation we had in the solutions team regarding this recently: https://inmanta.slack.com/archives/C01QH7140NT/p1710930606322199?thread_ts=1710930381.452039&cid=C01QH7140NT

> Yet you can not express this invariant in the model because you have to set the type to string?
That is true, you have to check yourself in the model if your service was created correctly, which you trust it is if you are in a state that passed validation.

> I was thinking more of a value being absent from the dict, than a special API value.
Which is exactly what `undefined` is in javascript (so in JSON as well), the issue is that you can not represent this in the model, so you can face a situation where the service of the api can not be represented in the model.

sanderr: e.g.

```
entity APIEntity:
    int? allocated_field = null
    lsm::attribute_modifier allocated_field__modifier="r"
end

entity ModelEntity:
    int allocated_field
end


for entity in lsm::all(entity_binding):
    APIEntity(allocated_field=entity["attributes"]["allocated_field"])
end

implementation imp for APIEntity:
    ModelEntity(
        allocated_field=self.allocated_field is defined ? self.allocated_field is defined : allocate(...)
    )
end
```

Major disadvantage: quite some duplication. But perhaps we can find a middle ground somehow. I don't see it yet.

edvgui: > I don't mean writable by the user, I mean provided by the lsm read API, because it was allocated in a previous state.

Okay, I misunderstood you.

> So you end up with static type violations all around

Yep, big downer for the static typing I agree.  But we never need to check everywhere in the model for it, because the value that will be assigned always comes out of the allocator, which will either return a valid allocated value, or `Unknown` if allocation failed previously (unknown value which we only meet in a deallocating state then, where the service is usually not even constructed in the model).

With all of this, I agree that the "nullable" notation of read-only attributes is not ideal, I would prefer if they were not typed like this in the model.  I do still think it would be better to have them typed "nullable" in the api, with a default of `null` than having them sometimes missing from the attributes, but this is a personal opinion.

Coming back to how things are now, and about the remark I made.  I didn't know that the api always added the default `null` value, even if we didn't provided it.  In that case, I see how the `null` default is not required.  To be honest, I am not a fan of it and the "allocation implementation" approach it goes with.  For one thing, it makes it impossible to create the service differently than via the lsm api (one existing use case we have for that is the translation of a service into another, i.e. connect l2Connect and connect CarrierEthernetEvc, two different apis for the same network configuration), and also it doesn't offer the nice separation of DSL domain vs LSM domain (allocation rather belongs in the lather).

sanderr: >  I do still think it would be better to have them typed "nullable" in the api, with a default of null than having them sometimes missing from the attributes, but this is a personal opinion.

I think I agree by now, but I have to consider it some more.

> Coming back to how things are now, and about the remark I made.

You make a good point. Especially combined with the fact that we're both leaning towards explicit defaults in the API in the future, I think it's probably best to stick with explicit model defaults in the model for now, at least in example code, which will establish the best practice.

docs/lsm/allocation/allocation_sources/allocation_v3/basic_example/v2_main.cf
```diff
@@ -0,0 +1,81 @@
...
+
+# Each top level service may have zero or more embedded services attached to it
+TopLevelService.embedded_services [0:] -- EmbeddedService
+
+entity EmbeddedService extends lsm::EmbeddedEntity:
+    """
+    An embedded service, attached to a TopLevelService instance.
+
+    :attr id: The identifier of this embedded service instance, it should be unique
+        across all other embedded service instances of the service it is attached to.
```
edvgui, line 35: This is not aligned with the index behavior, the index says it should be unique across all instances of all services

docs/lsm/allocation/allocation_sources/allocation_v3/basic_example/v2_main.cf
```diff
@@ -0,0 +1,81 @@
...
+        entity_binding=value_binding,
+        name=attributes["name"],
+        top_level_value=attributes["top_level_value"],
+    )
+
+    for embedded_service in attributes["embedded_services"]:
+        service.embedded_services += EmbeddedService(
+            **embedded_service
+        )
+    end
```
edvgui, line 80: Why not use the list comprehension and the constructor tree syntax here?

docs/lsm/allocation/allocation_sources/allocation_v3/ordering_example/plugin.py
```diff
@@ -0,0 +1,41 @@
...
+
+@allocator()
+def ordered_allocation(
+    service: "lsm::ServiceEntity",
+    attribute_path: "string",
+    *,
+    requires: "list?" = None
+) -> "string":
+    """
+    For demonstration purposes, this allocator returns the current time.
```
edvgui, line 32: Nice one!

changelogs/unreleased/667-document-allocation-v3.yml
```diff
@@ -0,0 +1,5 @@
+description: Add documentation for allocation v3
+issue-nr: 667
+issue-repo: lsm
+change-type: patch
+destination-branches: [master, iso7]
```
edvgui, line 5: I think it should either be a minor, or go into iso6, no?

edvgui (changes requested):
Starting to look good!

docs/lsm/allocation/allocation_sources/allocation_v3/basic_example/v3_main.cf
```diff
@@ -0,0 +1,93 @@
...
+                    lsm::format(
+                        "embedded_services[id={embedded_service_id}].embedded_value",
+                        [],
+                        {"embedded_service_id": embedded_service["id"]},
+                    ),
+                    value=3,
+                ),
+            )
+            for embedded_service in attributes["embedded_services"]
+        ],
```
edvgui, line 91:
```suggestion
        embedded_services=[
            EmbeddedService(
                id=embedded_service["id"],
                # Allocator call
                embedded_value=get_value(
                    service,
                    lsm::format(
                        "embedded_services[id={id}].embedded_value",
                        args=[],
                        kwargs=embedded_service,
                    ),
                    value=3,
                ),
            )
            for embedded_service in attributes["embedded_services"]
        ],
```
I saw this in a recent MR of @cukiernick and I think it is a lot more elegant and makes sense to be shown in the documentation

edvgui (approved):
Last minor comment

Hugo-Inmanta:
I removed `(usually ``self`` in the model).` in [this commit](https://github.com/inmanta/inmanta-core/pull/7380/commits/8cc73d27ca2336a7418d77525bfb569e14a44d02) because it is not reflected in the examples i.e. we allocate in the lsm unwrapping and not in an implementation where `self` would be used. 

I think I'm missing the subtlety between performing allocation in the unwrapping vs in the implementation

edvgui (approved):
I saw the change, I do agree with it, it slipped through my second review.  In the case where the allocation is not done in an implementation (the one we recommend), `self` does indeed not mean anything

PR #7390, opened 2024-03-19
Bugfix in wait condition for unstable test

sanderr:
# Description

Bugfix in wait condition for unstable test. The test case used to call `retry_limited` and attempted to force progress by calling `resource_container.wait_for_done_with_waiters` (#6565). However, this last method would always return immediately without notifying waiters because the current version already had all resources deployed. This PR changes the wait method to the one that waits on a custom condition instead. This method does not return immediately (because the condition we care about is not met), therefore it does notify any waiters.


arnaudsjs:
It feels rather strange to me the deploying two (empty) resources would take more than 10 seconds

sanderr:
You're right. I'm updating the description.

arnaudsjs (approved):

PR #7395, opened 2024-03-19
Issue/1484 in line compile listener

wouterdb:
# Description

Added  in line compile listener

part of https://github.com/inmanta/inmanta-lsm/issues/1484

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -64,13 +64,14 @@ COMPILER_LOGGER: Logger = LOGGER.getChild("report")
 class CompileStateListener:
     @abc.abstractmethod
     async def compile_done(self, compile: data.Compile) -> None:
-        """Receive notification of all completed compiles
+        """
+        Coroutine to execute when this listener is notified of a completed compile
 
         1- Notifications are delivered at least once (retry until all listeners have returned or raise an exception other
         than CancelledError)
-        2- Notification are delivered blocking or non-blocking (i.e. the next compile can already start or not)
+        2- If this listener is blocking, the next compile can only start after this coroutine is done
         3- Multiple notifications can be in flight at any given time, out-of-order delivery is possible but highly unlikely.
-        4- Notification are cancelled upon shutdown
+        4- This coroutine is cancelled upon shutdown
         """
 
     def is_blocking(self) -> bool:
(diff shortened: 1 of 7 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -68,11 +68,17 @@ async def compile_done(self, compile: data.Compile) -> None:
 
         1- Notifications are delivered at least once (retry until all listeners have returned or raise an exception other
         than CancelledError)
-        2- Notification are delivered out-of-band (i.e. the next compile can already start, multiple notifications can be in
-        flight at any given time, out-of-order delivery is possible but highly unlikely)
-        3- Notification are cancelled upon shutdown
+        2- Notification are delivered blocking or non-blocking (i.e. the next compile can already start or not)
+        3- Multiple notifications can be in flight at any given time, out-of-order delivery is possible but highly unlikely.
+        4- Notification are cancelled upon shutdown
```
Hugo-Inmanta, line 73: I find the terminology confusing here.

What about 
-`Receive notification of all completed compiles` -> `Coroutine to execute when this listener is notified of a completed compile`
- `Notification are delivered blocking or non-blocking (i.e. the next compile can already start or not)` -> `If this listener is blocking, the next compile can only start after this coroutine is done`
- `Notification are cancelled upon shutdown` -> `This coroutine is cancelled upon shutdown`

tests/server/test_compilerservice.py
```diff
@@ -232,6 +243,16 @@ async def request_compile(env: data.Environment) -> uuid.UUID:
...
 
+    async def get_compile(env: data.Environment, rid: uuid.UUID) -> data.Compile:
+        results = await data.Compile.get_by_remote_id(env.id, rid)
+        assert len(results) == 1
+        assert results[0].remote_id == rid
+        return results[0]
+
+    async def is_handled(env: data.Environment, rid: uuid.UUID) -> bool:
+        compile = await get_compile(env, rid)
+        return compile.handled
```
arnaudsjs, line 254: We normally use `rid` to refer to a resource_id. This can be confusing.
```suggestion
    async def get_compile(env: data.Environment, remote_id: uuid.UUID) -> data.Compile:
        results = await data.Compile.get_by_remote_id(env.id, remote_id)
        assert len(results) == 1
        assert results[0].remote_id == remote_id
        return results[0]

    async def is_handled(env: data.Environment, remote_id: uuid.UUID) -> bool:
        compile = await get_compile(env, remote_id)
        return compile.handled
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #7396, opened 2024-03-19
Cleanup code folder on agent restart

### LINKED ISSUE #7388: Stale code in agent's code directory for std module
# Action
- [x] drop the warning: it's too trigger-happy and we can't easily make it more accurate
- [x] mitigate the issue by clearing the agent code dir when the agent starts. This ensures that when a user restarts an agent, we start clean. It's not perfect, but acceptable considering that the issue will no longer be relevant with the iso8 agent rework.

# Motivation
All the development orchestrators now show this warning notification:
image

1. This is a bug: incremental deploys only deploy code for the resources in the increment. This makes the warning fire regularly for incremental deploys, while the scenario we're trying to detect has not occurred.
2. This is very spammy (why is it triggered 23 times here?)

wouterdb:
# Description

Cleanup code folder on agent restart

closes  #7388 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/7388-stale-code-warning.yml
+++ b/changelogs/unreleased/7388-stale-code-warning.yml
@@ -1,0 +1,6 @@
+description: Remove stale code warning
+issue-nr: 7388
+change-type: patch
+destination-branches: [master, iso7]
+sections:
+  minor-improvement: Ensure agent code folder is cleaned up on restart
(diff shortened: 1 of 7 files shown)


arnaudsjs (changes requested):
* We also need to remove the documentation that was added as part of https://github.com/inmanta/inmanta-core/commit/7e2819e27617cee72055ee6671c5de35e40bbdca
* The above-mentioned commit also added the `send_notification` API endpoint, that allows an agent to send_notifications. This PR removes the test case that verifies the behavior of this endpoint. So we should either remove the endpoint again or make sure it is properly tested.

arnaudsjs (approved):

Hugo-Inmanta (changes requested):
Changelog entry is missing, lgtm otherwise

Hugo-Inmanta (approved):

PR #7398, opened 2024-03-20
[WIP]

Hugo-Inmanta:
# Description

remove check in the cli options:

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -473,8 +473,6 @@ def export(options: argparse.Namespace) -> None:
             raise CLIException(
                 "The --delete-resource-set option should always be used together with the --partial option", exitcode=1
             )
-        if not options.partial_compile and options.soft_delete:
-            raise CLIException("The --soft-delete option should always be used together with the --partial option", exitcode=1)
         if options.environment is not None:
             Config.set("config", "environment", options.environment)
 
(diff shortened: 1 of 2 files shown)


src/inmanta/app.py
```diff
@@ -473,10 +473,8 @@ def export(options: argparse.Namespace) -> None:
             raise CLIException(
                 "The --delete-resource-set option should always be used together with the --partial option", exitcode=1
             )
-        if not options.delete_resource_set and options.soft_delete:
-            raise CLIException(
-                "The --soft-delete option should always be used together with the --delete-resource-set option", exitcode=1
-            )
+        if not options.partial_compile and options.soft_delete:
+            raise CLIException("The --soft-delete option should always be used together with the --partial option", exitcode=1)
```
arnaudsjs, line 477: This will not be enough to fix the issue. The lsm extension always sets `soft_delete=True`. Also when no partial compile is done.

arnaudsjs (approved):

src/inmanta/app.py
Hugo-Inmanta: Hmm right, do we just remove this check ?

arnaudsjs: Fine by me.

PR #7405, opened 2024-03-20
Issue/6625 string concatenation

### LINKED ISSUE #6625: f-strings don't support dict lookup
F-strings only expect simple variables to do string interpolation, the goal of this ticket is to add support for dict lookup.

For example:
```
t = {"name": "a"}
std::print(f"Dict value: {t['name']}")
```
This would output  `variable t['name'] not found (reported in std::print(Format(Dict value: {t['name']}))`

sanderr:
# Description

Extend '+' operator to allow concatenating strings. Includes small refactor of the `ArithmeticOperator` class.

Closes #6625. This ticket was deemed infeasible with the current state of the parser. This PR provides a functionality that satisfies the same need.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/string-concatenation.yml
+++ b/changelogs/unreleased/string-concatenation.yml
@@ -1,0 +1,7 @@
+description: Add support for string concatenation in the Inmanta modelling language
+change-type: minor
+sections:
+  minor-improvement: "{{description}}"
+destination-branches:
+  - master
+  - iso7
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -147,6 +147,12 @@ Example:
 Note that the result of the division operation is cast to the type ``int``. This is done because a division always
 results in a value of type ``float``.
 
+In addition to arithmetic, the ``+`` operator is also supported for string concatenation.
+
+.. code-block:: inmanta
+
+    var = "hello " + "world"
+
 
 Primitive types
 ==============================
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -147,12 +147,6 @@ Example:
 Note that the result of the division operation is cast to the type ``int``. This is done because a division always
 results in a value of type ``float``.
 
-In addition to arithmetic, the ``+`` operator is also supported for string concatenation.
-
-.. code-block:: inmanta
-
-    var = "hello " + "world"
-
 
 Primitive types
 ==============================
@@ -334,6 +328,15 @@ included in:
 
     # Output when displayed:
     # Welcome to serv1.example.org
+
+String concatenation
+####################
+
+Strings can be concatenated with the ``+`` operator.
+
+.. code-block:: inmanta
+
+    hello_world = "hello " + "world"
 
 
 .. _lang-conditions:
--- a/src/inmanta/ast/constraint/expression.py
+++ b/src/inmanta/ast/constraint/expression.py
@@ -518,32 +518,33 @@ class ArithmeticOperator(BinaryOperator):
     # but it's a lot simpler and it suffices for what we really need
 
     def _bin_op(self, arg1: object, arg2: object) -> object:
-        return self.dispatch(arg1, arg2)
+        result: object = self._dispatch(arg1, arg2)
+        if result is NotImplemented:
+            raise TypingException(
+                self,
+                (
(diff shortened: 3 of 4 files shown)


src/inmanta/ast/constraint/expression.py
```diff
@@ -513,30 +514,63 @@ def _bin_op(self, arg1: object, arg2: object) -> object:
...
-        if not isinstance(arg2, (int, float)):
-            raise TypingException(self, f"The {self.get_name()} operator in not supported for {arg2} of type {type(arg2)}.")
-        return self._arithmetic_op(arg1, arg2)
+        result: object = self._dispatch(arg1, arg2)
+        if result is NotImplemented:
+            raise TypingException(
+                self,
+                (
+                    f"Unsupported operand type(s) for {self.get_name()}:"
+                    f" '{type(arg1).__name__}' ({repr(arg1)}) and '{type(arg2).__name__}' ({repr(arg2)})."
```
sanderr, line 528: Borrowed this error message and the one for string concatenation from Python.

arnaudsjs (approved):
Looks good to me. I also prefer the approach without the `singledispatch`.

wouterdb (approved):

src/inmanta/ast/constraint/expression.py
```diff
@@ -514,29 +514,57 @@ class ArithmeticOperator(BinaryOperator):
...
     def __init__(self, op1: ExpressionStatement, op2: ExpressionStatement) -> None:
         ArithmeticOperator.__init__(self, "plus", op1, op2)
 
-    def _arithmetic_op(self, arg1: int | float, arg2: int | float) -> int | float:
+    def _execute_operator(self, arg1: object, arg2: object) -> object:
+        if isinstance(arg1, str):
+            if isinstance(arg2, str):
+                return self._arithmetic_op(arg1, arg2)
+            # raise more representative error to override generic one from ArithmeticOperator
+            raise TypingException(self, f"Can only concatenate str (not '{type(arg2).__name__}' ({repr(arg2)})) to str")
```
arnaudsjs, line 564: We will only get this more specific exception message if arg1 is a string and arg2 isn't. But not the other way around. Is that intentional?

arnaudsjs (changes requested):

docs/language.rst
```diff
@@ -147,6 +147,12 @@ Example:
 Note that the result of the division operation is cast to the type ``int``. This is done because a division always
 results in a value of type ``float``.
 
+In addition to arithmetic, the ``+`` operator is also supported for string concatenation.
+
+.. code-block:: inmanta
+
+    var = "hello " + "world"
+
```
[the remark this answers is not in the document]
sanderr: Good point, will fix.

src/inmanta/ast/constraint/expression.py
sanderr: Sort of, but I'm open to suggestions. This happens to be how Python reports it and I followed that:
```ipython
>>> "hello" + 1
TypeError: can only concatenate str (not "int") to str
>>> 1 + "hello"
TypeError: unsupported operand type(s) for +: 'int' and 'str'
>>>
```

I feel like it makes sense from the point of view of `x + y` meaning "add `y` to `x`". The type of `y` would then determine the type of the operation, and the type of `x` would determine whether the operation can be applied.

changelogs/unreleased/string-concatenation.yml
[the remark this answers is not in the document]
sanderr: Forgot to respond. I left it out intentionally because, while this PR closes a ticket, it does so more by fulfilling the same need than by implementing the requested feature.

src/inmanta/ast/constraint/expression.py
arnaudsjs: I find this a bit strange. On the other hand, both error messages are rather clear with respect to what is going wrong. So it's not really that important. You can ignore this comment.

arnaudsjs (approved):

src/inmanta/ast/constraint/expression.py
sanderr: Ok. We can always refine if we get a better idea.

PR #7407, opened 2024-03-20
Issue/7231 resource view

wouterdb:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/tests/server/test_resource_list.py
+++ b/tests/server/test_resource_list.py
@@ -25,6 +25,7 @@ import pytest
 from dateutil.tz import UTC
 from tornado.httpclient import AsyncHTTPClient, HTTPRequest
 
+import inmanta.util
 from inmanta import data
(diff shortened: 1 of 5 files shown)


tests/server/test_resource_list.py
```diff
@@ -525,3 +534,184 @@ async def test_deploy_summary(server, client, env_with_resources):
...
+    async def make_resource_set(tenant_index: int, iteration: int):
+        is_full = tenant_index == 0 and iteration == 0
+        if is_full:
+            version = await clienthelper.get_version()
+        else:
+            version = 0
+
+        resources = [
+            {
+                "id": f"test::XResource[agent{tenant_index},sub={ri}],v={version}",
```
wouterdb, line 572: We may want to vary the type, as it is the default order

PR #7423, opened 2024-03-22
Issue/6587 require env pause before deletion

### LINKED ISSUE #6587: require environment pause before deletion
Currently an environment can be deleted at any time. It seems to me that since the deletion is non-atomic the orchestrator might start misbehaving. I think it would be better to only allow delete while the environment is paused. It should not be allowed to unpause an environment when a delete is in progress. The same for deleting an active model version.

FloLey:
# Description

- [x]  delete first halts the environment
- [x]  environment can not be resumed while deleting
- [x]  reject deletion of active model version

closes #6587

# Self Check:

- [x] Changelog entry


FloLey:
I'm not sure how to test 'environment can not be resumed while deleting'

tests/agent_server/test_facts.py
```diff
@@ -381,6 +381,15 @@ async def params_are_available() -> bool:
     assert result.code == 200
     assert len(result.result["parameters"]) == 4
 
+    # Create version 3 to be able to delete version 2
+    version = await clienthelper.get_version()
+    assert version == 3
+
+    await clienthelper.put_version_simple([], version)
+
+    result = await client.release_version(environment, version, True, const.AgentTriggerMethod.push_full_deploy)
```
arnaudsjs, line 390:
```suggestion
    result = await client.release_version(environment, version, push=True, const.AgentTriggerMethod.push_full_deploy)
```

tests/server/test_env_settings.py
```diff
@@ -341,3 +344,42 @@ async def test_get_setting_no_longer_exist(server, client, environment):
...
+    Verify env will be halted before it is deleted.
+    """
+    result = await client.create_project("env-test")
+    assert result.code == 200
+    project_id = result.result["project"]["id"]
+
+    result = await client.create_environment(project_id=project_id, name="dev")
+    assert result.code == 200
+    env_id = result.result["environment"]["id"]
+
```
arnaudsjs, line 360: Why not use the `environment` fixture here?

tests/server/test_env_settings.py
```diff
@@ -341,3 +344,42 @@ async def test_get_setting_no_longer_exist(server, client, environment):
...
+    """
+    Cannot resume environment that is marked for deletion
+    """
+    result = await client.create_project("env-test")
+    assert result.code == 200
+    project_id = result.result["project"]["id"]
+
+    result = await client.create_environment(project_id=project_id, name="dev")
+    assert result.code == 200
+    env_id = result.result["environment"]["id"]
```
arnaudsjs, line 378: Same here.

tests/test_server.py
```diff
@@ -2027,3 +2027,27 @@ async def test_set_param_v2(server, client, environment):
...
+    res1 = "test::Resource[agent1,key=key1]"
+    res2 = "test::Resource[agent1,key=key2]"
+    resources = [
+        {"key": "key1", "value": "value", "id": f"{res1},v={version}", "requires": [], "purged": False, "send_event": False},
+        {"key": "key2", "value": "value", "id": f"{res2},v={version}", "requires": [], "purged": False, "send_event": False},
+    ]
+
+    await clienthelper.put_version_simple(resources, version)
+
+    result = await client.release_version(environment, version, True, const.AgentTriggerMethod.push_full_deploy)
```
arnaudsjs, line 2047: I also don't see why we need `push=True` here?
```suggestion
    result = await client.release_version(environment, version, push=True, const.AgentTriggerMethod.push_full_deploy)
```

arnaudsjs (changes requested):

Hugo-Inmanta (changes requested):

src/inmanta/server/services/environmentservice.py
```diff
@@ -269,6 +269,9 @@ async def halt(self, env: data.Environment) -> None:
 
     @handle(methods_v2.resume_environment, env="tid")
     async def resume(self, env: data.Environment) -> None:
+        if env.is_marked_for_deletion:
+            raise BadRequest("Cannot resume an environment that is marked for deletion.")
+
```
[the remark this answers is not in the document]
FloLey: As the `agent_state_lock` is already used for the halting/resuming I used the same one for the delete. I'm not sure this  is ok.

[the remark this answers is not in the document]
arnaudsjs: Indeed. There is already a lock for the agent_state. The thing I don't like about re-using the same lock for the mark_for_deletion flag, is that it becomes very implicit that this lock is used for two purposes. I would either introduce a second lock or I would rename the current lock so that it reflects its broader scope and add a comment that explain the scope of the lock.

src/inmanta/server/services/environmentservice.py
```diff
@@ -471,21 +482,31 @@ async def environment_list(self, details: bool = False) -> list[model.Environmen
...
+                is_protected_environment = await env.get(data.PROTECTED_ENVIRONMENT, connection)
+                if is_protected_environment:
+                    raise Forbidden(
+                        f"Environment {environment_id} is protected. See environment setting: {data.PROTECTED_ENVIRONMENT}"
+                    )
+
+                # Check if the environment is halted; if not, halt it
+                if not env.halted:
+                    LOGGER.info("Halting Environment %s", str(environment_id))
+                    await self.halt(env, connection=connection)
```
arnaudsjs, line 500: The `halt()` method doesn't have the `connection` argument yet, as far as I can see.

arnaudsjs (changes requested):

arnaudsjs (changes requested):
I now realize that I was wrong yesterday, saying that we needed a second lock. I misunderstood the scope of the current lock. Sorry for that. I really think we should add a clear docstring regarding the scope of this agent_state_lock.

FloLey:
@arnaudsjs I also renamed the lock as I didn't like is name if I reuse it. I hope that's ok

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -507,14 +507,20 @@ async def get_version(
...
+            if version is None:
+                return 404, {"message": "The given configuration model does not exist yet."}
+
+            active_version = await data.ConfigurationModel.get_latest_version(env.id, connection=connection)
+            if active_version and active_version.version == version.version:
+                raise BadRequest("Cannot delete the active version")
+
+            await version.delete_cascade()
+            # Make sure the ResourcePersistentState is consistent with resources
+            await ResourcePersistentState.trim(env.id)
```
arnaudsjs, line 522:
```suggestion
            await version.delete_cascade(connection=connection)
            # Make sure the ResourcePersistentState is consistent with resources
            await ResourcePersistentState.trim(env.id, connection=connection)
```

arnaudsjs (changes requested):

arnaudsjs (approved):

Hugo-Inmanta:
Do we also want to address this comment from the ticket now: https://github.com/inmanta/inmanta-core/issues/6587#issuecomment-1991933912 ? Or open a follow-up ticket

Hugo-Inmanta (approved):
LGTM, I still added one general comment regarding [this comment](https://github.com/inmanta/inmanta-core/issues/6587#issuecomment-1991933912) on the ticket

FloLey:
@Hugo-Inmanta what I understand from this discussion: https://inmanta.slack.com/archives/CKRF0C8R3/p1710257732265279 is that nothing needs to be done

PR #7429, opened 2024-03-22
Fix incorrect merging of undefined resources by put_partial

### LINKED ISSUE #7416: put_partial causes release_version to fail if undefined resources are removed from the resource set
1. when a resource is removed from a resource set, it is not removed from the 'undeployable' resource field on the model
2. this causes release_version to fail when it calls `resource_action_update`
3. this failure condition is not checked and silently ignored 

Todo
- [x] make `resource_action_update` raise exceptions when it fails
- [x] fix the underlying issue

arnaudsjs:
# Description

Fix bug where `undefined` and `skipped_for_undefined` resources are not correctly merged by the put_partial endpoint.

closes #7416 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5404,7 +5404,7 @@ class ConfigurationModel(BaseDocument):
                         WHERE r.environment=$1
                             AND r.model=$8
                             AND r.resource_id=t.rid
-                            -- Keep resources that belong to the shared resource set or a resource set that was not updated
+                            -- Keep only resources that belong to the shared resource set or a resource set that was not updated
                             AND (r.resource_set IS NULL OR NOT r.resource_set=ANY($9))
                     )
                 )
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5372,8 +5372,9 @@ class ConfigurationModel(BaseDocument):
         undeployable: abc.Sequence[ResourceIdStr],
         skipped_for_undeployable: abc.Sequence[ResourceIdStr],
         partial_base: int,
-        rids_in_partial_compile: abc.Set[ResourceIdStr],
         pip_config: Optional[PipConfig],
+        updated_resource_sets: abc.Set[str],
+        deleted_resource_sets: abc.Set[str],
         connection: Optional[Connection] = None,
     ) -> "ConfigurationModel":
         """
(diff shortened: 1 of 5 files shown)


src/inmanta/data/__init__.py
```diff
@@ -5390,14 +5391,42 @@ async def create_for_partial_compile(
...
-                WHERE c2.environment=$1 AND c2.version=$8
+                SELECT t.rid
+                FROM (
+                    SELECT DISTINCT unnest(c2.undeployable) AS rid
+                    FROM {cls.table_name()} AS c2
+                    WHERE c2.environment=$1 AND c2.version=$8
+                ) AS t(rid)
+                WHERE (
+                    EXISTS (
+                        SELECT 1
```
sanderr, line 5402: This is equivalent to a `JOIN` because it selects on (stricter than) the primary key? I'd find `JOIN` a bit more readable, but fine by me if you prefer it like this.

src/inmanta/data/__init__.py
```diff
@@ -5390,14 +5391,42 @@ async def create_for_partial_compile(
...
+                        FROM {Resource.table_name()} AS r
+                        WHERE r.environment=$1
+                            AND r.model=$8
+                            AND r.resource_id=t.rid
+                            -- Keep resources that belong to the shared resource set or a resource set that was not updated
```
sanderr, line 5407:
```suggestion
                            -- Keep only resources that belong to the shared resource set or a resource set that was not updated
```

src/inmanta/data/__init__.py
```diff
@@ -5421,9 +5450,7 @@ async def create_for_partial_compile(
                     FROM (
                         -- Undeployables in previous version of the model that are not part of the partial compile.
                         (
-                            SELECT rid
-                            FROM rids_undeployable_base_version AS undepl
-                            WHERE NOT undepl.rid=ANY($9)
+                            SELECT rid FROM rids_undeployable_base_version AS undepl
```
sanderr, line 5453: Doesn't this introduce a new bug, where we include shared resources even if they are being exported again? I'm not 100% confident off the top of my head what `ConfigurationModel.undeployable` is used for and whether it's bad to overpopulate it like this.

sanderr (commented):
Do I understand correcty that the bug was that we also kept undeployable and skipped_for_undeployable even if their entire set was being replaced/deleted?

sanderr (changes requested):
Clicked the wrong button before.

arnaudsjs:
> Do I understand correcty that the bug was that we also kept undeployable and skipped_for_undeployable even if their entire set was being replaced/deleted?

yes

sanderr (approved):

FloLey (approved):

PR #7442, opened 2024-03-26
improved f-string error reporting

### LINKED ISSUE #7418: Invalid f-string does not have any location information
consider the following code:

```
f"}"
```

this produces:

```

===================================================================== COMPILATION FAILURE =====================================================================
Error: Single '}' encountered in format string
````

If this happens in a file with 1000 lines there is no way to know where this is.

sanderr:
# Description

Catch Python's string format errors and report them as compiler exceptions (with location information) to the user.

@bartv I assigned you as reviewer for the error messages and the exhaustiveness of test scenarios. For both, see the test cases.

closes #7418

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/7418-f-string-error-reporting.yml
+++ b/changelogs/unreleased/7418-f-string-error-reporting.yml
@@ -1,0 +1,9 @@
+description: "Improved f-string error reporting"
+issue-nr: 7418
+change-type: patch
+sections:
+  bugfix: "{{description}}"
+destination-branches:
+  - master
+  - iso7
+  - iso6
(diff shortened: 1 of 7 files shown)


bartv (approved):

wouterdb (approved):
I don't see any issue, so I might miss the point.

sanderr:
> I don't see any issue, so I might miss the point.

If you would run the examples from the tests on master, you would get
1. Errors straight from Python, not wrapped in a compiler exception, therefore they provide no context at all about where it occurs.
2. Cryptic errors in some cases, e.g. `KeyError("'0'")` without further context.

Hugo-Inmanta (approved):

PR #7454, opened 2024-03-27
Issue/deadlock on boot

wouterdb:
# Description

Fix deadlock on server boot

This is a bit subtle because:
1. lsm validation listener is blocking 
2. lsm validation listener calls request compile

So the process has to be non-blocking both ways (boot can not block new requests, listeners can not block boot)  and order preserving (listener goes before compile, compiles go in order)

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -814,7 +814,7 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
             return
         env: Optional[data.Environment] = await data.Environment.get_by_id(environment)
         if env is None:
-            LOGGER.warning("Can't dequeue compile: environment %s does not exist" % environment)
+            LOGGER.warning("Can't dequeue compile: environment %s does not exist", environment)
             return
         if not env.halted and (compile.environment not in self._recompiles or self._recompiles[compile.environment].done()):
             task = self.add_background_task(self._run(compile))
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -790,14 +790,35 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
             unhandled = await data.Compile.get_unhandled_compiles()
             for u in unhandled:
                 await self._notify_listeners(u)
-            # We are ready to start!
-            self.fully_ready = True
             # one run per env max to get started
             runs = await data.Compile.get_next_run_all()
-            for run in runs:
-                await self._queue(run)
+
+            # prevent races by holding global lock
+            async with self._global_lock:
+                # We are ready to start!
+                self.fully_ready = True
+                for run in runs:
+                    await self.__queue_no_lock(run.environment, run)
 
         self.add_background_task(sub_recovery())
+
+    async def __queue_no_lock(self, environment: uuid.UUID, compile: data.Compile) -> None:
+        """
+        Internal helper that starts the next task if required
+
+
+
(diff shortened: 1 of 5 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -515,9 +515,14 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
     def __init__(self) -> None:
         super().__init__(SLICE_COMPILER)
         self._recompiles: dict[uuid.UUID, Task[None | Result]] = {}
+
         self._global_lock = asyncio.locks.Lock()
+        # boolean indicating when everything is up and processing can start
```
sanderr, line 520:
```suggestion
        # boolean indicating when everything is up and compile execution can start
```
If I understand correctly this determines only if we are ready to start executing compiles, not whether the service is ready to receive requests or anything like that. Right?

sanderr (approved):

arnaudsjs (changes requested):

src/inmanta/server/services/compilerservice.py
wouterdb: correct

src/inmanta/server/services/compilerservice.py
```diff
@@ -774,15 +781,23 @@ async def _sub_notify() -> None:
...
+        async def sub_recovery() -> None:
+            unhandled = await data.Compile.get_unhandled_compiles()
+            for u in unhandled:
+                await self._notify_listeners(u)
+            # We are ready to start!
+            self.fully_ready = True
+            # one run per env max to get started
+            runs = await data.Compile.get_next_run_all()
+            for run in runs:
+                await self._queue(run)
```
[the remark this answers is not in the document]
wouterdb: thx, missed that one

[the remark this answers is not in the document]
sanderr: Could you elaborate? `_queue` does not actually schedule the compile if another is already in progress.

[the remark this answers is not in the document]
wouterdb: replaced

src/inmanta/server/services/compilerservice.py
```diff
@@ -774,15 +781,44 @@ async def _sub_notify() -> None:
...
+
+
+
+            Must be called under the _global_lock
+        """
+        if self.is_stopping():
+            return
+        env: Optional[data.Environment] = await data.Environment.get_by_id(environment)
+        if env is None:
+            LOGGER.warning("Can't dequeue compile: environment %s does not exist" % environment)
```
arnaudsjs, line 817:
```suggestion
            LOGGER.warning("Can't dequeue compile: environment %s does not exist", environment)
```

src/inmanta/server/services/compilerservice.py
```diff
@@ -774,15 +781,44 @@ async def _sub_notify() -> None:
...
+            # prevent races by holding global lock
+            async with self._global_lock:
+                # We are ready to start!
+                self.fully_ready = True
+                for run in runs:
+                    await self.__queue_no_lock(run.environment, run)
+
+        self.add_background_task(sub_recovery())
+
+    async def __queue_no_lock(self, environment: uuid.UUID, compile: data.Compile) -> None:
```
arnaudsjs, line 805: Is there a reason why `_queue()` doesn't call into this method? Now this code is duplicates across two methods.

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #7466, opened 2024-03-28
Re-Added inline compile listeners

wouterdb:
# Description

Re-revert #7462 

Contains additional fixes not present in first 2 PR's 

For after the release


sanderr (approved):

src/inmanta/server/services/compilerservice.py
```diff
@@ -763,21 +781,57 @@ async def notify(listener: CompileStateListener) -> None:
...
+        # we run the recovery and start in the background to prevent deadlocks with sync handlers
+        async def sub_recovery() -> None:
+            unhandled = await data.Compile.get_unhandled_compiles()
+            for u in unhandled:
+                await self._notify_listeners(u)
+            # one run per env max to get started
+            runs = await data.Compile.get_next_run_all()
+
+            # prevent races by holding global lock
+            async with self._global_lock:
```
sanderr, line 810: Even with this lock I think there might be a race with `self._queue`: that method takes a compile as an argument. After acquiring the lock it does not check whether that compile may have already run.
1. server is started, `sub_recovery` runs as background task
2. `request_compile` is called for compile `c1`
3. `request_recompile` inserts compile in db and calls `self._queue(c1)`
4. `sub_recovery` wins race for global lock: discovers `c1` and starts running it
5. `c1` is a very fast compile and finishes almost immediately
6. `sub_recovery` has a bunch of other compiles to start because there are multiple environments. Only after that does it release the lock
7. event loop passes control back to `self._queue`, which acquires the lock. It sees that no compile is running for this environment, so it runs its argument `c1`, even though it has already run.

sanderr: This is probably extremely rare, but since it would be easy to address I think we should.

sanderr: Same in the other direction I think.

sanderr: From Slack (for future reference):

Ah, but perhaps the opposite is also possible?
1. recover fetches waiting compiles (`Compile.get_next_run_all()`) -> there are none because we're starting clean
2. compile is requested: `request_recompile` calls `self._queue(c1)` and wins the race -> returns without doing anything because `not self.fully_ready`
3. recovery acquires the lock, sets `fully_ready = True`, starts 0 compiles.
4. nothing happens. Compiles remain in the queue until a new one is requested, and only then do we start consuming again.

PR #7476, opened 2024-03-29
Issue 7247 raise warning jit

### LINKED ISSUE #7247: Raise warning/error when running against a DB that has jit enabled
With the PostgreSQL jit enabled, the query performance is quite bad. This issue should make sure we raise a warning when the Inmanta server starts against a DB with the jit enabled.

Also add to the PG documentation that we recommend to disable it.

FloLey:
# Description

Raise a warning if JIT is enabled on the psql db as this will degrade performances.

closes #7247 

# Self Check:

- [x] Changelog entry


FloLey:
I tested it manually as writing a test for this seems hard.

changelogs/unreleased/7247-raise-warning-jit.yml
```diff
@@ -0,0 +1,5 @@
+description: Raise a warning if JIT is enabled on the psql db as this will degrade performances.
```
arnaudsjs, line 1: We know it had some issues on older PostgreSQL versions, but it might be better on newer versions.
```suggestion
description: Raise a warning if JIT is enabled on the psql db as this might degrade performance.
```

docs/install/1-install-server.rst
```diff
@@ -357,8 +357,17 @@ to
   # IPv6 local connections:
   host    all             all             ::1/128                 md5
 
+Make sur JIT is disabled for the psql database as it will impact performances.
```
arnaudsjs, line 360:
```suggestion
Make sure JIT is disabled for the psql database as it might result in poor query performance.
```

docs/install/1-install-server.rst
```diff
@@ -357,8 +357,17 @@ to
...
+To disable JIT, set
 
-Restart the PostgreSQL server to apply the changes made in the ``pg_hba.conf`` file:
+.. code-block:: text
+
+  # disable JIT
+  jit = off
+
+in ``/etc/postgresql/13/main/postgresql.conf``.
```
arnaudsjs, line 368: I guess you are mentioning the path to the `postgresql.conf` file, like it is on Ubuntu. On RHEL, this file is present in a different location:
```suggestion
in ``/var/lib/pgsql/13/data/postgresql.conf``.
```

docs/install/1-install-server.rst
```diff
@@ -357,8 +357,17 @@ to
...
+
+Restart the PostgreSQL server to apply the changes made in the ``pg_hba.conf`` and  ``postgresql.conf`` files:
```
arnaudsjs, line 370: `pg_hba.conf`?

src/inmanta/server/services/databaseservice.py
```diff
@@ -90,6 +90,12 @@ async def connect_database(self) -> None:
         )
         LOGGER.info("Connected to PostgreSQL database %s on %s:%d", opt.db_name.get(), database_host, database_port)
 
+        # Check if JIT is enabled
+        async with self._pool.acquire() as connection:
+            jit_enabled = await connection.fetchval("SHOW jit;")
+            if jit_enabled == "on":
+                LOGGER.warning("JIT is enabled in the PostgreSQL database. This will affect performances.")
```
arnaudsjs, line 97:
```suggestion
                LOGGER.warning("JIT is enabled in the PostgreSQL database. This might result in poor query performance.")
```

arnaudsjs (changes requested):

docs/install/1-install-server.rst
FloLey: I don't know what this file is it was already in the docs I didn't add it

arnaudsjs: Ah. I see. This is a section with all the PostgreSQL-related config and not only the GIT config.

src/inmanta/server/services/databaseservice.py
```diff
@@ -90,6 +90,12 @@ async def connect_database(self) -> None:
         )
         LOGGER.info("Connected to PostgreSQL database %s on %s:%d", opt.db_name.get(), database_host, database_port)
 
+        # Check if JIT is enabled
+        async with self._pool.acquire() as connection:
+            jit_enabled = await connection.fetchval("SHOW jit;")
```
arnaudsjs, line 95: You will need to rely on `SELECT pg_jit_available();` instead of `SHOW jit;`. For the jit to work, two things are required: The jit compiler needs to be installed and the jit config option needs to be enabled. The `SHOW jit;` command only shows the value of the config option, while the `SELECT pg_jit_available();` returns only true if both requirements are met. Using the latter command will ensure we only raise the warning when the jit is actually enabled. This is especially relevant because the default value of the jit config option it `on`.

arnaudsjs (changes requested):

docs/install/1-install-server.rst
arnaudsjs: You forgot about this comment I think

FloLey: indeed sorry. It is changed

docs/install/1-install-server.rst
```diff
@@ -357,8 +357,17 @@ to
   # IPv6 local connections:
   host    all             all             ::1/128                 md5
 
+Make sure JIT is disabled for the psql database as it might result in poor query performance.
```
arnaudsjs, line 360: The comment of Hugo got reverted by applying my suggestion.
```suggestion
Make sure JIT is disabled for the PostgreSQL database as it might result in poor query performance.
```

arnaudsjs (changes requested):

PR #7482, opened 2024-04-02
Issue/7253 drop use generic

### LINKED ISSUE #7253: Drop `use_generic` from `Resource.deserialize` method
This is blocked on when the agent refactor integration is done.
Spawned from this discussion https://github.com/inmanta/inmanta-core/pull/7101/files#r1491241944

- double check that this isn't used anymore + drop the  `use_generic` parameter from `Resource.deserialize` [method](https://github.com/inmanta/inmanta-core/blob/a3281a3abb7c760e7573000b72e3f59c3d8beb42/src/inmanta/resources.py#L338
)
- iso8 only!

FloLey:
# Description

Drop `use_generic` from `Resource.deserialize` method

closes #7253 

# Self Check:

- [x] Changelog entry


Hugo-Inmanta:
The ticket mentions this should wait until the agent refactor integration is done, I don't think it's done yet

changelogs/unreleased/7253-remove-use-generic.yml
```diff
@@ -0,0 +1,5 @@
+description: Drop `use_generic` from `Resource.deserialize` method
+issue-nr: 7243
```
Hugo-Inmanta, line 2:
```suggestion
issue-nr: 7253
```

changelogs/unreleased/7253-remove-use-generic.yml
```diff
@@ -0,0 +1,5 @@
...
+change-type: patch
```
Hugo-Inmanta, line 3: I think merge tool will complain if we make this a patch but we don't merge it in branches that allow patches ie iso 6 and iso7. I think this has to be a major ?

Hugo-Inmanta (changes requested):

changelogs/unreleased/7253-remove-use-generic.yml
FloLey: I don't think so but I might be mistaken. I think it only complains the other way around. @sanderr?

sanderr: It does complain under certain circumstances (not these). If the branches are still aligned, a patch is not allowed to diverge them. But the branches are already diverged so then anything goes.

Hugo-Inmanta (approved):

PR #7485, opened 2024-04-02
Fix confusing code regaring handling of environment variables in compilerservice

arnaudsjs:
# Description

According to the docstring of `data.Compile`, the `used_environment_variables` field should be set to None until the compile starts. This was not in-line with the implementation. This in its turn made the interpretation of the `CompilerService._compile_merge_key()` method rather difficult. This PR fixes this issue by making the implementation in-line with the docstring.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -706,7 +706,12 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
         """
         return c.to_dto().model_dump_json(
             include={
-                "environment", "started", "do_export", "requested_environment_variables", "partial", "removed_resource_sets"
+                "environment",
+                "started",
+                "do_export",
+                "requested_environment_variables",
+                "partial",
+                "removed_resource_sets",
             },
         )
 
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -658,7 +658,7 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
             force_update=force_update,
             metadata=metadata,
             requested_environment_variables=env_vars,
-            used_environment_variables=env_vars,
+            used_environment_variables=None,
             mergeable_environment_variables=mergeable_env_vars,
             partial=partial,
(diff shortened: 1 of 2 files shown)


sanderr (commented):
Looks like there's a type violation at line 431: it assumes that the value will never be `None`. Could you address that as well?

I'm still looking over the context for the rest of the review.

sanderr (approved):
While reviewing I came across [one concern](https://inmanta.slack.com/archives/CKRF0C8R3/p1712154130560919) but it doesn't directly impact the functionality of this PR or what we currently use this feature for.

arnaudsjs:
> While reviewing I came across [one concern](https://inmanta.slack.com/archives/CKRF0C8R3/p1712154130560919) but it doesn't directly impact the functionality of this PR or what we currently use this feature for.

I converted the assertion into an exception and added a test case for this scenario.

src/inmanta/server/services/compilerservice.py
```diff
@@ -650,7 +650,11 @@ async def request_recompile(
         requested = datetime.datetime.now().astimezone()
 
         shared_keys = mergeable_env_vars.keys() & env_vars.keys()
-        assert not shared_keys, f"An env var can not be both mergeable and normal: {shared_keys}"
+        if shared_keys:
+            raise Exception(
```
sanderr, line 654:
```suggestion
            raise ValueError(
```

I think this would be slightly more to the point, but I'm fine with either.

tests/server/test_compilerservice.py
```diff
@@ -1847,3 +1847,26 @@ async def wait_for_compile() -> bool:
...
+            force_update=False,
+            do_export=False,
+            remote_id=uuid.uuid4(),
+            env_vars={"var": "val", "test": "123"},
+            mergeable_env_vars={"var": "otherval", "somekey": "someval"},
+        )
+    assert (
+        "Invalid compile request: The same environment variable cannot be present in the "
+        "env_vars and mergeable_env_vars dictionary simultaneously: {'var'}." in str(exc.value)
+    )
```
sanderr, line 1872: FYI, `pytest.raises` has a `matches` keyword to match on the error message.

Issue #7428, opened 2024-03-22
compatibility page improvements


sanderr:
As discussed in demo meeting:
- [ ] move page to Inmanta Reference
- [ ] add context to python requirements section
- [ ] rename / restructure json component_requirements? Discuss on Slack
- [ ] anything else?

PR #7490, opened 2024-04-04
Improve compatibility page

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/7428

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/docs/_templates/components_requirements.tmpl
+++ b/docs/_templates/components_requirements.tmpl
@@ -8,7 +8,7 @@ The python packages are of two types: one core package and several extension pac
 
 {% if data['component_requirements']%}
 
-The following table shows the range of compatible versions for Inmanta python components.
+The following table shows the ranges of the Inmanta Python components, this product version is compatible with.
 These compatible versions are defined for the whole lifetime of this major version i.e.
 they are not pinned to a specific three-digit version. Upper bounds may be added when a new major version is released.
 
--- a/docs/_templates/components_requirements.tmpl
+++ b/docs/_templates/components_requirements.tmpl
@@ -13,7 +13,7 @@ These compatible versions are defined for the whole lifetime of this major versi
 they are not pinned to a specific three-digit version. Upper bounds may be added when a new major version is released.
 
 This table is only meant as a reference for developers of Inmanta extensions. This information is not relevant
-when installing the orchestrator via rpm as is recommended in the :ref:`install-server` section.
+when installing the orchestrator via RPM as is recommended in the :ref:`install-server` section.
 
 .. list-table:: Inmanta core and extensions versions
    :header-rows: 1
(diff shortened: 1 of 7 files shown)


Hugo-Inmanta:
On hold until 08/04 for discussion about  `rename / restructure json component_requirements? Discuss on Slack`

docs/_templates/components_requirements.tmpl
```diff
@@ -3,17 +3,27 @@
...
+The python packages are of two types: one core package and several extension packages.
 
 {% if data['component_requirements']%}
+
+The following table shows the range of compatible versions for Inmanta python components.
+These compatible versions are defined for the whole lifetime of this major version i.e.
+they are not pinned to a specific three-digit version. Upper bounds may be added when a new major version is released.
+
+This table is only meant as a reference for developers of Inmanta extensions. This information is not relevant
+when installing the orchestrator via rpm as is recommended in the :ref:`install-server` section.
```
arnaudsjs, line 16:
```suggestion
when installing the orchestrator via RPM as is recommended in the :ref:`install-server` section.
```

docs/_templates/components_requirements.tmpl
```diff
@@ -3,17 +3,27 @@
 Inmanta core and extensions versions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+The orchestrator is constituted of several components (python packages and npm packages).
+The python packages are of two types: one core package and several extension packages.
 
 {% if data['component_requirements']%}
+
+The following table shows the range of compatible versions for Inmanta python components.
```
arnaudsjs, line 11: I think we should make it clear that the production version is the starting point and from that we define version constraints on the versions of the components.
```suggestion
The following table shows the ranges of the Inmanta Python components, this product version is compatible with.
```

arnaudsjs (changes requested):

docs/_templates/components_requirements.tmpl
```diff
@@ -3,17 +3,27 @@
...
+These compatible versions are defined for the whole lifetime of this major version i.e.
```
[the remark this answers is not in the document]
arnaudsjs: Based on Sander his comment on Slack I think we need to change the following here: 

* I would change the sentence "These compatible versions are defined for the whole lifetime of this major version i.e. they are not pinned to a specific three-digit version." to "the lower bound of these version constraints will remain the same for the whole lifetime of this major version". The latter is just a lot more clear to me.
* We also need to mention explicitly that these version constraints will only result in compatible versions of they are used as input towards a dependency resolver like pip. The current explanation gives the impression that if your installed packages satisfy these constraints, they are compatible with each other. But that's not necessarily true.

arnaudsjs (approved):

PR #7526, opened 2024-04-17
Fix bug on environment clear/delete endpoints

arnaudsjs:
# Description

* Fix bug that makes the endpoints to clear or delete an environment fail with the error message `(39, 'Directory not empty')`. This bug triggers when the compiler service updates the files in the environment directory, while the API endpoint to delete/clear the environment performs a `shutil.rmtree()` on that directory.
* Add the `SLICE_AGENT_MANAGER` to the dependencies of the EnvironmentService. It think this is a bug.
* Make the endpoint to delete an environment first calls into `mark_for_deletion()` and only then halt the environment, because the former is a guard to make sure that an environment, that is marked for deletion, cannot be resumed. The original implementation did it the other way around. This was not an issue, because these calls are protected by the `environment_state_operation_lock`, but having them swapped make the code easier to read.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -1068,3 +1068,18 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
     async def recalculate_queue_count_cache(self) -> None:
         async with self._queue_count_cache_lock:
             self._queue_count_cache = await data.Compile.get_total_length_of_all_compile_queues()
+
+    async def cancel_compile(self, env_id: uuid.UUID) -> None:
+        """
+        Cancel the ongoing compile for the given environment (if any) and don't start any new compiles.
+
+        :param env_id: The id of the environment for which the compile should be cancelled.
+        """
+        compile_task: Optional[Task[None | Result]] = self._recompiles.pop(env_id, None)
+        if compile_task:
+            compile_task.cancel()
+            try:
+                # Wait until cancellation has finished
+                await compile_task
+            except asyncio.CancelledError:
+                pass
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -1094,12 +1094,13 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
 
         :param env_id: The id of the environment for which the compile should be cancelled.
         """
-        compile_task: Optional[asyncio.Task[None | protocol.Result]] = self._env_to_compile_task.pop(env_id, None)
-        if compile_task:
-            compile_task.cancel()
-            try:
-                # Wait until cancellation has finished
-                await compile_task
-            except asyncio.CancelledError:
-                # Don't propagate CancelledError raised by invoking compile_task.cancel()
-                pass
+        async with self._write_lock_env_to_compile_task:
+            compile_task: Optional[asyncio.Task[None | protocol.Result]] = self._env_to_compile_task.pop(env_id, None)
+            if compile_task:
+                compile_task.cancel()
+                try:
+                    # Wait until cancellation has finished
+                    await compile_task
+                except asyncio.CancelledError:
+                    # Don't propagate CancelledError raised by invoking compile_task.cancel()
+                    pass
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -1071,7 +1071,9 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
 
     async def cancel_compile(self, env_id: uuid.UUID) -> None:
         """
-        Cancel the ongoing compile for the given environment (if any) and don't start any new compiles.
+        Cancel the ongoing compile for the given environment (if any). After the cancellation, this method
+        will not start the execution of a new compile, if a compile request would be present in the compile
+        queue for the given environment.
 
         :param env_id: The id of the environment for which the compile should be cancelled.
         """
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -520,9 +520,9 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
     def __init__(self) -> None:
         super().__init__(SLICE_COMPILER)
 
-        self._compiling_envs: set[uuid.UUID] = set()
+        self._env_to_compile_task: dict[uuid.UUID, asyncio.Task[None | protocol.Result]] = {}
         # write lock only, writers take care to uphold consistency so that readers don't need the lock
-        self._write_lock_compiling_envs = asyncio.locks.Lock()
+        self._write_lock_env_to_compile_task = asyncio.locks.Lock()
 
         # boolean indicating when everything is up and compile execution can start
         self.fully_ready = False
@@ -773,7 +773,7 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
 
         # acquire connection before lock to ensure lock contention remains short lived
         async with data.Environment.get_connection(connection) as con:
-            async with self._write_lock_compiling_envs:
+            async with self._write_lock_env_to_compile_task:
                 started_new_compile_run: bool = False
 
                 try:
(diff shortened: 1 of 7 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -1068,3 +1068,18 @@ async def environment_action_deleted(self, env: model.Environment) -> None:
     async def recalculate_queue_count_cache(self) -> None:
         async with self._queue_count_cache_lock:
             self._queue_count_cache = await data.Compile.get_total_length_of_all_compile_queues()
+
+    async def cancel_compile(self, env_id: uuid.UUID) -> None:
+        """
+        Cancel the ongoing compile for the given environment (if any) and don't start any new compiles.
```
Hugo-Inmanta, line 1074: > and don't start any new compiles

I don't think this is enforced by this method.

src/inmanta/server/services/environmentservice.py
```diff
@@ -524,11 +530,18 @@ async def environment_clear(self, env: data.Environment) -> None:
         if is_protected_environment:
             raise Forbidden(f"Environment {env.id} is protected. See environment setting: {data.PROTECTED_ENVIRONMENT}")
 
-        await self.autostarted_agent_manager.stop_agents(env, delete_venv=True)
+        await asyncio.gather(
+            self.autostarted_agent_manager.stop_agents(env, delete_venv=True),
+            self.compiler_service.cancel_compile(env.id),
```
Hugo-Inmanta, line 535: Isn't there a race condition here in the case where there are multiple compiles queued up ? We cancel the currently running one but couldn't the next one start right after ?
I must admit I'm not too clear on the semantics between clear and delete and
- why we need the env to be halted when deleting but not when clearing ?
- why we don't need the `environment_state_operation_lock` when clearing ?

Hugo-Inmanta (changes requested):

src/inmanta/server/services/compilerservice.py
arnaudsjs: I mean that this method doesn't start any new compiles. What do you propose?

src/inmanta/server/services/environmentservice.py
arnaudsjs: > Isn't there a race condition here in the case where there are multiple compiles queued up ?

No. Because the compile that finishes starts the next one. But there is one if a new compile is requested right after the previous one is cancelled. That's something I need to fix.

> why we need the env to be halted when deleting but not when clearing ?

I agree it would be better to halt the environment right before we clear is, but that would be a breaking change on the API. Because currently a clear of an environment doesn't halt it. But maybe I can work around that.

> why we don't need the environment_state_operation_lock when clearing ?

It protects against race conditions that would result in an inconsistent state on the agent state (paused or not) and the environment state (halted or not). The clear endpoint doesn't change the environment state, so there is no need for the lock.

Hugo-Inmanta: > Because the compile that finishes starts the next one.

But isn't this an issue ? i.e. deleting the env dir and clearing the env while there is an ongoing compile ?

arnaudsjs: I mean that the normal working of the compiler service is to start a new compile (if any is present in the queue) when a compile finishes. The `cancel_compile()` explicitly deviates from that to make sure that no compile is executing. But it's not waterproof, as mentioned in the your first comment. I will fix this.

Hugo-Inmanta: I completely misunderstood what self._recompiles was. For some reason I thought it was a queue of pending compiles, sorry for the noise. I see now how popping it in `cancel_compile()` prevents further compiles from being queued up.

arnaudsjs:
The new implementation first halts the environment, then performs the clear and then un-halts (if the environment was not initially halted). While the environment is halted, the compilerservice will not run any compiles for the halted environment.

src/inmanta/server/services/compilerservice.py
```diff
@@ -1068,3 +1068,20 @@ async def environment_action_deleted(self, env: model.Environment) -> None:
     async def recalculate_queue_count_cache(self) -> None:
         async with self._queue_count_cache_lock:
             self._queue_count_cache = await data.Compile.get_total_length_of_all_compile_queues()
+
+    async def cancel_compile(self, env_id: uuid.UUID) -> None:
```
sanderr, line 1072: Given that a compile may have side effects like allocation, export, ... I'm not sure if this is a good idea. As in, I honestly don't know. @wouterdb wdyt?

Apart from that, I'm afraid this will not work with the changes I made to the compiler service. For one thing, I removed the `Task` objects from `self._recompiles` (and renamed it and made it a set). But I also made it so that only a single method writes to it. I think we can extend that if we acquire the lock here, but I want to carefully consider it.

Let me know if you'd like me to make the required changes. After all, we only have this conflict because I rushed the other PR before reviewing this one.

src/inmanta/server/services/compilerservice.py
```diff
@@ -1068,3 +1068,20 @@ async def environment_action_deleted(self, env: model.Environment) -> None:
...
+        """
+        Cancel the ongoing compile for the given environment (if any). After the cancellation, this method
+        will not start the execution of a new compile, if a compile request would be present in the compile
+        queue for the given environment.
+
+        :param env_id: The id of the environment for which the compile should be cancelled.
+        """
+        compile_task: Optional[Task[None | Result]] = self._recompiles.pop(env_id, None)
+        if compile_task:
+            compile_task.cancel()
```
sanderr, line 1082: We'll have to carefully consider the potential race between the task finishing and this call. e.g. if the task finishes, we have to make sure it does not start the next compile while we only cancel this one.

I'm not going deeper until the other thread is resolved.

sanderr (commented):
I'm still reviewing but one of my comments may spawn a discussion so I'll unblock that already.

src/inmanta/server/services/compilerservice.py
wouterdb: 1. if we do this on clear only, I think we can safely do it
2. I don't know if this cancel will actually cancel the compile? (i.e. the process will keep going?)

src/inmanta/server/services/environmentservice.py
```diff
@@ -498,37 +509,57 @@ async def environment_delete(self, environment_id: uuid.UUID) -> None:
...
-                await asyncio.gather(
-                    self.autostarted_agent_manager.stop_agents(env, delete_venv=True),
-                    env.delete_cascade(connection=connection),
-                )
+                await self.compiler_service.cancel_compile(env.id)
+                # Delete the environment directory before deleting the database records. This ensures that
+                # this operation can be retried if the deletion of the environment directory fails. Otherwise,
+                # the environment directory would be left in an inconsistent state. This can cause problems if
+                # the user later on recreates an environment with the same environment id.
+                self._delete_environment_dir(environment_id)
```
wouterdb, line 525: we may want to spin this of onto a threadpool, as it blocks the ioloop

sanderr (changes requested):
No further comments

src/inmanta/server/services/compilerservice.py
arnaudsjs: 1. Indeed. I only call this method on clear and delete operations, because it's safe in these situations.
2. The compiler service spawns processes using `asyncio.create_subprocess_exec()`. In each situation the this code is surrounded by a try-finally block that kills the process if it's still running. So this method process guarantees that the process will be stopped by the times this method finishes execution.

src/inmanta/server/services/compilerservice.py
arnaudsjs: We always first halt the environment and then call into `cancel_compile()`. No new compiles are started on halted environment so this race condition shouldn't exist.

src/inmanta/server/services/compilerservice.py
```diff
@@ -800,18 +800,17 @@ async def _process_next_compile_in_queue(
...
-                            self._compiling_envs.add(compile.environment)
                             # Run the compile. self._run will call back into this method when done
-                            self.add_background_task(self._run(compile))
+                            self._env_to_compile_task[compile.environment] = self.add_background_task(self._run(compile))
                             started_new_compile_run = True
                 finally:
                     if finish_current_compile and not started_new_compile_run:
                         assert environment is not None
-                        self._compiling_envs.discard(environment)
+                        del self._env_to_compile_task[environment]
```
sanderr, line 813: I'm not sure `del` is necessarily safe here. `discard` is "remove if exists".

sanderr: One scenario where I think it may not be in the dict is when we clear the env near the end of the compile. This function will be called, and `cancel_compile` will be called. Both will race for the lock. I don't think either can make assumptions about pre-existing state just because they've been called.

src/inmanta/server/services/compilerservice.py
```diff
@@ -1095,11 +1094,13 @@ async def cancel_compile(self, env_id: uuid.UUID) -> None:
...
+        async with self._write_lock_env_to_compile_task:
+            compile_task: Optional[asyncio.Task[None | protocol.Result]] = self._env_to_compile_task.pop(env_id, None)
+            if compile_task:
+                compile_task.cancel()
+                try:
+                    # Wait until cancellation has finished
+                    await compile_task
+                except asyncio.CancelledError:
+                    # Don't propagate CancelledError raised by invoking compile_task.cancel()
+                    pass
```
sanderr, line 1106: I think we want to `pop` the task within the lock, but waiting for it to complete should be done outside of it to keep the lock shortlived. After all, we don't want to block other environments.

This is only safe if the environment is halted when we're here, but I believe this is the case?

sanderr (changes requested):

PR #7527, opened 2024-04-17
Improve support four-digit version on 'inmanta module release' command

### LINKED ISSUE #7521: Improve support four-digit version on 'inmanta module release' command
The `inmanta module release` command performs a patch-level version bump right after applying the git tag. But some modules use four-digit version numbers in general. This ticket should make sure that the `inmanta module release` command has support to bump to a four digit version number after a release.

Proposal: Make this configurable in the module metadata file (`setup.cfg` -> `inmanta.module.ModuleV2Metadata`)

FloLey:
# Description

The inmanta module release command performs a patch-level version bump right after applying the git tag. But some modules use four-digit version numbers in general. This ticket should make sure that the inmanta module release command has support to bump to a four digit version number after a release.
Make this configurable in the module metadata file (setup.cfg -> inmanta.module.ModuleV2Metadata)

closes #7521 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/7521-4-digits-module-release.yml
+++ b/changelogs/unreleased/7521-4-digits-module-release.yml
@@ -1,0 +1,7 @@
+---
+description: Improve support for four-digit version on 'inmanta module release' command. If the 'use_four_digit' is set to 'True' in the modules metadata, the version will be bumped to a 4 digit format after a release.
+issue-nr: 7521
+change-type: minor
+destination-branches: [master, iso7]
+sections:
+  minor-improvement: "{{description}}"
--- a/src/inmanta/moduletool.py
+++ b/src/inmanta/moduletool.py
@@ -1272,7 +1272,10 @@ version: 0.0.1dev0"""
         requested_version_bump: Optional[ChangeType] = ChangeType.parse_from_bools(revision, patch, minor, major)
         if not requested_version_bump and dev:
             # Dev always bumps
-            requested_version_bump = ChangeType.PATCH
+            if hasattr(module.metadata, "four_digit_version") and module.metadata.four_digit_version:
+                requested_version_bump = ChangeType.REVISION
+            else:
+                requested_version_bump = ChangeType.PATCH
 
         if requested_version_bump:
             new_version: Version = self._get_dev_version_with_minimal_distance_to_previous_stable_release(
@@ -1323,7 +1326,7 @@ version: 0.0.1dev0"""
             gitprovider.tag(repo=module_dir, tag=str(release_tag))
             print(f"Tag created successfully: {release_tag}")
             # bump to the next dev version
-            if hasattr(module.metadata, "use_four_digits") and module.metadata.use_four_digits:
+            if hasattr(module.metadata, "four_digit_version") and module.metadata.four_digit_version:
                 self.release(dev=True, message="Bump version to next development version", revision=True)
             else:
                 self.release(dev=True, message="Bump version to next development version", patch=True)
--- a/src/inmanta/moduletool.py
+++ b/src/inmanta/moduletool.py
@@ -70,6 +70,7 @@ from inmanta.module import (
     ModuleNotFoundException,
     ModuleV1,
     ModuleV2,
+    ModuleV2Metadata,
     ModuleV2Source,
     Project,
     gitprovider,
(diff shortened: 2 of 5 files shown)


tests/moduletool/test_release.py
```diff
@@ -140,7 +144,7 @@ def get_changelog_content(after_release: bool) -> str:
     assert gitprovider.get_version_tags(repo=path_module) == expected_tags
     # Verify version
     mod = Module.from_path(path_module)
-    assert mod.version == Version("1.2.4.dev0")
+    assert mod.version == Version("1.2.3.1dev0") if use_four_digits and not v1_module else Version("1.2.4.dev0")
```
Hugo-Inmanta, line 147: I would expect `1.2.3.1.dev0` here

tests/utils.py
```diff
@@ -557,6 +559,8 @@ def to_python_requires(
     config_file: str = os.path.join(dest_dir, module.ModuleV2.MODULE_FILE)
     config: configparser.ConfigParser = configparser.ConfigParser()
     config.read(config_file)
+    if use_four_digits:
+        config["metadata"]["use_four_digits"] = "True"
```
Hugo-Inmanta, line 563: Why not ?
```suggestion
        config["metadata"]["use_four_digits"] = True
```

changelogs/unreleased/7521-4-digits-module-release.yml
```diff
@@ -0,0 +1,7 @@
+---
+description: Improve support for four-digit version on 'inmanta module release' command. If the 'use_four_digit' is set to 'True' in the modules metadata, the version will be bumped to a 4 digit format after a release.
```
Hugo-Inmanta, line 2:
```suggestion
description: Improve support for four-digit version on 'inmanta module release' command. If the 'use_four_digit' is set to 'True' in the module's metadata, the version will be bumped to a 4 digit format after a release.
```

Hugo-Inmanta (changes requested):

tests/utils.py
FloLey: I think I tried it and this will be written to the .cfg and it only accepts strings

tests/moduletool/test_release.py
FloLey: indeed, I didn't pay attention to this. I think this means there is a bug in existing code. @sanderr could you confirm?

FloLey: there is no bug in the existing code. I find it weird but `Version("1.2.3.1dev0") == Version("1.2.3.1.dev0")`

sanderr: It might be, but I'm not entirely sure we want to stick with it. Please leave this thread unresolved for now as a reminder that I look into it.

Hugo-Inmanta (approved):
LGTM :+1:

tests/moduletool/test_release.py
sanderr: This looks good to me as it is now, but considering this new information I do think we should have at least one test that asserts the string as it is present in the metadata. Do you know if we already have such a test?

sanderr (changes requested):
Shouldn't this option also affect the default dev bump? i.e. line 1283?

tests/moduletool/test_release.py
```diff
@@ -57,17 +58,19 @@ def test_release_stable_version(
...
     def get_changelog_content(after_release: bool) -> str:
         if after_release:
             return f"""
 # Changelog
 
-## v1.2.4 - ?
+## {"v1.2.3.1" if use_four_digits and not v1_module else "v1.2.4"} - ?
 
 
 ## v1.2.3 - {datetime.date.today().isoformat()}
```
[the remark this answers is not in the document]
FloLey: As long as the version in the .cfg file and the one in the changelog are aligned it works.

src/inmanta/moduletool.py
```diff
@@ -1272,7 +1272,10 @@ def release(
         requested_version_bump: Optional[ChangeType] = ChangeType.parse_from_bools(revision, patch, minor, major)
         if not requested_version_bump and dev:
             # Dev always bumps
-            requested_version_bump = ChangeType.PATCH
+            if hasattr(module.metadata, "four_digit_version") and module.metadata.four_digit_version:
```
sanderr, line 1275: Why do we need `hasattr` here? Same further down.

sanderr (approved):

src/inmanta/moduletool.py
FloLey: without it I had issues with the v1 modules I think. let me test again

sanderr: Oh right, this is for either. I have a strong preference for an `isinstance` though.

sanderr: Or alternatively we could add the field to v1 metadata as an internal field, as `Literal[False]`. I'm not sure if that would be better or worse.

FloLey: i'll change it to isinstance

PR #7533, opened 2024-04-18
Issue/reference configuration reference

edvgui:
# Description

The first page you land on when searching for the configuration documentation is https://docs.inmanta.com/inmanta-service-orchestrator/7/administrators/configuration.html#configuration

And it doesn't refer to the actual config options documentation.  This PR adds the missing link.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/docs/administrators/configuration.rst
+++ b/docs/administrators/configuration.rst
@@ -1,6 +1,9 @@
 Configuration
 ===================
 
+.. note::
+    The documentation of the configuration options themselves can be found in the :ref:`inmanta config reference<config_reference>`.
+
 Inmanta server and Inmanta agent
 --------------------------------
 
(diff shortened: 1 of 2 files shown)


docs/administrators/configuration.rst
```diff
@@ -1,6 +1,9 @@
 Configuration
 ===================
 
+.. note::
+    The documentation of the configuration options themselves can be found in the :ref:`inmanta config reference<config_reference>`.
```
arnaudsjs, line 5: The name in the menu is "configuration reference"
```suggestion
    The documentation of the configuration options themselves can be found in the :ref:`Inmanta configuration reference<config_reference>`.
```

arnaudsjs (changes requested):

PR #7536, opened 2024-04-19
Clean up and patch compiler service queue processing

sanderr:
# Description

- Unify compiler service queue methods `_queue`, `__queue_no_lock` and `process_next_compile_in_queue` (very similar with subtle differences) to a single method. This change is intended to make the entire flow easier to understand.
- Fix potential (may or may not be related to actual observed problematic behavior, tbc) race condition between compiler service recovery and `request_recompile`. See https://github.com/inmanta/inmanta-core/pull/7466/files#r1570373482.

# Self Check:

- [x] Changelog entry
- [x] ~If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)~ (It does fix a race condition, but one not present on the stable branches.)

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -501,14 +501,16 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
     General design of compile scheduling:
 
     1. all compile request are stored in the database.
-    2. when a task starts, _queue is called, if no job is executing for that environment, it is started
-    3. when a task ends, it is marked as completed in the database and _dequeue is called and _notify_listeners is scheduled
-    as a background task
-    4. _dequeue checks the database for a runnable job and starts it (if present)
-    5. _notify_listeners runs all listeners in parallel, afterwards the task is marked as handled
+    2. when compile is queued, _process_next_compile_in_queue is called, if no job is executing for that environment, it is started
+    3. when a compile run ends:
+        a. it is marked as completed in the database
+        b. _notify_listeners notifies blocking listeners and schedules notification of async listeners as a background task.
+            When the background task finishes, the compile is marked as handled
+        c. _process_next_compile_in_queue is called again to start the next queued compile, if any
 
     Upon restart
-    1. find a runnable(incomplete) job for every environment (if any) and call _queue on it
+    1. call _process_next_compile_in_queue in all-environments mode to find a runnable (incomplete) request for every environment
+        and start it.
     2. find all unhandled but complete jobs and run _notify_listeners on them
     """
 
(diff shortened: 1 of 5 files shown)


sanderr:
@arnaudsjs @wouterdb I'm merging it now in the hopes of fixing the nightlies. I will still process review comments next week.

tests/conftest.py
```diff
@@ -1575,6 +1575,23 @@ def patch(compile: data.Compile, project_dir: str) -> CompileRun:
         return CompileRunnerMock(compile, make_compile_fail, runner_queue)
 
     monkeypatch.setattr(compilerslice, "_get_compile_runner", patch, raising=True)
+    monkeypatch.setattr(compilerslice, "_running_compiles", {}, raising=False)
```
wouterdb, line 1578: This could do with a comment.

src/inmanta/server/services/compilerservice.py
```diff
@@ -737,39 +744,74 @@ def _compile_merge_key(c: data.Compile) -> Hashable:
...
+
+                        # don't execute any compiles in a halted environment
+                        if env.halted:
+                            continue
+
+                        if (
+                            # no compile is running for this environment
+                            compile.environment not in self._compiling_envs
+                            # the previously running compile is finished
+                            or (finish_current_compile and compile.environment == environment)
```
arnaudsjs, line 805: I don't get this condition: `compile.environment == environment`?

arnaudsjs (changes requested):

tests/conftest.py
sanderr: I had to add this because I made the `_recompiles` dict into the `_compiling_envs` set. Since @arnaudsjs just had to change it back, I will revert this.

src/inmanta/server/services/compilerservice.py
sanderr: It's not strictly required because there will only ever be at most one compile in the list when `finish_current_compile` is set, due to the method invariants and guards, and it will be a compile for the given environment. In other words the condition will always be met in the current implementation.

So I had the choice between adding the condition or asserting it. I opted for the condition because I felt it's slightly less coupled. But I am not attached to it. If you feel like it is less readable I can go for the assert.

sanderr:
Opened new PR to address comments: #7566. Rerequesting review in case you have other comments.

src/inmanta/server/services/compilerservice.py
arnaudsjs: Isn't there a big behavioral difference between the assertion and the condition. Namely, the assertion will raise an exception and make us aware about the fact that the invariant was violated somehow. While the condition won't.

arnaudsjs (approved):

sanderr:
partially merged in bc5158c, rest delegated to #7566

Issue #6779, opened 2023-11-28
discovery resources: allow matching by id


sanderr:
allow matching discovered resources with managed resources. The following should be supported for sure:

- include link to corresponding managed resource
  - should this include orphans? If so, should we mark them? Double check with Bart and Guillaume
  - this returns a new API column: can be achieved with join between discovered resource table and persistent state table, optionally including resource table depending on orphan decision.
- exclude managed resources from the discovered resources endpoint's return values (via API filter)
  - this is considered to be a filter associated with the column from the first bullet point
- notify frontend that this has been added

Hugo-Inmanta:
> should this include orphans? If so, should we mark them? Double check with Bart and Guillaume

As discussed in standup Include orphans but don't specifically mark them as such

PR #7543, opened 2024-04-22
Issue/6779 discovery resources allow matching by

Hugo-Inmanta:
# Description

Closes https://github.com/inmanta/inmanta-core/issues/6779

# Self Check:

- [x] Changelog entry


arnaudsjs (changes requested):

changelogs/unreleased/6779-discovery-resources-allow-matching-by-id.yml
```diff
@@ -0,0 +1,6 @@
+description: >
+  Add filtering capabilities on DiscoveredResourceView. Discovered resource can now be filtered based on
+  whether they are already managed or not.
+issue-nr: 6779
+change-type: patch
+destination-branches: [master, iso7, iso6]
```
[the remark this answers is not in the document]
Hugo-Inmanta: Oh I see, the underlying classes are there but not the endpoints

arnaudsjs (changes requested):

src/inmanta/protocol/methods_v2.py
```diff
@@ -1599,6 +1602,14 @@ def discovered_resources_get_batch(
     :param sort: Return the results sorted according to the parameter value.
             The following sorting attributes are supported: 'discovered_resource_id'.
             The following orders are supported: 'asc', 'desc'
+    :param filter: Filter the list of returned resources.
+        Default behavior: return all discovered resources.
+        Filtering by 'resource_id' is supported:
+            - filter.resource_id=true: only return discovered resources that the orchestrator is already aware of i.e.
+            managed resources that make up the latest configuration model and orphaned resources (unpurged resources from
+            previous versions of the configuration model).
```
arnaudsjs, line 1610: I think this explanation is more complicated than required. I wouldn't use the term orphaned resource here, because it's ill-defined in this context.
```suggestion
            - filter.resource_id=true: only return discovered resources that the orchestrator is already aware of i.e.
            resources that are present in any configurationmodel of environment tid.
```

src/inmanta/protocol/methods_v2.py
```diff
@@ -1599,6 +1602,14 @@ def discovered_resources_get_batch(
...
+            - filter.resource_id=false: only return discovered resources that the orchestrator is unaware of i.e. resources
+            that are not part of any configuration model known to the orchestrator.
```
arnaudsjs, line 1612:
```suggestion
            that are not part of any configuration model of environment tid.
```

arnaudsjs (changes requested):

arnaudsjs:
@Hugo-Inmanta I think this PR doesn't handle the first bullet of the ticket: _"include link to corresponding managed resource"_

Hugo-Inmanta:
> @Hugo-Inmanta I think this PR doesn't handle the first bullet of the ticket: _"include link to corresponding managed resource"_

Ah indeed I'm missing a piece here.

src/inmanta/protocol/methods_v2.py
```diff
@@ -1599,6 +1602,14 @@ def discovered_resources_get_batch(
     :param sort: Return the results sorted according to the parameter value.
             The following sorting attributes are supported: 'discovered_resource_id'.
             The following orders are supported: 'asc', 'desc'
+    :param filter: Filter the list of returned resources.
+        Default behavior: return all discovered resources.
+        Filtering by 'resource_id' is supported:
+            - filter.resource_id=true: only return discovered resources that the orchestrator is already aware of i.e.
```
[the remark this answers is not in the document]
Hugo-Inmanta: Changed to `managed`

src/inmanta/data/dataview.py
```diff
@@ -1268,16 +1269,36 @@ def allowed_filters(self) -> dict[str, type[Filter]]:
...
     def get_base_url(self) -> str:
         return "/api/v2/discovered"
 
     def get_base_query(self) -> SimpleQueryBuilder:
         query_builder = SimpleQueryBuilder(
-            select_clause="SELECT environment, discovered_resource_id, values",
-            from_clause=f" FROM {data.DiscoveredResource.table_name()}",
-            filter_statements=["environment = $1"],
+            select_clause="SELECT *",
+            prelude=f"""
```
arnaudsjs, line 1282: I might be missing something here, but I don't see why you need the prelude.

src/inmanta/data/dataview.py
```diff
@@ -1268,16 +1269,36 @@ def allowed_filters(self) -> dict[str, type[Filter]]:
...
+                    (
+                        CASE
+                            WHEN rps.resource_id IS NOT NULL THEN true
+                            ELSE false
+                        END
+                    ) AS managed
+
+                FROM {data.DiscoveredResource.table_name()} as dr
+                LEFT JOIN {data.ResourcePersistentState.table_name()} rps
+                ON dr.discovered_resource_id = rps.resource_id
```
arnaudsjs, line 1297: I think you need to include the environment as well. Otherwise PostgreSQL might not be able to use the index when performing the join. Did you check the performance of this query?
```suggestion
                LEFT JOIN {data.ResourcePersistentState.table_name()} rps
                ON dr.environment=rps.environment AND dr.discovered_resource_id = rps.resource_id
```

src/inmanta/data/dataview.py
```diff
@@ -1268,16 +1269,36 @@ def allowed_filters(self) -> dict[str, type[Filter]]:
...
+                SELECT
+                    dr.environment,
+                    dr.discovered_resource_id,
+                    dr.values,
+                    (
+                        CASE
+                            WHEN rps.resource_id IS NOT NULL THEN true
+                            ELSE false
+                        END
+                    ) AS managed
```
arnaudsjs, line 1293:
```suggestion
                    (rps.resource_id IS NOT NULL)  AS managed
```

src/inmanta/data/dataview.py
```diff
@@ -1287,6 +1308,7 @@ def construct_dtos(self, records: Sequence[Record]) -> Sequence[dict[str, str]]:
             model.DiscoveredResource(
                 discovered_resource_id=res["discovered_resource_id"],
                 values=json.loads(res["values"]),
+                managed_resource_uri=f"/api/v2/resource/{res['discovered_resource_id']}" if res["managed"] else None,
```
arnaudsjs, line 1311: I assume this is correct, but don't know what would be correct here.

tests/test_discovered_resources.py
```diff
@@ -92,33 +95,102 @@ async def test_discovered_resource_create_batch(server, client, agent, environme
...
+        else:
+            expected_result = discovered_resources[-2:]
+
     result = await client.discovered_resources_get_batch(
         environment,
+        filter=filter,
     )
     assert result.code == 200
-    assert len(result.result["data"]) == 6
+    assert result.result["data"] == expected_result
```
arnaudsjs, line 188: This test case will be faster if you don't parameterize it and just put the three different call to the `discovered_resources_get_batch()` endpoint under each other. If you did it to prevent code duplication, then you can just put a for-loop around this code that loops through the different values of `apply_filter`.

arnaudsjs (changes requested):

src/inmanta/data/dataview.py
Hugo-Inmanta: Iirc I was getting `unknown column 'managed'` errors so I mimicked the behaviour from other queries involving filters.

src/inmanta/data/dataview.py
Hugo-Inmanta: Oof, much better :+1:

src/inmanta/data/dataview.py
Hugo-Inmanta: I'm not sure I understand what you mean ?

arnaudsjs: I mean that I don't know what the schema is of these URI's

Hugo-Inmanta: We have a ticket to document / standardize this [here](https://github.com/inmanta/inmanta-core/issues/7557)

src/inmanta/data/dataview.py
Hugo-Inmanta: It looks better indeed:
before:
```
                                                                       QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=27.10..27.12 rows=7 width=81) (actual time=0.098..0.104 rows=6 loops=1)
   ->  Sort  (cost=27.10..27.12 rows=7 width=81) (actual time=0.096..0.100 rows=6 loops=1)
         Sort Key: dr.discovered_resource_id
         Sort Method: quicksort  Memory: 25kB
         ->  Hash Right Join  (cost=12.69..27.01 rows=7 width=81) (actual time=0.058..0.070 rows=6 loops=1)
               Hash Cond: ((rps.resource_id)::text = (dr.discovered_resource_id)::text)
               ->  Seq Scan on resource_persistent_state rps  (cost=0.00..13.40 rows=340 width=32) (actual time=0.005..0.007 rows=4 loops=1)
               ->  Hash  (cost=12.64..12.64 rows=4 width=80) (actual time=0.038..0.039 rows=6 loops=1)
                     Buckets: 1024  Batches: 1  Memory Usage: 9kB
                     ->  Bitmap Heap Scan on discoveredresource dr  (cost=4.18..12.64 rows=4 width=80) (actual time=0.028..0.031 rows=6 loops=1)
                           Recheck Cond: (environment = '1a2d7489-c7dc-4e94-b0c1-1a35fd5170ba'::uuid)
                           Heap Blocks: exact=1
                           ->  Bitmap Index Scan on discoveredresource_pkey  (cost=0.00..4.18 rows=4 width=0) (actual time=0.016..0.016 rows=6 loops=1)
                                 Index Cond: (environment = '1a2d7489-c7dc-4e94-b0c1-1a35fd5170ba'::uuid)
 Planning Time: 0.372 ms
 Execution Time: 0.160 ms

```

After:
```
                                                                                      QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=22.20..22.24 rows=4 width=81) (actual time=0.081..0.096 rows=6 loops=1)
   ->  Merge Left Join  (cost=22.20..22.24 rows=4 width=81) (actual time=0.079..0.092 rows=6 loops=1)
         Merge Cond: ((dr.discovered_resource_id)::text = (rps.resource_id)::text)
         Join Filter: (dr.environment = rps.environment)
         ->  Sort  (cost=12.68..12.69 rows=4 width=80) (actual time=0.051..0.052 rows=6 loops=1)
               Sort Key: dr.discovered_resource_id
               Sort Method: quicksort  Memory: 25kB
               ->  Bitmap Heap Scan on discoveredresource dr  (cost=4.18..12.64 rows=4 width=80) (actual time=0.027..0.030 rows=6 loops=1)
                     Recheck Cond: (environment = '1a2d7489-c7dc-4e94-b0c1-1a35fd5170ba'::uuid)
                     Heap Blocks: exact=1
                     ->  Bitmap Index Scan on discoveredresource_pkey  (cost=0.00..4.18 rows=4 width=0) (actual time=0.016..0.016 rows=6 loops=1)
                           Index Cond: (environment = '1a2d7489-c7dc-4e94-b0c1-1a35fd5170ba'::uuid)
         ->  Sort  (cost=9.51..9.52 rows=2 width=48) (actual time=0.020..0.021 rows=4 loops=1)
               Sort Key: rps.resource_id
               Sort Method: quicksort  Memory: 25kB
               ->  Bitmap Heap Scan on resource_persistent_state rps  (cost=4.16..9.50 rows=2 width=48) (actual time=0.008..0.010 rows=4 loops=1)
                     Recheck Cond: (environment = '1a2d7489-c7dc-4e94-b0c1-1a35fd5170ba'::uuid)
                     Heap Blocks: exact=1
                     ->  Bitmap Index Scan on resource_persistent_state_environment_agent_resource_id_idx  (cost=0.00..4.16 rows=2 width=0) (actual time=0.004..0.004 rows=4 loops=1)
                           Index Cond: (environment = '1a2d7489-c7dc-4e94-b0c1-1a35fd5170ba'::uuid)
 Planning Time: 0.369 ms
 Execution Time: 0.165 ms
```
No seq scan when joining on env as well

changelogs/unreleased/6779-discovery-resources-allow-matching-by-id.yml
```diff
@@ -0,0 +1,8 @@
+description: >
+  Add filtering capabilities on DiscoveredResourceView. Discovered resource can now be filtered based on
+  whether they are already managed or not.
```
arnaudsjs, line 3: The word DiscoveredResourceView doesn't mean much to an end user.
```suggestion
  Add support to the `GET /api/v2/discovered` endpoint to filter the discovered resources on whether they are managed or not.
```

arnaudsjs (changes requested):

arnaudsjs (approved):

src/inmanta/data/dataview.py
```diff
@@ -1251,6 +1251,7 @@ def __init__(
         sort: str = "discovered_resource_id.asc",
         start: Optional[str] = None,
         end: Optional[str] = None,
+        filter: Optional[dict[str, list[str]]] = None,
```
sanderr, line 1254: Can we relax this to `Mapping` and `Sequence`?

src/inmanta/protocol/methods_v2.py
```diff
@@ -1588,8 +1588,12 @@ def discovered_resources_get_batch(
     start: Optional[str] = None,
     end: Optional[str] = None,
     sort: str = "discovered_resource_id.asc",
+    filter: Optional[dict[str, list[str]]] = None,
```
sanderr, line 1591: Same here, though we may not support the abstract types here.

src/inmanta/protocol/methods_v2.py
```diff
@@ -1599,6 +1603,13 @@ def discovered_resources_get_batch(
     :param sort: Return the results sorted according to the parameter value.
             The following sorting attributes are supported: 'discovered_resource_id'.
             The following orders are supported: 'asc', 'desc'
+    :param filter: Filter the list of returned resources.
+        Default behavior: return all discovered resources.
+        Filtering by 'managed' is supported:
+            - filter.managed=true: only return discovered resources that the orchestrator is already aware of i.e.
+            resources that are present in any configuration model of environment tid.
```
sanderr, line 1610:
```suggestion
            resources that are present in any released configuration model of environment tid.
```

src/inmanta/protocol/methods_v2.py
```diff
@@ -1599,6 +1603,13 @@ def discovered_resources_get_batch(
...
+            - filter.managed=false: only return discovered resources that the orchestrator is unaware of i.e. resources
+            that are not part of any configuration model of environment tid.
```
sanderr, line 1612:
```suggestion
            that are not part of any released configuration model of environment tid.
```

src/inmanta/server/services/resourceservice.py
```diff
@@ -1313,18 +1313,13 @@ async def discovered_resources_get_batch(
         start: Optional[str] = None,
         end: Optional[str] = None,
         sort: str = "discovered_resource_id.asc",
+        filter: Optional[dict[str, list[str]]] = None,
```
sanderr, line 1316: Same here

sanderr (changes requested):
Looks good. Some tiny remarks.

sanderr (approved):

PR #7545, opened 2024-04-22
added log shipping

wouterdb:
# Description

Added log shipper to IPC 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/protocol/ipc_light.py
+++ b/src/inmanta/protocol/ipc_light.py
@@ -18,6 +18,7 @@
 
 import abc
 import asyncio
+import functools
 import logging
 import pickle
 import struct
@@ -61,6 +62,14 @@ class IPCReplyFrame:
     id: uuid.UUID
     returnvalue: object
     is_exception: bool
+
+
+@dataclass
+class IPCLogRecord:
+
+    name: str
+    levelno: int
+    msg: str
 
 
 class IPCFrameProtocol(Protocol, typing.Generic[ServerContext]):
@@ -246,3 +255,41 @@ class FinalizingIPCClient(IPCClient[ServerContext]):
         super().connection_lost(exc)
         for fin in self.finalizers:
             asyncio.get_running_loop().create_task(fin())
+
+
+class LogReceiver(IPCFrameProtocol[ServerContext]):
+    """
+    IPC frame feature to receive log message
+    """
+
+    def frame_received(self, frame: IPCRequestFrame[ServerContext, ReturnType] | IPCReplyFrame) -> None:
+        if isinstance(frame, IPCLogRecord):
+            # calling log here is safe because if there are nu arguments, formatter is never called
+            logging.getLogger(frame.name).log(frame.levelno, frame.msg)
+        else:
+            super().frame_received(frame)
+
+
+class LogShipper(logging.Handler):
+    """
+    Log sender associated with the log receiver
+
+    This sender is threadsafe
+    """
+
+    def __init__(self, channel: IPCFrameProtocol[object], eventloop: asyncio.BaseEventLoop) -> None:
+        self.channel = channel
+        self.eventloop = eventloop
+        super().__init__()
+
+    def emit(self, record: logging.LogRecord) -> None:
+        self.eventloop.call_soon_threadsafe(
+            functools.partial(
+                self.channel.send_frame,
+                IPCLogRecord(
+                    record.name,
+                    record.levelno,
+                    self.format(record),
+                ),
+            )
+        )
--- a/tests/protocol/test_ipc.py
+++ b/tests/protocol/test_ipc.py
@@ -17,13 +17,18 @@
 """
 
 import asyncio
+import functools
+import logging
 import socket
 import struct
+import threading
 
 import pytest
 
 import inmanta.config
 import inmanta.protocol.ipc_light
+import inmanta.util
+import utils
 from inmanta.protocol.ipc_light import IPCClient, IPCServer
 
 
@@ -102,3 +107,38 @@ async def test_normal_flow(request):
     args = [1, 2, 3, 4]
     result = await client_protocol.call(Echo(args))
     assert args == result
+
+
+async def test_log_transport(caplog):
+
+    loop = asyncio.get_running_loop()
+    parent_conn, child_conn = socket.socketpair()
+
+    server_transport, server_protocol = await loop.connect_accepted_socket(
+        lambda: inmanta.protocol.ipc_light.LogReceiver("logs"), parent_conn
+    )
+    client_transport, client_protocol = await loop.connect_accepted_socket(lambda: IPCClient("Client"), child_conn)
+
+    log_drainer = inmanta.protocol.ipc_light.LogShipper(client_protocol)
+
+    with caplog.at_level(logging.INFO):
+        log_drainer.handle(logging.LogRecord("deep.in.test", logging.INFO, "xxx", 5, "Test %s", ("a",), exc_info=False))
+
+        def has_log(msg: str) -> bool:
+            try:
+                utils.log_contains(caplog, "deep.in.test", logging.INFO, f"Test {msg}")
+                return True
+            except AssertionError:
+                return False
+
+        await inmanta.util.retry_limited(functools.partial(has_log, "a"), 1)
+
+        # mess with threads
+        thread = threading.Thread(
+            target=lambda: log_drainer.handle(
+                logging.LogRecord("deep.in.test", logging.INFO, "xxx", 5, "Test %s", ("b",), exc_info=False)
+            )
+        )
+
+        thread.start()
+        await inmanta.util.retry_limited(functools.partial(has_log, "b"), 1)
--- a/changelogs/unreleased/ip_log_shipper.yml
+++ b/changelogs/unreleased/ip_log_shipper.yml
@@ -1,0 +1,3 @@
+description: Add log shipping to the IPCLight framework
+change-type: minor
+destination-branches: [master]


tests/protocol/test_ipc.py
```diff
@@ -102,3 +107,38 @@ def get_context(self) -> None:
     args = [1, 2, 3, 4]
     result = await client_protocol.call(Echo(args))
     assert args == result
+
+
+async def test_log_transport(caplog):
+
```
bartv, line 113: I think that every test case needs a description on what it does. The naming of the tests + variables is not enough imho.

Inconsistencies in terminology do not help either: shipper vs drainer

bartv (changes requested):

tests/protocol/test_ipc.py
```diff
@@ -102,3 +107,47 @@ def get_context(self) -> None:
...
+    loop = asyncio.get_running_loop()
+    parent_conn, child_conn = socket.socketpair()
+
+    server_transport, server_protocol = await loop.connect_accepted_socket(
+        lambda: inmanta.protocol.ipc_light.LogReceiver("logs"), parent_conn
+    )
+    request.addfinalizer(server_transport.close)
+    client_transport, client_protocol = await loop.connect_accepted_socket(lambda: IPCClient("Client"), child_conn)
+    request.addfinalizer(client_transport.close)
+    log_shipper = inmanta.protocol.ipc_light.LogShipper(client_protocol, loop)
```
bartv, line 131: I am quite confused here:

- you have a transport and a protocol
- but the logshipper has an argument channel to which you pass the client. However, based on naming I would have assumed that the transport is the channel?

wouterdb: updated

src/inmanta/protocol/ipc_light.py
```diff
@@ -246,3 +255,41 @@ def connection_lost(self, exc: Exception | None) -> None:
...
+    This sender is threadsafe
+    """
+
+    def __init__(self, protocol: IPCFrameProtocol[object], eventloop: asyncio.BaseEventLoop) -> None:
+        self.protocol = protocol
+        self.eventloop = eventloop
+        super().__init__()
+
+    def emit(self, record: logging.LogRecord) -> None:
+        self.eventloop.call_soon_threadsafe(
```
wouterdb, line 286: The failure behavior here is bit of an open question:

1. if the channel goes down
2. we raise a closed exception
3. that causes `call_soon_threadsafe`  to feed it to `self._loop.call_exception_handler`

Would it be better to silently drop log lines if we are disconnected?

bartv (approved):

src/inmanta/protocol/ipc_light.py
wouterdb: it will indeed terminate it, but logs produced during termination will be quite ugly (and invisible)

wouterdb: it even logs the failure, so it is a loop. I'll update it

src/inmanta/protocol/ipc_light.py
```diff
@@ -246,3 +255,55 @@ def connection_lost(self, exc: Exception | None) -> None:
...
+        self.eventloop = eventloop
+        self.logger_name = "inmanta.ipc.logs"
+        self.logger = logging.getLogger(self.logger_name)
+        super().__init__()
+
+    def _send_frame(self, record: IPCLogRecord) -> None:
+        try:
+            self.protocol.send_frame(record)
+        except:
+            # Stop excepption here
```
wouterdb, line 291:
```suggestion
            # Stop exception here
```

tests/protocol/test_ipc.py
```diff
@@ -102,3 +107,77 @@ def get_context(self) -> None:
...
+            )
+        )
+
+        thread.start()
+        await inmanta.util.retry_limited(functools.partial(has_log, "%d"), 1)
+
+        # Test we are not creating a log-loop when shut down
+        # This is a bit tricky to test, as we don't know how long to wait for a repeat log line to not appear
+
+        # This one should be ignored, as this is the loop protection
```
wouterdb, line 159:
```suggestion
        # This logger name should be ignored, as this is the loop protection
```

tests/utils.py
```diff
@@ -204,26 +204,26 @@ def __init__(self, caplog, index=0, allow_errors=True, ignore=[]):
         self.allow_errors = allow_errors
         self.ignore = ignore
 
-    def _find(self, loggerpart, level, msg, after=0):
+    def _find(self, loggerpart, level, msg, after=0, min_level=1):
```
arnaudsjs, line 207: Do you add a docstring about this new parameter? I think we also need some documentation regarding the default value that is used here.

tests/protocol/test_ipc.py
```diff
@@ -102,3 +107,77 @@ def get_context(self) -> None:
...
+
+        await inmanta.util.retry_limited(functools.partial(has_diverted_log), 1)
+        # Make sure we drain the buffer
+        await asyncio.sleep(0.1)
+        utils.log_doesnt_contain(caplog, log_shipper.logger_name, logging.INFO, "Test X")
+        # Log line is not repeated
+        # Not other log line after it
+        utils.LogSequence(caplog).contains(log_shipper.logger_name, logging.INFO, "Could not send log line").assert_not(
+            "", logging.INFO, "", logging.INFO
+        )
```
arnaudsjs, line 183:
```suggestion
        utils.LogSequence(caplog).contains(log_shipper.logger_name, logging.INFO, "Could not send log line").assert_not(
           loggerpart="", level=logging.INFO, msg="", min_level=logging.INFO
        )
```

tests/utils.py
```diff
@@ -204,26 +204,26 @@ def __init__(self, caplog, index=0, allow_errors=True, ignore=[]):
...
         for i, (logger_name, log_level, message) in enumerate(self.caplog.record_tuples[after:]):
             if msg in message:
-                if loggerpart in logger_name and level == log_level:
+                if loggerpart in logger_name and (level == log_level or (log_level < min_level)):
```
arnaudsjs, line 210: Isn't `min_level` a strange name here when doing the comparison `log_level < min_level`. It defines an upper limit. Shouldn't it be `max_level`?

src/inmanta/protocol/ipc_light.py
```diff
@@ -246,3 +255,55 @@ def connection_lost(self, exc: Exception | None) -> None:
...
+        except Exception:
+            # Stop exception here
+            # Log in own logger to prevent loops
+            self.logger.info("Could not send log line", exc_info=True)
+            return
+
+    def emit(self, record: logging.LogRecord) -> None:
+        if record.name == self.logger_name:
+            # avoid loops
+            return
```
arnaudsjs, line 299: Should you elaborate on this. The log only flow in one direction. How would it be possible to get a loop?

arnaudsjs (changes requested):

src/inmanta/protocol/ipc_light.py
wouterdb: When the logger fails to send a log message, it creates a log message, ...

tests/utils.py
wouterdb: this was plain wrong

tests/utils.py
```diff
@@ -204,26 +204,48 @@ def __init__(self, caplog, index=0, allow_errors=True, ignore=[]):
         self.allow_errors = allow_errors
         self.ignore = ignore
 
-    def _find(self, loggerpart, level, msg, after=0):
+    def _find(self, loggerpart, level, msg, after=0, min_level: int = 0):
```
arnaudsjs, line 207: This is a non-backwards compatible change. Is that intentional? The original implementation required the given log level to match the log level of a record. The new implementation will consider any log level by default, independent of the value of `level`.

arnaudsjs (changes requested):

tests/utils.py
arnaudsjs: Or do you mean `and` instead of `or`?

sanderr (changes requested):
Only had a high level overview because if I understand correctly Arnaud is reviewing thoroughly and you really only needed one of us to review it.

Overall, the idea looks good to me, but I wonder on how exactly it will integrate. The sender will need to have the appropriate logging config, but it will need to override all handlers with the IPC one. Right? Perhaps this invariant should be part of the `LogShipper`'s docstring (and potentially other classes')? Because a user has to know which side is supposed to do filtering, formatting, log level, ...

tests/utils.py
wouterdb: oh man, I managed to mess it up again.

wouterdb:
> Only had a high level overview because if I understand correctly Arnaud is reviewing thoroughly and you really only needed one of us to review it.
> 
> Overall, the idea looks good to me, but I wonder on how exactly it will integrate. The sender will need to have the appropriate logging config, but it will need to override all handlers with the IPC one. Right? Perhaps this invariant should be part of the `LogShipper`'s docstring (and potentially other classes')? Because a user has to know which side is supposed to do filtering, formatting, log level, ...

Well, the sender is a regular log handler, in any sense and the receiver a normal log produce, also in any sense. 

So nothing special applies, filtering can be done on either side.
The only thing that is special, is that it connects a produce to a consumer, so it will be an infinite loop if connected to itself.

sanderr:
Discussed in person. My comment was partially incorrect. The part that stands: docstrings could be more explicit about the scope of these classes, i.e. what do they do on the receiving side (pass control to the logging framework), and what is expected to be preprocessed on the sending side (nothing).

arnaudsjs (approved):

PR #7561, opened 2024-04-24
Scheduler code

wouterdb:
# Review

I would like to get general approval on what is here, before we decide how to proceed. 

# Description

Potential implementation for queue scheduler 

1. built a scheduler toolbox
2. integrated with the agent
3. added deploy, dryrun and getfact

Todo
- [x] add exception handling to the scheduler, now it stop working on exception
- [ ] the version based cache is a mess now. I would propose to make it a timed cache that keep every version open for a fixed amount of time
- [x] resume after suspend has to be replaced with priorities in the queue
- [x] cleanup the full queue

# Self Check:

- [ ] Changelog entry


src/inmanta/agent/agent.py
```diff
@@ -374,89 +350,72 @@ def reload(
...
+                    )
+                    return
+                else:
+                    self.logger.info("Upgrading run '%s' to '%s'", self.active_repair.origin.reason, new_request.reason)
+                    self.active_repair.run.cancel()
+            self.active_repair = build_run(PRIO_PERIODIC)
+        else:
+            # Just do as you are told
+            if self.active_deploy is not None and not self.active_deploy.run.finished():
+                self.logger.info("Canceling run '%s' in favor of '%s'", self.active_repair.origin.reason, new_request.reason)
```
arnaudsjs, line 416:
```suggestion
                self.logger.info("Canceling run '%s' in favor of '%s'", self.active_deploy.origin.reason, new_request.reason)
```

arnaudsjs, line 416: This part I don't understand.

src/inmanta/agent/agent.py
```diff
@@ -524,6 +476,9 @@ def __init__(self, process: "Agent", name: str, uri: str, *, ensure_deploy_on_st
 
         # init
         self._nq = ResourceScheduler(self, self._env_id, name)
+        self.work_queue = scheduler.TaskQueue()
+        self.work_queue_drainer = scheduler.TaskRunner(self.work_queue)
```
arnaudsjs, line 480: Why have two different terms for the same thing (`queue_drainer` vs. `TaskRunner`)?

src/inmanta/agent/agent.py
```diff
@@ -754,52 +713,49 @@ async def dryrun(self, dry_run_id: uuid.UUID, version: int) -> Apireturn:
         return 200
 
     async def do_run_dryrun(self, version: int, dry_run_id: uuid.UUID) -> None:
-        async with self.dryrunlock:
-            async with self.ratelimiter:
```
arnaudsjs, line 758: Why do we remove the ratelimiter here?

arnaudsjs (changes requested):

src/inmanta/agent/agent.py
wouterdb: the queue does that now

arnaudsjs: How?

wouterdb: We know that we can run only one task at a time, because only a single logical thread drains the queue,...

src/inmanta/scheduler/scheduler.py
```diff
@@ -0,0 +1,318 @@
...
+    """Logical thread to process work from a task queue
+
+    Can be restarted
+    """
+
+    def __init__(self, queue: TaskQueue) -> None:
+        self.queue = queue
+        self.running = False
+        self.should_run = True
+        self.finished: typing.Optional[asyncio.Task[None]] = None
```
sanderr, line 271: `finished` is a strange name for a task object.

src/inmanta/scheduler/scheduler.py
```diff
@@ -0,0 +1,318 @@
...
+
+    async def join(self) -> None:
+        assert not self.should_run
+        if self.finished:
+            await self.finished
+
+    async def run(self) -> None:
+        self.running = True
+        while self.should_run:
+            task = await self.queue.get()
```
sanderr, line 289: Doesn't this raise an `IndexError` when the queue is exhausted?

src/inmanta/scheduler/scheduler.py
```diff
@@ -0,0 +1,318 @@
...
+        self.queue = queue
+        self.running = False
+        self.should_run = True
+        self.finished: typing.Optional[asyncio.Task[None]] = None
+
+    def start(self) -> None:
+        self.should_run = True
+        self.finished = asyncio.create_task(self.run())
+
+    def stop(self) -> None:
```
sanderr, line 277: This stops the runner asyncronously. I think it could use a docstring on what exactly it achieves.

sanderr (commented):
Reviewed `scheduler.py`. Have yet to review the rest.

src/inmanta/agent/agent.py
wouterdb: changed

src/inmanta/scheduler/scheduler.py
```diff
@@ -0,0 +1,318 @@
...
+        """Are we cancelled"""
+        return self.cancelled
+
+    def is_running(self) -> bool:
+        return self.started and not self.done
+
+    def is_done(self) -> bool:
+        return self.done
+
+    def cancel(self) -> None:
```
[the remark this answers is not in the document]
wouterdb: I'm going to leave this as a todo for later stage I think. 

I'm not entirely clear on what we expect from the cancel.

e.g. 
- canceling a deploy for a resource that has already reached its desired state is equal to deploy done, but if it was previously failed, it equals failure. 
- if we store the future of the execution in the task, we can even cancel pre-emptively. 

It is not unlikely this entire dependency mechanism will change.

src/inmanta/scheduler/scheduler.py
```diff
@@ -0,0 +1,318 @@
...
+        if self._queue:
+            self._queue.full_queue.drop(self._entry)
+            if self.waitcount == 0 and not self.started:
+                self._queue.run_queue.drop(self._entry)
+
+    def wait_for(self, other: "BaseTask") -> None:
+        """
+        Declare we wait for some other task
+
+        Adding the same task twice will make this one wait forever!
```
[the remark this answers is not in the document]
wouterdb: good catch

src/inmanta/scheduler/scheduler.py
wouterdb: nope, it waits  (it is a blocking queue)



```    

Remove and return an item from the queue.
If queue is empty, wait until an item is available.
```

tests/agent_server/test_server_agent.py
```diff
@@ -2137,6 +2137,7 @@ async def test_reload(
     assert dep_state.index == resource_container.Provider.reloadcount("agent1", "key2")
 
 
+@pytest.mark.skip("Skipped for agent refactor")
```
arnaudsjs, line 2140: We have to be more descriptive in this message. Why do we need to skip this test?

sanderr (commented):
On hold until further notice.

PR #7565, opened 2024-04-25
Added docs

wouterdb:
# Description

Documentation for https://github.com/inmanta/lsm/pull/732
closes https://github.com/inmanta/inmanta-lsm/issues/1603
# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -29,7 +29,7 @@ SVG2PDF       = inkscape
 SVG2PDF_FLAGS =
 
 # Internal variables.
-ALLSPHINXOPTS   = -t oss -d $(BUILDDIR)/doctrees $(SPHINXOPTS) .
+ALLSPHINXOPTS   = -t iso -d $(BUILDDIR)/doctrees $(SPHINXOPTS) .
 # the i18n builder cannot share the environment and doctrees with the others
 I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
@@ -67,8 +67,7 @@ clean:
 	-rm -rf $(BUILDDIR)/*
 	-rm -rf $(PNGs)
 
-html: openapi latexpdf
-	cp $(BUILDDIR)/latex/inmanta.pdf $(SOURCEDIR)
+html: openapi
 	$(SPHINXBUILD) -b html -t include_redoc $(ALLSPHINXOPTS) $(BUILDDIR)/html
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -29,7 +29,7 @@ SVG2PDF       = inkscape
 SVG2PDF_FLAGS =
 
 # Internal variables.
-ALLSPHINXOPTS   = -t iso -d $(BUILDDIR)/doctrees $(SPHINXOPTS) .
+ALLSPHINXOPTS   = -t oss -d $(BUILDDIR)/doctrees $(SPHINXOPTS) .
 # the i18n builder cannot share the environment and doctrees with the others
 I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
--- a/docs/lsm/partial_compiles/partial_compiles.rst
+++ b/docs/lsm/partial_compiles/partial_compiles.rst
@@ -18,6 +18,7 @@ Implementation guidelines
 2. for every :ref:`Inter Service Relation<inter_service_relations>`
 
     1. indicate if this is the relation to the owner by setting :inmanta:attribute:`lsm::ServiceEntityBinding.relation_to_owner` and :inmanta:relation:`lsm::ServiceEntityBinding.owner`.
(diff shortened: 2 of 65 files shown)


docs/Makefile
```diff
@@ -29,7 +29,7 @@ SVG2PDF       = inkscape
 SVG2PDF_FLAGS =
 
 # Internal variables.
-ALLSPHINXOPTS   = -t oss -d $(BUILDDIR)/doctrees $(SPHINXOPTS) .
+ALLSPHINXOPTS   = -t iso -d $(BUILDDIR)/doctrees $(SPHINXOPTS) .
```
wouterdb, line 32:
```suggestion
ALLSPHINXOPTS   = -t oss -d $(BUILDDIR)/doctrees $(SPHINXOPTS) .
```

docs/lsm/partial_compiles/partial_compiles.rst
```diff
@@ -127,20 +129,57 @@ The underlying mechanism is that when we recompile for a state change on any Ser
...
+The selector is a piece of python code that determines which instances will be returned by ``lsm::all``. 
+
+The API for the selector is :py:class:`inmanta_plugins.lsm.partial.SelectorAPI`.
+To register the selector, use :py:meth:`inmanta_plugins.lsm.global_cache.set_selector_factory<inmanta_plugins.lsm.CacheManager.set_selector_factory>`.
+
+For example, consider a case where we have a `tunnel` service that connects two `ports`. To compile the tunnel, we also need to load the ports. 
+This is also not a tree, as each tunnel has not one, but two parents. As such, we can build a custom selector.
+
+
+.. literalinclude:: custom_partial.cf
```
wouterdb, line 147: This pull in a lot of docs, but requires a fix in the docs build!

sanderr:
My review may have to wait for a bit. I'm considering other reviews and the stuck orchestrator fix to be priorities right now.

sanderr (changes requested):
Looking at the example now makes me consider the implementation more critically. I wonder if this is not more complex for the end user than it needs to be.

1. the implementation of `select_all` looks pretty generic to me for the most part. Couldn't we just have the user implement a single `expand(service_entity_name: str, instance_id: ?) -> set[tuple[str, ?]]` that expands one instance one step by following its connected edges? All else can then be done in the default implementation in lsm.
2. The registration isn't very modular: a random module's plugins are now suddenly the authority for lsm's behavior. What if multiple modules need custom selectors? Wouldn't this be better set from the model through a plugin (explicitly)? And shouldn't we provide a mechanism to compose them somehow, e.g. have a `TunnelSelector` for tunnels and ports, and the default `TreeSelector` for all other entity types. If we simplify the interface for 1, composing them could be trivial, and we might even have a separate `TunnelSelector` to expand tunnels into ports and `PortSelector` to expand ports into tunnels (if that were required, which isn't the case here).

Requesting changes just to make sure my comments are not missed so we can discuss it.

wouterdb:
@sanderr 

for 1: we don't do that, because this is (expected to become) extremely performance critical. Forcing this algorithm to only take one instance at a time does not fit the domain very well. It is often easier to consider all instances (if they are all under one parent e.g.)  and it is easy to go from this api to what you suggest, the reverse is not. 
for 2: I agree. However, in this case (which has not happened yet, services are usually deployed on-per-environment) can easily be supported by adding a combiner. The reason the combiner doesn't exist yet, is that we don't know how it will distinguish what will go to which child.

sanderr:
1. Do you have an example where it doesn't work? The example in this PR does pretty much consider them one by one, but perhaps that is a simplification of the practical structures? I still feel like we expose a lot of complexity to the end user here, but I can accept it if you think this is the best approach.
2. Without 1 I think a combiner might get even more complicated. But again, something we can refine later if need be. ~~For now, how would you feel about making registration an explicit plugin call (which can be part of the project / driving module) rather than an load-time side effect?~~ Damn, I just realized we can not make it so reliably, because the compiler might execute `lsm::all` first. I'm giving this a few minutes more thought, but I'm afraid I may just have to accept this as well.

sanderr (approved):
I posted one more comment, mostly to ask some more context to get a better understanding of the context myself. Shouldn't block the merge.

wouterdb:
@sanderr examples:
- https://code.inmanta.com/solutions/bics/systemtenant/-/merge_requests/655
- https://code.inmanta.com/solutions/modules/connect/-/merge_requests/573

sanderr:
Thanks!

sanderr:
@wouterdb I have to confess, I still don't see how those two selectors are all that different. As far as I can tell, they loop over the requested instances, check their type, and each type has distinct dependencies / responsibilities to include others. e.g. for systemtenant, "standardgroup" includes "standardflow" and "systemtenant", "standardflow" does the other direction, and "systemtenant" includes all its children. Essentially we're building up an acyclic directed graph, where each node type determines the edges for its instances. As long as we only visit each node once, we build up the full graph, don't we?

The main difference I see is that some cases need to interact with the global cache more intimately because they have to do e.g. reverse relation lookups. But that doesn't mean we can't simplify the other cases, while making it more composable in the process.

PR #7566, opened 2024-04-25
compiler service refinement and review comments

sanderr:
# Description

- fixed incorrect error message regarding db transaction
- processed review comments from #7536

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -640,7 +640,8 @@ class CompilerService(ServerSlice, environmentservice.EnvironmentListener):
                                       if set to false, nothing will be notified. If not set then the default notifications are
                                       sent (failed pull stage and errors during the do_export)
         :param failed_compile_message: the message used in notifications if notify_failed_compile is set to True.
-        :param connection: Perform the database changes using this database connection.
+        :param connection: Perform the database changes using this database connection. If this connection is in a transaction
+                           context, in_db_transaction must be set to True.
         :param in_db_transaction: If set to True, the connection must be provided and the connection must be part of an ongoing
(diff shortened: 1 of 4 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -652,10 +653,13 @@ async def request_recompile(
         """
         if in_db_transaction and not connection:
             raise Exception("A connection should be provided when in_db_transaction is True.")
+        # TODO: can we afford this "breaking" change?
+        #   -> breaks current stable inmanta-lsm (can be worked around here), works without
+        #   -> breaks old inmanta-lsm, compiles may hang without
         if in_db_transaction != (connection is not None and connection.is_in_transaction()):
             raise Exception(
                 f"in_db_transaction is {in_db_transaction},"
```
sanderr, line 661: I added this error in the original PR. It is technically a breaking change but I think we can afford it because we always release a full product anyway, and the only user is inmanta-lsm, and the previous behavior was bugged:
1. For current stable inmanta-lsm, this change would raise an exception because of a typing error in inmanta-lsm. If the exception weren't raised, the behavior would be as expected.
2. For older inmanta-lsm than current stable, this change would also raise an exception because it uses a transaction without setting the flag. If the exception weren't raised, the behavior would still be unstable (this is why we added the flag in the first place).

Let me know if you think we have to remain compatible.

arnaudsjs (approved):

sanderr:
@arnaudsjs requesting rereview for e32647b31, as stated in https://github.com/inmanta/inmanta-core/pull/7536#discussion_r1580662179.

arnaudsjs (approved):

wouterdb (approved):

PR #7513, opened 2024-04-11
Issue/1921 extend version overview with more versions

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/irt/issues/1921

# Self Check:

Strike through any lines that are not applicable (`~~line~~`) then check the box

- [ ] Attached issue to pull request
- [ ] Changelog entry
- [ ] Type annotations are present
- [ ] Code is clear and sufficiently documented
- [ ] No (preventable) type errors (check using make mypy or make mypy-diff)
- [ ] Sufficient test cases (reproduces the bug/tests the requested feature)
- [ ] Correct, in line with design
- [ ] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )
- [ ] If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)

PR #7569, opened 2024-04-26
Fix core build failures

Hugo-Inmanta:
# Description

Revert the tox.ini changes until the irt changes are in again https://github.com/inmanta/irt/pull/2022

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/issue-fix-core-build-failure.yml
+++ b/changelogs/unreleased/issue-fix-core-build-failure.yml
@@ -1,0 +1,5 @@
+change-type: patch
+description: Fix core build failure
+destination-branches:
+- master
+sections: {}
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/issue-fix-core-build-failure.yml
```diff
@@ -0,0 +1,5 @@
+change-type: patch
+description: Fix core build failure
+destination-branches:
+- master
```
sanderr, line 4: Shouldn't this be all branches?

sanderr (approved):

changelogs/unreleased/issue-fix-core-build-failure.yml
Hugo-Inmanta: It only made it to master

Hugo-Inmanta: This [PR](https://github.com/inmanta/inmanta-core/pull/7513) I mean only made it to master and got rejected on iso7 and iso6 because of failing tests

sanderr: You'll probably need to delete/update its change entry then, or merge dummy ones into the other branches

PR #7582, opened 2024-04-30
Use updated env variable containing path for constraint file in tox.ini

Hugo-Inmanta:
# Description

Follow up to https://github.com/inmanta/inmanta-core/pull/7569
Original ticket https://github.com/inmanta/irt/issues/1921

The following PRs close the core part of the ticket:

* (this PR) master pr update tox ini https://github.com/inmanta/inmanta-core/pull/7582
* iso7 pr compat page improvement changes https://github.com/inmanta/inmanta-core/pull/7583
* iso6 pr compat page improvement changes https://github.com/inmanta/inmanta-core/pull/7584

This PR closes the irt part of the ticket:

- PR adding generation of new content for compat files: https://github.com/inmanta/irt/pull/2027

This PR adds a fix for autoclasses not being rendered properly for inmanta modules:

- fix autoclass docs PR https://github.com/inmanta/irt/pull/2028

# Self Check:

Strike through any lines that are not applicable (`~~line~~`) then check the box

- [ ] Attached issue to pull request
- [ ] Changelog entry
- [ ] Type annotations are present
- [ ] Code is clear and sufficiently documented
- [ ] No (preventable) type errors (check using make mypy or make mypy-diff)
- [ ] Sufficient test cases (reproduces the bug/tests the requested feature)
- [ ] Correct, in line with design
- [ ] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )
- [ ] If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)

PR #7583, opened 2024-04-30
Issue/1921 iso7 stage 2

Hugo-Inmanta:
# Description

iso7 specific PR (same as https://github.com/inmanta/inmanta-core/pull/7513)
Context here: https://github.com/inmanta/inmanta-core/pull/7582

except for:

change entry: `changelogs/unreleased/1921-extend-version-overview-with-more-versions-iso7.yml`
- new name
- iso7 only

# Self Check:

- [ ] Changelog entry


changelogs/unreleased/1921-extend-version-overview-with-more-versions-iso7.yml
sanderr, line 1: This should have the same name as the one that was merged in master. This is required for irt to recognize them as representing the same change. Why is that important?
1. In this specific case because the other one declares iso7 as destination branch, so we need to ensure the change is present on iso7 one way or another.
2. In general, but not relevant here because there is no changelog entry, `irt` does some filtering when we generate the change entries for a new major. It bases itself on the name of this file to know which changelog entries have already been mentioned in older releases of the previous major.

Hugo-Inmanta: First of all, sorry for the mess.
During the reverting PR I removed iso7 and iso6 from the dest branches for this change entry https://github.com/inmanta/inmanta-core/commit/280804f1e574ffc258957a847f912286ee45f4c9.
Do I understand correctly that I should add them back now that the files will be renamed ?

sanderr (approved):

PR #7597, opened 2024-05-06
issue/7274-Add command that output default logger config

### LINKED ISSUE #7274: Create commands to dump the default logging config
Create the following CLI command:

* inmanta server default-logging-config
* inmanta agent default-logging-config
* inmanta scheduler default-logging-config
* inmanta compile-service default-logging-config
* inmanta-cli default-logging-config

These command should write the default logging config for the relevant component to disk

FloLey:
# Description

Add command that output default logger config

closes #7274 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -324,6 +324,32 @@ def help_command(options: argparse.Namespace) -> None:
     sys.exit(0)
 
 
+@command("show-default-logger-config", help_msg="Display the default logger configuration")
+def show_default_logger_config(options: argparse.Namespace) -> None:
+    default_config = """
+version: 1
+formatters:
+  console_formatter:
+    (): inmanta.logging.MultiLineFormatter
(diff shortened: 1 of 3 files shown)


src/inmanta/app.py
```diff
@@ -324,6 +324,32 @@ def help_command(options: argparse.Namespace) -> None:
     sys.exit(0)
 
 
+@command("show-default-logger-config", help_msg="Display the default logger configuration")
+def show_default_logger_config(options: argparse.Namespace) -> None:
+    default_config = """
```
arnaudsjs, line 329: The idea was to generate this config dynamically based on the `LoggingConfigBuilder` class. Otherwise there is the risk that this hardcoded default config will over time become out of sync with the config generated by the `LoggingConfigBuilder` class.

arnaudsjs (changes requested):

src/inmanta/app.py
FloLey: I won't have time to do this ticket, I'll put it back in the next column.

PR #7598, opened 2024-05-06
Add get_current_resource_state() method

arnaudsjs:
# Description

Add the get_current_resource_state() method to the database layer. This method is required by https://github.com/inmanta/inmanta-lsm/pull/1718.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4635,12 +4635,18 @@ class Resource(BaseDocument):
         return {ResourceIdStr(r["resource_id"]): ResourceState[r["status"]] for r in out}
 
     @classmethod
-    async def get_current_resource_state(
+    async def get_status_for_v2(
         cls,
         env: uuid.UUID,
         model_version: int,
         rid: ResourceIdStr,
     ) -> Optional[ResourceState]:
+        """
+        Retrieve the ResourceState for the given resource. The difference between
+        this method and the get_status_for() method is that this method relies on the
+        resource_persistent_state table if the state is requested for a resource
+        in the latest released version of the configurationmodel.
+        """
         query = f"""
             SELECT (
                 SELECT max(c.version) AS version
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4635,44 +4635,36 @@ class Resource(BaseDocument):
         return {ResourceIdStr(r["resource_id"]): ResourceState[r["status"]] for r in out}
 
     @classmethod
-    async def get_status_for_v2(
-        cls,
-        env: uuid.UUID,
-        model_version: int,
-        rid: ResourceIdStr,
-    ) -> Optional[ResourceState]:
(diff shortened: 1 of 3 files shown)


src/inmanta/data/__init__.py
```diff
@@ -4634,6 +4634,43 @@ async def get_status_for(
         out = await cls.select_query(query, [env, model_version, rids], no_obj=True)
         return {ResourceIdStr(r["resource_id"]): ResourceState[r["status"]] for r in out}
 
+    @classmethod
+    async def get_current_resource_state(
```
wouterdb, line 4638: this could use some documentation. 

I don't understand why a 'get_current` also gets a model version as an argument?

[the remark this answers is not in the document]
arnaudsjs: That's a leftover from a previous implementation. I'll fix it.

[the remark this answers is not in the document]
arnaudsjs: Given the review of my other PR, we should actually just remove the `model_version` argument.

src/inmanta/data/__init__.py
```diff
@@ -4634,6 +4634,49 @@ async def get_status_for(
         out = await cls.select_query(query, [env, model_version, rids], no_obj=True)
         return {ResourceIdStr(r["resource_id"]): ResourceState[r["status"]] for r in out}
 
+    @classmethod
+    async def get_status_for_v2(
```
sanderr, line 4638: Perhaps a more informative name would be `get_status_for_version` for the other method, and plain `get_status_for` / `get_status` for this one? As I understand it this one gets the current status of a resource, while the other gets the status for a specific version of a resource.

src/inmanta/data/__init__.py
```diff
@@ -4634,6 +4634,49 @@ async def get_status_for(
         out = await cls.select_query(query, [env, model_version, rids], no_obj=True)
         return {ResourceIdStr(r["resource_id"]): ResourceState[r["status"]] for r in out}
 
+    @classmethod
```
sanderr, line 4637:
```suggestion
    @stable_api
    @classmethod
```

sanderr (changes requested):
I agree with the comment to drop the `model_version` parameter. Other than that two tiny comments.

sanderr (approved):

src/inmanta/data/__init__.py
```diff
@@ -4634,6 +4634,42 @@ async def get_status_for(
         out = await cls.select_query(query, [env, model_version, rids], no_obj=True)
         return {ResourceIdStr(r["resource_id"]): ResourceState[r["status"]] for r in out}
 
+    @stable_api
+    @classmethod
+    async def get_current_resource_state(cls, env: uuid.UUID, rid: ResourceIdStr) -> Optional[ResourceState]:
```
wouterdb, line 4639: Do we need to also include the check on the hash or this that for another PR?

wouterdb (approved):

PR #7609, opened 2024-05-08
Fix inconsistent state reporting on the resource endpoints

arnaudsjs:
# Description

Fix bug where the `GET /api/v2/resource/<rid>` and `GET /api/v2/resource` endpoints return an incorrect resourcestate if a resource moved back to the available state in a new version of the configurationmodel.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5111,7 +5111,7 @@ class Resource(BaseDocument):
         inner_query = f"""
         SELECT r.resource_id as resource_id,
         (
-            CASE WHEN (r.status = 'deploying')
+            CASE WHEN r.status IN ('deploying', 'undefined', 'skipped_for_undefined')
                 THEN
                     r.status::text
                 ELSE
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4641,34 +4641,40 @@ class Resource(BaseDocument):
         Return the state of the given resource in the latest version of the configuration model
         or None if the resource is not present in the latest version.
         """
-        query = f"""
+        query = """
+            WITH latest_released_version AS (
+                SELECT max(version) AS version
+                FROM configurationmodel
+                WHERE environment=$1 AND released
+            )
             SELECT (
-                SELECT r.status
-                FROM resource r
-                WHERE r.environment=$1
-                    AND r.model=(
-                        SELECT max(c.version) AS version
-                        FROM {ConfigurationModel.table_name()} c
-                        WHERE c.environment=$1 AND c.released
-                    )
-                    AND r.resource_id=$2
-            ) AS status,
-            (
-                SELECT rps.last_non_deploying_status
-                FROM resource_persistent_state rps
-                WHERE rps.environment=$1 AND rps.resource_id=$2
-            ) AS last_non_deploying_status
+                CASE
+                    -- The resource_persistent_state.last_non_deploying_status column is only populated for
+                    -- actual deployment operations to prevent locking issues. This case-statement calculates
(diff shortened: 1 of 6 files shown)


src/inmanta/data/__init__.py
```diff
@@ -4641,34 +4641,40 @@ async def get_current_resource_state(cls, env: uuid.UUID, rid: ResourceIdStr) ->
...
-                WHERE rps.environment=$1 AND rps.resource_id=$2
-            ) AS last_non_deploying_status
+                CASE
+                    -- The resource_persistent_state.last_non_deploying_status column is only populated for
+                    -- actual deployment operations to prevent locking issues. This case-statement calculates
+                    -- the correct state from the combination of the resource table and the
+                    -- resource_persistent_state table.
+                    WHEN r.status::text = 'deploying'
+                        THEN r.status::text
+                    WHEN r.status = 'available' AND rps.last_non_deploying_status IN('undefined', 'skipped_for_undefined')
```
wouterdb, line 4658: nice

sanderr (commented):
Commenting so you can rerequest when the dust has settled, i.e. the [discussion on Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1715161636398029) and associated change(s).

src/inmanta/data/__init__.py
```diff
@@ -4641,34 +4641,39 @@ async def get_current_resource_state(cls, env: uuid.UUID, rid: ResourceIdStr) ->
...
+                    -- actual deployment operations to prevent locking issues. This case-statement calculates
+                    -- the correct state from the combination of the resource table and the
+                    -- resource_persistent_state table.
+                    WHEN r.status::text IN('deploying', 'undefined', 'skipped_for_undefined')
+                        -- The deploying, undefined and skipped_for_undefined states are not tracked in the
+                        -- resource_persistent_state table.
+                        THEN r.status::text
+                    WHEN rps.last_deployed_attribute_hash != r.attribute_hash
+                        -- The hash changed since the last deploy -> new desired state
+                        THEN 'available'
```
sanderr, line 4662: Could this be `unavailable` as well? i.e. do we consider that "deployed" for the sake of the hash update? We may want to simply return `r.status::text` in this case as well, or am I missing something?

sanderr (commented):
I'm adding a link to the related [Slack discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1715161636398029) here because I bet I'll come looking for its context through git blame at some point.

I'll also release comments as I go to speed up the process.

src/inmanta/data/dataview.py
```diff
@@ -562,12 +562,21 @@ def get_base_query(self) -> SimpleQueryBuilder:
...
+                                    -- The deploying, undefined and skipped_for_undefined states are not tracked in the
+                                    -- resource_persistent_state table.
                                     THEN r.status::text
-                                ELSE
-                                    rps.last_non_deploying_status::text
+                                WHEN rps.last_deployed_attribute_hash != r.attribute_hash
+                                   -- The hash changed since the last deploy -> new desired state
+                                   THEN 'available'
+                                   -- No override required, use last known state from actual deployment
+                                   ELSE rps.last_non_deploying_status::text
```
sanderr, line 579: I'm a bit scared of the performance impact of this one. I'm writing down that we have to check on mgmt after updating.

sanderr (changes requested):
Looks mostly good, added a few minor comments.

src/inmanta/data/__init__.py
arnaudsjs: I think we can return `r.status::text` in this case.

sanderr (approved):

PR #7610, opened 2024-05-08
Issue/7356 remove deprecated std iso6

FloLey:
# Description

drop all usages of deprecated std resources from core test

(https://github.com/suite/inmanta-core/issues/7356)

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -1148,11 +1148,11 @@ async def test_mark_done(init_dataclasses_and_load_schema, resource_state, versi
 
 
 async def populate_model(env_id, version):
-    def get_path(n):
+    def get_name(n):
         return "/tmp/%d" % n
 
     def get_id(n):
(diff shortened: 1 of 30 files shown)


tests/test_data.py
```diff
@@ -422,33 +425,6 @@ async def assert_setting_in_db(expected_autostart_agent_map: dict[str, object])
     assert assert_setting_in_db(autostart_agent_map)
 
 
-async def test_environment_deprecated_setting(init_dataclasses_and_load_schema, caplog):
```
arnaudsjs, line 425: Why was this test case removed?

arnaudsjs (changes requested):

tests/test_data.py
FloLey: This was a mistake, I added it back. I'll leave it up to you to merge this PR :) Thanks for everything

arnaudsjs (approved):

PR #7612, opened 2024-05-10
Ensure agents improvements

sanderr:
# Description

Fixes bug in autostarted agent manager that occasionally causes agent's sessions to time out, dropping calls, resulting in a "stuck" orchestrator until a redeploy is triggered.

~~The bug has been there forever, but the conditions for it to trigger happen to be set up by #7278.~~

~~The core of the bug is the following. When we need a certain agent to be up, we call `AutostartedAgentManager._ensure_agents`. This makes sure an agent process is running for these agents and waits for them to be up. However, instead of starting a process to track the autostarted agent map and to trust that it would keep up with changes to it, we instead start a process with an explicit list of agent endpoints.~~ If a new call comes in to `_ensure_agents`, and the agent is not yet up, we would kill the current process and start a new one. In killing the process, we would not expire its sessions, letting them time out on the 30s heartbeat timeout, losing any calls made to it in that window.

EDIT: Scratched some of the above: I wrote a test case for the first bullet point below, which I thought to be the root cause of the reason for killing the old agent process. However, turns out the test also succeeds on master. The agent's `on_reconnect` actually updates the process' agent map. So, I think my root cause analysis may have been wrong (at least less likely), but the second and third bullet points should fix the issue anyway (doesn't matter exactly *how* we got in the situation where only part of the agent endpoints were up, as long as we handle it correctly), and even if part of the issue persists, the logging improvements should help future investigation. And the first bullet point is still a good consistency fix imo.

This PR makes the following changes:
- An agent process for autostarted agents (`use_autostart_agent_map` in the config file) now ignores any agent endpoints in the config file. Instead it runs purely in autostarted agent mode, starting instances for each of the endpoints in the agent map. It then tracks any changes made to the agent map. This last part was already in place, and resulted in a small inconsistency where the process would respect agent names in the config at start, and then suddenly move to consider the agent map the authority as soon as a change comes in. This inconsistency is now resolved by considering the agent map the authority for the entire lifetime of the agent process.
- The autostarted agent manager now trusts in its processes to track the agent map. If a process is already running for an environment, but the agent endpoints are not yet up, it waits for them rather than to kill the process and start fresh.
- When an explicit restart is requested, the autostarted agent manager now expires all sessions for the agents in the agent map, i.e. the endpoints for the killed process.
- Improved robustness of wait condition for an agent to be up, specifically make sure we don't consider expired sessions to be up.
- Made some log messages a bit more informative, no major changes.

# Self Check:

- [x] Changelog entry


src/inmanta/server/agentmanager.py
```diff
@@ -1295,6 +1267,75 @@ async def _fork_inmanta(
             if errhandle is not None:
                 errhandle.close()
 
+    async def _wait_for_agents(
```
sanderr, line 1270: Moved this from an embedded function in `_ensure_agents` to a dedicated method. I made very limited changes: update the signature and the docstring, and I'll mark all other changes with a comment.

src/inmanta/server/agentmanager.py
```diff
@@ -1295,6 +1267,75 @@ async def _fork_inmanta(
...
+                    len(actual_agents_in_up_state),
+                    len(expected_agents_in_up_state),
+                )
+            new_actual_agents_in_up_state = {
+                agent_name
+                for agent_name in expected_agents_in_up_state
+                if (
+                    (session := self._agent_manager.tid_endpoint_to_session.get((env.id, agent_name), None)) is not None
+                    # make sure to check for expiry because sessions are unregistered from the agent manager asyncronously
+                    and not session.expired
```
sanderr, line 1325: I added this condition: a session might still be in the `tid_endpoint_to_session` mapping while it has already been expired (if the agent manager hasn't processed the expiry event yet).

tests/test_agent_manager.py
```diff
@@ -1207,34 +1207,6 @@ async def _dummy_fork_inmanta(
     assert exception_message in str(excinfo.value)
 
 
-async def test_are_agents_active(server, client, environment, agent_factory) -> None:
```
sanderr, line 1210: These were unit tests for two methods that I dropped in this PR.

tests/agent_server/test_server_agent.py
```diff
@@ -1511,6 +1510,52 @@ async def assert_session_state(expected_agent_states: dict[str, AgentStatus], ex
     assert len(new_agent_processes) == 0, new_agent_processes
 
 
+@pytest.mark.parametrize("autostarted", (True, False))
+async def test_autostart_mapping_overrides_config(server, client, environment, async_finalizer, caplog, autostarted: bool):
```
sanderr, line 1514: This test succeeds on master as well (see note in PR description) but I left it because it's a good check to have.

tests/test_agent_manager.py
```diff
@@ -1278,6 +1253,99 @@ async def test_dont_start_paused_agent(server, client, environment, caplog) -> N
     assert "took too long to start" not in caplog.text
 
 
+async def test_wait_for_agent_map_update(server, client, environment) -> None:
```
sanderr, line 1256: I verified that both these tests fail on master. They verify the core change in this PR.

changelogs/unreleased/ensure-agents-improvements.yml
```diff
@@ -0,0 +1,10 @@
+description: "Made various improvements to the AutostartedAgent._ensure_agents method"
+sections:
+  bugfix: "Fixed a race condition where autostarted agents might become unresponsive for 30s when restarted"
```
sanderr, line 3: I found this one difficult to write. I think this captures the gist of the user-relevant part.

wouterdb: good enough for me

src/inmanta/agent/agent.py
```diff
@@ -977,16 +977,18 @@ async def _init_agent_map(self) -> None:
...
-                    if "$" in name:
-                        name = name.replace("$node-name", self.node_name)
-                    await self.add_end_point_name(name)
+        assert self.agent_map is not None
+        endpoints: Iterable[str] = (
+            [self.hostname]
+            if self.hostname is not None
+            else (
+                self.agent_map.keys()
+                if cfg.use_autostart_agent_map.get()
```
wouterdb, line 986: Doesn't this duplicate the logic of the method above (`_init_agent_map`)?

src/inmanta/server/agentmanager.py
```diff
@@ -635,20 +635,15 @@ def get_agent_client(self, tid: uuid.UUID, endpoint: str, live_agent_only: bool
         else:
             return None
 
-    async def are_agents_active(self, tid: uuid.UUID, endpoints: list[str]) -> bool:
+    async def expire_sessions_for_agents(self, env_id: uuid.UUID, endpoints: Set[str]) -> None:
```
wouterdb, line 638: I think this is not entirely as intended: if the same endpoint exists both on the autostarted agent and another one, if will kill the session of both agents.

I would expect the kill to affect the agent's session only (i.e. all session you close here are intended to be the same, the agent process maintains a single session, not one per endpoint)

Am I correct that the major complication is that there is no clear way to distinguish which one is the autostarted one?

src/inmanta/server/agentmanager.py
```diff
@@ -990,6 +990,33 @@ async def stop_agents(self, env: data.Environment, delete_venv: bool = False) ->
...
+            await self._wait_for_proc_bounded([subproc])
+            del self._agent_procs[env.id]
+
+        # fetch the agent map after stopping the process to prevent races with agent map update notifying the process
+        agent_map: Mapping[str, str] = cast(
+            Mapping[str, str], await env.get(data.AUTOSTART_AGENT_MAP, connection=connection)
+        )  # we know the type of this map
+
+        LOGGER.debug("Expiring sessions for autostarted agents %s", sorted(agent_map.keys()))
+        await self._agent_manager.expire_sessions_for_agents(env.id, agent_map.keys())
```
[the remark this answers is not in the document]
wouterdb: For specific debugging scenarios (when the autostarted agent is acting up or being restarted too often), we have resorted to taking over the endpoints from another agent process. It is a rare, but we have done it. 

But this fix may make that unnecessary.

src/inmanta/server/agentmanager.py
wouterdb: An alternative solution here would be to implement the `on_connection_close` method of the request handler. 

The autostarted agent socket is always local, so it closes fast. But it is a whole different line of work going into the depths of the RPC framework. So I just mention it, but I would not necessarily do it.

wouterdb (commented):
I would prefer the autostarted agent to indicate that it is the autostarted agent when setting up its session. 

This would make the behavior easier to understand, but I won't stand in the way either

src/inmanta/server/agentmanager.py
```diff
@@ -1300,6 +1272,75 @@ async def _fork_inmanta(
...
+                    now - started,
+                    len(actual_agents_in_up_state),
+                    len(expected_agents_in_up_state),
+                )
+            new_actual_agents_in_up_state = {
+                agent_name
+                for agent_name in expected_agents_in_up_state
+                if (
+                    (session := self._agent_manager.tid_endpoint_to_session.get((env.id, agent_name), None)) is not None
+                    # make sure to check for expiry because sessions are unregistered from the agent manager asyncronously
```
arnaudsjs, line 1329:
```suggestion
                    # make sure to check for expiry because sessions are unregistered from the agent manager asynchronously
```

tests/agent_server/test_server_agent.py
```diff
@@ -1511,6 +1510,53 @@ async def assert_session_state(expected_agent_states: dict[str, AgentStatus], ex
...
+    )
+    assert result.code == 200
+
+    # Start agent
+    a = agent.Agent(environment=env_uuid, code_loader=False)
+    await a.start()
+    async_finalizer(a.stop)
+
+    # Wait until agents are up
+    await retry_limited(lambda: len(agent_manager.tid_endpoint_to_session) == (2 if autostarted else 1), 2)
```
arnaudsjs, line 1543:
```suggestion
    await retry_limited(lambda: len(agent_manager.tid_endpoint_to_session) == (2 if autostarted else 1), timeout=2)
```

arnaudsjs (changes requested):

src/inmanta/agent/agent.py
sanderr: I don't think so. The agent map maps endpoint names to uris, this method determines which endpoints to add. Only in the case of the autostarted agent do we take the agent map as the authority for both.

src/inmanta/agent/agent.py
```diff
@@ -1069,6 +1071,9 @@ async def update_agent_map(self, agent_map: dict[str, str]) -> None:
             await self._update_agent_map(agent_map)
 
     async def _update_agent_map(self, agent_map: dict[str, str]) -> None:
+        if "internal" not in agent_map:
+            raise ValueError("The internal agent must be present in the agent map")
+
```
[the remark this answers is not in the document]
sanderr: I'm not sure. The current behavior was to drop the internal agent altogether, so I don't think we have to worry too much about it being a breaking change (the previous behavior is just breaking in a different way). In that sense I think we can afford it, and if we can afford it I think it's the sensible thing to do. It also corresponds with the environment setting update behavior.

src/inmanta/server/agentmanager.py
sanderr: Discussed and concluded [on Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1715593454444139).

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
sanderr: @wouterdb what's your opinion on this?

src/inmanta/server/agentmanager.py
```diff
@@ -1029,62 +1056,80 @@ async def _terminate_agents(self) -> None:
...
-        agent_map: dict[str, str] = cast(
-            dict[str, str], await env.get(data.AUTOSTART_AGENT_MAP, connection=connection)
-        )  # we know the type of this map
-
-        agents = [agent for agent in agents if agent in agent_map]
-        needsstart = restart
-        if len(agents) == 0:
-            return False
+        if connection is not None and connection.is_in_transaction():
+            raise Exception("_ensure_agents should not be called in a transaction context")
```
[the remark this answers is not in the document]
sanderr: Turns out I should have made a note for myself even, I'd forgotten the specifics of it. I made an effort to make it clear though.

src/inmanta/agent/agent.py
wouterdb: right

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
wouterdb: I think it is strange for the agent to reject the entire agent map, in case an entry is missing (that is can easily put in there itself).

I would understand if we reject the user input on the server side, here, I find it tricky, in that it can make existing, working setups stop working altogether. (this is code on the agent startup path)

wouterdb (changes requested):

src/inmanta/agent/agent.py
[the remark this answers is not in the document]
sanderr: Ok, I'll change this.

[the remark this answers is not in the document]
sanderr: 1. added a warning, I think this is definitely good.
2. auto-add it when it's missing. I could also just keep this part to the old behavior if you prefer.

wouterdb (approved):

arnaudsjs (approved):

sanderr:
iso6 merged in 97c04954f

PR #7634, opened 2024-05-16
Issue/auth on longpoll

wouterdb:
# Description

Fix longpoll issue introduced by https://github.com/inmanta/inmanta-core/pull/7627

1. #7627 centralized the auth code into `execute_call`
2. however,  `execute_call` is not only used on the server: the agent uses it to dispatch methods it received via the heartbeat. 
3. They have no authentication token, so now they fail.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/test_2way_protocol.py
+++ b/tests/test_2way_protocol.py
@@ -26,6 +26,7 @@ from pytest import fixture
 from tornado.gen import sleep
 
 from conftest import configure_auth
+
 # Methods need to be defined before the Client class is loaded by Python
 from inmanta import protocol  # NOQA
 from inmanta import data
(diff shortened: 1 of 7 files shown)


tests/test_2way_protocol.py
```diff
@@ -125,9 +127,9 @@ def is_same():
     await retry_limited(is_same, 10)
 
 
-async def test_2way_protocol(unused_tcp_port, no_tid_check, postgres_db, database_name):
-    configure(unused_tcp_port, database_name, postgres_db.port)
-
+async def test_2way_protocol(inmanta_config, server_config, no_tid_check, postgres_db, database_name):
+    # Authentication complicates this even further
+    configure_auth(True, False, False)
```
arnaudsjs, line 132:
```suggestion
    configure_auth(auth=True, ca=False, ssl=False)
```

arnaudsjs (changes requested):

bartv (approved):

arnaudsjs (approved):

Issue #7557, opened 2024-04-24
document the format used for reference to the api itself


wouterdb:
# Problem

Currently, the notifications can carry links to compiles. 
We may also pass other links to the front-end, but it is unclear if and where.

The front-end must be able to dispatch these links to the correct page.

We should document all possible links we can emit, so they can handle them all.

Hugo-Inmanta:
For reference, this PR https://github.com/inmanta/inmanta-core/pull/7543 adds another reference to the api:
- the `discovered_resources_get_batch` endpoint returns a link with format `"/api/v2/resource/<rid>"`

PR #7648, opened 2024-05-21
Issue/7557 document the format used to reference the api itself

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/7557

~~TODO: When this is approved, open a separate PR for iso6 without the "managed resource" link.~~

# Self Check:

- [x] Changelog entry


Hugo-Inmanta:
image

docs/reference/apilinks.rst
```diff
@@ -0,0 +1,16 @@
+API self-referencing links
+==========================
+
+Some API endpoints can provide additional information or context by returning links to other API endpoints.
+The following table documents all these links and their format:
+
+
+.. list-table:: Self-referencing links
```
wouterdb, line 8: Can we reference this from the api spec itself, where it is used?

wouterdb:
I would keep the full doc in ISO6, only makes it easier

docs/reference/apilinks.rst
Hugo-Inmanta: Added a reference to the description of the endpoints (see image below). I tried adding this reference further down the tree, e.g. in the description of the uri of the response object, but I didn't managed to.
image

wouterdb: Good enough I guess.

wouterdb:
I would definitely send out a link to the team to look at it once it is public

PR #7660, opened 2024-05-23
Cleaner exception when influxdb connection fails

### LINKED ISSUE #7622: Produce cleaner exception if influxdb connection fails
We currently log 

```
2024-05-14 09:40:32,450 WARNING  inmanta.reporter Cannot create database inmanta to ####
  File "/opt/inmanta/lib64/python3.11/site-packages/inmanta/reporter.py", line 125, in _create_database
    response = await http_client.fetch(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tornado.simple_httpclient.HTTPTimeoutError: Timeout while connecting
``` 

This doesn't provide any indication this is about influxdb at all. 
It should clearly mention that it fail to connect to influxdb.

jptrindade:
# Description

Changed the exception to explicitly mention influxdb so it is more clear

closes #7622 

# Self Check:

- [x] Changelog entry
- [x] ~~No (preventable) type errors (check using make mypy or make mypy-diff)~~

### DIFF
--- a/src/inmanta/reporter.py
+++ b/src/inmanta/reporter.py
@@ -127,7 +127,7 @@ class InfluxReporter(AsyncReporter):
             # Only set if we actually were able to get a successful response
             self._did_create_database = True
         except Exception:
-            LOGGER.warning("Cannot create database %s to %s", self.database, self.server, exc_info=True)
+            LOGGER.warning("Failed to connect or create to influx database %s to %s", self.database, self.server, exc_info=True)
 
     async def report_now(self, registry: Optional[MetricsRegistry] = None, timestamp: Optional[float] = None) -> None:
         http_client = AsyncHTTPClient()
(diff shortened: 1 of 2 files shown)


src/inmanta/reporter.py
```diff
@@ -127,7 +127,7 @@ async def _create_database(self, http_client: AsyncHTTPClient) -> None:
             # Only set if we actually were able to get a successful response
             self._did_create_database = True
         except Exception:
-            LOGGER.warning("Cannot create database %s to %s", self.database, self.server, exc_info=True)
+            LOGGER.warning("Failed to connect or create to influx database %s to %s", self.database, self.server, exc_info=True)
```
wouterdb, line 130:
```suggestion
            LOGGER.warning("Failed to connect to or create the influx database %s on %s", self.database, self.server, exc_info=True)
```

wouterdb (approved):

PR #7678, opened 2024-05-27
Issue/1667 dont run background tasks during migration tests

Hugo-Inmanta:
# Description

Closes https://github.com/inmanta/inmanta-lsm/issues/1667

Since background taks are now halted by default, we can revert https://github.com/inmanta/inmanta-core/pull/7399/files where the env had to be halted in tests/db/migration_tests/dumps/v202403110.sql 

# Self Check:

- [x] Changelog entry


arnaudsjs (changes requested):

arnaudsjs (changes requested):
This PR doesn't disable all background jobs. Some missing jobs:

* Purging of old model versions
* Purging resource action logs
* Renewing of facts
* Purging of agent processes
* Check of database pool exhaustion (But maybe we want to keep this one and then we should mention that clearly in the docstring of the fixture).

Hugo-Inmanta:
> Purging of old model versions

I think this one is already taken care of by this auto-use fixture:

https://github.com/inmanta/inmanta-core/blob/e92be0cdc79519018060188c10eb57f6f2ccb389/tests/conftest.py#L1921-L1930

arnaudsjs:
> > Purging of old model versions
> 
> I think this one is already taken care of by this auto-use fixture:
> 
> https://github.com/inmanta/inmanta-core/blob/e92be0cdc79519018060188c10eb57f6f2ccb389/tests/conftest.py#L1920-L1930

I see. Let's mention this in the docstring of the `disable_background_tasks` fixture.

arnaudsjs (changes requested):
See comment.

arnaudsjs (approved):

src/inmanta/server/services/__init__.py
```diff
@@ -15,3 +15,6 @@
 
     Contact: code@inmanta.com
 """
+
+# This variable can be updated by the test suite to disable the scheduling of background jobs on service start.
+DONT_RUN_BACKGROUND_JOBS = False
```
sanderr, line 20: Perhaps `DISABLE_BACKGROUND_JOBS`? It's the same in essence but it feels less like a confusion double negation to me.

sanderr (changes requested):
Wouldn't it be better to make this a responsibility of `inmanta.util.Scheduler` instead of each client seperately?

sanderr (approved):
Nice, I think this is a lot cleaner. ~I'm rerequesting @arnaudsjs because I requested drastic changes.~ EDIT: you already did

tests/conftest.py
```diff
@@ -642,6 +643,26 @@ def server_pre_start(server_config):
     """This fixture is called by the server. Override this fixture to influence server config"""
 
 
+@pytest.fixture
+def disable_background_jobs(monkeypatch):
+    """
+    This fixture disables the scheduling of background jobs that could interfere with testing
+    e.g. cleanup jobs.
```
arnaudsjs, line 650:
```suggestion
    This fixture disables the scheduling of all background jobs.
```

tests/db/migration_tests/dumps/v202403110.sql
```diff
@@ -616,7 +616,7 @@ COPY public.dryrun (id, environment, model, date, total, todo, resources) FROM s
 
 COPY public.environment (id, name, project, repo_url, repo_branch, settings, last_version, halted, description, icon) FROM stdin;
 834fde5a-e5d3-4443-9ab7-04034bfdbec9	dev-2	0873d9b8-1964-4fbb-b26a-193ed5307d90			{"auto_full_compile": ""}	0	f		
-e03eabc9-5342-4e6b-b72e-6a062bab9e52	dev-1	0873d9b8-1964-4fbb-b26a-193ed5307d90			{"auto_deploy": false, "server_compile": true, "auto_full_compile": "", "recompile_backoff": 0.1, "autostart_agent_map": {"internal": "local:", "localhost": "local:"}, "autostart_agent_deploy_interval": "0", "autostart_agent_repair_interval": "600", "autostart_agent_deploy_splay_time": 0, "autostart_agent_repair_splay_time": 0}	7	t		
+e03eabc9-5342-4e6b-b72e-6a062bab9e52	dev-1	0873d9b8-1964-4fbb-b26a-193ed5307d90			{"auto_deploy": false, "server_compile": true, "auto_full_compile": "", "recompile_backoff": 0.1, "autostart_agent_map": {"internal": "local:", "localhost": "local:"}, "autostart_agent_deploy_interval": "0", "autostart_agent_repair_interval": "600", "autostart_agent_deploy_splay_time": 0, "autostart_agent_repair_splay_time": 0}	7	f
```
arnaudsjs, line 619: Why is this change required?

arnaudsjs (changes requested):

tests/db/migration_tests/dumps/v202403110.sql
Hugo-Inmanta: We halted the environment as part of https://github.com/inmanta/inmanta-core/pull/7399/files
Now that we disable the background tasks I think this we can revert it ?
Maybe this is not necessary / not relevant and it might add some complication because the above-mentioned PR was only merged into master and iso7, while the current PR will go into iso6 also.

Hugo-Inmanta: Remove it as it is not strictly required and might cause complications for iso6

PR #7684, opened 2024-05-29
Add a local client that uses RPC but does not use a network connection

bartv:
# Description

This PR does a number of things

- Introduce a local client that does not use a network connection.
- Introduce a typed client that already parses some of the json and restores exceptions
- Move protocol tests in the protocol folder

# Self Check:

- [ ] Changelog entry


wouterdb (approved):

src/inmanta/protocol/endpoints.py
```diff
@@ -474,3 +477,65 @@ async def _call(
...
+            case _:
+                self._raise_exception(exceptions.ServerError, response.result)
+
+    async def _call(
+        self, method_properties: common.MethodProperties, args: list[object], kwargs: dict[str, object]
+    ) -> types.ReturnTypes:
+        """Execute a call and return the result"""
+        return self._process_response(method_properties, await self._transport_instance.call(method_properties, args, kwargs))
+
+    def __getattr__(self, name: str) -> Callable[..., Coroutine[Any, Any, types.ReturnTypes]]:
```
wouterdb, line 539: why is this here? (It looks like the default behavior of python re-defined)

wouterdb (approved):

src/inmanta/protocol/endpoints.py
bartv: I removed it

tests/protocol/test_typed_client.py
```diff
@@ -0,0 +1,108 @@
...
+from inmanta import protocol
+from inmanta.data.model import BaseModel
+from inmanta.postgresproc import PostgresProc
+from inmanta.protocol import exceptions
+from inmanta.server.protocol import LocalClient, Server, ServerSlice, common
+from utils import configure
+
+
+async def test_local_client(unused_tcp_port: int, postgres_db: PostgresProc, database_name: str, async_finalizer) -> None:
+    """The the local clien"""
```
arnaudsjs, line 32:
```suggestion
    """Test the local client"""
```

arnaudsjs (approved):

Issue #7498, opened 2024-04-05
Documentation menu collapses


arnaudsjs:
When clicking a sub-element from the `Inmanta Lifecycle Service Manager` menu entry from the documentation pages, the menu tree collapses.

[Screencast from 2024-04-05 16-27-30.webm](https://github.com/inmanta/inmanta-core/assets/2684622/4f4cfd93-07ad-413c-a0e7-daac5e1552df)

PR #7687, opened 2024-05-30
Issue/7498 documentation menu collapses

Hugo-Inmanta:
# Description

Fix bug where clicking a sub-element from the ``Inmanta Lifecycle Service Manager`` menu would collapse the whole menu.
Fix similar behavior for the iso-only pages of the `Administrator Documentation` section (`Operational Procedures With LSM` and `Support Procedure`)

Bug was caused because we were using the `.. only:: ...` directive to control structure (which sections to include) whereas it should only be used for including/excluding content within a fixed structure. 
Use the `exclude_patterns` functionality of the `conf.py` file instead.

Behavior on iso:
- check that clicking items no longer collapses the menu
- check that previous page / next page links work as intended. i.e. from the `Platform Developers Guide` page, the next page should be `Inmanta Lifecycle Service Manager`

[Screencast from 30-05-2024 10:46:12.webm](https://github.com/inmanta/inmanta-core/assets/98876549/6739e102-a7b8-4e85-81e2-7e12422d7e74)

Behavior on oss:
- check that the `Inmanta Lifecycle Service Manager` page is not present in the menu
- and that it is not accessible either via next page links. i.e. from the `Platform Developers Guide` page, the next page should be `Administrator documentation`
- also check it is not accessible via entering url directly e.g. by navigating to `http://0.0.0.0:8000/lsm`

[Screencast from 30-05-2024 10:43:29.webm](https://github.com/inmanta/inmanta-core/assets/98876549/e282620a-425f-408f-958d-54588cc0c981)

The current approach generates the following warnings when building oss:

```console
WARNING: toctree contains reference to excluded document 'administrators/operational_procedures_with_lsm'
WARNING: toctree contains reference to excluded document 'administrators/support'
WARNING: toctree contains reference to nonexisting document 'lsm/index'
```

~~I'm not sure what to do with these warnings, either accept them or suppress them via `conf.py` e.g.:~~
EDIT: After discussion I think it's fine to leave these warnings as they are.

```python

if tags.has("oss"):
    # Suppress warnings when building oss docs because some toc trees
    # reference files that we explicitely exclude (e.g. administrators/support.rst)
    suppress_warnings = ["toc.excluded", "toc.not_readable"]
```

closes https://github.com/inmanta/inmanta-core/issues/7498

# Self Check:

- [x] Changelog entry


docs/administrators/operational_procedures_with_lsm.rst
```diff
@@ -1,9 +1,9 @@
-.. only:: iso
+.. _operational_procedures_lsm:
 
-    .. _operational_procedures_lsm:
+Operational Procedures With LSM
+--------------------------------
 
-    Operational Procedures With LSM
-    --------------------------------
+.. only:: iso
```
arnaudsjs, line 6: I don't think we need the `only` directive here and on the other pages that are iso only, because those pages are excluded using the `exclude_patterns` configuration option. Like that we would have two way to exclude things from oss:

* The `exclude_patterns` configuration option if a whole page needs to be excluded.
* The `only` directive if need to exclude only parts of a page.

Combining both mechanisms on a single page is not very useful I think. Wdyt?

docs/conf.py
```diff
@@ -156,7 +156,19 @@ def check_dot_command():
...
+default_pattern = ['adr/*.md']
+exclude_from_oss = [
+    "lsm",
+    "administrators/operational_procedures_with_lsm.rst",
+    "administrators/support.rst"
+]
+exclude_patterns = default_pattern + exclude_from_oss
+
+if tags.has("iso"):
+    exclude_patterns = default_pattern
```
arnaudsjs, line 171: I think we need to add a comment here to indicate why this cannot be achieved using the `only` directive.

arnaudsjs (changes requested):

docs/administrators/operational_procedures_with_lsm.rst
Hugo-Inmanta: > Combining both mechanisms on a single page is not very useful I think. Wdyt?

I agree, my rationale was to have extra safety to make sure the iso-only content is not shown e.g. if we inadvertently change the exclude config of the conf.py file. I think it is more confusing than useful, I'll remove it

docs/conf.py
sanderr: I've been doing some reading on this so I'll add some links that I found interesting:
- [docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-only) state "This directive is designed to control only content of document. It could not control sections, labels and so on."
- sphinx-doc/sphinx#9819
- sphinx-doc/sphinx#10642

docs/conf.py
```diff
@@ -156,7 +156,22 @@ def check_dot_command():
...
```
sanderr, line 174: This is a very strange pattern imo:
```
default = ...
x = default + ...  # add something to it
if c:
  x = default  # changed our minds
```

Instead of this back-and-forth, wouldn't it be a lot more straightforward to either define it at once (functional, my preference), or at least to only move in one direction (not add first, then redefine: only add).

In addition to this, the behavior is not exactly the same as with the `only` directives. Perhaps we want to condition on `not tags.has("iso")` and `not tags.has("oss")` instead.

sanderr (changes requested):

docs/conf.py
```diff
@@ -156,7 +156,22 @@ def check_dot_command():
...
+
+default_pattern = ['adr/*.md']
+exclude_from_oss = [
+    "lsm",
+    "administrators/operational_procedures_with_lsm.rst",
+    "administrators/support.rst"
+]
+exclude_patterns = default_pattern + exclude_from_oss
+
+if tags.has("iso"):
```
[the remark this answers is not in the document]
Hugo-Inmanta: Quickly checked them and as far as I can tell they seem to render fine even though we use them to control the structure in some places e.g.:
https://github.com/inmanta/inmanta-core/blame/c484bc4273274ba3e8579166f7b3ae62f820876e/docs/install/1-install-server.rst#L13-L16

Issue #7531, opened 2024-04-18
race condition in exporter file upload


sanderr:
# Solution
- [ ] Make the file upload endpoint idempotent: if the file already exists, silently ignore the request
- [ ] Double check that the exporter can handle this new behavior

# Problem
The exporter calls `stat_files` to determine which files to upload. However, files are identified by their hash (across environments), while the compiler/exporter is only serialized within an environment. Therefore a race condition exists where one environment would upload a file (e.g. an empty `__init__.py`) in the window between another environment's `stat_file` and `upload_file`, causing that export to fail.

I observed this behavior by running `pytest --log-level-cli=DEBUG --count=1000 -x tests/server/test_workon.py::test_workon_list_no_api`. It occurs now and then, and it's visible as an error in the test setup log. This may not be a stable reproduction, given that this test case is currently unstable and therefore its behavior may be changed to address that.

PR #7707, opened 2024-06-05
Issue/7531 race condition in exporter file upload

Hugo-Inmanta:
# Description

- [x] Make the file upload endpoint idempotent: if the file already exists, silently ignore the request
- [x] Double check that the exporter can handle this new behavior

closes https://github.com/inmanta/inmanta-core/issues/7531

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
+++ b/changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
@@ -1,4 +1,6 @@
-description: Trying to upload a file that is already present in the database will now silently fail.
+description: >
+  The upload_file endpoint will now silently ignore attempts to upload a file if a file with the same
+  hash was previously uploaded.
 issue-nr: 7531
 change-type: minor
 destination-branches: [master, iso7]
--- a/changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
+++ b/changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
@@ -2,8 +2,8 @@ description: >
   The upload_file endpoint will now silently ignore attempts to upload a file if a file with the same
   hash was previously uploaded.
 issue-nr: 7531
-change-type: minor
-destination-branches: [master, iso7]
+change-type: patch
+destination-branches: [master, iso7, iso6]
 sections:
   bugfix: >
     Fix race condition where exporting a file might fail if a file with similar content
--- a/changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
+++ b/changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
@@ -6,5 +6,5 @@ change-type: patch
 destination-branches: [master, iso7, iso6]
 sections:
   bugfix: >
-    Fix race condition where exporting a file might fail if a file with similar content
+    Fix race condition where exporting a file might fail if a file with the same content
     was uploaded between the file existence check in the database and the export itself.
--- a/changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
+++ b/changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
@@ -1,0 +1,9 @@
+description: Trying to upload a file that is already present in the database will now silently fail.
+issue-nr: 7531
+change-type: minor
+destination-branches: [master, iso7]
+sections:
+  minor-improvement: "{{description}}"
+  bugfix: >
+    Fix race condition where exporting a file might fail if a file with similar content
+    was uploaded between the file existence check in the database and the export itself.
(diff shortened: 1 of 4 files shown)


Hugo-Inmanta:
> Double check that the exporter can handle this new behavior

For this point this seems ok as far as I can tell, and the test suite didn't go up in smoke. I'm not sure if this is enough or what else needs to be done to check this.

tests/protocol/test_protocol.py
```diff
@@ -136,6 +136,28 @@ async def test_client_files_stat(client):
...
+        assert len(result.result["files"]) == 0
+
+    (hash, content, body) = make_random_file()
+
+    result = await client.stat_files(files=[hash])
+    assert result.code == 200
+    assert len(result.result["files"]) == 1
+
+    await upload_and_stat(body, client, hash)
+    await upload_and_stat(body, client, hash)
```
arnaudsjs, line 158: To test whether the concurrency issue is resolved, you should run these two API calls concurrently.
```suggestion
    await asyncio.gather(upload_and_stat(body, client, hash),  upload_and_stat(body, client, hash))
```

changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
```diff
@@ -0,0 +1,11 @@
+description: >
+  The upload_file endpoint will now silently ignore attempts to upload a file if a file with the same
+  hash was previously uploaded.
+issue-nr: 7531
+change-type: minor
+destination-branches: [master, iso7]
+sections:
+  minor-improvement: "{{description}}"
```
arnaudsjs, line 8: To me this change is just a bugfix. I don't really see why it needs a minor-improvement entry.

changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
```diff
@@ -0,0 +1,10 @@
+description: >
+  The upload_file endpoint will now silently ignore attempts to upload a file if a file with the same
+  hash was previously uploaded.
+issue-nr: 7531
+change-type: minor
+destination-branches: [master, iso7]
```
arnaudsjs, line 6: Is there a specific reason why we don't want this fix into iso6?

arnaudsjs (changes requested):

changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
Hugo-Inmanta: Hmm no actually. It felt more like a minor than a patch to me, but looking back again at it, this shouldn't drive to which branch it gets merged. I'll add iso6 and I'll try to be more careful about this in the future.

arnaudsjs (approved):

changelogs/unreleased/7531-race-condition-in-exporter-file-upload.yml
```diff
@@ -0,0 +1,10 @@
+description: >
+  The upload_file endpoint will now silently ignore attempts to upload a file if a file with the same
+  hash was previously uploaded.
+issue-nr: 7531
+change-type: patch
+destination-branches: [master, iso7, iso6]
+sections:
+  bugfix: >
+    Fix race condition where exporting a file might fail if a file with similar content
```
sanderr, line 9:
```suggestion
    Fix race condition where exporting a file might fail if a file the same content
```

src/inmanta/protocol/methods.py
```diff
@@ -876,7 +876,7 @@ def upload_code_batched(tid: uuid.UUID, id: int, resources: dict):
     :param tid: The id of the environment to which the code belongs.
     :param id: The version number of the configuration model.
     :param resources: A dictionary where each key is a string representing a resource type.
-                  For each resource type, the value is a dictionary. This nested dictionary's keys are file names,
+                  For each resource type, the value is a dictionary. This nested dictionary's keys are file hashes,
```
sanderr, line 879: I think they're technically both, but perhaps this is an improvement.

sanderr (changes requested):

src/inmanta/protocol/methods.py
Hugo-Inmanta: To me `names` implies a clear text file name

sanderr: Well, it is. Iirc this is the name we use to store this on the file system. But perhaps I don't remember correctly.

sanderr (approved):

src/inmanta/protocol/methods.py
Hugo-Inmanta: Yes I think these keys are hashes: we use the hash to call into upload_code_batched:  https://github.com/inmanta/inmanta-core/blob/a6b57a5ba24d9bccbee9884482de0510a5ab153e/src/inmanta/export.py#L107-L112

PR #7717, opened 2024-06-10
Issue/736 also enable batched partial compiles for termination states

Hugo-Inmanta:
# Description

Core part of https://github.com/inmanta/lsm/issues/736

# Self Check:

- [x] Changelog entry
   PR adding test in lsm module: https://github.com/inmanta/lsm/pull/752

### DIFF
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -451,9 +451,10 @@ def export_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
     parser.add_argument(
         "--delete-resource-set",
         dest="delete_resource_set",
-        help="Remove a resource set as part of a partial compile. This option can be provided multiple times and should always "
-        "be used together with the --partial option.",
-        action="append",
+        help="Remove resource set(s) passed in the ENV_REMOVED_SET_ID env variable as part of a partial compile. "
+        "This option should always be used together with the --partial option",
+        action="store_true",
+        default=False,
     )
     parser.add_argument(
         "--soft-delete",
@@ -473,6 +474,15 @@ def export(options: argparse.Namespace) -> None:
             raise CLIException(
                 "The --delete-resource-set option should always be used together with the --partial option", exitcode=1
             )
+        resource_sets_to_remove: list[str] = []
+        if options.delete_resource_set is True:
+            if "ENV_REMOVED_SET_ID" not in os.environ or not (removed_sets := os.environ["ENV_REMOVED_SET_ID"]):
+                raise CLIException(
+                    "The --delete-resource-set cli option is set but no resource set was passed via "
+                    "the ENV_REMOVED_SET_ID env variable."
+                )
+            resource_sets_to_remove = removed_sets.split(" ")
+
         if options.environment is not None:
             Config.set("config", "environment", options.environment)
 
@@ -551,7 +561,7 @@ def export(options: argparse.Namespace) -> None:
                 model_export=options.model_export,
                 export_plugin=options.export_plugin,
                 partial_compile=options.partial_compile,
-                resource_sets_to_remove=options.delete_resource_set,
+                resource_sets_to_remove=resource_sets_to_remove,
             )
 
         if not summary_reporter.is_failure() and options.deploy:
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -442,8 +442,8 @@ def export_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
         help=(
             "Execute a partial export. Does not upload new Python code to the server: it is assumed to be unchanged since the"
             " last full export. Multiple partial exports for disjunct resource sets may be performed concurrently but not"
-            " concurrent with a full export. When used in combination with the `--json` option, 0 is used as a placeholder for"
-            " the model version."
+            " concurrent with a full export. When used in combination with the ``--json`` option, 0 is used as a placeholder "
+            "for the model version."
         ),
         action="store_true",
         default=False,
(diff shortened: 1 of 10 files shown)


src/inmanta/app.py
```diff
@@ -451,9 +451,10 @@ def export_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
     parser.add_argument(
         "--delete-resource-set",
         dest="delete_resource_set",
-        help="Remove a resource set as part of a partial compile. This option can be provided multiple times and should always "
-        "be used together with the --partial option.",
-        action="append",
+        help="Remove resource set(s) passed in the INMANTA_REMOVED_SET_ID env variable as part of a partial compile. "
+        "This option should always be used together with the --partial option",
+        action="store_true",
```
wouterdb, line 456: I think 
1. we should always handle the env var if it is set
2. I think we can allow both the option and the env var to be set at the same time, this makes the backward compatibility trivial.

src/inmanta/server/services/compilerservice.py
```diff
@@ -645,10 +644,10 @@ async def request_recompile(
         :param in_db_transaction: If set to True, the connection must be provided and the connection must be part of an ongoing
                                   database transaction. If this parameter is set to True, is required to call
                                   `CompileService.notify_compile_request_committed()` right after the transaction commits.
-        :param soft_delete: Silently ignore deletion of resource sets in removed_resource_sets if they contain
-            resources that are being exported.
-        :param mergeable_env_vars: a set of env vars that can be compacted over multiple compiles.
-            If multiple values are compacted, they will be joined using spaces
+        :param soft_delete: Silently ignore deletion of resource sets passed in the env_vars or mergeable_env_vars if
+            they contain resources that are being exported.
+        :param mergeable_env_vars: a set of env vars that can be compacted over multiple compiles when batched partial compiles
```
wouterdb, line 649: At this level, we don't care about partial compile (i.e mergeable env vars can be used for other things as well)

src/inmanta/server/services/compilerservice.py
```diff
@@ -623,7 +623,6 @@ async def request_recompile(
         metadata: Optional[JsonType] = None,
         env_vars: Optional[Mapping[str, str]] = None,
         partial: bool = False,
-        removed_resource_sets: Optional[list[str]] = None,
```
wouterdb, line 626: I wonder if we should keep this in first instance to make the migration smoother (i.e. split this MR in two parts). 
(not a hard requirement, I just wonder)

tests/server/test_compilerservice.py
```diff
@@ -1052,12 +1092,31 @@ async def test_compileservice_queue(mocked_compiler_service_block: queue.Queue,
     remote_id6 = uuid.uuid4()
     compile_id6, _ = await compilerslice.request_recompile(env=env, force_update=False, do_export=True, remote_id=remote_id6)
 
+    # Queue = [
```
wouterdb, line 1095: nice!

wouterdb (changes requested):

src/inmanta/app.py
```diff
@@ -469,10 +471,21 @@ def export_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
 def export(options: argparse.Namespace) -> None:
     logger_mode_manager = LoggerModeManager.get_instance()
     with logger_mode_manager.run_in_logger_mode(LoggerMode.COMPILER):
-        if not options.partial_compile and options.delete_resource_set:
+        resource_sets_to_remove: set[str] = set(options.delete_resource_set) if options.delete_resource_set else set()
+
+        if "INMANTA_REMOVED_SET_ID" in os.environ:
```
wouterdb, line 476: This string should be in const with the other env vars

src/inmanta/const.py
```diff
@@ -340,3 +340,7 @@ class NotificationSeverity(str, Enum):
 DATETIME_MIN_UTC = datetime.datetime.min.replace(tzinfo=datetime.timezone.utc)
 
 MODULE_PKG_NAME_PREFIX = "inmanta-module-"
+
+# Resource sets marked for deletion during a partial export can be passed via this env
+# variable as a space separated list of set ids.
+INMANTA_REMOVED_SET_ID = "INMANTA_REMOVED_SET_ID"
```
sanderr, line 346:
```suggestion
INMANTA_REMOVED_SET_ID = "INMANTA_REMOVED_RESOURCE_SET_ID"
```

sanderr (approved):

sanderr (changes requested):
Wrong button

src/inmanta/server/services/compilerservice.py
Hugo-Inmanta: I don't have a good feel for how much pain this might cause, so I'll keep the parameter for now but simply ignore it.
Follow-up PR that removes it [here](https://github.com/inmanta/inmanta-core/pull/7730)

wouterdb (approved):

src/inmanta/server/services/compilerservice.py
```diff
@@ -648,6 +649,8 @@ async def request_recompile(
             they contain resources that are being exported.
         :param mergeable_env_vars: a set of env vars that can be compacted over multiple compiles.
             If multiple values are compacted, they will be joined using spaces.
+        :param removed_resource_sets: [DEPRECATED] Resource sets marked for removal should be passed by setting the
+            INMANTA_REMOVED_RESOURCE_SET_ID environment variable in the env_vars or the mergeable_env_vars parameter.
```
sanderr, line 653: I think we can afford to simply drop this. Afaik the only interface here is between core and inmanta-lsm, which we will upgrade together. @wouterdb do you agree?

sanderr (changes requested):

src/inmanta/server/services/compilerservice.py
wouterdb: yes

Issue #7722, opened 2024-06-11
agent executor code version caching


sanderr:
From [the design (WIP)](https://github.com/inmanta/designs/blob/c351ba26f7316c00b10e2d3e8f881c58dd37d8f0/agent-refactor.md)

"With incremental deploys (and even more so with more fine-grained deploys and repairs from chunk 2), we may often get only a partial view on an agent's code files / dependencies. We have to make sure that this does not lead to an increase in the number of code versions, and therefore the number of venvs and executors. We should either make sure to get a full view of an agent's code version for a given model version, or we should allow supersets of a code version. The first seems like the safer choice. The initial implementation of the external executors may or may not already satisfy this requirement. This requires further investigation."

PR #7745, opened 2024-06-18
Issue/7722 agent executor code version caching

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/7722

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -752,7 +752,7 @@ class AgentInstance:
                         result.result["version"],
                         const.ResourceAction.deploy,
                         result.result["resources"],
-                        result.result["resource_types"],
+                        set(result.result["resource_types"]),
                     )
                     if len(resources) > 0 and executor is not None:
                         self._nq.reload(resources, undeployable, deploy_request, executor)
@@ -774,7 +774,7 @@ class AgentInstance:
                     return
 
                 undeployable, resource_refs, executor = await self.setup_executor(
-                    version, const.ResourceAction.dryrun, response.result["resources"], response.result["resource_types"]
+                    version, const.ResourceAction.dryrun, response.result["resources"], set(response.result["resource_types"])
                 )
                 deployable_resources: list[ResourceDetails] = []
                 for resource in resource_refs:
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -797,7 +797,7 @@ class AgentInstance:
     async def get_facts(self, resource: JsonType) -> Apireturn:
         async with self.ratelimiter:
             undeployable, resource_refs, executor = await self.setup_executor(
-                resource["model"], const.ResourceAction.getfact, [resource], [resource["resource_type"]]
+                resource["model"], const.ResourceAction.getfact, [resource], {resource["resource_type"]}
             )
 
             if undeployable or not resource_refs:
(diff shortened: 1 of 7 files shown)


wouterdb (approved):
Nice!

src/inmanta/agent/agent.py
```diff
@@ -800,14 +810,17 @@ async def get_facts(self, resource: JsonType) -> Apireturn:
...
     ) -> tuple[dict[ResourceVersionIdStr, const.ResourceState], list[ResourceDetails], Optional[executor.Executor]]:
-        # TODO refactor wiring of tuples around -> not sure how
-
         """
         Setup an executor for resource deployment
 
-        returns a tuple
+        :param version: model version being deployed
+        :param action: the resource action associated with this deploy
+        :param resources: list of resources in serialized form: a dict with information about
```
sanderr, line 820: As far as I can tell we now no longer need `resources` to actually set up the executor. Perhaps we can extract the executor setup from this method? i.e. give this one a different name (because it does a lot more than set up the executor, and have it call into the new `setup_executor(version, resource_types) -> Optional[executor.Executor])`?

src/inmanta/agent/agent.py
```diff
@@ -800,14 +810,17 @@ async def get_facts(self, resource: JsonType) -> Apireturn:
...
+            this resource and its desired state
+        :param resource_types: set of all resource types composing this version of the model
```
sanderr, line 822:
```suggestion
        :param resource_types: set of all resource types composing this version of the model for this agent, even if they are not being deployed right now.
```
Tried to make this a bit more explicit, wdyt?

sanderr (changes requested):
Nice! Added some notes for further opportunities.

src/inmanta/agent/agent.py
```diff
@@ -843,11 +844,40 @@ async def setup_executor(
...
+    async def load_configuration_model(
+        self,
+        model_version: int,
+        action: const.ResourceAction,
+        resource_batch: list[JsonType],
+        resource_types: set[ResourceType],
+    ) -> tuple[dict[ResourceVersionIdStr, const.ResourceState], list[ResourceDetails], Optional[executor.Executor]]:
+        """
+        Load (part of) a configuration model. This method is expected to be called as an
+        initialization step when interacting with resources.
```
sanderr, line 858: We don't really load the model, do we? Just the resources? And we don't even really load them, just request an executor and "parse" them (for lack of a better word). I fail to come up with satisfying name though.Things that come to mind are `get_executable_resources`, `get_executor_resources`, `prepare_resources`, `setup_for_execution`.

sanderr (changes requested):
I hadn't really considered the naming and semantical difficulties my proposal would bring. If you don't feel like the change is an improvement, feel free to revert. I personally still think it is, but less than I assumed when I requested it.

src/inmanta/agent/agent.py
```diff
@@ -810,29 +816,24 @@ async def get_facts(self, resource: JsonType) -> Apireturn:
...
-
+        Set up an executor for a given version of a model. This executor is the interface
+        to interact with the resources.
+
+        :param model_version: model version being loaded
+        :param resource_types: set of all resource types we want to be able to interact with via this executor
+        :return: A tuple of:
+            - The executor (if creation/retrieval was successful)
+            - invalid_resource_types: resource types for which we're missing a handler or pip config
+            - failed_resource_types: resource types for which handler code installation failed
```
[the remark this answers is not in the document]
Hugo-Inmanta: I don't see a reason to differentiate anymore either

src/inmanta/agent/agent.py
Hugo-Inmanta: I think I like `prepare_resources` the most, since it will be used in `dryrun` and `getfacts` contexts as well. I always associate (maybe wrongfully so?) the `execute` semantic with the `deploy` itself

wouterdb (approved):

sanderr (approved):

PR #7746, opened 2024-06-19
Add ability to update worker name

wouterdb:
# Description

I added the ability to update the name of the process, to make debugging easier

1. It now identifies the agent, not the code version or anything
2. it shows the status (connecting/connected/disconnected) 
2. I could take it further and show every action (deploying resource x_y_z)
3. this requires an additional dependency, which I don't like, but I see no alternative

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/forking_agent/test_executor.py
+++ b/tests/forking_agent/test_executor.py
@@ -20,6 +20,7 @@ import asyncio
 import base64
 import logging
 
+import psutil
 import pytest
 
 import inmanta.agent
@@ -164,6 +165,9 @@ async def test_executor_server_dirty_shutdown(mpmanager: MPManager, caplog):
     result = await child1.connection.call(Echo(["aaaa"]))
     assert ["aaaa"] == result
     print("Child there")
+
+    process_name = psutil.Process(pid=child1.process.pid).name()
+    assert process_name == "inmanta: executor test - connected"
 
     await asyncio.get_running_loop().run_in_executor(None, child1.process.kill)
     print("Kill sent")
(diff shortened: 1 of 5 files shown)


tests/forking_agent/test_executor.py
```diff
@@ -165,6 +166,9 @@ async def test_executor_server_dirty_shutdown(mpmanager: MPManager, caplog):
     assert ["aaaa"] == result
     print("Child there")
 
+    process_name = psutil.Process(pid=child1.process.pid).name()
+    assert process_name == "inmanta: executor test - connected"
```
arnaudsjs, line 170: I wondering whether we should include the full identity of an executor here, because multiple agents with the same name can exist at the same time. For example when a dry-run is requested for an unreleased version. With this implementation it will be hard to keep them apart.

arnaudsjs (approved):

tests/forking_agent/test_executor.py
wouterdb: I agree (see description), I just don't see how to make this identity meaningful...

sanderr (approved):
Nice!

tests/forking_agent/test_executor.py
wouterdb: it is volatile and not part of the current interfaces, but I kind of like it.

The ranges may compact quite badly, because not every version may reach the executor so it will not be aware of everything. 

I don't think I'm going to add it now, but it is something to keep in mind

arnaudsjs: I like it as well.

Issue #7693, opened 2024-05-31
Vevn Cleanup


wouterdb:
Implement and document a policy for cleanup of virtual envs created for the executors

# Proposal

1. every thing the env is used, we touch a specific file (and any running executor regularly touches this file for his venv)
2. when this file has not been modified for X days (configurable, default 7), the venv is removed

PR #7748, opened 2024-06-19
Venv cleanup

hlloreda:
# Description

- Venv can now be cleaned

closes https://github.com/inmanta/inmanta-core/issues/7693

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/config.py
+++ b/src/inmanta/agent/config.py
@@ -160,6 +160,16 @@ agent_get_resource_backoff: Option[float] = Option(
     is_float,
 )
 
+agent_virtual_environment_cleanup: Option[int] = Option(
+    "config",
+    "agent-venv-cleanup",
+    7,
+    "This is the number of days to wait before unused Python virtual environments are cleaned from the inmanta "
+    "server. Setting this option too low may result in a high load on the Inmanta server. Setting it too high"
+    " may result in long deployment times.",
+    is_int,
+)
+
 
 class AgentExcutorMode(str, enum.Enum):
     threaded = "threaded"
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -404,8 +404,9 @@ class VirtualEnvironmentManager:
             if not file.is_dir():
                 continue
 
-            if file in reverse_environment_map:
-                blueprint = reverse_environment_map[file]
+            absolute_path_file = str(file.absolute())
+            if absolute_path_file in reverse_environment_map:
+                blueprint = reverse_environment_map[absolute_path_file]
                 async with self._locks.get(blueprint.blueprint_hash()):
                     del self._environment_map[blueprint]
 
@@ -417,10 +418,10 @@ class VirtualEnvironmentManager:
 
                 modification_datetime = datetime.datetime.fromtimestamp(timestamp)
                 if (current_datetime - modification_datetime).days >= number_days_before_venv_cleanup:
-                    environments_on_disk_to_clean.add(str(file.absolute()))
+                    environments_on_disk_to_clean.add(absolute_path_file)
             else:
                 # The Virtual Environment could be incomplete or broken
-                environments_on_disk_to_clean.add(str(file.absolute()))
+                environments_on_disk_to_clean.add(absolute_path_file)
 
         for path_env_to_clean in environments_on_disk_to_clean:
             shutil.rmtree(path_env_to_clean)
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -39,7 +39,7 @@ import pkg_resources
 
 import inmanta.types
 from inmanta.agent import config as cfg
-from inmanta.data.model import PipConfig, ResourceIdStr, ResourceVersionIdStr
+from inmanta.data.model import PipConfig, ResourceIdStr, ResourceVersionIdStr, VirtualEnvStatus
 from inmanta.env import PythonEnvironment
 from inmanta.loader import ModuleSource
 from inmanta.resources import Id
@@ -315,10 +315,18 @@ class VirtualEnvironmentManager:
             return env_dir, False  # Returning the path and False for existing directory
 
     def read_environment_status(self, inmanta_env_status_file: str) -> Optional[str]:
-        current_status = None
-        if os.path.exists(inmanta_env_status_file):
-            with open(inmanta_env_status_file) as f:
-                current_status = f.read().strip()
+        """
+        Read the specific inmanta file which gives an indication on the status of the Python Virtual Environment:
+            - Is it broken / empty (no specific file)
+            - Is it created / running (should be complete and good to be used with an executor)
+
+        :param inmanta_env_status_file: The inmanta file that gives us the status of the Python Virtual Environment
+        """
+        if not os.path.exists(inmanta_env_status_file):
+            return None
+
+        with open(inmanta_env_status_file) as f:
+            current_status = f.read().strip()
         return current_status
 
     async def create_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPoolExecutor) -> ExecutorVirtualEnvironment:
@@ -339,7 +347,7 @@ class VirtualEnvironmentManager:
             current_status = self.read_environment_status(inmanta_env_status)
 
             match current_status:
-                case "built" | "running":
+                case VirtualEnvStatus.created | VirtualEnvStatus.running:
                     pass
                 case _:
                     shutil.rmtree(env_storage)
@@ -352,7 +360,7 @@ class VirtualEnvironmentManager:
         self._environment_map[blueprint] = process_environment
 
         with open(inmanta_env_status, "w") as f:
-            f.write("built")
+            f.write(VirtualEnvStatus.created)
 
         return process_environment
 
@@ -388,13 +396,18 @@ class VirtualEnvironmentManager:
         number_days_before_venv_cleanup = cfg.agent_virtual_environment_cleanup.get()
         envs_dir = pathlib.Path(self.envs_dir)
         environments_on_disk_to_clean: set[str] = set()
-        environments_in_memory_to_clean_mapping: dict[str, EnvBlueprint] = dict()
+        reverse_environment_map = {v.env_path: k for k, v in self._environment_map.items()}
 
         # We check every folder and we save the ones that contain the special file and that should be removed
         current_datetime = datetime.datetime.now()
         for file in envs_dir.iterdir():
             if not file.is_dir():
                 continue
+
+            if file in reverse_environment_map:
+                blueprint = reverse_environment_map[file]
+                async with self._locks.get(blueprint.blueprint_hash()):
+                    del self._environment_map[blueprint]
 
             inmanta_env_status_file = file / ".inmanta_env_status"
 
@@ -409,20 +422,7 @@ class VirtualEnvironmentManager:
                 # The Virtual Environment could be incomplete or broken
                 environments_on_disk_to_clean.add(str(file.absolute()))
 
-        for blueprint, env in self._environment_map.items():
-            env_path = env.env_path
-            if env_path not in environments_on_disk_to_clean:
-                continue
-
-            # We need to save this information to remove the entry from `_environment_map`
-            environments_in_memory_to_clean_mapping[env_path] = blueprint
-
-        for path_env_to_clean, blueprint in environments_in_memory_to_clean_mapping.items():
-            async with self._locks.get(blueprint.blueprint_hash()):
-                del self._environment_map[blueprint]
-                shutil.rmtree(path_env_to_clean)
-
-        for path_env_to_clean in environments_on_disk_to_clean.difference(set(environments_in_memory_to_clean_mapping.keys())):
+        for path_env_to_clean in environments_on_disk_to_clean:
             shutil.rmtree(path_env_to_clean)
 
 
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -383,44 +383,47 @@ class VirtualEnvironmentManager:
 
     async def clean_environments(self) -> None:
         """
-        Retrieves an existing virtual environment that matches the given blueprint or creates a new one if no match is found.
-        Utilizes NamedLock to ensure thread-safe operations for each unique blueprint.
+        Remove Python Virtual Environments that were not used since a number of days (configurable in the agent config).
         """
-
-        env_folder = pathlib.Path(self.envs_dir)
-        environments_to_clean: set[pathlib.Path] = set()
-        environments_to_clean_mapping: dict[str, EnvBlueprint] = dict()
         number_days_before_venv_cleanup = cfg.agent_virtual_environment_cleanup.get()
-        for file in env_folder.iterdir():
+        envs_dir = pathlib.Path(self.envs_dir)
+        environments_on_disk_to_clean: set[str] = set()
+        environments_in_memory_to_clean_mapping: dict[str, EnvBlueprint] = dict()
+
+        # We check every folder and we save the ones that contain the special file and that should be removed
+        current_datetime = datetime.datetime.now()
+        for file in envs_dir.iterdir():
             if not file.is_dir():
                 continue
 
             inmanta_env_status_file = file / ".inmanta_env_status"
 
-            # get modification time
-            # platform dependent (time of most recent metadata change on Unix, or the time of creation on Windows)
-            # Could cause issue one Windows systems
-            timestamp = inmanta_env_status_file.stat().st_mtime
+            if inmanta_env_status_file.exists():
+                # Retrieve modification time
+                timestamp = inmanta_env_status_file.stat().st_mtime
 
-            # convert time to dd-mm-yyyy hh:mm:ss
-            modification_datetime = datetime.datetime.fromtimestamp(timestamp)
-            if (datetime.datetime.now() - modification_datetime).days >= number_days_before_venv_cleanup:
-                environments_to_clean.add(file)
+                modification_datetime = datetime.datetime.fromtimestamp(timestamp)
+                if (current_datetime - modification_datetime).days >= number_days_before_venv_cleanup:
+                    environments_on_disk_to_clean.add(str(file.absolute()))
+            else:
+                # The Virtual Environment could be incomplete or broken
+                environments_on_disk_to_clean.add(str(file.absolute()))
 
         for blueprint, env in self._environment_map.items():
             env_path = env.env_path
-            if env_path not in environments_to_clean:
+            if env_path not in environments_on_disk_to_clean:
                 continue
 
-            environments_to_clean_mapping[env_path] = blueprint
+            # We need to save this information to remove the entry from `_environment_map`
+            environments_in_memory_to_clean_mapping[env_path] = blueprint
 
-        for env_to_clean, blueprint in environments_to_clean_mapping.items():
+        for path_env_to_clean, blueprint in environments_in_memory_to_clean_mapping.items():
             async with self._locks.get(blueprint.blueprint_hash()):
-                shutil.rmtree(env_to_clean)
                 del self._environment_map[blueprint]
+                shutil.rmtree(path_env_to_clean)
 
-        for env_to_clean in environments_to_clean:
-            shutil.rmtree(env_to_clean)
+        for path_env_to_clean in environments_on_disk_to_clean.difference(set(environments_in_memory_to_clean_mapping.keys())):
+            shutil.rmtree(path_env_to_clean)
 
 
 class CacheVersionContext(contextlib.AbstractAsyncContextManager[None]):
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -20,11 +20,14 @@ import abc
 import asyncio
 import contextlib
 import dataclasses
+import datetime
 import functools
 import hashlib
 import json
 import logging
 import os
+import pathlib
+import shutil
 import types
 import typing
 import uuid
@@ -311,6 +314,13 @@ class VirtualEnvironmentManager:
             )
             return env_dir, False  # Returning the path and False for existing directory
 
+    def read_environment_status(self, inmanta_env_status_file: str) -> Optional[str]:
+        current_status = None
+        if os.path.exists(inmanta_env_status_file):
+            with open(inmanta_env_status_file) as f:
+                current_status = f.read().strip()
+        return current_status
+
     async def create_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPoolExecutor) -> ExecutorVirtualEnvironment:
         """
         Creates a new virtual environment based on the provided blueprint or reuses an existing one if suitable.
@@ -320,14 +330,29 @@ class VirtualEnvironmentManager:
         :param threadpool: A ThreadPoolExecutor
         :return: An instance of ExecutorVirtualEnvironment representing the created or reused environment.
 
-        TODO: Improve handling of bad venv scenarios, such as when the folder exists but is empty or corrupted.
         """
         env_storage, is_new = self.get_or_create_env_directory(blueprint)
         process_environment = ExecutorVirtualEnvironment(env_storage, threadpool)
+        inmanta_env_status = os.path.join(env_storage, ".inmanta_env_status")
+
+        if not is_new:
+            current_status = self.read_environment_status(inmanta_env_status)
+
+            match current_status:
+                case "built" | "running":
+                    pass
+                case _:
+                    shutil.rmtree(env_storage)
+                    os.makedirs(env_storage)
+                    is_new = True
+
         if is_new:
             LOGGER.info("Creating venv for content %s, content hash: %s", str(blueprint), blueprint.blueprint_hash())
             await process_environment.create_and_install_environment(blueprint)
         self._environment_map[blueprint] = process_environment
+
+        with open(inmanta_env_status, "w") as f:
+            f.write("built")
 
         return process_environment
 
@@ -355,6 +380,47 @@ class VirtualEnvironmentManager:
                 )
                 return self._environment_map[blueprint]
             return await self.create_environment(blueprint, threadpool)
+
+    async def clean_environments(self) -> None:
+        """
+        Retrieves an existing virtual environment that matches the given blueprint or creates a new one if no match is found.
+        Utilizes NamedLock to ensure thread-safe operations for each unique blueprint.
+        """
+
+        env_folder = pathlib.Path(self.envs_dir)
+        environments_to_clean: set[pathlib.Path] = set()
+        environments_to_clean_mapping: dict[str, EnvBlueprint] = dict()
+        number_days_before_venv_cleanup = cfg.agent_virtual_environment_cleanup.get()
+        for file in env_folder.iterdir():
+            if not file.is_dir():
+                continue
+
+            inmanta_env_status_file = file / ".inmanta_env_status"
+
+            # get modification time
+            # platform dependent (time of most recent metadata change on Unix, or the time of creation on Windows)
+            # Could cause issue one Windows systems
+            timestamp = inmanta_env_status_file.stat().st_mtime
+
+            # convert time to dd-mm-yyyy hh:mm:ss
+            modification_datetime = datetime.datetime.fromtimestamp(timestamp)
+            if (datetime.datetime.now() - modification_datetime).days >= number_days_before_venv_cleanup:
+                environments_to_clean.add(file)
+
+        for blueprint, env in self._environment_map.items():
+            env_path = env.env_path
+            if env_path not in environments_to_clean:
+                continue
+
+            environments_to_clean_mapping[env_path] = blueprint
+
+        for env_to_clean, blueprint in environments_to_clean_mapping.items():
+            async with self._locks.get(blueprint.blueprint_hash()):
+                shutil.rmtree(env_to_clean)
+                del self._environment_map[blueprint]
+
+        for env_to_clean in environments_to_clean:
+            shutil.rmtree(env_to_clean)
 
 
 class CacheVersionContext(contextlib.AbstractAsyncContextManager[None]):
--- a/src/inmanta/agent/forking_executor.py
+++ b/src/inmanta/agent/forking_executor.py
@@ -442,7 +442,7 @@ class MPExecutor(executor.Executor):
         await self.connection.call(ExecuteCommand(gid, resource_details, reason))
         self.check_modification_time_venv()
 
-    def check_modification_time_venv(self):
+    def check_modification_time_venv(self) -> None:
         """
         Make sure that the time of most recent content modification of the `inmanta_env_status` file is within the last two
         hours. If it's not, the file is overwritten.
--- a/src/inmanta/agent/forking_executor.py
+++ b/src/inmanta/agent/forking_executor.py
@@ -20,6 +20,7 @@ import asyncio
 import collections
 import concurrent.futures
 import concurrent.futures.thread
+import datetime
 import functools
 import logging
 import logging.config
@@ -417,9 +418,11 @@ class MPExecutor(executor.Executor):
 
     async def close_version(self, version: int) -> None:
         await self.connection.call(CloseVersionCommand(version))
+        self.check_env_status()
 
     async def open_version(self, version: int) -> None:
         await self.connection.call(OpenVersionCommand(version))
+        self.check_env_status()
 
     async def dry_run(
         self,
@@ -427,6 +430,7 @@ class MPExecutor(executor.Executor):
         dry_run_id: uuid.UUID,
     ) -> None:
         await self.connection.call(DryRunCommand(resources, dry_run_id))
+        self.check_env_status()
 
     async def execute(
         self,
@@ -435,9 +439,24 @@ class MPExecutor(executor.Executor):
         reason: str,
     ) -> None:
         await self.connection.call(ExecuteCommand(gid, resource_details, reason))
+        self.check_env_status()
+
+    def check_env_status(self):
+        inmanta_env_status = os.path.join(self.executor_virtual_env.env_path, ".inmanta_env_status")
+        timestamp_env_modification = os.stat(inmanta_env_status).st_mtime
+        last_access_env_status = datetime.datetime.fromtimestamp(timestamp_env_modification)
+
+        current_datetime = datetime.datetime.now()
+        if ((current_datetime - last_access_env_status).seconds / 3600) < 2:
+            return
+
+        with open(inmanta_env_status, "w") as f:
+            f.write("running")
 
     async def get_facts(self, resource: "inmanta.agent.executor.ResourceDetails") -> inmanta.types.Apireturn:
-        return await self.connection.call(FactsCommand(resource))
+        facts = await self.connection.call(FactsCommand(resource))
+        self.check_env_status()
+        return facts
 
 
 class MPManager(executor.ExecutorManager[MPExecutor]):
--- a/src/inmanta/agent/forking_executor.py
+++ b/src/inmanta/agent/forking_executor.py
@@ -44,6 +44,7 @@ import inmanta.protocol.ipc_light
 import inmanta.signals
 import inmanta.util
 from inmanta.agent import executor
+from inmanta.data.model import VirtualEnvStatus
 from inmanta.protocol.ipc_light import FinalizingIPCClient, IPCServer, LogReceiver, LogShipper
 
 LOGGER = logging.getLogger(__name__)
@@ -418,11 +419,11 @@ class MPExecutor(executor.Executor):
 
     async def close_version(self, version: int) -> None:
         await self.connection.call(CloseVersionCommand(version))
-        self.check_env_status()
+        self.check_modification_time_venv()
 
     async def open_version(self, version: int) -> None:
         await self.connection.call(OpenVersionCommand(version))
-        self.check_env_status()
+        self.check_modification_time_venv()
 
     async def dry_run(
         self,
@@ -430,7 +431,7 @@ class MPExecutor(executor.Executor):
         dry_run_id: uuid.UUID,
     ) -> None:
         await self.connection.call(DryRunCommand(resources, dry_run_id))
-        self.check_env_status()
+        self.check_modification_time_venv()
 
     async def execute(
         self,
@@ -439,9 +440,13 @@ class MPExecutor(executor.Executor):
         reason: str,
     ) -> None:
         await self.connection.call(ExecuteCommand(gid, resource_details, reason))
-        self.check_env_status()
+        self.check_modification_time_venv()
 
-    def check_env_status(self):
+    def check_modification_time_venv(self):
+        """
+        Make sure that the time of most recent content modification of the `inmanta_env_status` file is within the last two
+        hours. If it's not, the file is overwritten.
+        """
         inmanta_env_status = os.path.join(self.executor_virtual_env.env_path, ".inmanta_env_status")
         timestamp_env_modification = os.stat(inmanta_env_status).st_mtime
         last_access_env_status = datetime.datetime.fromtimestamp(timestamp_env_modification)
@@ -451,11 +456,11 @@ class MPExecutor(executor.Executor):
             return
 
         with open(inmanta_env_status, "w") as f:
-            f.write("running")
+            f.write(VirtualEnvStatus.running)
 
     async def get_facts(self, resource: "inmanta.agent.executor.ResourceDetails") -> inmanta.types.Apireturn:
         facts = await self.connection.call(FactsCommand(resource))
-        self.check_env_status()
+        self.check_modification_time_venv()
         return facts
 
 
--- a/tests/forking_agent/test_agent_executor.py
+++ b/tests/forking_agent/test_agent_executor.py
@@ -346,54 +346,28 @@ def test():
     blueprint1 = executor.ExecutorBlueprint(pip_config=pip_config, requirements=requirements1, sources=sources1)
     blueprint2 = executor.ExecutorBlueprint(pip_config=pip_config, requirements=requirements2, sources=sources2)
 
-    # class TestDatetime(datetime.datetime):
-    #     return_to_the_past = datetime.datetime(2015, 12, 25, 17, 5, 55)
-    #
-    #     @classmethod
-    #     def now(cls, tz=None):
-    #         return cls.return_to_the_past
-    #
-    # monkeypatch.setattr(datetime, "datetime", TestDatetime)
-
-
     executor_manager = mpmanager_light
     executor_1, executor_2 = await asyncio.gather(
         executor_manager.get_executor("agent1", "local:", code_for(blueprint1)),
         executor_manager.get_executor("agent1", "local:", code_for(blueprint2)),
     )
 
-    old_check_executor1 = executor_1.last_access_env_status
-    old_check_executor2 = executor_2.last_access_env_status
-
-    assert ((datetime.datetime.now() - old_check_executor1).seconds / 3600) < 2
-    assert ((datetime.datetime.now() - old_check_executor2).seconds / 3600) < 2
-
-    # subprocess.check_output(
-    #     ["touch", "-m", "-t", "202211161200.10", f"{executor_2.executor_virtual_env.env_path}/.inmanta_env_status"]
-    # )
-
     old_datetime = datetime.datetime(year=2022, month=9, day=22, hour=12, minute=51, second=42)
     os.utime(f"{executor_2.executor_virtual_env.env_path}/.inmanta_env_status", (old_datetime.timestamp(), old_datetime.timestamp()))
-    # assert executor_1.last_access_env_status == TestDatetime.return_to_the_past
-    # assert executor_2.last_access_env_status == TestDatetime.return_to_the_past
-    #
-    # monkeypatch.undo()
+
+    def get_modification_datetime(file: str) -> datetime.datetime:
+        return datetime.datetime.fromtimestamp(os.stat(file).st_mtime)
+
+    old_check_executor1 = get_modification_datetime(f"{executor_1.executor_virtual_env.env_path}/.inmanta_env_status")
 
     executor_1.check_env_status()
     executor_2.check_env_status()
 
-    # old_check_executor1 = executor_1.last_access_env_status
-    # old_check_executor2 = executor_2.last_access_env_status
+    new_check_executor1 = get_modification_datetime(f"{executor_1.executor_virtual_env.env_path}/.inmanta_env_status")
+    new_check_executor2 = get_modification_datetime(f"{executor_2.executor_virtual_env.env_path}/.inmanta_env_status")
 
-    # assert ((datetime.datetime.now() - old_check_executor1).seconds / 3600) < 2
-    # assert ((datetime.datetime.now() - old_check_executor2).seconds / 3600) < 2
-
-    # This should not override `last_access_env_status`
-    # executor_1.check_env_status()
-    # executor_2.check_env_status()
-
-    assert executor_1.last_access_env_status == old_check_executor1
-    assert ((datetime.datetime.now() - old_check_executor2).seconds / 3600) < 2
+    assert new_check_executor1 == old_check_executor1
+    assert ((datetime.datetime.now() - new_check_executor2).seconds / 3600) < 2
 
     # Now we want to check if the cleanup is working correctly
     # First we want to override the modification date of the `inmanta_env_status` file
--- a/tests/forking_agent/test_agent_executor.py
+++ b/tests/forking_agent/test_agent_executor.py
@@ -353,15 +353,17 @@ def test():
     )
 
     old_datetime = datetime.datetime(year=2022, month=9, day=22, hour=12, minute=51, second=42)
-    os.utime(f"{executor_2.executor_virtual_env.env_path}/.inmanta_env_status", (old_datetime.timestamp(), old_datetime.timestamp()))
+    os.utime(
+        f"{executor_2.executor_virtual_env.env_path}/.inmanta_env_status", (old_datetime.timestamp(), old_datetime.timestamp())
+    )
 
     def get_modification_datetime(file: str) -> datetime.datetime:
         return datetime.datetime.fromtimestamp(os.stat(file).st_mtime)
 
     old_check_executor1 = get_modification_datetime(f"{executor_1.executor_virtual_env.env_path}/.inmanta_env_status")
 
-    executor_1.check_env_status()
-    executor_2.check_env_status()
+    executor_1.check_modification_time_venv()
+    executor_2.check_modification_time_venv()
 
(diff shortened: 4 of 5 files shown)


src/inmanta/agent/config.py
```diff
@@ -160,6 +160,16 @@
     is_float,
 )
 
+agent_virtual_environment_cleanup: Option[int] = Option(
+    "config",
+    "agent-venv-cleanup",
+    7,
+    "This is the number of days to wait before unused Python virtual environments are cleaned from the inmanta "
+    "server. Setting this option too low may result in a high load on the Inmanta server. Setting it too high"
+    " may result in long deployment times.",
```
wouterdb, line 169:
```suggestion
    " may result in increased disk usage.",
```

src/inmanta/agent/config.py
```diff
@@ -160,6 +160,16 @@
...
+    is_int,
```
wouterdb, line 170: I wonder if we can make it float? (if you want to go really short, it is easy for us to allow and may come in handy)

src/inmanta/agent/executor.py
```diff
@@ -311,6 +314,21 @@ def get_or_create_env_directory(self, blueprint: EnvBlueprint) -> tuple[str, boo
             )
             return env_dir, False  # Returning the path and False for existing directory
 
+    def read_environment_status(self, inmanta_env_status_file: str) -> Optional[str]:
```
wouterdb, line 317: style wise, I would make this method take a folder as argument and have it return a VirtualEnvStatus

```suggestion
    def read_environment_status(self, venv_folder: str) -> Optional[VirtualEnvStatus]:
```

i.e. I would make this method encapsulate the manner in which this value is persisted. 

And I would make an associated write method

(you could even group these methods on a class, or add them to the `ExecutorVirtualEnvironment` but I'm less opinionated about that)

src/inmanta/agent/executor.py
```diff
@@ -320,15 +338,30 @@ async def create_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPo
...
         """
         env_storage, is_new = self.get_or_create_env_directory(blueprint)
         process_environment = ExecutorVirtualEnvironment(env_storage, threadpool)
+        inmanta_env_status = os.path.join(env_storage, ".inmanta_env_status")
+
+        if not is_new:
+            current_status = self.read_environment_status(inmanta_env_status)
+
+            match current_status:
+                case VirtualEnvStatus.created | VirtualEnvStatus.running:
```
wouterdb, line 350: Does the enum match on the string, I wonder?

src/inmanta/agent/executor.py
```diff
@@ -320,15 +338,30 @@ async def create_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPo
...
+                    pass
+                case _:
+                    shutil.rmtree(env_storage)
```
wouterdb, line 353: I would log something here, this is a pretty invasive and slow operation

wouterdb, line 353: We may want to run the removal on a threadpool, to not block the ioloop

src/inmanta/agent/executor.py
```diff
@@ -320,15 +338,30 @@ async def create_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPo
...
+                    os.makedirs(env_storage)
+                    is_new = True
+
         if is_new:
             LOGGER.info("Creating venv for content %s, content hash: %s", str(blueprint), blueprint.blueprint_hash())
             await process_environment.create_and_install_environment(blueprint)
```
wouterdb, line 359: Looking at this code again, I noticed that `self.init_env()` in `ExecutorVirtualEnvironment.create_and_install_environment` is not running on a thread either, I think this can block the ioloop as well and should be on a thread.

src/inmanta/agent/executor.py
```diff
@@ -356,6 +389,43 @@ async def get_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPoolE
...
+        current_datetime = datetime.datetime.now()
+        for file in envs_dir.iterdir():
+            if not file.is_dir():
+                continue
+
+            absolute_path_file = str(file.absolute())
+            if absolute_path_file in reverse_environment_map:
+                blueprint = reverse_environment_map[absolute_path_file]
+                async with self._locks.get(blueprint.blueprint_hash()):
+                    del self._environment_map[blueprint]
```
wouterdb, line 411: why?

src/inmanta/agent/forking_executor.py
```diff
@@ -435,9 +440,28 @@ async def execute(
...
     ) -> None:
         await self.connection.call(ExecuteCommand(gid, resource_details, reason))
+        self.check_modification_time_venv()
+
+    def check_modification_time_venv(self) -> None:
+        """
+        Make sure that the time of most recent content modification of the `inmanta_env_status` file is within the last two
+        hours. If it's not, the file is overwritten.
+        """
+        inmanta_env_status = os.path.join(self.executor_virtual_env.env_path, ".inmanta_env_status")
```
wouterdb, line 450: 1. I think it is simpler to just update it always  (e.g. using https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch)  (it makes it more accurate and requires less calls to the OS)

2. I think this is risky: if an executor is running in the venv, but not doing anything, the venv can still be removed. I think it is easier to just start a timer (Inside the executor, not on the agent side like you did now) that updates the file every minute (I propose one minute, because it is sufficiently fast for all cases and sufficiently slow to not matter at all wrt performance)

tests/forking_agent/test_agent_executor.py
```diff
@@ -312,3 +314,71 @@ def test():
...
+    # First we want to override the modification date of the `inmanta_env_status` file
+    os.utime(
+        f"{executor_1.executor_virtual_env.env_path}/.inmanta_env_status", (old_datetime.timestamp(), old_datetime.timestamp())
+    )
+
+    venv_dir = pathlib.Path(mpmanager_light.environment_manager.envs_dir)
+    assert len([e for e in venv_dir.iterdir()]) == 2, "We should have two Virtual Environments for our 2 executors!"
+    # We remove the old VirtualEnvironment
+    await mpmanager_light.environment_manager.clean_environments()
+    assert len([e for e in venv_dir.iterdir()]) == 1, "Only one Virtual Environment should exist!"
```
wouterdb, line 384: it shouldn't be removed, because the executor is still running,... 
(i.e. this testcase test a scenario that should never happen)

wouterdb (changes requested):
Nice!

src/inmanta/agent/forking_executor.py
```diff
@@ -175,13 +194,19 @@ def __init__(
         sources: list[inmanta.loader.ModuleSource],
         environment: uuid.UUID,
         uri: str,
+        venv_checkup_interval: typing.Optional[int] = None,
```
hlloreda, line 197: I don't know if you see another way to change the interval only in the test

src/inmanta/agent/executor.py
```diff
@@ -341,26 +325,41 @@ async def create_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPo
...
-                case _:
-                    shutil.rmtree(env_storage)
-                    os.makedirs(env_storage)
-                    is_new = True
+        loop = asyncio.get_running_loop()
+
+        if not is_new:
+            if not inmanta_env_status.exists():
+                LOGGER.info(
+                    "Removing venv for content %s, content hash: %s located in",
```
wouterdb, line 339:
```suggestion
                    "Removing venv for content %s, content hash: %s located in %s",
```

src/inmanta/agent/executor.py
```diff
@@ -341,26 +325,41 @@ async def create_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPo
...
+                    env_storage,
+                )
+                await loop.run_in_executor(threadpool, reset_virtual_environment, env_storage)
+                is_new = True
 
         if is_new:
             LOGGER.info("Creating venv for content %s, content hash: %s", str(blueprint), blueprint.blueprint_hash())
-            await process_environment.create_and_install_environment(blueprint)
+
+            def async_func_wrapper(current_blueprint: EnvBlueprint) -> None:
```
wouterdb, line 350: I would not do this, I would either move the threadpool deeper down or make this method sync, but this is a bit too complex for comfort

src/inmanta/agent/executor.py
```diff
@@ -423,8 +417,16 @@ async def clean_environments(self) -> None:
...
         for path_env_to_clean in environments_on_disk_to_clean:
-            shutil.rmtree(path_env_to_clean)
+            # If the path to clean is part of the `_environment_map` then we can use the threadpool of the executor to clean
+            # the virtual environment
+            if path_env_to_clean in reverse_environment_map:
+                blueprint = reverse_environment_map[path_env_to_clean]
+                async with self._locks.get(blueprint.blueprint_hash()):
+                    await loop.run_in_executor(self._environment_map[blueprint].thread_pool, shutil.rmtree, path_env_to_clean)
+            else:
+                shutil.rmtree(path_env_to_clean)
```
wouterdb, line 429: The more import question is, why is it OK to block the IOloop if it isn't?

and two what makes this race-free with creation?

src/inmanta/agent/forking_executor.py
wouterdb: I think this is fine

src/inmanta/agent/forking_executor.py
```diff
@@ -224,6 +246,22 @@ async def call(self, context: ExecutorContext) -> typing.Sequence[inmanta.loader
                 logger.info("Failed to load sources: %s", module_source, exc_info=True)
                 failed.append(module_source)
 
+        if context.server.timer_venv_checkup is not None:
```
wouterdb, line 249: init should not be called twice, so I don't know if this is required 

(we could make it fail if called twice, might be safer)

tests/forking_agent/test_agent_executor.py
```diff
@@ -361,17 +362,21 @@ def get_modification_datetime(file: str) -> datetime.datetime:
         return datetime.datetime.fromtimestamp(os.stat(file).st_mtime)
 
     old_check_executor1 = get_modification_datetime(f"{executor_1.executor_virtual_env.env_path}/.inmanta_env_status")
+    old_check_executor2 = get_modification_datetime(f"{executor_2.executor_virtual_env.env_path}/.inmanta_env_status")
 
-    executor_1.check_modification_time_venv()
-    executor_2.check_modification_time_venv()
+    time.sleep(2)
```
wouterdb, line 367: 1. async sleep is more appropriate 
2. we can probably take even shorter timings?

src/inmanta/agent/forking_executor.py
```diff
@@ -224,6 +246,22 @@ async def call(self, context: ExecutorContext) -> typing.Sequence[inmanta.loader
...
+        if self._venv_checkup_interval is not None:
+            context.server.timer_venv_checkup = tornado.ioloop.PeriodicCallback(
+                callback=context.server.touch_inmanta_env_status,
+                callback_time=datetime.timedelta(seconds=self._venv_checkup_interval),
+            )
+        else:
+            context.server.timer_venv_checkup = tornado.ioloop.PeriodicCallback(
+                callback=context.server.touch_inmanta_env_status,
+                callback_time=datetime.timedelta(minutes=1),
+            )
```
wouterdb, line 262:
```suggestion
        interval = self._venv_checkup_interval or 60
        
        context.server.timer_venv_checkup = tornado.ioloop.PeriodicCallback(
            callback=context.server.touch_inmanta_env_status,
            callback_time=datetime.timedelta(seconds=interval),
        )
      
```

wouterdb (changes requested):

src/inmanta/agent/forking_executor.py
wouterdb: I would also start the timer before initializing the venv, take decrease time for a race

src/inmanta/agent/executor.py
```diff
@@ -356,6 +388,46 @@ async def get_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPoolE
...
+        reverse_environment_map = {v.env_path: k for k, v in self._environment_map.items()}
+
+        # We check every folder and we save the ones that contain the special file and that should be removed
+        current_datetime = datetime.datetime.now()
+        for file in envs_dir.iterdir():
+            if not file.is_dir():
+                continue
+
+            absolute_path_file = str(file.absolute())
+            inmanta_env_status_file = file / ".inmanta_env_status"
```
wouterdb, line 407: I would put this file name in a constant

src/inmanta/agent/executor.py
hlloreda: About 2. Isn't the lock enough? We rely on it when we create the environment and if the environment exists in memory in the cleanup, we use the same lock. Or am I missing something?

src/inmanta/agent/forking_executor.py
hlloreda: > I would also start the timer before initializing the venv, take decrease time for a race

I'm not sure to see why this would decrease time for a race

src/inmanta/agent/executor.py
hlloreda: Yes, it does

src/inmanta/agent/forking_executor.py
hlloreda: Going back to the second point: but now we could have a race condition between the touch of the file and the removal of the old virtual envs right?

src/inmanta/const.py
```diff
@@ -344,3 +344,6 @@ class NotificationSeverity(str, Enum):
 # Resource sets marked for deletion during a partial export can be passed via this env
 # variable as a space separated list of set ids.
 INMANTA_REMOVED_SET_ID = "INMANTA_REMOVED_RESOURCE_SET_ID"
+
+
+INMANTA_ENV_STATUS_FILENAME = "inmanta_env_status"
```
wouterdb, line 349: would document what this is for.

src/inmanta/agent/executor.py
```diff
@@ -397,14 +394,14 @@ async def clean_environments(self) -> None:
...
-        # We check every folder and we save the ones that contain the special file and that should be removed
+        # We check every folder, and we save the ones that contain the special file and that should be removed
         current_datetime = datetime.datetime.now()
         for file in envs_dir.iterdir():
             if not file.is_dir():
                 continue
 
             absolute_path_file = str(file.absolute())
-            inmanta_env_status_file = file / ".inmanta_env_status"
+            inmanta_env_status_file = file / f".{const.INMANTA_ENV_STATUS_FILENAME}"
```
wouterdb, line 404: the `.` has to be part for the constant as well (as it is part of the file name)

src/inmanta/agent/forking_executor.py
wouterdb: As it is now, we can only race when the executor is starting.

src/inmanta/agent/executor.py
wouterdb: this code is not below the lock

src/inmanta/agent/forking_executor.py
wouterdb: 1. When the executor is starting, up until the point it touches the file, it is racing with the cleanup.
    - I might be wrong, perhaps the entire startup is under lock, keeping the cleanup out? If so, it might be good to document this
    -  Doing it under lock is probably for the best
2. Loading the venv may take a bit of time
3. so writing to the file earlier is better as it reduces the race time

tests/forking_agent/test_agent_executor.py
```diff
@@ -312,3 +315,84 @@ def test():
     assert executor_1 is not executor_2, "Expected a different executor instance for different sources"
     assert executor_1 is not executor_3, "Expected a different executor instance for different requirements and sources"
     assert executor_2 is not executor_3, "Expected different executor instances for different requirements"
+
+
+async def test_executor_creation_and_venv_usage(pip_index, mpmanager_light, monkeypatch) -> None:
```
arnaudsjs, line 320: Type annotations are missing.

tests/forking_agent/test_agent_executor.py
```diff
@@ -312,3 +315,84 @@ def test():
...
+    sources1 = ()
+    sources2 = (module_source1,)
+
+    blueprint1 = executor.ExecutorBlueprint(pip_config=pip_config, requirements=requirements1, sources=sources1)
+    blueprint2 = executor.ExecutorBlueprint(pip_config=pip_config, requirements=requirements2, sources=sources2)
+
+    executor_manager = mpmanager_light
+    executor_1, executor_2 = await asyncio.gather(
+        executor_manager.get_executor("agent1", "local:", code_for(blueprint1), 0.1),
+        executor_manager.get_executor("agent2", "local:", code_for(blueprint2), 0.1),
```
arnaudsjs, line 353:
```suggestion
        executor_manager.get_executor("agent1", "local:", code_for(blueprint1), venv_checkup_interval=0.1),
        executor_manager.get_executor("agent2", "local:", code_for(blueprint2), venv_checkup_interval=0.1),
```

tests/forking_agent/test_agent_executor.py
```diff
@@ -312,3 +315,84 @@ def test():
...
+
+    def get_modification_datetime(file: str) -> datetime.datetime:
+        return datetime.datetime.fromtimestamp(os.stat(file).st_mtime)
+
+    old_check_executor1 = get_modification_datetime(
+        f"{executor_1.executor_virtual_env.env_path}/{const.INMANTA_ENV_STATUS_FILENAME}"
+    )
+    old_check_executor2 = get_modification_datetime(
+        f"{executor_2.executor_virtual_env.env_path}/{const.INMANTA_ENV_STATUS_FILENAME}"
+    )
```
arnaudsjs, line 370: Let's add a comment here to indicate that this only works because there is no `await` statement between the `os.utime()` call this these statements. This is very subtle.

tests/forking_agent/test_agent_executor.py
```diff
@@ -312,3 +315,84 @@ def test():
...
+    assert new_check_executor2 > old_check_executor2
+    assert (datetime.datetime.now() - new_check_executor2).seconds <= 2
+
+    monkeypatch.undo()
+
+    # Now we want to check if the cleanup is working correctly
+    await executor_manager.stop_for_agent("agent1")
+    # First we want to override the modification date of the `inmanta_env_status` file
+    os.utime(
+        f"{executor_1.executor_virtual_env.env_path}/.inmanta_env_status", (old_datetime.timestamp(), old_datetime.timestamp())
```
arnaudsjs, line 391: By updating both the modification and the access timestamp we lose test coverage. If the `clean_environments()` method only takes into account the access timestamp, this test case will succeed, but in production we will get different behavior, because the code that touches the file only updates the modification timestamp.

tests/forking_agent/test_agent_executor.py
```diff
@@ -312,3 +315,84 @@ def test():
...
+    )
+
+    venv_dir = pathlib.Path(mpmanager_light.environment_manager.envs_dir)
+    assert len([e for e in venv_dir.iterdir()]) == 2, "We should have two Virtual Environments for our 2 executors!"
+    # We remove the old VirtualEnvironment
+    await mpmanager_light.environment_manager.clean_environments()
+    assert len([e for e in venv_dir.iterdir()]) == 1, "Only one Virtual Environment should exist!"
```
arnaudsjs, line 398: This assertion only verifies that a venv was cleaned up. It doesn't verify that the correct one (the oldest) was removed.

src/inmanta/agent/config.py
```diff
@@ -160,6 +160,16 @@
...
 )
 
+agent_virtual_environment_cleanup: Option[float] = Option(
+    "config",
+    "agent-venv-cleanup",
+    7,
+    "This is the number of days to wait before unused Python virtual environments are cleaned from the inmanta "
+    "server. Setting this option too low may result in a high load on the Inmanta server. Setting it too high"
+    " may result in increased disk usage.",
+    is_float,
```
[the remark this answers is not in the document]
hlloreda: A float could be preferable: https://github.com/inmanta/inmanta-core/pull/7748#discussion_r1647196658

src/inmanta/agent/executor.py
```diff
@@ -258,7 +262,7 @@ async def create_and_install_environment(self, blueprint: EnvBlueprint) -> None:
                 requirements=list(pkg_resources.parse_requirements(req)),
                 config=blueprint.pip_config,
             )
-            await loop.run_in_executor(self.thread_pool, install_for_config)
+            install_for_config()
```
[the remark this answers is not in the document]
hlloreda: The whole `create_and_install_environment` method is running in an executor, so we are not blocking the IO loop, right?

tests/forking_agent/test_agent_executor.py
arnaudsjs: The reason why it's not possible get a race condition where the status file is touched between the call to `os.utime()` and the call to `get_modification_datetime()` is because there is no switching happening on the ioloop. Or am I missing something?

arnaudsjs (changes requested):
I see that many of the comments were not addressed. I am going to move this back to "request changes". Let me know if some comments are not clear to you.

src/inmanta/agent/config.py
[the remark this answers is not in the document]
wouterdb: I would also add a check to make very sure this is larger than the period with which we touch the file.

wouterdb (changes requested):

src/inmanta/agent/executor.py
```diff
@@ -354,8 +383,58 @@ async def get_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPoolE
...
+        loop = asyncio.get_running_loop()
+        for path_env_to_clean in environments_on_disk_to_clean:
+            # If the path to clean is part of the `_environment_map` then we can use the threadpool of the executor to clean
+            # the virtual environment
+            if path_env_to_clean in reverse_environment_map:
+                blueprint = reverse_environment_map[path_env_to_clean]
+                async with self._locks.get(blueprint.blueprint_hash()):
+                    await loop.run_in_executor(self._environment_map[blueprint].thread_pool, shutil.rmtree, path_env_to_clean)
+            else:
+                async with self._locks.get(path_env_to_clean):
```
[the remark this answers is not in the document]
hlloreda: This comment appears to be based on outdated code. The lock is on the directory name

src/inmanta/agent/forking_executor.py
```diff
@@ -517,14 +554,20 @@ def init_once(cls) -> None:
...
+        code: typing.Collection[executor.ResourceInstallSpec],
+        venv_checkup_interval: typing.Optional[int] = None,
     ) -> MPExecutor:
         """
         Retrieves an Executor based on the agent name and ResourceInstallSpec.
         If an Executor does not exist for the given configuration, a new one is created.
 
         :param agent_name: The name of the agent for which an Executor is being retrieved or created.
         :param code: The set of sources to be installed on the executor.
+        :param venv_checkup_interval: The time interval after which the virtual environment must be touched. Only used for
```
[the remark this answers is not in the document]
hlloreda: When we discussed this, it was decided to do this check every minute, but we could make it more flexible and allow the user to modify this value.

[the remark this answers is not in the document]
wouterdb: I don't necessarily want it to be more flexible. 

I just want it
1. to be clear that this has a default
2. what this default is 
3. why this is here


so for 1 ad 2 I would use a default argument
for 3, I would document that this value can be overridden to speed up tests

tests/forking_agent/test_agent_executor.py
```diff
@@ -312,3 +315,79 @@ def test():
...
+    assert new_check_executor2 > old_check_executor2
+    assert (datetime.datetime.now() - new_check_executor2).seconds <= 2
+
+    monkeypatch.undo()
+
+    # Now we want to check if the cleanup is working correctly
+    await executor_manager.stop_for_agent("agent1")
+    # First we want to override the modification date of the `inmanta_venv_status` file
+    os.utime(
+        f"{executor_1.executor_virtual_env.env_path}/.inmanta_venv_status", (old_datetime.timestamp(), old_datetime.timestamp())
```
arnaudsjs, line 384: Shouldn't it be like this:
```suggestion
        f"{executor_1.executor_virtual_env.env_path}/.inmanta_venv_status", (datetime.datetime.now(), old_datetime.timestamp())
```

arnaudsjs (changes requested):

src/inmanta/agent/agent.py
```diff
@@ -939,6 +940,8 @@ def __init__(
         if remote_executor and can_have_remote_executor:
             LOGGER.info("Selected forking agent executor mode")
             env_manager = inmanta.agent.executor.VirtualEnvironmentManager(self._storage["executor"])
+            # We need to do this otherwise, the scheduler would crash because no event loop would be running
+            IOLoop.current().add_callback(env_manager.start)
```
wouterdb, line 944: I would just add this to the start method (that is where we start stuff)

src/inmanta/agent/executor.py
```diff
@@ -284,13 +326,32 @@ class VirtualEnvironmentManager:
...
+        # allows us to prevent creating and deleting the same venv at a given time. The keys of this named lock are the hash of
+        # venv
         self._locks: NamedLock = NamedLock()
+        self._cleanup_scheduler = util.Scheduler("venv_cleanup_scheduler")
+
+    async def start(self) -> None:
+        interval = datetime.timedelta(days=1).total_seconds()
+        executor_venv_retention_time = cfg.executor_venv_retention_time.get()
+        assert datetime.timedelta(days=executor_venv_retention_time).total_seconds() > interval, (
+            "The `executor-venv-retention-time` should be larger than the period " "with which the executor touches its file!"
```
wouterdb, line 339: I would also log the values

```suggestion
            "The `executor-venv-retention-time` should be larger then 1second"
```

src/inmanta/agent/executor.py
```diff
@@ -284,13 +326,32 @@ class VirtualEnvironmentManager:
...
         self._environment_map: dict[EnvBlueprint, ExecutorVirtualEnvironment] = {}
         self.envs_dir: str = envs_dir
+        # We rely on a Named lock (`self._locks`) to be able to lock specific entries of the `_environment_map` dict. This
+        # allows us to prevent creating and deleting the same venv at a given time. The keys of this named lock are the hash of
+        # venv
         self._locks: NamedLock = NamedLock()
+        self._cleanup_scheduler = util.Scheduler("venv_cleanup_scheduler")
+
+    async def start(self) -> None:
+        interval = datetime.timedelta(days=1).total_seconds()
```
wouterdb, line 336: What is this value? it is used as the `the period with which the executor touches its file!` and it is how often we do the cleanup check? 

Why one day?

src/inmanta/agent/executor.py
```diff
@@ -354,8 +432,54 @@ async def get_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPoolE
...
+        """
+        current_datetime = datetime.datetime.now()
+        executor_venv_retention_time = cfg.executor_venv_retention_time.get()
+        envs_dir = pathlib.Path(self.envs_dir)
+        venv_path_to_blueprint = {pathlib.Path(v.env_path): k for k, v in self._environment_map.items()}
+
+        loop = asyncio.get_running_loop()
+        for root, folders, _ in os.walk(self.envs_dir):
+            for folder in folders:
+                async with self._locks.get(folder):
```
wouterdb, line 451:
```suggestion
                # the name of the folder is the hash!
                async with self._locks.get(folder):
```

src/inmanta/agent/executor.py
```diff
@@ -284,13 +326,32 @@ class VirtualEnvironmentManager:
...
         """
         Retrieves the directory path for a virtual environment based on the given blueprint.
-        If the directory does not exist, it creates a new one. This method ensures that each
-        virtual environment has a unique storage location.
+        If the directory does not exist, it creates a new one. This method ensures that each virtual environment has a unique
+        storage location. This method must be executed under the self._locks.get(<blueprint-hash>) lock to ensure thread-safe
+        operations for each unique blueprint
 
         :param blueprint: The blueprint of the environment for which the storage is being determined.
         :return: A tuple containing the path to the directory and a boolean indicating whether the directory was newly created.
```
wouterdb, line 357: I would add a comment to line 360 stating that the folder name has to be the hash, otherwise locking breaks in `clean_virtual_environments`

src/inmanta/agent/forking_executor.py
```diff
@@ -175,15 +200,27 @@ def __init__(
...
         self.sources = sources
         self.environment = environment
         self.uri = uri
+        self._venv_checkup_interval = venv_checkup_interval
 
     async def call(self, context: ExecutorContext) -> typing.Sequence[inmanta.loader.ModuleSource]:
+        assert context.server.timer_venv_checkup is None, "InitCommand should be only called once!"
+
+        interval = self._venv_checkup_interval or 60
+        context.server.start_timer_venv_checkup(interval)
```
wouterdb, line 222: `self.context.env` is not set yet, if any await call is added between here and line 243, it will give an assertion failure

tests/forking_agent/test_agent_executor.py
```diff
@@ -312,3 +315,86 @@ def test():
...
+    await asyncio.sleep(0.2)
+
+    new_check_executor1 = executor_1.executor_virtual_env.get_last_used_timestamp()
+    new_check_executor2 = executor_2.executor_virtual_env.get_last_used_timestamp()
+
+    assert new_check_executor1 > old_check_executor1
+    assert new_check_executor2 > old_check_executor2
+    assert (datetime.datetime.now() - new_check_executor2).seconds <= 2
+
+    monkeypatch.undo()
```
wouterdb, line 380: What is this for?

src/inmanta/agent/forking_executor.py
```diff
@@ -175,15 +200,27 @@ def __init__(
...
         self.gid = session_gid
         self.sources = sources
         self.environment = environment
         self.uri = uri
+        self._venv_checkup_interval = venv_checkup_interval
 
     async def call(self, context: ExecutorContext) -> typing.Sequence[inmanta.loader.ModuleSource]:
+        assert context.server.timer_venv_checkup is None, "InitCommand should be only called once!"
+
+        interval = self._venv_checkup_interval or 60
```
wouterdb, line 221: I find it hard to articulate why, but this way of setting a default doesn't feel right to me. 

1. it makes it hard to find here the default is set (I expect it to be in the signature of the constructor or at least in the constructor).
2. it opens up the opportunity for duplication of the default value: if other code want to now this interval, it can't get it, because the value we store is not always the value we use

wouterdb (changes requested):

src/inmanta/agent/executor.py
hlloreda: This is the interval of time for the cleanup check. It seems like a good balance if the default venv retention time is set to 7 days. It's not too aggressive

src/inmanta/agent/executor.py
```diff
@@ -333,16 +333,16 @@ def __init__(self, envs_dir: str) -> None:
         self._cleanup_scheduler = util.Scheduler("venv_cleanup_scheduler")
 
     async def start(self) -> None:
-        interval = datetime.timedelta(days=1).total_seconds()
+        interval_cleanup_check = datetime.timedelta(days=1).total_seconds()
```
wouterdb, line 336: Why check only once per day? 

I would expect we support retention times that are much shorter? We write the keep alive every minute, so we can go as low as e.g. 61 seconds here? 



I'm also going to point to @sanderr remark here: 

https://github.com/inmanta/inmanta-core/pull/7765#discussion_r1655064745

wouterdb (changes requested):

src/inmanta/util/__init__.py
```diff
@@ -384,6 +384,59 @@ def action_function() -> None:
...
+                except Exception:
+                    LOGGER.exception("Uncaught exception while executing scheduled action")
+                finally:
+                    # Final iteration
+                    del self._scheduled[task_spec]
+
+        handle: asyncio.TimerHandle = asyncio.get_running_loop().call_later(schedule_typed.get_initial_delay(), action_function)
+        self._scheduled[task_spec] = handle
+        return task_spec
+
```
hlloreda, line 439: @wouterdb In the latest commit, I tried to implement sander's suggestion. However, the scheduler is part of the stable api, so I'm not sure what I'm allowed (and not allowed) to do. Because of this, the latest commit is more of a draft than an actual implementation.

src/inmanta/agent/executor.py
```diff
@@ -338,7 +338,8 @@ def __init__(self, envs_dir: str) -> None:
         self._locks: NamedLock = NamedLock()
         # We know that the .inmanta venv status file is touched every minute, so `61` seconds is the lowest default we can use
         self._default_expiry_cleanup = 61
-        self._cleanup_scheduler = util.Scheduler("venv_cleanup_scheduler")
+        # self._cleanup_scheduler = util.Scheduler("venv_cleanup_scheduler")
+        asyncio.ensure_future(self.start())
```
wouterdb, line 342: No object should call its own start method 

The pattern we follow in most code is that the constructor is passive (it has no side effects) and sync.
 - This allows objects to be constructed with no side effects. 
 - It allows find grained control over start order. 
 - It allows validation of start conditions over many objects prior to actual start
 (many objects even had a two phase start, where pre_start does active checks )

The start method is active and async. 

The code constructing the `VirtualEnvironmentManager` should also start it.

src/inmanta/agent/executor.py
```diff
@@ -349,11 +350,7 @@ async def start(self) -> None:
         ), "The `executor-venv-retention-time` should be larger than 1 day!"
```
wouterdb, line 350: Why?

wouterdb (changes requested):
I don't understand the scheduling any more

arnaudsjs (changes requested):
Moving this to "request changes", because it needs the integration with the executor cleanup code.

hlloreda:
@wouterdb I tried to implement [this idea](https://inmanta.slack.com/archives/CKRF0C8R3/p1719825692066209?thread_ts=1719824573.908049&cid=CKRF0C8R3). If you feel that it adds too much complexity, we can revert the latest commit. Every comment should have been addressed, but I think some doc strings are missing (following the latest changes)

src/inmanta/agent/forking_executor.py
```diff
@@ -768,65 +834,26 @@ def _make_child(self, name: str, log_level: int, cli_log: bool) -> tuple[multipr
...
-        self.running = False
+        await super().stop()
         await asyncio.gather(*(child.stop() for child in self.children))
 
     async def force_stop(self, grace_time: float) -> None:
+        await super().stop()
         await asyncio.gather(*(child.force_stop(grace_time) for child in self.children))
 
     async def join(self, thread_pool_finalizer: list[concurrent.futures.ThreadPoolExecutor], timeout: float) -> None:
+        await super().stop()
```
arnaudsjs, line 848: This call was already made in the `stop()` method, so why do it here again?

arnaudsjs (changes requested):

src/inmanta/agent/executor.py
```diff
@@ -244,36 +247,153 @@ class ResourceInstallSpec:
...
+                self.env_path,
+                self.get_idle_time().total_seconds(),
+                cfg.executor_venv_retention_time.get(),
+            )
+        else:
+            LOGGER.debug(
+                "Removing venv %s because it is incomplete (broken venv),.",
+                self.env_path,
+            )
+        await loop.run_in_executor(self.thread_pool, self.remove_venv)
```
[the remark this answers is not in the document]
hlloreda: [It is not an issue](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor), right? 

If the `venv` is not present in the `self._environment_map`, then we don't know which thread pool this `ExecutorVirtualEnvironment` can use. Therefore, we must rely on the default one.

src/inmanta/agent/forking_executor.py
```diff
@@ -115,6 +115,7 @@ def __init__(self, name: str, logger: logging.Logger, take_over_logging: bool =
         self.ctx = ExecutorContext(self)
         self.log_transport: typing.Optional[LogShipper] = None
         self.take_over_logging = take_over_logging
+        self.timer_venv_checkup: typing.Optional[util.Scheduler] = None
```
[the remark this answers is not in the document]
hlloreda: Because we use this to make sure the `InitCommand` is called only once.

src/inmanta/agent/forking_executor.py
hlloreda: It was a precaution in case we would call join without calling stop afterward, but in the current implementation, we don't need this

src/inmanta/agent/executor.py
```diff
@@ -289,27 +409,120 @@ def initialize_envs_directory() -> str:
...
+                        if pool_member.can_be_cleaned_up(self.retention_time):
+                            await pool_member.clean()
+                            self.clean_pool_member_from_manager(pool_member)
+                else:
+                    reschedule_interval = min(
+                        reschedule_interval,
+                        (
+                            datetime.timedelta(seconds=self.retention_time) - (cleanup_start - pool_member.last_used())
+                        ).total_seconds(),
+                    )
```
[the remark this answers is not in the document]
hlloreda: We want to have the shortest `reschedule_interval` where we would actually do something (clean an executor / venv). 

If a pool member was not used for a long time, the `reschedule_interval` would be very low. With this, we will be able to clean it immediately after it expires.

src/inmanta/agent/executor.py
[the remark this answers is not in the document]
hlloreda: What I said is not entirely correct because we know that the thread pool is set by the `MPManager`.
image

tests/forking_agent/test_agent_executor.py
```diff
@@ -330,3 +332,95 @@ def test():
...
+    new_check_executor1 = executor_1.executor_virtual_env.last_used()
+    new_check_executor2 = executor_2.executor_virtual_env.last_used()
+
+    assert new_check_executor1 > old_check_executor1
+    assert new_check_executor2 > old_check_executor2
+    assert (datetime.datetime.now().astimezone() - new_check_executor2).seconds <= 2
+
+    # Now we want to check if the cleanup is working correctly
+    await executor_manager.stop_for_agent("agent1")
+    await asyncio.sleep(0.2)
```
arnaudsjs, line 397: Why this sleep?

tests/forking_agent/test_agent_executor.py
```diff
@@ -330,3 +332,95 @@ def test():
...
+
+    venv_dir = pathlib.Path(mpmanager_light.environment_manager.envs_dir)
+    assert len([e for e in venv_dir.iterdir()]) == 2, "We should have two Virtual Environments for our 2 executors!"
+    # We remove the old VirtualEnvironment
+    logger.debug("Calling cleanup_virtual_environments")
+    mpmanager_light.environment_manager.running = True
+    try:
+        await asyncio.wait_for(mpmanager_light.environment_manager.cleanup_inactive_pool_members(), 0.5)
+    except TimeoutError:
+        pass
```
arnaudsjs, line 411: Why pass here? The remaining of the test will not succeed if this call didn't finish completely? By doing a pass we will also on the CI not know whether the timeout happened or not. We will just see a failure on one of the assertions that follow.

tests/forking_agent/test_agent_executor.py
```diff
@@ -330,3 +332,95 @@ def test():
...
+    assert [executor_2.executor_virtual_env.env_path] == venvs
+
+    # Let's stop the other agent and pretend that the venv is broken
+    await executor_manager.stop_for_agent("agent2")
+    await asyncio.sleep(0.2)
+    executor_2_venv_status_file.unlink()
+    try:
+        await asyncio.wait_for(mpmanager_light.environment_manager.cleanup_inactive_pool_members(), 0.5)
+    except TimeoutError:
+        pass
```
arnaudsjs, line 424: Same here.

tests/forking_agent/test_agent_executor.py
```diff
@@ -330,3 +332,95 @@ def test():
...
+    logger.debug("cleanup_virtual_environments ended")
+    venvs = [str(e) for e in venv_dir.iterdir()]
+    assert len(venvs) == 1, "Only one Virtual Environment should exist!"
+    assert [executor_2.executor_virtual_env.env_path] == venvs
+
+    # Let's stop the other agent and pretend that the venv is broken
+    await executor_manager.stop_for_agent("agent2")
+    await asyncio.sleep(0.2)
```
arnaudsjs, line 419: Why this sleep?

src/inmanta/agent/forking_executor.py
```diff
@@ -144,6 +146,19 @@ def _detach_log_shipper(self) -> None:
...
+        self.timer_venv_checkup.add_action(
+            action=self.touch_inmanta_venv_status,
+            schedule=util.IntervalSchedule(
+                interval=interval,
+            ),
+        )
+
+    def stop_timer_venv_checkup(self) -> None:
+        if self.timer_venv_checkup is not None:
+            asyncio.ensure_future(self.timer_venv_checkup.stop())
```
arnaudsjs, line 160: I am not sure what to do with this. Like this the `stop()` method will only get called when the ioloop picks up the task. This might never happen. The annoying thing is that the `connection_lost()` method is a sync method, so there is no easy way to make sure that the `self.timer_venv_checkup.stop()` call gets awaited. @sanderr Do you see an easy way out here?

arnaudsjs (changes requested):

tests/forking_agent/test_agent_executor.py
hlloreda: To make sure the `StopCommand` has been processed by the `ExecutorServer` as we don't receive a reply to confirm that everything has been stopped

arnaudsjs: You can use `psutil` to list the subprocesses and verify that the executor process actually stopped. Sleeps like this should be prevented as much as possible. They make the test suite unstable/unpredictable and also slower than required.

tests/forking_agent/test_agent_executor.py
```diff
@@ -330,3 +334,121 @@ def test():
...
+    assert (datetime.datetime.now().astimezone() - new_check_executor2).seconds <= 2
+
+    async def wait_for_agent_stop_running(pid_agent: int) -> None:
+        """
+        Wait for the agent to stop running
+
+        :param pid_agent: The PID of the agent
+        """
+        for i in range(10):
+            if psutil.pid_exists(pid_agent):
```
arnaudsjs, line 420:
```suggestion
            if not psutil.pid_exists(pid_agent):
```

tests/forking_agent/test_agent_executor.py
```diff
@@ -330,3 +334,121 @@ def test():
...
+
+    def retrieve_pid_agent(agent_name: str) -> int:
+        """
+        Retrieve the PID of the agent
+
+        :param agent_name: The name of the agent from which we want to retrieve the PID
+        """
+        for proc in psutil.process_iter(["pid", "name"]):
+            if f"inmanta: executor {agent_name}" in proc.name():
+                return proc.pid
```
arnaudsjs, line 383: It's better to return the process object than to return the pid. You can then use `process.is_running()` to verify whether the process is still there. Using the pid has a race condition, because the pid might be re-used by a different process after the process stopped.

arnaudsjs (changes requested):

arnaudsjs (approved):

src/inmanta/agent/executor.py
```diff
@@ -289,27 +404,137 @@ def initialize_envs_directory() -> str:
...
+
+    async def stop(self) -> None:
+        """
+        Cancel the cleaning job of the Pool Manager -> We don't want to wait for the cleaning to finish because it could take a
+        long time.
+        """
+        self.running = False
+        if self.cleanup_job is not None:
+            try:
+                self.cleanup_job.cancel()
```
wouterdb, line 436: Why do we have both running=False and also cancel? 

They both terminate the run, but differently?

src/inmanta/agent/executor.py
```diff
@@ -289,27 +404,137 @@ def initialize_envs_directory() -> str:
...
+        cleanup_end = datetime.datetime.now().astimezone()
+
+        return max(0.0, run_next_cleanup_job_at - (cleanup_end - cleanup_start).total_seconds())
+
+    async def cleanup_inactive_pool_members_task(self) -> None:
+        """
+        This task periodically cleans up idle pool member
+        """
+        while self.running:
+            await asyncio.sleep(await self.cleanup_inactive_pool_members())
```
wouterdb, line 505: I would find it easier to read if the `cleanup_inactive_pool_members` was part of this method as well (now the sleep is an odd place)

src/inmanta/agent/executor.py
```diff
@@ -369,7 +608,46 @@ async def get_environment(self, blueprint: EnvBlueprint, threadpool: ThreadPoolE
...
+            # the name of the folder is the hash!
+            async with self._locks.get(folder):
+                current_folder = envs_dir / folder
+                blueprint = venv_path_to_blueprint.get(current_folder, None)
+                if blueprint is not None and blueprint in self._environment_map:
+                    current_executor_environment = self._environment_map[blueprint]
+                else:
+                    current_executor_environment = ExecutorVirtualEnvironment(env_path=str(current_folder.absolute()))
+
+                pool_members.append(current_executor_environment)
```
wouterdb, line 639: I wonder if we shouldn't cache all this: load from disk when initializing, then work from memory?

src/inmanta/agent/forking_executor.py
```diff
@@ -163,12 +187,21 @@ def _sync_stop(self) -> None:
 
     def connection_lost(self, exc: Exception | None) -> None:
         """We lost connection to the controler, bail out"""
+        self.stop_timer_venv_checkup()
```
wouterdb, line 190: this one is not  required, as we call `_sync_stop` later on

src/inmanta/agent/forking_executor.py
```diff
@@ -428,7 +472,7 @@ def __init__(
         self.failed_resource_sources: typing.Sequence[inmanta.loader.ModuleSource] = list()
         self.failed_resource_types: set[ResourceType] = set()
 
-    async def stop(self) -> None:
+    def stop(self) -> None:
```
wouterdb, line 475: too risky

src/inmanta/agent/forking_executor.py
```diff
@@ -144,6 +151,22 @@ def _detach_log_shipper(self) -> None:
...
+                interval=interval,
+            ),
+        )
+
+    def stop_timer_venv_checkup(self) -> None:
+        # Given that the stop method of the Scheduler is async, we have no way to make sure that the stop method will get called
+        # (only if the ioloop picks up the task -> might never happen). Therefore, we rely on the sync method that will cancel
+        # the checkup task
+        if self.timer_venv_scheduler is not None and self.timer_venv_checkup is not None:
+            self.timer_venv_scheduler.remove(self.timer_venv_checkup)
```
wouterdb, line 168:
```suggestion
while not self.stopping:
    self.touch_inmanta_venv_status()
    await asyncio.sleep(self.interval)
```

src/inmanta/agent/forking_executor.py
```diff
@@ -768,65 +844,30 @@ def _make_child(self, name: str, log_level: int, cli_log: bool) -> tuple[multipr
...
+        await super().start()
+        # We need to do this here, otherwise, the scheduler would crash because no event loop would be running
+        await self.environment_manager.start()
 
     async def stop(self) -> None:
-        self.running = False
-        await asyncio.gather(*(child.stop() for child in self.children))
+        loop = asyncio.get_running_loop()
+        await super().stop()
+        await asyncio.gather(*(loop.run_in_executor(self.thread_pool, child.stop) for child in self.children))
```
wouterdb, line 854: this is not thread pool safe code

src/inmanta/agent/forking_executor.py
```diff
@@ -768,65 +844,30 @@ def _make_child(self, name: str, log_level: int, cli_log: bool) -> tuple[multipr
...
+        await self.environment_manager.stop()
 
     async def force_stop(self, grace_time: float) -> None:
+        await super().stop()
```
wouterdb, line 858: this is very odd, calling the super method of another method

src/inmanta/agent/forking_executor.py
```diff
@@ -768,65 +844,30 @@ def _make_child(self, name: str, log_level: int, cli_log: bool) -> tuple[multipr
...
 
     async def stop(self) -> None:
-        self.running = False
-        await asyncio.gather(*(child.stop() for child in self.children))
+        loop = asyncio.get_running_loop()
+        await super().stop()
+        await asyncio.gather(*(loop.run_in_executor(self.thread_pool, child.stop) for child in self.children))
+        await self.environment_manager.stop()
 
     async def force_stop(self, grace_time: float) -> None:
```
wouterdb, line 857: This can be remove I think

tests/forking_agent/test_agent_executor.py
```diff
@@ -330,3 +342,121 @@ def test():
...
+            if not process_agent.is_running():
+                return
+            else:
+                await asyncio.sleep(0.2)
+
+        raise RuntimeError("The agent was still running after 2 seconds")
+
+    # Now we want to check if the cleanup is working correctly
+    await executor_manager.stop_for_agent("agent1")
+    await wait_for_agent_stop_running(process_agent_1)
```
wouterdb, line 437: I would use the retry limited here

wouterdb (changes requested):

src/inmanta/agent/executor.py
```diff
@@ -427,22 +427,11 @@ async def start(self) -> None:
...
     async def stop(self) -> None:
         """
-        Cancel the cleaning job of the Pool Manager -> We don't want to wait for the cleaning to finish because it could take a
-        long time.
+        Stop the cleaning job of the Pool Manager.
         """
+        # We don't want to cancel the task because it could lead to an inconsistent state (e.g. Venv half removed). Therefore,
+        # we need to wait for the completion of the task
         self.running = False
-        if self.cleanup_job is not None:
```
wouterdb: It would be good to have a join method as well

src/inmanta/agent/executor.py
```diff
@@ -500,9 +488,13 @@ async def cleanup_inactive_pool_members(self) -> float:
...
         """
         This task periodically cleans up idle pool member
+
+        We split up `cleanup_inactive_pool_members` and `cleanup_inactive_pool_members_task` in order to be able to call the
+        cleanup method in the test without being blocked in a loop.
         """
         while self.running:
-            await asyncio.sleep(await self.cleanup_inactive_pool_members())
+            sleep_interval = await self.cleanup_inactive_pool_members()
+            await asyncio.sleep(sleep_interval)
```
wouterdb, line 497: you wait even when not running, it might be better to not wait when stopping

wouterdb (changes requested):

wouterdb (approved):

PR #7749, opened 2024-06-19
Fix log shipper

wouterdb:
# Description

Make log shipper include exception traces 

# Self Check:

- [x] Changelog entry


src/inmanta/protocol/ipc_light.py
```diff
@@ -329,7 +329,7 @@ def emit(self, record: logging.LogRecord) -> None:
                 IPCLogRecord(
                     record.name,
                     record.levelno,
-                    record.getMessage(),
```
wouterdb, line 332: This was not sufficient...

arnaudsjs (approved):

PR #7754, opened 2024-06-20
allow indexes on nullable attributes

### LINKED ISSUE #7204: Allow nullable attributes to be used in indexes
They must follow the normal rules, that they either need a default of be set in the constructor explicitly

sanderr:
# Description

Allow indexes on nullable attributes

closes #7204

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -33,9 +33,9 @@ from inmanta.ast import (
 )
 from inmanta.ast.blocks import BasicBlock
 from inmanta.ast.statements.generator import SubConstructor
-from inmanta.ast.type import Float, NamedType, Type
+from inmanta.ast.type import Float, NamedType, NullableType, Type
 from inmanta.execute.runtime import Instance, QueueScheduler, Resolver, dataflow
-from inmanta.execute.util import AnyType
+from inmanta.execute.util import AnyType, NoneValue
 
 try:
     from typing import TYPE_CHECKING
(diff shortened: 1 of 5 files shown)


tests/compiler/test_index.py
```diff
@@ -81,21 +81,6 @@ def test_issue_140_index_error(snippetcompiler):
         assert re.match(".*No index defined on std::Service for this lookup:.*", str(e))
 
 
-def test_issue_745_index_on_nullable(snippetcompiler):
```
sanderr, line 84: Deleted because it no longer applies (tested the opposite of this feature)

tests/compiler/test_typing.py
```diff
@@ -534,22 +534,6 @@ def test_float_type(snippetcompiler):
     assert Number().validate(u)
 
 
-def test_lookup_on_float_with_int(snippetcompiler):
```
sanderr, line 537: Moved this to `test_index.py` where I felt it was more appropriate.

sanderr: And updated + extended implementation

src/inmanta/ast/entity.py
```diff
@@ -433,12 +433,24 @@ def lookup_index(
                 stmt, self.get_full_name(), "No index defined on %s for this lookup: " % self.get_full_name() + str(params)
             )
 
+        def coerce(t: Type, v: object) -> object:
+            """
+            Coerce float-typed values to float (e.g. 1 -> 1.0)
+            """
+            match t:
```
wouterdb, line 440: cool!

tests/compiler/test_index.py
```diff
@@ -592,3 +577,128 @@ def test_index_collision_exception(snippetcompiler) -> None:
...
+def test_index_on_nullable(snippetcompiler) -> None:
+    """
+    Verify that indexes on nullable attributes are allowed and behave as expected.
+
+    Indexes on optional relations are not supported and are tested in test_issue_745_2689_index_on_optional.
+    """
+    snippetcompiler.setup_for_snippet(
+        """
+        entity A:
+            int? n = null
```
wouterdb, line 591: this should also work if no default value is set, but then it must be set explicit in the constructor I think.

This case is not tested

wouterdb (changes requested):
one remark

tests/compiler/test_index.py
sanderr: Can you elaborate? The following model works, and I believe that it should. Do you think differently?

```
entity A:
    int? n
end
index A(n)

implement A using std::none

a_null = A(n=null)
a_zero = A(n=0)


assert = true
assert = (a_null != a_zero)
assert = (a_null == A(n=null))
assert = (a_null == A[n=null])
assert = (a_zero == A(n=0))
assert = (a_zero == A[n=0])
```

sanderr: Requesting rereview to bring this to your attention

wouterdb: it should

wouterdb: and this one should fail

```
entity A: 
  int? n 
end 

index A(n) 

implement A using std::none 

A()
```

wouterdb (approved):

Hugo-Inmanta (approved):
Pattern matching seems super powerful

Issue #7694, opened 2024-05-31
Executor Cleanup


wouterdb:
We currently never remove executors.

We should have a policy to clean them up.

# Proposal

1. every agent can have at most X executors (configurable, default 3 (one for deploy, one dryrun and one spare)
2. every executor that hasn't been used for X time unit is removed (configurable, default 1 minute)

Link to [design](https://github.com/inmanta/designs/blob/c351ba26f7316c00b10e2d3e8f881c58dd37d8f0/agent-refactor.md)

PR #7765, opened 2024-06-24
Issue/7694 executor cleanup

Hugo-Inmanta:
# Description

Closes https://github.com/inmanta/inmanta-core/issues/7694

Add 2 config options:
- `executor_retention`: executors idle for longer than this get cleaned up
- `executor_cap_per_agent`: if this limit is already reached when attempting to create a new executor for this agent, the oldest executor is first stopped.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -992,7 +992,6 @@ class Agent(SessionEndpoint):
         can_have_remote_executor = code_loader
 
         self.executor_manager: executor.ExecutorManager[executor.Executor]
-        LOGGER.error(f"{LOGGER.level=}")
         if remote_executor and can_have_remote_executor:
             LOGGER.info("Selected forking agent executor mode")
             env_manager = inmanta.agent.executor.VirtualEnvironmentManager(self._storage["executor"])
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -537,7 +537,6 @@ class AgentInstance:
         self._repair_splay_value = random.randint(0, repair_splay_time)
 
         self._executor_retention: int = cfg.executor_retention.get()
-        self._executor_cap_per_agent: int = cfg.executor_cap_per_agent.get()
 
         self._getting_resources = False
         self._get_resource_timeout = 0
@@ -631,23 +630,33 @@ class AgentInstance:
             )
 
         async def cleanup_executors_action() -> None:
-            now = datetime.datetime.now().astimezone()
-            await self.executor_manager.cleanup_inactive_executors(now, self._executor_retention)
+            await self.executor_manager.cleanup_inactive_executors(self._executor_retention)
 
         def periodic_schedule(
             kind: str,
             action: Callable[[], Coroutine[object, None, object]],
             interval: Union[int, str],
             splay_value: int,
-            initial_time: datetime.datetime,
         ) -> bool:
+            """
+            Schedule a periodic task
+
+            :param kind: Name of the task (value to display in logs)
+            :param action: The action to schedule periodically
+            :param interval: The interval at which to schedule the task. Can be specified as either a number of
+                seconds, or a cron string.
+            :param splay_value: When specifying the interval as a number of seconds, this parameter specifies
+                the number of seconds by which to delay the initial execution of this action.
+            """
+            now = datetime.datetime.now().astimezone()
+
             if isinstance(interval, int) and interval > 0:
                 self.logger.info(
                     "Scheduling periodic %s with interval %d and splay %d (first run at %s)",
                     kind,
                     interval,
                     splay_value,
-                    (initial_time + datetime.timedelta(seconds=splay_value)).strftime(const.TIME_LOGFMT),
+                    (now + datetime.timedelta(seconds=splay_value)).strftime(const.TIME_LOGFMT),
                 )
                 interval_schedule: IntervalSchedule = IntervalSchedule(
                     interval=float(interval), initial_delay=float(splay_value)
@@ -675,9 +684,9 @@ class AgentInstance:
             )
             self.ensure_deploy_on_start = False
 
-        periodic_schedule("deploy", deploy_action, self._deploy_interval, self._deploy_splay_value, now)
-        periodic_schedule("repair", repair_action, self._repair_interval, self._repair_splay_value, now)
-        periodic_schedule("executor cleanup", cleanup_executors_action, self._executor_retention, 0, now)
+        periodic_schedule("deploy", deploy_action, self._deploy_interval, self._deploy_splay_value)
+        periodic_schedule("repair", repair_action, self._repair_interval, self._repair_splay_value)
+        periodic_schedule("executor cleanup", cleanup_executors_action, self._executor_retention, 0)
 
     def _enable_time_trigger(self, action: TaskMethod, schedule: TaskSchedule) -> None:
         self.process._sched.add_action(action, schedule)
@@ -905,7 +914,6 @@ class Agent(SessionEndpoint):
         :param environment: environment id
         """
         super().__init__("agent", timeout=cfg.server_timeout.get(), reconnect_delay=cfg.agent_reconnect_delay.get())
-        # agent_trigger_method_on_autodeploy = cast(str, await env.get(data.AGENT_TRIGGER_METHOD_ON_AUTO_DEPLOY))
 
         self.hostname = hostname
         self.ratelimiter = asyncio.Semaphore(1)
@@ -959,6 +967,7 @@ class Agent(SessionEndpoint):
                 self._storage["executor"],
                 LOGGER.level,
                 False,
+                max_executors_per_agent=cfg.executor_cap_per_agent,
             )
         else:
             LOGGER.info("Selected threaded agent executor mode")
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -628,9 +628,6 @@ class AgentInstance:
                 )
             )
 
-        async def cleanup_executors_action() -> None:
-            await self.executor_manager.cleanup_inactive_executors()
-
         def periodic_schedule(
             kind: str,
             action: Callable[[], Coroutine[object, None, object]],
@@ -685,7 +682,6 @@ class AgentInstance:
 
         periodic_schedule("deploy", deploy_action, self._deploy_interval, self._deploy_splay_value)
         periodic_schedule("repair", repair_action, self._repair_interval, self._repair_splay_value)
-        periodic_schedule("executor cleanup", 2, self._executor_retention, 0)
 
     def _enable_time_trigger(self, action: TaskMethod, schedule: TaskSchedule) -> None:
         self.process._sched.add_action(action, schedule)
@@ -966,7 +962,6 @@ class Agent(SessionEndpoint):
                 self._storage["executor"],
                 LOGGER.level,
                 False,
-                max_executors_per_agent=cfg.executor_cap_per_agent.get(),
             )
         else:
             LOGGER.info("Selected threaded agent executor mode")
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -536,7 +536,6 @@ class AgentInstance:
         repair_splay_time = cfg.agent_repair_splay_time.get()
         self._repair_splay_value = random.randint(0, repair_splay_time)
 
-        self._executor_retention: int = cfg.executor_retention.get()
 
         self._getting_resources = False
         self._get_resource_timeout = 0
@@ -630,7 +629,7 @@ class AgentInstance:
             )
 
         async def cleanup_executors_action() -> None:
-            await self.executor_manager.cleanup_inactive_executors(self._executor_retention)
+            await self.executor_manager.cleanup_inactive_executors()
 
         def periodic_schedule(
             kind: str,
@@ -686,7 +685,7 @@ class AgentInstance:
 
         periodic_schedule("deploy", deploy_action, self._deploy_interval, self._deploy_splay_value)
         periodic_schedule("repair", repair_action, self._repair_interval, self._repair_splay_value)
-        periodic_schedule("executor cleanup", cleanup_executors_action, self._executor_retention, 0)
+        periodic_schedule("executor cleanup", 2, self._executor_retention, 0)
 
     def _enable_time_trigger(self, action: TaskMethod, schedule: TaskSchedule) -> None:
         self.process._sched.add_action(action, schedule)
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -992,6 +992,7 @@ class Agent(SessionEndpoint):
         can_have_remote_executor = code_loader
 
         self.executor_manager: executor.ExecutorManager[executor.Executor]
+        LOGGER.error(f"{LOGGER.level=}")
         if remote_executor and can_have_remote_executor:
             LOGGER.info("Selected forking agent executor mode")
             env_manager = inmanta.agent.executor.VirtualEnvironmentManager(self._storage["executor"])
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -1079,7 +1079,6 @@ class Agent(SessionEndpoint):
         await super().start()
         await self.executor_manager.start()
 
-
     async def add_end_point_name(self, name: str) -> None:
         async with self._instances_lock:
             await self._add_end_point_name(name)
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -536,6 +536,9 @@ class AgentInstance:
         repair_splay_time = cfg.agent_repair_splay_time.get()
         self._repair_splay_value = random.randint(0, repair_splay_time)
 
+        self._executor_retention: int = cfg.executor_retention.get()
+        self._executor_cap_per_agent: int = cfg.executor_cap_per_agent.get()
+
         self._getting_resources = False
         self._get_resource_timeout = 0
 
@@ -627,6 +630,10 @@ class AgentInstance:
                 )
             )
 
+        async def cleanup_executors_action() -> None:
+            now = datetime.datetime.now().astimezone()
+            await self.executor_manager.cleanup_inactive_executors(now, self._executor_retention)
+
         def periodic_schedule(
             kind: str,
             action: Callable[[], Coroutine[object, None, object]],
@@ -667,8 +674,10 @@ class AgentInstance:
                 )
             )
             self.ensure_deploy_on_start = False
+
         periodic_schedule("deploy", deploy_action, self._deploy_interval, self._deploy_splay_value, now)
         periodic_schedule("repair", repair_action, self._repair_interval, self._repair_splay_value, now)
+        periodic_schedule("executor cleanup", cleanup_executors_action, self._executor_retention, 0, now)
 
     def _enable_time_trigger(self, action: TaskMethod, schedule: TaskSchedule) -> None:
         self.process._sched.add_action(action, schedule)
@@ -896,6 +905,7 @@ class Agent(SessionEndpoint):
         :param environment: environment id
         """
         super().__init__("agent", timeout=cfg.server_timeout.get(), reconnect_delay=cfg.agent_reconnect_delay.get())
+        # agent_trigger_method_on_autodeploy = cast(str, await env.get(data.AGENT_TRIGGER_METHOD_ON_AUTO_DEPLOY))
 
         self.hostname = hostname
         self.ratelimiter = asyncio.Semaphore(1)
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -1077,6 +1077,8 @@ class Agent(SessionEndpoint):
         # cache reference to THIS ioloop for handlers to push requests on it
         self._io_loop = asyncio.get_running_loop()
         await super().start()
+        await self.executor_manager.start()
+
 
     async def add_end_point_name(self, name: str) -> None:
         async with self._instances_lock:
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -967,7 +967,7 @@ class Agent(SessionEndpoint):
                 self._storage["executor"],
                 LOGGER.level,
                 False,
-                max_executors_per_agent=cfg.executor_cap_per_agent,
+                max_executors_per_agent=cfg.executor_cap_per_agent.get(),
             )
         else:
             LOGGER.info("Selected threaded agent executor mode")
--- a/src/inmanta/agent/agent.py
+++ b/src/inmanta/agent/agent.py
@@ -41,7 +41,14 @@ from inmanta.agent import config as cfg
 from inmanta.agent import executor, forking_executor, in_process_executor
 from inmanta.agent.executor import ResourceDetails, ResourceInstallSpec
 from inmanta.agent.reporting import collect_report
-from inmanta.data.model import LEGACY_PIP_DEFAULT, AttributeStateChange, PipConfig, ResourceIdStr, ResourceVersionIdStr
+from inmanta.data.model import (
+    LEGACY_PIP_DEFAULT,
+    AttributeStateChange,
+    PipConfig,
+    ResourceIdStr,
+    ResourceType,
+    ResourceVersionIdStr,
+)
 from inmanta.loader import CodeLoader, ModuleSource
 from inmanta.protocol import SessionEndpoint, SyncClient, methods, methods_v2
 from inmanta.resources import Id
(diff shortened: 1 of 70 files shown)


src/inmanta/agent/forking_executor.py
```diff
@@ -656,13 +680,21 @@ def _child_closed(self, child_handle: MPExecutor) -> None:
...
         p.start()
-        child_conn.close()
+        # child_conn.close() -> Why ?
+
+        # https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.connect_accepted_socket
+        #
+        # Note
+        #
+        # The sock argument transfers ownership of the socket to the transport created.
+        # To close the socket, call the transport’s close() method.
```
Hugo-Inmanta, line 697: Open question 1: Why do we `child_conn.close()` here ?

src/inmanta/agent/agent.py
```diff
@@ -627,20 +629,34 @@ async def repair_action() -> None:
                 )
             )
 
+        async def cleanup_executors_action() -> None:
+            await self.executor_manager.cleanup_inactive_executors(self._executor_retention)
```
wouterdb, line 633: I would move this into the executor manager itself. I see little reason for the agent to be responsible for the timer of the manager

src/inmanta/agent/agent.py
```diff
@@ -667,8 +683,10 @@ def periodic_schedule(
                 )
             )
             self.ensure_deploy_on_start = False
-        periodic_schedule("deploy", deploy_action, self._deploy_interval, self._deploy_splay_value, now)
-        periodic_schedule("repair", repair_action, self._repair_interval, self._repair_splay_value, now)
+
+        periodic_schedule("deploy", deploy_action, self._deploy_interval, self._deploy_splay_value)
+        periodic_schedule("repair", repair_action, self._repair_interval, self._repair_splay_value)
+        periodic_schedule("executor cleanup", cleanup_executors_action, self._executor_retention, 0)
```
wouterdb, line 689: I don't think this produces the expected behavior: it clean up every  `executor-retention` it doesn't clean after `executor-retention` 


(i.e. clean up after 24h is not the same as cleaning up every 24h)

wouterdb (changes requested):

wouterdb (changes requested):

src/inmanta/agent/forking_executor.py
```diff
@@ -681,3 +708,14 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
         children = [self.executor_map[child_id] for child_id in children_ids]
         await asyncio.gather(*(child.stop() for child in children))
         return children
+
+    async def cleanup_inactive_executors(self) -> None:
+        now = datetime.datetime.now().astimezone()
+
+        for _executor in self.executor_map.values():
```
sanderr, line 715: Instead of calling this on a schedule and to go over all executors every two seconds, how about the following?

1. in this method, stop all timed out executors and save `next_expiry = min(retention_time - (now - last_used))` (pseudo-code)
2. at the end of this method, reschedule itself on the IO loop in `next_expiry` seconds.
3. When the setting is changed, we may need to take some action to reschedule.

We would gain the following advantages:
a. More responsive: clean up after 60 seconds rather than after 61.9 seconds. Not too important IMO
b. Safer: no risk of being in this method twice.
c. More to the point: on low-load times we may end up just calling this method every 60 seconds.

They're not big gains, but I'd say it's an easy win, and it feels like a cleaner approach to me.

@wouterdb FYI

sanderr (changes requested):

src/inmanta/agent/forking_executor.py
```diff
@@ -681,3 +708,14 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
...
+            if now - _executor.connection.last_used_at > timedelta(seconds=self.executor_retention_time):
+                LOGGER.info(
```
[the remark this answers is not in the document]
wouterdb: I think you are right

src/inmanta/agent/forking_executor.py
```diff
@@ -681,3 +708,14 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
...
+                    f"Stopping executor {_executor.executor_id.identity()} because it was inactive for longer "
+                    f"than the configured allowed idling time."
+                )
+                await _executor.stop()
```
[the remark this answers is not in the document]
wouterdb: I would assume 10's of milliseconds, but should be easy to measure.

src/inmanta/agent/forking_executor.py
```diff
@@ -360,7 +400,7 @@ def __init__(
         self,
         owner: "MPManager",
         process: multiprocessing.Process,
-        connection: FinalizingIPCClient[ExecutorContext],
+        connection: ExecutorClient,
```
Hugo-Inmanta, line 403: I think this is correct ?

src/inmanta/agent/forking_executor.py
```diff
@@ -159,7 +169,37 @@ def connection_lost(self, exc: Exception | None) -> None:
...
+
+    @typing.overload
+    def call(self, method: IPCMethod[ExecutorContext, ReturnType], has_reply: typing.Literal[False]) -> None: ...
+
+    def call(self, method: IPCMethod[ExecutorContext, ReturnType], has_reply: bool = True) -> Future[ReturnType] | None:
+        """Call a method with given arguments"""
+        self.last_used_at = datetime.datetime.now().astimezone()
+        response = super().call(method, has_reply)
+        assert response is None or isinstance(response, Future)
+        return response
```
Hugo-Inmanta, line 194: This part I'm unsure of as well and I don't understand the resulting mypy error
```
+ src/inmanta/agent/forking_executor.py:192: error: No overload variant of "call" of "IPCClient" matches argument types "IPCMethod[ExecutorContext, ReturnType]", "bool"  [call-overload]
+ src/inmanta/agent/forking_executor.py:192: note: Possible overload variants:
+ src/inmanta/agent/forking_executor.py:192: note:     def [ReturnType] call(self, method: IPCMethod[ExecutorContext, ReturnType], has_reply: Literal[True] = ...) -> Future[ReturnType]
+ src/inmanta/agent/forking_executor.py:192: note:     def [ReturnType] call(self, method: IPCMethod[ExecutorContext, ReturnType], has_reply: Literal[False]) -> None

```

src/inmanta/agent/forking_executor.py
```diff
@@ -548,11 +559,14 @@ async def get_executor(
         self, agent_name: str, agent_uri: str, code: typing.Collection[executor.ResourceInstallSpec]
```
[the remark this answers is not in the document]
Hugo-Inmanta: > 
    do the cleanup under lock (or at least the stop part)
    we need to 'touch' the last_used time before letting go of the lock in this method.

Done

src/inmanta/agent/forking_executor.py
wouterdb: https://github.com/python/mypy/issues/14764

src/inmanta/agent/forking_executor.py
wouterdb: yes

src/inmanta/agent/forking_executor.py
```diff
@@ -709,13 +749,35 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
         await asyncio.gather(*(child.stop() for child in children))
         return children
 
-    async def cleanup_inactive_executors(self) -> None:
-        now = datetime.datetime.now().astimezone()
+    def start_periodic_executor_cleanup(self) -> None:
+        async def cleanup_inactive_executors() -> None:
```
wouterdb, line 753: this is quire smart!

src/inmanta/agent/forking_executor.py
```diff
@@ -709,13 +749,35 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
...
+                if _executor.can_be_cleaned_up(self.executor_retention_time):
+                    LOGGER.debug(
+                        "Stopping executor %s because it "
+                        "was inactive for %d s, which is longer then the retention time of %d s.",
+                        _executor.executor_id.identity(),
+                        _executor.connection.get_idle_time().total_seconds(),
+                        self.executor_retention_time,
+                    )
+                    async with self._locks.get(_executor.executor_id.identity()):
+                        await _executor.stop()
```
wouterdb, line 770: I would add a try block here, because any exception kills the cleanup and it will never recover.

wouterdb (changes requested):
Nive, minor remarks

src/inmanta/agent/forking_executor.py
```diff
@@ -682,3 +752,41 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
         children = [self.executor_map[child_id] for child_id in children_ids]
         await asyncio.gather(*(child.stop() for child in children))
         return children
+
+    def start_periodic_executor_cleanup(self) -> None:
+        async def cleanup_inactive_executors() -> None:
```
wouterdb, line 757: If the manager is stopped or stopping, you have to break out of this loop as well (or cancel it or both)

wouterdb (changes requested):

src/inmanta/agent/forking_executor.py
```diff
@@ -163,7 +169,40 @@ def connection_lost(self, exc: Exception | None) -> None:
...
 
 class ExecutorClient(FinalizingIPCClient[ExecutorContext], LogReceiver):
-    pass
+    def __init__(self, name: str):
+        super().__init__(name)
+
+        # Keeps track of when this client was active last
+        self.last_used_at: datetime.datetime = datetime.datetime.now().astimezone()
+
+    def get_idle_time(self) -> timedelta:
```
sanderr, line 178: IMO this should be the qualified `datetime.timedelta`. Not a strong requirement for me, I only mention it because we try to be more aware of our import conventions.

src/inmanta/agent/forking_executor.py
```diff
@@ -163,7 +169,40 @@ def connection_lost(self, exc: Exception | None) -> None:
...
+    @typing.overload
+    def call(
+        self, method: IPCMethod[ExecutorContext, ReturnType], has_reply: typing.Literal[True] = True
+    ) -> Future[ReturnType]: ...
+
+    @typing.overload
+    def call(self, method: IPCMethod[ExecutorContext, ReturnType], has_reply: typing.Literal[False]) -> None: ...
+
+    @typing.overload
+    def call(self, method: IPCMethod[ExecutorContext, ReturnType], has_reply: bool = True) -> Future[ReturnType] | None: ...
```
sanderr, line 190: Isn't this one overload too many?

src/inmanta/agent/forking_executor.py
```diff
@@ -517,6 +566,11 @@ def __init__(
 
         self._locks: inmanta.util.NamedLock = inmanta.util.NamedLock()
 
+        self.executor_retention_time = inmanta.agent.config.agent_executor_retention_time.get()
+        self.max_executors_per_agent = inmanta.agent.config.agent_executor_cap.get()
+
+        self.start_periodic_executor_cleanup()
```
wouterdb, line 572: I would move this to the start method

src/inmanta/agent/forking_executor.py
```diff
@@ -163,7 +169,40 @@ def connection_lost(self, exc: Exception | None) -> None:
...
+        response = super().call(method, has_reply)
+        assert response is None or isinstance(response, Future)
+        return response
+
+    def has_outstanding_calls(self) -> bool:
+        """Is this client still waiting for replies"""
+        return len(self.requests) > 0
+
+    def process_reply(self, frame: IPCReplyFrame) -> None:
+        self.last_used_at = datetime.datetime.now().astimezone()
```
sanderr, line 204: I have a slight preference to set this after processing the reply, because that is actually when it was last used imo (the last action was taken then).

src/inmanta/agent/forking_executor.py
```diff
@@ -710,13 +753,40 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
...
+            for _executor in self.executor_map.values():
+                if _executor.can_be_cleaned_up(self.executor_retention_time):
+                    LOGGER.debug(
+                        "Stopping executor %s because it "
+                        "was inactive for %d s, which is longer then the retention time of %d s.",
+                        _executor.executor_id.identity(),
+                        _executor.connection.get_idle_time().total_seconds(),
+                        self.executor_retention_time,
+                    )
+                    async with self._locks.get(_executor.executor_id.identity()):
```
sanderr, line 773: Shouldn't we check again inside the lock whether it can be cleaned up? This is an async call, so things may move while we wait for the lock.

src/inmanta/agent/forking_executor.py
```diff
@@ -710,13 +753,40 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
...
+                else:
+                    reschedule_interval = min(
+                        reschedule_interval,
+                        (
+                            now - _executor.connection.last_used_at + timedelta(seconds=self.executor_retention_time)
+                        ).total_seconds(),
+                    )
+
+            LOGGER.debug(f"Scheduling next cleanup_inactive_executors in {reschedule_interval} s.")
+            await asyncio.sleep(reschedule_interval)
```
sanderr, line 789: What if the setting changes? i.e. if it is set first to 1 day, then changed to 1 second, but we're already sleeping for a day, there will be a day's delay.

sanderr, line 789: Any reason to choose `sleep` over `asyncio.call_later`?

src/inmanta/agent/forking_executor.py
```diff
@@ -710,13 +753,40 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
...
+                            await _executor.stop()
+                        except Exception:
+                            LOGGER.debug(
+                                "Unexpected error during executor %s cleanup:", _executor.executor_id.identity(), exc_info=True
+                            )
+                else:
+                    reschedule_interval = min(
+                        reschedule_interval,
+                        (
+                            now - _executor.connection.last_used_at + timedelta(seconds=self.executor_retention_time)
```
sanderr, line 784: This will always be higher than `reschedule_interval`, won't it? i.e. `x >= x + now - past`. I think you want this:
```suggestion
                            timedelta(seconds=self.executor_retention_time) - (now - _executor.connection.last_used_at)
```

e.g. suppose retention time is one minute, and one of the executors has already been idle for 50 seconds (`min(60, 60 - 50)`), reschedule in 10 seconds rather than 60 seconds (`min(60, 60 + 50)`).

src/inmanta/agent/forking_executor.py
sanderr: Looks like you added it to the parent as well, so it doesn't seem like a mistake. Why do we need it?

src/inmanta/agent/forking_executor.py
```diff
@@ -710,13 +753,40 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
...
+                    LOGGER.debug(
+                        "Stopping executor %s because it "
+                        "was inactive for %d s, which is longer then the retention time of %d s.",
+                        _executor.executor_id.identity(),
+                        _executor.connection.get_idle_time().total_seconds(),
+                        self.executor_retention_time,
+                    )
+                    async with self._locks.get(_executor.executor_id.identity()):
+                        try:
+                            await _executor.stop()
```
sanderr, line 775: If there's any chance this might hang we should take into account the difference between `now` and the time when we reschedule at the end of the method. But I get the impression that this should always return quickly?

sanderr (changes requested):

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: Without this overload, mypy will complain that some overloads are missing because it doesn't realize that all cases are covered (see discussion [here](https://github.com/inmanta/inmanta-core/pull/7765#discussion_r1658262146))

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: Maybe we should update it before and after then ? Because if we only do it after we process the reply I think there is a chance that we `del self.requests[frame.id]` before updating `last_used_at` and end up cleaning up the executor wrongfully.

sanderr: I'm fine with before and after, but I don't think there's realistically a problem with doing it just after because we're in a synchronous context here. That's why I asked [this question](https://inmanta.slack.com/archives/CKRF0C8R3/p1719826294775269) on Slack BTW.

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: I think you're right, nice catch

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: I see :+1:

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: I think `call_later` takes a regular function as an argument for the callback (see [here](https://stackoverflow.com/a/34004028) ) but `cleanup_inactive_executors` is a coroutine so I was getting a type error: `TypeError: coroutines cannot be used with call_at()`.

This solution felt more elegant than wrapping this coroutine in a regular function ie 

```python
  asyncio.get_running_loop().call_later(reschedule_interval, lambda: asyncio.ensure_future(cleanup_inactive_executors()))
```

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: Oof nice catch

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: Added a watcher task that periodically checks the setting's value and reschedules the cleanup if it changed

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: Added mechanism to take into account the difference between start and end of cleanup

Hugo-Inmanta:
I added the `PoolManager` and `PoolMember` interfaces suggested [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1719825692066209?thread_ts=1719824573.908049&cid=CKRF0C8R3) as a first step, but I think further integration should go into another ticket ?

src/inmanta/agent/forking_executor.py
```diff
@@ -576,6 +678,18 @@ async def get_executor(
                         "Found stale executor for agent %s with id %s, waiting for close", agent_name, executor_id.identity()
                     )
                     await it.join(2.0)
+            n_executors_for_agent = len(self.agent_map[executor_id.agent_name])
+            if n_executors_for_agent >= self.max_executors_per_agent:
```
wouterdb, line 682: has to be strictly  larger than 1

src/inmanta/agent/forking_executor.py
```diff
@@ -682,3 +805,97 @@ async def stop_for_agent(self, agent_name: str) -> list[MPExecutor]:
         children = [self.executor_map[child_id] for child_id in children_ids]
         await asyncio.gather(*(child.stop() for child in children))
         return children
+
+    def start_executor_retention_time_watcher(self) -> None:
```
wouterdb, line 809: This value can not be updated

wouterdb (changes requested):

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: As discussed with @wouterdb, this is an agent setting i.e. not currently possible to be updated on the fly while the agent is running.

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: Added a validator to the setting

src/inmanta/agent/config.py
```diff
@@ -188,7 +189,7 @@ def is_executor_mode(value: str | AgentExecutorMode) -> AgentExecutorMode:
     3,
     "Maximum number of concurrent executors to keep per environment, per agent. If this limit is already reached "
     "when creating a new executor, the oldest one will be stopped first.",
-    is_int,
+    functools.partial(is_lower_bounded_int, 1),
```
wouterdb, line 192: We use the docstring of the function to generate documentation, so this won't look so pretty in the docs

src/inmanta/agent/forking_executor.py
```diff
@@ -517,6 +602,14 @@ def __init__(
 
         self._locks: inmanta.util.NamedLock = inmanta.util.NamedLock()
 
+        self.executor_retention_time = inmanta.agent.config.agent_executor_retention_time.get()
+        self.max_executors_per_agent = inmanta.agent.config.agent_executor_cap.get()
+
+        self.stopping: bool = False
+        # We keep a reference to the periodic cleanup task to prevent it
+        # from disappearing mid-execution https://docs.python.org/3.11/library/asyncio-task.html#creating-tasks
```
wouterdb, line 610: wow, nice!

src/inmanta/agent/forking_executor.py
```diff
@@ -658,16 +768,22 @@ def _child_closed(self, child_handle: MPExecutor) -> None:
...
             args=(child_conn, name, log_level, cli_log, inmanta.config.Config.config_as_dict()),
             name=f"agent.executor.{name}",
         )
         p.start()
         child_conn.close()
+
         return p, parent_conn
 
+    async def start(self) -> None:
+        self.cleanup_job = asyncio.create_task(self.cleanup_inactive_executors())
```
wouterdb, line 783: in the join method, you could `await self.cleanup_job` to make sure it is down.

src/inmanta/config.py
```diff
@@ -262,6 +262,14 @@ def is_list(value: str | list[str]) -> list[str]:
     return [] if value == "" else [x.strip() for x in value.split(",")]
 
 
+def is_lower_bounded_int(lower_bound: int, value: str | int) -> int:
+    """lower-bounded int"""
+    to_int = int(value)
+    if to_int < lower_bound:
+        raise ValueError(f"Value can not be lower than {lower_bound}")
+    return to_int
```
wouterdb, line 270: to make the docs prettier

```suggestion
def is_lower_bounded_int(lower_bound: int) -> Callable[[str | int],int]:
    """lower-bounded int factory"""
    def inner(value: str | int):
         to_int = int(value)
         if to_int < lower_bound:
              raise ValueError(f"Value can not be lower than {lower_bound}")
         return to_int
    inner.__doc__ = f"int < {lower_bound}"
    return inner 
```

wouterdb (changes requested):

src/inmanta/agent/forking_executor.py
```diff
@@ -517,6 +602,14 @@ def __init__(
 
         self._locks: inmanta.util.NamedLock = inmanta.util.NamedLock()
 
+        self.executor_retention_time = inmanta.agent.config.agent_executor_retention_time.get()
+        self.max_executors_per_agent = inmanta.agent.config.agent_executor_cap.get()
+
+        self.stopping: bool = False
```
wouterdb, line 608: I've already given more comment than I'm comfortable, so feel free to ignore this one: when used like this, I usually call this 'running' (and invert it) to make sure it doesn't only indicate 'stopping' but also 'stopped'

tests/forking_agent/test_executor.py
```diff
@@ -86,8 +86,8 @@ def set_custom_executor_policy():
     inmanta.agent.config.agent_executor_cap.set("2")
 
     old_retention_value = inmanta.agent.config.agent_executor_retention_time.get()
-    # Clean up executors after 2s of inactivity
-    inmanta.agent.config.agent_executor_retention_time.set("2")
+    # Clean up executors after 3s of inactivity
+    inmanta.agent.config.agent_executor_retention_time.set("3")
```
Hugo-Inmanta, line 90: Increased this a bit because it was causing intempestive cleanup in the test case:  the old executor is supposed to get cleaned up when the agent cap is reached, but with this setting too low, it would automatically get cleaned up before then.

wouterdb (approved):

sanderr (changes requested):

src/inmanta/agent/forking_executor.py
```diff
@@ -518,7 +537,24 @@ async def get_facts(self, resource: "inmanta.agent.executor.ResourceDetails") ->
         return await self.connection.call(FactsCommand(resource))
 
 
-class MPManager(executor.ExecutorManager[MPExecutor]):
+class PoolManager:
```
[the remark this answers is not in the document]
Hugo-Inmanta: Agreed, this is not useful right now, I'll remove it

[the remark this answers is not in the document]
sanderr: That's not entirely what I'm trying to say. I think the concept is very good, if it allows us to share behavior between both use cases. So when @wouterdb proposed it, I'm assuming that's what he had in mind? By removing it, we're back to square one, aren't we?

sanderr (commented):
My review is blocked on Wouter's response to [this discussion](https://github.com/inmanta/inmanta-core/pull/7765#discussion_r1664065991).

sanderr (approved):

PR #7784, opened 2024-06-28
Prevent reload

wouterdb:
# Description

Explicitly prevent reloads, as this still caused issues 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/forking_executor.py
+++ b/src/inmanta/agent/forking_executor.py
@@ -49,8 +49,6 @@ from setproctitle import setproctitle
 
 LOGGER = logging.getLogger(__name__)
 
-import asyncio
-
 
 class ExecutorContext:
     """The context object used by the executor to expose state to the incoming calls"""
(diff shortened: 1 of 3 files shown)


src/inmanta/agent/forking_executor.py
```diff
@@ -236,7 +236,8 @@ async def call(self, context: ExecutorContext) -> typing.Sequence[inmanta.loader
         for module_source in in_place:
             try:
                 await loop.run_in_executor(
-                    context.threadpool, functools.partial(loader._load_module, module_source.name, module_source.hash_value)
+                    context.threadpool,
+                    functools.partial(loader._load_module, module_source.name, module_source.hash_value, False),
```
arnaudsjs, line 240:
```suggestion
                    functools.partial(loader._load_module, module_source.name, module_source.hash_value, require_reload=False),
```

arnaudsjs (changes requested):

arnaudsjs (approved):

Issue #7759, opened 2024-06-21
Provide the actual excpetion that cause a resource load to fail in the deploy log


wouterdb:
Currently, we log 

```
Failed to load handler code or install handler code dependencies. Check the agent log for details.
```

It would be better to log the actual exception 

This implies adding the exception to the FailedResources

PR #7785, opened 2024-06-28
Propagate exception for failed resources

hlloreda:
# Description

Provide the exception that caused a resource load to fail in the deploy log

closes https://github.com/inmanta/inmanta-core/issues/7759

# Self Check:

- [x] Changelog entry


src/inmanta/agent/agent.py
```diff
@@ -874,37 +885,54 @@ async def prepare_resources(
...
                 loaded_resources.append(ResourceDetails(res))
 
                 state = const.ResourceState[res["status"]]
                 if state in const.UNDEPLOYABLE_STATES:
-                    undeployable[res["id"]] = state
+                    undeployable[res_id] = state
             else:
-                failed_resources.append(res["id"])
-                undeployable[res["id"]] = const.ResourceState.unavailable
+                logs.append(
```
wouterdb, line 907: Isn't this going to repeat the same message over and over again?

I.e. We duplicate it for every instance of the type. 

I would just log the exception once per type (as we only produce one resource_action_update with all log messages at once)

src/inmanta/agent/executor.py
```diff
@@ -45,7 +45,14 @@
...
 
-FailedResourcesSet: typing.TypeAlias = set[ResourceType]
+
+@dataclass(frozen=True)
+class FailedResource:
+    resource_type: ResourceType
+    exception: Exception
+
+
+FailedResourcesSet: typing.TypeAlias = set[FailedResource]
```
wouterdb, line 55: I wonder if this shouldn't be a dict. As it is now, keeping duplicates (same type, different exception) out is not trivial.

tests/agent_server/test_agent_code_loading.py
```diff
@@ -322,3 +324,98 @@ def test():
...
+    idx1 = log_index(
+        caplog,
+        "inmanta.agent.agent",
+        logging.ERROR,
+        "Failed to get source code for test::Test2 version=-1",
+    )
+
+    log_index(caplog, "inmanta.agent.agent", logging.ERROR, "Failed to install handler test::Test version=1", idx1)
+
+    log_index(caplog, "inmanta.agent.agent", logging.ERROR, "Failed to install handler test::Test2 version=1", idx1)
```
wouterdb, line 421: It would be interesting to also make assertions on the resource action log

wouterdb (changes requested):
Nice approach!

tests/agent_server/test_agent_code_loading.py
hlloreda: These assertions have been added to the other test

src/inmanta/agent/agent.py
```diff
@@ -1239,13 +1233,13 @@ async def get_code(
             - set of invalid resource_types (no handler code and/or invalid pip config)
         """
         if self._loader is None:
-            return [], set()
+            return [], dict()
```
wouterdb, line 1236:
```suggestion
            return [], {}
```

src/inmanta/agent/agent.py
```diff
@@ -1239,13 +1233,13 @@ async def get_code(
...
 
         # store it outside the loop, but only load when required
         pip_config: Optional[PipConfig] = None
 
         resource_install_specs: list[ResourceInstallSpec] = []
-        invalid_resource_types: executor.FailedResourcesSet = set()
+        invalid_resource: executor.FailedResources = dict()
```
wouterdb, line 1242:
```suggestion
        invalid_resource: executor.FailedResources = {}
```

tests/agent_server/test_agent_code_loading.py
```diff
@@ -375,17 +375,15 @@ async def get_version() -> int:
...
+    assert len(invalid_resources_1.keys()) == 3
+    for resource_type, exception in invalid_resources_1.items():
         assert (
-            "Failed to get source code for " + invalid_resource.resource_type + " version=-1, result={'message': 'Request or "
-            "referenced resource does not exist: The version of the code does not exist. "
-            + invalid_resource.resource_type
-            + ", -1'}"
-        ) == str(invalid_resource.exception)
+            "Failed to get source code for " + resource_type + " version=-1, result={'message': 'Request or "
+            "referenced resource does not exist: The version of the code does not exist. " + resource_type + ", -1'}"
```
wouterdb, line 385: What is the `-1`?

src/inmanta/agent/agent.py
```diff
@@ -887,31 +879,33 @@ async def prepare_resources(
...
-                        logging.ERROR,
-                        "%(res_id)s failed due to `%(error)s`.",
-                        res_id=res_id,
-                        error=str(invalid_resources_types[res_type]),
+                if res_type not in logged_resource_types:
+                    logged_resource_types.add(res_type)
+                    logs.append(
+                        data.LogLine.log(
+                            logging.ERROR,
+                            "This type of resource `%(res_type)s` failed due to `%(error)s`.",
```
wouterdb, line 904:
```suggestion
                            "All resources of `%(res_type)s` failed due to `%(error)s`.",
```

src/inmanta/agent/agent.py
```diff
@@ -1299,21 +1288,16 @@ async def get_code(
...
 
-        return resource_install_specs, invalid_resource_types
+        return resource_install_specs, invalid_resource
 
-    async def ensure_code(self, code: Collection[ResourceInstallSpec]) -> executor.FailedResourcesSet:
+    async def ensure_code(self, code: Collection[ResourceInstallSpec]) -> executor.FailedResources:
         """Ensure that the code for the given environment and version is loaded"""
 
-        failed_to_load: executor.FailedResourcesSet = set()
+        failed_to_load: executor.FailedResources = dict()
```
wouterdb, line 1300:
```suggestion
        failed_to_load: executor.FailedResources = {}
```

src/inmanta/agent/forking_executor.py
```diff
@@ -387,7 +387,7 @@ def __init__(
 
         # Set by init and parent class that const
         self.failed_resource_results: typing.Sequence[inmanta.loader.FailedModuleSource] = list()
-        self.failed_resource_types: executor.FailedResourcesSet = set()
+        self.failed_resources: executor.FailedResources = dict()
```
wouterdb, line 390:
```suggestion
        self.failed_resources: executor.FailedResources = {}
```

wouterdb (changes requested):

tests/agent_server/test_agent_code_loading.py
wouterdb: ah, this comes from  the server

wouterdb (approved):

arnaudsjs (changes requested):
Regarding the two test cases: Why are these tests specified in two different files? They clearly belong together and highly related logic in a very similar way.

tests/agent_server/test_server_agent.py
```diff
@@ -3887,3 +3899,211 @@ def test_deploy_response_matrix_invariants():
...
+
+    monkeypatch.undo()
+
+    idx1 = log_index(
+        caplog,
+        f"resource_action_logger.{environment}",
+        logging.ERROR,
+        "All resources `test::Resource` failed to load handler code or install handler code dependencies: "
+        "`Could not set up executor for agent1: Failed to "
+        "install handler `test` version=1`.",
```
arnaudsjs, line 3975: I would expect an error message + stack trace here. Because the stack trace is mostly the thing that allows you to determine the root cause of an issue.

arnaudsjs (changes requested):

src/inmanta/agent/agent.py
```diff
@@ -886,39 +893,53 @@ async def prepare_resources(
...
-        executor, invalid_resource_types = await self.setup_executor(model_version, resource_types)
+        executor, invalid_resources = await self.setup_executor(model_version, resource_types)
 
         loaded_resources: list[ResourceDetails] = []
-        failed_resources: list[ResourceVersionIdStr] = []
         undeployable: dict[ResourceVersionIdStr, const.ResourceState] = {}
 
+        logs: dict[str, set[data.LogLine]] = defaultdict(set)
+        failed_resource_ids: dict[str, set[str]] = defaultdict(set)
+        logged_resource_types = set()
```
sanderr, line 903: This one seems redundant to me, because it's basically a copy of `logs.keys()`. But if you prefer it style-wise, I'm fine with it.

sanderr (changes requested):
Nice!

src/inmanta/agent/agent.py
```diff
@@ -886,39 +893,53 @@ async def prepare_resources(
...
-                resource_ids=failed_resources,
-                action_id=uuid.uuid4(),
-                action=action,
-                started=started,
-                finished=datetime.datetime.now().astimezone(),
-                messages=[log],
-                status=const.ResourceState.unavailable,
-            )
+        if len(failed_resource_ids) > 0:
+            for resource_type, error_messages in logs.items():
```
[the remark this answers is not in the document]
sanderr: Come to think of it, `failed_resources` is completely redundant now, isn't it? It's just a copy of `invalid_resources` now?

tests/agent_server/test_server_agent.py
```diff
@@ -3887,3 +3892,221 @@ def test_deploy_response_matrix_invariants():
...
+    )
+
+    idx1 = log_index(
+        caplog,
+        f"resource_action_logger.{environment}",
+        logging.ERROR,
+        expected_error_message_without_tb,
+    )
+    # A traceback should be present in the logs, so the lengths of these logs should be longer
+    assert len(caplog.record_tuples[idx1][2]) > len(expected_error_message_without_tb)
```
arnaudsjs, line 3975: The new code performs some magic with the stack traces. The tests should verify that the resulting stack trace is correct. This assertion gives very minimal guarantees about that. We should improve the assertion to verify the actual stack trace.

arnaudsjs (changes requested):

tests/agent_server/test_server_agent.py
```diff
@@ -3887,3 +3893,225 @@ def test_deploy_response_matrix_invariants():
...
+    expected_error_message = f"""multiple resources: All resources of type `test::Resource` failed to load handler code or \
+install handler code dependencies: `Could not set up executor for agent1: Failed to install handler `test` version=1`
+  File "{module_folder}/src/inmanta/agent/agent.py", line 854, in setup_executor
+    current_executor = await self.executor_manager.get_executor(self.name, self.uri, code)
+                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  File "{module_folder}/src/inmanta/agent/in_process_executor.py", line 502, in get_executor
+    out.failed_resources = await self.process.ensure_code(code)
+                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  File "{module_folder}/tests/agent_server/test_server_agent.py", line 3953, in ensure_code
+"""
```
arnaudsjs, line 3974: This is a bit too strict to be used in an assertion, because it contains line numbers. As such, the test case will break every time the line number changes. It's sufficient to verify that the exception comes from the location where it was raised. So the following pattern should match: `File "{module_folder}/tests/agent_server/test_server_agent.py", line \d+, in ensure_code`. All other parts of the stack trace are not relevant.

arnaudsjs (changes requested):

arnaudsjs (approved):

src/inmanta/agent/agent.py
```diff
@@ -900,9 +899,8 @@ async def prepare_resources(
         loaded_resources: list[ResourceDetails] = []
         undeployable: dict[ResourceVersionIdStr, const.ResourceState] = {}
 
-        logs: dict[str, set[data.LogLine]] = defaultdict(set)
-        failed_resource_ids: dict[str, set[str]] = defaultdict(set)
-        logged_resource_types = set()
+        # {resource_type -> (set{resource_ids}, LogLine)}
```
sanderr, line 902: `typing.NewType` or plain type aliases can also be useful here, but I have no objections to this.

src/inmanta/agent/agent.py
```diff
@@ -914,32 +912,33 @@ async def prepare_resources(
                 if state in const.UNDEPLOYABLE_STATES:
                     undeployable[res_id] = state
             else:
-                if res_type not in logged_resource_types:
-                    logged_resource_types.add(res_type)
-                    logs[res_type].add(
+                if res_type not in failed_resources:
+                    failed_resources[res_type] = (
+                        {res_id},
```
sanderr, line 917: Slightly simpler than the if-else is probably to use the `if` only for initialization (to `set()`), and always execute the `add`.

Just a suggestion, feel free to leave as is.

sanderr (approved):

PR #7789, opened 2024-06-29
Issue/dict path improvs

edvgui:
# Description

Add one more functionality to the dict path library: the possibility to emit all the dict path expression that would match one of the elements matched by a wild dict path for a given container. (see test and docstrings for example)

**Context**: When unwrapping an lsm service, in an update state, we often want to check which embedded entities have been created, updated, or removed.  For each service we always need to implement this logic custom, which can be tedious if the embedded entity we care about is deep in the service attributes tree.  One thing that would make it a lot easier was the possibility to say "for all the elements matching this [wild] dict path, tell me which ones have been added, updated, removed".  But this is also very inconvenient to calculate because the notion of identity of each layer of embedded entity needs to be extracted from the dict path, which is not really what the library is meant to be used for.  This new feature allows an easier approach: resolving all the paths that point to embedded entities in the previous set of attributes and in the current one, and just making the diff on the list of dict paths.

```python
@inmanta.plugins.plugin()
def get_removed_entities(service: "lsm::ServiceEntity", wild_path: "string") -> "dict[]":
    previous_attributes = ...
    current_attributes = ...
    
    parsed_wild_path = dict_path.to_wild_path(wild_path)
    removed_entitites = (
        {str(p) for p in parsed_wild_path.resolve_wild_cards(previous_attributes)}
        - {str(p) for p in parsed_wild_path.resolve_wild_cards(current_attributes)}
    )

    return [
        dict_path.to_path(path).get_element(previous_attributes)
        for path in removed_entities
    ]
```

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/extend-dict-path-library.yml
+++ b/changelogs/unreleased/extend-dict-path-library.yml
@@ -1,0 +1,4 @@
+---
+description: Add get_paths method to dict path library.
+change-type: minor
+destination-branches: [master, iso7]
--- a/changelogs/unreleased/extend-dict-path-library.yml
+++ b/changelogs/unreleased/extend-dict-path-library.yml
@@ -1,4 +1,6 @@
 ---
-description: Add get_paths method to dict path library.
+description: Add resolve_wild_cards method to dict path library.
 change-type: minor
 destination-branches: [master, iso7]
+sections:
+  minor-improvement: Extend dict_path library, allow to resolve the wild cards of a wild dict path for a specific container.
--- a/src/inmanta/util/dict_path.py
+++ b/src/inmanta/util/dict_path.py
@@ -228,7 +228,7 @@ class WildDictPath(abc.ABC):
     def get_paths(self, container: object) -> list["DictPath"]:
         """
         Get all the dict paths that this wild dict path contains, in the given
-        containers.  That is, the set of dict path that would resolve the same
+        container.  That is, the set of dict path that would resolve the same
         set of elements as this wild dict path, when searching in the given
         container.
 
--- a/src/inmanta/util/dict_path.py
+++ b/src/inmanta/util/dict_path.py
@@ -225,12 +225,16 @@ class WildDictPath(abc.ABC):
         """
 
     @abc.abstractmethod
-    def get_paths(self, container: object) -> Sequence["DictPath"]:
+    def resolve_wild_cards(self: TWDP, container: object) -> Sequence[TWDP]:
         """
-        Get all the dict paths that this wild dict path contains, in the given
-        container.  That is, the set of dict path that would resolve the same
-        set of elements as this wild dict path, when searching in the given
-        container.
+        Resolve all the wild cards contained in this wild dict path, based on the
+        given container.  For each possible set of values that can replace the wild
+        cards and still resolve a subset of the elements matched by this expression,
+        return a wild dict path expression without the use of any wild card.
+
+        The set of all elements that can be resolved by all returned wild dict paths
+        for the given container, should be equal to the set of all elements matched by
+        this wild dict path expression for the given container.
 
         :param container: the container to search in
         """
@@ -324,38 +328,38 @@ class WildInDict(WildDictPath):
         else:
             raise ContainerStructureException(f"{container} is not a Dict")
 
-    def get_paths(self, container: object) -> Sequence["InDict"]:
+    def resolve_wild_cards(self, container: object) -> Sequence["WildInDict"]:
         """
-        Get one InDict by key matching the wild key of this object.
+        Get one WildInDict by key matching the wild key of this object.
 
         ..code-block:: python
 
-            assert WildInDict("a").get_paths(
+            assert WildInDict("a").resolve_wild_cards(
                 {
                     "a": "b",
                     "aa": "bb",
                     "c": "d",
                 }
             ) == [
-                InDict("a"),
+                WildInDict("a"),
             ]
 
-            assert WildInDict("*").get_paths(
+            assert WildInDict("*").resolve_wild_cards(
                 {
                     "a": "b",
                     "aa": "bb",
                     "c": "d",
                 }
             ) == [
-                InDict("a"),
-                InDict("aa"),
-                InDict("c"),
+                WildInDict("a"),
+                WildInDict("aa"),
+                WildInDict("c"),
             ]
 
         """
         if self._validate_container(container):
             try:
-                return [InDict(str(key)) for key in container if self.key.matches(key)]
+                return [WildInDict(str(key)) for key in container if self.key.matches(key)]
             except KeyError:
                 return []
         else:
@@ -538,14 +542,14 @@ class WildKeyedList(WildDictPath):
             and all(any(key.matches(k) and value.matches(v) for k, v in dct.items()) for key, value in self.key_value_pairs)
         ]
 
-    def get_paths(self, container: object) -> Sequence["KeyedList"]:
+    def resolve_wild_cards(self, container: object) -> Sequence["WildKeyedList"]:
         """
-        Get one KeyedList by set of matching key value pairs that can be found
+        Get one WildKeyedList by set of matching key value pairs that can be found
         in the given container at the relation specified in this object.
 
         ..code-block:: python
 
-            assert WildKeyedList("relation","key_attribute","*").get_paths(
+            assert WildKeyedList("relation","key_attribute","*").resolve_wild_cards(
                 {
                     "relation":[
                         {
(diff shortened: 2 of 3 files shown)


src/inmanta/util/dict_path.py
```diff
@@ -593,6 +682,24 @@ def get_elements(self, container: object) -> list[object]:
...
+
+        if len(self.expanded_path) == 1:
+            # The path is equivalent to its unique element
+            return self.expanded_path[0].get_paths(container)
+
+        return [
+            path + next_part
+            for path in self.expanded_path[0].get_paths(container)
+            for next_part in WildComposedPath(path=self.expanded_path[1:]).get_paths(path.get_element(container))
+        ]
```
edvgui, line 795: Could it be done without recursion? I couldn't think of an algorithm

changelogs/unreleased/extend-dict-path-library.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: Add get_paths method to dict path library.
```
edvgui, line 2: Should there be an entry in the public changelog? (minor-improvement section)

src/inmanta/util/dict_path.py
```diff
@@ -223,6 +224,17 @@ def get_elements(self, container: object) -> list[object]:
...
+    @abc.abstractmethod
+    def get_paths(self, container: object) -> list["DictPath"]:
+        """
+        Get all the dict paths that this wild dict path contains, in the given
+        container.  That is, the set of dict path that would resolve the same
+        set of elements as this wild dict path, when searching in the given
+        container.
+
+        :param container: the container to search in
+        """
```
edvgui, line 236: This is probably not clear enough, but I am really struggling with the explanation and the naming of such a feature.  Any suggestion?

src/inmanta/util/dict_path.py
```diff
@@ -489,6 +538,139 @@ def get_elements(self, container: object) -> list[object]:
...
+                        {
+                            "key_attribute":"other_value",
+                        },
+                        {
+                            "other_key_attribute":"other_value",
+                        },
+                    [
+                }
+            ) == [
+                KeyedList("relation", [("key_attribute","key_value")]),
```
wouterdb, line 624: I don't see how this works with the described usecase? 

I expect, from the explanation, that this would be used to find a difference. 

let say we go from the set you have now to one that adds 
```python
"relation":[
                        {
                            "key_attribute":"key_value",
                            "other_attribute":"other_value",
                        },
                        {
                            "key_attribute":"other_value",
                        },
                        {
                            "other_key_attribute":"other_value",
                        },
                        {
                           "key_attribute":"other_value",
                           "other_key_attribute":"other_value",
                        },
                    [
                        
```

We try to detect change by doing 

```python
  removed_entitites = (
        {str(p) for p in parsed_wild_path.get_paths(previous_attributes)}
        - {str(p) for p in parsed_wild_path.get_paths(current_attributes)}
    )
```
Which is empty 


i.e. this filter is much weaker then expected, in that it doesn't identify any individual members. 

I think this is because the interface definition is to weak/wide for the intended use case.

src/inmanta/util/dict_path.py
```diff
@@ -223,6 +224,17 @@ def get_elements(self, container: object) -> list[object]:
         :param container: the container to search in
         """
 
+    @abc.abstractmethod
+    def get_paths(self, container: object) -> Sequence["DictPath"]:
+        """
+        Get all the dict paths that this wild dict path contains, in the given
+        container.  That is, the set of dict path that would resolve the same
+        set of elements as this wild dict path, when searching in the given
+        container.
```
wouterdb, line 233: `when searching in the given container.` is a very weak constraint. 

i.e. when the container has 0 or 1 element, an empty filter is always correct. 

The intended usage is can't be supported under this definition.

I think you have to assume the path describes an identity and that identity is made concrete based on the instances present.

src/inmanta/util/dict_path.py
```diff
@@ -593,6 +775,25 @@ def get_elements(self, container: object) -> list[object]:
...
+        if container is None:
+            raise IndexError("Can not get anything from None")
+
+        if len(self.expanded_path) == 0:
+            # No need to dig further
+            return []
+
+        if len(self.expanded_path) == 1:
+            # The path is equivalent to its unique element, but we have to wrap
+            # each of the possible path into a ComposedPath object to be consistent
```
wouterdb, line 788: Why?

wouterdb: I think it is because of the recursion?

src/inmanta/util/dict_path.py
```diff
@@ -620,6 +821,12 @@ def get_elements(self, container: object) -> list[object]:
         else:
             raise ContainerStructureException(f"{container} is not a Dict")
 
+    def get_paths(self, container: object) -> Sequence["NullPath"]:
+        if self._validate_container(container):
+            return [NullPath()]
```
wouterdb, line 826: could it not return `[]` ?

wouterdb (changes requested):

wouterdb:
Cool stuff

src/inmanta/util/dict_path.py
edvgui: > I don't see how this works with the described usecase?

It doesn't, I couldn't think of a use case where we would want to use the wild card on the key.  But I didn't feel like ignoring this usage of the wild keyed list either.  So I added support for it and illustrated it in the example.

To refer to your other comment:
> I think you have to assume the path describes an identity and that identity is made concrete based on the instances present.

I completely agree with that, in the use case I have it is the only thing that makes sense, but it seemed easier to me to implement the full, and less strict functionality, leaving the responsibility to the developer of making a good usage of it.

src/inmanta/util/dict_path.py
edvgui: I am not sure.  An empty list here would mean "there is no path matching this filter in the give container", but here there is

src/inmanta/util/dict_path.py
edvgui: Another way of seeing it is: the context I provided above is there to explain why I want this feature but the feature itself is not aware of this context, or lsm, and any notion of identity. Dict path makes it easy to interact with json-like structure with an identity mechanism defined, but doesn't enforce anything like that, so I don't think we should start with it now

src/inmanta/util/dict_path.py
```diff
@@ -656,12 +860,6 @@ def get_element(self, container: object, construct: bool = False) -> object:
         :raises KeyError: if the element is not found or if more than one occurrence was found.
         """
 
-    def get_elements(self, container: object) -> list[object]:
-        try:
-            return [self.get_element(container, False)]
-        except LookupError:
-            return []
-
```
edvgui, line 664: Should this still be removed?

src/inmanta/util/dict_path.py
```diff
@@ -593,6 +771,26 @@ def get_elements(self, container: object) -> list[object]:
 
         return containers
 
+    def resolve_wild_cards(self, container: object) -> Sequence["WildComposedPath"]:
```
wouterdb, line 774: This could be considered a tree if I understand correctly, where very wildcard creates/can create a branching point?


(not actionable, just for me to understand)

edvgui: I would rather say that every wild path is a branching point, but yes, it can be seen as a tree

src/inmanta/util/dict_path.py
```diff
@@ -489,6 +542,131 @@ def get_elements(self, container: object) -> list[object]:
...
+                        },
+                    [
+                }
+            ) == [
+                WildKeyedList("relation", [("other_attribute","other_value")]),
+                WildKeyedList("relation", [("key_attribute","other_value")]),
+                WildKeyedList("relation", [("other_key_attribute","other_value")]),
+            ]
+
+            assert WildKeyedList("relation",[("*","*")]).resolve_wild_cards(
```
wouterdb, line 612: What about

```
items= WildKeyedList("relation", [("*", "other_value")]).resolve_wild_cards(
        {
            "relation": [
                {
                    "key_attribute": "key_value",
                    "other_attribute": "other_value",
                },
                {
                    "key_attribute": "other_value",
                    "key_attribute_2": "other_value",
                },
                {
                    "other_key_attribute": "other_value",
                }
                ]
        }
    )
```

edvgui: It will give 
```
[
    WildKeyedList("relation", [("other_attribute", "other_value"]),
    WildKeyedList("relation", [("key_attribute", "other_value"]),
    WildKeyedList("relation", [("key_attribute_2", "other_value"]),
    WildKeyedList("relation", [("other_key_attribute", "other_value"]),
]
```

edvgui: We can have more dict path than there are matching elements (and so more than one dict path by matching element).  Do you think that is an issue?

edvgui: And also, we can have more than one matching element by dict path (as the identity/uniqueness is not checked)

src/inmanta/util/dict_path.py
edvgui: Not gonna risk it the day before the release (I don't want the full commit to get reverted)

wouterdb (approved):

PR #7791, opened 2024-07-01
Fix plugin call re-execution

### LINKED ISSUE #7792: Compiler stuck on plugin call
For this model

```
entity Test:

end

entity Child:
    string name
end

Test.child [0:] -- Child

Test(
    child=std::key_sort([Child(name="x")], "name")
)

implement Child using std::none
```

The compiler never terminates

wouterdb:
# Description

Fixes #7792 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/ast/statements/call.py
+++ b/src/inmanta/ast/statements/call.py
@@ -19,7 +19,7 @@
 import logging
 from collections import abc
 from itertools import chain
-from typing import Optional, Sequence, Tuple
+from typing import Optional
 
 import inmanta.ast.type as InmantaType
 import inmanta.execute.dataflow as dataflow
@@ -92,7 +92,7 @@ class FunctionCall(ReferenceStatement):
         else:
             raise RuntimeException(self, "Can not call '%s', can only call plugin or primitive type cast" % self.name)
 
-    def requires_emit(self, resolver, queue):
+    def requires_emit(self, resolver: Resolver, queue: QueueScheduler) -> dict[object, VariableABC[object]]:
         requires: dict[object, VariableABC] = self._requires_emit_promises(resolver, queue)
         sub = ReferenceStatement.requires_emit(self, resolver, queue)
         # add lazy vars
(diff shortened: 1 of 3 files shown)


src/inmanta/ast/statements/call.py
```diff
@@ -114,17 +114,21 @@ def execute_direct(self, requires: abc.Mapping[str, object]) -> object:
                 kwargs[k] = v
         return self.function.call_direct(arguments, kwargs)
 
-    def resume(self, requires, resolver, queue, result):
-        """
-        Evaluate this statement.
-        """
+    def execute_args(self, requires:  abc.Mapping[str, object], resolver: Resolver, queue: QueueScheduler) -> Tuple[Sequence[object], dict[str, object]]:
```
sanderr, line 117:
```suggestion
    def execute_args(self, requires:  abc.Mapping[str, object], resolver: Resolver, queue: QueueScheduler) -> tuple[list[object], dict[str, object]]:
```

src/inmanta/ast/statements/call.py
```diff
@@ -277,11 +281,24 @@ def __init__(self, queue_scheduler, resolver, result: ResultVariable, requires,
         for r in requires.values():
             self.waitfor(r)
         self.ready(self)
+        # resolved args and kwargs
+        self.args = None
+        self.kwargs = None
```
sanderr, line 286: Could you annotate these?

src/inmanta/ast/statements/call.py
```diff
@@ -277,11 +281,24 @@ def __init__(self, queue_scheduler, resolver, result: ResultVariable, requires,
...
+            requires = {k: v.get_value() for (k, v) in self.base_requires.items()}
+            try:
+                self.args, self.kwargs = self.function.execute_args(requires, self.resolver, self.queue)
+            except RuntimeException as e:
+                e.set_statement(self.function)
+                raise e
+
         try:
-            self.function.resume(requires, self.resolver, self.queue, self.result)
+            self.function.execute_call(self.args, self.kwargs, self.resolver, self.queue, self.result)
```
sanderr, line 301: Are we confident that the `FunctionUnit` object gets garbage collected or should we set `self.args` and `self.kwargs` back to `None`?

sanderr (changes requested):
Tiny comments, otherwise looks good.

sanderr (approved):

PR #7810, opened 2024-07-02
Issue/executor hash

wouterdb:
# Description

Make blueprint hash resilient vs duplicates

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -174,8 +174,8 @@ class ExecutorBlueprint(EnvBlueprint):
             blueprint_dict = {
                 "pip_config": self.pip_config.dict(),
                 "requirements": self.requirements,
-                # Use the hash values of the sources, sorted to ensure consistent ordering
-                "sources": sorted(source.hash_value for source in self.sources),
+                # Use the hash values and name to create a stable identity
+                "sources": [[source.hash_value, source.name, source.is_byte_code] for source in self.sources],
             }
 
             # Serialize the extended blueprint dictionary to a JSON string, ensuring consistent ordering
@@ -195,9 +195,9 @@ class ExecutorBlueprint(EnvBlueprint):
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, ExecutorBlueprint):
             return False
-        return (self.pip_config, set(self.requirements), self.sources) == (
+        return (self.pip_config, self.requirements, self.sources) == (
             other.pip_config,
-            set(other.requirements),
+            other.requirements,
             other.sources,
         )
 
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -91,7 +91,7 @@ class EnvBlueprint:
     requirements: Sequence[str]
     _hash_cache: Optional[str] = dataclasses.field(default=None, init=False, repr=False)
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         # remove duplicates and make uniform
         self.requirements = sorted(list(set(self.requirements)))
 
@@ -136,7 +136,7 @@ class ExecutorBlueprint(EnvBlueprint):
     sources: Sequence[ModuleSource]
     _hash_cache: Optional[str] = dataclasses.field(default=None, init=False, repr=False)
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         super().__post_init__()
         # remove duplicates and make uniform
         self.sources = sorted(list(set(self.sources)))
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -91,6 +91,10 @@ class EnvBlueprint:
     requirements: Sequence[str]
     _hash_cache: Optional[str] = dataclasses.field(default=None, init=False, repr=False)
 
+    def __post_init__(self):
+        # remove duplicates and make uniform
+        self.requirements = sorted(list(set(self.requirements)))
+
     def blueprint_hash(self) -> str:
         """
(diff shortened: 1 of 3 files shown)


src/inmanta/agent/executor.py
```diff
@@ -91,6 +91,10 @@ class EnvBlueprint:
     requirements: Sequence[str]
     _hash_cache: Optional[str] = dataclasses.field(default=None, init=False, repr=False)
 
+    def __post_init__(self) -> None:
+        # remove duplicates and make uniform
+        self.requirements = sorted(list(set(self.requirements)))
```
sanderr, line 96: No need for the `list` in between I think.
```suggestion
        self.requirements = sorted(set(self.requirements))
```

src/inmanta/agent/executor.py
```diff
@@ -132,6 +136,11 @@ class ExecutorBlueprint(EnvBlueprint):
     sources: Sequence[ModuleSource]
     _hash_cache: Optional[str] = dataclasses.field(default=None, init=False, repr=False)
 
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        # remove duplicates and make uniform
+        self.sources = sorted(list(set(self.sources)))
```
sanderr, line 142:
```suggestion
        self.sources = sorted(set(self.sources))
```

src/inmanta/agent/executor.py
```diff
@@ -164,9 +173,9 @@ def blueprint_hash(self) -> str:
         if self._hash_cache is None:
             blueprint_dict = {
                 "pip_config": self.pip_config.dict(),
-                "requirements": sorted(self.requirements),
+                "requirements": self.requirements,
                 # Use the hash values of the sources, sorted to ensure consistent ordering
-                "sources": sorted(source.hash_value for source in self.sources),
+                "sources": sorted(list(set(source.hash_value for source in self.sources))),
```
sanderr, line 178: I'm not convinced we want to remove duplicates here. We might end up with different executors that have the same hash, e.g. if one of them contains an additional empty `__init__.py` file somewhere.

src/inmanta/agent/executor.py
```diff
@@ -186,10 +195,10 @@ def to_env_blueprint(self) -> EnvBlueprint:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, ExecutorBlueprint):
             return False
-        return (self.pip_config, set(self.requirements), sorted(self.sources)) == (
+        return (self.pip_config, set(self.requirements), sorted(list(set(self.sources)))) == (
```
sanderr, line 198: Shouldn't we check simple equality now that we sort these at construction? i.e. `@dataclass(frozen=True, eq=True)`? May have to use `__setattr__` if you go for frozen though: https://docs.python.org/3/library/dataclasses.html#frozen-instances.

src/inmanta/agent/executor.py
```diff
@@ -91,6 +91,10 @@ class EnvBlueprint:
     requirements: Sequence[str]
```
sanderr, line 91: It feels a bit off to me that we still require users to pass a `Sequence` rather than a `Set` or `Collection`. This would require a different input type vs output/internal type, but afaik `dataclasses` does not support that. So there's probably nothing to be done about it (apart from dropping dataclass but that may be worse).

sanderr (changes requested):

src/inmanta/agent/executor.py
wouterdb: I'm not super happy with it either and the fact we can't make it frozen and use `__post_init__` is unfortunate as well.

src/inmanta/agent/executor.py
```diff
@@ -175,7 +175,7 @@ def blueprint_hash(self) -> str:
                 "pip_config": self.pip_config.dict(),
                 "requirements": self.requirements,
                 # Use the hash values of the sources, sorted to ensure consistent ordering
-                "sources": sorted(list(set(source.hash_value for source in self.sources))),
+                "sources": sorted(source.hash_value for source in self.sources),
```
sanderr, line 178: Actually, I think we don't even want to sort, because again we would have cases where we get the same hash for different executors (e.g. file `a.py` renamed to `z.py` -> `sources` order is different from `source.hash_value` order).

As an extension to this, file rename `a0.py` to `a1.py` would still suffer from this behavior, because we do not include module/file names in the blueprint hash at all. But that's not new in this PR.

src/inmanta/agent/executor.py
```diff
@@ -195,10 +195,10 @@ def to_env_blueprint(self) -> EnvBlueprint:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, ExecutorBlueprint):
             return False
-        return (self.pip_config, set(self.requirements), sorted(list(set(self.sources)))) == (
+        return (self.pip_config, set(self.requirements), self.sources) == (
             other.pip_config,
             set(other.requirements),
```
sanderr, line 200: Why keep this `set()` if we already do it in the post init? Why not just `dataclasses.__eq__`?

sanderr (changes requested):

src/inmanta/agent/executor.py
wouterdb: But that means hashing the file name into it as well?

src/inmanta/agent/executor.py
wouterdb: `_hash_cache` would mess it out

sanderr: FYI, `dataclasses.field` has a `compare` boolean: https://docs.python.org/3/library/dataclasses.html#dataclasses.field

sanderr (approved):

Issue #7777, opened 2024-06-26
Prevent fact refresh on resources that are undeployable


wouterdb:
We currently push out fact refreshes to the agent for unknown resources, this can drown out the actual deploy

PR #7835, opened 2024-07-08
Issue/7777 prevent fact refresh on resources that are undeployable

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/7777

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/agentmanager.py
+++ b/src/inmanta/server/agentmanager.py
@@ -858,7 +858,7 @@ class AgentManager(ServerSlice, SessionListener):
                     "Ignore fact request for %s, resource is in an undeployable state.",
                     resource_id,
                 )
-                return 404, {"message": "The resource is in an undeployable state."}
+                return 503, {"message": "The resource is in an undeployable state."}
 
             rid: Id = Id.parse_id(res.resource_version_id)
             version: int = rid.version
--- a/src/inmanta/server/agentmanager.py
+++ b/src/inmanta/server/agentmanager.py
@@ -37,7 +37,7 @@ from inmanta import config as global_config
 from inmanta import const, data
 from inmanta.agent import config as agent_cfg
 from inmanta.config import Config
-from inmanta.const import AgentAction, AgentStatus, UNDEPLOYABLE_NAMES
+from inmanta.const import UNDEPLOYABLE_NAMES, AgentAction, AgentStatus
 from inmanta.data import APILIMIT, InvalidSort, model
 from inmanta.data.model import ResourceIdStr
 from inmanta.protocol import encode_token, handle, methods, methods_v2
--- a/src/inmanta/server/agentmanager.py
+++ b/src/inmanta/server/agentmanager.py
@@ -37,7 +37,7 @@ from inmanta import config as global_config
 from inmanta import const, data
 from inmanta.agent import config as agent_cfg
 from inmanta.config import Config
-from inmanta.const import AgentAction, AgentStatus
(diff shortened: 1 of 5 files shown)


tests/agent_server/test_facts.py
```diff
@@ -226,8 +226,9 @@ async def test_get_facts_extended(server, client, agent, clienthelper, resource_
     ]
 
     resource_states = {
-        "test::Fact[agent1,key=key4],v=%d" % version: const.ResourceState.undefined,
-        "test::Fact[agent1,key=key5],v=%d" % version: const.ResourceState.undefined,
```
Hugo-Inmanta, line 230: I think including the resource versions in here was a bug. The `resource_state` arg of `put_version()` is typed as `dict[ResourceIdStr, ...]` And this check was failed because of it: https://github.com/inmanta/inmanta-core/blob/48a780475ae561c988deefc0ba667b919ca98793/src/inmanta/server/services/orchestrationservice.py#L573

tests/agent_server/test_facts.py
```diff
@@ -270,28 +271,11 @@ async def get_fact(rid, result_code=200, limit=10, lower_limit=2):
...
-                log_sequence = log_sequence.contains("inmanta.agent.agent.agent1", logging.ERROR, "Unable to retrieve fact")
-            log_sequence.no_more_errors()
-        except AssertionError:
-            return False
-        else:
-            return True
-
-    # The get_parameter() API calls from the server to the agent are executed asynchronously with respect the
-    # get_param() API calls done from the test case to the server. Here we wait until all log records are available.
-    await retry_limited(wait_until_log_records_are_available, timeout=10)
```
Hugo-Inmanta, line 294: I think this is not needed anymore since we return earlier for undeployable resources

src/inmanta/server/agentmanager.py
```diff
@@ -853,6 +853,13 @@ async def request_parameter(self, env_id: uuid.UUID, resource_id: ResourceIdStr)
             if res is None:
                 return 404, {"message": "The resource has no recent version."}
 
+            if res.status in UNDEPLOYABLE_NAMES:
+                LOGGER.debug(
+                    "Ignore fact request for %s, resource is in an undeployable state.",
+                    resource_id,
+                )
+                return 404, {"message": "The resource is in an undeployable state."}
```
sanderr, line 861: I don't think 404 is appropriate here. Perhaps 405? But I'm honestly not very familiar with HTTP error codes so someone else might know better.

tests/agent_server/test_facts.py
sanderr: I think you're right.

sanderr (changes requested):

src/inmanta/server/agentmanager.py
Hugo-Inmanta: 405 seems more appropriate :+1:

wouterdb (approved):

src/inmanta/server/agentmanager.py
wouterdb: I don't think 405 is appropriate,because that seems to imply it would work for e.g. a  a PUT on a different URL. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405

also 

```
The server must generate an [Allow](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow) header field in a 405 status code response. The field must contain a list of methods that the target resource currently supports.
```

503 Service Unavailable is most appropriate I think

sanderr (approved):

Issue #7696, opened 2024-05-31
Fix agent cache


wouterdb:
# Problem

1. The agent cache is now version based, this is complicated for end-users and becomes difficult in the new agent design
2. Time based expiry is broken: https://github.com/inmanta/inmanta-core/issues/833

# Proposed solution

1. we keep version open for a fixed time X (e.g. 1 minute after last use ended)
2. we generate a 'tick' to the cache regularly (e.g. 1s) to make it run finalizes), on a threadpool  (taking care not to race actual usage.

sanderr:
Discussed in planning meeting. Main difficulty that needs to be addressed: discarding an entry from the cache means calling a custom finalizer, which is not async, therefore has to be run on a thread. It must only be called when the agent is not actively using the thing cached.

sanderr:
See status [on Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1722611435774309)

PR #7846, opened 2024-07-10
Issue/7696 fix agent cache

Hugo-Inmanta:
# Description

part 1 of https://github.com/inmanta/inmanta-core/issues/7696

- externalize cleanup: cache is now periodically cleaned up by the `cleanup_stale_cache_entries` job.
- convert `timerqueue` to a heap queue

part 2 [PR](https://github.com/inmanta/inmanta-core/pull/7850)
- converts version-based caching to time based

# Self Check:

- [x] Changelog entry


src/inmanta/agent/forking_executor.py
```diff
@@ -287,8 +287,21 @@ async def call(self, context: ExecutorContext) -> typing.Sequence[inmanta.loader
                 logger.info("Failed to load sources: %s", module_source, exc_info=True)
                 failed.append(module_source)
 
+        self.periodic_cache_cleanup_job = asyncio.create_task(self.cleanup_stale_cache_entries(context))
+
         return failed
 
+    async def cleanup_stale_cache_entries(self, context: ExecutorContext) -> None:
```
wouterdb, line 294: I would expect this code to be in the local executor, as it is the one actually contains it

wouterdb (changes requested):

src/inmanta/agent/cache.py
```diff
@@ -40,12 +40,21 @@ def __init__(self, timeout: int = 24 * 3600, version: int = 0) -> None:
...
+        value: Any,
+        call_on_delete: Optional[Callable[[Any], None]],
+        freshness_period: float,
+    ) -> None:
         self.key = key
         self.scope = scope
         self.value = value
         self.time: float = time.time() + scope.timeout
         self.call_on_delete = call_on_delete
+        self.last_used_at = datetime.datetime.now().astimezone()
```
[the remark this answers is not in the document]
Hugo-Inmanta: Changed this to:
- the cache item will mark itself as stale after the given `freshness_period`
- this will put this key in a list of keys to cleanup

During cleanup, under exclusive work_in_progress lock:
- go through this list and remove this item from the cache (if they're still in it)

During work (deploy/dry_run/get_facts), under exclusive work_in_progress lock:
- if we hit a stale item, synchronously remove it from the cache before recreating it

src/inmanta/agent/cache.py
```diff
@@ -40,13 +40,24 @@ def __init__(self, timeout: int = 24 * 3600, version: int = 0) -> None:
 
 
 class CacheItem:
-    def __init__(self, key: str, scope: Scope, value: Any, call_on_delete: Optional[Callable[[Any], None]]) -> None:
```
wouterdb, line 43: This while cache evalidation is getting too complicated, I would revert this to what it was and rename `_advance_time` to `clean_stale_entries`  and remove all calls to `_advance_time`

src/inmanta/agent/in_process_executor.py
```diff
@@ -67,12 +67,29 @@ def __init__(
...
+        self.periodic_cache_cleanup_job: Optional[asyncio.Task[None]] = None
+
+    async def start(self) -> None:
+        self.periodic_cache_cleanup_job = asyncio.create_task(self.cleanup_stale_cache_entries())
+
+    async def cleanup_stale_cache_entries(self) -> None:
+        """
+        This task periodically cleans up stale entries in the cache
+        """
+        while True:
```
wouterdb, line 88: when stopping, this should stopped as well (and awaited in join)

src/inmanta/agent/in_process_executor.py
```diff
@@ -67,12 +67,29 @@ def __init__(
...
+            reschedule_interval: float = 1.0
+            async with self.wip_lock:
+                self._cache.clean_stale_entries()
+            await asyncio.sleep(reschedule_interval)
```
wouterdb, line 92: A few lines below, the `self._cache.close()` should probably go on the threadpool as well  (requiring stop to be an async method), because it is called from the ioloop and it calls the finalizers

tests/conftest.py
```diff
@@ -1937,3 +1938,22 @@ def disable_version_and_agent_cleanup_job():
     orchestrationservice.PERFORM_CLEANUP = False
     yield
     orchestrationservice.PERFORM_CLEANUP = old_perform_cleanup
+
+
+@pytest.fixture(scope="session")
+def pip_index(tmp_path_factory) -> utils.PipIndex:
```
wouterdb, line 1944: Why is this added?

src/inmanta/agent/in_process_executor.py
```diff
@@ -67,12 +67,29 @@ def __init__(
...
+        self.periodic_cache_cleanup_job = asyncio.create_task(self.cleanup_stale_cache_entries())
+
+    async def cleanup_stale_cache_entries(self) -> None:
+        """
+        This task periodically cleans up stale entries in the cache
+        """
+        while True:
+            reschedule_interval: float = 1.0
+            async with self.wip_lock:
+                self._cache.clean_stale_entries()
```
wouterdb, line 91: This has to go on the threadpool

wouterdb (approved):

tests/conftest.py
Hugo-Inmanta: Not needed -> removed it

src/inmanta/agent/cache.py
```diff
@@ -202,11 +221,7 @@ def _cache(self, item: CacheItem) -> None:
                 self.keysforVersion[scope.version].add(item.key)
             except KeyError:
                 raise Exception("Added data to version that is not open")
-
-        bisect.insort_right(self.timerqueue, item)
-        if item.time < self.nextAction:
-            self.nextAction = item.time
```
wouterdb, line 208: this is still needed

wouterdb (approved):

src/inmanta/agent/cache.py
Hugo-Inmanta: indeed :+1:

arnaudsjs (changes requested):

src/inmanta/agent/in_process_executor.py
```diff
@@ -68,20 +68,42 @@ def __init__(
         self.thread_pool: ThreadPoolExecutor = ThreadPoolExecutor(1, thread_name_prefix="Pool_%s" % self.name)
 
         self._cache = inmanta.agent.cache.AgentCache(self)
+        # This lock ensures cache entry can not be cleaned up when
+        # the executor is actively working and vice versa
+        self.wip_lock = asyncio.Lock()
```
sanderr, line 73: I find `wip` a strange name. How about `activity_lock`?

src/inmanta/agent/in_process_executor.py
```diff
@@ -90,6 +112,8 @@ def join(self, thread_pool_finalizer: list[ThreadPoolExecutor]) -> None:
         assert self._stopped
         thread_pool_finalizer.append(self.provider_thread_pool)
         thread_pool_finalizer.append(self.thread_pool)
+        if self.periodic_cache_cleanup_job:
+            await self.periodic_cache_cleanup_job
```
sanderr, line 116: I believe this will suffer from the same issue as raised by @arnaudsjs [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1721120710485599).

sanderr (changes requested):

tests/test_cache.py
```diff
@@ -39,37 +42,72 @@ class MyResource(Resource):
...
+    )
+
+    cache = myagent_instance._cache
+    value = "test too"
+    cache.cache_value("test", value, timeout=0.1)
+    cache.cache_value("test2", value)
+
+    assert value == cache.find("test")
+    # Cache cleanup job is periodically triggered with a 1s delay
+    await asyncio.sleep(2)
```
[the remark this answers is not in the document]
Hugo-Inmanta: Added the `agent_cache_cleanup_tick_rate` config option to set this interval. 
Added a fixture (`set_custom_cache_cleanup_policy`) that sets this interval to 0.1s for this test. I'm monkeypatching the validator in this fixture to accept a float value, I'm not sure this is the cleanest approach.

src/inmanta/agent/in_process_executor.py
```diff
@@ -68,20 +68,42 @@ def __init__(
...
+        self.periodic_cache_cleanup_job = asyncio.create_task(self.cleanup_stale_cache_entries())
+
+    async def cleanup_stale_cache_entries(self) -> None:
+        """
+        Periodically cleans up stale entries in the cache. The clean_stale_entries
+        has to be called on the thread pool because it might call finalizers.
+        """
+
+        while not self._stopped:
+            reschedule_interval: float = 1.0
```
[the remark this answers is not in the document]
Hugo-Inmanta: I'm not sure actually, your other [comment](https://github.com/inmanta/inmanta-core/pull/7846#discussion_r1677892873) about the test being slow with a long expiry interval made me realize it would be nice to add an option to control this interval easily. I used the `is_time` validator for the option (which I had to monkey patch in the test to allow for <1s interval).
Maybe it would be less convoluted to use float everywhere then ?

src/inmanta/agent/in_process_executor.py
```diff
@@ -68,20 +68,51 @@ def __init__(
...
+        """
+        Periodically cleans up stale entries in the cache. The clean_stale_entries
+        has to be called on the thread pool because it might call finalizers.
+        """
+        reschedule_interval: int = self.cache_cleanup_tick_rate
+        while not self._stopped:
+            async with self.activity_lock:
+                if self._stopped:
+                    return
+                await asyncio.get_running_loop().run_in_executor(self.thread_pool, self._cache.clean_stale_entries)
```
arnaudsjs, line 97: Let's add a try-catch around this to make sure we don't drop out of the while loop if an exception occurs.

arnaudsjs (changes requested):

src/inmanta/agent/in_process_executor.py
```diff
@@ -68,20 +68,55 @@ def __init__(
...
+        reschedule_interval: int = self.cache_cleanup_tick_rate
+        while not self._stopped:
+            async with self.activity_lock:
+                if self._stopped:
+                    return
+                try:
+                    await asyncio.get_running_loop().run_in_executor(self.thread_pool, self._cache.clean_stale_entries)
+                except Exception:
+                    # Make sure we don't drop out of the while loop if an exception occurs.
+                    pass
```
arnaudsjs, line 101: Let's also log a message. Otherwise we are not aware about the fact that there is an issue.
```suggestion
                    # Make sure we don't drop out of the while loop if an exception occurs.
                    LOGGER.exception("An unexpected exception occurred while cleaning up the cache of agent %s.", self.name)
```

arnaudsjs (changes requested):

src/inmanta/agent/config.py
```diff
@@ -211,6 +211,14 @@ def is_executor_mode(value: str | AgentExecutorMode) -> AgentExecutorMode:
     is_time,
 )
 
+agent_cache_cleanup_tick_rate = Option[int](
+    "agent",
+    "cache-cleanup-tick-rate",
+    1,
+    "The rate (in seconds) at which the agent will periodically attempt to remove stale entries from the cache.",
```
wouterdb, line 218:
```suggestion
    "The rate (in seconds) at which the agent will periodically attempt to remove stale entries from the cache when idle.",
```

tests/test_cache.py
```diff
@@ -39,37 +44,89 @@ class MyResource(Resource):
...
+        "agent1", "local:", [executor.ResourceInstallSpec("test::Test", 5, blueprint1)]
+    )
+
+    cache = myagent_instance._cache
+    value = "test too"
+    cache.cache_value("test", value, timeout=0.1)
+    cache.cache_value("test2", value)
+
+    assert value == cache.find("test")
+    # Cache cleanup job is periodically triggered with a 1s delay
```
wouterdb, line 93:
```suggestion
    # Cache cleanup job is periodically triggered with a 0.1s delay
```

wouterdb (approved):
Nice

tests/forking_agent/test_executor.py
```diff
@@ -224,8 +224,12 @@ async def oldest_gone():
 
     await simplest.request_shutdown()
     await simplest.join()
+
+    async def check_connection_lost() -> bool:
+        return await simplest.call(GetName()) != simplest_blueprint.blueprint_hash()
+
     with pytest.raises(ConnectionLost):
-        await simplest.call(GetName())
+        await retry_limited(check_connection_lost, 1)
```
Hugo-Inmanta, line 232: Comment for linking purpose

PR #7866, opened 2024-07-18
Fix pkgresources import

sanderr:
# Description

Fixed import of `pkg_resources.extern`, which was dropped in the latest `setuptools` release (v71).

# Self Check:

- [x] Changelog entry
- [x] ~~Code is clear and sufficiently documented~~

### DIFF
--- a/changelogs/unreleased/fix-pkgresources-import.yml
+++ b/changelogs/unreleased/fix-pkgresources-import.yml
@@ -2,7 +2,7 @@ description: Fixed dropped import from pkg_resources
 change-type: patch
 sections:
   bugfix: "Addressed breaking change in setuptools (core Python library)"
-  upgrade-note: "If you had previously constrained `setuptools<71` in your `project.yml`, you may now drop the constraint"
+  upgrade-note: "If you had previously constrained `setuptools<71` in your project's `requirements.txt`, you may now drop the constraint"
 destination-branches:
   - master
   - iso7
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/fix-pkgresources-import.yml
```diff
@@ -0,0 +1,7 @@
+description: Fixed dropped import from pkg_resources
+sections:
+  bugfix: "Addressed breaking change in setuptools (core Python library)"
+  upgrade-note: "If you had previously constrained `setuptools<71` in your `pyproject.yml`, you may now drop the constraint"
+destination-branches:
+  - master
+  - iso7
```
sanderr, line 7: iso6 will have its own PR because it has a second occurrence of this import.

arnaudsjs (approved):

sanderr:
@edvgui requesting your review for the changelog entries (what's in "sections" will be visible to the customers in the generated changelog for the patch release in the documentation).

edvgui (approved):

sanderr:
@arnaudsjs I had to apply one more fix due to pypa/setuptools#4482. It's not pretty but it's the best I can think of on short notice, and we can hopefully get rid of the hack when we migrate away from `pkg_resources`.

arnaudsjs (approved):

hlloreda (approved):

Issue #7589, opened 2024-05-02
Code loading code in the agent constructor can be moved to in_process_executor.InProcessExecutorManager(


wouterdb:
Currently, when the agent is constructed, the agent code loader is set up in the constructor. 

It can be moved into the InProcessExecutorManager, to further separate both execution modes

PR #7867, opened 2024-07-18
Issue/7589 move agent code loader setup to in process executor manager

Hugo-Inmanta:
# Description

closes  https://github.com/inmanta/inmanta-core/issues/7589

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/agent_server/test_agent_code_loading.py
+++ b/tests/agent_server/test_agent_code_loading.py
@@ -339,7 +339,6 @@ def test():
          - all elements from <must_contain> are present
          - no element from <must_not_contain> are present
         """
-        must_contain = list(must_contain) if must_contain else []
         for package in package_list:
             assert package not in must_not_contain
             if package in must_contain:


src/inmanta/agent/agent.py
```diff
@@ -1283,14 +1277,35 @@ async def get_code(
 
         return resource_install_specs, invalid_resource_types
 
+    def initialize_loader(self) -> None:
```
sanderr, line 1280: I think we can move the entire functionality to `InProcessManager`, including all these bookkeeping attributes and the `_install` method. Wdyt?

sanderr (changes requested):

src/inmanta/agent/agent.py
Hugo-Inmanta: Yes, I agree :+1:

tests/agent_server/test_agent_code_loading.py
Hugo-Inmanta, line 1: I'm not 100% about the direction to go with the tests in this file.

I adapted them to pass with the current implementation but I'm not sure whether they still are relevant.

src/inmanta/agent/agent.py
```diff
@@ -1223,7 +1198,7 @@ async def get_code(
             - collection of ResourceInstallSpec for resource_types with valid handler code and pip config
             - set of invalid resource_types (no handler code and/or invalid pip config)
         """
-        if self._loader is None:
+        if not self.executor_manager.can_load_code():
```
sanderr, line 1201: I have a slight preference for making this a field `_code_loader: True` on `self`, so that we don't have to pollute the generic executor interface with a testing artifact specific to the in-process executor.

But I'm not opposed to how you did it, if you prefer it like that.

src/inmanta/agent/executor.py
```diff
@@ -527,3 +527,11 @@ async def join(self, thread_pool_finalizer: list[ThreadPoolExecutor], timeout: f
         Any threadpools that need to be closed can be handed of to the parent via thread_pool_finalizer
         """
         pass
+
+    @abc.abstractmethod
+    def can_load_code(self) -> bool:
+        """
+        Mechanism to speed up tests using the old (<= iso7) agent mechanism
+        by avoiding spawning a virtual environment.
```
sanderr, line 535: Good addition (the comment)

src/inmanta/agent/executor.py
```diff
@@ -527,3 +527,11 @@ async def join(self, thread_pool_finalizer: list[ThreadPoolExecutor], timeout: f
...
+        """
+        pass
```
sanderr, line 537: If you keep the method, I'd just `return True` and make it non-abstract.

tests/agent_server/test_agent_code_loading.py
```diff
@@ -23,10 +23,12 @@
 import tempfile
 import uuid
 from logging import DEBUG, INFO
+from typing import Sequence
```
sanderr, line 26:
```suggestion
from collections.abc import Sequence
```

tests/agent_server/test_agent_code_loading.py
```diff
@@ -71,8 +73,15 @@ async def make_source_structure(
         return hv
 
 
+@pytest.fixture(scope="function")
+def server_pre_start(server_config, request):
+    config.Config.set("agent", "executor-mode", request.param)
+    yield config
+
+
+@pytest.mark.parametrize("server_pre_start", ["threaded"], indirect=True)
```
sanderr, line 82: What has changed to require this?

tests/agent_server/test_agent_code_loading.py
```diff
@@ -318,7 +329,24 @@ def test():
...
     )
-    await agent.ensure_code(install_spec)
-
-    assert agent._env.are_installed(["pkg", "dep-a"])
-    assert not agent._env.are_installed(["dep-b", "dep-c"])
+    executor = await agent.executor_manager.get_executor("agent1", "localhost", install_spec)
+
+    installed_packages = executor.executor_virtual_env.get_installed_packages()
+
+    def check_packages(
```
sanderr, line 336: Same question here, why is this change required?

sanderr (changes requested):

tests/agent_server/test_agent_code_loading.py
sanderr: I think they are still relevant.

src/inmanta/agent/in_process_executor.py
```diff
@@ -446,6 +452,7 @@ def __init__(
         eventloop: asyncio.AbstractEventLoop,
         parent_logger: logging.Logger,
         process: "agent.Agent",
+        code_loader: bool = True,
```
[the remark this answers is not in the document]
Hugo-Inmanta: It is passed here: 
https://github.com/inmanta/inmanta-core/blob/2d997152002c8b2ff6bd727008709019e0dd8f36/src/inmanta/agent/agent.py#L991

Or did you mean apart from here ? It is indirectly passed in tests where we explicitly disable code loading for speed up e.g. 

https://github.com/inmanta/inmanta-core/blob/5166f236c8eb4188ecd920654a2684c77348ef6f/tests/test_server.py#L422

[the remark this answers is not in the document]
sanderr: Wow, not sure how I missed that.

src/inmanta/agent/agent.py
```diff
@@ -970,21 +957,8 @@ def __init__(
...
-            # Lock to ensure only one actual install runs at a time
-            self._loader_lock = Lock()
-            # Keep track for each resource type of the last loaded version
-            self._last_loaded_version: dict[str, executor.ExecutorBlueprint | None] = defaultdict(lambda: None)
-            # Cache to prevent re-fetching the same resource-version
-            self._previously_loaded: dict[tuple[str, int], ResourceInstallSpec] = {}
-            # Per-resource lock to serialize all actions per resource
-            self._resource_loader_lock = NamedLock()
+        # Cache to prevent re-fetching the same resource-version
+        self._previously_loaded: dict[tuple[str, int], ResourceInstallSpec] = {}
```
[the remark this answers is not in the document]
Hugo-Inmanta: Renamed to _code_cache

tests/agent_server/test_agent_code_loading.py
Hugo-Inmanta: I will remember this is the correct one of the two someday, but this is rather ingrained.

sanderr: I get it

tests/agent_server/test_agent_code_loading.py
Hugo-Inmanta: I'm not sure, this ties into my comment on the file https://github.com/inmanta/inmanta-core/pull/7867#discussion_r1691485611
As I understand it:
- `test_agent_code_loading` is aimed at testing executors in `executor-mode==threaded`,  because it is testing agent code reloading (and incidentally using in-process specific code e.g. `agent.executor_manager.ensure_code()`
- `test_agent_installs_dependency_containing_extras` (in its state prior to this PR) was only testing executors in `executor-mode==forking` (the default mode) and using forking mode specific code e.g. `executor.executor_virtual_env`

So I adapted the tests to each run against a specific executor mode. At the root of this is the fact that `ensure_code()` was moved from the `Agent` to the `InProcessExecutorManager` and I'm not sure how to adapt the tests that were relying on this.

src/inmanta/agent/agent.py
Hugo-Inmanta: I think this is a lot cleaner with `_code_loader` indeed :+1:

tests/agent_server/test_agent_code_loading.py
sanderr: > I adapted the tests to each run against a specific executor mode

This is the part I don't get. You say that they were already running against a specific executor mode, so what does this change exactly? Is it that previously it didn't use the executor so we didn't care if it picked the wrong one for the mode we were testing?

sanderr: That would answer the question for the `server_pre_start` change, but why did these assertions need to be changed? `executor.executor_virtual_env` is the same as `agent._env` was, isn't it?

Hugo-Inmanta: > Is it that previously it didn't use the executor so we didn't care if it picked the wrong one for the mode we were testing?

- For the `test_agent_code_loading` I think so, yes. Since it is testing `ensure_code()` which was moved from the `Agent` to the `InProcessExecutorManager` I think this test is only relevant for `threaded` mode.

> executor.executor_virtual_env is the same as agent._env was, isn't it?

Not exactly, the former is a https://github.com/inmanta/inmanta-core/blob/cfdd76f035e218dda683c34cd3d26d37b130c825/src/inmanta/agent/executor.py#L247 whereas the latter is a 
https://github.com/inmanta/inmanta-core/blob/82842dfb76b1fb120576b884723758fcaf066e5d/src/inmanta/env.py#L1195

and so it doesn't have a `are_installed()` method.

sanderr: Right, it's not activated.

sanderr (approved):

tests/agent_server/test_agent_code_loading.py
```diff
@@ -318,7 +329,24 @@ def test():
...
+        """
+        must_contain = list(must_contain) if must_contain else []
+        for package in package_list:
+            assert package not in must_not_contain
+            if package in must_contain:
+                must_contain.remove(package)
+
+        assert not must_contain
+
+    check_packages(package_list=installed_packages, must_contain=["pkg", "dep-a"], must_not_contain=["dep-b", "dep-c"])
```
arnaudsjs, line 352: Very minor, but using sets here improves the performance of the `in` operation. Although the benefits will be limited as these are very small lists.
```suggestion
    check_packages(package_list=installed_packages, must_contain={"pkg", "dep-a"}, must_not_contain={"dep-b", "dep-c"})
```

arnaudsjs (changes requested):

arnaudsjs (approved):

Issue #6668, opened 2023-10-24
drop usage of pkg_resources in core


sanderr:
This library has been deprecated. It should be replaced with alternatives for iso7+. The `<` constraint on `setuptools` should then be raised and the associated dependabot ignore dropped.

See note in [pkg_resources docs](https://setuptools.pypa.io/en/latest/pkg_resources.html).

PR #7878, opened 2024-07-22
Setuptools constrain

sanderr:
# Description

- Reverted previous fix
- Added upper bound to setuptools dependency. I went for `<70` rather than `<71` because the identity of the `InvalidRequirement` exception changed between 69 and 70. This is not as breaking as the issue we were observing with 71, but it might result in some less helpful error messages. 
- Follow-up ticket #6668 scheduled for next product release. We will be able to start tracking the latest release for the upper bound at that point.
- I'll apply the same fix to iso6

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ requires = [
     "pynacl~=1.5",
     "python-dateutil~=2.0",
     "pyyaml~=6.0",
-    "setuptools",
+    "setuptools<70",
     "texttable~=1.0",
     "tornado~=6.0",
     # lower bound because of ilevkivskyi/typing_inspect#100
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -30,7 +30,7 @@
     "pynacl~=1.5",
     "python-dateutil~=2.0",
     "pyyaml~=6.0",
-    "setuptools",
+    "setuptools<70",
```
edvgui, line 33: Why not 
```suggestion
    "setuptools<71",
```
This is the one we have been using in the workaround

edvgui (changes requested):

setup.py
sanderr: I know. I deviated from that deliberately. From the PR description: "I went for <70 rather than <71 because the identity of the InvalidRequirement exception changed between 69 and 70. This is not as breaking as the issue we were observing with 71, but it might result in some less helpful error messages."

edvgui: ok!

edvgui (approved):

Issue #7625, opened 2024-05-15
Remove unused shell argument from run_command_and_stream_output


wouterdb:
Find more live information in Aikido here: https://app.aikido.dev/queue?sidebarIssue=1157612&groupId=3504&sidebarIssueTask=18026&sidebarTab=tasks

==== Scope ====
This task includes issues in the following code repository:
- inmanta-core: src/inmanta/env.py

==== TLDR ====
Using subprocess.Popen with expressions based on user input can execute arbitrary code. This is especially true when shell=true, because it adds the extra danger of shell injection attacks

PR #7908, opened 2024-07-25
Remove shell arguments

hlloreda:
# Description

closes https://github.com/inmanta/inmanta-core/issues/7625

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/module.py
+++ b/src/inmanta/module.py
@@ -98,7 +98,10 @@ class InmantaModuleRequirement:
 
     def __init__(self, requirement: Requirement) -> None:
         if requirement.project_name.startswith(ModuleV2.PKG_NAME_PREFIX):
-            raise ValueError("InmantaModuleRequirement instances work with inmanta module names, not python package names.")
+            raise ValueError(
+                f"InmantaModuleRequirement instances work with inmanta module names, not python package names. "
+                f"Problematic case: {str(requirement)}"
+            )
         self._requirement: Requirement = requirement
 
     @property
--- a/src/inmanta/module.py
+++ b/src/inmanta/module.py
@@ -3079,7 +3079,7 @@ class Module(ModuleLike[TModuleMetadata], ABC):
     def execute_command(self, cmd: str) -> None:
         print(f"executing {cmd} on {self.name} in {self._path}")
         print("=" * 10)
-        subprocess.call(cmd, shell=True, cwd=self._path)
+        subprocess.call(cmd, cwd=self._path)
         print("=" * 10)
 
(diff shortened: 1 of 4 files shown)


src/inmanta/module.py
```diff
@@ -3079,7 +3079,7 @@ def _get_fq_mod_name_for_py_file(self, py_file: str, plugin_dir: str, mod_name:
     def execute_command(self, cmd: str) -> None:
         print(f"executing {cmd} on {self.name} in {self._path}")
         print("=" * 10)
-        subprocess.call(cmd, shell=True, cwd=self._path)
+        subprocess.call(cmd, cwd=self._path)
         print("=" * 10)
```
hlloreda, line 3083: I saw another usage of this argument, this method is only used in a `do` method in `ModuleTool`. Can it be taken out?

hlloreda: cc @arnaudsjs

arnaudsjs: I don't think we can remove it on the `do()` method, because it would be a breaking change on the `do` command. But I don't think the `do` command is used in practice. Maybe it would be a good idea to just drop that command entirely. Let's create a ticket to discuss this in the planning meeting.

hlloreda: Ticket has been created: https://github.com/inmanta/inmanta-core/issues/7913

tests/moduletool/test_install.py
```diff
@@ -1263,32 +1262,16 @@ def test_module_install_logging(local_module_package_index: str, snippetcompiler
         )
 
 
-@pytest.mark.slowtest
-def test_real_time_logging(caplog):
```
arnaudsjs, line 1267: The goal of this test is to verify `run_command_and_stream_output` logs in real time. We should keep that test. The only thing we cannot keep is the command it executes, because it requires `shell=True`, which is removed by this PR. But there are ways to work around this by changing the command in the following way for example:

```
cmd: list[str] = ["sh", "-c", "echo one && sleep 1 && echo two"]
```

tests/moduletool/test_install.py
```diff
@@ -1263,32 +1262,16 @@ def test_module_install_logging(local_module_package_index: str, snippetcompiler
         )
 
 
-@pytest.mark.slowtest
```
arnaudsjs, line 1266: This test performs a sleep so it's still slow. We should keep this marker.

arnaudsjs (changes requested):

arnaudsjs (approved):

Issue #7528, opened 2024-04-17
discovered resources: link to discovery resource


sanderr:
include id and link to discovery resource that reported it

Follow the managed resource link format https://docs.inmanta.com/inmanta-service-orchestrator/latest/reference/apilinks.html#api-self-referencing-links

sanderr:
See status [on Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1722611435774309)

PR #7912, opened 2024-07-29
Issue/7528 add link from discovered resource to discovery resource

Hugo-Inmanta:
# Description

Add link to the discovery resource responsible for discovering the resources in the discovered resource view.

closes https://github.com/inmanta/inmanta-core/issues/7528

# Self Check:

- [x] Changelog entry
> src/inmanta/data/dataview.py:1326: error: Argument "discovery_resource_id" to "DiscoveredResource" has incompatible type "object"; expected "ResourceIdStr | None"  [arg-type]


tests/test_discovered_resources.py
```diff
@@ -209,26 +278,85 @@ async def test_discovered_resource_get_paging(server, client, agent, environment
...
     assert "Failed to validate argument" in result.result["message"]
 
+    result = await agent._client.discovered_resource_create(
+        tid=environment,
+        discovered_resource_id="test::Resource[agent1,key=key]",
+        values={"value1": "test1", "value2": "test2"},
+        discovery_resource_id="invalid_rid",
+    )
+    assert result.code == 400
+    assert "Failed to validate argument" in result.result["message"]
```
arnaudsjs, line 316: Shouldn't we assert on which argument it's failing? The error message might be about some other argument than the one we are expecting.

src/inmanta/data/dataview.py
```diff
@@ -1313,7 +1321,8 @@ def construct_dtos(self, records: Sequence[Record]) -> Sequence[dict[str, str]]:
                 discovered_resource_id=res["discovered_resource_id"],
                 values=json.loads(res["values"]),
                 managed_resource_uri=f"/api/v2/resource/{res['discovered_resource_id']}" if res["managed"] else None,
-            ).model_dump()
+                discovery_resource_id=res["discovery_resource_id"],
+            ).model_dump(exclude={"discovery_resource_id"})
```
arnaudsjs, line 1325: Shouldn't we make this exclude part of the config of the pydantic class, instead of setting this parameter in every location where we serialize the object.

changelogs/unreleased/7528-add-link-to-discovery-resource.yml
```diff
@@ -0,0 +1,6 @@
+description: Add link to the resource responsible for the discovery to the discovered resource view.
+issue-nr: 7528
+change-type: patch
```
arnaudsjs, line 3:
```suggestion
change-type: minor
```

changelogs/unreleased/7528-add-link-to-discovery-resource.yml
```diff
@@ -0,0 +1,6 @@
+description: Add link to the resource responsible for the discovery to the discovered resource view.
```
arnaudsjs, line 1: What do you mean with "the discovered resource view". I don't think this has a lot of meaning to an end-user.

arnaudsjs (changes requested):

src/inmanta/data/dataview.py
arnaudsjs: I have a suspicion that the regular GET endpoint to retrieve a specific discovered resource is returning this field, while the batch endpoint isn't. Is that correct? If this is the case, we need to improve the test coverage.

changelogs/unreleased/7528-add-link-to-discovery-resource.yml
Hugo-Inmanta: I meant the Discovered Resource page, but this is inaccurate, as the displaying of this link itself is not yet implemented on the frontend side. Changed this to:
```
  Update the endpoints returning discovered resource (`GET /api/v2/discovered/<discovered_resource_id>` and
  `GET /api/v2/discovered`) to include a link to the resource responsible for the discovery.
```

src/inmanta/agent/handler.py
```diff
@@ -1023,12 +1023,17 @@ def _call_discovered_resource_create_batch(
                 discovered_resources: abc.Sequence[DiscoveredResource],
             ) -> typing.Awaitable[Result]:
                 return self.get_client().discovered_resource_create_batch(
-                    tid=self._agent.environment, discovered_resources=discovered_resources
+                    tid=self._agent.environment,
+                    discovered_resources=discovered_resources,
```
Hugo-Inmanta, line 1027: Comment for linking purpose

src/inmanta/data/model.py
```diff
@@ -747,31 +749,48 @@ class LoginReturn(BaseModel):
...
     managed_resource_uri: Optional[str] = None
 
-    @field_validator("discovered_resource_id")
-    @classmethod
-    def discovered_resource_id_is_resource_id(cls, v: str) -> str:
-        if resources.Id.is_resource_id(v):
-            return v
-        raise ValueError(f"id {v} is not of type ResourceIdStr")
+    # This field is excluded during serialization in favour of the discovery_resource_uri
+    discovery_resource_id: Optional[ResourceIdStr] = Field(default=None, exclude=True)
```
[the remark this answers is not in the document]
sanderr: @arnaudsjs tagging you because this is a deviation of what you reviewed.

src/inmanta/protocol/methods_v2.py
```diff
@@ -1546,13 +1546,17 @@ def set_password(username: str, password: str) -> None:
     varkw=True,
 )
 def discovered_resource_create(
-    tid: uuid.UUID, discovered_resource_id: str, **kwargs: object  # bypass the type checking for the values
+    tid: uuid.UUID,
+    discovered_resource_id: str,
+    discovery_resource_id: str | None = None,
```
sanderr, line 1551: In the same vein as my comment regarding optionality, if accepted there, we need to make it non-optional here as well.

src/inmanta/data/dataview.py
```diff
@@ -1294,11 +1293,19 @@ def get_base_query(self) -> SimpleQueryBuilder:
                     dr.environment,
                     dr.discovered_resource_id,
                     dr.values,
-                    (rps.resource_id IS NOT NULL)  AS managed
+                    (rps_1.resource_id IS NOT NULL)  AS managed,
+                    rps_2.resource_id as discovery_resource_id
```
sanderr, line 1297:
```suggestion
                    (rps_1.resource_id IS NOT NULL) AS managed,
                    rps_2.resource_id AS discovery_resource_id
```

src/inmanta/data/dataview.py
```diff
@@ -1294,11 +1293,19 @@ def get_base_query(self) -> SimpleQueryBuilder:
...
-                LEFT JOIN {data.ResourcePersistentState.table_name()} rps
-                ON dr.environment=rps.environment AND dr.discovered_resource_id = rps.resource_id
+
+                -- Retrieve the corresponding managed resource id (if any)
+                LEFT JOIN {data.ResourcePersistentState.table_name()} rps_1
+                ON dr.environment=rps_1.environment AND dr.discovered_resource_id = rps_1.resource_id
+
+                -- Retrieve the id of the resource responsible for discovering this resource
+                LEFT JOIN {data.ResourcePersistentState.table_name()} rps_2
+                ON dr.environment=rps_2.environment AND dr.discovery_resource_id = rps_2.resource_id
```
sanderr, line 1307: What's the purpose of this join? We join on resource id, then select resource id?

sanderr (changes requested):

src/inmanta/data/model.py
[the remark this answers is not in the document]
sanderr: @Hugo-Inmanta out of my comments it's probably best to consider this one first, because it may invalidate some of the others.

src/inmanta/data/model.py
```diff
@@ -747,31 +749,48 @@ class LoginReturn(BaseModel):
     user: User
 
 
+def is_resource_id(v: Optional[str], info: ValidationInfo, allow_none: bool = False) -> str | None:
+    if v is None:
+        if allow_none:
+            return v
+        raise ValidationError(f"{info.field_name} is None")
```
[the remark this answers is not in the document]
Hugo-Inmanta: For `discovered_resource_id`, yes, but I use the same validator for `discovery_resource_id` which can be None

src/inmanta/data/dataview.py
Hugo-Inmanta: This is to make sure that the `discovery_resource_id` corresponds to a proper resource that exists in the db.

sanderr: Ah right, because we link to it. Ok, fine by me.

src/inmanta/data/model.py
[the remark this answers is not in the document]
Hugo-Inmanta: > Do I understand correctly that this is optional because we may currently have None values in the database?

Yes, and also wouldn't that be a breaking change ? Won't this break in all the places where this endpoint is called without a `discovery_resource_id` ?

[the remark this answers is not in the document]
sanderr: Technically yes, but the method is defined as an agent-server API, and we only really ever use autostarted agents. So client and server should always use the same code afaik.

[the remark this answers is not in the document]
Hugo-Inmanta: Went with your approach for the typing. And the following:
------
DB side:
`DiscoveredResource`: add an optional field `discovery_resource_id`
The value of this field drives the choice of model class for serialization:
- if None -> `m.DiscoveredResource`
- else -> `m.LinkedDiscoveredResource`
------
Model side:
Parent class `DiscoveredResource` remains moslty unchanged
Add child class `LinkedDiscoveredResource` that links to an existing discovery resource

src/inmanta/data/model.py
```diff
@@ -755,23 +767,44 @@ class DiscoveredResource(BaseModel):
...
     values: dict[str, object]
     managed_resource_uri: Optional[str] = None
 
-    @field_validator("discovered_resource_id")
-    @classmethod
-    def discovered_resource_id_is_resource_id(cls, v: str) -> str:
-        if resources.Id.is_resource_id(v):
-            return v
-        raise ValueError(f"id {v} is not of type ResourceIdStr")
+    _validate_discovered_rid = field_validator("discovered_resource_id")(is_resource_id)
```
arnaudsjs, line 774: Aren't we now performing this validation twice. Once because of the type validation `ResourceId` and once here by setting the `_validate_discovered_rid` field.

arnaudsjs (changes requested):

src/inmanta/data/model.py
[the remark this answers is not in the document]
sanderr: > The value of this field drives the choice of model class for serialization

I don't think this is required, and it adds a lot of verbosity. The `LinkedDiscoveredResource` I proposed is just to declare the requirement that the field be set. This is relevant for the inbound API. For the outbound API it doesn't make any sense to promise we'll set the field, maybe. We might as well not promise and use the generic `DiscoveredResource`.

We should probably create a follow-up ticket to consider how to phase out the unlinked resources.

src/inmanta/data/dataview.py
```diff
@@ -1315,15 +1315,26 @@ def get_base_query(self) -> SimpleQueryBuilder:
...
-                discovered_resource_id=res["discovered_resource_id"],
-                values=json.loads(res["values"]),
-                managed_resource_uri=f"/api/v2/resource/{res['discovered_resource_id']}" if res["managed"] else None,
-                discovery_resource_id=res["discovery_resource_id"],
-            ).model_dump()
-            for res in records
-        ]
+        out = []
+        for res in records:
+            if res["discovery_resource_id"]:
```
sanderr, line 1320: I don't see the value in this, see [here](https://github.com/inmanta/inmanta-core/pull/7912#discussion_r1709054592). I think we can revert this entire block to what it was in db7e7ca.

src/inmanta/data/model.py
```diff
@@ -755,23 +767,44 @@ class DiscoveredResource(BaseModel):
...
+class LinkedDiscoveredResource(DiscoveredResource):
+    """
+    DiscoveredResource linked to the discovery resource that discovered it.
+
+    :param discovery_resource_id: Resource id of the (managed) discovery resource that reported this
+           discovered resource.
+    """
+
+    discovery_resource_id: ResourceId
+    _validate_discovery_rid = field_validator("discovery_resource_id")(functools.partial(is_resource_id))
```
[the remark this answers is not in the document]
sanderr: It's also redundant because `ResourceId` is an annotated type with this validator built-in.

src/inmanta/data/model.py
```diff
@@ -763,28 +765,40 @@ class DiscoveredResource(BaseModel):
...
-        discovered resource.
     """
 
-    discovered_resource_id: ResourceIdStr
+    discovered_resource_id: ResourceId
     values: dict[str, object]
     managed_resource_uri: Optional[str] = None
 
-    # This field is excluded during serialization in favour of the discovery_resource_uri
-    discovery_resource_id: Optional[ResourceIdStr] = Field(default=None, exclude=True)
```
sanderr: I think we should keep this as `Optional[ResourceId]`

src/inmanta/data/model.py
```diff
@@ -763,28 +765,40 @@ class DiscoveredResource(BaseModel):
...
+    def to_dao(self, env: uuid.UUID) -> "data.DiscoveredResource":
+        return data.DiscoveredResource(
+            discovered_resource_id=self.discovered_resource_id,
+            values=self.values,
+            discovered_at=datetime.datetime.now(),
+            environment=env,
+        )
+
+
+class LinkedDiscoveredResource(DiscoveredResource):
```
sanderr, line 785: Perhaps add a comment (not docstring because that's included in the docs) as to why we need this class and what its purpose is (i.e. API input).

src/inmanta/data/__init__.py
```diff
@@ -6404,10 +6404,15 @@ class DiscoveredResource(BaseDocument):
     __primary_key__ = ("environment", "discovered_resource_id")
 
     def to_dto(self) -> m.DiscoveredResource:
+        if self.discovery_resource_id:
```
sanderr, line 6407: Same here, I would just make it a `DiscoveredResource`

src/inmanta/data/model.py
```diff
@@ -749,12 +749,14 @@ class LoginReturn(BaseModel):
...
-    if v is None:
-        if allow_none:
-            return v
-        raise ValidationError(f"{info.field_name} is None")
-    return resources.Id.parse_id(v).resource_str()
+def is_resource_id(v: ResourceIdStr, info: pydantic.ValidationInfo) -> str:
+    try:
+        return resources.Id.parse_id(v).resource_str()
+    except Exception:
+        raise ValueError(f"Validation failed for {info.field_name}")
```
sanderr, line 756: The parse method already raises a `ValueError` with a more useful error message. I would drop the try-catch, perhaps with a comment that we rely on the method's `ValueError` for validation.

src/inmanta/data/model.py
```diff
@@ -763,28 +765,40 @@ class DiscoveredResource(BaseModel):
...
+    discovery_resource_id: ResourceId
+    _validate_discovery_rid = field_validator("discovery_resource_id")(functools.partial(is_resource_id))
 
     @computed_field  # type: ignore[misc]
     @property
     def discovery_resource_uri(self) -> str | None:
-        if self.discovery_resource_id is None:
-            return None
-        return f"/api/v2/resource/{parse.quote(self.discovery_resource_id)}"
+        return f"/api/v2/resource/{urllib.parse.quote(self.discovery_resource_id)}"
```
sanderr, line 799: This will need the `is None` check again.

sanderr (changes requested):

src/inmanta/data/dataview.py
```diff
@@ -1308,14 +1315,26 @@ def get_base_query(self) -> SimpleQueryBuilder:
...
+                )
+            else:
+                out.append(
+                    model.DiscoveredResource(
+                        discovered_resource_id=res["discovered_resource_id"],
+                        values=json.loads(res["values"]),
+                        managed_resource_uri=f"/api/v2/resource/{res['discovered_resource_id']}" if res["managed"] else None,
+                    ).model_dump()
+                )
+        return out
```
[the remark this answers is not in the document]
Hugo-Inmanta: Nice catch, I think it is.

src/inmanta/data/model.py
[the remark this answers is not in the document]
Hugo-Inmanta: Follow up ticket -> https://github.com/inmanta/inmanta-core/issues/8004

arnaudsjs (changes requested):

src/inmanta/data/model.py
```diff
@@ -747,31 +748,67 @@ class LoginReturn(BaseModel):
     user: User
 
 
+def is_resource_id(v: ResourceIdStr) -> str:
+    # For validation, we rely on the parse_id method raising a ValueError for invalid resource ids.
+    return resources.Id.parse_id(v).resource_str()
```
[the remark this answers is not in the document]
Hugo-Inmanta: Mypy was complaining when typing this as `def is_resource_id(v: str) -> ResourceIdStr:` 
```
src/inmanta/data/model.py:753: error: Argument 1 to "parse_id" of "Id" has incompatible type "str"; expected "ResourceVersionIdStr | ResourceIdStr"  [arg-type]
```

Added a cast to `ResourceIdStr` in the call to `parse_id`

src/inmanta/data/model.py
```diff
@@ -747,31 +748,67 @@ class LoginReturn(BaseModel):
...
+            return None
+        return f"/api/v2/resource/{urllib.parse.quote(self.discovery_resource_id, safe='')}"
+
+    def to_dao(self, env: uuid.UUID) -> "data.DiscoveredResource":
+        return data.DiscoveredResource(
+            discovered_resource_id=self.discovered_resource_id,
+            values=self.values,
+            discovered_at=datetime.datetime.now(),
+            environment=env,
+        )
```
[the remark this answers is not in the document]
Hugo-Inmanta: Nice catch, I think this method is actually redundant, we only want to insert `LinkedDiscoveredResource`s in the db from now on.

src/inmanta/data/model.py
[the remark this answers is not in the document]
arnaudsjs: Ah. I see. That is unexpected to me. I would expect the `Id.parse_id()` method to accept a string as well, instead of only `ResourceVersionIdStr | ResourceIdStr`. I would do the following:
* Remove the type cast, because at the moment the type cast is actually a lie. We are not certain the string is actually a ResourceIdStr.
* I would ask on Slack whether there is a reason to not accept the string type here. The change was made in https://github.com/inmanta/inmanta-core/commit/85aa5e4b629d287b1e2c5ca625843192cd3b00f5

arnaudsjs (changes requested):
See comment.

arnaudsjs (approved):

src/inmanta/data/model.py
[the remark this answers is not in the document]
Hugo-Inmanta: [Slack discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1724827928487769) for reference

src/inmanta/data/dataview.py
```diff
@@ -18,12 +18,11 @@
 
 import abc
 import json
+import urllib
```
sanderr, line 21:
```suggestion
import urllib.parse
```

src/inmanta/data/model.py
```diff
@@ -749,14 +748,13 @@ class LoginReturn(BaseModel):
     user: User
 
 
-def is_resource_id(v: ResourceIdStr, info: pydantic.ValidationInfo) -> str:
-    try:
-        return resources.Id.parse_id(v).resource_str()
-    except Exception:
-        raise ValueError(f"Validation failed for {info.field_name}")
+def check_resource_id_str(v: str) -> ResourceIdStr:
```
sanderr, line 751: Tiny preference to make this "private"
```suggestion
def _check_resource_id_str(v: str) -> ResourceIdStr:
```

Not required.

sanderr (changes requested):
One tiny change request (the `urllib` import). If need be (because I'll be off next week) you can dismiss my review at the bottom of the page under "1 change requested".

PR #7922, opened 2024-07-30
Issue/document transfer optimization

arnaudsjs:
# Description

Add documentation about `transfer optimization` feature.

Closes https://github.com/inmanta/inmanta-lsm/issues/1794

# Self Check:

- [x] Changelog entry


hlloreda (approved):
Nice documentation!

docs/lsm/transfer_optimization/transfer_optimization.rst
```diff
@@ -0,0 +1,59 @@
...
+infrastructure). This assumption doesn't hold in all situations. The service model could for example change the
+low-level intent based on the current state of the service instance. Caution is advised when modelling a lifecycle
+with state-preserving transfer, as incorrectly marking a transfer as state-preserving will cause the orchestrator
+to behave incorrectly.
+
+Enabling the transfer optimization feature
+==========================================
+
+The environment setting :inmanta.environment-settings:setting:`enable_lsm_transfer_optimization` can be used to enable
+the transfer optimization feature. When enabled, the LSM extension will perform a compile when transitioning between
```
sanderr, line 47:
```suggestion
the transfer optimization feature. When enabled, the LSM extension will perform a compile only when transitioning between
```

docs/lsm/transfer_optimization/transfer_optimization.rst
```diff
@@ -0,0 +1,59 @@
...
+two states that don't preserve the desired state. When disabled, a recompile will be done for each state transition.
+Validation compiles are not impacted by this setting. They are always executed.
+
+Testing
+=======
+
+The Inmanta server validates every lifecycle that is exported to the server and it will reject any lifecycle with
```
sanderr, line 54: > will reject any lifecycle with incorrect state-preserving transfers

This is contradicted later when you say we can not exhaustively detect all cases. Indeed we can't, so I think we should word this more as that the server "attempts to", or "has some validation for general invariants" or something rather than that it will reject any incorrectly marked transfer.

sanderr (changes requested):
Nice, looks clear to me. Good introduction. Two small remarks.

sanderr (approved):

PR #7956, opened 2024-08-07
Issue/slow tests compiler

wouterdb:
# Description

I attempted to speed up the compiler tests

Performance trace prior to optimization:
- 20% time spent on module loading -> reduced reliance on std  
- 29% of time spent verifying python dependencies -> patched it out of the compiler tests (caching was too complicated to be worth the effort, I tried) -> down to 8.28%

On master, prior to change, compiler tests take 191 on this branch 116, 75s time reduction 

# Performance graph before
baseline

# Performance graph after

combined

it does make the test a bit more verbose 

# Self Check:

- [x] Changelog entry


src/inmanta/env.py
```diff
@@ -975,9 +975,8 @@ def install_for_config(
         finally:
             self.notify_change()
 
-    @classmethod
     def get_constraint_violations_for_check(
```
wouterdb, line 978: This may be a bit tricky and perhaps no required after all

src/inmanta/env.py
```diff
@@ -1177,16 +1177,16 @@ def notify_change(self) -> None:
...
     environment. This function should only be called when a Python environment has been set up dynamically and this environment
     should be treated as if this process was spawned from it, and even then with great care.
 
-    :param python_path: The path to the python binary. Only one of `python_path` and `env_path` should be set.
-    :param env_path: The path to the python environment directory. Only one of `python_path` and `env_path` should be set.
+    :param new_process_env: new env to use
     """
-    process_env.__init__(python_path=python_path, env_path=env_path)  # type: ignore
+    global process_env
+    process_env = new_process_env
```
wouterdb, line 1189: I simplified this, I hope it is OK

tests/compiler/dataflow/test_model_assignment.py
```diff
@@ -106,10 +106,13 @@ def test_dataflow_model_attribute_assignment_responsible(dataflow_test_helper: D
 entity Test:
     int n
 end
-implement Test using std::none
+implement Test using none
```
wouterdb, line 109: this change is repeated a few 100 times

arnaudsjs: I don't get why this results in a performance improvement.

arnaudsjs (changes requested):

tests/compiler/dataflow/test_model_assignment.py
wouterdb: loading std is very slow compared to the code under test (much larger, plugins, ...) 

So I make it not load std at all 

(std::Entity is not in std, btw)

src/inmanta/env.py
```diff
@@ -1177,16 +1176,16 @@ def notify_change(self) -> None:
 
 
 @stable_api
-def mock_process_env(*, python_path: Optional[str] = None, env_path: Optional[str] = None) -> None:
+def mock_process_env(new_process_env: ActiveEnv) -> None:
```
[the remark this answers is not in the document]
wouterdb: right, I'll fix that

src/inmanta/env.py
wouterdb: @sanderr do you think this is OK?

tests/compiler/dataflow/test_model_assignment.py
arnaudsjs: > (std::Entity is not in std, btw)

Right! I forgot about that.

src/inmanta/env.py
sanderr: I'm not sure it is. I'd have to check where exactly we use it, but I believe this will break any non-qualified imports. We could have explicitly disallowed it, but I don't think we did. We may be lucky in that all our uses are qualified, in which case perhaps we can get away with it.

Hugo-Inmanta (approved):
LGTM appart from the subtleties in src/inmanta/env.py which are lost on me.

src/inmanta/env.py
```diff
@@ -1182,13 +1183,28 @@ def mock_process_env(*, python_path: Optional[str] = None, env_path: Optional[st
     Overrides the process environment information. This forcefully sets the environment that is recognized as the outer Python
     environment. This function should only be called when a Python environment has been set up dynamically and this environment
     should be treated as if this process was spawned from it, and even then with great care.
-
     :param python_path: The path to the python binary. Only one of `python_path` and `env_path` should be set.
     :param env_path: The path to the python environment directory. Only one of `python_path` and `env_path` should be set.
+
+    When using this method in a fixture to set an reset virtualenv, it is preferable to use `store_venv()`
```
wouterdb, line 1189:
```suggestion
    When using this method in a fixture to set and reset virtualenv, it is preferable to use `store_venv()`
```

src/inmanta/env.py
```diff
@@ -1182,13 +1183,28 @@ def mock_process_env(*, python_path: Optional[str] = None, env_path: Optional[st
...
-
     :param python_path: The path to the python binary. Only one of `python_path` and `env_path` should be set.
     :param env_path: The path to the python environment directory. Only one of `python_path` and `env_path` should be set.
+
+    When using this method in a fixture to set and reset virtualenv, it is preferable to use `store_venv()`
     """
     process_env.__init__(python_path=python_path, env_path=env_path)  # type: ignore
 
 
+def swap_process_env(env: ActiveEnv) -> ActiveEnv:
```
sanderr, line 1194: I think this will be OK, but I'd suggest not mergint it on the same day as the pkg_resources PR so we can identify fallout.

src/inmanta/env.py
```diff
@@ -1281,3 +1297,64 @@ class VenvActivationFailedError(Exception):
     def __init__(self, msg: str) -> None:
         super().__init__(msg)
         self.msg = msg
+
+
+@dataclass
+class VenvSnapshot:
```
sanderr, line 1303: Nice! Well, once you accept the horrible things we do here, but that's an established fact.

sanderr (changes requested):

sanderr (approved):
I think I clicked the wrong button on my previous review.

PR #7961, opened 2024-08-08
drop netifaces and enable uv

wouterdb:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,7 @@ black = black src tests tests_common
 .PHONY: install
 install:
 	pip install -U setuptools pip uv
-	uv pip install -U -r requirements.txt -r requirements.dev.txt
-	uv pip install -e .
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -5,9 +5,9 @@ black = black src tests tests_common
 
 .PHONY: install
 install:
-	pip install -U setuptools pip
-	pip install -U -r requirements.txt -r requirements.dev.txt
-	pip install -e .
+	pip install -U setuptools pip uv
+	uv pip install -U -r requirements.txt -r requirements.dev.txt
+	uv pip install -e .
```
sanderr, line 10: Might as well take the opportunity to fix this while we're at it.

```suggestion
	uv pip install -U --upgrade-strategy=eager -e . -c requirements.txt -r requirements.dev.txt
```

sanderr (approved):

PR #7964, opened 2024-08-08
Allow co-hosting executors 

wouterdb:
# Description

1. I upgrade the caching/pooling infrastructure to be able to handle the more advanced case of stacking executors on processes 
2. One process can now host several executors 

# TODO

- [x] move env manager to new framework
- [x] naming  (closing, closed, ...)
- [x]  review log output
- [x] make sure docs are in the right place 
- [x] hunt todos 

# Self Check:

- [x] Changelog entry


bartv (commented):
I think it looks good and is not to complicated. The only remark I have is that "close" is maybe not the best term and it is also not that consistent:

close vs is_stopping

Wouldn't shutdown be the best term here? And is_shutting_down?

src/inmanta/agent/forking_executor.py
```diff
@@ -685,179 +821,237 @@ def init_once(cls) -> None:
...
+    async def join(self) -> None:
+        await super().join()
+        await self.environment_manager.join()
+        await asyncio.gather(*(child.join() for child in self.pool.values()))
+
+    def _id_to_internal(self, ext_id: executor.ExecutorBlueprint) -> executor.ExecutorBlueprint:
+        return ext_id
+
+    async def create_member(self, blueprint: executor.ExecutorBlueprint) -> MPProcess:
+        # TODO: logging
```
wouterdb, line 846: todo alert

src/inmanta/agent/executor.py
```diff
@@ -248,56 +254,20 @@ class ResourceInstallSpec:
     blueprint: ExecutorBlueprint
 
 
-class PoolMember(abc.ABC):
```
wouterdb, line 251: moved to resourcepool.py

src/inmanta/agent/resourcepool.py
```diff
@@ -0,0 +1,395 @@
...
+        self.shut_down = True
+        self.shutting_down = True
+        for listener in self.termination_listeners:
+            await listener(self)
+
+
+TPoolMember = TypeVar("TPoolMember", bound=PoolMember)
+
+
+class PoolManager(abc.ABC, Generic[TPoolID, TIntPoolID, TPoolMember]):
```
wouterdb, line 148: I failed to type this in a way that binds the pool member and the ids entirely correctly

src/inmanta/agent/forking_executor.py
```diff
@@ -685,179 +874,240 @@ def init_once(cls) -> None:
...
+                        my_executor.failed_resources[rtype] = failed_resource_result.exception
+
+        # FIXME: recovery. If loading failed, we currently never rebuild https://github.com/inmanta/inmanta-core/issues/7695
+        return my_executor
+
+    async def create_member(self, executor_id: executor.ExecutorId) -> MPExecutor:
+        process = await self.process_pool.get(executor_id.blueprint)
+        # TODO: we have a race here: the process can become empty between these two calls
+        result = await process.get(executor_id)
+        self.agent_map.get(executor_id.agent_name).add(result)
```
wouterdb, line 1073: type error I don't know how to resolve

```diff
src/inmanta/agent/forking_executor.py:1093: error: Item "None" of "set[MPExecutor] | None" has no attribute "add"  [union-attr]
src/inmanta/agent/forking_executor.py:1097: error: Item "None" of "set[MPExecutor] | None" has no attribute "discard"  [union-attr]

```

src/inmanta/agent/resourcepool.py
```diff
@@ -0,0 +1,395 @@
...
+    async def _create_or_replace(self, member_id: TPoolID, interal_id: TIntPoolID) -> TPoolMember:
+        """
+        MUST BE CALLED UNDER LOCK!
+        """
+        await self.pre_create_capacity_check(member_id)
+
+        my_executor = await self.create_member(member_id)
+        assert my_executor.id == interal_id
+        self.pool[interal_id] = my_executor
+        my_executor.termination_listeners.append(self.notify_member_shutdown)
```
wouterdb, line 282: type error 


```
src/inmanta/agent/resourcepool.py:145: error: Missing type parameters for generic type "PoolMember"  [type-arg]
src/inmanta/agent/resourcepool.py:270: error: Argument 1 to "append" of "list" has incompatible type "Callable[[TPoolMember], Coroutine[Any, Any, bool]]"; expected "Callable[[PoolMember[TPoolID]], Coroutine[Any, Any, Any]]"  [arg-type]
src/inmanta/agent/resourcepool.py:274: error: Missing type parameters for generic type "PoolMember"  [type-arg]
```

src/inmanta/agent/forking_executor.py
hlloreda: The value of the dict could be saved in a variable and asserted to be not None.

tests/forking_agent/test_venv_manager.py
```diff
@@ -0,0 +1,141 @@
...
+    # Define two sets of requirements, identical but in different orders
+    requirements1 = ("pkg1", "pkg2")
+    requirements2 = ("pkg2", "pkg1")
+
+    blueprint1 = executor.EnvBlueprint(pip_config=pip_config, requirements=requirements1)
+    blueprint2 = executor.EnvBlueprint(pip_config=pip_config, requirements=requirements2)
+
+    hash1 = blueprint1.blueprint_hash()
+    hash2 = blueprint2.blueprint_hash()
+    print(hash1)
```
hlloreda, line 44: Leftover?

hlloreda (changes requested):

src/inmanta/agent/forking_executor.py
wouterdb: It is a bit silly default dict can't be typed more strictly, https://github.com/python/typeshed/tree/main/stdlib/collections#L399 because the factory argument is optional

src/inmanta/agent/resourcepool.py
```diff
@@ -0,0 +1,384 @@
...
+        """Convert an external id to an internal id"""
+        pass
+
+    def my_name(self) -> str:
+        """Method to improve logging output by naming this executor"""
+        return "PoolManager"
+
+    def render_id(self, member: TPoolID) -> str:
+        """Method to improve logging output by naming external ids"""
+        return "PoolMember"
```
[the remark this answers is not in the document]
wouterdb: This is simply a default that is not very informative. It makes it easier to start using this class and then refine later.

src/inmanta/agent/forking_executor.py
```diff
@@ -337,66 +447,84 @@ async def call(self, context: ExecutorContext) -> typing.Sequence[inmanta.loader
...
 
-    def __init__(self, version: int) -> None:
+    def __init__(self, agent_name: str, version: int) -> None:
         self.version = version
+        self.agent_name = agent_name
 
     async def call(self, context: ExecutorContext) -> None:
-        assert context.executor is not None
-        await context.executor.open_version(self.version)
+        await context.get(self.agent_name).open_version(self.version)
```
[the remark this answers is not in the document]
wouterdb: Good question. 

Logically, it should never happen, so I didn't think about it too much.
KeyError is correct, but not very informative I must admit. 
What would you suggest?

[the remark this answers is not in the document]
hlloreda: I think we could raise a `LookupError` to provide more information about the current context

src/inmanta/agent/forking_executor.py
```diff
@@ -685,179 +903,235 @@ def init_once(cls) -> None:
...
+        """
+        super().__init__(
+            retention_time=inmanta.agent.config.agent_executor_retention_time.get(),
+        )
+
+        self.process_pool = MPPool(thread_pool, session_gid, environment, log_folder, storage_folder, log_level, cli_log)
+
+        self.environment = environment
+
+        typing.DefaultDict
```
hlloreda, line 1031: Leftover

src/inmanta/agent/forking_executor.py
```diff
@@ -685,179 +906,239 @@ def init_once(cls) -> None:
...
-    async def join(self, thread_pool_finalizer: list[concurrent.futures.ThreadPoolExecutor], timeout: float) -> None:
-        await super().join(thread_pool_finalizer, timeout)
-        await self.environment_manager.join(thread_pool_finalizer, timeout)
+    async def request_shutdown(self) -> None:
+        await self.process_pool.request_shutdown()
+        await super().request_shutdown()
 
-        thread_pool_finalizer.append(self.thread_pool)
-        await asyncio.gather(*(child.join(timeout) for child in self.children))
+    async def join(self, thread_pool_finalizer: list[ThreadPoolExecutor] = [], timeout: float = 0.0) -> None:
```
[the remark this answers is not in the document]
wouterdb: In general yes, but here, there is no good solution here I'm afraid:

1. I need a default to splice the two superclasses together
2. if I don't want a type violation, it has to be a list
3. it is absolutely not used at all (and there is a comment below it to explain that)

hlloreda (approved):

PR #7970, opened 2024-08-13
Improved config parser output

wouterdb:
# Description

Override code from the python sdk

Original

```
    def _validate_value_types(self, *, section="", option="", value=""):
        """Raises a TypeError for non-string values.

        The only legal non-string value if we allow valueless
        options is None, so we need to check if the value is a
        string if:
        - we do not allow valueless options, or
        - we allow valueless options but the value is not None

        For compatibility reasons this method is not used in classic set()
        for RawConfigParsers. It is invoked in every case for mapping protocol
        access and in ConfigParser.set().
        """
        if not isinstance(section, str):
            raise TypeError("section names must be strings")
        if not isinstance(option, str):
            raise TypeError("option keys must be strings")
        if not self._allow_no_value or value:
            if not isinstance(value, str):
                raise TypeError("option values must be strings")
```

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/config.py
+++ b/src/inmanta/config.py
@@ -58,6 +58,7 @@ class LenientConfigParser(ConfigParser):
             raise TypeError(f"option keys must be strings, instead received {option} of type {type(option)}")
         if not isinstance(value, str):
             raise TypeError(f"option values must be strings, instead received {value} of type {type(value)}")
+        super()._validate_value_types(section=section, option=option, value=value)
 
 
 class Config:
--- a/src/inmanta/config.py
+++ b/src/inmanta/config.py
@@ -48,7 +48,7 @@ class LenientConfigParser(ConfigParser):
         name = _normalize_name(name)
         return super().optionxform(name)
 
-    def _validate_value_types(self, *, section="", option="", value=""):
+    def _validate_value_types(self, *, section: str = "", option: str = "", value: str = "") -> None:
         """
         Override parent class to get clear exceptions
         """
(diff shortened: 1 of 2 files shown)


src/inmanta/config.py
```diff
@@ -48,6 +48,18 @@ def optionxform(self, name: str) -> str:
         name = _normalize_name(name)
         return super().optionxform(name)
 
+    def _validate_value_types(self, *, section="", option="", value=""):
```
edvgui, line 51:
```suggestion
    def _validate_value_types(self, *, section: str = "", option: str = "", value: str = ""):
```

src/inmanta/config.py
```diff
@@ -48,6 +48,18 @@ def optionxform(self, name: str) -> str:
...
+        Override parent class to get clear exceptions
+        """
+        if not isinstance(section, str):
+            raise TypeError(f"section names must be strings, instead received {section} of type {type(section)}")
+        if not isinstance(option, str):
+            raise TypeError(f"option keys must be strings, instead received {option} of type {type(option)}")
+        if not self._allow_no_value or value:
+            if not isinstance(value, str):
+                raise TypeError(f"option values must be strings, instead received {value} of type {type(value)}")
+
```
edvgui, line 62: Should we call the parent at the end? To stay compatible with any potential later addition?

jptrindade (approved):
Other than Guillaume's comments, I have nothing to remark

src/inmanta/config.py
wouterdb: I wouldn't. We have the behavior we want now. But I could.

edvgui (approved):

PR #7976, opened 2024-08-16
Remove io

bartv:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry


src/inmanta/agent/in_process_executor.py
```diff
@@ -101,9 +100,7 @@ def is_stopped(self) -> bool:
         return self._stopped
 
     async def get_provider(self, resource: Resource) -> HandlerAPI[Any]:
-        provider = await asyncio.get_running_loop().run_in_executor(
-            self.provider_thread_pool, handler.Commander.get_provider, self._cache, self, resource
-        )
+        provider = handler.Commander.get_provider(agent=self, resource=resource)
```
wouterdb, line 103: provider threadpool can be removed I think

wouterdb (changes requested):

wouterdb (approved):

PR #7990, opened 2024-08-21
Issue/drop pkg resources - first part

hlloreda:
# Description

First part that migrates the use of the deprecated library: pkg_resources. This part preserves the venv caching mechanism that allows us to speed up the test suite. The second part should migrate this mechanism

Related to https://github.com/inmanta/inmanta-core/issues/6668

# Self Check:

- [ ] Changelog entry


src/inmanta/main.py
```diff
@@ -182,7 +182,7 @@ def get_table(header: list[str], rows: list[list[str]], data_type: Optional[list
     return table.draw()
 
 
-@with_plugins(iter_entry_points("inmanta.cli_plugins"))
+@with_plugins(importlib_metadata.entry_points(group="inmanta.cli_plugins"))
```
hlloreda, line 185: mypy is raising a warning here:
```
src/inmanta/main.py:185: error: Argument 1 to "with_plugins" has incompatible type "EntryPoints"; expected "Generator[EntryPoint, None, None]"  [arg-type]
```
I'm not sure how to resolve this

hlloreda, line 185: mypy is raising a warning here:
```
src/inmanta/main.py:185: error: Argument 1 to "with_plugins" has incompatible type "EntryPoints"; expected "Generator[EntryPoint, None, None]"  [arg-type]
```
I don't know how should I fix this

wouterdb:
```suggestion
@with_plugins(iter(importlib_metadata.entry_points(group="inmanta.cli_plugins")))
```

like so?

wouterdb: [this doesn't seem to be maintained](https://github.com/click-contrib/click-plugins) I think it is good to figure out why it is here at all (I suspect LSM). 

We may want to vendor it or find a fork if we need it.

wouterdb: https://github.com/click-contrib/click-plugins/issues/37

src/inmanta/module.py
```diff
@@ -39,6 +39,7 @@
 from enum import Enum
```
wouterdb, line 39: Note, I got here, but I'm going to wait with the res until we are certain the ` SafeRequirement` is here to stay, because it makes the review very hard

wouterdb (changes requested):
1. is there a migration guide that was followed? (I don't know if it exists, didn't find it either)

src/inmanta/env.py
```diff
@@ -61,6 +62,17 @@ class PipInstallError(Exception):
     pass
 
 
+def safe_parse_requirement(requirement_name: str) -> Requirement:
+    # Packaging Requirement is not able to parse requirements with comment. Therefore, we need to remove the `comment` part
+    drop_comment = requirement_name.split("#")[0].strip()
+    assert len(drop_comment) > 0, "The name of the requirement cannot be an empty string!"
```
wouterdb, line 68: I expect this to go wrong in many places, I would also make a method to parse an entire file, that can contain empty lines and ` # `

src/inmanta/main.py
wouterdb: yes, click-plugin is just outdated, perhaps we can vendor it?

src/inmanta/agent/in_process_executor.py
```diff
@@ -572,7 +571,7 @@ async def _install(self, blueprint: executor.ExecutorBlueprint) -> None:
             await loop.run_in_executor(
                 self.process.thread_pool,
                 self._env.install_for_config,
-                list(pkg_resources.parse_requirements(blueprint.requirements)),
+                [safe_parse_requirement(requirement_name=e) for e in blueprint.requirements],
```
sanderr, line 574: Does the install command really require canonical requires? Same comment applies in different places.

wouterdb (changes requested):

src/inmanta/file_parser.py
```diff
@@ -86,21 +86,22 @@ def get_content_with_dep_removed(cls, filename: str, remove_dep_on_pkg: str) ->
         if not os.path.exists(filename):
             raise Exception(f"File {filename} doesn't exist")
 
+        removed_dependency = safe_parse_requirement(requirement_name=remove_dep_on_pkg)
```
sanderr, line 89: I'd prefer the canonical name method over the requirement one here.

src/inmanta/file_parser.py
```diff
@@ -86,21 +86,22 @@ def get_content_with_dep_removed(cls, filename: str, remove_dep_on_pkg: str) ->
...
-                        if Requirement.parse(line_continuation_buffer).key != remove_dep_on_pkg:
+                        if safe_parse_requirement(requirement_name=line_continuation_buffer).name != removed_dependency.name:
                             result += line_continuation_buffer
                         line_continuation_buffer = ""
                 elif not line.strip() or line.strip().startswith("#"):
                     result += line
                 elif line.endswith("\\"):
                     line_continuation_buffer = line
-                elif Requirement.parse(line).key != remove_dep_on_pkg.lower():
+                elif safe_parse_requirement(requirement_name=line).name != removed_dependency.name:
```
sanderr, line 104: Same here

src/inmanta/module.py
```diff
@@ -107,15 +109,15 @@ def __init__(self, requirement: Requirement) -> None:
     @property
     def project_name(self) -> str:
```
sanderr, line 110: I'm inclined to follow `Requirement` here, i.e. add `name` and deprecate `project_name` and `key`. Perhaps make `self._requirement: CanonicalRequirement` (though we probably should not change the constructor argument, we'll have to convert it).
If we don't, we'll need to have another look at the diff here, because chaning `key` to `name` doesn't hold up.

src/inmanta/module.py
```diff
@@ -124,7 +126,7 @@ def __eq__(self, other: object) -> bool:
         return self._requirement == other._requirement
 
     def __contains__(self, version: str) -> bool:
```
sanderr, line 128:
```suggestion
    def __contains__(self, version: version.Version | str) -> bool:
```

src/inmanta/module.py
```diff
@@ -134,7 +136,7 @@ def __hash__(self) -> int:
 
     @property
     def specs(self) -> Sequence[tuple[str, str]]:
```
sanderr, line 138: We could probably phase this out if we wanted. But I'll leave the decision up to you.

src/inmanta/module.py
```diff
@@ -690,12 +692,13 @@ def get_installed_version(cls, module_name: str) -> Optional[version.Version]:
...
-            dist: Distribution = pkg_resources.get_distribution(ModuleV2Source.get_package_name_for(module_name))
-            return version.Version(dist.version)
-        except DistributionNotFound:
+            dist: Distribution = Distribution.from_name(ModuleV2Source.get_package_name_for(module_name))
+            try:
+                return version.Version(dist.version)
+            except version.InvalidVersion:
+                raise InvalidModuleException(f"Package {dist.name} was installed but it has no valid version.")
+        except PackageNotFoundError:
             return None
```
sanderr, line 701: Why not simply use `importlib.metadata.version()` here?

src/inmanta/module.py
```diff
@@ -1175,14 +1178,16 @@ def has_requirement_for(self, pkg_name: str) -> bool:
         Returns True iff this requirements.txt file contains the given package name. The given `pkg_name` is matched
         case insensitive against the requirements in this RequirementsTxtFile.
         """
-        return any(r.key == pkg_name.lower() for r in self._requirements)
+        return any(r.name == pkg_name for r in self._requirements)
```
sanderr, line 1181: This is a breaking change, isn't it?

src/inmanta/module.py
sanderr: I just realized it might hold up because you just don't have the concept of an explicit canonical requirement yet. But this makes it clear to me that there really is a need for it, so we always know what assumptions we can safely make about `r.name`.

sanderr (changes requested):
I think we can remove the constraint on `setuptools` from `setup.py` now? Or is it still relevant for the final `pkg_resources` parts? Iirc it was related to the `InvalidRequirement` exception, so if we don't use it anymore we should be good.

src/inmanta/env.py
```diff
@@ -197,24 +208,32 @@ def _are_installed_recursive(
...
                 # the marker evaluation fail.
                 environment_marker_evaluation = {"extra": contained_in_extra} if contained_in_extra else None
                 if r.marker and not r.marker.evaluate(environment=environment_marker_evaluation):
                     # The marker of the requirement doesn't apply on this environment
                     continue
-                if r.key not in installed_packages or str(installed_packages[r.key]) not in r:
+                if (
+                    r.name.lower() not in installed_packages
+                    or (len(r.specifier) > 0 and str(installed_packages[r.name.lower()]) not in r.specifier)
+                    # If no specifiers are provided, the `in` operation will return `False`
```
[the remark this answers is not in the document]
hlloreda: With an existing version, it doesn't work
Screenshot from 2024-09-02 09-13-15
Screenshot from 2024-09-02 09-12-57
Screenshot from 2024-09-02 09-12-47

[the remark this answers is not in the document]
wouterdb: This is super fishy, could you make a testcase to prove that this is as you say , because my testcase shows the opposite?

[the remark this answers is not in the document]
wouterdb:
```python
>>> packaging.version.Version("1.2.3b2") in packaging.requirements.Requirement('pip').specifier
False
>>> packaging.version.Version("1.2.3") in packaging.requirements.Requirement('pip').specifier
``` 

```
>>> packaging.version.Version("1.2.3b0") in packaging.requirements.Requirement('pip>1.0').specifier
False
>>> packaging.version.Version("1.2.3") in packaging.requirements.Requirement('pip>1.0').specifier
True
>>> 

```

it is not about it being empty, it is about the `b`

src/inmanta/util/__init__.py
```diff
@@ -863,3 +868,66 @@ def check_for_pool_exhaustion(self) -> None:
...
+
+
+def remove_comment_part(to_clean: str) -> str:
+    """
+    Remove the comment part of a given string and ensure that the lenght of the string is greater than 0
+
+    :param to_clean: The string to clean
+    :return: A cleaned string
+    """
+    drop_comment, _, _ = to_clean.partition("#")
```
wouterdb, line 880: This assumed a requires can't contain a '#' , it can (as ugly as it is)
https://peps.python.org/pep-0508/ 

to refer to the spec: https://pip.pypa.io/en/latest/reference/requirements-file-format/#comments

```
A line that begins with # is treated as a comment and ignored. Whitespace followed by a # causes the # and the remainder of the line to be treated as a comment.
```

src/inmanta/util/__init__.py
```diff
@@ -863,3 +868,66 @@ def check_for_pool_exhaustion(self) -> None:
...
+    if len(drop_comment) == 0:
+        raise ValueError("The name of the requirement cannot be an empty string!")
```
wouterdb, line 882: what with a line containing only a comment>

wouterdb (changes requested):

src/inmanta/util/__init__.py
hlloreda: Indeed, I missed that a # could be part of a requirement name

src/inmanta/env.py
```diff
@@ -61,6 +62,17 @@ class PipInstallError(Exception):
...
 
 
+def safe_parse_requirement(requirement_name: str) -> Requirement:
+    # Packaging Requirement is not able to parse requirements with comment. Therefore, we need to remove the `comment` part
+    drop_comment = requirement_name.split("#")[0].strip()
+    assert len(drop_comment) > 0, "The name of the requirement cannot be an empty string!"
+    # We canonicalize the name of the requirement to be able to compare requirements and check if the requirement is
+    # already installed
+    requirement = Requirement(requirement_string=drop_comment)
+    requirement.name = utils.canonicalize_name(requirement.name)
```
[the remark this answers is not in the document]
wouterdb: I don't see clean a way to make that work? 

The copy will have the same problem?

mypy-baseline.txt
```diff
@@ -892,13 +893,8 @@ src/inmanta/execute/scheduler.py:0: error: Missing type parameters for generic t
 src/inmanta/execute/scheduler.py:0: error: "<subclass of "DelayedResultVariable" and "RelationAttributeVariable">" has no attribute "attribute"  [attr-defined]
 src/inmanta/execute/scheduler.py:0: error: "<subclass of "DelayedResultVariable" and "RelationAttributeVariable">" has no attribute "attribute"  [attr-defined]
 src/inmanta/execute/scheduler.py:0: error: "<subclass of "DelayedResultVariable" and "RelationAttributeVariable">" has no attribute "attribute"  [attr-defined]
+src/inmanta/env.py:0: error: Argument 1 to "parse_requirements" has incompatible type "Sequence[Requirement]"; expected "Sequence[str]"  [arg-type]
```
wouterdb, line 896: this looks suspicious?

src/inmanta/module.py
```diff
@@ -96,52 +97,49 @@ class InmantaModuleRequirement:
...
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, InmantaModuleRequirement):
             return NotImplemented
         return self._requirement == other._requirement
 
-    def __contains__(self, version: str) -> bool:
-        return version in self._requirement
+    def __contains__(self, version: packaging.version.Version | str) -> bool:
+        return version in self._requirement.specifier if len(self._requirement.specifier) > 0 else True
```
wouterdb, line 135: this looks like the same issue as before with the pre-release??

src/inmanta/util/__init__.py
```diff
@@ -863,3 +871,118 @@ def check_for_pool_exhaustion(self) -> None:
...
+    if not file_path.exists():
+        raise RuntimeError(f"The provided path does not exist: `{file_path}`!")
+
+    requirements = []
+    with open(file_path) as f:
+        for line in f.readlines():
+            try:
+                requirements.append(parse_requirement(line))
+            except ValueError:
+                LOGGER.warning("This line was skipped because the requirement could not be parsed: %s", line)
```
wouterdb, line 945: This should not be a warning. This also happens if you have a comment in the file. This will drown the log in warnings about things that are perfectly normal.

src/inmanta/util/__init__.py
```diff
@@ -863,3 +871,118 @@ def check_for_pool_exhaustion(self) -> None:
 
     def _reset_counter(self) -> None:
         self._exhausted_pool_events_count = 0
+
+
+def remove_comment_part(to_clean: str) -> str:
```
wouterdb, line 876: I want unit tests for all of these parsing methods.

src/inmanta/util/__init__.py
```diff
@@ -863,3 +871,118 @@ def check_for_pool_exhaustion(self) -> None:
...
+    """
+    Remove the comment part of a given string and ensure that the length of the string is greater than 0
+
+    :param to_clean: The string to clean
+    :return: A cleaned string
+    """
+    # Refer to PEP 508. A requirement could contain a hashtag
+    to_clean = to_clean.strip()
+    if to_clean.startswith("#"):
+        raise ValueError("The name of the requirement cannot be a comment!")
```
wouterdb, line 886: Why can't we have lines with only comments, as described in the pep?

This method should document when it raises an exception, because now it is super confusing.

Also, using the same exception for errors and comments is very weird to me.

src/inmanta/util/__init__.py
```diff
@@ -863,3 +871,118 @@ def check_for_pool_exhaustion(self) -> None:
...
+    if len(drop_comment) == 0:
+        raise ValueError("The name of the requirement cannot be an empty string!")
+    return drop_comment
+
+
+CanonicalRequirement = typing.NewType("CanonicalRequirement", packaging.requirements.Requirement)
+
+
+def parse_requirement(requirement: str) -> CanonicalRequirement:
+    """
```
wouterdb, line 899: it should be documented if this method handles comments or not and how it handles them.

src/inmanta/util/__init__.py
```diff
@@ -863,3 +871,118 @@ def check_for_pool_exhaustion(self) -> None:
...
+    # We canonicalize the name of the requirement to be able to compare requirements and check if the requirement is
+    # already installed
+    # This instance is considered as doomed because the requirement name that this instance is using is not "canonicalized"
+    requirement_instance = packaging.requirements.Requirement(requirement_string=drop_comment)
+    requirement_instance.name = packaging.utils.canonicalize_name(requirement_instance.name)
+    canonical_requirement_instance = CanonicalRequirement(requirement_instance)
+    return canonical_requirement_instance
+
+
+def parse_requirements(requirements: Sequence[str]) -> list[CanonicalRequirement]:
```
wouterdb, line 917: it should be documented if this method handles comments or not and how it handles them.

src/inmanta/util/__init__.py
```diff
@@ -863,3 +871,118 @@ def check_for_pool_exhaustion(self) -> None:
...
+    Parameters
+    ----------
+    plugins : iter
+        An iterable producing one `pkg_resources.EntryPoint()` per iteration.
+    attrs : **kwargs, optional
+        Additional keyword arguments for instantiating `click.Group()`.
+
+    Returns
+    -------
+    click.Group()
```
wouterdb, line 966: I would like to have normal docs, not google style

tests/compiler/test_basics.py
```diff
@@ -743,6 +743,15 @@ def test_implementation_import_missing_error(snippetcompiler) -> None:
     assert exception.value.location.start_char == 20
 
 
+@pytest.mark.parametrize("name", ["", "#", " # ", "#this is a comment"])
```
wouterdb, line 746: 1. I think tests for methods in util should not get into the compiler folder
2.  I would also test positive examples
3. I would also test the handling of lists or files that should ignore comments

wouterdb (changes requested):

mypy-baseline.txt
hlloreda: I don't know why the line is not reported correctly but the actual error is:
```
src/inmanta/env.py:176: error: Argument 1 to "parse_requirements" has incompatible type "Sequence[Requirement]"; expected "Sequence[str]"  [arg-type]
```

This is the code triggering this error:
```
    def _get_as_requirements_type(cls, requirements: req_list) -> Sequence[packaging.requirements.Requirement]:
        """
        Convert requirements from Union[Sequence[str], Sequence[Requirement]] to Sequence[Requirement]
        """
        if isinstance(requirements[0], str):
            return util.parse_requirements(requirements)
        else:
            return requirements
```

src/inmanta/util/__init__.py
hlloreda: I moved the raising of the error to the caller: makes more sense and was probably creating more confusion than necessary

I agree that different exception errors should be used, even though in those particular cases, they could be considered a ValueError.

src/inmanta/util/__init__.py
```diff
@@ -922,7 +924,16 @@ def parse_requirements(requirements: Sequence[str]) -> list[CanonicalRequirement
     :param requirements: The names of the different requirements
```
wouterdb, line 924:
```suggestion
    
    This function supposes to receive actual requirements. Commented strings will not be handled and result in a ValueError

    :param requirements: The names of the different requirements
```

wouterdb (approved):

src/inmanta/util/__init__.py
```diff
@@ -863,3 +871,110 @@ def check_for_pool_exhaustion(self) -> None:
...
+    To be able to compare requirements, we need to make sure that every requirement's name is canonicalized otherwise issues
+    could arise when checking if packages are installed in a particular Venv.
+    This function supposes to receive an actual requirement. Commented strings will not be handled and result in a ValueError
+
+    :param requirement: The requirement's name
+    :return: A new requirement instance
+    :raise: ValueError when commented or empty strings are provided
+    """
+    # Packaging Requirement is not able to parse requirements with comment. Therefore, we need to remove the `comment` part
+    drop_comment = remove_comment_part(to_clean=requirement)
```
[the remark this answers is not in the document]
hlloreda: I wanted to say both things:
- `lorem # nice package` is not supported
- And `# lorem`
image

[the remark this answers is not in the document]
sanderr: But you do support `lorem # nice package` here by stripping the comment. And it is in fact allowed in a requirement file, so it's good that you do. Though, as stated in the comment, I don't think this is the place to do it.

src/inmanta/env.py
```diff
@@ -1000,15 +984,15 @@ def get_protected_inmanta_packages(cls) -> list[str]:
...
         to make sure that no Inmanta packages gets overridden.
         """
         protected_inmanta_packages: list[str] = cls.get_protected_inmanta_packages()
-        workingset: dict[str, version.Version] = PythonWorkingSet.get_packages_in_working_set()
+        workingset: dict[str, packaging.version.Version] = PythonWorkingSet.get_packages_in_working_set()
         return [
-            safe_parse_requirement(requirement_name=f"{pkg}=={workingset[pkg]}")
+            util.parse_requirement(requirement=f"{pkg}=={workingset[pkg]}")
             for pkg in workingset
             if pkg in protected_inmanta_packages
```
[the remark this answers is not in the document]
hlloreda: I'm not sure to understand the end of your comment, `get_protected_inmanta_package_names` is already returning canonicalized names
image

src/inmanta/module.py
```diff
@@ -98,22 +97,29 @@ class InmantaModuleRequirement:
             used by distinguishing the two on a type level.
     """
 
-    def __init__(self, requirement: Requirement) -> None:
+    def __init__(self, requirement: util.CanonicalRequirement) -> None:
```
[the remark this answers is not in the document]
hlloreda: I didn't find any occurrence in `pytest-inmanta` and `pytest-inmanta-lsm`:
image
image

Makefile
```diff
@@ -6,7 +6,7 @@ black = black src tests tests_common
 .PHONY: install
 install:
 	pip install -U setuptools pip uv -c requirements.txt
-	uv pip install -U -e . -c requirements.txt -r requirements.dev.txt
+	uv pip install -U -e . -c requirements.txt -r requirements.txt -r requirements.dev.txt
```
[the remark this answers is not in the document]
hlloreda: I think I was confused that requirements.txt is a constraint file and not a requirement file (perhaps naming it constraints.txt would have made more sense?). Constraints will not necessarily be installed, only if they are required.

mypy-baseline.txt
```diff
@@ -892,13 +893,8 @@ src/inmanta/execute/scheduler.py:0: error: Missing type parameters for generic t
 src/inmanta/execute/scheduler.py:0: error: "<subclass of "DelayedResultVariable" and "RelationAttributeVariable">" has no attribute "attribute"  [attr-defined]
 src/inmanta/execute/scheduler.py:0: error: "<subclass of "DelayedResultVariable" and "RelationAttributeVariable">" has no attribute "attribute"  [attr-defined]
 src/inmanta/execute/scheduler.py:0: error: "<subclass of "DelayedResultVariable" and "RelationAttributeVariable">" has no attribute "attribute"  [attr-defined]
+src/inmanta/env.py:0: error: Argument 1 to "parse_requirements" has incompatible type "Sequence[CanonicalRequirement]"; expected "Sequence[str]"  [arg-type]
```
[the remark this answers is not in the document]
hlloreda: This should be the same as https://github.com/inmanta/inmanta-core/pull/7990#discussion_r1763307866

Makefile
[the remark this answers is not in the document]
sanderr: I agree that the naming is confusing. But it's also the convention, so naming it differently would also be confusing.

src/inmanta/env.py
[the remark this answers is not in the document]
sanderr: But that's not the method I'm referring to. Am I missing something?

src/inmanta/agent/in_process_executor.py
hlloreda: So the motivation behind this is the following: At some point, the requirements will be checked (to see if they are already installed), we will need canonical requirements for this.
- `install_from_list` depends on `install_from_index`
- `install_from_index` depends on `Install_from_config`
- `Install_from_config` will check if the provided requirements are installed

sanderr: Ok, it's unfortunate that we have to lift it to the user interface, but it's probably the lesser of two evils.

src/inmanta/util/__init__.py
```diff
@@ -888,37 +887,37 @@ def remove_comment_part(to_clean: str) -> str:
     return drop_comment
 
 
+"""
+A CanonicalRequirement is a packaging.requirements.Requirement except that the name of this Requirement is canonicalized, which
+allows us to compare requirements without dealing afterwards with the format of these requirements.
```
sanderr, line 892:
```suggestion
allows us to compare names without dealing afterwards with the format of these requirements.
```

Requirements can be compared regardless because `packaging` compares canonical forms.

src/inmanta/util/__init__.py
```diff
@@ -888,37 +887,37 @@ def remove_comment_part(to_clean: str) -> str:
...
+"""
```
sanderr, line 893: Conventionally this goes below the name.

src/inmanta/util/__init__.py
```diff
@@ -888,37 +887,37 @@ def remove_comment_part(to_clean: str) -> str:
...
-    drop_comment = remove_comment_part(to_clean=requirement)
-
-    if drop_comment.startswith("#") or len(drop_comment) == 0:
-        raise ValueError(f"The requirement is invalid: Cannot be a comment or empty -> `{drop_comment}`!")
-
     # We canonicalize the name of the requirement to be able to compare requirements and check if the requirement is
     # already installed
-    # This instance is considered as doomed because the requirement name that this instance is using is not "canonicalized"
-    requirement_instance = packaging.requirements.Requirement(requirement_string=drop_comment)
+    # /!\ The following line could cause issue because we are not supposed to modify fields of an existing instance
```
sanderr, line 908: Ok, I see now. Perhaps make it a bit less alarming by expanding?
```suggestion
    # /!\ The following line could cause issue because we are not supposed to modify fields of an existing instance
    # The version of packaging is constrained to ensure this can not cause problems in production.
```

Is the "/!\" intentional?

sanderr (changes requested):
Final (minor) notes

src/inmanta/env.py
```diff
@@ -43,10 +43,11 @@
 
 import pkg_resources
 
+import inmanta.util
```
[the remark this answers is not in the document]
sanderr: (not a change request per se, just wondering)

sanderr:
@wouterdb I caused quite some changes to be made after your review so I'm rerequesting to be sure you're OK with all of it.

src/inmanta/env.py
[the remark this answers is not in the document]
hlloreda: Just as a precaution, given that we have multiple files named `util`

src/inmanta/util/__init__.py
hlloreda: The "/!\" was intentional but maybe a bit too alarmist, I have removed it

src/inmanta/env.py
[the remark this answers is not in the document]
sanderr: Typically we (at least I do and I think the rest of the team does the same) use partially qualified `util` for `inmanta.util`, fully qualified for anything external, and in the rare case of a nested util like `inmanta.db.util`, I might use the partially qualified `db.util`.

src/inmanta/util/__init__.py
sanderr: Oh, it's a danger triangle? I was very confused by the slashes.

sanderr (approved):

wouterdb (approved):
Didn't check all details, looks ok

src/inmanta/module.py
```diff
@@ -107,7 +105,7 @@ def __init__(self, requirement: packaging.requirements.Requirement) -> None:
                 f"Problematic case: {str(requirement)}"
             )
         self._requirement: inmanta.util.CanonicalRequirement = inmanta.util.parse_requirement(str(requirement))
-        self._project_name = re.sub('[^A-Za-z0-9.]+', '-', requirement.name)
+        self._project_name = re.sub("[^A-Za-z0-9.]+", "-", requirement.name)
```
hlloreda, line 108: Maybe it would be better to only receive a string?

wouterdb (approved):

sanderr (approved):
As discussed

src/inmanta/module.py
```diff
@@ -104,13 +104,11 @@ def __init__(self, requirement: inmanta.util.CanonicalRequirement) -> None:
                 f"InmantaModuleRequirement instances work with inmanta module names, not python package names. "
                 f"Problematic case: {str(requirement)}"
             )
-        self._requirement: inmanta.util.CanonicalRequirement = requirement
+        self._requirement = inmanta.util.parse_requirement(str(requirement))
```
sanderr, line 107: This is still a change. There's no need to parse it again, is there?

sanderr (changes requested):

sanderr (approved):

hlloreda:
Everything has been merged (ISO7 and master)

PR #7998, opened 2024-08-23
Mock out logfire when disabled 

wouterdb:
# Description

Draft at replacing logfire with no-ops 

# Self Check:

- [x] Changelog entry


src/inmanta/protocol/rest/__init__.py
```diff
@@ -631,7 +630,7 @@ async def _execute_call(
                 arguments.auth_username if arguments.auth_username else "<>",
             )
 
-            with logfire.span("Calling method " + config.method_name, **arguments.call_args):
```
wouterdb, line 634: I really don't like this kwargs at the end, because they can allow injection of arguments and are impossible to type.

src/inmanta/tracing.py
```diff
@@ -40,15 +45,83 @@ def configure_logfire(service: str) -> None:
...
+    *,
+    _tags: Sequence[str] | None = None,
+    _span_name: str | None = None,
+    _level: LevelName | None = None,
+    **attributes: Any,
+) -> LogfireSpan:
+    if enabled:
+        return logfire.span(msg_template, _tags=_tags, _span_name=_span_name, _level=_level, **attributes)
+    else:
+        return no_span
```
wouterdb, line 87: this is a type violation inherited from  logfire:

```
src/inmanta/tracing.py:87: error: Incompatible return value type (got "NoopSpan", expected "LogfireSpan")  [return-value]
```

src/inmanta/tracing.py
```diff
@@ -16,16 +16,25 @@
...
 import logging
 import os
+from typing import Any, Callable, ContextManager, LiteralString, Mapping, ParamSpec, Sequence, TypeVar
 
 import logfire
+import logfire._internal.config
 import logfire.integrations
 import logfire.integrations.pydantic
+from logfire import LevelName, LogfireSpan
+from logfire._internal.main import NoopSpan
```
wouterdb, line 29: This internal import can be avoided by copy-pasting the class.

arnaudsjs (approved):

bartv (approved):

PR #8001, opened 2024-08-23
resource scheduler: implement first stage

### LINKED ISSUE #7548: agent rework: deploy sequencing first steps
See [design meeting slides](https://docs.google.com/presentation/d/1rj-hrtZ1_ZyoDW-HCmMT4HafTw2i8fzcm_8FDhaDjzY/edit#slide=id.p), especially slide 8 (Sequencing: first iteration).

Wouter already made a part of it here: https://github.com/inmanta/inmanta-core/pull/7561

sanderr:
# Description

First stage of the resource scheduler (#8007, [ZenHub epic](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/8007)). Focus is mainly on the data structures for model state and scheduled work, and the algorithms to update that state when a new version comes in or as tasks are executed.

There is still some polishing to do. I propose to delegate that to #8008. That means that the focus of this review is on the stability of the server and agent interfaces, so that work on those will not conflict with further refinements in #8008.

Of course, feel free to already leave feedback for the other parts, I'll make sure to include it in #8008. I've already marked a lot of notes to self in the code as `FIXME[#8008]`.

Please make sure to also consider the module structure. Do you agree that `inmanta.deploy` makes sense as a namespace?

closes #7548

# Self Check:

- [x] Changelog entry
- [x] ~~Sufficient test cases (reproduces the bug/tests the requested feature)~~ -> #8008


src/inmanta/deploy/scheduler.py
sanderr, line 1: I'd suggest to start your review here and to follow the flow from there.

src/inmanta/deploy/scheduler.py
```diff
@@ -0,0 +1,163 @@
...
+        # TODO: this lock may have little value left now. Consider.
+        self._scheduler_lock: asyncio.Lock = asyncio.Lock()
+        # - lock to serialize scheduler state updates (i.e. process new version)
+        self._update_lock: asyncio.Lock = asyncio.Lock()
+
+    def start(self) -> None:
+        # TODO (ticket): read from DB instead
+        pass
+
+    async def deploy(self) -> None:
```
wouterdb, line 54: this I assume is the external deploy trigger (i.e. current deploy button)

src/inmanta/deploy/scheduler.py
```diff
@@ -0,0 +1,163 @@
...
+    # TODO: name
+    # TODO (ticket): design step 2: read new state from DB instead of accepting as parameter (method should be notification only, i.e. 0 parameters)
+    async def new_version(
+        self,
+        version: int,
+        resources: Mapping[ResourceIdStr, ResourceDetails],
+        requires: Mapping[ResourceIdStr, Set[ResourceIdStr]],
+    ) -> None:
+        async with self._update_lock:
+            # Inspect new state and mark resources as "update pending" where appropriate. Since this method is the only writer
```
wouterdb, line 72: neat

src/inmanta/deploy/scheduler.py
```diff
@@ -0,0 +1,163 @@
...
+            # for "update pending", and a stale read is acceptable, we can do this part before acquiring the exclusive scheduler
+            # lock.
+            # TODO: what to do when an export changes handler code without changing attributes? Consider in deployed state? What
```
wouterdb, line 75: desires state is considered to be defined by the attributes and nothing else. 

Given that updates are rare, we can keep the behavior like that. 
(Apart from that argument, I would keep the behavior as it is, but I'm not going to linger on this)

src/inmanta/deploy/scheduler.py
```diff
@@ -0,0 +1,163 @@
...
+            #   does current implementation do?
+            deleted_resources: Set[ResourceIdStr] = self._state.resources.keys() - resources.keys()
+            for resource in deleted_resources:
+                self._work.delete_resource(resource)
+
+            # TODO: make sure this part (before scheduler lock is acquired) doesn't block queue until lock is acquired: either
+            # run on thread or make sure to regularly pass control to IO loop (preferred)
```
wouterdb, line 82: running it on a thread requires locking and usage of concurrency safe data structures, which can be hard to get any benefit out off (definitely in python)

src/inmanta/util/collections.py
```diff
@@ -0,0 +1,167 @@
...
+            m[k] = v
+        time_updated: float = time.process_time()
+
+        print(
+            f"Initial population with {self.size} items with {self.nb_targets} each: {time_populated - time_start},"
+            f" update: of {self.nb_updates} items: {time_updated - time_populated}."
+        )
+
+
+ScalingTest(size=100, nb_targets=10, nb_updates=10).test()
```
wouterdb, line 160: How does it perform?

src/inmanta/util/collections.py
```diff
@@ -0,0 +1,167 @@
...
+K = TypeVar("K", bound=Hashable)
+V = TypeVar("V", bound=Hashable)
+
+
+# TODO: review
+# TODO: unit tests
+
+
+# TODO: better name?
+class BidirectionalManyToManyMapping(MutableMapping[P, Set[S]], Generic[P, S]):
```
wouterdb, line 35: Cool! 

(If we are very looking for a candidate to turn into C or or Rust, might be this one)

src/inmanta/deploy/work.py
```diff
@@ -0,0 +1,424 @@
...
+        Removes the given task from its associated agent queue if it is present.
+        Returns the priority at which the deleted task was queued, if it was at all.
+        """
+        try:
+            return self.remove(task)
+        except KeyError:
+            return None
+
+    def _queue_item_for_task(self, task: Task) -> Optional[TaskQueueItem]:
+        return self._tasks_by_resource.get(task.resource, {}).get(task, None)
```
wouterdb, line 148: Why not make Tasks the same as TaskQueueItems or carry a reference to them?

src/inmanta/deploy/work.py
```diff
@@ -0,0 +1,424 @@
...
+            return NotImplemented
+        return (self.task.priority, self.insert_order) < (other.task.priority, other.insert_order)
+
+
+class AgentQueues(Mapping[Task, PrioritizedTask[Task]]):
+    """
+    Per-agent priority queue for ready-to-execute tasks. Clients must not interact with the underlying priority queues directly,
+    only through the queue manipulation methods offered by this class.
+
+    Each Task will be queued at most once. If an identical Task is queued later on, it will not be queued separately. It may
```
wouterdb, line 102: elegant

src/inmanta/deploy/work.py
```diff
@@ -0,0 +1,424 @@
...
+    #########################
+
+    def __getitem__(self, key: Task) -> PrioritizedTask[Task]:
+        return self._tasks_by_resource.get(key.resource, {})[key].task
+
+    def __iter__(self) -> Iterator[Task]:
+        return itertools.chain.from_iterable(self._tasks_by_resource.values())
+
+    def __len__(self) -> int:
+        return sum(len(items) for items in self._tasks_by_resource.values())
```
wouterdb, line 222: Looks expensive, but may not be used often?

src/inmanta/deploy/work.py
```diff
@@ -0,0 +1,424 @@
...
+        added_requires: Optional[Mapping[ResourceIdStr, Set[ResourceIdStr]]] = None,
+        dropped_requires: Optional[Mapping[ResourceIdStr, Set[ResourceIdStr]]] = None,
+    ) -> None:
+        """
+        Update the scheduled work state to reflect the new model state and scheduler intent. May defer tasks that were
+previously considered ready to execute if any of their dependencies are added to the scheduled work.
+
+        :param ensure_scheduled: Set of resources that should be deployed. Adds a deploy task to the scheduled work for each
+            of these, unless it is already scheduled.
+        :param running_deploys: Set of resources for which a deploy is currently running, for the latest desired state, i.e.
```
wouterdb, line 272: this I don't understand

wouterdb (approved):
Cool stuff

src/inmanta/deploy/work.py
sanderr: As I see it, `Task` is the high level declaration that such a resource action should be performed. By containing it to only the declaration of this intent, it can be used for comparison, e.g. to check whether a certain task is already scheduled. Perhaps `Task` is not the best name for it.

That said, I'll consider the value of keeping it like that versus what you suggest. I do think that there is one easy win in the direction you suggest in merging `PrioritizedTask` and `TaskQueueItem` into one. I think they had a separate value in my initial implementation but that ceased to be, so it would be cleaner to merge them. Then we'd only have two levels: the "intent" and the "task".

sanderr: `Task` also allows the queue consumer to work with a simpler interface, i.e. just the intent.

src/inmanta/util/collections.py
sanderr: On my laptop:
```
Initial population with 100 items with 10 each: 0.00030825400000000086, update: of 10 items: 3.2813999999999066e-05.
Initial population with 100000 items with 10 each: 0.6959348740000001, update: of 100 items: 0.0007234259999999715.
Initial population with 100000 items with 10 each: 0.665464032, update: of 1000 items: 0.00753114900000007.
Initial population with 100000 items with 10 each: 0.6574984850000001, update: of 10000 items: 0.077292629.
Initial population with 100000 items with 10 each: 0.7974232150000002, update: of 100000 items: 0.8767623789999996.
Initial population with 1000000 items with 5 each: 6.596546615999998, update: of 1000 items: 0.005180475000003071.
Initial population with 1000000 items with 10 each: 10.285557354999998, update: of 1000 items: 0.009985014000001513.
Initial population with 10000 items with 10000 each: 16.487250724000006, update: of 10 items: 0.014592215999996938.
```

I think the second to last is representative of the scale we're aiming for? 1M resources, with 10 requires each (which is probably a lot for the average), in resources sets of 1000 each. First `put_version` might take 10s, but an update for a single `put_partial` should be snappy at `1ms`.

So (as can be expected) it scales pretty linearly with the size of the resources and the number of requires edges. The fact that it's linear is great, because it means that simply speeding up the raw efficiency (e.g. with Rust as you pointed out in another comment) can do a lot for us as our use case scales.

src/inmanta/deploy/scheduler.py
sanderr: External deploy trigger yes, and deploy button is one case, but it can also be the server's deploy timer triggering it (for now).

src/inmanta/deploy/scheduler.py
sanderr: Indeed, I decided against it but hadn't removed the note yet.

src/inmanta/deploy/work.py
sanderr: Not used at all atm, it's just one of the methods you need to implement the `Mapping` abc.

src/inmanta/deploy/work.py
sanderr: It was a bit of a last-minute addition, may need more refinement. I'll see what I can do in #8008 and if nothing else I'll clarify it there.

sanderr:
@wouterdb I updated the description with some more context.

src/inmanta/deploy/scheduler.py
```diff
@@ -97,50 +105,73 @@ async def new_version(
                 if dropped:
                     self._state.update_pending.add(resource)
                     dropped_requires[resource] = dropped
-
+                # this loop is race-free, potentially slow, and completely synchronous
+                # => regularly pass control to the event loop to not block scheduler operation during update prep
+                await asyncio.sleep(0)
```
wouterdb, line 110: this is a point to keep in mind, once for every resource might be too much, otoh, it might be perfect

wouterdb (approved):

arnaudsjs (dismissed):

src/inmanta/deploy/work.py
```diff
@@ -0,0 +1,431 @@
...
+
+    # FIXME[#8019]: relies on undocumented asyncio.PriorityQueue._queue field and the fact that it's a heapq,
+    #               can we do something about that?
+
+    def __init__(self) -> None:
+        self._agent_queues: dict[str, asyncio.PriorityQueue[TaskQueueItem]] = {}
+        # can not drop tasks from queue without breaking the heap invariant, or potentially breaking asyncio.Queue invariants
+        # => take approach suggested in heapq docs: simply mark as deleted.
+        # => Keep view on all active tasks for a given resource, which also doubles as lookup for client operations
+        # Only tasks in this collection are considered queued, as far as this class' client interface is concerned.
```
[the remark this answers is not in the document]
sanderr: Indeed it does not, but the `heapq` documentation does: https://docs.python.org/3/library/heapq.html#priority-queue-implementation-notes

And indeed, `asyncio.PriorityQueue` doesn't offer a `remove` method or similar.

src/inmanta/deploy/scheduler.py
```diff
@@ -0,0 +1,182 @@
...
+            dropped_requires: dict[ResourceIdStr, Set[ResourceIdStr]] = {}
+            for resource, details in resources.items():
+                if (
+                    resource not in self._state.resources
+                    or details.attribute_hash != self._state.resources[resource].attribute_hash
+                ):
+                    self._state.update_pending.add(resource)
+                    new_desired_state.append(resource)
+                old_requires: Set[ResourceIdStr] = requires.get(resource, set())
+                new_requires: Set[ResourceIdStr] = self._state.requires.get(resource, set())
```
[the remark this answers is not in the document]
sanderr: Yes, thank you.

sanderr:
@arnaudsjs thanks for the review. I committed the bugfix suggestion you commented, but I'm leaving the rest of your comments for #8008 to unstuck @hlloreda for #8009.

src/inmanta/util/collections.py
```diff
@@ -0,0 +1,115 @@
...
+"""
+
+from collections.abc import Hashable, Iterator, Mapping, MutableMapping, Set
+from typing import Generic, Optional, TypeVar
+
+P = TypeVar("P", bound=Hashable)
+S = TypeVar("S", bound=Hashable)
+# type vars not bound to the class
+K = TypeVar("K", bound=Hashable)
+V = TypeVar("V", bound=Hashable)
```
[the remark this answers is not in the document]
sanderr: I found it hard to come up with meaningful names here. I changed it to the conventional `S` and `T` now. I'm not sure comments would be helpful, since the names really don't have much meaning.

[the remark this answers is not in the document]
sanderr: (If you get notified about these comments, no need to look at them now, I'll make sure to link here from the new PR.)

PR #8002, opened 2024-08-26
Mypy baseline

bartv:
# Description

* Short description here *

# Self Check:

- [x] Changelog entry


changelogs/unreleased/enforce-mypy.yml
```diff
@@ -0,0 +1,3 @@
+description: Enforece the use of mypy using mypy baseline
+change-type: patch
+destination-branches: [master]
```
wouterdb, line 3: I think we will need this on iso7 as well, to prevent merge hell. 
(but might be separate MR, due to different baseline )

requirements.dev.txt
```diff
@@ -8,6 +8,7 @@ types-setuptools==74.0.0.20240831
 types-toml==0.10.8.20240310
 flake8-junit-report==2.1.0
 time-machine==2.15.0
+mypy-baseline==0.7.1
```
wouterdb, line 11: I think this in the dev deps

wouterdb: https://github.com/inmanta/inmanta-dev-dependencies/blob/master/pyproject.toml#L26

wouterdb (changes requested):

wouterdb (approved):

Issue #7849, opened 2024-07-11
Code loading problems on packaged V2 modules


wouterdb:
# Problem

When you have a .py file in your files folder. This file will be loaded by the compiler.

# Problem explanation

-  When we package it, we copy the files, templates and model folder into the python package
-  we load all python files in the package (they become sub-packages)
-  we don't allow a model sub-package in any module
-  we do allow a files or template sub package, but it doesn't work in practice 

# Solution

- Don't allow files or templates sub packages either
- Don't load from them

```diff
diff --git a/src/inmanta/module.py b/src/inmanta/module.py
index 35f245254..3c77dfab5 100644
--- a/src/inmanta/module.py
+++ b/src/inmanta/module.py
@@ -2981,10 +2981,11 @@ class Module(ModuleLike[TModuleMetadata], ABC):
 
         files: dict[str, str] = {}
         model_dir_path: str = os.path.join(plugin_dir, "inmanta_plugins", self.name, "model")
+        files_dir_path: str = os.path.join(plugin_dir, "inmanta_plugins", self.name, "files")
 
         for dirpath, dirnames, filenames in os.walk(plugin_dir, topdown=True):
             # Modify dirnames in-place to stop os.walk from descending into any more subdirectories of the model directory
-            if dirpath.startswith(model_dir_path):
+            if dirpath.startswith(model_dir_path) or dirpath.startswith(files_dir_path):
                 dirnames[:] = []
                 continue
 
 ```

PR #8020, opened 2024-08-28
Fix code loading on packaged V2 modules #7849

hlloreda:
# Description

The compiler was loading any python file in the `files`, `model` and `templates` directory. This is no longer the case

closes https://github.com/inmanta/inmanta-core/issues/7849

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/7849-code-loading.yml
+++ b/changelogs/unreleased/7849-code-loading.yml
@@ -1,0 +1,3 @@
+description: Make sure the compiler will not load Python code code present in the `files`, `model` or `templates` directory
+change-type: patch
+destination-branches: [master, iso7]
--- a/src/inmanta/module.py
+++ b/src/inmanta/module.py
@@ -2976,18 +2976,24 @@ class Module(ModuleLike[TModuleMetadata], ABC):
         """
         Generate a list of all Python files in the given plugin directory.
         This method prioritizes .pyc files over .py files, uses caching to avoid duplicate directory walks,
-        includes namespace packages and excludes the model directory.
+        includes namespace packages and excludes the `model`, `files` and `templates` directories.
         """
         # Return cached results if this directory has been processed before
         if plugin_dir in self._dir_cache:
             return self._dir_cache[plugin_dir]
 
         files: dict[str, str] = {}
-        model_dir_path: str = os.path.join(plugin_dir, "inmanta_plugins", self.name, "model")
+        model_dir_path: str = os.path.join(plugin_dir, "model")
+        files_dir_path: str = os.path.join(plugin_dir, "files")
+        templates_dir_path: str = os.path.join(plugin_dir, "templates")
 
         for dirpath, dirnames, filenames in os.walk(plugin_dir, topdown=True):
             # Modify dirnames in-place to stop os.walk from descending into any more subdirectories of the model directory
-            if dirpath.startswith(model_dir_path):
+            if (
+                dirpath.startswith(model_dir_path)
+                or dirpath.startswith(files_dir_path)
+                or dirpath.startswith(templates_dir_path)
+            ):
                 dirnames[:] = []
                 continue
 
--- a/src/inmanta/moduletool.py
+++ b/src/inmanta/moduletool.py
@@ -1761,7 +1761,7 @@ setup(name="{ModuleV2Source.get_package_name_for(self._module.name)}",
                 elif problematic_dir_name == "files":
                     bundling_description = "inmanta files for managed machines"
                 elif problematic_dir_name == "templates":
-                    bundling_description = "templates that will be used to generate configuration files"
+                    bundling_description = "inmanta templates that will be used to generate configuration files"
                 else:
                     raise RuntimeError(f"Unexpected bundling case: {problematic_dir_name}!")
 
--- a/src/inmanta/moduletool.py
+++ b/src/inmanta/moduletool.py
@@ -1751,13 +1751,32 @@ setup(name="{ModuleV2Source.get_package_name_for(self._module.name)}",
         """
         python_pkg_dir: str = os.path.join(build_path, "inmanta_plugins", self._module.name)
         model_dir: str = os.path.join(python_pkg_dir, "model")
-        if os.path.exists(model_dir):
-            raise ModuleBuildFailedError(
-                msg="There is already a model directory in %s. "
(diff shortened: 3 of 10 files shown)


src/inmanta/module.py
```diff
@@ -2976,18 +2976,24 @@ def _list_python_files(self, plugin_dir: str) -> list[str]:
...
-        includes namespace packages and excludes the model directory.
+        includes namespace packages and excludes the `model`, `files` and `templates` directories.
         """
         # Return cached results if this directory has been processed before
         if plugin_dir in self._dir_cache:
             return self._dir_cache[plugin_dir]
 
         files: dict[str, str] = {}
-        model_dir_path: str = os.path.join(plugin_dir, "inmanta_plugins", self.name, "model")
+        model_dir_path: str = os.path.join(plugin_dir, "model")
```
wouterdb, line 2986: where did the `"inmanta_plugins", self.name` part go?

tests/compiler/test_basics.py
```diff
@@ -731,3 +735,56 @@ def test_implementation_import_missing_error(snippetcompiler) -> None:
     assert "could not find type tests::length in namespace __config__" in exception.value.msg
     assert exception.value.location.lnr == 6
     assert exception.value.location.start_char == 20
+
+
+@pytest.mark.slowtest
+@pytest.mark.parametrize("problematic_folder", ["files", "model", "templates"])
```
wouterdb, line 741: Why not test all of them in one go to speed this up?

tests/compiler/test_basics.py
```diff
@@ -731,3 +735,56 @@ def test_implementation_import_missing_error(snippetcompiler) -> None:
...
+
+    # set up project with a v2 module
+    snippetcompiler_clean.setup_for_snippet(
+        """
+import std
+import custom_mod_one
+        """.strip(),
+        python_package_sources=[local_module_package_index],
+        project_requires=[
+            module.InmantaModuleRequirement.parse("std~=4.3.3,<4.3.4"),
```
wouterdb, line 776: why this constraint?

tests/compiler/test_basics.py
```diff
@@ -731,3 +735,56 @@ def test_implementation_import_missing_error(snippetcompiler) -> None:
...
+            module.InmantaModuleRequirement.parse("custom_mod_one>0"),
+        ],
+        python_requires=[
+            module.InmantaModuleRequirement.parse("custom_mod_one<999").get_python_package_requirement(),
```
wouterdb, line 780: Why is this?

tests/compiler/test_basics.py
```diff
@@ -731,3 +735,56 @@ def test_implementation_import_missing_error(snippetcompiler) -> None:
...
+    # set up project with a v2 module
+    snippetcompiler_clean.setup_for_snippet(
+        """
+import std
+import custom_mod_one
+        """.strip(),
+        python_package_sources=[local_module_package_index],
+        project_requires=[
+            module.InmantaModuleRequirement.parse("std~=4.3.3,<4.3.4"),
+            module.InmantaModuleRequirement.parse("custom_mod_one>0"),
```
wouterdb, line 777:
```suggestion
            module.InmantaModuleRequirement.parse("custom_mod_one"),
```

wouterdb (changes requested):

src/inmanta/module.py
hlloreda: We saw with @sanderr that the plugin_dir was already containing the `"inmanta_plugins", self.name` part:
Screenshot from 2024-08-28 13-24-08

sanderr: We believe it was bugged from the start and it doesn't seem to be covered by a test.

changelogs/unreleased/7849-code-loading.yml
```diff
@@ -0,0 +1,3 @@
+description: Make sure the compiler will not load Python code code present in the `files`, `model` or `templates` directory
+change-type: patch
+destination-branches: [master, iso7]
```
sanderr, line 3: This is technically a breaking change because not only do we fix the bug, we also prohibit the `files` and `templates` namespaces.

I think we have to make this iso8 only. @wouterdb wdyt?

sanderr (changes requested):
[This part](https://github.com/inmanta/inmanta-core/blob/2ad9b500457d77aaeac5dcc9fa05fa4f4de019f1/src/inmanta/moduletool.py#L1754) seems to be missing. I think it's relevant here as well.

changelogs/unreleased/7849-code-loading.yml
bartv: It would be really nice to have this bug fixed in iso7 as well. It has bitten us already a few times.

src/inmanta/moduletool.py
```diff
@@ -1751,13 +1751,32 @@ def _move_data_files_into_namespace_package_dir(self, build_path: str) -> None:
...
-                msg="There is already a model directory in %s. "
-                "The `inmanta_plugins.%s.model` package is reserved for bundling the inmanta model files. "
-                "Please use a different name for this Python package."
-                % (os.path.join(self._module.path, "inmanta_plugins", self._module.name), self._module.name)
-            )
+        files_dir: str = os.path.join(python_pkg_dir, "files")
+        templates_dir: str = os.path.join(python_pkg_dir, "templates")
+        for problematic_dir_path in [model_dir, files_dir, templates_dir]:
+            if os.path.exists(problematic_dir_path):
+                problematic_dir_name = os.path.basename(problematic_dir_path)
```
sanderr, line 1758: The general flow here feels a bit off to me. You start with a name for each of these, then build the paths for each seperately, loop over them, and use `os` to get the names you started with.

And as an added bonus, if you start out with `Collection[tuple[<dir>, <description>]]` you don't need the if-else with exhaustiveness check either.

sanderr (changes requested):

changelogs/unreleased/7849-code-loading.yml
bartv: Then we break that hypothetical module that we are sure of that does not exist.

wouterdb: We need this in iso7, it really is a bug.

(and the module build would crash if there is both a files folder and a files submodule )

wouterdb (approved):

sanderr (approved):

PR #8023, opened 2024-08-28
Interrupt timer on shutdown

wouterdb:
# Description

In order to prevent races, we interrupt the sleep timer.

I don't like this solution at all. 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/resourcepool.py
+++ b/src/inmanta/agent/resourcepool.py
@@ -50,7 +50,7 @@ import abc
 import asyncio
 import datetime
 import logging
-from asyncio import FIRST_COMPLETED, CancelledError, Event, Task
+from asyncio import CancelledError, Task
 from typing import Any, Callable, Coroutine, Generic, Optional, TypeVar
 
 import inmanta.util
--- a/src/inmanta/agent/resourcepool.py
+++ b/src/inmanta/agent/resourcepool.py
@@ -50,6 +50,7 @@ import abc
 import asyncio
 import datetime
 import logging
+from asyncio import FIRST_COMPLETED, CancelledError, Event, Task
 from typing import Any, Callable, Coroutine, Generic, Optional, TypeVar
 
 import inmanta.util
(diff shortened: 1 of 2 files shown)


src/inmanta/agent/resourcepool.py
```diff
@@ -337,11 +346,17 @@ async def cleanup_inactive_pool_members_task(self) -> None:
...
+            while self.running:
+                sleep_interval = await self.cleanup_inactive_pool_members()
+                if self.running:
+                    LOGGER.log(LOG_LEVEL_TRACE, "Manager will clean in %.2f seconds", sleep_interval)
+                    # Allow wait to be cancelled on shutdown
+                    self.shutdown_sleep = asyncio.create_task(asyncio.sleep(sleep_interval))
+                    await self.shutdown_sleep
+        except CancelledError:
+            # We are woken up by shutdown
+            pass
```
wouterdb, line 359: I don't like this at all.

It is too messy to be able to
1. wait for completion in the join method
2. prevent cancel of the cleanup itself. 

I wonder if I should step to something more like our scheduler? 

or am I missing something?

src/inmanta/agent/resourcepool.py
```diff
@@ -337,11 +346,17 @@ async def cleanup_inactive_pool_members_task(self) -> None:
...
         cleanup method in the test without being blocked in a loop.
         """
-        while self.running:
-            sleep_interval = await self.cleanup_inactive_pool_members()
-            if self.running:
-                LOGGER.log(LOG_LEVEL_TRACE, "Manager will clean in %.2f seconds", sleep_interval)
-                await asyncio.sleep(sleep_interval)
+        try:
+            while self.running:
+                sleep_interval = await self.cleanup_inactive_pool_members()
```
arnaudsjs, line 351: Don't we also want to wrap this method into a try-catch block? Otherwise we might drop out of the while loop when it raises an exception.

arnaudsjs (approved):

src/inmanta/agent/resourcepool.py
arnaudsjs: This is fine for me. I don't have the feeling that the alternatives will be much better.

wouterdb:
blocked on #8024

PR #8024, opened 2024-08-28
Fix broken tests with Std >= 6.0

hlloreda:
# Description

Fixed broken tests following the removal of some resources in std 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/tests/compiler/test_execution.py
+++ b/tests/compiler/test_execution.py
@@ -30,6 +30,28 @@ def test_issue_139_scheduler(snippetcompiler):
 import std
 import std::testing
 
+typedef service_state as string matching self == "running" or self == "stopped"
+
+entity Service:
+    string name
+    service_state state
+    bool onboot
+    bool reload=false
+    bool send_event=false
+end
+
+implement Service using serviceHost
+
+Service.host [1] -- Host.services [0:]
+
+index Service(host, name)
+
+implementation serviceHost for Service:
+    self.requires = self.host.requires
+end
+
+implement Service using serviceHost
+
 entity Host extends std::Host:
     string attr
 end
@@ -38,7 +60,8 @@ implement Host using std::none
 host = Host(name="vm1", os=std::linux)
 
 f = std::testing::NullResource(name=host.name)
-
+Service(host=host, name="svc", state="running", onboot=true, requires=[f])
+ref = Service[host=host, name="svc"]
 """
     )
     with pytest.raises(MultiException):
--- a/tests/compiler/test_execution.py
+++ b/tests/compiler/test_execution.py
@@ -39,9 +39,6 @@ host = Host(name="vm1", os=std::linux)
 
 f = std::testing::NullResource(name=host.name)
 
-std::Service(host=host, name="svc", state="running", onboot=true, requires=[f])
-ref = std::Service[host=host, name="svc"]
-
 """
     )
     with pytest.raises(MultiException):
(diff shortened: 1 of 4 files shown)


tests/compiler/test_execution.py
```diff
@@ -39,9 +39,6 @@ def test_issue_139_scheduler(snippetcompiler):
 
 f = std::testing::NullResource(name=host.name)
 
-std::Service(host=host, name="svc", state="running", onboot=true, requires=[f])
-ref = std::Service[host=host, name="svc"]
-
```
wouterdb, line 44: This is the entire purpose of the test, you have to replace it with some other index lookup

tests/compiler/test_index.py
```diff
@@ -141,9 +149,19 @@ def test_issue_212_bad_index_defintion(snippetcompiler):
...
+
+        index A(host, path)
+
+        implement A using std::none
+
         host = std::Host(name="a",os=std::linux)
-        a=std::DefaultDirectory(host=host,path="/etc")
-        b=std::DefaultDirectory(host=host,path="/etc")
+        a=A(host=host,path="/etc")
+        b=A(host=host,path="/etc")
```
wouterdb, line 164: This is removes the entire purpose of this test: to test if indexes on supertypes also apply to subtypes. There is not subtyping here.

wouterdb (changes requested):

tests/compiler/test_execution.py
hlloreda: Am I missing something? This is on the old version (with the std::Service and std==5.2.4):
```
        else:
>           raise MultiException(excns)
E           inmanta.ast.MultiException: Reported 2 errors:
E           	The object __config__::Host (instantiated at /tmp/tmpdb_a7l7j/main.cf:10) is not complete: attribute attr (/tmp/tmpdb_a7l7j/main.cf:6:12) is not set
E           	The object __config__::Host (instantiated at /tmp/tmpdb_a7l7j/main.cf:10) is not complete: attribute host_config (/home/hugo/Documents/Inmanta/copies/inmanta-core/venv/lib/python3.11/site-packages/inmanta_plugins/std/model/_init.cf:463:6) is not set
```
This is on the new one (std==6.0.0):
```
>           raise MultiException(excns)
E           inmanta.ast.MultiException: Reported 2 errors:
E           	The object __config__::Host (instantiated at /tmp/tmpcj4mkegw/main.cf:10) is not complete: attribute attr (/tmp/tmpcj4mkegw/main.cf:6:12) is not set
E           	The object __config__::Host (instantiated at /tmp/tmpcj4mkegw/main.cf:10) is not complete: attribute host_config (/home/hugo/Documents/Inmanta/copies/inmanta-core/venv/lib/python3.11/site-packages/inmanta_plugins/std/model/_init.cf:286:6) is not set
```

wouterdb: yes, which are both the good behavior. But the testcase no longer reproduces the original bug, so the fact that it doesn't exhibit the bad behavior doesn't indicate it is absent. 

(the annoying thing is that it is now very hard to reproduce the bug, because it has been fixed a long time ago, but we want to keep this regression test in here)

tests/compiler/test_index.py
```diff
@@ -141,9 +149,23 @@ def test_issue_212_bad_index_defintion(snippetcompiler):
...
+        entity A:
+            string path
+        end
+
+        A.host [1] -- std::Host.ases [0:]
+
+        entity B extends A:
+        end
+
+        index B(host, path)
```
wouterdb, line 161:
```suggestion
        index A(host, path)
```

wouterdb (approved):

PR #8072, opened 2024-09-09
Strange solution

wouterdb:
# Description

1. I cleaned up some bad shutdown code
2. I tried to sidestep our race condition and it seems to work (for now)

# The strange solution:

We tru to bypass bug on https://github.com/python/cpython/issues/103911#issuecomment-2333963137
by taking this branch  https://github.com/python/cpython/blob/d8f3c1e8f9c384fa6a473c3923f3b26ad6257cac/Lib/subprocess.py#L1850

May have security implications, keeping FD's open is definitely not clean 

# result

Running on a branch in a loop , ideally this produces a clean crash that leads to a  fixable cause 
After a few hunderd tests: it seems to actually fix it 

# Self Check:

- [x] Changelog entry


arnaudsjs (approved):

src/inmanta/agent/agent.py
```diff
@@ -539,13 +539,14 @@ def __init__(self, process: "Agent", name: str, uri: str, *, ensure_deploy_on_st
...
         self.executor_manager: executor.ExecutorManager[executor.Executor] = process.executor_manager
+        self._executors_to_join: Optional[list[executor.Executor]] = None
 
     async def stop(self) -> None:
         self._stopped = True
         self._enabled = False
         self._disable_time_triggers()
         self._nq.cancel()
-        await self.executor_manager.stop_for_agent(self.name)
+        self._executors_to_join = await self.executor_manager.stop_for_agent(self.name)
```
sanderr, line 549: Why make it optional? It then requires the assertion later, while a list can just as well have 0 elements.

src/inmanta/env.py
```diff
@@ -1054,8 +1054,11 @@ async def async_run_command_and_stream_output(
         of the sub-process.
         """
         full_output = []
+        # We use close_fds=False to use an alternative path to spawn to avoid
+        # the bug https://github.com/python/cpython/issues/103911#issuecomment-2333963137
+        # This uses spawn instead of fork+execve
         process = await asyncio.create_subprocess_exec(
-            cmd[0], *cmd[1:], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, env=env_vars
+            cmd[0], *cmd[1:], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, env=env_vars, close_fds=False
```
sanderr, line 1061: I can only review this part superficially because I don't understand all intricacies of what's going on.

sanderr (changes requested):

src/inmanta/env.py
wouterdb: same as everyone 'm afraid

sanderr (approved):

PR #8079, opened 2024-09-10
Add indexes to improve performance diagnose endpoint

arnaudsjs:
# Description

This PR adds indexes that are required by the diagnose endpoint.

Part of https://github.com/inmanta/inmanta-lsm/pull/1881

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (approved):

src/inmanta/db/versions/v202409090.py
```diff
@@ -0,0 +1,29 @@
...
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    """
+    Add indexes required by the diagnose endpoint.
+    """
+    schema = """
+    CREATE INDEX report_started_compile_returncode ON report(started DESC, compile, returncode);
```
sanderr, line 27: Are you sure that this index is actually used by the query? I'm surprised to see it here, since we first select on compile and returncode, and only then sort on `started`. Futhermore, I'd expect the set of reports to sort on to be either one or none.

sanderr (changes requested):

src/inmanta/db/versions/v202409090.py
arnaudsjs: You are right. I swaped it around. It should have been `(compile, returncode, started DESC)`. I'll change that. Thanks!

sanderr: But even then, is the index worth it? Do we ever expect the `compile, returncode` match to return more than one record?

sanderr: i.e. we stop after the first failed compile stage, don't we?

sanderr (approved):

PR #8090, opened 2024-09-13
Add failed state and test second deploy

### LINKED ISSUE #8010: resource scheduler: agent instrumentation
Implement the interaction between scheduler and agent. Make sure the scheduler can drive the agent on a per-resource basis. Make sure the old mode still works when the config option (part of #8009) is set.

- [x] start scheduler on the agent if the new mode is selected
- [x] implement `Scheduler._run_task` to request an executor and to delegate the resource action request to it, focus on deploy only

- [x] add test with dummy executor

- [x] clean up
  - collect discussion points in slides? 
- [x] merge PR 1
- [ ] look into state representation and propagation
- [ ] test DB state Vs scheduler state
- [ ] process failures appropriately 

#7561 already contains parts of this, though it was created before the resource scheduler design, it may or may not be easier to just start from scratch

wouterdb:
# Description

Basic end-to-end testing

close #8010 

# Self Check:

- [X] Attached issue to pull request
- [x] Changelog entry


src/inmanta/deploy/work.py
```diff
@@ -20,47 +20,31 @@
...
-from typing import Callable, Generic, Optional, TypeAlias, TypeVar
+from typing import Callable, Generic, Optional, TypeVar
 
 from inmanta.data.model import ResourceIdStr
 from inmanta.resources import Id
 
 
 @dataclass(frozen=True, kw_only=True)
-class _Task(abc.ABC):
+class Task(abc.ABC):
```
sanderr, line 32: Why still define the ABC here? I think it would make sense to move it to `tasks` as well, so that `work` depends on `tasks` but not vice versa.

src/inmanta/deploy/tasks.py
```diff
@@ -0,0 +1,165 @@
...
+
+    It is used to make sure all workers wake up to observe that they have been closed.
+    It functions mostly as a no-op
+    """
+
+    async def execute(self, scheduler: "scheduler.ResourceScheduler", agent: str) -> None:
+        pass
+
+
+class WithHashMatchTask(work.Task):
```
sanderr, line 45: Where does this name come from? i.e. how does `RefreshFact` care about hash matching? From the implementation I see only the fact that they act on the latest version.

sanderr (changes requested):
Tiny review so far, pending discussion on [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1726232081294179?thread_ts=1726221115.255039&cid=CKRF0C8R3).

src/inmanta/deploy/work.py
wouterdb: right, that could make things better 

I found it more natural to have the abstract in one place, but I see the point

src/inmanta/deploy/tasks.py
wouterdb: It may not be the best name, but given the plan to eliminate it, I would not invest too much time in it

sanderr: > the plan to eliminate it

Do you mean given my proposal on Slack? I'm not sure it would go in its entirety but we'll see.

src/inmanta/deploy/tasks.py
```diff
@@ -0,0 +1,182 @@
...
+            # refresh resource details for latest model state
+            new_details: Optional[state.ResourceDetails] = scheduler._state.resources.get(self.resource, None)
+            my_state: state.ResourceState | None = scheduler._state.resource_state.get(self.resource, None)
+            if new_details is not None and new_details.attribute_hash == resource_details.attribute_hash:
+                assert my_state is not None
+                if is_success:
+                    my_state.status = state.ResourceStatus.UP_TO_DATE
+                    my_state.deployment_result = state.DeploymentResult.DEPLOYED
+                else:
+                    # FIXME: WDB to Sander: do we set status here as well?
```
sanderr, line 103: According to the current definition of the status yes, but there's a FIXME attached to that as well, so perhaps we can mark this accordingly?
```suggestion
                    # FIXME[#8008]: WDB to Sander: do we set status here as well?
```

tests/agent_server/deploy/e2e/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+    yield server_config
+    agentmanager.no_auto_start_scheduler = False
+
+
+@pytest.fixture(scope="function")
+async def agent(server, environment):
+    """Construct an agent that can execute using the resource container"""
+    agentmanager = server.get_slice(SLICE_AGENT_MANAGER)
+
+    # First part - test the ResourceScheduler (retrieval of data from DB)
```
sanderr, line 58: Did this comment get lost? It doesn't seem to apply here.

sanderr (changes requested):

src/inmanta/deploy/tasks.py
wouterdb: I mean that

sanderr (approved):

Issue #7556, opened 2024-04-24
Refuse deletion of projects with environments


arnaudsjs:
This ticket should make sure that the API endpoints to delete a project, refuse the deletion is that project still has environments attached. The error message should mention which environments the project still has. This change should go into ISO7+.

PR #8097, opened 2024-09-17
Prevent deletion project

hlloreda:
# Description

Prevent the deletion of projects with environments

closes https://github.com/inmanta/inmanta-core/issues/7556

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/7556-prevent-project-deletion-if-environment.yml
+++ b/changelogs/unreleased/7556-prevent-project-deletion-if-environment.yml
@@ -5,4 +5,4 @@ issue-repo: inmanta-core
 change-type: minor
 destination-branches: [master, iso7]
 sections:
-  feature: "{{description}}"
+  bugfix: "{{description}}"
--- a/changelogs/unreleased/7556-prevent-project-deletion-if-environment.yml
+++ b/changelogs/unreleased/7556-prevent-project-deletion-if-environment.yml
@@ -1,0 +1,8 @@
+---
+description: Prevent the deletion of a project if this project contains one or multiple environments attached to it.
+issue-nr: 7556
+issue-repo: inmanta-core
+change-type: minor
+destination-branches: [master, iso7]
+sections:
+  feature: "{{description}}"
--- a/src/inmanta/server/services/projectservice.py
+++ b/src/inmanta/server/services/projectservice.py
@@ -16,7 +16,6 @@
     Contact: code@inmanta.com
 """
 
-import asyncio
 import logging
 import uuid
 from typing import Optional, cast
@@ -112,7 +111,8 @@ class ProjectService(protocol.ServerSlice):
         environments = await data.Environment.get_list(project=project.id)
         if len(environments) > 0:
             raise ServerError(
-                f"Cannot remove the project `{project_id}` because it still contains some environments: {','.join([str((env.name, str(env.id))) for env in environments])}"
+                f"Cannot remove the project `{project_id}` because it still contains some environments: "
+                f"{','.join([str((env.name, str(env.id))) for env in environments])}"
             )
 
         await project.delete()
--- a/src/inmanta/server/services/projectservice.py
+++ b/src/inmanta/server/services/projectservice.py
@@ -110,8 +110,10 @@ class ProjectService(protocol.ServerSlice):
             raise NotFound("The project with given id does not exist.")
 
         environments = await data.Environment.get_list(project=project.id)
-        for env in environments:
-            await asyncio.gather(self.autostarted_agent_manager.stop_agents(env, delete_venv=True), env.delete_cascade())
+        if len(environments) > 0:
+            raise ServerError(
+                f"Cannot remove the project `{project_id}` because it still contains some environments: {','.join([str((env.name, str(env.id))) for env in environments])}"
+            )
 
         await project.delete()
 
--- a/src/inmanta/server/services/projectservice.py
+++ b/src/inmanta/server/services/projectservice.py
@@ -25,7 +25,7 @@ import asyncpg
 from inmanta import data
 from inmanta.data import model
 from inmanta.protocol import handle, methods, methods_v2
-from inmanta.protocol.exceptions import NotFound, ServerError
+from inmanta.protocol.exceptions import Conflict, NotFound, ServerError
 from inmanta.server import (
     SLICE_AUTOSTARTED_AGENT_MANAGER,
     SLICE_DATABASE,
@@ -110,7 +110,7 @@ class ProjectService(protocol.ServerSlice):
 
         environments = await data.Environment.get_list(project=project.id)
         if len(environments) > 0:
-            raise ServerError(
+            raise Conflict(
                 f"Cannot remove the project `{project_id}` because it still contains some environments: "
                 f"{','.join([str((env.name, str(env.id))) for env in environments])}"
             )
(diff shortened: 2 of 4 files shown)


src/inmanta/server/services/projectservice.py
hlloreda, line 1: This will impact the behaviour of V1 and V2 endpoints, is it a problem?

tests/test_project.py
```diff
@@ -333,20 +406,6 @@ async def test_create_env_same_name(client):
...
-    result = await client.create_environment(project_id=project_id, name="dev")
-    result = await client.create_environment(project_id=project_id, name="prod")
-
-    result = await client.delete_project(project_id)
-    assert result.code == 200
-
-    result = await client.list_environments()
-    assert len(result.result["environments"]) == 0
-
-
```
hlloreda, line 349: No such behaviour anymore

changelogs/unreleased/7556-prevent-project-deletion-if-environment.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: Prevent the deletion of a project if this project contains one or multiple environments attached to it.
+issue-nr: 7556
+issue-repo: inmanta-core
+change-type: minor
+destination-branches: [master, iso7]
+sections:
+  feature: "{{description}}"
```
arnaudsjs, line 8: The `feature` section is intended for big new features. This isn't. So we should either move it to `minor-improvement` if we consider this new functionality or to `bugfix` if we consider the old behavior to be a bug. I am fine with both.

src/inmanta/server/services/projectservice.py
```diff
@@ -110,8 +109,11 @@ async def project_delete(self, project_id: uuid.UUID) -> None:
             raise NotFound("The project with given id does not exist.")
 
         environments = await data.Environment.get_list(project=project.id)
-        for env in environments:
-            await asyncio.gather(self.autostarted_agent_manager.stop_agents(env, delete_venv=True), env.delete_cascade())
+        if len(environments) > 0:
+            raise ServerError(
```
arnaudsjs, line 113: A `ServerError` is an HTTP 500, so an unexpected exception on the server. This should be ` Conflict` (409), because the current state of the resource doesn't allow the action.

tests/test_project.py
```diff
@@ -83,6 +83,40 @@ async def test_project_api_v1(client):
     assert response.code == 404
 
 
+async def test_delete_environment_project(client):
```
arnaudsjs, line 86: This method uses the variable `result` and `response` for the response from an API endpoint. Let's keep it consistent and use one of both names for that. Same in the other test case.

arnaudsjs (changes requested):

tests/agent_server/test_server_agent.py
```diff
@@ -1889,24 +1889,6 @@ async def test_stop_autostarted_agents_on_environment_removal(server, client, re
     ps_diff_inmanta_agent_processes(original=inmanta_agent_child_processes, current_process=current_process, diff=0)
 
 
-async def test_stop_autostarted_agents_on_project_removal(server, client, resource_container, no_agent_backoff):
```
arnaudsjs, line 1892: Why was this test case removed?

tests/test_project.py
```diff
@@ -333,20 +406,6 @@ async def test_create_env_same_name(client):
...
-    project_id = result.result["project"]["id"]
-
-    result = await client.create_environment(project_id=project_id, name="dev")
-    result = await client.create_environment(project_id=project_id, name="prod")
-
-    result = await client.delete_project(project_id)
-    assert result.code == 200
-
-    result = await client.list_environments()
-    assert len(result.result["environments"]) == 0
```
arnaudsjs, line 347: This test case also calls into the `list_environments()` endpoint after the deletion of the project. The new test cases don't do that. Make sure to be very careful when deleting tests, so that we don't lose test coverage.

arnaudsjs (changes requested):

tests/agent_server/test_server_agent.py
arnaudsjs: I see. We have another test case for the environment. You can ignore this comment.

tests/test_project.py
```diff
@@ -83,6 +83,40 @@ async def test_project_api_v1(client):
     assert response.code == 404
 
 
+async def test_delete_environment_project(client):
+    """Test modifying the project of an environment"""
```
Hugo-Inmanta, line 87:
```suggestion
    """Test deleting a project containing an environment."""
```

tests/test_project.py
```diff
@@ -260,6 +294,45 @@ async def test_modify_environment_project(client_v2):
     assert result.code == 200
 
 
+async def test_delete_environment_project_v2(client_v2):
+    """Test modifying the project of an environment"""
```
Hugo-Inmanta, line 298: Is this test exactly the same as the one above but with 2 environments instead of one ? Do we need both tests, maybe this one is enough ?

tests/test_project.py
```diff
@@ -260,6 +294,45 @@ async def test_modify_environment_project(client_v2):
...
+
+    # Create two projects and two environments
```
Hugo-Inmanta, line 300:
```suggestion
    # Create a project with two environments
```

Hugo-Inmanta (changes requested):

tests/test_project.py
hlloreda: I tend to agree, I wanted to cover the V1 endpoint (even if it relies on the same logic as V2)

tests/test_project.py
```diff
@@ -260,6 +260,45 @@ async def test_modify_environment_project(client_v2):
...
+
+    result = await client_v2.environment_create(project_id=project_id_a, name="env")
+    assert result.code == 200
+    env1_id = result.result["data"]["id"]
+
+    result = await client_v2.environment_create(project_id=project_id_a, name="env2")
+    assert result.code == 200
+    env2_id = result.result["data"]["id"]
+
+    # Delete project a
```
Hugo-Inmanta, line 279:
```suggestion
    # Delete project
```

tests/test_project.py
```diff
@@ -260,6 +260,45 @@ async def test_modify_environment_project(client_v2):
     assert result.code == 200
 
 
+async def test_delete_environment_project_v2(client_v2):
+    """Test deleting a project containing an environment."""
+
+    # Create a projects and two environments
```
Hugo-Inmanta, line 266:
```suggestion
    # Create a project and two environments
```

tests/test_project.py
```diff
@@ -260,6 +260,45 @@ async def test_modify_environment_project(client_v2):
     assert result.code == 200
 
 
+async def test_delete_environment_project_v2(client_v2):
+    """Test deleting a project containing an environment."""
```
Hugo-Inmanta, line 264:
```suggestion
    """Test deleting a project containing environments."""
```

Hugo-Inmanta (changes requested):

tests/test_project.py
Hugo-Inmanta: I see, I completely missed the fact that they were using different endpoints. Maybe we should err on the side of caution and test both versions after all, in case we update this in the future. @arnaudsjs  wdyt ?

arnaudsjs (approved):

Hugo-Inmanta (approved):

Issue #8052, opened 2024-09-05
Update cache  parameter `for_version` to `lingering`.


Hugo-Inmanta:
Follow up to https://github.com/inmanta/inmanta-core/issues/7696

Implement the suggestions in [this comment](https://github.com/inmanta/inmanta-core/pull/7850#discussion_r1744902102)

Improve the naming as suggested https://inmanta.slack.com/archives/CKRF0C8R3/p1725543194151569

Make sure to remove all mentions of `lingering` everywhere (tests, docstrings ...)

PR #8109, opened 2024-09-19
Issue/8052 agent cache follow up improve naming

Hugo-Inmanta:
# Description

- Remove all mentions of `lingering`
- Silently deprecate `for_version` parameter in favour of 2 new parameters `refresh_after_access` and `evict_after_creation`.
- Set sensible default values for these new parameters

closes https://github.com/inmanta/inmanta-core/issues/8052

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/8052-agent-cache-improve-naming.yml
+++ b/changelogs/unreleased/8052-agent-cache-improve-naming.yml
@@ -1,3 +1,7 @@
-description: "Clarify user facing interface."
+description: >
+  The retention policy of items in the agent cache can be set via
+  the `evict_after_creation` and `evict_after_creation` parameters.
 change-type: patch
 destination-branches: [master]
+sections:
+  minor-improvement: "{{description}}"
--- a/changelogs/unreleased/8052-agent-cache-improve-naming.yml
+++ b/changelogs/unreleased/8052-agent-cache-improve-naming.yml
@@ -1,0 +1,3 @@
+description: "Clarify user facing interface."
+change-type: patch
+destination-branches: [master]
--- a/changelogs/unreleased/8052-agent-cache-improve-naming.yml
+++ b/changelogs/unreleased/8052-agent-cache-improve-naming.yml
@@ -5,3 +5,10 @@ change-type: minor
 destination-branches: [master]
 sections:
   minor-improvement: "{{description}}"
+  update-note: >
+    Agent cache retention policy parameter `timeout` is now an alias for
+    the `evict_after_creation` parameter.
+  deprecation-note: >
+    Agent cache retention policy parameters `for_version` and `timeout` are deprecated.
+    The `evict_after_creation` and `evict_after_last_access` parameters should be
+    used instead.
--- a/changelogs/unreleased/8052-agent-cache-improve-naming.yml
+++ b/changelogs/unreleased/8052-agent-cache-improve-naming.yml
@@ -5,7 +5,7 @@ change-type: minor
 destination-branches: [master]
 sections:
   minor-improvement: "{{description}}"
-  update-note: >
+  upgrade-note: >
     Agent cache retention policy parameter `timeout` is now an alias for
     the `evict_after_creation` parameter.
   deprecation-note: >
--- a/changelogs/unreleased/8052-agent-cache-improve-naming.yml
+++ b/changelogs/unreleased/8052-agent-cache-improve-naming.yml
@@ -1,7 +1,7 @@
 description: >
   The retention policy of items in the agent cache can be set via
-  the `evict_after_creation` and `evict_after_creation` parameters.
-change-type: patch
+  the `evict_after_creation` and `evict_after_last_access` parameters.
+change-type: minor
 destination-branches: [master]
 sections:
   minor-improvement: "{{description}}"
--- a/docs/model_developers/handlers.rst
+++ b/docs/model_developers/handlers.rst
@@ -236,12 +236,13 @@ second time with the same arguments, it will not be executed again, but the cach
 returned instead. To exclude specific arguments from the cache key, use the ``ignore`` parameter.
 
 Cache entries will be dropped from the cache when they become stale. Use the following parameters to control when an entry is considered stale:
-  * Setting ``evict_after_creation=True`` will mark entries as stale after a timeout controlled by the ``timeout`` parameter (5000s by default).
-  * Setting ``refresh_after_access=True`` will reset the expiry time of entries to 60s anytime they're used.
+  * ``evict_after_creation``: mark entries as stale after this amount of time (in seconds) has elapsed since they entered the cache (5000s by default).
+  * ``evict_after_last_access``: mark entries as stale after this amount of time (in seconds) has elapsed since they were last accessed (60 by default).
+
 
 .. note::
 
-    If both ``evict_after_creation=True`` and ``refresh_after_access=True`` are set,
+    If both ``evict_after_creation=True`` and ``evict_after_last_access=True`` are set,
     the entry will become stale when the shortest of the two timers is up.
 
 
@@ -249,12 +250,12 @@ For example, to cache the connection to a specific device for 120 seconds:
 
 .. code-block:: python
 
-    @cache(timeout=120, ignore=["ctx"], evict_after_creation=True)
+    @cache(ignore=["ctx"], evict_after_creation=120)
     def get_client_connection(self, ctx, device_id):
        # ...
        return connection
 
-Setting ``refresh_after_access=True`` (or omitting the parameter) will reset the lifetime
+Setting ``evict_after_last_access=60`` (or omitting the parameter) will reset the lifetime
 of the cached connection to 60s everytime it is read from the cache.
 
 .. code-block:: python
@@ -271,7 +272,7 @@ from the cache. It gets the cached item as argument.
 
 .. code-block:: python
 
-    @cache(timeout=120, ignore=["ctx"], refresh_after_access=True,
+    @cache(ignore=["ctx"], evict_after_last_access=60,
        call_on_delete=lambda connection:connection.close())
     def get_client_connection(self, ctx, device_id, version):
        # ...
--- a/docs/model_developers/handlers.rst
+++ b/docs/model_developers/handlers.rst
@@ -235,8 +235,8 @@ the method will form the cache key, the return value will be cached. When the me
 second time with the same arguments, it will not be executed again, but the cached result is
 returned instead. To exclude specific arguments from the cache key, use the ``ignore`` parameter.
 
-Cache entries will be dropped from the cache when they become stale. Use the following parameters to control when an entry is considered stale:
-  * ``evict_after_creation``: mark entries as stale after this amount of time (in seconds) has elapsed since they entered the cache (5000s by default).
+Cache entries will be dropped from the cache when they become stale. Use the following parameters to set the retention policy:
+  * ``evict_after_creation``: mark entries as stale after this amount of time (in seconds) has elapsed since they entered the cache.
   * ``evict_after_last_access``: mark entries as stale after this amount of time (in seconds) has elapsed since they were last accessed (60 by default).
 
 
@@ -255,8 +255,8 @@ For example, to cache the connection to a specific device for 120 seconds:
        # ...
        return connection
 
-Setting ``evict_after_last_access=60`` (or omitting the parameter) will reset the lifetime
-of the cached connection to 60s everytime it is read from the cache.
+Setting ``evict_after_last_access=60`` (or omitting the parameter) will evict
+the connection from the cache 60s after it was last read from the cache.
 
 .. code-block:: python
 
--- a/docs/model_developers/handlers.rst
+++ b/docs/model_developers/handlers.rst
@@ -50,6 +50,7 @@ entity it is serializing.
 
     from inmanta.resources import resource, Resource
 
+
     @resource("std::File", agent="host.name", id_attribute="path")
     class File(Resource):
         fields = ("path", "owner", "hash", "group", "permissions", "purged", "reload")
@@ -63,6 +64,7 @@ entity it is serializing.
         @staticmethod
         def get_permissions(_, obj):
             return int(x.mode)
+
 
 
 Classes decorated with :func:`@resource<inmanta.resources.resource>` do not have to inherit directly from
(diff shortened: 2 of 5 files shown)


src/inmanta/agent/cache.py
```diff
@@ -40,25 +40,31 @@ def __init__(
...
-            When False, this cache item will be evicted from the cache <timeout> seconds after
+        :param evict_after_last_access: When True, this cache item will stay in the cache for 60s after its last use.
+        :param evict_after_creation: When True, this cache item will be evicted from the cache <timeout> seconds after
             entering the cache.
         """
         self.key = key
         self.value = value
         self.call_on_delete = call_on_delete
-        self.lingering = lingering
+        self.evict_after_last_access = evict_after_last_access
```
hlloreda, line 59: I think I would name it like this
```suggestion
        self.referesh_after_access = evict_after_last_access
```

tests/test_cache.py
```diff
@@ -600,35 +678,44 @@ def short_lived_non_linger(self):
...
     # The cache cleanup method is called upon entering the AgentCache context manager
     with agent_cache:
-        assert "x1" == test.non_linger_test_method_1()  # cache hit
-        assert "x2" == test.non_linger_test_method_2()  # cache hit
-        assert "recurring_read" == test.non_linger_test_method_3(dummy_arg="recurring_read")  # cache hit
-        assert "x3" == test.linger_test_method()  # +1 miss
-        assert "x4" == test.short_lived_non_linger()  # +1 miss
-        test.check_n_cache_misses(2)
+        assert "x1" == test.test_method_1()  # cache hit
+        assert "x1" == test.test_method_11()  # cache hitm
```
hlloreda, line 682:
```suggestion
        assert "x1" == test.test_method_11()  # cache hit
```

tests/test_cache.py
```diff
@@ -600,35 +678,44 @@ def short_lived_non_linger(self):
...
+        assert "x2" == test.test_method_2()  # cache hit
+        assert "x2" == test.test_method_22()  # cache hitm
```
hlloreda, line 684:
```suggestion
        assert "x2" == test.test_method_22()  # cache hit
```

tests/test_cache.py
```diff
@@ -600,35 +678,44 @@ def short_lived_non_linger(self):
...
+        assert "recurring_read" == test.test_method_3(dummy_arg="recurring_read")  # cache hit
+        assert "recurring_read" == test.test_method_33(dummy_arg="recurring_read")  # cache hitm
```
hlloreda, line 686:
```suggestion
        assert "recurring_read" == test.test_method_33(dummy_arg="recurring_read")  # cache hit
```

hlloreda (changes requested):
Minor comments

Hugo-Inmanta:
I just realized I forgot to update the docs

src/inmanta/agent/cache.py
Hugo-Inmanta: I like this suggestion :+1:

arnaudsjs (changes requested):
> I just realized I forgot to update the docs

I am moving this to "request changes", so that I get a notification when it's ready for review again.

Hugo-Inmanta:
Latest looks:
image

changelogs/unreleased/8052-agent-cache-improve-naming.yml
```diff
@@ -0,0 +1,3 @@
+description: "Clarify user facing interface."
+change-type: patch
+destination-branches: [master]
```
arnaudsjs, line 3: As far as I can see we don't have a changelog entry yet for the changes to the cache. So we need to add one.

docs/model_developers/handlers.rst
```diff
@@ -233,28 +233,28 @@ method annotated with this annotation will be cached, similar to the way
...
 the method will form the cache key, the return value will be cached. When the method is called a
 second time with the same arguments, it will not be executed again, but the cached result is
-returned instead. To exclude specific arguments from the cache key, use the `ignore` parameter.
+returned instead. To exclude specific arguments from the cache key, use the ``ignore`` parameter.
 
-Cache entries will be dropped from the cache when they become stale.
-The for_version and timeout parameters control this behaviour:
+Cache entries will be dropped from the cache when they become stale. Use the following parameters to control when an entry is considered stale:
+  * Setting ``evict_after_creation=True`` will mark entries as stale after a timeout controlled by the ``timeout`` parameter (5000s by default).
+  * Setting ``refresh_after_access=True`` will reset the expiry time of entries to 60s anytime they're used.
```
arnaudsjs, line 240: I don't see a reason to hardcode this to 60s. As a default 60s is ok, but we should give more freedom to the user to configure the duration of the timeout like is the case for the `evict_after_creation` timer.

Wouldn't it be a good idea to change the type of these two parameters to integers, so that we can have a different value for each timer?

hlloreda, line 240: Perhaps it was already discussed, and I'm not aware of the conclusion of the discussion, but don't we want to be a bit more flexible than the 60s: be able to set any timeout for this kind of cache as well?

src/inmanta/agent/cache.py
```diff
@@ -40,25 +40,31 @@ def __init__(
         timeout: float,
```
hlloreda, line 40: If it was decided to keep 60 sec, don't we want this parameter to be optional? To me, it feels weird to have a parameter that will not be considered in some cases.

docs/model_developers/handlers.rst
arnaudsjs, line 240: I am not really a big fan of changing this from `evict_after_last_access` to `refresh_after_access`. The term "evict" makes it really clear this is about cache invalidation and the term "after_last_access" makes it clear the we will evict after the last access. `refresh_after_access` is a lot more vague. Refresh can mean a lot of different things. I also don't really like that we had a naming schema for the variables `evict_<something>`. That's something we lost now.

arnaudsjs, line 240: The fact that we reset timers is an implementation detail. We should more describe the behavior instead.

```suggestion
  * Setting ``evict_after_creation=True`` will mark entries as stale after a timeout controlled by the ``timeout`` parameter (5000s by default).
  * Setting ``refresh_after_access=True`` will mark the cache entry stale 60s after is has been last accessed.
```

tests/test_cache.py
```diff
@@ -415,183 +423,253 @@ def test_method_2(self, dummy_arg):
...
+            self.increment_miss_counter()
+            return "x2"
+
+        @cache(timeout=1, refresh_after_access=True)
+        def test_method_44(self):
+            self.increment_miss_counter()
+            return "x4"
+
+        @cache(refresh_after_access=True)
+        def test_method_55(self, dummy_arg: str):
```
hlloreda, line 491: Imo, I think it would be easier to read if the different test methods were named according to what we are trying to test instead of number. However, it could become too long and verbose.

arnaudsjs (changes requested):
I only had a look at the interface of the cache up to now. I have some remarks about that, so I am going to leave the remainder of the review for when discussion on the interface has settled down. I am going to pull in @bartv as well, because he has a clear opinion on what the interface should look like.

docs/model_developers/handlers.rst
hlloreda: I agree with this. As originally thought, the renaming I proposed was specific to the `CacheItem` class because checking: 
```
if item.evict_after_last_access
```
to see if an item needs to be refreshed is not so clear to me. But maybe I am the only one bothered by this

Hugo-Inmanta: I agree, I'll revert this and only apply the renaming to the `CacheItem` class.

src/inmanta/agent/handler.py
```diff
@@ -145,15 +153,47 @@ def wrapper(self: HandlerAPI[TResource], *args: object, **kwds: object) -> objec
...
+                        _evict_after_last_access = 60
+                else:
+                    _evict_after_last_access = 0
+
+                    # Successive overrides with a 5000s default
+                    _evict_after_creation = 5000
+                    if evict_after_creation > 0:
+                        _evict_after_creation = evict_after_creation
+                    if timeout and timeout > 0:
+                        _evict_after_creation = timeout
```
hlloreda, line 181: I think it would be helpful to warn people that an override will happen if they use both versions.

src/inmanta/agent/cache.py
```diff
@@ -76,21 +81,36 @@ def delete(self) -> None:
...
         self.delete()
 
+    def refresh(self, now: float) -> None:
+        """
+        Refresh this cache item. Resets the expiry time for items marked
+        for eviction a certain time after their last access.
+
+        :parameter now: Baseline 'now' time.
+        """
+        if self.refresh_after_access < 0:
```
wouterdb, line 91:
```suggestion
        if self.refresh_after_access <= 0:
```

changelogs/unreleased/8052-agent-cache-improve-naming.yml
```diff
@@ -0,0 +1,7 @@
+description: >
+  The retention policy of items in the agent cache can be set via
+  the `evict_after_creation` and `evict_after_last_access` parameters.
+change-type: minor
+destination-branches: [master]
+sections:
```
arnaudsjs, line 6: We probably also need to add an `update-note` section regarding the change in behavior of the `timeout` and `for_version` parameters on the cache. And also a `deprecation-note` to say to people should migrate away from the `timeout` and `for_version` parameters.

arnaudsjs (changes requested):

src/inmanta/agent/handler.py
```diff
@@ -145,15 +153,53 @@ def wrapper(self: HandlerAPI[TResource], *args: object, **kwds: object) -> objec
...
```
arnaudsjs, line 181: Don't we want to take into account the new configuration parameters when they are provided and take the old ones into account they are provided without any new configuration parameters being set? This implementation uses the old parameters also when the new options are there.

src/inmanta/agent/handler.py
```diff
@@ -145,15 +153,53 @@ def wrapper(self: HandlerAPI[TResource], *args: object, **kwds: object) -> objec
...
+
+                    LOGGER.warning(
+                        "Both the `evict_after_creation` and the `timeout` parameter are set "
+                        f"for cached method {f.__name__}. `evict_after_creation` will be "
+                        "overridden."
+                    )
```
arnaudsjs, line 187: We should also mention the actual value that was used.

src/inmanta/agent/handler.py
```diff
@@ -145,15 +153,53 @@ def wrapper(self: HandlerAPI[TResource], *args: object, **kwds: object) -> objec
...
+                    # Successive overrides with a 5000s default
+                    _evict_after_creation = 5000
+                    if evict_after_creation > 0:
+                        _evict_after_creation = evict_after_creation
+                    if timeout and timeout > 0:
+                        _evict_after_creation = timeout
+
+                    LOGGER.warning(
+                        "Both the `evict_after_creation` and the `timeout` parameter are set "
+                        f"for cached method {f.__name__}. `evict_after_creation` will be "
```
arnaudsjs, line 185: Log messages should not use f-strings.

arnaudsjs (changes requested):

src/inmanta/agent/handler.py
```diff
@@ -123,11 +127,15 @@ def cache(
     If an argument named resource is present,
     it is assumed to be a resource and its ID is used, without the version information
 
-    :param timeout: Hard timeout for non-lingering cache item i.e. when `for_version=False`.
+    :param timeout: Hard timeout for cache items when `evict_after_creation=True`.
```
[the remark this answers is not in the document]
Hugo-Inmanta: Nice catch. Removed the docstring for this parameter since it is deprecated.

src/inmanta/agent/handler.py
Hugo-Inmanta: Oof, nice catch

src/inmanta/agent/handler.py
```diff
@@ -145,15 +149,57 @@ def wrapper(self: HandlerAPI[TResource], *args: object, **kwds: object) -> objec
...
+                    elif timeout and timeout > 0:
+                        _evict_after_creation = timeout
+                    else:
+                        _evict_after_creation = 5000
+
+                    if evict_after_creation > 0 and timeout and timeout > 0:
+                        LOGGER.warning(
+                            msg="Both the `evict_after_creation` and the deprecated `timeout` parameter are set "
+                            "for cached method %s. Cached entries will be kept in the cache for %.2fs "
+                            "after entering it." % (f.__name__, evict_after_creation)
```
arnaudsjs, line 179: I know this is logically equivalent to what you wrote, but it's more easy to verify it's correct like this.
```suggestion
                            "after entering it." % (f.__name__, _evict_after_creation)
```

src/inmanta/agent/handler.py
```diff
@@ -145,15 +149,57 @@ def wrapper(self: HandlerAPI[TResource], *args: object, **kwds: object) -> objec
...
+                        )
```
arnaudsjs, line 180:
```suggestion
                        LOGGER.warning(
                            msg="Both the `evict_after_creation` and the deprecated `timeout` parameter are set "
                            "for cached method %s. Cached entries will be kept in the cache for %.2fs "
                            "after entering it.", f.__name__, evict_after_creation
                        )
```

src/inmanta/agent/cache.py
```diff
@@ -37,28 +37,35 @@ class CacheItem:
...
     ) -> None:
         """
         :param key: The full key identifying this item in the cache.
-        :param timeout: Hard timeout for non-lingering cache item.
         :param value: The value being cached associated to the key.
         :param call_on_delete: Optional finalizer to call when the cache item is deleted. This is
             a callable expecting the cached value as an argument.
-        :param lingering: When True, this cache item will linger in the cache for 60s after its last use.
-            When False, this cache item will be evicted from the cache <timeout> seconds after
+        :param evict_after_creation: This cache item will be considered stale this number of seconds after
```
hlloreda, line 50: Minor comment: It's a bit weird that the order of parameters is not following the order in the method definition. This comment applies at different places

src/inmanta/agent/cache.py
```diff
@@ -37,28 +37,35 @@ class CacheItem:
...
 
         now = time.time()
-        self.expiry_time = (now + 60) if lingering else (now + timeout)
+        self.expiry_time: float = sys.float_info.max
+        self.after_creation_expiry_time: float = sys.float_info.max
+
+        if evict_after_last_access > 0:
+            self.expiry_time = now + evict_after_last_access
+        if evict_after_creation > 0:
+            self.after_creation_expiry_time = now + evict_after_creation
```
hlloreda, line 67: Do we need this field?

src/inmanta/agent/cache.py
```diff
@@ -76,21 +83,37 @@ def delete(self) -> None:
...
+        Refresh this cache item. Resets the expiry time for items marked
+        for eviction a certain time after their last access.
+
+        :parameter now: Baseline 'now' time to make sure expiry times are synced up
+            across cache items.
+        """
+        if self.refresh_after_access <= 0:
+            # Refreshing on access is disabled on the CacheItem.
+            return
+        self.expiry_time = min(now + self.refresh_after_access, self.after_creation_expiry_time)
```
hlloreda, line 97: I see that there is a usage now, but it feels weird that it could impact the refresh time. I would definitively log a warning here

hlloreda: While I agree that it's explicitly written:
```
If both are set, the shortest duration of the
    two will trigger the item to become stale
```
I did not understand that this would also apply for the refresh variant as well. Am I the only one?

Hugo-Inmanta: As I understood it, if both are set, whichever timer expires first makes the item stale.

hlloreda: Never mind, after some discussions, I understood why it was like this

arnaudsjs (approved):

hlloreda (approved):

PR #8116, opened 2024-09-20
Issue/8008 resource scheduler polishing

### LINKED ISSUE #8008: resource scheduler: polish scheduling algorithm
#7548 will be merged ASAP, as soon as the interface towards server and agent is finalized. It will still contain a lot of TODO's that are specific to the scheduling algorithm. These should be resolved.

- [ ] polish
- [ ] unit tests (+ if feasible test with a dummy executor)
- [ ] TODO's

sanderr:
# Description

Polished the resource scheduler and addressed todos and review comments from #8001. I attached some comments to the diff with more details / motivation.

Please also have another look at the comments you made in #8001 to see if they have been resolved to your satisfaction.

closes #8008

# Self Check:

- [x] Changelog entry


src/inmanta/deploy/scheduler.py
```diff
@@ -36,11 +39,63 @@
 LOGGER = logging.getLogger(__name__)
 
 
-# FIXME[#8008] review code structure + functionality + add docstrings
-# FIXME[#8008] add import entry point test case
+class TaskManager(abc.ABC):
```
sanderr, line 42: I added this to have a restricted interface with the tasks, so they don't need to access private attributes. As I see it, this makes it easier to reason on scheduler state, because it's all in one place (here), and we know it won't be misused elsewhere. The two modules / classes are still relatively tightly coupled, but at least in a well defined manner.

src/inmanta/deploy/scheduler.py
```diff
@@ -60,30 +115,38 @@ def __init__(
         self._work: work.ScheduledWork = work.ScheduledWork(
             requires=self._state.requires.requires_view(),
             provides=self._state.requires.provides_view(),
-            consumer_factory=self.start_for_agent,
+            new_agent_notify=self._start_for_agent,
```
sanderr, line 118: Just a suggestion: I found the name `consumer_factory` a bit vague / confusing. I can revert this if you don't like it.

src/inmanta/deploy/state.py
```diff
@@ -16,40 +16,31 @@
...
-
-    # FIXME: flatten out inheritance?
-
-    def __init__(self, id: ResourceIdStr, version: int, attributes: dict[str, object], attribute_hash: AttributeHash) -> None:
-        super().__init__(id, version, attributes)
-        self.attribute_hash = attribute_hash
-
-    attribute_hash: AttributeHash
+@dataclass(frozen=True)
+class ResourceDetails:
```
sanderr, line 41: This is one of the bigger changes I made, as proposed on Slack. I felt that the inheritance from `executor.ResourceDetails` was not appropriate here, because `executor.ResourceDetails` contains the full specification to deploy a resource, while this dataclass here is simply meant to express the current intent of a resource.

Most importantly imo, it should be version agnostic: the scheduler knows which version it reflects, and it makes sure that its model state reflects that version. If a new version is read from the database, unchanged resources should not be affected. The resource should become versioned only at the point where we actually commit to a version. Therefore I moved the construction of `executor.ResourceDetails` to the `tasks` module.

src/inmanta/deploy/tasks.py
```diff
@@ -87,81 +96,57 @@ class PoisonPill(Task):
...
-
-    @abc.abstractmethod
-    async def execute_on_resource(
-        self, scheduler: "scheduler.ResourceScheduler", agent: str, resource_details: "state.ResourceDetails"
-    ) -> None:
-        pass
+        intent = await task_manager.get_resource_intent(self.resource, for_deploy=True)
+        if intent is None:
+            # Stale resource, can simply be dropped.
+            return
```
sanderr, line 110: The common part here became so trivial that I felt it made no sense anymore to group them together under a common subclass, so I dropped `OnLatestState`.

src/inmanta/deploy/work.py
```diff
@@ -250,75 +267,82 @@ class ScheduledWork:
...
+        and scheduler intent. May defer tasks that were previously considered ready to execute if any of their dependencies are
+        added to the scheduled work or if new dependencies are added.
 
-        :param ensure_scheduled: Set of resources that should be deployed. Adds a deploy task to the scheduled work for each
+        :param resources: Set of resources that should be deployed. Adds a deploy task to the scheduled work for each
             of these, unless it is already scheduled.
-        :param running_deploys: Set of resources for which a deploy is currently running, for the latest desired state, i.e.
-            stale deploys should be excluded.
+        :param stale_deploys: Set of resources for which a stale deploy is in progress, i.e. a deploy for an outdated resource
+            intent.
```
sanderr, line 312: @wouterdb this was previously `running_deploys`, on which you commented that it was unclear. I turned the approach around, I hope it's more clear now.

tests/agent_server/deploy/test_scheduler_agent.py
```diff
@@ -180,9 +180,7 @@ async def test_basic_deploy(agent: TestAgent, make_resource_minimal):
         ResourceIdStr(rid2): make_resource_minimal(rid2, {"value": "a"}, [rid1], 5),
     }
 
-    # FIXME: SANDER: It seems we immediatly deploy if a new version arrives, we don't wait for an explicit deploy call?
-    # Is this by design?
```
sanderr, line 184: It was, with the intention to have the same behavior as the one we have now. But I just realized that it is actually a deviation since this is currently managed by a setting.

I have to give it more thought, but it may be difficult to change this, damn.

src/inmanta/deploy/scheduler.py
```diff
@@ -143,86 +207,87 @@ async def get_facts(self, resource: dict[str, object]) -> None:
...
 
-    async def build_resource_mappings_from_db(self, version: int | None = None) -> Mapping[ResourceIdStr, ResourceDetails]:
+    async def _build_resource_mappings_from_db(self, version: int | None = None) -> Mapping[ResourceIdStr, ResourceDetails]:
         """
         Build a view on current resources. Might be filtered for a specific environment, used when a new version is released
 
         :return: resource_mapping {id -> resource details}
         """
         if version is None:
+            # TODO: create ticket: BUG: not necessarily released
```
sanderr, line 217: I'll delegate these to a new ticket but that will be for Monday.

src/inmanta/deploy/scheduler.py
```diff
@@ -60,30 +115,38 @@ def __init__(
...
+        # Set of resources for which a concrete non-stale deploy is in progress, i.e. we've committed for a given intent and
+        # that intent still reflects the latest resource intent
+        # Apart from the obvious, this differs from the agent queues' in-progress deploys in the sense that those are simply
+        # tasks that have been picked up while this set contains only those tasks for which we've already committed. For each
+        # deploy task, there is a (typically) short window of time where it's considered in progress by the agent queue, but
+        # it has not yet started on the actual deploy, i.e. it will still see updates to the resource intent.
+        self._deploying: set[ResourceIdStr] = set()
+        # Set of resources for which a concrete stale deploy is in progress, i.e. we've committed for a given intent and
+        # that intent has gone stale since
+        self._deploying_stale: set[ResourceIdStr] = set()
```
wouterdb, line 144: Why do we need to keep track of this?

src/inmanta/deploy/tasks.py
```diff
@@ -87,81 +96,55 @@ class PoisonPill(Task):
...
-            # The deploy that finished has become stale (state has changed since the deploy started).
-            # Nothing to report on a stale deploy.
-            # A new deploy for the current model state will have been queued already.
+        is_success: bool = status == const.ResourceState.deployed
+        await task_manager.report_resource_state(
+            resource=self.resource,
+            attribute_hash=resource_details.attribute_hash,
+            status=state.ResourceStatus.UP_TO_DATE if is_success else state.ResourceStatus.HAS_UPDATE,
+            deployment_result=state.DeploymentResult.DEPLOYED if is_success else state.DeploymentResult.FAILED,
+        )
```
wouterdb, line 121: I wonder if this should be in a finaly block? i.e. where is the hard exception boundary for e.g.  all the remoting? 

If this is not called, due to an exception, the state of the scheduler is toast, if I understand correctly?

tests/agent_server/deploy/scheduler_test_util.py
```diff
@@ -68,7 +68,7 @@ def convert_resource(resource: JsonType) -> ResourceDetails:
     m.update(character.encode("utf-8"))
     attribute_hash = m.hexdigest()
 
-    return ResourceDetails(out, attribute_hash)
+    return ResourceDetails(attributes=out, attribute_hash=attribute_hash)
```
wouterdb, line 71: oh, now I understand. The resource details `attribute` field doesn't just contain the attributes, but the serialized resource? Then I think the field should have a different name? 

or this is a mistaken update of the testcase, and it should be
```suggestion
    return ResourceDetails(attributes=attributes, attribute_hash=attribute_hash)
```

tests/agent_server/deploy/test_scheduler_agent.py
wouterdb: We will have to dampen this somehow, but not now.

wouterdb (commented):
nice!

src/inmanta/deploy/state.py
sanderr: I have to confess I'm always on the fence for these types of things. Do you duplicate the id in the object, adding an invariant that it matches the key in the mapping, or do you keep it contained to just the "data"? I'm not particularly attached to this, so I can add the id.

src/inmanta/deploy/scheduler.py
sanderr: 1. We want to track the in-progress tasks, so that when we trigger a deploy, we know what's already running.
2. We want to make sure we only consider tasks as in-progress for the purposes of 1, if they are running for the latest intent, or they haven't committed to an intent yet.

This set is to inform the `ScheduledWork` that while there are tasks running for these resources, they should not be taken into consideration because they are known to be stale.

I have to confess (I meant to add a comment but I forgot) that I'm not completely satisfied with how this ended up. But I also don't see an alternative that I prefer. The knowledge of which tasks are running is in the agent queues mostly, but the semantics of what the tasks mean and what a stale deploy is, live in the scheduler. I tried to stick with those responsibilities as much as I could, but there are some rough edges to it that I couldn't get rid of.

src/inmanta/deploy/tasks.py
sanderr: You're right. I made the change, but I had to move some code around to keep it coherent (it felt a bit strange to handle some exceptions in `do_deploy` and some others outside of it.

tests/agent_server/deploy/scheduler_test_util.py
sanderr: Definitely a mistaken update, thanks.

mypy-baseline.txt
```diff
@@ -978,6 +978,7 @@ src/inmanta/agent/resourcepool.py:0: error: Argument 1 to "append" of "list" has
 src/inmanta/agent/resourcepool.py:0: error: Missing type parameters for generic type "PoolMember"  [type-arg]
 src/inmanta/deploy/scheduler.py:0: error: Argument 1 to "parse_id" of "Id" has incompatible type "object"; expected "ResourceVersionIdStr | ResourceIdStr"  [arg-type]
 src/inmanta/deploy/scheduler.py:0: error: Argument "attribute_hash" to "ResourceDetails" has incompatible type "str | None"; expected "str"  [arg-type]
+src/inmanta/deploy/scheduler.py:0: error: "object" has no attribute "__iter__"; maybe "__dir__" or "__str__"? (not iterable)  [attr-defined]
```
sanderr, line 981: Simply a new instance of a known shortcoming in our typing (untyped resource attributes, in this case "requires").

src/inmanta/deploy/state.py
```diff
@@ -39,9 +39,16 @@ def provides_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
...
 class ResourceDetails:
+    resource_id: ResourceIdStr
     attribute_hash: str
     attributes: Mapping[str, object] = dataclasses.field(hash=False)
 
+    id: Id = dataclasses.field(init=False, compare=False, hash=False)
+
+    def __post_init__(self) -> None:
+        # use object.__setattr__ because this is a frozen dataclass, see dataclasses docs
+        object.__setattr__(self, "id", Id.parse_id(self.resource_id))
```
wouterdb, line 50: so this is how it's done! Oh the horror.

src/inmanta/deploy/tasks.py
wouterdb: It is normal, different types of failure are handled differently and at some level, no exceptions are expected, but they still have to be stopped to protect the main loop.

wouterdb (approved):

src/inmanta/deploy/tasks.py
sanderr: Exactly, but by lifting it up a level, the two cases become one and we're no longer dealing with vagueries purely for the sake of protecting the main loop, we now have the context to know what we're protecting against.

arnaudsjs (approved):

PR #8122, opened 2024-09-23
fix bug when loading versions from the DB

### LINKED ISSUE #8118: resource scheduler may deploy unreleased resources
`_build_resource_mappings_from_db` calls `get_resources_in_latest_version`. As the name implies, this returns the resources for the latest version, which might not have been released yet.

Start with a test case to confirm the hypothesis, then fix the bug.

### LINKED ISSUE #8119: resource scheduler may track inconsistent version
`read_version` sets version to `environment.last_version`. However, it also calls `_build_resource_mappings_from_db` to fetch the resources for a version. There are two issues here:

1. `environment.last_version` gives the highest reserved version number. This version might not have been released, or it might not even have been exported (yet or at all, e.g. export interrupted).
2. There might be a race where this version differs from the resources' version

wouterdb:
# Description

closes #8118 
closes #8119 

## The proverbial can of worms

It is very likely the method 'get_resources_in_latest_version' was always intended to return only released version. 

It's only use I can find is in lsm to find all resources that are required by a service.

```python
lifecycle_transfer_resources = await self.resource_service.get_resources_in_latest_version(
                environment=env,
                resource_type="lsm::LifecycleTransfer",
                attributes={"instance_id": str(service_id)},
                connection=con,
            )
```

In this context, 
1. it makes little sense to take into account unreleased version (it does make a little sense, because if all versions are produced by lsm, the last version, released or not, best reflects the latest inventory state)
2. it is absolutely whack to pull **every** `lsm::LifecycleTransfer` out of the DB to do what is essentially a match on the primary key.

I propose to make a ticket to 
1. remove the usage of this method from lsm and just do a query. (it is probably a remnant from the time where we expect a need to run lsm on a separate DB)
2. drop the additional argument and make the behavior needed here the default 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -207,18 +207,13 @@ class ResourceScheduler(TaskManager):
             )
         )
 
-    async def _build_resource_mappings_from_db(self, version: int | None = None) -> Mapping[ResourceIdStr, ResourceDetails]:
+    async def _build_resource_mappings_from_db(self, version: int) -> Mapping[ResourceIdStr, ResourceDetails]:
         """
         Build a view on current resources. Might be filtered for a specific environment, used when a new version is released
 
         :return: resource_mapping {id -> resource details}
         """
-        if version is None:
-            resources_from_db: list[Resource] = await data.Resource.get_resources_in_latest_version(
-                environment=self.environment, released_only=True
-            )
-        else:
-            resources_from_db = await data.Resource.get_resources_for_version(self.environment, version)
+        resources_from_db = await data.Resource.get_resources_for_version(self.environment, version)
 
         resource_mapping = {
             resource.resource_id: ResourceDetails(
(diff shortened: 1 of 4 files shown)


sanderr (approved):
Your proposal sounds good to me.

hlloreda (approved):

sanderr (changes requested):
Doesn't this still suffer from the same race condition, or at least part of it? Sure, you'll always get a release version now, but you may get a different one for each of the two calls.

I feel like this needs to be addressed. That said, if this partial fix gets the lsm testing unstuck I have no objection against delegating it to a new PR.

src/inmanta/deploy/scheduler.py
```diff
@@ -250,14 +244,11 @@ async def read_version(
...
-        # FIXME[8119]: version does not necessarily correspond to resources' version
-        #       + last_version is reserved, not necessarily released
-        #       -> call ConfigurationModel.get_version_nr_latest_version() instead?
-        version = environment.last_version
-        resources_from_db = await self._build_resource_mappings_from_db()
+        cm_version = await ConfigurationModel.get_latest_version(self.environment)
+        if cm_version is None:
+            return
+        version = cm_version.version
+        resources_from_db = await self._build_resource_mappings_from_db(version=version)
```
wouterdb, line 251: @sanderr this was the most elegant way I could come up with.

It guarantees consistency and it works for versions with no resources

sanderr: I like it

sanderr (approved):

PR #8126, opened 2024-09-24
Introduce repair and redeploy timers to the new agent

jptrindade:
# Description

Copied over some logic from the old agent to introduce repair and redeploy timers

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent_new.py
+++ b/src/inmanta/agent/agent_new.py
@@ -74,7 +74,10 @@ class Agent(SessionEndpoint):
 
         self._set_deploy_and_repair_intervals()
 
-    def _set_deploy_and_repair_intervals(self):
+    def _set_deploy_and_repair_intervals(self) -> None:
+        """
+        Fetch the settings related to automatic deploys and repairs from the config
+        """
         # do regular deploys
         self._deploy_interval = cfg.agent_deploy_interval.get()
         deploy_splay_time = cfg.agent_deploy_splay_time.get()
--- a/src/inmanta/agent/agent_new.py
+++ b/src/inmanta/agent/agent_new.py
@@ -16,11 +16,13 @@
     Contact: code@inmanta.com
 """
 
+import datetime
 import logging
 import os
+import random
 import uuid
 from concurrent.futures.thread import ThreadPoolExecutor
-from typing import Any, Optional
+from typing import Any, Callable, Coroutine, Optional, Union
 
 from inmanta import config, const, protocol
 from inmanta.agent import config as cfg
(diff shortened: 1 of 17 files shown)


tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -84,3 +88,159 @@ async def test_auto_deploy_no_splay(server, client, clienthelper, resource_conta
...
+    await verify_deployment_result()
+
+    # Manual change
+    resource_container.Provider.set("agent1", "key1", "another_value")
+    # Wait until repair restores the state
+    now = time.time()
+    while resource_container.Provider.get("agent1", "key1") != "value1":
+        if time.time() > now + 10:
+            raise Exception("Timeout occurred while waiting for repair run")
+        await asyncio.sleep(0.1)
```
Hugo-Inmanta, line 243: The [`retry_limited`](https://github.com/inmanta/inmanta-core/blob/c4913c81e47e6dd01aaaa64a4372085d0dcf962b/tests/utils.py#L68) method can be used for this pattern.

Hugo-Inmanta (changes requested):
Minor change on the test, lgtm otherwise :+1:

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -235,12 +235,12 @@ async def verify_deployment_result():
...
     # Manual change
     resource_container.Provider.set("agent1", "key1", "another_value")
+
     # Wait until repair restores the state
-    now = time.time()
-    while resource_container.Provider.get("agent1", "key1") != "value1":
-        if time.time() > now + 10:
-            raise Exception("Timeout occurred while waiting for repair run")
-        await asyncio.sleep(0.1)
+    def repaired() -> None:
```
Hugo-Inmanta, line 240:
```suggestion
    def repaired() -> bool:
```

Hugo-Inmanta (changes requested):

tests/agent_server/deploy/e2e/test_autostarted.py
jptrindade: yup, my bad, rushed it a little bit

sanderr (changes requested):
I'll start with a general concern: we have to make sure that these interval options can be changed and that we respond to it when they are. I have yet to consider how best to address this though.

src/inmanta/agent/agent_new.py
```diff
@@ -67,6 +69,78 @@ def __init__(
...
+        self._set_deploy_and_repair_intervals()
+
+    def _set_deploy_and_repair_intervals(self) -> None:
+        """
+        Fetch the settings related to automatic deploys and repairs from the config
+        """
+        # do regular deploys
+        self._deploy_interval = cfg.agent_deploy_interval.get()
+        deploy_splay_time = cfg.agent_deploy_splay_time.get()
+        self._deploy_splay_value = random.randint(0, deploy_splay_time)
```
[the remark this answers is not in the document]
wouterdb: I think we plan to replace these timer mechanisms entirely, so yes

This implementation triggers all agents at once, so also yes

sanderr (changes requested):
> I think we plan to replace these timer mechanisms entirely, so yes

@jptrindade considering Wouter's comment, I think we can leave the issue with the updating of the setting as is for now. But can you add a FIXME, noting that these settings are not currently updated (unlike the old agent) and that we should either fix it or rip this whole legacy timer out again?

sanderr (approved):

PR #8125, opened 2024-09-24
Issue/8012 scheduler event propagation

sanderr:
# Description

Quickfix implementation of event propagation for the resource scheduler. A follow-up PR will address the remaining todo's.

# Self Check:

Strike through any lines that are not applicable (`~~line~~`) then check the box

- [ ] Attached issue to pull request
- [ ] Changelog entry
- [ ] Type annotations are present
- [ ] Code is clear and sufficiently documented
- [ ] No (preventable) type errors (check using make mypy or make mypy-diff)
- [ ] Sufficient test cases (reproduces the bug/tests the requested feature)
- [ ] Correct, in line with design
- [ ] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )
- [ ] If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)

PR #8131, opened 2024-09-27
Issue/8012 scheduler event propagation refinement

### LINKED ISSUE #8012: resource scheduler: implement event propagation
Since the scheduler does not receive / calculate increments, it needs to propagate events on the fly. When a task finishes, if `resource.send_event=True`, its dependencies can be added to the scheduled work. Implementation may need to acquire scheduler lock.

sanderr:
# Description

Polishing of the changes in #8125

- Fixed a rare race condition where an event would be skipped if the listener happens to be deploying at the time the event is produced.
- Added test cases, both for #8125 and for the general scheduler behavior.
- Implemented core side of inmanta/std#549 (`receive_events` boolean on resources)

closes #8012

# Self Check:

- [x] Changelog entry


tests/conftest.py
```diff
@@ -760,7 +760,7 @@ def map_data_line(line):
 
 
 @pytest.fixture(scope="function")
-async def server_config(event_loop, inmanta_config, postgres_db, database_name, clean_reset, unused_tcp_port_factory):
+async def server_config(inmanta_config, postgres_db, database_name, clean_reset, unused_tcp_port_factory):
```
sanderr, line 763: I got warnings that this fixture should not be used with async test cases so I dropped it where I found it.

wouterdb (approved):
I didn't go over all testcases, but it sure looks thorough

jptrindade (approved):

PR #8132, opened 2024-09-27
added ManagedExecutor test helper

sanderr:
# Description

Adds a test helper class for scheduler testing. `ManagedExecutor` expects to be driven by the test case itself. It hangs at the start of deploy on a future for which the result should be set by the test case. You can find usage examples in #8131.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/tests/agent_server/deploy/test_scheduler_agent.py
+++ b/tests/agent_server/deploy/test_scheduler_agent.py
@@ -18,6 +18,7 @@
     This file is intended to contain test that use the agent/scheduler combination in isolation: no server, no executor
 """
 
+import asyncio
 import hashlib
 import json
 import typing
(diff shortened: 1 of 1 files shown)


tests/agent_server/deploy/test_scheduler_agent.py
```diff
@@ -71,14 +72,53 @@ async def open_version(self, version: int) -> None:
...
+        return self._deploys
+
+    async def stop(self) -> None:
+        # resolve hanging futures to prevent test hanging during teardown
+        for deploy in self._deploys.values():
+            deploy.set_result(const.ResourceState.undefined)
+
+    async def execute(self, gid: uuid.UUID, resource_details: ResourceDetails, reason: str) -> const.ResourceState:
+        assert resource_details.rid not in self._deploys
+        self._deploys[resource_details.rid] = asyncio.Future()
```
wouterdb, line 105: didn't know you could do this, I expect to have to use the create_future of the ioloop.

wouterdb (approved):
pretty

tests/agent_server/deploy/test_scheduler_agent.py
sanderr: You can do this, but I checked the docs just now and it turns out that they recommend your approach: "the recommended way to create a Future object is to call [loop.create_future()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_future)".

sanderr:
```suggestion
        self._deploys[resource_details.rid] = asyncio.get_running_loop().create_future()
```

jptrindade:
Could you add a simple test to show how we should use it?

jptrindade (approved):

PR #8135, opened 2024-09-30
Fix bug that causes pytest logging options to be ignored

arnaudsjs:
# Description

This PR fixes a bug that causes the pytest logging options (`--log-cli-level`, `--log-file-level`, etc.)  to be ignored. The bug was causes by the fact that a call to `logging.config.dictConfig()` closes all existing handlers, including the handlers added by pytest to handle the logs.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -69,6 +69,7 @@ environment.
 """
 
 import asyncio
+import weakref
 import concurrent
 import csv
 import datetime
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -1896,12 +1897,29 @@ async def dont_remove_caplog_handlers(request, monkeypatch):
     original_apply_config = inmanta_logging.FullLoggingConfig.apply_config
 
     def patched_apply_config(self) -> None:
-        caplog_handlers = [h for h in logging.root.handlers if is_caplog_handler(h)]
+        # Make sure the root log level is not altered.
```
sanderr, line 1900: Is this always desired? What with test cases that set an explicit log level?Could we pick the higher (more verbose) of the two?

tests/conftest.py
```diff
@@ -1896,12 +1897,29 @@ async def dont_remove_caplog_handlers(request, monkeypatch):
     original_apply_config = inmanta_logging.FullLoggingConfig.apply_config
 
     def patched_apply_config(self) -> None:
```
sanderr, line 1899: This method could use some documentation as to what it achieves. Do I understand correctly that it wraps the original `apply_config` to make sure we keep caplog handlers enabled? It does this by inspecting the handlers before the original `apply_config` is called so we can restore the caplog subset afterwards?

tests/conftest.py
```diff
@@ -1896,12 +1897,29 @@ async def dont_remove_caplog_handlers(request, monkeypatch):
...
+            logging.root.removeHandler(current_handler)
+        # Make sure that the weak references to the caplog handlers in the logging._handlerList are not removed.
+        # This list is used to tear down the handlers in the reverse order with respect to the setup order.
+        # As such, this method should not alter the order. We assume the caplog handlers are entirely independent
+        # from any other handler. Like that the order only matters within the set of caplog handlers.
+        re_add_to_handler_list: list[weakref.ReferenceType] = [
+            weak_ref for weak_ref in logging._handlerList if is_caplog_handler(weak_ref())
+        ]
+        for weak_ref in re_add_to_handler_list:
+            logging._handlerList.remove(weak_ref)
```
sanderr, line 1914: I don't get this part, or the other remove call. What's the difference between `caplog_root_handler` and `_handlerList`? Why do we need to remove these if we add them back in the same order anyway?

sanderr (changes requested):

sanderr (approved):

Hugo-Inmanta (approved):

PR #8139, opened 2024-09-30
Implement priority mechanism in new scheduler

### LINKED ISSUE #8015: resource scheduler: priorities
The scheduler already has the concept of priorities built-in, but it doesn't actually use them yet. Implement the priorities as in the design. Before starting on implementation, confirm with team that designed priorities are (still) relevant.

jptrindade:
# Description

Added priority mechanism to the new scheduler with priorities for every existing task.

closes #8015 

# Self Check:

- [x] Changelog entry


src/inmanta/deploy/work.py
```diff
@@ -34,6 +35,17 @@
...
 
+class TaskPriority(IntEnum):
+    TERMINATED = -1
+    USER_DEPLOY = 0
+    NEW_VERSION_DEPLOY = 1
+    USER_REPAIR = 2
+    DRYRUN = 3
+    INTERVAL_DEPLOY = 4
+    FACT_REFRESH = 5
+    INTERVAL_REPAIR = 6
```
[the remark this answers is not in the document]
jptrindade: I don't think it is much of an issue, we can change the numbers how we like because in the code we just use the task's name. @sanderr what do you think?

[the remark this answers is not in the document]
sanderr: Indeed, I care mostly about the relative order. I do think that a comment stating that these can be freely updated would be useful.

If we ever start writing priorities to the database it's different, but afaik we don't plan to.

Hugo-Inmanta (approved):

src/inmanta/deploy/scheduler.py
```diff
@@ -305,6 +305,7 @@ async def _new_version(
                 # ensure deploy for ALL dirty resources, not just the new ones
                 self._work.deploy_with_context(
                     self._state.dirty,
+                    TaskPriority.NEW_VERSION_DEPLOY,
```
sanderr, line 308: Nice, I hadn't considered this to be a separate case.

src/inmanta/deploy/scheduler.py
```diff
@@ -384,7 +385,11 @@ async def report_resource_state(
                 if details.attributes.get("send_event", False):
                     provides: Set[ResourceIdStr] = self._state.requires.provides_view().get(resource, set())
                     if provides:
-                        self._work.deploy_with_context(provides, deploying=self._deploying)
+                        # Not sure about this priority. I believe this method is called when a resource has a new state
+                        # and hence, new version.
```
sanderr, line 389: Reminder to drop this comment.

As to the question: this method is called when a resource finishes deploy (or at least we only reach this branch of it in that case). The deploy that is triggered here is the propagation of events to its dependents. e.g. a service's resource deploys successfully -> we schedule the lsm state transfer resource for deploy.

I wonder if we might want a separate priority level for event propagation. I would rank it pretty high (urgent) for fast event response, but perhaps given how eagerly we currently send them out (because we'll still default to `receive_events=True`), that might not be the best idea at the moment.

Overall, perhaps we should simply schedule it with the same priority as the resource that just finished? We could pass in the information to this method, or we could just query the agent queues mapping. The first is more direct, but it feels a bit clunky, given how generic this method is, so I think I prefer the second.

sanderr: Hmm, we don't even keep that information at the moment. I need to give this some more thought. Preliminary suggestion: make `AgentQueues._in_progress` a mapping from `Task` to `PrioritizedTask` instead of the current set, then query that here.

sanderr: Or perhaps have the `queue_get` method return `PrioritizedTask` instead of `Task`, and pass on the task context to this method?

I'll get back to this. I have to consider what fits best.

src/inmanta/deploy/work.py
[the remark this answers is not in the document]
sanderr: @jptrindade could you add such a comment?

src/inmanta/deploy/work.py
```diff
@@ -34,6 +35,17 @@
 T = TypeVar("T", bound=tasks.Task, covariant=True)
 
 
+class TaskPriority(IntEnum):
+    TERMINATED = -1
+    USER_DEPLOY = 0
+    NEW_VERSION_DEPLOY = 1
```
sanderr, line 41: I'm inclined to swap these two (0 and 1). @wouterdb WDYT?

src/inmanta/deploy/work.py
```diff
@@ -294,6 +307,7 @@ def reset(self) -> None:
     def deploy_with_context(
         self,
         resources: Set[ResourceIdStr],
+        priority: TaskPriority,
```
sanderr, line 310: I'm very much in favor of making this kw-only like the other args.

src/inmanta/deploy/work.py
```diff
@@ -377,11 +392,10 @@ def extend_blocked_on(resource: ResourceIdStr, new_blockers: set[ResourceIdStr])
                 # discard rather than remove because task may already be running, in which case we leave it run its course
                 # and simply add a new one
                 task: tasks.Deploy = tasks.Deploy(resource=resource)
-                priority: Optional[int] = self.agent_queues.discard(task)
+                task_priority: Optional[TaskPriority] = self.agent_queues.discard(task)
                 queued.remove(resource)
                 self._waiting[resource] = BlockedDeploy(
-                    # FIXME[#8015]: default priority
-                    task=PrioritizedTask(task=task, priority=priority if priority is not None else 0),
+                    task=PrioritizedTask(task=task, priority=task_priority if task_priority is not None else priority),
```
sanderr, line 398: Shouldn't we bump the priority here if it's higher?

src/inmanta/deploy/work.py
```diff
@@ -391,17 +405,23 @@ def extend_blocked_on(resource: ResourceIdStr, new_blockers: set[ResourceIdStr])
...
         # ensure desired resource deploys are scheduled
         for resource in resources:
+            prioritized_task = PrioritizedTask(task=tasks.Deploy(resource=resource), priority=priority)
             if is_scheduled(resource):
-                # Deploy is already scheduled / running. No need to do anything here. If any of its dependencies are to be
+                # Deploy is already scheduled / running. Check to see if this task has a higher priority than the one already
+                # scheduled. If it has, update the priority. If any of its dependencies are to be
                 # scheduled as well, they will follow the provides relation to ensure this deploy waits its turn.
+                if prioritized_task.task in self.agent_queues:
+                    self.agent_queues.queue_put_nowait(prioritized_task)
```
sanderr, line 414:
```suggestion
                    # simply add it again, the queue will make sure only the highest priority is kept
                    self.agent_queues.queue_put_nowait(prioritized_task)
```

sanderr (changes requested):
I still have to go over the tests.

src/inmanta/deploy/work.py
jptrindade: If we do add this as kw-only, what do you suggest we use as default?

sanderr: It can be kw-only without having a default.

sanderr: But you do make me realize that it's the only non-trivial required field, so perhaps having it positional is acceptable after all.

sanderr: I still prefer kw-only, but I'm not as strongly opinionated on it as in my first comment.

src/inmanta/deploy/scheduler.py
jptrindade: Have you thought more about this case?

src/inmanta/deploy/scheduler.py
```diff
@@ -169,20 +169,20 @@ async def stop(self) -> None:
...
         await asyncio.gather(*self._workers.values())
 
-    async def deploy(self) -> None:
+    async def deploy(self, priority: TaskPriority) -> None:
         """
         Trigger a deploy
         """
         async with self._scheduler_lock:
-            self._work.deploy_with_context(self._state.dirty, deploying=self._deploying)
+            self._work.deploy_with_context(self._state.dirty, priority, deploying=self._deploying_stale)
```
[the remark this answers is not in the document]
jptrindade: I merged in master now,  is it `deploying_latest`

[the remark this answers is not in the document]
sanderr: Yes

src/inmanta/deploy/scheduler.py
```diff
@@ -169,20 +169,20 @@ async def stop(self) -> None:
         self._work.agent_queues.send_shutdown()
         await asyncio.gather(*self._workers.values())
 
-    async def deploy(self) -> None:
+    async def deploy(self, priority: TaskPriority) -> None:
```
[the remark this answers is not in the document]
jptrindade: I made the changes but still don't quite understand why this is preferred . Is it so the user does not have to set the priority when triggering the deploy? But from what I understand they don't call this method directly, right?

src/inmanta/deploy/work.py
jptrindade: Why do you prefer kw-only?

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
sanderr: As I see it it's simply a matter of expertise. We structure code in modules and classes to encapsulate some behavior and associated knowledge. You then expose an interface on that encapsulation to the outside world. The goal is to make that as simple and intuitive as possible, by not requiring the user to reason on things that are really in the component's domain.

In this case: there is a sane default for the priority, i.e. when `deploy` is called, we expect it to be for a deploy. If the user really wants to override the priority, sure we'll let them, but we won't require them to know what the appropriate priority is.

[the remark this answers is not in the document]
sanderr: EDIT: "expertise" -> "domain expertise". I meant expertise of a code component, not of you or I.

src/inmanta/deploy/work.py
sanderr: 1. It's more explicit when the method is called, which I prefer for non-trivial arguments to a method. When you see the method name, you'll think "ok, we need to tell it what to deploy". It makes sense for this to be positional, and it makes sense for this to be the first argument because it's trivially expected to be there. But the same does not go for the priority.

Consider a `def add(x: int, y: int) -> int`, where it's obvious that it expects two arguments. However, when you add a context argument, it's not trivially expected there, e.g. `def add(x: int, y: int, *, print_result: bool = False)` where calling `add(1, 2, True)` would be unintuitive because based on the name you don't expect a boolean argument there. `add(1, 2, print_result=True)` is obvious to the reader, `add(1, 2, True) is vague at best, often confusing, and error-prone (see 2) at worst.

Now, because the priority is an enum rather than a simple int, it will always be clear what sort of context we're adding. So this argument only partially applies, which is why I don't feel so strongly about it.

2. In case of multiple non-trivial context arguments, there is no meaningful ordering, so you just pick an arbitrary order. For one, this can often lead to the confusion from 1, but it can also make refactors more difficult. Suppose you want to add an argument later. Either you have to put it at the end, or you have to make sure to update all callers. We have had (more than once) issues with exactly this: we refactor a method with multiple args of the same type (some with defaults), and we forgot to update some callers, leading to unexpected behavior.

Again, in this case, there's only the one argument, and it doesn't have a default, so argument 2 also doesn't fully apply.

Overall, I'm fine to leave it as is here, but the above has led me try to apply the reasoning as a general principle when I create an interface. I.e. think about "where does this argument make sense semantically?", "how will this read when called?", ... E.g. is it a trivial primary argument, or a context argument (as in the example with `add`)?

Sometimes there are strong motivations, sometimes it's more of a feeling that kw-only fits the context better. This situation here is more that subjective case, hence I express my preference but otherwise leave it up to you.

jptrindade: Thanks for the thorough reply. It's also something that I subconsciously think but reading it this way made perfect sense. I always try to do it when I'm the user of the function I try to do:  `add(1, 2, print_result=True)` but as a developer I was only using kw-only when I had optional arguments.

src/inmanta/agent/agent_new.py
```diff
@@ -132,8 +134,18 @@ def periodic_schedule(
                 return True
             return False
 
-        periodic_schedule("deploy", self.scheduler.deploy, self._deploy_interval, self._deploy_splay_value)
-        periodic_schedule("repair", self.scheduler.repair, self._repair_interval, self._repair_splay_value)
+        periodic_schedule(
+            "deploy",
+            functools.partial(self.scheduler.deploy, TaskPriority.INTERVAL_DEPLOY).func,
```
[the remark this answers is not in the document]
jptrindade: I used `func` because of this line that is called further in the code:
`LOGGER.info("Calling %s", action.__name__)`

and `functools.partial` does not have the `__name__` attr.  So this solution does not work.
I'll just go back to having dedicated function for this

Hugo-Inmanta (approved):

src/inmanta/agent/agent_new.py
```diff
@@ -132,8 +133,14 @@ def periodic_schedule(
                 return True
             return False
 
-        periodic_schedule("deploy", self.scheduler.deploy, self._deploy_interval, self._deploy_splay_value)
-        periodic_schedule("repair", self.scheduler.repair, self._repair_interval, self._repair_splay_value)
+        async def interval_deploy() -> None:
+            await self.scheduler.deploy(TaskPriority.INTERVAL_DEPLOY)
+
+        async def interval_repair() -> None:
+            await self.scheduler.repair(TaskPriority.INTERVAL_REPAIR)
```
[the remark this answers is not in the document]
sanderr: Hmm, I don't like this at all. I mean the fact that `mypy` is completely fine with the following, yet it gives an `AttributeError`:

```python
import functools
from collections.abc import Callable


def f() -> int:
    return 42


x: Callable[[], int] = functools.partial(f)
print(x.__name__)
```

But given those circumstances, I agree with your assessment.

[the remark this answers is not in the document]
sanderr: We could also make the scheduler more robust by trying `__name__` and falling back to `repr(func)`, but since that's out of scope for this ticket I'll leave the decision up to you.

src/inmanta/deploy/scheduler.py
sanderr: I would have liked to just slam a hardcoded priority on this. But I'm not sure we can:
1. if we pick a high priority (low number), we essentially bump 80% of deploys to that priority, because event propagation is so eager at the moment. i.e. if `std::File("hello/world")` depends on `std::Directory("hello")`, then the file would get the high priority, even though there is no need for it at all.
2. if we pick a low priority (high number), we might starve actually relevant events (e.g. lsm) after repairs, facts, ... on the same agent.

As I see it, 1 is out of the question. 2 could be argued to be acceptable if real event listeners are expected to be on a separate agent. But iirc this is not the case for policy install, which is one of the two known event listeners (the other being lsm).

@wouterdb do you agree that this invalidates our conclusion? Mainly 1: I initially posed that we would bump the priority on event listeners, but I was overlooking that we would actually bump it on non-listeners as well (because `receive_events=True` by default), which is the core of the problem.

So, if Wouter agrees, how do we implement it? My preference is to put it in `AgentQueues._in_progress` as `Mapping[Task, TaskPriority]`, mainly because this allows us to easily bump in-progress tasks' priority as well, in the `deploy_with_context` path. By bumping in-progress tasks' priorities, we essentially bump their events as well. Wdyt?

src/inmanta/deploy/work.py
```diff
@@ -394,7 +396,10 @@ def extend_blocked_on(resource: ResourceIdStr, new_blockers: set[ResourceIdStr])
                 task_priority: Optional[TaskPriority] = self.agent_queues.discard(task)
                 queued.remove(resource)
                 self._waiting[resource] = BlockedDeploy(
-                    task=PrioritizedTask(task=task, priority=task_priority if task_priority is not None else priority),
+                    task=PrioritizedTask(
+                        task=task,
+                        priority=task_priority if task_priority is not None and task_priority <= priority else priority,
```
sanderr, line 401: 100% optional suggestion that I find reads slightly easier. Feel free to leave as is if you prefer it like that.
```suggestion
                        priority=max(priority, task_priority) if task_priority is not None else priority,
```

sanderr (changes requested):
LGTM, requesting changes purely for the open discussions.

tests/agent_server/deploy/test_scheduler_agent.py
```diff
@@ -1127,3 +1128,111 @@ async def done():
     await retry_limited_fast(done)
 
     assert agent.executor_manager.executors["agent1"].facts_count == 1
+
+
+async def test_scheduler_priority(agent: TestAgent, environment, make_resource_minimal):
```
sanderr, line 1133: Nice, looks thorough!

src/inmanta/deploy/work.py
jptrindade: I believe here is `min` right? The lowest priority is the most urgent

jptrindade: But I agree that is much more readable

src/inmanta/deploy/work.py
wouterdb: Given the current api, they will both be about the exact same set of tasks, I think, so I don't think this is very important. 

Assuming it would matter, I like this order, because it gives the user control

jptrindade: (Meaning that I support swapping these two)

src/inmanta/deploy/scheduler.py
```diff
@@ -386,7 +387,15 @@ async def report_resource_state(
...
-                        self._work.deploy_with_context(event_listeners, deploying=set())
+                        priority = self._work.agent_queues.in_progress.get(Deploy(resource=resource))
+                        if priority is None:
+                            LOGGER.warning(
+                                "Deploy task for resource %s not found on in_progress."
+                                "Assigning a priority of NEW_VERSION_DEPLOY to the events it produced",
+                                resource,
+                            )
+                            priority = TaskPriority.NEW_VERSION_DEPLOY
+                        self._work.deploy_with_context(event_listeners, priority=priority, deploying=set())
```
jptrindade, line 398: @wouterdb @sanderr This is the part that we discussed. 
If we don't find the task in the queue I opted for logging a warning and assigning a "default" priority. What do you think?

sanderr (dismissed):

src/inmanta/deploy/work.py
```diff
@@ -232,7 +232,7 @@ async def queue_get(self, agent: str) -> tasks.Task:
                 continue
             # remove from the queue since it's been picked up
             self.discard(item.task.task)
-            self._in_progress.add(item.task.task)
+            self._in_progress[item.task.task] = item.task.priority
```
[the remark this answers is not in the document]
sanderr: > if that turns out to be cleaner

It probably is.

wouterdb (approved):

PR #8141, opened 2024-10-01
Environment halting and Agents pausing

hlloreda:
# Description

Remaining things:
- [x] Clean
- [x] Remove old todos
- [x] Fix broken test

# Self Check:

- [ ] Changelog entry


src/inmanta/server/agentmanager.py
```diff
@@ -210,14 +210,16 @@ async def halt_agents(self, env: data.Environment, connection: Optional[asyncpg.
         Halts all agents for an environment. Persists prior paused state.
         """
         await data.Agent.persist_on_halt(env.id, connection=connection)
-        await self._pause_agent(env, connection=connection)
+        await self._stop_agent(env, connection=connection)
```
hlloreda, line 213: before creating new executor if env not paused and agent not paused

hlloreda:
@wouterdb Indeed, as discussed, the first approach would be better:
image

image

The list of things I would need to implement are:
- [ ]  The Scheduler will need to read from the DB about agent instances + Environment when it starts / is notified by the Server that something has changed
- [ ] Refactor the current implementation to make sure that we don't restart an agent if the environment is resumed (this will depend on the boolean of the agent)
- [ ] Add test where the agent is paused and then the Scheduler dies -> make sure that the agent stays in paused state
- [ ] Add test where the agent is paused and then the the environment is resumed -> make sure that the agent stays in paused state (this will depend on the boolean of the agent, need to test both cases)

Do you see anything else?

wouterdb:
one more thing: an agent with no resources is paused, it receives first resource, the scheduler doesn't start it

src/inmanta/server/agentmanager.py
```diff
@@ -243,57 +243,87 @@ async def agent_action(self, env: data.Environment, name: str, action: AgentActi
...
+                if not live_session:
+                    await self._autostarted_agent_manager._ensure_scheduler(env)
+                    live_session = self.tid_endpoint_to_session.get(key)
+                    assert live_session
+
+                if endpoint is not None and endpoint != const.AGENT_SCHEDULER_ID:
+                    # We don't need to do this when the environment is resumed because the scheduler will need to have updated
+                    # information related to agents (being paused) and in order to have that, this transaction needs to
+                    # finish first
+                    await live_session.get_client().set_state(endpoint, enabled=False)
```
hlloreda, line 278: Maybe I should refactor this to avoid having almost identical blocks of code

src/inmanta/deploy/scheduler.py
```diff
@@ -357,6 +382,43 @@ async def _run_for_agent(self, agent: str) -> None:
...
+            - If the agent is not paused: It will make sure that the agent is running.
+            - If the agent is paused: Stop a particular agent.
+
+        :param name: The name of the agent
+        """
+        requested_agent = await data.Agent.get(env=self.environment, endpoint=name)
+
+        # This should never occur as the Scheduler is creating these 'agents'!
+        if requested_agent and name not in self._state.agent_status:
+            raise RuntimeError("Agent exists in Database but not in the Scheduler state!")
```
wouterdb, line 403: How do you know?  You have no control over when this method is called? (The server calls it)

src/inmanta/deploy/scheduler.py
```diff
@@ -357,6 +382,43 @@ async def _run_for_agent(self, agent: str) -> None:
 
             self._work.agent_queues.task_done(agent, task)
 
+    async def is_running(self) -> bool:
```
wouterdb, line 385: this is a bit of a weird name, is doesn't say of the scheduler is running or not, but if, according to the DB, it should be running?

src/inmanta/deploy/scheduler.py
```diff
@@ -158,6 +158,7 @@ def reset(self) -> None:
     async def start(self) -> None:
         self.reset()
         self._running = True
+        await self.read_agent_instances()
```
wouterdb, line 161: Why is this? It will read them again if/when they are started?

src/inmanta/deploy/scheduler.py
```diff
@@ -357,6 +382,43 @@ async def _run_for_agent(self, agent: str) -> None:
...
+        :param name: The name of the agent
+        """
+        requested_agent = await data.Agent.get(env=self.environment, endpoint=name)
+
+        # This should never occur as the Scheduler is creating these 'agents'!
+        if requested_agent and name not in self._state.agent_status:
+            raise RuntimeError("Agent exists in Database but not in the Scheduler state!")
+
+        if name not in self._state.agent_status and requested_agent is None:
+            raise LookupError(f"The agent `{name}` does not exist!")
```
wouterdb, line 406: How do you know? You have no control over when this method is called? (The server calls it)

src/inmanta/deploy/scheduler.py
```diff
@@ -357,6 +382,43 @@ async def _run_for_agent(self, agent: str) -> None:
...
+
+        if requested_agent.paused:
+            # We don't need to stop it, through the executor manager, because it will taking new task from the queue,
+            # so it will time out eventually
+            self._state.agent_status[name] = AgentStatus.STOPPED
+        else:
+            old_status = self._state.agent_status[name]
+            self._state.agent_status[name] = AgentStatus.STARTED
+            if old_status != AgentStatus.STARTED and (name not in self._workers or self._workers[name].done()):
+                self._start_for_agent(name)
```
wouterdb, line 420: this code path will go to the database twice, seems a bit wasteful?

src/inmanta/deploy/scheduler.py
```diff
@@ -344,11 +358,22 @@ async def _new_version(
...
+        sure that it's allowed to run."""
+        current_environment = await Environment.get_by_id(self.environment)
+        assert current_environment
+        if current_environment.halted:
+            return
+
+        await data.Agent(environment=self.environment, name=agent).insert_if_not_exist()
+        # We need to retrieve the actual state to see if the agent is actually paused (if it existed before)
+        agent_state = await data.Agent.get(env=self.environment, endpoint=agent)
+        self._state.agent_status[agent] = AgentStatus.STOPPED if agent_state.paused else AgentStatus.STARTED
```
wouterdb, line 374: I wonder if we want to have `started`, `stopped` and `stopping` states?

src/inmanta/deploy/scheduler.py
wouterdb: I think this may also cause an agent to not run at all. The fact that the task isn't done doesn't mean it won't fall flat on its face soon.... 


I think this code should go into a nice little object, so there is an overview.

src/inmanta/deploy/state.py
```diff
@@ -149,6 +161,7 @@ class ResourceState:
     status: ResourceStatus
     deployment_result: DeploymentResult
     blocked: BlockedStatus
+    agent_status: AgentStatus
```
wouterdb, line 164: This is believe is not something we have in the design?

src/inmanta/server/agentmanager.py
```diff
@@ -243,57 +243,87 @@ async def agent_action(self, env: data.Environment, name: str, action: AgentActi
...
         elif action is AgentAction.keep_paused_on_resume:
             await self._set_unpause_on_resume(env, should_be_unpaused_on_resume=False, endpoint=name)
         elif action is AgentAction.unpause_on_resume:
             await self._set_unpause_on_resume(env, should_be_unpaused_on_resume=True, endpoint=name)
         else:
             raise BadRequest(f"Unknown agent action: {action.name}")
 
-    async def _pause_agent(
+    async def _stop_agent(
         self, env: data.Environment, endpoint: Optional[str] = None, connection: Optional[asyncpg.connection.Connection] = None
```
wouterdb, line 257: Could you describe what is changed here, as I can't make sense of it.  (and this code is extremely sensitive)

src/inmanta/server/agentmanager.py
```diff
@@ -243,57 +243,87 @@ async def agent_action(self, env: data.Environment, name: str, action: AgentActi
...
-                        self.tid_endpoint_to_session[key] = session
-                        await session.get_client().set_state(agent_name, enabled=True)
-                        endpoints_with_new_primary.append((agent_name, session.id))
-            await data.Agent.update_primary(
-                env.id, endpoints_with_new_primary, now=datetime.now().astimezone(), connection=connection
-            )
+                    await self._autostarted_agent_manager._ensure_scheduler(env)
+                    live_session = self.tid_endpoint_to_session.get(key)
+                    assert live_session
+                    self.tid_endpoint_to_session[key] = live_session
```
wouterdb, line 305: why assign it back?

src/inmanta/server/agentmanager.py
```diff
@@ -1025,16 +1061,29 @@ async def stop_agents(
...
-                self._remove_venv_for_agent_in_env(env.id)
-
-            LOGGER.debug("Expiring all sessions for %s", env.id)
-            await self._agent_manager.expire_all_sessions_for_environment(env.id)
+            if opt.server_use_resource_scheduler.get():
+                agent_client = self._agent_manager.get_agent_client(
+                    tid=env.id, endpoint=const.AGENT_SCHEDULER_ID, live_agent_only=False
+                )
+                if agent_client is None:
+                    await self._ensure_scheduler(env)
```
wouterdb, line 1069: Why start it if we want to stop it?

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -244,3 +252,677 @@ def repaired() -> bool:
...
+                    or len(process_mapping["pg_ctl"]) == 1
+                    and len(process_mapping["python"]) == 1
+                    and len(process_mapping["inmanta fork server"]) == 1
+                )
+            case 4:
+                return (
+                    len(process_mapping["pg_ctl"]) == 1
+                    and len(process_mapping["python"]) == 2
+                    and len(process_mapping["inmanta fork server"]) == 1
+                )
```
wouterdb, line 360: Why this odd structure? Why not just pass in a list of processes and remove these that are allowed in stead of enumerating the possibilities?

Is it sufficient to check only `python` and nothing more?

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -244,3 +252,677 @@ def repaired() -> bool:
...
+    environment,
+    no_agent_backoff,
+    auto_start_agent: bool,
+    should_time_out: bool,
+    time_to_sleep: int,
+):
+    """
+    Verify that the new scheduler can actually halt an ongoing deployment and can resume it when the user requests it
+    """
+    current_pid = ensure_consistent_starting_point
```
wouterdb, line 409: this could do with some inline comments to show what is going on, for all of them

wouterdb (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -1494,6 +1494,21 @@ async def insert(self, connection: Optional[asyncpg.connection.Connection] = Non
...
+        Insert a new document based on the instance passed if it does not exist. Validation is done based on the defined fields.
+        """
+        (column_names, values) = self._get_column_names_and_values()
+        column_names_as_sql_string = ",".join(column_names)
+        values_as_parameterized_sql_string = ",".join(["$" + str(i) for i in range(1, len(values) + 1)])
+        query = (
+            f"INSERT INTO {self.table_name()} "
+            f"({column_names_as_sql_string}) "
+            f"VALUES ({values_as_parameterized_sql_string}) "
+            "ON CONFLICT DO NOTHING"
```
wouterdb, line 1508: I would be nice to add `returning paused` that saves one query and one change of a race

src/inmanta/agent/agent_new.py
```diff
@@ -181,48 +181,70 @@ async def start_working(self) -> None:
...
+
+        if agent == AGENT_SCHEDULER_ID:
+            if enabled:
+                await self.executor_manager.start()
+                await self.scheduler.start()
+                self._enable_time_triggers()
+            else:
+                await self.executor_manager.stop()
+                await self.executor_manager.join([], timeout=const.EXECUTOR_GRACE_HARD)
+                await self.scheduler.stop()
```
hlloreda, line 216: Need to be uncommented and removed

src/inmanta/server/agentmanager.py
hlloreda: The only difference is that if the scheduler is used, we call one of its endpoints to perform the action (and we first make sure that the agent client is up and running)

src/inmanta/agent/agent_new.py
```diff
@@ -181,40 +181,38 @@ async def start_working(self) -> None:
...
-    async def stop_working(self) -> None:
+    async def stop_working(self, timeout: Optional[int] = None) -> None:
         """Stop working, connection lost"""
         if not self.working:
             return
         self.working = False
         self._disable_time_triggers()
         await self.executor_manager.stop()
+        if timeout is not None:
+            await self.executor_manager.join([], timeout=timeout)
```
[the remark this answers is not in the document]
hlloreda: If a task (created by the scheduler) is taking a lot of time, cancelling the task is not enough, we need to kill the executor, right? Or did I miss something?

src/inmanta/agent/agent_new.py
```diff
@@ -181,48 +181,70 @@ async def start_working(self) -> None:
...
-            return 404, "No such agent"
+    @protocol.handle(methods.set_state)
+    async def set_state(self, agent: str, enabled: bool) -> Apireturn:
+        # if agent == AGENT_SCHEDULER_ID:
+        #     return (
+        #         400,
+        #         "Invalid request, cannot pause the scheduler! The closest action to that would be to pause the environment",
+        #     )
+
+        if agent == AGENT_SCHEDULER_ID:
```
hlloreda, line 208: Need to be uncommented and removed

hlloreda: But in order to do so, I need help with this comment: https://github.com/inmanta/inmanta-core/pull/8141#discussion_r1803063778

src/inmanta/deploy/scheduler.py
```diff
@@ -95,6 +96,71 @@ async def report_resource_state(
...
+        current_agent = await data.Agent.get(env=self._scheduler.environment, endpoint=self.endpoint)
+        if current_agent is None:
+            return
+        should_be_running = not (current_environment.halted or current_agent.paused)
+
+        match self.status:
+            case AgentStatus.STARTED if not should_be_running:
+                self.stop()
+                if current_environment.halted and self._task is not None:
+                    self._task.cancel("Environment is being halted")
```
hlloreda, line 132: Not sure if that's actually useful, as this would never get called in time

src/inmanta/agent/agent_new.py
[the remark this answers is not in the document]
wouterdb: 1. yes but if you don't stop the scheduler before joining, the scheduler keeps on pushing work to the executors, causing exception for no reason
2. we may even want to stop the scheduler first
3. why the if? i.e. it is not very expected for the method to return before the scheduler if down unless you pass in a timeout?

src/inmanta/deploy/scheduler.py
wouterdb: I would not do that, the run method is absolutely not sufficiently robust to handle the exception (e.g. `self.status = stopped` is not in a final block, the code below it may not handle it well either. )

src/inmanta/deploy/scheduler.py
```diff
@@ -345,18 +412,29 @@ async def _new_version(
...
+            return
+
+        requested_agent = await data.Agent.get(env=self.environment, endpoint=name)
+        if requested_agent:
+            self._start_for_agent(agent=requested_agent.name)
+            return
 
-            self._work.agent_queues.task_done(agent, task)
+        if name not in self._state.agent_status and requested_agent is None:
+            raise LookupError(f"The agent `{name}` does not exist!")
```
wouterdb, line 437: When is this code path relevant?

src/inmanta/deploy/state.py
wouterdb: can you link to it, I can't find it.

src/inmanta/deploy/state.py
```diff
@@ -171,6 +186,7 @@ class ModelState:
     types_per_agent: dict[str, dict[ResourceType, int]] = dataclasses.field(
         default_factory=lambda: defaultdict(lambda: defaultdict(lambda: 0))
     )
+    agent_status: dict[str, AgentStatus] = dataclasses.field(default_factory=dict)
```
wouterdb, line 189: what is this for?

wouterdb (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -3394,6 +3394,26 @@ async def get(
...
+        )
+        new_agent_paused_status = await self._fetchval(query, *values, connection=connection)
+        # The entry already exists, we ended up in a conflict, so None is returned
+        if new_agent_paused_status is None:
+            current_agent = await self.get(env=self.environment, endpoint=self.name)
+            assert current_agent is not None
+            return current_agent.paused
+        else:
+            return bool(new_agent_paused_status)
+
```
hlloreda, line 3416: Primary + this db request

src/inmanta/server/agentmanager.py
```diff
@@ -238,62 +238,107 @@ async def all_agents_action(self, env: data.Environment, action: AgentAction) ->
...
+                    # We don't need to do this when the environment is resumed because the scheduler will need to have updated
+                    # information related to agents (being paused) and in order to have that, this transaction needs to
+                    # finish first
+                    await live_session.get_client().set_state(endpoint, enabled=False)
+                elif endpoint is None:
+                    # We need to update information in the DB
+                    endpoints_with_new_primary.append((const.AGENT_SCHEDULER_ID, None))
+                    await data.Agent.update_primary(
+                        env.id, endpoints_with_new_primary, now=datetime.now().astimezone(), connection=connection
+                    )
```
hlloreda, line 288: @wouterdb  I'm not really convinced by my implementation (the same can be said for the `start_agent` method): because if the scheduler is used, we cannot remove the live session here (as we would do if the scheduler wasn't used) because we still need to notify the scheduler that something has changed in the DB. But to do that, we need the transaction to go through. Any suggestion?

src/inmanta/agent/agent_new.py
```diff
@@ -198,48 +198,44 @@ async def start_working(self) -> None:
...
 
     async def on_reconnect(self) -> None:
-        name = AGENT_SCHEDULER_ID
-        result = await self._client.get_state(tid=self._env_id, sid=self.sessionid, agent=name)
+        result = await self._client.get_state(tid=self._env_id, sid=self.sessionid, agent=AGENT_SCHEDULER_ID)
         if result.code == 200 and result.result is not None:
             state = result.result
             if "enabled" in state and isinstance(state["enabled"], bool):
-                await self.set_state(name, state["enabled"])
+                await self.start_working()
```
Hugo-Inmanta, line 238: Only if `state["enabled"]` is True I would assume ?

src/inmanta/deploy/scheduler.py
```diff
@@ -95,6 +96,57 @@ async def report_resource_state(
...
+            case AgentStatus.STARTED if not should_be_running:
+                self.stop()
+            case AgentStatus.STOPPING | AgentStatus.STOPPED if should_be_running:
+                self.start()
+
+    async def run(self) -> None:
+        """Main loop for one agent. It will first fetch or create its actual state from the DB to make sure that it's
+        allowed to run."""
+        paused_status = await data.Agent.retrieve_paused_status(environment=self._scheduler.environment, endpoint=self.endpoint)
+        # The entry already exists, we ended up in a conflict, so None is returned
```
Hugo-Inmanta, line 133:
```suggestion

```

src/inmanta/deploy/scheduler.py
```diff
@@ -345,18 +400,43 @@ async def _new_version(
...
-        while self._running:
-            task: Task = await self._work.agent_queues.queue_get(agent)
-            try:
-                await task.execute(self, agent)
-            except Exception:
-                LOGGER.exception("Task %s for agent %s has failed and the exception was not properly handled", task, agent)
+    async def should_be_running(self, endpoint: str) -> bool:
+        """
+        Return if the agent (or the Scheduler if endpoint == Scheduler id) should be running. This will also
+            check if the environment is halted
```
Hugo-Inmanta, line 409: I like the "authoritative entity" phrasing from the other docstring. What about:
```suggestion
        Check in the DB (authoritative entity) if the agent (or the Scheduler if endpoint == Scheduler id) should be running i.e. if it is not paused and its environment is not halted.
```

src/inmanta/agent/agent_new.py
```diff
@@ -198,48 +198,44 @@ async def start_working(self) -> None:
...
+            else:
+                await self.stop_working(timeout=const.EXECUTOR_GRACE_HARD)
         else:
-            return await self.pause(agent)
+            try:
+                await self.scheduler.refresh_agent_state_from_db(name=agent)
+            except LookupError:
+                return 404, f"No such agent: {agent}"
+
+        return 200, f"{agent} has been {'started' if enabled else 'stopped'}"
```
Hugo-Inmanta, line 231: The msg `f"{agent} has been {'started' if enabled else 'stopped'}"` relies on `enabled` but `refresh_agent_state_from_db` doesn't. We might get inconsistent behaviour no ?

src/inmanta/deploy/scheduler.py
```diff
@@ -345,18 +400,43 @@ async def _new_version(
...
+        :param name: The name of the agent
+        """
+
+        if name in self._workers:
+            await self._workers[name].notify()
+            return
+
+        requested_agent = await data.Agent.get(env=self.environment, endpoint=name)
+        if requested_agent:
+            self._start_for_agent(agent=requested_agent.name)
```
Hugo-Inmanta, line 435: Only if it is not paused in the DB I would assume ?

Hugo-Inmanta (changes requested):
Partial review so far, haven't looked at `tests/agent_server/deploy/e2e/test_autostarted.py` and `src/inmanta/server/agentmanager.py` in details yet.

src/inmanta/deploy/scheduler.py
hlloreda: This one was safe because even if a `TaskRunner` is started, it will first check its actual state from the DB, so it would see at some point that it is not allowed to run. But I refactored that part, to avoid making an additional sql request if we already have the information.

src/inmanta/deploy/state.py
```diff
@@ -177,6 +177,11 @@ class ModelState:
...
     resources: dict[ResourceIdStr, ResourceDetails] = dataclasses.field(default_factory=dict)
     requires: RequiresProvidesMapping = dataclasses.field(default_factory=RequiresProvidesMapping)
+    """
+    Resources that have a new desired state (might be simply a change of its dependencies), which are still being processed by
+    the resource scheduler. This is a short-lived transient state, used for internal concurrency control. Kept separate from
+    ResourceStatus so that it lives outside of the scheduler lock's scope.
+    """
     resource_state: dict[ResourceIdStr, ResourceState] = dataclasses.field(default_factory=dict)
     # resources with a known or assumed difference between intent and actual state
     dirty: set[ResourceIdStr] = dataclasses.field(default_factory=set)
```
Hugo-Inmanta, line 187: I think this is describing the `dirty` set.
```suggestion
    resource_state: dict[ResourceIdStr, ResourceState] = dataclasses.field(default_factory=dict)
    # resources with a known or assumed difference between intent and actual state
    # (might be simply a change of its dependencies), which are still being processed by
    # the resource scheduler. This is a short-lived transient state, used for internal concurrency control. Kept separate from
    # ResourceStatus so that it lives outside of the scheduler lock's scope.
    dirty: set[ResourceIdStr] = dataclasses.field(default_factory=set)
```

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -244,3 +253,750 @@ def repaired() -> bool:
...
+            - a postgres server, the scheduler and the resource tracker
+            - a postgres server, the scheduler and one inmanta fork server (unrelated to the new scheduler)
+        - 4 processes: a postgres server, the scheduler, one inmanta fork server and the resource tracker
+
+    :param agent: The agent fixture (that we want to stop before running the test)
+    :param ensure_resource_tracker_is_started: The fixture that creates a Resource Tracker process
+    """
+    current_pid = os.getpid()
+    await agent.stop()
+    return current_pid
```
wouterdb, line 305: I don't understand this?

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -244,3 +253,750 @@ def repaired() -> bool:
...
+                    relevant_processes["scheduler"] = child
+            case "pg_ctl":
+                pass
+            case _:
+                assert False
+    return relevant_processes
+
+
+@pytest.mark.parametrize(
+    "auto_start_agent,should_time_out,time_to_sleep,", [(True, False, 2), (True, True, 120)]
```
wouterdb, line 357: Why two variants?

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -244,3 +253,750 @@ def repaired() -> bool:
...
+    expected_agents_status = {e["name"]: e["paused"] for e in result.result["agents"]}
+    assert set(expected_agents_status.keys()) == {const.AGENT_SCHEDULER_ID, "agent1"}
+    assert not expected_agents_status[const.AGENT_SCHEDULER_ID]
+    assert not expected_agents_status["agent1"]
+
+    # Now let's halt the environment
+    result = await client.halt_environment(tid=environment)
+    assert result.code == 200
+
+    # Let's recheck the agent table and check that the scheduler and agent1 are present and paused
```
wouterdb, line 467: I would release a new version here and assert nothing gets deployed

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -244,3 +253,750 @@ def repaired() -> bool:
...
+        "These processes should be present: The Scheduler and the fork server. "
+        "The agent created by the scheduler should have been killed!"
+    )
+    for child in current_halted_children.values():
+        assert child.is_running()
+
+    await client.resume_environment(environment)
+
+    # Let's check the agent table and check that the scheduler and agent1 are present and not paused
+    result = await client.list_agents(tid=environment)
```
wouterdb, line 499: I would assert resource get deployed

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -244,3 +253,750 @@ def repaired() -> bool:
...
+    )
+
+    # Let's recheck the number of processes after pausing the agent
+    halted_children = get_process_state(current_pid)
+    assert len(halted_children) == 1
+    assert len(halted_children.values()) == 1
+    current_halted_children = filter_relevant_processes(halted_children)
+
+    # We should have lost one process in the "process"
+    assert len(current_halted_children) == len(current_children_after_deployment) - 1
```
wouterdb, line 657: Why? All these agent share an executor process  and filter_relevant_processes collapse all executors into one entry anyways. So you just wait for the executor cleanup, which has no relation to halting.

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -244,3 +253,750 @@ def repaired() -> bool:
...
+    for children in current_children_children_after_restart.values():
+        assert children.is_running()
+
+    result = await client.resource_list(environment, deploy_summary=True)
+    assert result.code == 200
+    summary = result.result["metadata"]["deploy_summary"]
+    assert summary["total"] == 1, f"Unexpected summary: {summary}"
+    # FIXME this should be fixed -> old resource is still in deploying state, should be available
+    # Uncomment this once fixed, see https://github.com/inmanta/inmanta-core/issues/8216
+    # assert summary["by_state"]["available"] == 1, f"Unexpected summary: {summary}"
```
wouterdb, line 825: I would assert that it is in deploying, as this doesn't really test anything like this. It could be deploying for all we care

tests/agent_server/deploy/test_scheduler_agent.py
```diff
@@ -178,6 +214,7 @@ def __init__(
     ):
         super().__init__(environment)
         self.executor_manager = DummyManager()
+        self.scheduler = TestScheduler(self.scheduler.environment, self.scheduler.executor_manager, self.scheduler.client)
```
wouterdb, line 217: We mix two styles of mocking: subclass Vs overwriting methods. 

I would choose one.

tests/agent_server/deploy/test_scheduler_agent.py
```diff
@@ -170,6 +173,39 @@ async def pass_method():
...
+    def __init__(self, environment: uuid.UUID, executor_manager: executor.ExecutorManager[executor.Executor], client: Client):
+        super().__init__(environment, executor_manager, client)
+
+    async def should_be_running(self, endpoint: str) -> bool:
+        return True
+
+    def _create_agent(self, agent: str, should_start: bool = True) -> None:
+        """Start processing for the given agent"""
+        self._workers[agent] = TestTaskRunner(endpoint=agent, scheduler=self)
+        self._workers[agent].start()
```
wouterdb, line 186: with some changes, we may be able to get awayt with just replacing the should_run method

tests/conftest.py
```diff
@@ -812,7 +812,8 @@ async def server(server_pre_start) -> abc.AsyncIterator[Server]:
     yield ibl.restserver
 
     try:
-        await ibl.stop(timeout=15)
+        # If the shutdown is only applied after `const.SHUTDOWN_GRACE_HARD`, we cannot assume this will be done instantly
```
wouterdb, line 815: Could you explain a bit more?

wouterdb (changes requested):

src/inmanta/deploy/scheduler.py
```diff
@@ -343,20 +400,62 @@ async def _new_version(
...
+            # The None represent the all agents action
+            agents_to_wake_up = await data.Agent.get_list(env=self.environment)
+            for agent in agents_to_wake_up:
+                if agent.name == const.AGENT_SCHEDULER_ID:
+                    continue
+
+                if agent.name in self._workers:
+                    await self._workers[agent.name].notify()
+                else:
+                    self._create_agent(agent=agent.name)
```
hlloreda, line 450: This should be moved, but I propose to do this in another ticket

src/inmanta/agent/agent_new.py
```diff
@@ -198,48 +198,51 @@ async def start_working(self) -> None:
...
     @protocol.handle(methods.set_state)
     async def set_state(self, agent: str, enabled: bool) -> Apireturn:
-        if enabled:
-            return await self.unpause(agent)
+        if agent == AGENT_SCHEDULER_ID:
+            should_be_running = await self.scheduler.should_be_running(agent)
+            enabled = should_be_running
+
+            if should_be_running:
+                await self.start_working()
```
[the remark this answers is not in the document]
hlloreda: What we could do is to notify all agents, wdyt?

Hugo-Inmanta (commented):
Partial review so far, holding off until after wouter's comments are addressed.

src/inmanta/agent/agent_new.py
```diff
@@ -206,51 +208,65 @@ async def start_working(self) -> None:
...
+            should_be_running = await self.scheduler.should_be_running(agent)
+
+            if should_be_running:
+                if self.working != should_be_running:
+                    await self.start_working()
+                else:
+                    # Special cast that the server considers us disconnected, but the Scheduler thinks we are still connected.
+                    # In that case, the Scheduler may have missed some event, but it would get a start after a start.
+                    # Therefore, we need to refresh everything (Scheduler side) to make sure we are up to date
+                    await self.scheduler.read_version()
```
hlloreda, line 242: Maybe it would be better to use the initialize method to only compute the increment?

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,720 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+
+    await client.resume_environment(environment)
+
+    # Let's check the agent table and check that the scheduler and agent1 are present and not paused
+    await assert_is_paused(client, environment, {const.AGENT_SCHEDULER_ID: False, "agent1": False})
+
+    if should_time_out:
+        # Wait for at least one resource to be in deploying
+        await retry_limited(are_resources_being_deployed, timeout=5)
+        await asyncio.sleep(0.5)
```
hlloreda, line 615: I did this because I had this assertion error (forked executor had the same name of the fork server for a small amount of time):
```
            match child.name():
                case "inmanta: multiprocessing fork server":
>                   assert not fork_server
E                   AssertionError: assert not psutil.Process(pid=81210, name='inmanta: multiprocessing fork server', status='sleeping', started='14:55:36')

/home/hugo/Documents/Inmanta/copies/inmanta-core/tests/deploy/e2e/test_autostarted.py:439: AssertionError
-> assert not fork_server
(Pdb) children
[psutil.Process(pid=81209, name='python', status='sleeping', started='14:55:36'), psutil.Process(pid=81210, name='inmanta: multiprocessing fork server', status='sleeping', started='14:55:36'), psutil.Process(pid=81213, name='inmanta: executor process 78f726cec1879275a4cc2fa2c103afb5 - connected', status='sleeping', started='14:55:36')]
(Pdb) fork_server
psutil.Process(pid=81210, name='inmanta: multiprocessing fork server', status='sleeping', started='14:55:36')
(Pdb) child
psutil.Process(pid=81213, name='inmanta: executor process 78f726cec1879275a4cc2fa2c103afb5 - connected', status='sleeping', started='14:55:36')
(Pdb) 
```
Is there a better (or cleaner) solution to fix this?

src/inmanta/agent/agent_new.py
```diff
@@ -206,51 +208,65 @@ async def start_working(self) -> None:
...
-        LOGGER.info("Scheduler stopped for environment %s", self.environment)
-        await self.stop_working()
-        return 200
-
     @protocol.handle(methods.set_state)
     async def set_state(self, agent: str, enabled: bool) -> Apireturn:
-        if enabled:
-            return await self.unpause(agent)
+        if agent == AGENT_SCHEDULER_ID:
+            should_be_running = await self.scheduler.should_be_running()
```
wouterdb, line 233:
```suggestion
            should_be_running = enabled
```
it becomes too complicated otherwise

src/inmanta/agent/agent_new.py
```diff
@@ -190,7 +191,8 @@ async def stop(self) -> None:
         self.thread_pool.shutdown(wait=False)
 
         await join_threadpools(threadpools_to_join)
-        await super().stop()
+        # We need to shield to avoid CancelledTask exception
+        await asyncio.shield(super().stop())
```
wouterdb, line 195: do we still need the shield?

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,745 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+        current_children=state_after_deployment.children,
+        expected_terminated_process=3,
+    )
+
+    # Let's recheck the number of processes after pausing the environment
+    halted_state = construct_scheduler_children(current_pid)
+
+    assert (
+        len(halted_state.children) == 0
+    ), "The Scheduler and the fork server and the agent created by the scheduler should have been killed!"
```
wouterdb, line 623:
```suggestion
    ), "The Scheduler and the fork server and the executor created by the scheduler should have been killed!"
```

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,745 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+    for child in halted_state.children:
+        assert child.is_running()
```
wouterdb, line 625: there are none?

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,745 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+
+    # Wait for at least one resource to be deployed
+    await retry_limited(check_resource_in_state, timeout=10)
+
+    # Retrieve the current processes, we should have more processes than `start_children`
+    state_after_deployment = construct_scheduler_children(current_pid)
+    expected_additional_children_after_deployment = 3
+    assert (
+        len(state_after_deployment.children) == expected_additional_children_after_deployment
+    ), "These processes should be present: The Scheduler, the fork server and the actual agent!"
```
wouterdb, line 719:
```suggestion
    ), "These processes should be present: The Scheduler, the fork server and the executor!"
```

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,745 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+
+    result = await client.agent_action(tid=environment, name="agent1", action=AgentAction.pause.value)
+    assert result.code == 200
+
+    # Let's pretend that the server crashes
+    await asyncio.wait_for(server.stop(), timeout=20)
+    ibl = InmantaBootloader(configure_logging=False)
+    async_finalizer.add(partial(ibl.stop, timeout=20))
+    # Let's restart the server
+    await ibl.start()
```
wouterdb, line 867: this is a clean shutdown, not a crash, but I guess getting closer to real crash is very hard. 

This is more a test for restart than crash

tests/utils.py
```diff
@@ -379,14 +379,12 @@ async def sufficient_versions():
...
 
-async def wait_until_deployment_finishes(client: Client, environment: str, version: int = -1, timeout: int = 10) -> None:
+async def wait_until_deployment_finishes(client: Client, environment: str, timeout: int = 10) -> None:
 
     async def done():
         result = await client.resource_list(environment, deploy_summary=True)
         assert result.code == 200
         summary = result.result["metadata"]["deploy_summary"]
-        # {'by_state': {'available': 3, 'cancelled': 0, 'deployed': 12, 'deploying': 0, 'failed': 0, 'skipped': 0,
-        #               'skipped_for_undefined': 0, 'unavailable': 0, 'undefined': 0}, 'total': 15}
```
wouterdb, line 389: I like this comment, it helps if I need to update this code, I want it back

src/inmanta/deploy/scheduler.py
```diff
@@ -520,13 +524,7 @@ async def refresh_agent_state_from_db(self, name: str) -> None:
...
             await self._workers[name].notify()
-            return
-
-        requested_agent = await data.Agent.get(env=self.environment, endpoint=name)
-        if requested_agent:
-            self._create_agent(agent=requested_agent.name)
-
-        if name not in self._workers and requested_agent is None:
+        else:
             raise LookupError(f"The agent `{name}` does not exist!")
```
wouterdb, line 528: This can go as well, it is possible/normal if we don't have it.

wouterdb (changes requested):
one last minor remark

wouterdb (approved):

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,774 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+        should_fork_server_be_defined=True,
+        executor_to_be_defined=1,
+    )
+
+    current_children_mapping = construct_scheduler_children(current_pid)
+    for children in current_children_mapping.children:
+        assert children.is_running()
+
+    # Retrieve the current processes, we should have more processes than `start_children`
+    children_after_deployment = construct_scheduler_children(current_pid)
```
hlloreda, line 873: Duplicate + need to mark these tests as slow

tests/conftest.py
```diff
@@ -743,7 +743,7 @@ async def server(server_pre_start) -> abc.AsyncIterator[Server]:
     yield ibl.restserver
 
     try:
-        await ibl.stop(timeout=15)
+        await ibl.stop(timeout=20)
```
Hugo-Inmanta, line 746: Necessary or test artifact ?

tests/server/test_notifications.py
```diff
@@ -415,7 +415,7 @@ async def notification_cleaned_up(env_id: uuid.UUID, expected_length_after_clean
 
 @pytest.mark.parametrize("halted", [True, False])
 @pytest.mark.parametrize("no_agent", [True])
-async def test_cleanup_notifications(server, client, halted_env_with_old_notifications, halted):
+async def test_cleanup_notifications(server, client, halted_env_with_old_notifications, halted, no_agent: bool):
```
Hugo-Inmanta, line 418: This doesn't seem to be used in the test ? Or am I missing something ? (same in a couple of other places)

Hugo-Inmanta (changes requested):
Partial review so far, haven't looked at `tests/deploy/e2e/test_autostarted.py` yet.

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,779 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+        try:
+            if process.status() == "terminated":
+                terminated_process.append(process)
+        except psutil.NoSuchProcess:
+            terminated_process.append(None)
+
+    return len(terminated_process) == expected_terminated_process
+
+
+def construct_scheduler_children(current_pid: int) -> SchedulerChildren:
```
Hugo-Inmanta, line 397: Missing docstring alert

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,779 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+    def get_process_state(current_pid: int) -> list[Process]:
```
Hugo-Inmanta, line 398: `get_process_children` maybe more accurate?

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,779 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+    # Now let's halt the environment
+    result = await client.halt_environment(tid=environment)
+    assert result.code == 200
+
+    snippetcompiler.setup_for_snippet(
+        f"""
+    import minimalwaitingmodule
+
+    a = minimalwaitingmodule::Sleep(name="test_sleep", agent="agent1", time_to_sleep={time_to_sleep})
+    """,
```
Hugo-Inmanta, line 618: Very minor detail the model could be factorized and reused from L546

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,779 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+
+    # Now let's pause agent1
+    result = await client.agent_action(tid=environment, name="agent1", action=AgentAction.pause.value)
+    assert result.code == 200
+
+    # Let's recheck the agent table and check that agent1 is present and paused
+    await assert_is_paused(client, environment, {const.AGENT_SCHEDULER_ID: False, "agent1": True})
+
+    # Let's also check if the state of resources are consistent with what we expect: one resource still needs to be
+    # deployed
```
Hugo-Inmanta, line 757:
```suggestion
    # Let's also check if the state of resources are consistent with what we expect:
    # The agent finished deploying resource n°2 before pausing and resource n°3
    # still needs to be deployed
```

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,779 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+    await retry_limited(check_resource_in_state, timeout=6, expected_resources=2)
+    result = await client.resource_list(environment, deploy_summary=True)
+    assert result.code == 200
+    summary = result.result["metadata"]["deploy_summary"]
+    assert summary["total"] == 3, f"Unexpected summary: {summary}"
+    assert summary["by_state"]["available"] == 1, f"Unexpected summary: {summary}"
+    assert summary["by_state"]["deployed"] == 2, f"Unexpected summary: {summary}"
+
+    # Let's wait for the executor to die
```
Hugo-Inmanta, line 766: Why does this kill the executor ?

tests/deploy/e2e/test_autostarted.py
```diff
@@ -315,3 +360,779 @@ async def test_autostart_clear_agent_venv_on_delete(
...
+    for children in state_after_restart.children:
+        assert children.is_running()
+
+    result = await client.resource_list(environment, deploy_summary=True)
+    assert result.code == 200
+    summary = result.result["metadata"]["deploy_summary"]
+    assert summary["total"] == 1, f"Unexpected summary: {summary}"
+    # FIXME this should be fixed -> old resource is still in deploying state, should be available
+    # Uncomment this once fixed, see https://github.com/inmanta/inmanta-core/issues/8216
+    # assert summary["by_state"]["available"] == 1, f"Unexpected summary: {summary}"
```
Hugo-Inmanta, line 913: Could you add a permalink to this part in the referenced ticket if it is still relevant ?

src/inmanta/agent/agent_new.py
```diff
@@ -206,51 +207,60 @@ async def start_working(self) -> None:
...
-        if enabled:
-            return await self.unpause(agent)
+    async def set_state(self, agent: Optional[str], enabled: bool) -> Apireturn:
+        if agent == AGENT_SCHEDULER_ID:
+            if enabled:
+                if self.working != enabled:
+                    await self.start_working()
+                else:
+                    # Special cast that the server considers us disconnected, but the Scheduler thinks we are still connected.
+                    # In that case, the Scheduler may have missed some event, but it would get a start after a start.
```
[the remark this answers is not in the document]
hlloreda: I refactored that part, is it better?

tests/conftest.py
hlloreda: It's necessary, I added a comment

tests/test_agent_manager.py
```diff
@@ -787,10 +787,11 @@ async def test_session_creation_fails(server, environment, async_finalizer, capl
 
 
 @pytest.mark.parametrize("no_agent", [True])
-async def test_agent_actions(server, client, async_finalizer):
+async def test_agent_actions(server, client, async_finalizer, no_agent: bool):
     """
     Test the agent_action() and the all_agents_action() API call.
     """
+    pytest.skip("Need to be migrated with the new Scheduler")
```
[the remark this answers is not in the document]
hlloreda: yes

tests/server/test_notifications.py
hlloreda: My IDE was complaining that this argument was not given, whereas the decorator provides it

tests/deploy/e2e/test_autostarted.py
hlloreda: Would you like to change this to a function? If that's the goal, I think the test case will be harder to read

tests/deploy/e2e/test_autostarted.py
hlloreda: It's not being killed but cleaned up. I refactored these comments

tests/deploy/e2e/test_autostarted.py
hlloreda: It's already been fixed in another PR, so I think we don't need to create a ticket as it should be merged soon.

tests/deploy/e2e/test_autostarted.py
Hugo-Inmanta: No I meant this string into a variable, but this is fine either way really.

tests/server/test_notifications.py
Hugo-Inmanta: Right I see, let's leave it up for the big test review

src/inmanta/agent/agent_new.py
```diff
@@ -206,51 +206,64 @@ async def start_working(self) -> None:
...
+                if self.working != enabled:
+                    await self.start_working()
+                else:
+                    # Special case that the server considers us disconnected, but the Scheduler thinks we are still connected.
+                    # In that case, the Scheduler may have missed some event, but it would get a start after a start:
+                    # agent_new --Start--> Scheduler (the scheduler is started)
+                    # Server --disconnected-- agent_new (after some time, the server considers us disconnected)
+                    # When the server sees the scheduler again, this would happen (logic in agent_new.on_reconnect):
+                    # agent_new --Start--> Scheduler (second start since we have already received one)
+                    # Therefore, we need to refresh everything (Scheduler side) to make sure we are up to date
```
Hugo-Inmanta, line 241:
```suggestion
                    # Special case that the server considers us disconnected, but the Scheduler thinks we are still connected.
                    # In that case, the Scheduler may have missed some event, therefore, we need to refresh everything 
                    # (Scheduler side) to make sure we are up to date.
```

Hugo-Inmanta (approved):

PR #8144, opened 2024-10-02
update examples

Hugo-Inmanta:
# Description

Core part of https://github.com/inmanta/lsm/pull/816

# Self Check:

- [ ] Changelog entry


docs/lsm/embedded_entities/embedded_entities.rst
```diff
@@ -147,12 +147,33 @@ Tracking embedded entities across updates
...
+.. note::
+    The "simple" lifecycle defines out-of-the-box which pair of sets should be compared at each step of the update.
+    Please refer to the :ref:`Tracking embedded entities when using a custom lifecycle<using_custom_lifecycle>` section
+    below for more information on how to define which pairs should be compared when using a custom lifecycle.
+
+.. note::
+    To set a different naming scheme for these tracking attributes, use the ``removed_attribute`` and
+    ``added_attribute`` parameters of the ``insert_removed_embedded_entities`` plugin.
+
+
```
wouterdb, line 170: I think this plugin only works when `strict_modifier_enforcement=true` but I'm not sure. I think it is because embedded entities need an identity to be able to compare them and  `strict_modifier_enforcement` adds that. 

Something to test and document 

On closer inspection, it needs to have '__lsm_key_attributes' set IF it is of arity higher then 1?

(see slack)

docs/lsm/embedded_entities/embedded_entities.rst
```diff
@@ -208,6 +230,46 @@ migration from cluster A to cluster B:
...
+    * Previous set
+        - The previous set is the set defined by the ``previous_attr_set_on_validate`` parameter.
+
+2. During an exporting compile:
+    * Current set
+        - The current set is the active attributes set
+
+    * Previous set
+        - The previous set is the set defined by the ``previous_attr_set_on_export`` parameter.
+
```
wouterdb, line 267:
```suggestion
The following logic is used to determine which is the current and which is the previous attribute set
|                                 | p r e v i o u s                  | current           |
|---------------------------------|----------------------------------|-------------------|
| instance is being validate      | ``previous_attr_set_on_validate` | ``validate_self`` |
| instance is not being validated | ``previous_attr_set_on_export``  | active_attributes |
```

may need some styling

docs/lsm/embedded_entities/embedded_entities.rst
```diff
@@ -208,6 +230,46 @@ migration from cluster A to cluster B:
...
+
+We recommend to manually go through all the states that are part of the update, think about which
```
wouterdb, line 269: I would add the actual method we used as well

In order to analyze a lifecycle, make a table as follows:

collumns: state, validation, previous attributes, current attributes, is_like, operation to is_like.

1. add one line for each state that is involved on an update or rollback scenario 
1. add the states just before and after the update (e.g. up and up)
1. add an additional line for each state that has a validating transfer 
    1. for the validating line, set `is_like` to the state we validating for (which state do we pretend to be, for which state do we want to know where are valid). Usually this state is soon after the validating transfer
    2. for the non-validating line for this transfer, set `is_like` to the state we try to maintain while validating. This is usually the state immediately before
    3. set the operations performed (promote, rollback) between the state and the `is_like` in `operation to is_like`  (for the non-validating line, this should be no operation)
1. for the main state in which the update is performed, set the `current attributes` to `active_attributes`  and the `previous_attributes ` to either `rollback attributes` for an update scenario and `candidate attributes` for a rollback scenario
2. for states with `is_like` set, apply the operations in reverse to the state they are like and previous attributes and  current attributes appropriately 


I would work this out for the simple lifecycle and add method and the example to the docs.

What I describe may be incomplete.

docs/lsm/embedded_entities/embedded_entities_sources/example_lines_in_file.cf
```diff
@@ -3,6 +3,7 @@ import lsm
 import lsm::fsm
 import fs
 import mitogen
+import lsm::embedded_entities_tracker
```
wouterdb, line 6: I would put it in the root package

wouterdb (changes requested):

docs/lsm/embedded_entities/embedded_entities.rst
```diff
@@ -208,6 +230,46 @@ migration from cluster A to cluster B:
 
 For these more involved update scenarios we recommend updating the lifecycle specifically for this update.
 
+.. _using_custom_lifecycle:
+
+
+Tracking embedded entities when using a custom lifecycle
+********************************************************
+
```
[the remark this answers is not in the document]
wouterdb: Added on my version

docs/lsm/embedded_entities/embedded_entities.rst
wouterdb: Added on my version

docs/lsm/embedded_entities/embedded_entities.rst
Hugo-Inmanta: Added a note regarding `strict_modifier_enforcement`

wouterdb (approved):

PR #8153, opened 2024-10-03
Fix test cases related to logging

arnaudsjs:
# Description

This PR fixes the logging-related test cases that broke due to https://github.com/inmanta/inmanta-core/commit/fc81b36711235eb10bf9aa0a4764b8422d42d34a.

**Root cause of failures:**

The above-mentioned commit made sure that the root log level is not changed by the `apply_config()` method. This broke the logging-related test cases because they assume that they can change the root log level.

**Changes made by this PR:**

* Add `allow_overriding_root_log_level` fixture that makes sure that the `apply_config` can change the root log level.
* Simplify the `InmantaLoggerConfig.clean_instance()` method. In the previous implementation it made a call to `logging.shutdown()`, which closes all the handlers. This PR re-adds that method invocation to simplify the method. This is not an issue for pytest, because pytest re-adds their handlers after each test case.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/logging.py
+++ b/src/inmanta/logging.py
@@ -585,13 +585,7 @@ class InmantaLoggerConfig:
         By default, this method removes and closes all root handlers from the logging framework. If the
         root_handlers_to_remove argument is not None, only the provided root handlers will be removed and closed.
         """
-        to_remove = root_handlers_to_remove if root_handlers_to_remove is not None else logging.root.handlers
-        for handler in to_remove:
-            # File-based handlers automatically re-open after close() if log records are written to them.
-            # As such, we explicitly remove the handler from the root logger here.
-            logging.root.removeHandler(handler)
-            handler.flush()
-            handler.close()
+        logging.shutdown()
         cls._instance = None
 
     def _get_path_to_logging_config_file(self, options: Options) -> Optional[str]:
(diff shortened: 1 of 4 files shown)


sanderr (approved):

src/inmanta/logging.py
```diff
@@ -585,13 +585,7 @@ def clean_instance(cls, root_handlers_to_remove: Optional[abc.Sequence[logging.H
         By default, this method removes and closes all root handlers from the logging framework. If the
         root_handlers_to_remove argument is not None, only the provided root handlers will be removed and closed.
         """
-        to_remove = root_handlers_to_remove if root_handlers_to_remove is not None else logging.root.handlers
-        for handler in to_remove:
-            # File-based handlers automatically re-open after close() if log records are written to them.
-            # As such, we explicitly remove the handler from the root logger here.
-            logging.root.removeHandler(handler)
-            handler.flush()
-            handler.close()
```
Hugo-Inmanta, line 594: Shouldn't we remove the `root_handlers_to_remove` from the signature + docstring?

Hugo-Inmanta (changes requested):

src/inmanta/logging.py
arnaudsjs: I have been considering that, but this method is annotated with `@stable_api`, so we can't if we want to keep it backwards compatible. Although, I think this parameter is not used outside of inmanta-core.

PR #8156, opened 2024-10-04
Add hard timeout with tox

hlloreda:
# Description

Added hard timeout with tox to prevent Jenkins from killing the job without being able to retrieve some data

# Self Check:

- [ ] Changelog entry


changelogs/unreleased/hard_timeout_tox.yml
```diff
@@ -0,0 +1,4 @@
+description: Add hard timeout in tox to retrieve some data, in case of stuck test suite, before being killed by Jenkins
```
wouterdb, line 1:
```suggestion
description: Add soft timeout in tox to retrieve some data, in case of stuck test suite, before being killed by Jenkins
```

wouterdb (approved):

PR #8165, opened 2024-10-07
Handle undefined resources in the scheduler

arnaudsjs:
# Description

Make sure that the scheduler handles undefined resources properly.

# Self Check:

- [x] Changelog entry


tests/agent_server/deploy/test_scheduler_agent.py
```diff
@@ -1111,19 +1113,153 @@ async def test_get_facts(agent: TestAgent, make_resource_minimal):
...
+    assert len(agent.scheduler.get_types_for_agent("agent1")) == 1
+
+    assert_resource_state(rid8, state.ResourceStatus.UNDEFINED, state.DeploymentResult.NEW, state.BlockedStatus.YES)
+    assert_resource_state(rid9, state.ResourceStatus.UNDEFINED, state.DeploymentResult.NEW, state.BlockedStatus.YES)
+
+    # rid8 is no longer undefined
+    resources = {
+        rid8: make_resource_minimal(rid=rid8, values={"value": "a"}, requires=[], status=const.ResourceState.available),
+        rid9: make_resource_minimal(rid=rid9, values={"value": "a"}, requires=[rid8], status=const.ResourceState.undefined),
+    }
```
hlloreda, line 1258:
```suggestion
    resources[rid8] = make_resource_minimal(rid=rid8, values={"value": "a"}, requires=[], status=const.ResourceState.available)

```

hlloreda (changes requested):

src/inmanta/deploy/scheduler.py
```diff
@@ -263,13 +264,26 @@ async def _new_version(
...
+            # Resources that were undeployable in a previous model version, but got unblocked. Not the transitive ones.
+            unblocked_resources: set[ResourceIdStr] = set()
             added_requires: dict[ResourceIdStr, Set[ResourceIdStr]] = {}
             dropped_requires: dict[ResourceIdStr, Set[ResourceIdStr]] = {}
             for resource, details in resources.items():
-                if (
-                    resource not in self._state.resources
-                    or details.attribute_hash != self._state.resources[resource].attribute_hash
-                ):
+                if details.status is const.ResourceState.undefined:
```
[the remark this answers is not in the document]
arnaudsjs: I see your point. It's rather confusing. The two types represent different things, but they have an almost identical name. Sticking with one will not be easy. But given that `ResourceStatus`  was introduced recently, I think we can still quite easily rename it. Maybe we should call it `ResourceOpStatus`, to indicate this represents the operational state of the resource. Do you have something else in mind? @sanderr I think you introduced the `ResourceStatus`. Do you have an opinion on this?

src/inmanta/deploy/scheduler.py
```diff
@@ -283,19 +297,30 @@ async def _new_version(
                 # => regularly pass control to the event loop to not block scheduler operation during update prep
                 await asyncio.sleep(0)
 
+            assert len(new_desired_state | blocked_resources | unblocked_resources) == len(new_desired_state) + len(
+                blocked_resources
+            ) + len(unblocked_resources)
```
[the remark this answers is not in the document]
arnaudsjs: Yes. That's correct. I'll add a comment.

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
sanderr: I have to go through the full pull request before I can give a definite answer, but for now I'll say this:
1. I wouldn't expect `const.ResourceStatus` to be used here, because that is the legacy status that we're trying to replace here.
2. I'm open to a name change, and perhaps we can even drop the resource part and go for simply `OperationalStatus` but I'm not sure we can make that decision here. I would involve the people that were present for the [improved resource states design session](https://docs.google.com/presentation/d/1F3bFNy2BZtzZgAxQ3Vbvdw7BWI9dq0ty5c3EoLAtUUY/edit#slide=id.g292b508a90d_0_5). Iirc that's Bart, Wouter and Guillaume. I think this PR is the final piece in the resource state that we want to include for iso8. Perhaps, once review has settled, we can have a short session where we propose the subset of the improved resource states that we want to expose, with the names we intend to use, and only then make resulting changes?

src/inmanta/deploy/state.py
```diff
@@ -36,15 +38,45 @@ def requires_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
     def provides_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
         return self.reverse_mapping()
 
+    def get_all_provides_transitively(self, resource: ResourceIdStr | set[ResourceIdStr]) -> list[ResourceIdStr]:
```
sanderr, line 41:
```suggestion
    def get_all_provides_transitively(self, resource: ResourceIdStr | Set[ResourceIdStr]) -> list[ResourceIdStr]:
```

src/inmanta/deploy/state.py
```diff
@@ -36,15 +38,45 @@ def requires_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
...
+        This method returns all the provides (transitively) of the given resource.
+
+        :param resource: The resource or set of resources for which the provides have to be resolved transitively.
+        """
+        input_set = {resource} if isinstance(resource, str) else set(resource)
+        work: SimpleQueue[ResourceIdStr] = SimpleQueue()
+        for elem in input_set:
+            work.put_nowait(elem)
+        provides_mapping = self.provides_view()
+        # Use a dict here to not lost the order of the elements.
```
sanderr, line 52:
```suggestion
        # Use a dict here to not lose the order of the elements.
```

src/inmanta/deploy/state.py
```diff
@@ -36,15 +38,45 @@ def requires_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
...
         return self.reverse_mapping()
 
+    def get_all_provides_transitively(self, resource: ResourceIdStr | set[ResourceIdStr]) -> list[ResourceIdStr]:
+        """
+        This method returns all the provides (transitively) of the given resource.
+
+        :param resource: The resource or set of resources for which the provides have to be resolved transitively.
+        """
+        input_set = {resource} if isinstance(resource, str) else set(resource)
+        work: SimpleQueue[ResourceIdStr] = SimpleQueue()
```
sanderr, line 48: Do you know anything of the performance of this vs a simple list? I'm assuming this is better because it has efficient pops and appends. But it might also have more bookkeeping logic.

I ask because everything executed under the lock is somewhat performance-critical.

src/inmanta/deploy/state.py
```diff
@@ -36,15 +38,45 @@ def requires_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
...
+            work.put_nowait(elem)
+        provides_mapping = self.provides_view()
+        # Use a dict here to not lost the order of the elements.
+        result: dict[ResourceIdStr, None] = {}
+        while not work.empty():
+            current_resource = work.get()
+            result[current_resource] = None
+            provides = provides_mapping.get(current_resource, set())
+            for elem in provides:
+                work.put_nowait(elem)
```
sanderr, line 59: Only if it's not yet in `result`, right? We might even need a set to keep track of the union of result and work if we don't want to queue the same element a bunch of times.

src/inmanta/deploy/state.py
```diff
@@ -36,15 +38,45 @@ def requires_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
...
+        for elem in input_set:
+            result.pop(elem, None)
```
sanderr, line 61: Could you add to the docstring that the result is exclusive of the input, even if multiple resources are provided, where a provides edge exists between them?

src/inmanta/deploy/state.py
```diff
@@ -36,15 +38,45 @@ def requires_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
...
+        return list(result.keys())
+
 
 @dataclass(frozen=True)
 class ResourceDetails:
     resource_id: ResourceIdStr
     attribute_hash: str
     attributes: Mapping[str, object] = dataclasses.field(hash=False)
+    status: const.ResourceState
```
sanderr, line 70: So, as mentioned in the other thread: I do not think this belongs here. This is the old-style resource state we're trying to replace. That said, we can consider to add it now and to drop it when we make the scheduler the authority for resource state (#8187?). If we decide to take that approach, we should add a FIXME here.

Overall, I think I would prefer to drop it now already, and simply populate the `ResourceStatus` when we read from the database.

@wouterdb WDYT?

src/inmanta/deploy/state.py
sanderr, line 1: I think it would be useful to add @wouterdb as a reviewer at some point, only for the resource states part of this PR, because he was heavily involved in that.

src/inmanta/util/collections.py
```diff
@@ -50,7 +50,7 @@ def _set(primary: dict[K, set[V]], reverse: dict[V, set[K]], key: K, values: Set
         Set the key-value pair on the primary dict and update the reverse mapping to uphold the bidirectionality.
 
         :param primary: The dict to consider as primary for this set operation, i.e. the one to set the key-value pair on.
-        :param primary: The dict to consider as secondary for this set operation, i.e. the one to apply the reverse key-value
+        :param reverse: The dict to consider as secondary for this set operation, i.e. the one to apply the reverse key-value
```
sanderr, line 53: Oops, thanks.

src/inmanta/deploy/state.py
```diff
@@ -125,6 +181,73 @@ def reset(self) -> None:
         self.resource_state.clear()
         self.types_per_agent.clear()
 
+    def block_resource(self, resource: ResourceIdStr, details: ResourceDetails, transient: bool) -> None:
+        """
+        Mark the given resource as blocked, i.e. it's not deployable.
```
sanderr, line 186: This method also updates the resource details / intent. I think that should be mentioned.

src/inmanta/deploy/state.py
```diff
@@ -125,6 +181,73 @@ def reset(self) -> None:
...
+
+        _unblock(resource)
+        for res in self.requires.get_all_provides_transitively(resource):
+            is_blocked = self.resource_state[res].status is ResourceStatus.UNDEFINED or any(
+                self.resource_state[r].blocked.is_blocked() for r in self.requires.get(res, set())
+            )
+            if not is_blocked:
+                _unblock(res)
+
     def update_desired_state(
```
sanderr, line 251: Do I understand correctly a caller would call either this method or `block_resource` depending on whether it's blocked or not. If so, could you add to the docstring that `block_resource` should be called instead for blocked resources?

src/inmanta/deploy/state.py
```diff
@@ -125,6 +181,73 @@ def reset(self) -> None:
...
+        :param resource: The resource that should be blocked.
+        :param details: The details of the resource that should be blocked.
+        :param transient: True iff the given resource is blocked transitively. It's blocked because one of its dependencies
+                          is blocked.
+        """
+        self.resources[resource] = details
+        blocked_status = BlockedStatus.TRANSIENT if transient else BlockedStatus.YES
+        if resource in self.resource_state:
+            if not transient:
+                self.resource_state[resource].status = ResourceStatus.UNDEFINED
```
sanderr, line 197: Note to self: TODO: should this also set `HAS_UPDATE`? What about the dirty set?

sanderr (changes requested):
Releasing intermediate comments because I have a meeting now.

Not reviewed so far:
- `state.block_resource` and the methods below
- `scheduler`
- tests

src/inmanta/deploy/scheduler.py
```diff
@@ -307,7 +333,7 @@ async def _new_version(
                     self._state.drop(resource)
             # Once more, drop all resources that do not exist in this version from the scheduled work, in case they got added
             # again by a deploy trigger (because we dropped them outside the lock).
-            for resource in deleted_resources:
+            for resource in deleted_resources | blocked_resources | transitively_blocked_resources:
                 self._work.delete_resource(resource)
```
sanderr, line 337: As implemented now, I don't think blocked resources (especially undefined ones) are safe with respect to races.

The reason we call this method outside of the lock, is that for `deleted_resources`, we don't care too much if we lose the race. The worst that can happen is that it gets deployed once too many. However, if we apply that logic to the blocked resources, its deploy task would choke on the unknowns, wouldn't it? So I think that we have two options:

- make sure that the task can handle the race
- move this call within the lock, for the blocked resources

src/inmanta/deploy/scheduler.py
```diff
@@ -283,19 +297,31 @@ async def _new_version(
...
                 self._state.version = version
+                for resource in blocked_resources:
+                    self._state.block_resource(resource, details, transient=False)
                 for resource in new_desired_state:
                     self._state.update_desired_state(resource, resources[resource])
                 for resource in added_requires.keys() | dropped_requires.keys():
                     self._state.update_requires(resource, requires[resource])
+                transitively_blocked_resources: set[ResourceIdStr] = self._state.block_provides(resources=blocked_resources)
+                for resource in unblocked_resources:
+                    self._state.unblock_resource(resource)
```
sanderr, line 320: What if we just marked it as transitively blocked? We'd unmark it here but that's not valid, is it?

Also potentially affects the `difference_update two lies down.

src/inmanta/deploy/state.py
```diff
@@ -36,15 +38,45 @@ def requires_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
...
 @dataclass(frozen=True)
 class ResourceDetails:
     resource_id: ResourceIdStr
     attribute_hash: str
     attributes: Mapping[str, object] = dataclasses.field(hash=False)
+    status: const.ResourceState
 
     id: Id = dataclasses.field(init=False, compare=False, hash=False)
 
+    def is_undeployable(self) -> bool:
```
sanderr, line 74: I don't think this method is used in the current implementation.

src/inmanta/deploy/state.py
sanderr: Concretely, I think that the difference between old resource states `undefined` and `skipped_for_undefined` are the following in the new resources states:
- `undefined`: `status=UNDEFINED, blocked=YES`
- `skipped_for_undefined`: `status=UP_TO_DATE/HAS_UPDATE, blocked=YES`

sanderr (dismissed):

src/inmanta/deploy/state.py
```diff
@@ -81,6 +115,26 @@ class DeploymentResult(StrEnum):
     FAILED = enum.auto()
 
 
+class BlockedStatus(StrEnum):
+    """
+    YES: The resource will retain its blocked status within this model version. For example: A resource that has unknowns
+         or depends on a resource with unknowns.
+    TRANSIENT: The resource is blocked but the blocked status may get resolved with the model version.
+               For example: The dependency of the resource failed to deploy.
+    NO: The resource is not blocked
```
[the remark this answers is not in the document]
arnaudsjs: You are right. Somehow I misinterpreted the slides, while they are actually rather clear on what the TRANSIENT blocked state means. I'll fix this.

src/inmanta/deploy/state.py
arnaudsjs: I tested the performance and it seems that pop and push execute in O(1).

src/inmanta/deploy/state.py
```diff
@@ -125,6 +181,73 @@ def reset(self) -> None:
...
+            self.resource_state[resource].blocked = BlockedStatus.NO
+            if self.resource_state[resource].status is ResourceStatus.UNDEFINED:
+                self.resource_state[resource].status = ResourceStatus.HAS_UPDATE
+            if self.resource_state[resource].status is ResourceStatus.HAS_UPDATE:
+                self.dirty.add(resource)
+
+        _unblock(resource)
+        for res in self.requires.get_all_provides_transitively(resource):
+            is_blocked = self.resource_state[res].status is ResourceStatus.UNDEFINED or any(
+                self.resource_state[r].blocked.is_blocked() for r in self.requires.get(res, set())
```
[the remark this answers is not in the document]
arnaudsjs: The `get_all_provides_transitively()` method returns the element sorted. That's why it's a list instead of a set. But I didn't document this. I added the documentation now.

src/inmanta/deploy/scheduler.py
arnaudsjs: I moved it under the lock.

src/inmanta/deploy/scheduler.py
```diff
@@ -263,13 +264,26 @@ async def _new_version(
...
             added_requires: dict[ResourceIdStr, Set[ResourceIdStr]] = {}
             dropped_requires: dict[ResourceIdStr, Set[ResourceIdStr]] = {}
             for resource, details in resources.items():
-                if (
-                    resource not in self._state.resources
-                    or details.attribute_hash != self._state.resources[resource].attribute_hash
-                ):
+                if details.status is const.ResourceState.undefined:
+                    blocked_resources.add(resource)
+                    self._work.delete_resource(resource)
```
hlloreda, line 276: This is not necessary, right? As we do this a bit below:
```
                for resource in blocked_resources | transitively_blocked_resources:
                    self._work.delete_resource(resource)
```

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
wouterdb: For now, I expect both to be used (one towards the DB and one towards the scheduler). Once the DB is more evolved, the `const.ResourceState.undefined` will go away 

We could raw the line a bit higher up on the stack, but this fine for me as well

src/inmanta/deploy/scheduler.py
```diff
@@ -283,19 +297,31 @@ async def _new_version(
                 # => regularly pass control to the event loop to not block scheduler operation during update prep
                 await asyncio.sleep(0)
 
+            # A resource should not be present in more than one of these resource sets
+            assert len(new_desired_state | blocked_resources | unblocked_resources) == len(new_desired_state) + len(
+                blocked_resources
+            ) + len(unblocked_resources)
```
wouterdb, line 303: this can get expensive, but I would leave it in for now. Perhaps add a comment that this can become expensive.

src/inmanta/deploy/state.py
wouterdb: I'm going to remember this

src/inmanta/deploy/state.py
wouterdb: We handle it now or later, no strong preference.

src/inmanta/deploy/state.py
```diff
@@ -125,20 +184,89 @@ def reset(self) -> None:
...
+        Marks the provides of the given resources as blocked.
+
+        Must be called under the scheduler lock. This method assumes that all the resources of the model version
+        are populated in the resources dictionary.
+
+        :param resources: The set of resources for which the provides have to be blocked.
+        :return: The set of dependent resources that were marked as blocked (transitively).
+        """
+        result = set()
+        for dependent_resource in self.requires.get_all_provides_transitively(resources):
```
wouterdb, line 221: if may become relevant in the future to only traverse up the provides relation until we find one that is already blocked, but this will do for now I think.

wouterdb (approved):

src/inmanta/deploy/scheduler.py
arnaudsjs: I am not sure I understand what you mean. Here we delete the blocked resources (and the transitively blocked resources). This needs to happen under the scheduler lock, otherwise we risk these resources may get deployed. This is undesired. A bit below we delete the deleted resources. We do this outside the scheduler lock for improved performance, i.e. we don't want to keep the lock acquired longer than strictly required. If a deleted resource gets deployed one time too many, it's not a big deal.

src/inmanta/deploy/state.py
```diff
@@ -38,24 +38,36 @@ def requires_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
     def provides_view(self) -> Mapping[ResourceIdStr, Set[ResourceIdStr]]:
         return self.reverse_mapping()
 
-    def get_all_provides_transitively(self, resource: ResourceIdStr | set[ResourceIdStr]) -> list[ResourceIdStr]:
+    def get_all_provides_transitively(self, resource: ResourceIdStr | Set[ResourceIdStr]) -> list[ResourceIdStr]:
         """
-        This method returns all the provides (transitively) of the given resource.
+        This method returns all the provides (transitively) of the given resource. The returned list will be sorted as such
```
sanderr, line 43: What is the motivation for sorting them? The current approach for sorting them seems like it may do a lot of double (or triple, quadruple, ...) work if the provides tree is heavily non-treelike.

src/inmanta/deploy/state.py
```diff
@@ -183,29 +186,28 @@ def reset(self) -> None:
...
         self.resources[resource] = details
-        blocked_status = BlockedStatus.TRANSIENT if transient else BlockedStatus.YES
         if resource in self.resource_state:
             if not transient:
                 self.resource_state[resource].status = ResourceStatus.UNDEFINED
-            if self.resource_state[resource].blocked is not BlockedStatus.YES:
-                self.resource_state[resource].blocked = blocked_status
+            self.resource_state[resource].blocked = BlockedStatus.YES
         else:
             resource_status = ResourceStatus.UNDEFINED if not transient else ResourceStatus.HAS_UPDATE
```
sanderr, line 203: I missed this in the previous review, but shouldn't this be `ResourceStatus.NEW` instead of `HAS_UPDATE`?

src/inmanta/deploy/state.py
sanderr: Added to #8187

src/inmanta/deploy/state.py
sanderr: > What is the motivation for sorting them?

I see now that if we don't, it would complicate the `unblock_resources` logic because we would need to traverse the requires relation more deeply. i.e. what I said there about checking "if `r` is in `self.requires.get_all_provides_transitively(resource)`" doesn't suffice because `r` might have other blocked requires outside of that provides forest.

sanderr: With this in mind, I wonder if it might not be simpler to not use the method and simply traverse the relation directly, unblocking resources along the way and stopping recursion when either a resource is already blocked (will have already informed its provides) or when one of its requires is still blocked (this one won't be unblocked and neither will its children, unless we reach this node again).

Issue #7898, opened 2024-07-24
paging behind last page returns 0 objects "before"


sanderr:
If the filtering results in 0 objects, the API returns `{"data": [], "metadata": {"total": 1, "before": 0, "after": 0, "page_size": 20}` even if there are matching resources before it. The opposite is probably true for pages before the first one. This results in issues like inmanta/web-console#5894. Fixing this would prevent the need to reset the paging when a filter is added, as is the proposed solution in the linked ticket.

I believe that the reason for this is that we typically use the result set to determine the first and last element on the current page, and then use those as boundary elements to count the before and after. But I think it should be simple enough to handle this case: iff the result set is empty, simply drop all paging filters for the before count (in case of ASC) or the after count (in case of DESC). i.e. if the current page is empty, **all** other resources are before / after it.

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/7898?planning-poker) with Zenhub @arnaudsjs @hlloreda @Hugo-Inmanta @jptrindade

hlloreda:
https://github.com/inmanta/inmanta-core/pull/8172 will not completely address this issue because we cannot decide, in the current implementation, if the boundaries are exclusive or inclusive. They're inclusive at the moment, but that causes issues, as shown below:

![image](https://github.com/user-attachments/assets/c825fb76-7330-4264-8018-0e9c867a130e)
 
To fix this, the idea would be to update the `DatabaseOrder.as_filter` method:
- add an `inclusive` boolean (by default set to `False`)

The links remain a difficulty though because they would require a like-named argument on all the paging methods. There are a lot of them, and not all of them in core.
```
(inmanta-core-312) sander@bedevere:~/documents/projects/inmanta/inmanta-core$ grep 'def as_filter' -R src/inmanta
src/inmanta/data/__init__.py:    def as_filter(
src/inmanta/data/__init__.py:    def as_filter(
src/inmanta/data/__init__.py:    def as_filter(
(inmanta-core-312) sander@bedevere:~/documents/projects/inmanta/inmanta-core$ grep 'def as_filter' -R ../inmanta-lsm/src/inmanta_lsm
(inmanta-core-312) sander@bedevere:~/documents/projects/inmanta/inmanta-core$ grep 'as_filter' -R ../inmanta-lsm/src/inmanta_lsm
[...]

PR #8172, opened 2024-10-08
Fix pagination when returning 0 objects

hlloreda:
# Description

Fixed metadata + link when returning 0 objects

Related to https://github.com/inmanta/inmanta-core/issues/7898

# Self Check:

- [ ] Changelog entry


src/inmanta/data/dataview.py
```diff
@@ -139,6 +155,21 @@ def has_start(self) -> bool:
...
 
+    def get_empty_page_boundaries(self) -> PagingBoundaries:
+        """
+        Return the virtual paging boundaries that corresponds to the boundaries of an empty page response for this request.
+
+        Concretely, this means swapping start and end, because the semantics of PagingBoundaries are opposite those of this
+        class: PagingBoundaries represents "current page" while this class represents "requested page";
+        max of current page = min of requested page and vice versa.
+        """
+        return PagingBoundaries(
```
sanderr, line 166: Perhaps we should add a comment here noting that there is a potential off-by-one error here because both classes are exclusive of the boundary values, and add a reference to the remaining ticket?

tests/server/test_agents_api.py
```diff
@@ -216,6 +216,22 @@ async def test_agents_paging(server, client, env_with_agents: None, environment:
 
     assert result.result["metadata"] == {"total": 7, "before": 0, "after": 0, "page_size": 100}
 
+    fake_id_url = f"{base_url}/api/v2/agents?limit=1&sort=last_failover.asc&filter.status=paused&first_id=zzzz"
```
sanderr, line 219:
```suggestion
    # verify that paging beyond the last page returns correct counts and links
    fake_id_url = f"{base_url}/api/v2/agents?limit=1&sort=last_failover.asc&filter.status=paused&first_id=zzzz"
```

tests/server/test_agents_api.py
```diff
@@ -216,6 +216,22 @@ async def test_agents_paging(server, client, env_with_agents: None, environment:
...
+    )
+    response = await http_client.fetch(request, raise_error=False)
+    assert response.code == 200
+    response = json.loads(response.body.decode("utf-8"))
+    # We don't have a way to reconstruct the previous link
+    assert response["links"] == {
+        "first": "/api/v2/agents?limit=1&sort=last_failover.asc&filter.status=paused",
+        "prev": "/api/v2/agents?limit=1&sort=last_failover.asc&filter.status=paused&last_id=zzzz",
+    }
+    assert response["metadata"] == {"total": 2, "before": 2, "after": 0, "page_size": 1}
```
sanderr, line 232: What does `page_size` represent? Should it be 1?

tests/server/test_agents_api.py
```diff
@@ -216,6 +216,22 @@ async def test_agents_paging(server, client, env_with_agents: None, environment:
...
 
+    fake_id_url = f"{base_url}/api/v2/agents?limit=1&sort=last_failover.asc&filter.status=paused&first_id=zzzz"
+    request = HTTPRequest(
+        url=fake_id_url,
+        headers={"X-Inmanta-tid": str(environment)},
+    )
+    response = await http_client.fetch(request, raise_error=False)
+    assert response.code == 200
+    response = json.loads(response.body.decode("utf-8"))
+    # We don't have a way to reconstruct the previous link
```
sanderr, line 227: How do you mean? We do get a valid prev link, don't we?

tests/server/test_resource_list.py
```diff
@@ -449,6 +450,43 @@ async def test_resources_paging(server, client, order_by_column, order, env_with
     assert response["metadata"] == {"total": 5, "before": 2, "after": 1, "page_size": 2}
 
 
+async def test_none_resources_paging(server, client, env_with_resources):
+    """Test querying resources with paging, using different sorting parameters."""
```
sanderr, line 454: This docstring doesn't seem to apply to what's being tested.

tests/server/test_resource_list.py
```diff
@@ -449,6 +450,43 @@ async def test_resources_paging(server, client, order_by_column, order, env_with
...
+    env = env_with_resources
+    order_by_column = "agent"
+    order = "DESC"
+
+    no_result = await client.resource_list(
+        env.id,
+        limit=2,
+        sort=f"{order_by_column}.{order}",
+        start="ausf",
+        first_id="anet_co%3A%3Aresource%3A%3AResource%5BINMANTA-INMANTA-INMT%2Cresource_id%3Dinmt%5D",
```
sanderr, line 464: Apart from the confidentiality thing, which has been resolved, is there any reason we use so complex (difficult to read) resource here, rather than something simple and obviously meaningless like you have in the other test case with "zzz"?

tests/server/test_resource_list.py
sanderr, line 1: Can you add some scenarios for the other sorting orders: DESC-prev, ASC-next, ASC-prev?

sanderr (changes requested):

src/inmanta/data/model.py
```diff
@@ -459,10 +459,22 @@ class PagingBoundaries:
...
+    neighbouring pages. Which represents next and which prev depends on sorting order (ASC or DESC).
+
+    Boundary values represent max and min values, regardless of sorting direction (ASC or DESC), i.e.
+    - ASC sorting based on next links: (>) end,last_id
+    - ASC sorting based on prev links: (<) start,first_id
+    - DESC sorting based on next links: (<) start,first_id
+    - DESC sorting based on prev links: (>) end,last_id
+
+    So, while the names "start" and "end" might seem to indicate "left" and "right" of the page, they actually mean "highest"
+    and "lowest".
```
Hugo-Inmanta, line 472: While trying to understand this I added a tangible example with 2 columns (int, letter) that shows which param is taken into consideration on which page depending on the sorting order. It also shows the inclusive/exclusive ie what is on which page. Is this accurate ? (Feel free to ignore)
```suggestion
    and "lowest".
    
                                                                                                                                     
                  |        prev             |                current page                    |    next                           
     -------------|-------------------------|------------------------------------------------|----------------------             
     ASC sorting  | [  ...  ] (100 c)       | [ (100 c) ( 100 d) ... (10 000 r) (10 000 s) ] | (10 000 s) [ ...   ]              
                  |            start = 100  |     end = 100            start = 10 000        |     end = 10 000                  
                  |            first_id = c |     last_id = c          first_id = s          |     last_id = s                   
                  |                                                                                                              
     DESC sorting | [  ...  ] (10 000 s)    | [ (10 000 s) (10 000 r) ... (100 d) ( 100 c) ] | (100 c) [ ...   ]                 
                  |            end = 10 000 |          start = 10000          end = 100      |     start = 100                   
                  |            last_id = s  |          first_id = s           last_id = c    |     first_id = c                  
                                                                                                                                 
```

Hugo-Inmanta (changes requested):

src/inmanta/data/model.py
hlloreda: I adjusted some values, but it's way cleaner, thank you!

tests/server/test_agents_api.py
hlloreda: It's the number of objects that we asked for. I rewrote that part to make it clearer

src/inmanta/data/dataview.py
```diff
@@ -96,7 +96,28 @@
...
+    Boundary values represent min and max values, regardless of sorting direction (ASC or DESC), i.e.
+                  |        prev             |                current page                    |    next
+     -------------|-------------------------|------------------------------------------------|----------------------
+     ASC sorting  | [  ...  ] (99 c)        | [ (100 d) ( 100 e) ... (10 000 r) (10 000 s) ] | (10 001 t) [ ...   ]
+                  |            end = 99     |     start = 100          end = 10 000          |     start = 10 001
+                  |            last_id = c  |     first_id = d         last_id = s           |     last_id = t
+                  |
+     DESC sorting | [  ...  ] (9 999 s)     | [ (10 000 t) (10 000 r) ... (100 d) ( 100 c) ] |  (99 b) [ ...   ]
+                  |          start = 9 999  |         end = 10000          start = 100       |     end = 99
+                  |          first_id = s   |         last_id = t          first_id = c      |     last_id = b
```
Hugo-Inmanta, line 111: I think ?
```suggestion
     DESC sorting | [  ...  ] (10 001 t)    | [ (10 000 s) (10 000 r) ... (100 d) ( 100 c) ] |  (99 b) [ ...   ]
                  |          start = 10 001 |         end = 10000          start = 100       |     end = 99
                  |          first_id = t   |         last_id = s          first_id = c      |     last_id = b
```

Hugo-Inmanta (changes requested):

tests/server/test_resource_list.py
```diff
@@ -449,6 +450,136 @@ async def test_resources_paging(server, client, order_by_column, order, env_with
...
+    assert actual_result_desc_prev.result["links"] == expected_result_desc_prev
+    assert actual_result_desc_prev.result["metadata"] == {"after": 0, "before": 4, "page_size": 2, "total": 6}
+    assert len(actual_result_desc_prev.result["data"]) == 2
+
+    no_result_desc_next = await client.resource_list(
+        env.id,
+        limit=2,
+        sort="agent.DESC",
+        start="ausf",
+        first_id="anet_co",
```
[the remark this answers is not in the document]
hlloreda: We cannot do this because we want to have a page boundary without a next in with a descending order.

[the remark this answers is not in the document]
Hugo-Inmanta: >  without a next

Without a prev you mean right?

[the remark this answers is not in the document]
Hugo-Inmanta: I would expect in desc order `start="zzzz", first_id="zz"` to have the bulk of the data to the right and nothing as prev

src/inmanta/data/dataview.py
```diff
@@ -96,7 +96,23 @@
...
 class RequestedPagingBoundaries:
-    """Represents the lower and upper bounds that the user requested for the paging boundaries, if any."""
+    """
+    Represents the lower and upper bounds that the user requested for the paging boundaries, if any.
+
+    Boundary values represent max and min values, regardless of sorting direction (ASC or DESC), i.e.
+    - ASC sorting based on next links: (>) start,first_id
+    - ASC sorting based on prev links: (<) end,last_id
+    - DESC sorting based on next links: (<) end,last_id
+    - DESC sorting based on prev links: (>) start,first_id
```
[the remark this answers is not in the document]
sanderr: My bad. gt, lt, lt, gt. I found it difficult to put in words. It's supposed to clarify the direction of the bounds, e.g. the first one means a page is requested for which all elements are `> (start, first_id)`.

changelogs/unreleased/7898-fix-pagination.yml
```diff
@@ -0,0 +1,6 @@
+description: "Fixed pagination when on a page returning 0 results (due to pagination criteria) that is strictly after or strictly before the bulk of the actual results"
```
hlloreda, line 1: behind the last result  / before first result

src/inmanta/data/dataview.py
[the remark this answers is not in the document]
hlloreda: Add this to the docstring

src/inmanta/data/dataview.py
```diff
@@ -400,48 +434,45 @@ async def prepare_paging_links(
...
+        if meta.before > 0:
+            if self.order.get_order() == "DESC":
+                links["prev"] = link_with_start
+            else:
+                links["prev"] = link_with_end
+            # First page
+            links["first"] = make_link()
+
+        # Same page
+        if dtos:
```
hlloreda, line 475: api could assume self entry even if page empty

src/inmanta/data/dataview.py
```diff
@@ -98,7 +98,28 @@
...
+    Boundary values represent min and max values, regardless of sorting direction (ASC or DESC), i.e.
+                  |        prev             |                current page                    |    next
+     -------------|-------------------------|------------------------------------------------|----------------------
+     ASC sorting  | [  ...  ] (99 c)        | [ (100 d) ( 100 e) ... (10 000 r) (10 000 s) ] | (10 001 t) [ ...   ]
+                  |            end = 99     |     start = 100          end = 10 000          |     start = 10 001
+                  |            last_id = c  |     first_id = d         last_id = s           |     first_id = t
+                  |
+     DESC sorting | [  ...  ] (10 001 t)    | [ (10 000 s) (10 000 r) ... (100 d) ( 100 c) ] |  (99 b) [ ...   ]
+                  |          start = 10 001 |         end = 10000          start = 100       |     end = 99
+                  |          first_id = t   |         last_id = s          first_id = c      |     last_id = b
```
Hugo-Inmanta, line 113: Same here, I'm not so sure of what we're conveying here, maybe adds more confusion than anything

src/inmanta/data/model.py
```diff
@@ -471,10 +471,27 @@ class PagingBoundaries:
...
+    Boundary values represent max and min values, regardless of sorting direction (ASC or DESC), i.e.
+                  |        prev             |                current page                    |    next
+     -------------|-------------------------|------------------------------------------------|----------------------
+     ASC sorting  | [  ...  ] (99 c)        | [ (100 d) ( 100 e) ... (10 000 r) (10 000 s) ] | (10 001 t) [ ...   ]
+                  |            start = 99   |     end = 100            start = 10 000        |     end = 10 001
+                  |            first_id = c |    last_id = d          first_id = s           |     first_id = t
+                  |
+     DESC sorting | [  ...  ] (10 001 s)    | [ (10 000 t) (10 000 r) ... (100 d) ( 100 c) ] |  (99 b) [ ...   ]
+                  |            end = 10 001 |          start = 10000          end = 100      |     start = 99
+                  |            last_id = t  |          first_id = t           last_id = c    |     first_id = b
```
Hugo-Inmanta, line 486: I'm not sure I understand what's conveyed here anymore. We should remove it if it adds more confusion than anything.
Alternatively we could maybe add a single scenario where we describe what happens in the simple case eg:

User requests all Resources with name  >= foo in ASC order. page size = 10
```
RequestPagingBoundary:
    start = foo
    end = None
```
    
Actual data = [foo1 ... foo10]

```
Pagingboundary:
    end = foo1
    start = foo10    # Reversed because these are meant to be used by the neighbouring pages
```
    
Does this make sense / is actually correct ?

hlloreda: Yes, well explained, thank you!

Hugo-Inmanta (approved):

src/inmanta/data/dataview.py
```diff
@@ -141,6 +157,24 @@ def has_start(self) -> bool:
     def has_end(self) -> bool:
         return (self.end is not None) or (self.last_id is not None)
 
+    def get_empty_page_boundaries(self) -> PagingBoundaries:
+        """
+        Return the virtual paging boundaries that corresponds to the boundaries of an empty page response for this request.
+
+        Concretely, this means swapping start and end, because the semantics of PagingBoundaries are opposite those of this
+        class: PagingBoundaries represents "current page" while this class represents "requested page";
+        max of current page = min of requested page and vice versa.
```
sanderr, line 166:
```suggestion
        class: PagingBoundaries' boundary values map to RequestedPagingBoundaries` boundary values for the next and previous
        pages, e.g. requested max = requested min of next page and vice versa.
```

src/inmanta/data/model.py
```diff
@@ -473,16 +473,30 @@ class PagingBoundaries:
...
-    - ASC sorting based on next links: (>) end,last_id
-    - ASC sorting based on prev links: (<) start,first_id
-    - DESC sorting based on next links: (<) start,first_id
-    - DESC sorting based on prev links: (>) end,last_id
-
     So, while the names "start" and "end" might seem to indicate "left" and "right" of the page, they actually mean "highest"
     and "lowest".
 
+    Let's show this in an example: a user requests the following:
+     - all Resources with name >= foo
```
sanderr, line 480:
```suggestion
     - all Resources with name > foo
```

src/inmanta/data/model.py
```diff
@@ -473,16 +473,30 @@ class PagingBoundaries:
...
+            end = None
+        ```
+
+    The fetched data will be: [foo1 ... foo10]
+
+    But the Pagingboundary will be constructed this way:
+        ```
+        Pagingboundary:
+            end = foo1
+            start = foo10    # Reversed because these are meant to be used by the neighbouring pages
```
sanderr, line 497:
```suggestion
            start = foo10 # Reversed because these are meant to map to like-named fields on neighbouring RequestedPagingBoundary
```

sanderr (changes requested):

sanderr:
@Hugo-Inmanta I made some final small changes to try to clarify so I'm rerequesting your review.

Hugo-Inmanta (approved):

sanderr (approved):

PR #8188, opened 2024-10-09
bump compiler version to 2025.1

Hugo-Inmanta:
# Description

As part of 7.3 release.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/bump-compiler-version.yml
+++ b/changelogs/unreleased/bump-compiler-version.yml
@@ -1,0 +1,3 @@
+description: Bump COMPILER_VERSION to 2025.1
+change-type: patch
+destination-branches: [master, iso7]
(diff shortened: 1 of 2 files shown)


sanderr (changes requested):
Should be a patch: https://internal.inmanta.com/development/core/technical-designs/compiler_version#bumping-the-compiler-version

sanderr (approved):

changelogs/unreleased/bump-compiler-version.yml
```diff
@@ -0,0 +1,3 @@
+description: Bump COMPILER_VERSION to 2025.1
+change-type: patch
+destination-branches: [master, iso7]
```
sanderr, line 3:
```suggestion
destination-branches: [master]
```

sanderr: iso7 will have a different version

Issue #8154, opened 2024-10-03
Make logfire an optional dependency


sanderr:
- Make it an optional dependency
- Make sure core still works if it's not installed (e.g. `ImportError`)
- keep pinned in `requirements.txt`
- Add (floating) to `requirements.dev.txt` iff we need it for testing and/or type checks

PR #8195, opened 2024-10-09
Make logfire as an optional dependency

hlloreda:
# Description

- `Logfire` is now an optional dependency

closes https://github.com/inmanta/inmanta-core/issues/8154

# Self Check:

- [ ] Changelog entry


requirements.dev.txt
```diff
@@ -1,5 +1,6 @@
 inmanta-dev-dependencies[pytest,async,core]==2.137.0
 bumpversion==0.6.0
+#-e .[tracing]
```
hlloreda, line 3: This needs to be uncommented (only used for testing purpose)

src/inmanta/tracing.py
```diff
@@ -19,21 +19,79 @@
...
+    import logfire._internal.config
+    import logfire.integrations
+    import logfire.integrations.pydantic
+    from logfire import LevelName, LogfireSpan
+    from logfire._internal.main import NoopSpan
+    from logfire.propagate import ContextCarrier
+
+    # Make sure we don't get warnings when it is off
+    logfire._internal.config.GLOBAL_CONFIG.ignore_no_config = True
+else:
```
hlloreda, line 41: I'm a bit on the fence if this or the try except should be better. Any Opinion?

I'm a bit worried that with a `try except` approach, we try to enable logging even though the dependencies are not installed. We could still assert that the env var is not set if the library is not present

tests/conftest.py
```diff
@@ -1983,13 +1982,3 @@ def disable_version_and_agent_cleanup_job():
...
     orchestrationservice.PERFORM_CLEANUP = old_perform_cleanup
-
-
-@pytest.fixture(scope="session", autouse=not PYTEST_PLUGIN_MODE)
-def configure_logfire():
-    """Configure logfire to ensure all the instrumentation works correctly and does not provide warnings. This does not
-    setup tracing for tests"""
-    logfire.configure(
-        send_to_logfire=False,
-        console=False,
```
hlloreda, line 1994: I don't think we need to keep this? As we are already calling `configure_logfire` in multiple places (components)

changelogs/unreleased/8154-optional-logfire-dependency.yml
hlloreda, line 1: test in another venv (or discuss it further)

src/inmanta/tracing.py
hlloreda: Change to try and except, we should not crash if this token was used for something else

changelogs/unreleased/8154-optional-logfire-dependency.yml
hlloreda: Let it because mypy and use "TYPING"

src/inmanta/tracing.py
```diff
@@ -82,16 +137,16 @@ def span(
     *,
     _tags: Sequence[str] | None = None,
     _span_name: str | None = None,
-    _level: LevelName | None = None,
+    _level: LevelName | None | "logfire.LevelName" = None,
```
hlloreda, line 140: This does not work, any suggestion?

setup.py
```diff
@@ -15,7 +15,6 @@
     "docstring-parser>=0.10,<0.17",
     "email-validator>=1,<3",
     "jinja2~=3.0",
-    "logfire>=0.46,<2.0",
     "more-itertools>=8,<11",
     "opentelemetry-instrumentation-asyncpg~=0.46b0",
```
wouterdb, line 19: this one has to be in the extra as well

src/inmanta/tracing.py
wouterdb: what doesn't work/ how does it nor work?

tox.ini
```diff
@@ -8,8 +8,8 @@ requires = pip
 # We need this to be allowed to use the timeout command
 allowlist_externals=timeout
 deps=
+    .
     -rrequirements.dev.txt
-    -rrequirements.txt
```
wouterdb, line 12: ?

wouterdb (changes requested):

tox.ini
hlloreda: I thought it was enough to install inmanta-core using `.` (setup.py)

PR #8218, opened 2024-10-15
Fine tune soft timeout 

hlloreda:
# Description

Fine-tuned soft timeout to retrieve some output

# Self Check:

- [ ] Changelog entry


tests/test_server.py
```diff
@@ -2123,3 +2126,46 @@ async def test_delete_active_version(client, clienthelper, server, environment):
...
+            ).decode("utf-8").strip()
+            assert False, "Subprocess should have failed!"
+        except subprocess.CalledProcessError as e:
+            actual_logs = e.output.decode("utf-8").strip()
+            expected_trace = """
+    def test_timeout_output():
+>       time.sleep(30)
+E       KeyboardInterrupt
+"""
+            assert expected_trace in actual_logs
```
wouterdb, line 2171: this is a bit too meta

Let's not add a test for it.

wouterdb (changes requested):

wouterdb (approved):

Issue #7590, opened 2024-05-02
Redesign on-disk layout of the agent/server


wouterdb:
Currently, the on disk layout in `/var/lib/inmanta` is quite confused. In iso8, this should be cleaned up

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/7590?planning-poker) with Zenhub @arnaudsjs @hlloreda @Hugo-Inmanta @jptrindade

PR #8220, opened 2024-10-16
Issue/refactor on disk layout

Hugo-Inmanta:
# Description

- [ ] ensure pytest-inmanta-lsm supports both old and new layout 
    - tentative PR: https://github.com/inmanta/pytest-inmanta-lsm/pull/453

closes https://github.com/inmanta/inmanta-core/issues/7590

Old vs New layout:
```                                                                                                            
       OLD                                                       NEW                                           
/var/lib/inmanta/                                         /var/lib/inmanta/                                    
    ├─ env_uuid_1/                                            ├─ server/                                       
    │   ├─ agent/                                                 ├─ env_uuid_1/                               
    │       ├─ code/                                              │   ├─ executors/                            
    │       ├─ env/                                               │   │   ├─ venvs/                            
    │       ├─ executor/                                          │   │   │  ├─ venv_blueprint_hash_1/         
    │          ├─ venv/                                           │   │   │  ├─ venv_blueprint_hash_2/         
    │          │  ├─ venv_blueprint_hash_1/                       │   │   │  ├─ ...                            
    │          │  ├─ venv_blueprint_hash_2/         ==>           │   │   ├─ code/                             
    │          │  ├─ ...                                          │   │      ├─ executor_blueprint_hash_1/     
    │          ├─ code/                                           │   │      ├─ executor_blueprint_hash_2/     
    │             ├─ executor_blueprint_hash_1/                   │   │      ├─ ...                            
    │             ├─ executor_blueprint_hash_2/                   │   ├─ scheduler.cfg                         
    │             ├─ ...                                          │   ├─ compiler/                             
    │                                                             │                                            
    ├─ env_uuid_2/                                                │                                            
    │   ├─ ( ... )                                                ├─ env_uuid_2/                               
    │                                                             │   ├─ ( ... )                                
    ├─ server/                                                                                                 
        ├─ environments/                                  /var/logs/inmanta/                                   
        │  ├─ env_uuid_1/                                     ├─ logs/                                         
        │  ├─ env_uuid_2/                                                                                      
        │  ├─ ...                                                                                              
        ├─ agents/                                  ==>                                                        
           ├─ env_uuid_1/                                                                                      
           │  ├─ agent.cfg                                                                                     
           ├─ env_uuid_2/                                                                                      
              ├─ agent.cfg                                                                                     
                                                                                                               
/var/logs/inmanta/                                                                                             
    ├─ logs/                                                                                                   
                                                                                                               
```                                                                                                            

# Self Check:

- [x] Changelog entry


src/inmanta/server/server.py
```diff
@@ -80,18 +81,15 @@ def check_storage(self) -> dict[str, str]:
...
         state_dir = config.state_dir.get()
         server_state_dir = os.path.join(state_dir, "server")
-        dir_map = {"server": _ensure_directory_exist(state_dir, "server")}
-        dir_map["environments"] = _ensure_directory_exist(server_state_dir, "environments")
-        dir_map["agents"] = _ensure_directory_exist(server_state_dir, "agents")
-        dir_map["logs"] = _ensure_directory_exist(config.log_dir.get())
+        dir_map = {
+            "server": ensure_directory_exist(state_dir, "server"),
+            "environments": ensure_directory_exist(server_state_dir, "environments"),
+            "agents": ensure_directory_exist(server_state_dir, "agents"),
```
hlloreda, line 89: Will this folder still exist after we switch to the Scheduler? I don't see it in your new layout (although it might be for compatibility reasons)

src/inmanta/server/agentmanager.py
```diff
@@ -1263,11 +1263,14 @@ async def __do_start_agent(
...
-        config_dir = os.path.join(self._server_storage["agents"], str(env.id))
+        top_dir: str = self._server_storage["scheduler_config"] if use_resource_scheduler else self._server_storage["agents"]
+        config_dir = os.path.join(top_dir, str(env.id))
+
         if not os.path.exists(config_dir):
             os.mkdir(config_dir)
 
-        config_path = os.path.join(config_dir, "agent.cfg")
+        file_name: str = "scheduler.cfg" if use_resource_scheduler else "agent.cfg"
+        config_path = os.path.join(config_dir, file_name)
```
hlloreda, line 1273: It's not something you should change, but I feel like we are relying sometimes on `pathlib` and sometimes on `os` for the same operations. I think it would be better to use only one (and pathlib might be the best candidate). But feel free to ignore this

tests/agent_server/deploy/e2e/conftest.py
hlloreda, line 1: Could you add a test for this?

src/inmanta/server/server.py
Hugo-Inmanta: No indeed it should be removed at some point but we have to keep it until the old agent is stripped out.

src/inmanta/agent/agent.py
Hugo-Inmanta, line 1: Not very relevant since old agent will be stripped out.

wouterdb (approved):

Issue #8186, opened 2024-10-09
Fix the agent view in the front-end


wouterdb:
The current agent view in the web console is based on the old agent, we want the same api to return data on the scheduler. 

This is mostly for backward compatibility, some fields may lose their full meaning, we want to retain them nonetheless.

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/8186?planning-poker) with Zenhub @arnaudsjs @hlloreda @Hugo-Inmanta @jptrindade

PR #8221, opened 2024-10-16
Make sure the agent view is working with the Scheduler

hlloreda:
# Description

Updated a test to make sure the agent view is working with the Scheduler

closes https://github.com/inmanta/inmanta-core/issues/8186

# Self Check:

- [ ] Changelog entry


hlloreda:
Data retrieved from the view:
```
[{'environment': '9f2158f8-b999-4b17-bbf7-048776497f81', 'name': '$__scheduler', 'last_failover': '2024-10-16T15:18:56.342750+00:00', 'paused': False, 'process_id': 'f5fc2fbc-8bd1-11ef-8330-42375fa4af91', 'process_name': 'fedora', 'unpause_on_resume': None, 'status': 'up'}, {'environment': '9f2158f8-b999-4b17-bbf7-048776497f81', 'name': 'agent1', 'last_failover': None, 'paused': False, 'process_id': None, 'process_name': None, 'unpause_on_resume': None, 'status': 'down'}]
```

tests/agent_server/deploy/e2e/test_autostarted.py
```diff
@@ -688,6 +688,13 @@ def wait_for_new_executor() -> bool:
     for children in current_resumed_children.values():
         assert children.is_running()
 
+    result = await client.get_agents(environment)
+    assert result.code == 200
+    actual_data = result.result["data"]
+    assert len(actual_data) == 2
+    assert all([not agent["paused"] for agent in actual_data])
+    assert {e["name"] for e in actual_data} == {const.AGENT_SCHEDULER_ID, "agent1"}
```
wouterdb, line 696: 1. I don't think we want to show the `AGENT_SCHEDULER_ID` 
2. I think we should try to set a meaningfull up/down  state (e.g. `up iff scheduler.state  == UP and not agent.paused`

wouterdb (changes requested):

wouterdb (changes requested):

tests/deploy/e2e/test_autostarted.py
```diff
@@ -1131,3 +1200,152 @@ async def check_resource_in_state(expected_resources: int = 1, state: str = "dep
...
+        "environment": environment,
+        "last_failover": actual_data[0]["last_failover"],
+        "name": "agent1",
+        "paused": False,
+        "process_id": None,
+        "process_name": None,
+        "status": "down",
+        "unpause_on_resume": None,
+    }
+    assert actual_data[0] == expected_data
```
hlloreda, line 1351: @wouterdb  I added locally this part in the test (for testing purposes only):
```
    await asyncio.sleep(10)
    await retry_limited(
        wait_for_terminated_status,
        timeout=const.EXECUTOR_GRACE_HARD + 2,
        current_children=children_after_deployment.children,
        expected_terminated_process=2,
    )
    await retry_limited(
        wait_for_terminated_status,
        timeout=const.EXECUTOR_GRACE_HARD + 2,
        current_children=children_after_deployment.children,
        expected_terminated_process=3,
    )
```
But this fails because the fork server and the executor seem somehow to survive: 
```
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/hugo/Documents/Inmanta/copies/inmanta-core/tests/utils.py(87)retry_limited()
-> raise AssertionError("Bounded wait failed")
(Pdb) up
> /home/hugo/Documents/Inmanta/copies/inmanta-core/tests/deploy/e2e/test_autostarted.py(1354)test_scheduler_killed()
-> await retry_limited(
(Pdb) children_after_deployment.children
[psutil.Process(pid=29456, name='python', status='terminated', started='11:45:53'), psutil.Process(pid=29495, name='inmanta: multiprocessing fork server', status='sleeping', started='11:45:55'), psutil.Process(pid=29497, name='inmanta: executor process 78f726cec1879275a4cc2fa2c103afb5 - disconnected', status='sleeping', started='11:45:55')]
```
The executor seems dead but not cleaned, is this expected?

hlloreda:
@wouterdb This test is failing: `test_agent_process_details_with_report` because it tries to fetch process details of a process that does not exist anymore (+ the fact that the scheduler is not returned from the agent view). How should I handle this? Is it still something we want to support (fetch process details) for scheduler processes?

src/inmanta/data/dataview.py
```diff
@@ -1211,19 +1211,21 @@ def get_base_url(self) -> str:
...
+                                                ELSE 'up'
                                             END) as status""",
-            from_clause=f" FROM {Agent.table_name()} as a LEFT JOIN public.agentinstance ai ON a.id_primary=ai.id "
-            " LEFT JOIN public.agentprocess ap ON ai.process = ap.sid",
-            filter_statements=[" a.environment = $1 "],
-            values=[self.environment.id],
+            from_clause=f" FROM  (SELECT sched.id, sched.halted FROM {Environment.table_name()} AS sched "
+            "WHERE sched.id =$1) AS sched "
+            f"RIGHT JOIN {Agent.table_name()} a ON a.environment = sched.id "
+            f"RIGHT JOIN (SELECT schedagent.name, schedagent.id_primary FROM {Agent.table_name()} AS schedagent "
```
wouterdb, line 1222:
```suggestion
            from_clause= f"RIGHT JOIN (SELECT schedagent.name, schedagent.id_primary FROM {Agent.table_name()} AS schedagent "
```

I think only the last bit is used?

wouterdb: Why a right join?

src/inmanta/data/dataview.py
```diff
@@ -1195,7 +1196,6 @@ def __init__(
     def allowed_filters(self) -> dict[str, type[Filter]]:
         return {
             "name": ContainsPartialFilter,
-            "process_name": ContainsPartialFilter,
```
wouterdb, line 1198: This will break the front-end. I would just make this filter do nothing.

src/inmanta/data/dataview.py
```diff
@@ -1211,19 +1211,21 @@ def get_base_url(self) -> str:
...
                                             END) as status""",
-            from_clause=f" FROM {Agent.table_name()} as a LEFT JOIN public.agentinstance ai ON a.id_primary=ai.id "
-            " LEFT JOIN public.agentprocess ap ON ai.process = ap.sid",
-            filter_statements=[" a.environment = $1 "],
-            values=[self.environment.id],
+            from_clause=f" FROM  (SELECT sched.id, sched.halted FROM {Environment.table_name()} AS sched "
+            "WHERE sched.id =$1) AS sched "
+            f"RIGHT JOIN {Agent.table_name()} a ON a.environment = sched.id "
+            f"RIGHT JOIN (SELECT schedagent.name, schedagent.id_primary FROM {Agent.table_name()} AS schedagent "
+            f"WHERE schedagent.name ='{const.AGENT_SCHEDULER_ID}') AS schedagent ON a.name <> schedagent.name ",
```
wouterdb, line 1223: 1. This 'AS' statement in here looks a bit weird to me
2. You have to make sure we select the scheduler from the right environment

tests/server/test_agents_api.py
```diff
@@ -118,12 +118,12 @@ def agent_names(agents: list[dict[str, str]]) -> list[str]:
     return [agent["name"] for agent in agents]
 
 
-@pytest.mark.parametrize("order_by_column", ["name", "status", "process_name", "last_failover", "paused"])
```
wouterdb, line 121: if would be  interesting to make only the absolute minimal amount of modification to this testcase, so we know we don't break the front-end

wouterdb:
> `test_agent_process_details_with_report`

Good question, I suppose you could report the process of the scheduler as the process of every agent and then it would all still work.

src/inmanta/data/dataview.py
```diff
@@ -1211,16 +1213,19 @@ def get_base_url(self) -> str:
...
                                             END) as status""",
-            from_clause=f" FROM {Agent.table_name()} as a LEFT JOIN public.agentinstance ai ON a.id_primary=ai.id "
-            " LEFT JOIN public.agentprocess ap ON ai.process = ap.sid",
-            filter_statements=[" a.environment = $1 "],
-            values=[self.environment.id],
+            from_clause=f" FROM {Agent.table_name()} a "
+            "LEFT JOIN (SELECT sa.name, sa.id_primary, ap.hostname, ai.process FROM agent sa  "
+            f"LEFT JOIN {AgentInstance.table_name()} ai ON sa.id_primary=ai.id "
+            f"LEFT JOIN {AgentProcess.table_name()} ap ON ai.process = ap.sid  "
+            "WHERE sa.environment = $1 AND sa.name = $2) sa_join ON a.name <> sa_join.name ",
```
wouterdb, line 1226: the join on `<>` I don't know about, my SQL is too weak. 
@Hugo-Inmanta can you check this out?

tests/server/test_agents_api.py
```diff
@@ -123,7 +123,7 @@ def agent_names(agents: list[dict[str, str]]) -> list[str]:
 async def test_agents_paging(server, client, env_with_agents: None, environment: str, order_by_column: str, order: str) -> None:
     result = await client.get_agents(
         environment,
-        filter={"status": ["paused", "up"]},
```
wouterdb, line 126: Why is this changed?

wouterdb (approved):

src/inmanta/data/dataview.py
Hugo-Inmanta: I think it is working as expected i.e. for every nominal agent we retrieve the info for the hostname and pid of the scheduler in the env.
@hlloreda  do I understand correctly that this `ON` clause is there just to have something to join on ? ie we would expect the same result if we passed `ON true` ? In this case a `CROSS JOIN` would be preferable I think

src/inmanta/data/dataview.py
```diff
@@ -1211,16 +1213,19 @@ def get_base_url(self) -> str:
 
     def get_base_query(self) -> SimpleQueryBuilder:
         base = SimpleQueryBuilder(
-            select_clause="""SELECT a.name, a.environment, last_failover, paused, unpause_on_resume,
-                                            ap.hostname as process_name, ai.process as process_id,
-                                            (CASE WHEN paused THEN 'paused'
-                                                WHEN id_primary IS NOT NULL THEN 'up'
-                                                ELSE 'down'
+            select_clause="""SELECT a.name, a.environment, a.last_failover, a.paused, a.unpause_on_resume,
```
Hugo-Inmanta, line 1216: Could you clean up the formatting a bit while you're at it, to improve readability ?

Hugo-Inmanta (changes requested):

PR #8238, opened 2024-10-21
Make logfire an optional dependency iso7

hlloreda:
# Description

Same changes were applied on master: https://github.com/inmanta/inmanta-core/pull/8195

# Self Check:

- [ ] Changelog entry


hlloreda:
Logfire is not used in ISO7

PR #8248, opened 2024-10-22
Scheduler recovery

### LINKED ISSUE #8014: resource scheduler: back up state to DB
The scheduler keeps some internal state. Some of it is directly derived from the DB state, but some of it is not. Examples include the current scheduled work, metadata related to event propagation (if any), ... This state should be somehow recoverable in case of failure.

- [ ] determine which state needs to be written to the DB for recovery
- [ ] design back up + recovery approach + request review
- [ ] implement

arnaudsjs:
# Description

This PR ensures that the scheduler state is recovered from the database at startup time.

closes #8014

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/in_process_executor.py
+++ b/src/inmanta/agent/in_process_executor.py
@@ -501,7 +501,8 @@ class InProcessExecutorManager(executor.ExecutorManager[InProcessExecutor]):
             await child.stop()
 
     async def start(self) -> None:
-        pass
+        assert all(e.is_stopped() for e in self.executors.values())
+        self.executors.clear()
 
     async def stop_for_agent(self, agent_name: str) -> list[InProcessExecutor]:
         if agent_name in self.executors:
--- a/tests/agent_server/deploy/e2e/test_scheduler_initialization.py
+++ b/tests/agent_server/deploy/e2e/test_scheduler_initialization.py
@@ -15,10 +15,12 @@
 
     Contact: code@inmanta.com
 """
+
 import pytest
 
-from inmanta import config, const
 import utils
+from inmanta import config, const
+
 
 @pytest.fixture
 async def server_pre_start(server_config):
@@ -26,9 +28,7 @@ async def server_pre_start(server_config):
     config.Config.set("config", "agent-repair-interval", "0")
 
 
-async def test_scheduler_initialization(
-    agent, resource_container, clienthelper, server, client, environment
-) -> None:
+async def test_scheduler_initialization(agent, resource_container, clienthelper, server, client, environment) -> None:
     """
     Ensure that when the scheduler starts, it only deploys the resources that need to be deployed,
     i.e. the resources that are not up-to-date or have outstanding events.
@@ -75,7 +75,7 @@ async def test_scheduler_initialization(
     for rid, expected_status in [
         ("test::Resource[agent1,key=key1]", const.ResourceState.deployed),
         ("test::Resource[agent1,key=key2]", const.ResourceState.failed),
-        ("test::Resource[agent1,key=key3]", const.ResourceState.skipped)
+        ("test::Resource[agent1,key=key3]", const.ResourceState.skipped),
     ]:
         result = await client.resource_details(tid=environment, rid=rid)
         assert result.code == 200
@@ -91,7 +91,6 @@ async def test_scheduler_initialization(
     assert len(result.result["data"]) == 1
     assert result.result["data"][0]["resource_version_ids"] == ["test::Resource[agent1,key=key1],v=1"]
 
-
     result = await client.get_resource_actions(
         tid=environment,
         resource_type="test::Resource",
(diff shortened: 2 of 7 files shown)


src/inmanta/agent/in_process_executor.py
```diff
@@ -501,7 +501,8 @@ async def stop(self) -> None:
             await child.stop()
 
     async def start(self) -> None:
-        pass
+        assert all(e.is_stopped() for e in self.executors.values())
```
wouterdb, line 504: is_stopped doesn't return if the executor is stopped, but if stop was requested.

I wonder if we have to do a join as well? (I don't exactly know why the assert is added)

tests/agent_server/deploy/e2e/test_scheduler_initialization.py
```diff
@@ -0,0 +1,135 @@
...
+        assert result.code == 200
+        assert result.result["data"]["status"] == expected_status.value, f"{rid} has unexpected state"
+
+    result = await client.get_resource_actions(
+        tid=environment,
+        resource_type="test::Resource",
+        agent="agent1",
+        exclude_changes=[const.Change.nochange.value],
+    )
+    assert result.code == 200
```
wouterdb, line 123: why is this called made twice?

tests/agent_server/deploy/e2e/test_scheduler_initialization.py
```diff
@@ -0,0 +1,135 @@
...
+
+    result = await client.get_resource_actions(
+        tid=environment,
+        resource_type="test::Resource",
+        agent="agent1",
+        exclude_changes=[const.Change.nochange.value],
+    )
+    assert result.code == 200
+    assert len(result.result["data"]) == 3
```
wouterdb, line 132: This is unexpected? I would expect 5?

1. one for key1, first deploy
2. one for key2, first deploy
3. one for key3, first deploy
4. one for key2, second deploy
5. one for key3, second deploy

wouterdb (changes requested):

tests/agent_server/deploy/e2e/test_scheduler_initialization.py
```diff
@@ -0,0 +1,135 @@
...
+    assert result.code == 200
+    assert len(result.result["data"]) == 1
+    assert result.result["data"][0]["resource_version_ids"] == ["test::Resource[agent1,key=key1],v=1"]
+
+    result = await client.get_resource_actions(
+        tid=environment,
+        resource_type="test::Resource",
+        agent="agent1",
+    )
+    assert result.code == 200
```
Hugo-Inmanta, line 99: Why is this required ?

Hugo-Inmanta (changes requested):

src/inmanta/agent/in_process_executor.py
arnaudsjs: > is_stopped doesn't return if the executor is stopped, but if stop was requested.

Indeed. The naming is rather confusing.

> I wonder if we have to do a join as well? (I don't exactly know why the assert is added)

I just wanted to make sure that if we call `start()` after `stop()` + `join()`, that the executors are actually stopped. So that we can be sure that no executors keep hanging after clearing the executors dictionary. I think this use case is only relevant for the tests.

tests/agent_server/deploy/e2e/test_scheduler_initialization.py
arnaudsjs: For no reason. I removed them. Thanks!

tests/agent_server/deploy/e2e/test_scheduler_initialization.py
arnaudsjs: The call to `get_resource_actions()` filters out the "nochange" actions. Failure events are logged as "nochange". As such the following happens:

First deploy (before restart):

* Only `test::Resource[agent1,key=key1]` gets deployed successfully so we only have one resource action

Second deploy (after restart):

* `test::Resource[agent1,key=key1]` was already successfully deployed, so the recovery process of the scheduler doesn't schedule it again.
* `test::Resource[agent1,key=key2]` and ``test::Resource[agent1,key=key3]`` are deployed successfully.

So after the restart there are three resource actions without the "nochange" type. `test::Resource[agent1,key=key1]` (from before the restart) and `test::Resource[agent1,key=key2]` and `test::Resource[agent1,key=key3]` (after the restart). Do I need to document this better. Or do I need to remove the filter on nochange?

tests/agent_server/deploy/e2e/test_scheduler_initialization.py
arnaudsjs: It isn't. I removed it. Thanks!

wouterdb (approved):

Hugo-Inmanta (approved):

PR #8249, opened 2024-10-22
Add reason field on task

### LINKED ISSUE #8192: Add back the reason field to deploys
When scheduling a deploy task, we want to add an end-user oriented reason to the deploy, so the user can check why a deploy was done.

It is possible to have multiple reasons, but we want to retain only one of the highest priority.
In practice, this means we want to set it when scheduling the task and update it when re-scheduling to to priority increase

hlloreda:
# Description

- Added reason field on task

closes https://github.com/inmanta/inmanta-core/issues/8192

# Self Check:

- [x] Changelog entry


src/inmanta/deploy/scheduler.py
```diff
@@ -169,16 +169,36 @@ async def deploy(self, priority: TaskPriority = TaskPriority.USER_DEPLOY) -> Non
...
         Trigger a deploy
         """
+        match priority:
+            case TaskPriority.USER_DEPLOY:
+                reason = "User has requested a deploy"
+            case TaskPriority.INTERVAL_DEPLOY:
+                reason = "Deploy is being triggered by interval"
+            case _ as e:
+                raise RuntimeError(f"Unhandled case: {e}!")
+
```
jptrindade, line 179: We are missing the `NEW_VERSION_DEPLOY`

src/inmanta/deploy/work.py
```diff
@@ -323,6 +324,7 @@ def deploy_with_context(
         :param resources: Set of resources that should be deployed. Adds a deploy task to the scheduled work for each
             of these, unless it is already scheduled.
         :param priority: The priority of this deploy.
+        :param priority: The reason of this deploy.
```
jptrindade, line 327:
```suggestion
        :param reason: The reason of this deploy.
```

src/inmanta/deploy/work.py
```diff
@@ -345,7 +347,22 @@ def deploy_with_context(
         for resourceId in queued:
             if resourceId in resources:
                 task = tasks.Deploy(resource=resourceId)
-                self.agent_queues._in_progress[task] = min(self.agent_queues._in_progress[task], priority)
+                assert task in self.agent_queues._in_progress
+                actual_task = [e for e in self.agent_queues._in_progress.keys() if e == task][0]
```
jptrindade, line 351: Can we make this a method in `agent_queues`?

src/inmanta/deploy/scheduler.py
```diff
@@ -422,7 +444,11 @@ async def report_resource_state(
                         task = Deploy(resource=resource)
                         assert task in self._work.agent_queues.in_progress
                         priority = self._work.agent_queues.in_progress[task]
-                        self._work.deploy_with_context(event_listeners, priority=priority, deploying=set())
+                        task_with_reason = [e for e in self._work.agent_queues.in_progress.keys() if e == task][0]
+                        assert task_with_reason.reason is not None
```
jptrindade, line 448: I would like to see some explanation if we reach this assert, it will make it a bit easier for debugging. Same for the others

jptrindade: Either a comment or the message of the assertion error

src/inmanta/deploy/work.py
```diff
@@ -345,7 +347,22 @@ def deploy_with_context(
...
+                    self.agent_queues._in_progress[task] == priority and actual_task.reason == reason
+                ):
+                    # We don't need to do anything in that particular case, the lowest priority is already set
+                    pass
+                else:
+                    raise RuntimeError(
+                        f"Current task is equals to 2 different priorities: currently stored `{actual_task}` "
+                        f"- `{self.agent_queues._in_progress[task]}` | new priority `{reason}` - "
+                        f"`{priority}"
+                    )
```
jptrindade, line 365: I don't think this is necessary (especially if you add the resource id in the reason)

jptrindade (changes requested):
Overall looks good. What do you feel about also adding the resource id of the resource that triggers the task? Might be redundant for some, but helpful with event propagation

src/inmanta/deploy/scheduler.py
```diff
@@ -169,16 +169,36 @@ async def deploy(self, priority: TaskPriority = TaskPriority.USER_DEPLOY) -> Non
         """
         Trigger a deploy
         """
+        match priority:
+            case TaskPriority.USER_DEPLOY:
+                reason = "User has requested a deploy"
+            case TaskPriority.INTERVAL_DEPLOY:
+                reason = "Deploy is being triggered by interval"
+            case _ as e:
+                raise RuntimeError(f"Unhandled case: {e}!")
```
wouterdb, line 178: It would be good to look at the reasons we reported on the old agent and start from that.

src/inmanta/deploy/tasks.py
```diff
@@ -45,13 +46,17 @@ class Task(abc.ABC):
     """
 
     resource: ResourceIdStr
-
     id: resources.Id = dataclasses.field(init=False, compare=False, hash=False)
+    reason: typing.Optional[str] = dataclasses.field(compare=False, hash=False, default=None)
```
arnaudsjs, line 50: Adding this field here has some implications. The `AgentQueues` class, for example, uses this object as key in a dictionary. This means that it relies on the fact that the hash of the object is identifying for the object. So in the previous implementation every `Deploy()` object for the same resource_id would be considered equal (also with respect to the `__eq__()` method). Now that we add the reason field, that's no longer true. Deploy object will be considered different if they have a different reason. I don't think that's something we want.

arnaudsjs, line 50: Why is this field optional? In which situations is this None?

src/inmanta/deploy/tasks.py
```diff
@@ -45,13 +46,17 @@ class Task(abc.ABC):
...
 
     def __post_init__(self) -> None:
         # use object.__setattr__ because this is a frozen dataclass, see dataclasses docs
         object.__setattr__(self, "id", resources.Id.parse_id(self.resource))
 
+    def set_reason(self, reason: str) -> None:
+        # use object.__setattr__ because this is a frozen dataclass, see dataclasses docs
+        object.__setattr__(self, "reason", reason)
```
arnaudsjs, line 58: This is something we should not do. It will change the hash of the object, which is rather bad, since this object is used as the key of the dictionary.

```
>>> import abc
>>> from dataclasses import dataclass
>>> @dataclass(frozen=True, kw_only=True)
... class Task(abc.ABC):
...     x: str
... 
>>> t = Task(x="test")
>>> dct = {}
>>> dct[t] = "aaa"
>>> 
>>> object.__setattr__(t, "x", "bbb")
>>> dct[t]
KeyError: Task(x='bbb'')
```

arnaudsjs (changes requested):

src/inmanta/deploy/work.py
hlloreda: I'm not sure to understand your comment

src/inmanta/deploy/work.py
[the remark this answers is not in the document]
hlloreda: When a task is in progress, it could not exist in the `task_by_resource`:
image

src/inmanta/deploy/tasks.py
hlloreda: There is a subtilty that was not highlighted enough: It's not taken into consideration in the hash of the object
```
>>> import abc
>>> import dataclasses
>>> from dataclasses import dataclass
>>> @dataclass(frozen=True, kw_only=True)
... class Task(abc.ABC):
...     x: str
...     y: typing.Optional[str] = dataclasses.field(compare=False, hash=False, default=None)
... 
>>> t = Task(x="test")
>>> dct = {}
>>> dct[t] = "aaa"
>>> object.__setattr__(t, "y", "bbb")
>>> dct[t]
'aaa'
>>> 
```

src/inmanta/deploy/tasks.py
hlloreda: You're right, but given the attributes in the `dataclasses.field`, everything should work as expected.

hlloreda: I have added a comment, let me know if it is better

src/inmanta/deploy/work.py
jptrindade: > What do you feel about also adding the resource id of the resource that triggers the task? 

Related to this. If we add the resource id of the resource that triggers the task we might get into a situation where this would change and we would be fine with it.

src/inmanta/deploy/scheduler.py
```diff
@@ -422,7 +432,14 @@ async def report_resource_state(
                         task = Deploy(resource=resource)
                         assert task in self._work.agent_queues.in_progress
                         priority = self._work.agent_queues.in_progress[task]
-                        self._work.deploy_with_context(event_listeners, priority=priority, deploying=set())
+                        task_with_reason = self._work.agent_queues.retrieve_task_in_progress(task)
+                        assert task_with_reason, "The task retrieved from the queue should exist"
+                        assert task_with_reason.reason, (
+                            "Every scheduled task should have a reason " "(linked to a given priority)!"
```
jptrindade, line 438:
```suggestion
                            "Every scheduled task should have a reason (linked to a given priority)!"
```

src/inmanta/deploy/work.py
jptrindade: The wording is also a bit weird

jptrindade (changes requested):

src/inmanta/deploy/work.py
jptrindade: But overall, no strong feelings regarding this

src/inmanta/deploy/scheduler.py
```diff
@@ -422,7 +432,14 @@ async def report_resource_state(
                         task = Deploy(resource=resource)
                         assert task in self._work.agent_queues.in_progress
                         priority = self._work.agent_queues.in_progress[task]
-                        self._work.deploy_with_context(event_listeners, priority=priority, deploying=set())
+                        task_with_reason = self._work.agent_queues.retrieve_task_in_progress(task)
+                        assert task_with_reason, "The task retrieved from the queue should exist"
+                        assert (
+                            task_with_reason.reason
+                        ), "Every scheduled task should have a reason (linked to a given priority)!"
```
wouterdb, line 439: can't we enforce this as the type level?

src/inmanta/deploy/tasks.py
wouterdb: it might make sense to add this to `PrioritizedTask` instead of task, it resolve the issue of optionality as well.  (or even QueueItem)

src/inmanta/deploy/tasks.py
```diff
@@ -45,13 +46,21 @@ class Task(abc.ABC):
...
 
     def __post_init__(self) -> None:
         # use object.__setattr__ because this is a frozen dataclass, see dataclasses docs
         object.__setattr__(self, "id", resources.Id.parse_id(self.resource))
 
+    def set_reason(self, reason: str) -> None:
+        # use object.__setattr__ because this is a frozen dataclass, see dataclasses docs
+        # This does not impact the __hash__ method of the dataclass because this field is not taken into consideration when
+        # computing the hash of the object (see dataclass field attributes)
+        object.__setattr__(self, "reason", reason)
```
wouterdb, line 62: if it is frozen, it is frozen, no hacking around it.

src/inmanta/deploy/work.py
```diff
@@ -246,6 +246,17 @@ def task_done(self, agent: str, task: tasks.Task) -> None:
...
+    def retrieve_task_in_progress(self, task: tasks.Task) -> Optional[tasks.Task]:
+        """
+        Retrieve the task that is currently stored in the Agent Queue. This task should contain information on why it was
+        scheduled.
+        """
+        actual_search = [e for e in self._in_progress.keys() if e == task]
+        if len(actual_search) == 0:
+            return None
+        else:
+            return actual_search[0]
```
wouterdb, line 258: linear search is a no-go in this code. All of this code is hyper efficient, this is not

src/inmanta/deploy/work.py
```diff
@@ -345,7 +358,23 @@ def deploy_with_context(
...
+                    self.agent_queues._in_progress[task] == priority and actual_task.reason == reason
+                ):
+                    # We don't need to do anything in that particular case, the lowest priority is already set
+                    pass
+                else:
+                    raise RuntimeError(
+                        f"Two different priorities with the same weight have been associated with the current task: "
+                        f"Old priority -> (`{self.agent_queues._in_progress[task]}`, reason: {actual_task.reason}) "
+                        f"- New priority -> (`{priority}`, reason: {reason})"
+                    )
```
wouterdb, line 377: I would not do this, I would expect `queue_put_nowait` to be updated to take the reason along with the priority when updating the priority?

wouterdb (changes requested):

src/inmanta/deploy/scheduler.py
wouterdb: yes, I would put this on slack

wouterdb: Comments on this one:

I would make all things concrete:

e.g.
`Deploy was triggered by timer` -> `Deploy start by timer at %s`
`A new version has been released` -> `Version %d has been released`

src/inmanta/deploy/work.py
wouterdb, line 377: We very explicitly don't care of two different priorities with the same weight have been associated with the current task and definitely don't want to raise an exception then, as it is expected to happen

src/inmanta/deploy/work.py
```diff
@@ -345,7 +358,23 @@ def deploy_with_context(
         for resourceId in queued:
             if resourceId in resources:
                 task = tasks.Deploy(resource=resourceId)
-                self.agent_queues._in_progress[task] = min(self.agent_queues._in_progress[task], priority)
+                assert task in self.agent_queues._in_progress
+                actual_task = self.agent_queues.retrieve_task_in_progress(task)
+                assert actual_task, "The task retrieved from the queue should exist"
+                if self.agent_queues._in_progress[task] > priority:
+                    self.agent_queues._in_progress[task] = priority
+                    actual_task.set_reason(reason=reason)
```
wouterdb, line 366: in progress tasks can and should not be modified

src/inmanta/deploy/tasks.py
arnaudsjs: > You're right, but given the attributes in the dataclasses.field, everything should work as expected.

Indeed. I overlooked that.

src/inmanta/deploy/tasks.py
arnaudsjs: I agree.

Issue #8193, opened 2024-10-09
Add new database connection pool setting


wouterdb:
The server and scheduler now share a database pool, postgresql+asynpg can't handle when we request more then postgresql will allow. 

- [ ] Add new settings for scheduler DB pool
- [ ] Add new setting for the server DB pool
- [ ] we default the new setting such that the server half of the old setting as default. Each scheduler gets 10% of the pool. 
- [ ] Document that every environment now needs a pool of its own
- [ ] add also to upgrade notes

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/8193?planning-poker) with Zenhub @arnaudsjs @hlloreda @Hugo-Inmanta @jptrindade

PR #8261, opened 2024-10-24
Issue/8193 add new database connection pool setting

Hugo-Inmanta:
# Description

- [x] Add new settings for scheduler DB pool
- [x] Add new setting for the server DB pool
- [x] we default the new setting such that the server half of the old setting as default. Each scheduler gets 10% of the pool. 
- [ ] Document that every environment now needs a pool of its own
- [x] add also to upgrade notes 

closes https://github.com/inmanta/inmanta-core/issues/8193

# Self Check:

- [x] Changelog entry


sanderr (commented):
I'm only partially in the loop for this one at the moment, I'm requesting Wouter instead.

@wouterdb if you don't have capacity for this, feel free to bounce it back.

src/inmanta/server/config.py
```diff
@@ -44,13 +44,39 @@
...
+
+db_service_connection_pool_min_size = Option(
+    "database",
+    "service_connection_pool_min_size",
+    5,
+    "Number of connections the database connection pool will be initialized with",
+    is_int,
+)
+db_service_connection_pool_max_size = Option(
+    "database", "service_connection_pool_max_size", 35, "Max number of connections in the database connection pool", is_int
```
jptrindade, line 78: Maybe we should mention that these are specific per environment (from what I understood, could be wrong)

jptrindade (approved):

src/inmanta/data/__init__.py
```diff
@@ -2644,6 +2647,29 @@ def to_dto(self) -> m.Environment:
...
+            "the resource scheduler in this environment.",
+            validator=convert_int,
+        ),
+        SCHEDULER_DB_CONNECTION_TIMEOUT: Setting(
+            name=SCHEDULER_DB_CONNECTION_TIMEOUT,
+            typ="positive_float",
+            default=60.0,
+            doc="Database connection timeout in seconds.",
+            validator=convert_positive_float,
+        ),
```
wouterdb, line 2672: I don't think we want to make this into environment settings. 

typically, the people with access to the DB setup have no access to the environment setting, and vice versa.

(also, setting these requires accessing the DB, so this is very risky business)

wouterdb (changes requested):
see comment

changelogs/unreleased/8193-add-new-database-connection-pool-setting.yml
```diff
@@ -0,0 +1,18 @@
...
+change-type: major
+destination-branches: [master]
+sections:
+  upgrade-note: >
+    Previously, a single database connection pool was shared among the server and the agent, configured
+    via the ``database.connection_pool_max_size`` and ``database.connection_pool_min_size`` options.
+    For each environment, the new resource scheduler uses a database connection pool of its own.
+    Please review the following connection pool options and make sure they fit your
+    needs. Do keep in mind that these options will apply to all active environments:
+      - For the scheduler: ["agent.db_connection_timeout", agent.db_connection_pool_max_size", agent.db_connection_pool_min_size"]
```
wouterdb, line 12: let's call it scheduler, not agent 

```suggestion
      - For the scheduler: ["scheduler.db_connection_timeout", scheduler.db_connection_pool_max_size", scheduler.db_connection_pool_min_size"]
```

changelogs/unreleased/8193-add-new-database-connection-pool-setting.yml
```diff
@@ -0,0 +1,18 @@
...
+      - For the server: [``database.service_connection_pool_max_size``, ``database.service_connection_pool_min_size``, ``database.connection_timeout``]
```
wouterdb, line 13: I would format these as 

`server.xxx` where xx is the same as the one for the scheduler

src/inmanta/server/agentmanager.py
```diff
@@ -1365,9 +1365,11 @@ async def _make_agent_config(
 name={opt.db_name.get()}
 username={opt.db_username.get()}
 password={opt.db_password.get()}
+service_connection_pool_min_size={agent_cfg.agent_db_connection_pool_min_size.get()}
+service_connection_pool_max_size={agent_cfg.agent_db_connection_pool_max_size.get()}
+service_connection_timeout={agent_cfg.agent_db_connection_timeout.get()}
```
wouterdb, line 1370: This is unexpected?

I would expect the scheduler to use it's own settings, not those of the server?

src/inmanta/agent/config.py
```diff
@@ -173,6 +173,31 @@
     is_lower_bounded_int(60),
 )
 
+agent_db_connection_pool_min_size: Option[int] = Option(
```
wouterdb, line 176: These need a default  that is derived from the actual value of the deprecated setting, not of the default of that setting.

src/inmanta/server/config.py
```diff
@@ -44,13 +44,39 @@
 db_name = Option("database", "name", "inmanta", "The name of the database on the postgresql server", is_str)
 db_username = Option("database", "username", "postgres", "The username to access the database in the PostgreSQL server", is_str)
 db_password = Option("database", "password", None, "The password that belong to the database user", is_str)
+
 db_connection_pool_min_size = Option(
-    "database", "connection_pool_min_size", 10, "Number of connections the pool will be initialized with", is_int
+    "database",
+    "connection_pool_min_size",
+    10,
+    "[DEPRECATED USE :inmanta.config:option:`database.service_connection_pool_min_size`] Number of connections the database connection pool will be initialized with",
```
wouterdb, line 52: There is a way to indicate deprecation that neatly links in the docs

Hugo-Inmanta: Added `predecessor` to the new server options

src/inmanta/agent/config.py
Hugo-Inmanta: Added dynamic defaults to pool size options for server and scheduler

src/inmanta/server/agentmanager.py
Hugo-Inmanta: Agreed, cleaned this up and refactored the `server_db_connect` function into `initialize_database_connection_pool` with proper arguments.

src/inmanta/server/config.py
Hugo-Inmanta: I cleaned this up a bit and mentioned for the scheduler settings that they are set for each environment.

src/inmanta/server/services/databaseservice.py
```diff
@@ -133,23 +143,41 @@ async def _check_database_pool_exhaustion(self) -> None:
...
+        username=database_username,
+        password=database_password,
+        create_db_schema=create_db_schema,
         connection_pool_min_size=connection_pool_min_size,
         connection_pool_max_size=connection_pool_max_size,
         connection_timeout=connection_timeout,
     )
-    LOGGER.info("Connected to PostgreSQL database %s on %s:%d", opt.db_name.get(), database_host, database_port)
+    LOGGER.info("Connected to PostgreSQL database %s on %s:%d", database_name, database_host, database_port)
     return out
```
jptrindade, line 183: What do you think about making these arguments optional, and if they are not provided, using the environment values?

wouterdb (approved):

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: Do you have a specific use case in mind ? I think these parameters are better off being set explicitly when starting up the server / the scheduler.

PR #8265, opened 2024-10-25
Make sure log lines can be pickled

wouterdb:
# Description

Make sure log items are both json and pickle compatible 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/handler.py
+++ b/src/inmanta/agent/handler.py
@@ -18,6 +18,7 @@
 
 import base64
 import inspect
+import json
 import logging
 import traceback
 import typing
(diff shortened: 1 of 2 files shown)


src/inmanta/agent/handler.py
```diff
@@ -424,20 +425,23 @@ def _log_msg(self, level: int, msg: str, *args: object, exc_info: bool = False,
         if exc_info:
             kwargs["traceback"] = traceback.format_exc()
 
-        for k, v in kwargs.items():
+        def clean_arg_value(k: str, v: object) -> object:
             try:
-                json_encode(v)
+                # make sure we have clean dict
```
arnaudsjs, line 430: This comment explains what the code does. What I am missing here is an explanation on why we have to do this.

arnaudsjs (changes requested):

src/inmanta/agent/handler.py
```diff
@@ -426,8 +426,18 @@ def _log_msg(self, level: int, msg: str, *args: object, exc_info: bool = False,
             kwargs["traceback"] = traceback.format_exc()
 
         def clean_arg_value(k: str, v: object) -> object:
+            """
+            Make sure we have a clean dict.
+
+            These values will be pickled and json serialized later down the stream.
+            As such we have to make sure both thigns are possible.
```
arnaudsjs, line 433:
```suggestion
            As such we have to make sure both things are possible.
```

arnaudsjs (approved):

PR #8266, opened 2024-10-25
Small improvements to the unknowns handling

arnaudsjs:
# Description

* Improve performance to the code path that transitively blocks/unblocks resources because the undefined state of a resource changed.
* Improve naming of methods.

# Self Check:

- [x] Changelog entry


src/inmanta/deploy/scheduler.py
```diff
@@ -385,14 +385,14 @@ async def _new_version(
...
+                    self._state.block_resource(resource, resources[resource], is_transitive=False)
                 for resource in new_desired_state:
                     self._state.update_desired_state(resource, resources[resource])
                 for resource in added_requires.keys() | dropped_requires.keys():
                     self._state.update_requires(resource, requires[resource])
-                transitively_blocked_resources: set[ResourceIdStr] = self._state.block_provides(resources=blocked_resources)
+                transitively_blocked_resources: Set[ResourceIdStr] = self._state.block_provides(resources=blocked_resources)
                 for resource in unblocked_resources:
-                    self._state.unblock_resource(resource)
+                    self._state.mark_as_defined(resource, resources[resource])
```
sanderr, line 395: Why not simply add these resources to `new_desired_state` as well? It's semantically accurate and it prevents having a third place where we update resource intent.

It does have the disadvantage that it adds it to `dirty`, while that may only be appropriate if it's not transitively blocked. But the same actually goes for any resource who's desired state is updated while it's transitively blocked, so we should probably simply check if a resource blocked prior to marking it dirty in `update_desired_state`?

src/inmanta/deploy/state.py
```diff
@@ -234,38 +199,60 @@ def block_provides(self, resources: Set[ResourceIdStr]) -> set[ResourceIdStr]:
...
-        _unblock(resource)
-        for res in self.requires.get_all_provides_transitively(resource):
-            is_blocked = self.resource_state[res].status is ResourceStatus.UNDEFINED or any(
-                self.resource_state[r].blocked.is_blocked() for r in self.requires.get(res, set())
-            )
-            if not is_blocked:
-                _unblock(res)
+        provides_view: Mapping[ResourceIdStr, Set[ResourceIdStr]] = self.requires.provides_view()
+        self.resources[resource] = details
+        self.resource_state[resource].status = ResourceStatus.HAS_UPDATE
```
sanderr, line 249: Iff we follow my suggestion from the other comment (can't link to in-progress review comments so quoting below), I think these two lines should go.

> Why not simply add these resources to new_desired_state as well? It's semantically accurate and it prevents having a third place where we update resource intent.

src/inmanta/deploy/state.py
```diff
@@ -234,38 +199,60 @@ def block_provides(self, resources: Set[ResourceIdStr]) -> set[ResourceIdStr]:
...
 
-        def _unblock(resource: ResourceIdStr) -> None:
-            self.resource_state[resource].blocked = BlockedStatus.NO
+        def update_blocked_status(resource: ResourceIdStr) -> bool:
+            """
+            Check if the deployment of the given resource is still blocked and mark it as unblocked if it is.
+
+            :return: True iff the given resource was unblocked.
+            """
             if self.resource_state[resource].status is ResourceStatus.UNDEFINED:
```
sanderr, line 235: We should probably also have a check for `self.resource_state[resource].blocked is BlockedStatus.NO` (or keep track of these in a set) to make sure we don't go down the same subtree twice in the case of diamond provides.

src/inmanta/deploy/state.py
```diff
@@ -234,38 +199,60 @@ def block_provides(self, resources: Set[ResourceIdStr]) -> set[ResourceIdStr]:
...
-                self.resource_state[resource].status = ResourceStatus.HAS_UPDATE
+                # The resource is undefined.
+                return False
+            if any(self.resource_state[r].blocked.is_blocked() for r in self.requires.get(resource, set())):
```
sanderr, line 238: This may still be expensive in the case of diamond provides, because we'll run this check N times. But optimizing it isn't trivial. Perhaps if we cache known blockers (`known_blockers: dict[ResourceIdStr, Optional[ResourceIdStr]]`) and only run the `any` check if the previously known blocker has since been unblocked?

src/inmanta/deploy/state.py
```diff
@@ -234,38 +199,60 @@ def block_provides(self, resources: Set[ResourceIdStr]) -> set[ResourceIdStr]:
...
+        while todo:
+            resource_id: ResourceIdStr = todo.pop()
+            if self.resource_state[resource_id].blocked.is_blocked():
+                # Resource is already blocked. All provides will be blocked as well.
+                continue
+            else:
+                # Block resource and its provides.
+                self.block_resource(resource_id, self.resources[resource_id], is_transitive=True)
+                result.add(resource_id)
+                todo.extend(provides_view.get(resource_id, set()))
```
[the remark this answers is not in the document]
arnaudsjs: I was assuming that the overhead of calling `is_blocked()` would be very small, but it doesn't hurt to add the `seen` set. I'll add it.

sanderr (commented):
I noticed one more thing (again, I can create a ticket if you prefer because this is stretching long, and it's less urgent than the other fixes). When the scheduler triggers a repair, it adds all resources to the dirty set. But we should really exclude blocked ones, shouldn't we?

src/inmanta/deploy/scheduler.py
sanderr: For future reference: the first concern was deemed irrelevant in the current status of the ticket, though it will likely become relevant in the context of follow-up bugfix tickets, somewhat related to the other two points. I will create relevant tickets and link them here.

src/inmanta/deploy/state.py
```diff
@@ -226,17 +232,38 @@ def mark_as_defined(self, resource: ResourceIdStr, details: ResourceDetails) ->
         :param details: The details of resource.
         """
 
+        # Cache used by the `update_blocked_status()` method to prevent the expensive check on the blocked
+        # status of the requirements of a resource. If the dictionary contains key value pair (X, Y), it
+        # indicates that resource Y is a requirement of resource X that is blocked.
```
sanderr, line 237:
```suggestion
        # indicates that resource Y is a requirement of resource X that was blocked, last we checked
```

src/inmanta/deploy/state.py
```diff
@@ -201,71 +166,120 @@ def add_up_to_date_resource(self, resource: ResourceIdStr, details: ResourceDeta
...
+                continue
+            else:
+                # Block resource and its provides.
+                self.block_resource(resource_id, self.resources[resource_id], is_transitive=True)
+                result.add(resource_id)
+                todo.extend(provides_view.get(resource_id, set()))
         return result
 
-    def unblock_resource(self, resource: ResourceIdStr) -> None:
+    def mark_as_defined(self, resource: ResourceIdStr, details: ResourceDetails) -> None:
```
sanderr, line 223: I think it might be beneficial to do this in bulk for all relevant resources so we can share the cache, but we'll probably get that for free when we fix the bug we discussed earlier, so I don't think we need to at the moment.

sanderr (changes requested):
Tiny change request on a comment, other than that it's good to go for me

sanderr (approved):

src/inmanta/deploy/scheduler.py
sanderr: #8284

PR #8268, opened 2024-10-25
Make executor report get facts to scheduler

### LINKED ISSUE #8190: Make executor report facts to the scheduler
On get fact, the executor reportsthe facts to the server, we want to move this to the scheduler, such that the scheduler write it to the DB.

this is similar to #8189 and make have a synergy with #8183

jptrindade:
# Description

* Short description here *

closes #8190 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -462,6 +462,16 @@ class VirtualEnvironmentManager(resourcepool.TimeBasedPoolManager[EnvBlueprint, 
 
 
 @dataclass
+class FactResult:
+    resource_id: ResourceVersionIdStr
+    action_id: uuid.UUID
+    parameters: list[dict[str, Any]]
+    started: datetime.datetime
+    finished: datetime.datetime
+    messages: list[LogLine]
+
+
+@dataclass
 class DeployResult:
     rvid: ResourceVersionIdStr
     action_id: uuid.UUID
@@ -542,7 +552,7 @@ class Executor(abc.ABC):
         pass
 
     @abc.abstractmethod
-    async def get_facts(self, resource: ResourceDetails) -> inmanta.types.Apireturn:
+    async def get_facts(self, resource: ResourceDetails) -> Optional[FactResult]:
         """
         Get facts for a given resource
         :param resource: The resource for which to get facts.
--- a/src/inmanta/deploy/persistence.py
+++ b/src/inmanta/deploy/persistence.py
@@ -25,7 +25,7 @@ from uuid import UUID
 from asyncpg import UniqueViolationError
 
 from inmanta import const, data
-from inmanta.agent.executor import DeployResult
+from inmanta.agent.executor import DeployResult, FactResult
 from inmanta.const import TERMINAL_STATES, TRANSIENT_STATES, VALID_STATES_ON_STATE_UPDATE, Change, ResourceState
 from inmanta.data.model import AttributeStateChange, ResourceIdStr, ResourceVersionIdStr
 from inmanta.protocol import Client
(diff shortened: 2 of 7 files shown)


src/inmanta/agent/write_barier_executor.py
```diff
@@ -21,10 +21,10 @@
 import uuid
 from copy import deepcopy
 
-import inmanta.types
+import inmanta.types  # noqa: F401
```
jptrindade, line 24: Not exactly sure if this is needed or we can remove it

src/inmanta/agent/executor.py
```diff
@@ -35,8 +35,6 @@
 from typing import Any, Dict, Optional, Sequence, cast
 from uuid import UUID
 
-import inmanta.types
-import inmanta.util
```
jptrindade, line 39: This may be wrong to do

src/inmanta/agent/executor.py
```diff
@@ -471,6 +469,19 @@ async def create_member(self, member_id: EnvBlueprint) -> ExecutorVirtualEnviron
...
 
 
+@dataclass
+class FactResult:
+    resource_id: ResourceVersionIdStr
+    # Failed fact checks may have empty action_id
+    action_id: Optional[uuid.UUID]
+    started: datetime.datetime
+    finished: datetime.datetime
+    succeeded: bool
```
wouterdb, line 479: I think we tend to use success instead of suceeded

src/inmanta/agent/write_barier_executor.py
wouterdb: it not used, remove

src/inmanta/deploy/persistence.py
```diff
@@ -331,3 +347,15 @@ async def dryrun_update(self, env: UUID, dryrun_result: DryrunResult) -> None:
...
+        await self.client.set_parameters(tid=self.environment, parameters=fact_result.parameters)
+        await self.client.resource_action_update(
+            tid=self.environment,
+            resource_ids=[fact_result.resource_id],
+            action_id=fact_result.action_id,
+            action=const.ResourceAction.getfact,
+            started=fact_result.started,
+            finished=fact_result.finished,
+            messages=fact_result.messages,
+        )
```
wouterdb, line 361: We could choose to write this to the DB directly, but not required

tests/deploy/test_scheduler_agent.py
```diff
@@ -251,6 +250,9 @@ def check_with_scheduler(self, scheduler: ResourceScheduler) -> None:
             if status:
                 assert scheduler._state.resource_state[resource].status == status
 
+    def set_parameters(self, fact_result: FactResult) -> None:
+        self.state[Id.parse_id(fact_result.resource_id).resource_str()] = const.ResourceState.deploying
```
wouterdb, line 254: Why deploying?

jptrindade: Just copy pasted it and forgot to change

jptrindade: I noticed there is not a specific state that I associate with the fact refresh, maybe available. What do you suggest?
```
    unavailable = "unavailable"  # This state is set by the agent when no handler is available for the resource
    skipped = "skipped"  #
    dry = "dry"
    deployed = "deployed"
    failed = "failed"
    deploying = "deploying"
    available = "available"
```

src/inmanta/deploy/persistence.py
jptrindade: I chose to leave it like this to be consistent with the `dryrun_update` I figured we would do this on the next step. But if you want, we could do this (otherwise I'll create a ticket for it)

tests/deploy/test_scheduler_agent.py
wouterdb: not do anything I think

wouterdb (approved):

Hugo-Inmanta (approved):

PR #8275, opened 2024-10-28
refactor for reason field

### LINKED ISSUE #8192: Add back the reason field to deploys

wouterdb:
# Description
closes #8192 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -36,7 +36,7 @@ from inmanta.data.model import ResourceIdStr, ResourceType, ResourceVersionIdStr
 from inmanta.deploy import work
 from inmanta.deploy.persistence import StateUpdateManager, ToDbUpdateManager
 from inmanta.deploy.state import DeploymentResult, ModelState, ResourceDetails, ResourceState, ResourceStatus
-from inmanta.deploy.tasks import Deploy, DryRun, RefreshFact, Task
+from inmanta.deploy.tasks import Deploy, DryRun, RefreshFact
 from inmanta.deploy.work import PrioritizedTask, TaskPriority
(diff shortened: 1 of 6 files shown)


src/inmanta/deploy/work.py
```diff
@@ -59,6 +59,7 @@ class PrioritizedTask(Generic[T]):
 
     task: T
     priority: TaskPriority
+    reason: str | None = None
```
wouterdb, line 62: Do I take this out of `__eq__` ?

sanderr (commented):
The changes are not complete (e.g. bumping is missing) but I think that you are aware of that, given that this is a draft review of the concept rather than a full solution?

hlloreda (approved):

src/inmanta/deploy/scheduler.py
```diff
@@ -202,24 +202,28 @@ async def stop(self) -> None:
         self._work.agent_queues.send_shutdown()
         await asyncio.gather(*self._workers.values())
 
-    async def deploy(self, priority: TaskPriority = TaskPriority.USER_DEPLOY) -> None:
+    async def deploy(self, reason: str, priority: TaskPriority = TaskPriority.USER_DEPLOY) -> None:
```
sanderr, line 205: I'd prefer this as a kw-only arg but I'm not against it like this.

sanderr (changes requested):

sanderr (approved):

PR #8276, opened 2024-10-28
Issue/8145 reschedule deferred deploys order

### LINKED ISSUE #8145: resource scheduler: reschedule deferred deploys at previous insert order
When a deploy task is moved from its agent queue back to the waiting set (e.g. because a dependency was added to the scheduled work), it should retain its queue insert order so that when it is re-queued later on, it can jump (part of) the queue. This to prevent starvation.

Add a test case.

Related to #8015. It was initially assumed to be already implemented so it was deemed out of scope for the ticket.

sanderr:
# Description

- Ensure that when a deploy task gets deferred because of an incoming dependency, it is added back to its original position in the queue when it becomes unblocked again, mitigating starvation.
- Small refactor of PrioritizedTask and related classes because they were becoming unwieldy. Replaced with two protocols: one for the fields required internally, the other for the ones to be exposed to the scheduler (and a third combining the two). This makes it easy to have different dataclasses with variations, while keeping them consistent.
- Small bugfix on `_new_version`: take *all* resources, and then a set of up-to-date resources, rather than to disjunct dicts. This caused us to consider all up-to-date ones as "deleted resources" later on, essentially acting as if they were not there at all. I expect the first export after server start would in practice lead to an almost complete repair (everything that was in a known good state when the server came up).

closes #8145

# Self Check:

- [x] Changelog entry


sanderr (commented):
Note to self: #8275 has an impact on the dropping of the `PrioritizedTask` class. There are two options:
- add it back
- add the reason field to `QueueItem` and `BlockedDeploy`

The second still seems the most readable to me. Could be implemented with a `TaskWithMetadata` (working name) protocol for `(task: Task, priority: TaskPriority, tiebreaker: Optional[Tiebreaker], reason: Reason)`

src/inmanta/deploy/work.py
```diff
@@ -47,50 +47,41 @@ class TaskPriority(IntEnum):
...
+    insert_order: Tiebreaker
 
     # Mutable state
     deleted: bool = False
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, TaskQueueItem):
             return NotImplemented
-        return (self.task.priority, self.insert_order) == (other.task.priority, other.insert_order)
+        return (self.priority, self.insert_order, self.deleted) == (other.priority, other.insert_order, other.deleted)
```
sanderr, line 76: Note to self: should this consider `self.task` as final tiebreaker value, purely for semantical reasons?

src/inmanta/agent/cache.py
sanderr, line 1: This is a simple cleanup I did when I happened to encounter it. `sys.maxsize` is not in any way a max int, and so it should not be used like this.

jptrindade:
Looks nice!

jptrindade (approved):

src/inmanta/deploy/work.py
```diff
@@ -406,42 +459,63 @@ def extend_blocked_on(resource: ResourceIdStr, new_blockers: set[ResourceIdStr])
...
+                if task in self.agent_queues:
                     # simply add it again, the queue will make sure only the highest priority is kept
-                    self.agent_queues.queue_put_nowait(prioritized_task)
-                if resource in self._waiting:
-                    if self._waiting[resource].task.priority > priority:
-                        self._waiting[resource].task = prioritized_task
+                    self.agent_queues.queue_put_nowait(task, priority=priority, reason=reason)
+                if resource in self._waiting and priority < self._waiting[resource].priority:
+                    self._waiting[resource].priority = priority
+                    self._waiting[resource].priority_tiebreaker = None
```
wouterdb, line 507: I would expect the tie breaker to remain when bumping priority?

src/inmanta/deploy/scheduler.py
```diff
@@ -311,16 +311,11 @@ async def _initialize(self) -> None:
...
-            rid: resources[rid] for rid in resources.keys() if rid not in increment
-        }
-
         await self._new_version(
             version,
-            resources=resources_to_deploy,
+            resources=resources,
             requires=requires,
-            up_to_date_resources=up_to_date_resources,
+            up_to_date_resources=resources.keys() - increment,
```
wouterdb, line 318: this also happens to be the second return value of `get_increment`

tests/deploy/test_scheduler_agent.py
```diff
@@ -1816,15 +1816,303 @@ async def test_scheduler_priority(agent: TestAgent, environment, make_resource_m
...
+
+    # add some more tasks to the middle priority level
+    work.deploy_with_context(
+        resources={rid1, rid2, rid3, rid4, rid5},
+        reason="Bulk deploy",
+        priority=TaskPriority(2),
+    )
+    # add one more task to the more urgent priority level for some relative ordering asserts later on
+    # -> priority 1: rid2, rid6
+    # -> priority 2: rid1, rid3, rid4, rid5 (latter three in any order)
```
wouterdb, line 1886: I wonder if we want to always make the tie breaker unique?

wouterdb (commented):
Very cool

src/inmanta/deploy/work.py
sanderr: We're already moving to a more urgent bucket, so we're moving forward from where we're scheduled. I wouldn't expect it to jump the queue of that bucket, considering that other items in the bucket were requested first *for that priority*.

Does that reasoning change your mind, or should we take this to Slack?

src/inmanta/deploy/scheduler.py
sanderr: Right, will fix.

tests/deploy/test_scheduler_agent.py
sanderr: We could address it in this test case by requesting three separate deploys instead of the one bulk deploy, but I think that would add verbosity for little gain (we don't need to know the order, we only need to inspect it so we can assert that it is maintained further on).

src/inmanta/deploy/work.py
sanderr: Discussed further in person. Conclusion: current resetting of the tiebreaker is good, but we should assign it when a task gets scheduled rather than when it gets queued. Reworked in 507ea3126.

src/inmanta/deploy/work.py
```diff
@@ -243,31 +259,41 @@ def queue_put_nowait(
...
-        :param priority_tiebreaker: The tiebreaker value for comparing this task's priority equal-priority tasks.
-            Should only be set if this task was previously dropped from the agent queues, in order to insert it back
-            at its previous position.
+        :param requested_at: Integer value representing when this task was initially requested for the given priority (e.g. if
+            it was deferred for some time. Serves as tiebreaker for comparing this task with equal-priority tasks. If none, the
+            next value in line is used, resulting in the task being added at the end of its priority queue. A value can be
+            reserved (for adding to the queue in the future) through the reserve_requested_at() method.
         """
-        priority_tiebreaker = priority_tiebreaker if priority_tiebreaker is not None else self._entry_count
+        requested_at = requested_at if requested_at is not None else self._entry_count
```
wouterdb, line 275:
```suggestion
        requested_at = requested_at if requested_at is not None else self.reserve_requested_at()
```

tests/deploy/test_scheduler_agent.py
```diff
@@ -2112,6 +2122,8 @@ def get_sorted_tasks() -> Sequence[tuple[tasks.Task, str]]:
     # queued priority should have been shifted as well
     assert work.agent_queues.queued()[task_fourth_requires].priority == TaskPriority(-1)
 
+    # TODO: bump priority on queued task without adding requires
```
wouterdb, line 2125: todo alert

sanderr: Right, thanks

src/inmanta/deploy/work.py
sanderr: Actually, this line should go in its entirety. I moved this further down and forgot to remove this.

wouterdb (approved):

jptrindade (approved):
Looks nice! Added a small comment

src/inmanta/deploy/work.py
```diff
@@ -174,34 +230,71 @@ def discard(self, task: tasks.Task) -> Optional[TaskPriority]:
...
             already_queued.deleted = True
-        item: TaskQueueItem = TaskQueueItem(task=prioritized_task, insert_order=self._entry_count)
+        item: TaskQueueItem = TaskQueueItem(
+            task=task,
+            priority=priority,
+            requested_at=(
+                self.reserve_requested_at()
+                if (
+                    # no requested_at given => use default
+                    requested_at is None
```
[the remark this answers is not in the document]
sanderr: Good catch, thanks

Issue #8182, opened 2024-10-09
Add per resource timers


wouterdb:
# Goal

Make the repair and deploy timers per-resource instead of per-scheduler

- [ ] per resource deploy timer
- [ ] per resource repair timer
- [ ] remove per-scheduler timer 

The idea is to remove the global timers and replace them with per-resource timers which behave the same, except for their scope. When a deploy happens, the timers will have to be reset. 

The main challenge it to find a good way to create these timers, as we will have many of them.
Using the eventloop timer or our own scheduler framework are options.

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/8182?planning-poker) with Zenhub @arnaudsjs @hlloreda @Hugo-Inmanta @jptrindade

PR #8292, opened 2024-10-30
Issue/8182 add per resource timers

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/8182

# Self Check:

- [x] Changelog entry


src/inmanta/deploy/scheduler.py
```diff
@@ -162,6 +168,9 @@ def reset(self) -> None:
         self._work.reset()
         self._workers.clear()
         self._deploying_latest.clear()
+        for task in self.resource_periodic_deploys.values():
+            task.cancel()
```
sanderr, line 172: Don't you still need to await a task you cancel?
```suggestion
            task.cancel()
        with contextlib.suppress(asyncio.CancelledError):
            await asyncio.gather(*self.resource_periodic_deploys.keys())
```

(not sure if this works but I think it will)

sanderr (changes requested):
At the moment this only adds a repair timer, not a deploy one, but I assume you are aware of that.

src/inmanta/agent/config.py
```diff
@@ -174,6 +174,16 @@
     is_lower_bounded_int(60),
 )
 
+scheduler_resource_compliance_check_window = Option(
```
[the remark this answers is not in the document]
Hugo-Inmanta: If we reuse the old ones I will have to add support for the case a cron schedule is passed but I think it's doable.

src/inmanta/deploy/scheduler.py
```diff
@@ -433,6 +442,50 @@ async def _run_for_agent(self, agent: str) -> None:
...
+            # TODO Add dependents ?
+            async with self._scheduler_lock:
+                LOGGER.debug(
+                    "Periodic deploy for resource %s. Triggering deploy for resources: %s", str(resource), str(to_deploy)
+                )
+                self._work.deploy_with_context(to_deploy, priority=TaskPriority.INTERVAL_REPAIR, deploying=set())
+
+        async def _recurrent_compliance_check(resource: ResourceIdStr) -> None:
+            await _trigger_deploy(resource)
+            await delay(_recurrent_compliance_check(resource), self._compliance_check_window)
```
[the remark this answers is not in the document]
Hugo-Inmanta: You were correct, I tried with a very small interval and got a recursion depth error. I refactored this whole bit to use an infinite while loop with a sleep.

src/inmanta/agent/config.py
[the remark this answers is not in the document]
sanderr: That's a fair point. The cron expressions sort of go against the grain of what we're trying to express here. I'd pass this by @wouterdb.

src/inmanta/deploy/scheduler.py
```diff
@@ -433,6 +442,50 @@ async def _run_for_agent(self, agent: str) -> None:
 
             self._work.agent_queues.task_done(agent, task)
 
+    async def _start_compliance_check(self, resource: ResourceIdStr) -> None:
+        """
+        This method is expected to be called when the given resource has just finished deploying,
+        no matter the result itself i.e. success or failure.
+        """
+        if (previous_task := self.resource_periodic_deploys.get(resource)) is not None:
+            previous_task.cancel()
```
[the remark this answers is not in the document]
Hugo-Inmanta: 1. I think new implementation circumvents this issue
2. fixed this

changelogs/unreleased/8182-add-per-resource-timers.yml
```diff
@@ -0,0 +1,6 @@
+description: >
+  Add ``scheduler.resource-compliance-check-window`` agent config option to set the maximum time window
+  between two compliance checks on a per-resource basis.
```
Hugo-Inmanta, line 3: TODO reuse previous config option instead

src/inmanta/deploy/scheduler.py
```diff
@@ -433,6 +442,50 @@ async def _run_for_agent(self, agent: str) -> None:
...
+
+        self.resource_periodic_deploys[resource] = self.create_recurring_compliance_check(resource)
+
+    def create_recurring_compliance_check(self, resource: ResourceIdStr) -> asyncio.Task[None]:
```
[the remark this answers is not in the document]
Hugo-Inmanta: I'll stick with deploy and repair

src/inmanta/deploy/scheduler.py
```diff
@@ -433,6 +442,50 @@ async def _run_for_agent(self, agent: str) -> None:
...
+        """
+        Schedule recurring deploys for the given resource every <scheduler_resource_compliance_check_window> seconds
+        and return the associated task. This task will be canceled and scheduled again when the next deploy finishes
+        for this resource.
+        """
+
+        async def delay(coro: Coroutine[Any, Any, None], seconds: float) -> None:
```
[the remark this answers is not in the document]
Hugo-Inmanta: refactored this bit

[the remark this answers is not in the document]
Hugo-Inmanta: refactored this bit

src/inmanta/agent/config.py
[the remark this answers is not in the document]
wouterdb: Otoh, the cron expressions are a very important feature. For operations, to have all repairs at night is very important. 

So we can't just drop that.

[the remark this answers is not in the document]
sanderr: For the cron expression we probably don't want to make it per-resource btw (as long as the setting is global at least). Otherwise we'd have *a lot* of per-resource deploys/repairs all at the same time, all potentially pulling in dependencies. It makes a lot more sense to consider it a global (environment-wide) trigger then.

src/inmanta/deploy/scheduler.py
```diff
@@ -433,6 +442,50 @@ async def _run_for_agent(self, agent: str) -> None:
...
+            """
+            Helper function to run a coroutine after a delay
+            """
+            await asyncio.sleep(seconds)
```
wouterdb: This delay mechanism was not what I had in mind when thinking about timers. 

I can't determine if it is very smart or very dangerous.

I am suspicious about it and I wonder why we don't use any of the existing scheduling tools we have.

src/inmanta/deploy/scheduler.py
```diff
@@ -433,6 +442,50 @@ async def _run_for_agent(self, agent: str) -> None:
...
+            await coro
+
+        async def _trigger_deploy(resource: ResourceIdStr) -> None:
+            """
+            Check that the given resource is in a known good state with no pending updates and schedule a repair if need be
+            """
+            to_deploy: set[ResourceIdStr] = set()
+            to_deploy.add(resource)
+            # TODO Add dependents ?
```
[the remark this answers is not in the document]
wouterdb: This is the big one right! 

For repair, I think we should only be concerned with individual resources. We repair the one-by-one, independently. They respond only to the know states.

For deploy, it becomes more subtle. 
If we define the deploy as the minimal interval between subsequent deploys of a resource, it is clear: anything in the dirty set gets an individual re-try timer. They will naturally line up in the correct order. (except that a pure skip for dependencies is also a deploy for the scheduler, making the external observable behavior complicated again)

If we define it as the retry interval for individual dirty items, we can just run the timer for individual items. they will also naturally line up (deepest ones will expire first and 'block' the others, when they are done, they will immediately queue them, resetting the timer).

If we define it as the retry interval for all resources, we arrive back to the mechanism we already had. 

Retry for a disconnected subtree is what is suggested here, feels a bit weird to me?


I think we need a clear storry here, that we want to present to the wider team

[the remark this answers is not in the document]
sanderr: Regarding how we define it, I assumed we were talking "he retry interval for individual dirty items" in this iteration.

As to "they will naturally line up", I'm not sure that is the case (in this interpretation / definition). e.g. consider a resource `host` in the common (not the right name but you get the point) resource set and two `file` resources `hello` and `world` that depend on that `host`, each in their own resource set. We first export and release the `hello` resource set, causing `host` and `hello` to be scheduled for deploy. `host` fails, `hello` is skipped. Sometime later we export and release the `world` resource set, causing `host` and `world` to be scheduled. `host` fails again and `world` is skipped. Now the retry times for `host` and `world` are out of sync, unless we just keep stacking triggers for the same resource, but that doesn't seem right to me either.

The worst instance of this issue would occur with a long retry interval, and the two resource deploys only a few seconds after one another. This would cause `hello` to be skipped again, then a few seconds after `host` and `world` might deploy successfully, and `hello` would have to wait for almost the entire retry interval once more.

> Retry for a disconnected subtree is what is suggested here, feels a bit weird to me?

I think all approaches have something to say for them. This one solves some of the issues that others have, but I see your point.

> I think we need a clear storry here, that we want to present to the wider team

I wholeheartedly agree. We probably should have gotten this more clear earlier. We should start from what makes sense conceptually and go from there. Perhaps we could do the simple thing where we have simple per-resource triggers, and then address the issue with the out-of-order retries through a different mechanism, e.g. (draft idea) force "success event propagation" towards skipped resources, even if `receive_events=false`. This would cause skipped resources to be retried when they become unblocked. We'd probably want to move ahead the plan to differentiate between default `deploy` skip and explicit handler skip.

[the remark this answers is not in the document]
sanderr: Moved discussion to [Slack thread](https://inmanta.slack.com/archives/CKRF0C8R3/p1730735805023259)

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: This is how I interpreted this part of the ticket `Using the eventloop timer`. I'll go with the scheduling framework instead.

src/inmanta/agent/config.py
```diff
@@ -102,54 +102,35 @@
...
     " interpreted in UTC. The expected format is ``[sec] min hour dom month dow [year]`` (If only 6 values are provided, they"
     " are interpreted as ``min hour dom month dow year``)."
-    " A deploy will be requested at the scheduled time. Note that if a cron"
-    " expression is used the :inmanta.config:option:`config.agent_deploy_splay_time` setting will be ignored."
+    " A deploy will be requested at the scheduled time."
     " Set this to 0 to disable the scheduled deploy runs.",
     is_time_or_cron,
     predecessor_option=agent_interval,
 )
-agent_deploy_splay_time = Option(
```
Hugo-Inmanta, line 115: Can we just remove this and `agent_repair_splay_time` here or should we first deprecate them ?

src/inmanta/deploy/scheduler.py
```diff
@@ -615,37 +780,102 @@ async def report_resource_state(
...
+
+                # Set of dependant resources that are interested in deploys of the current resource
+                concerned_resources: set[ResourceIdStr] = set()
+                provides: Set[ResourceIdStr]
                 if deployment_result is DeploymentResult.DEPLOYED:
                     self._state.dirty.discard(resource)
+
+                    if previous_deployment_result != DeploymentResult.DEPLOYED:
+                        # This resource went from not deployed to deployed
+                        # we have to inform its dependants regardless of event propagation
```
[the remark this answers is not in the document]
Hugo-Inmanta: Added some rationale as far as I understand it from the slack discussion, could you double check that it is correct, I'm not confident about the event propagation part.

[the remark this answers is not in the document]
sanderr: What you linked is the concrete proposal. I had also forgotten about the rationale behind it, which is [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1730735864437679?thread_ts=1730735805.023259&cid=CKRF0C8R3).

changelogs/unreleased/8182-add-per-resource-timers.yml
```diff
@@ -0,0 +1,9 @@
+description: >
+  Add support for periodic per-resource deploys and repairs.
+  When specified as a number of seconds, the ``config.agent-repair-interval`` agent config option will now be
```
wouterdb, line 3: I would refer to the environment config

src/inmanta/agent/agent_new.py
```diff
@@ -104,93 +89,9 @@ def _set_deploy_and_repair_intervals(self) -> None:
...
-        self._sched.add_action(action, schedule)
-        self._time_triggered_actions.add(ScheduledTask(action=action, schedule=schedule))
-
-    def _disable_time_triggers(self) -> None:
-        for task in self._time_triggered_actions:
-            self._sched.remove(task)
-        self._time_triggered_actions.clear()
+        # quick and dirty, fix later:
+        self.scheduler._repair_timer = cfg.agent_repair_interval.get()
+        self.scheduler._deploy_timer = cfg.agent_deploy_interval.get()
```
wouterdb, line 94: Why not load these from the scheduler itself?

src/inmanta/agent/config.py
wouterdb: I would remove them, the agent is gone anyways, this is now purely internal between server  and scheduler. 

We could even read it from the DB directly is we would like to.

src/inmanta/deploy/scheduler.py
```diff
@@ -243,7 +245,32 @@ def __init__(
...
+        self._repair_timer: Union[int, str] = agent_config.agent_repair_interval.get()
+
+        self.global_periodic_repair_task: ScheduledTask | None = None
+
+        self.resource_periodic_deploys: dict[ResourceIdStr, ScheduledTask] = {}
+        self._deploy_timer: Union[int, str] = agent_config.agent_deploy_interval.get()
+
+        self.global_periodic_deploy_task: ScheduledTask | None = None
+
+    async def clear_periodic_tasks(self) -> None:
```
wouterdb, line 259: Why is this async?

I think it shouldn't be, as it creates a race where the timer fires after clearing all other things.

src/inmanta/deploy/scheduler.py
```diff
@@ -243,7 +245,32 @@ def __init__(
         self.executor_manager = executor_manager
         self._state_update_delegate = ToDbUpdateManager(client, environment)
 
-    def reset(self) -> None:
+        self._sched = Scheduler("Resource scheduler")
+        self.resource_periodic_repairs: dict[ResourceIdStr, ScheduledTask] = {}
+        self._repair_timer: Union[int, str] = agent_config.agent_repair_interval.get()
```
wouterdb, line 250: This needs some docs to explain the subtle difference in type here.

src/inmanta/deploy/scheduler.py
```diff
@@ -464,9 +552,10 @@ async def _new_version(
             added_requires: dict[ResourceIdStr, Set[ResourceIdStr]] = {}
             dropped_requires: dict[ResourceIdStr, Set[ResourceIdStr]] = {}
             for resource, details in up_to_date_resources.items():
-                self._state.add_up_to_date_resource(resource, details)
+                self._state.add_up_to_date_resource(resource, details)  # Removes from the dirty set
 
             for resource, details in resources.items():
+                await self.create_periodic_repair_and_deploy(resource)
```
wouterdb, line 558: This is not correct right?

This timer is wrt to when the resource as first seen, not wrt to the last time it was deployed?

I would expect a non-periodic timer that reset on every deploy?

wouterdb (changes requested):

src/inmanta/deploy/scheduler.py
```diff
@@ -243,7 +245,32 @@ def __init__(
         self.executor_manager = executor_manager
         self._state_update_delegate = ToDbUpdateManager(client, environment)
 
-    def reset(self) -> None:
+        self._sched = Scheduler("Resource scheduler")
+        self.resource_periodic_repairs: dict[ResourceIdStr, ScheduledTask] = {}
+        self._repair_timer: Union[int, str] = agent_config.agent_repair_interval.get()
+
+        self.global_periodic_repair_task: ScheduledTask | None = None
```
wouterdb, line 252: The globals took me a while to understand, cool mechanism

src/inmanta/deploy/scheduler.py
wouterdb: Correction: I would start timer only for new resources, not for new versions.

src/inmanta/deploy/scheduler.py
```diff
@@ -593,6 +682,74 @@ async def is_agent_running(self, name: str) -> bool:
...
+                    reason=f"Individual repair triggered for resource {resource} because last "
+                    f"repair happened more than {self._repair_timer}s ago.",
+                    priority=TaskPriority.INTERVAL_REPAIR,
+                    deploying=self._deploying_latest,
+                )
+
+        assert isinstance(self._repair_timer, int)
+        interval_schedule: IntervalSchedule = IntervalSchedule(
+            interval=float(self._repair_timer), initial_delay=self._repair_timer
+        )
```
wouterdb, line 708: Does it make sense to scheduler more then one deploy? Don't we need single shot tasks?

(question, not suggestion)

src/inmanta/deploy/scheduler.py
```diff
@@ -651,37 +808,107 @@ async def report_resource_state(
...
                 state.deployment_result = deployment_result
                 self._work.finished_deploy(resource)
+
+                # Set of dependant resources that are interested in deploys of the current resource
+                concerned_resources: set[ResourceIdStr] = set()
+                provides: Set[ResourceIdStr]
                 if deployment_result is DeploymentResult.DEPLOYED:
                     self._state.dirty.discard(resource)
+
+                    if previous_deployment_result != DeploymentResult.DEPLOYED:
```
wouterdb, line 823: These changes are mostly unrelated to the rest, I think?

src/inmanta/deploy/tasks.py
```diff
@@ -121,9 +121,10 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
                 # Stale resource, can simply be dropped.
                 return
 
+            assert isinstance(task_manager, scheduler.ResourceScheduler)  # Make mypy happy
```
wouterdb, line 124: You should extend the interface class, instead of the isinstance.

src/inmanta/deploy/tasks.py
```diff
@@ -232,6 +234,7 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
                     ),
                     deployment_result=scheduler_deployment_result,
                 )
+                await task_manager.create_periodic_repair_and_deploy(executor_resource_details.rid)
```
wouterdb, line 237: I wonder if this should be part of `report_resource_state`

wouterdb (commented):
I will continue reviewing tomorrow

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: Nice catch, this was an artifact from early implementation.

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: > Does it make sense to scheduler more then one deploy?

What do you mean here ?

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: I'm not sure anymore, this is what was proposed [here (last paragraph)](https://github.com/inmanta/inmanta-core/pull/8292#discussion_r1827968749).

src/inmanta/deploy/scheduler.py
wouterdb: Currently, the interval timer keeps firing every X seconds . 

So let's say, every 10 s

timeline:
1. t=0, timer fires, resource goes in queue, 
2. t=10, timer fires, resource goes in queue again 
3. t=15, resources is de-queued, removes timer

So if the queue is full, we keep on re-queueing, but it is only the first one that has any relevance.

src/inmanta/deploy/scheduler.py
```diff
@@ -593,6 +695,74 @@ async def is_agent_running(self, name: str) -> bool:
...
+        assert isinstance(self._deploy_timer, int)
+        interval_schedule: IntervalSchedule = IntervalSchedule(
+            interval=float(self._deploy_timer), initial_delay=self._deploy_timer
+        )
+        task = self._sched.add_action(_deploy, interval_schedule)
+
+        # Omitting this yields mypy error:
+        # Returning Any from function declared to return "ScheduledTask | None"  [no-any-return]
+        assert task is None or isinstance(task, ScheduledTask)
+        return task
```
wouterdb, line 764: I wonder if we really want to run these timer for resources that are not in the dirty set. 

if we have, let's say 50k resources and a 10s deploy interval and an empty dirty set, we still process 5k timers per second for no reason.  (these numbers are high, but not crazy either)

sanderr (commented):
Holding off on review as discussed.

src/inmanta/deploy/scheduler.py
```diff
@@ -243,6 +244,8 @@ def __init__(
         self.executor_manager = executor_manager
         self._state_update_delegate = ToDbUpdateManager(client, environment)
 
+        self._timer_manager = inmanta.deploy.timers.TimerManager(self)
```
Hugo-Inmanta, line 247: I don't like passing `self` here. Only do this to be able to schedule global repair/deploys from the timer manager. Is there a cleaner alternative ?

src/inmanta/deploy/persistence.py
```diff
@@ -43,12 +43,14 @@ class StateUpdateManager(abc.ABC):
     """
 
     @abc.abstractmethod
-    async def send_in_progress(self, action_id: UUID, resource_id: ResourceVersionIdStr) -> None:
+    async def send_in_progress(
+        self, action_id: UUID, resource_version_id: ResourceVersionIdStr, resource_id: ResourceIdStr
```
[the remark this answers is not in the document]
Hugo-Inmanta: I agree, I did this in preparation for this `# FIXME: get rid of version in the id` but this is actually for another PR.
I'll parse the `rid` from the `rvid` for now

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        Make sure the underlying resource is marked for execution
+
+        :param periodic_deploy_interval: Per-resource deploy interval, assumed to be a positive int or None
+        :param periodic_repair_interval: Per-resource repair interval, assumed to be a positive int or None
+        :param is_dirty: Last known state of the resource at the timer install request time.
+        :param action_function: The function to execute
+        """
+        if self.is_installed:
+            return
```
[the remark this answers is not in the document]
Hugo-Inmanta: This was indeed incorrect. Made this api idempotent and take into consideration new shorter timers.

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def reset(self) -> None:
+        self.stop()
+
+        self.global_periodic_repair_task = None
+        self.global_periodic_deploy_task = None
+        self.periodic_repair_interval = None
+        self.periodic_deploy_interval = None
+
+        self._initialize()
```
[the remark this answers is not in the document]
Hugo-Inmanta: Moved it out and explicitly call initialize from the scheduler's initialize

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: See @sanderr 's answer https://inmanta.slack.com/archives/CKRF0C8R3/p1732551457970899?thread_ts=1732551090.789439&cid=CKRF0C8R3

src/inmanta/deploy/scheduler.py
```diff
@@ -675,37 +713,89 @@ async def report_resource_state(
...
+                    if previous_deployment_result != DeploymentResult.DEPLOYED:
+                        # This resource went from not deployed to deployed
+                        # we have to inform its dependants regardless of event propagation
+                        # so that we can try to re-deploy them in case this resource's previous
+                        # failure was the reason they were skipped
+
+                        # FIXME: stop informing resource skipped because of
+                        #  a custom handler skip via this mechanism
+
+                        dependant_resources: set[ResourceIdStr] = set()
```
wouterdb, line 739: I would find it easier to understand if you would update `concerned_resources` immediately instead of first building up `dependant_resources` and then copy it into `concerned_resources`

src/inmanta/deploy/scheduler.py
```diff
@@ -675,37 +713,89 @@ async def report_resource_state(
...
+                        # Also add back dependents to the dirty set
+                        # For now we add all dependents
+
+                        dependant_resources = {
+                            dependant for dependant in provides if self._state.resources.get(dependant, None) is not None
+                        }
+                        self._state.dirty.update(dependant_resources)
+                        for dependant in dependant_resources:
+                            # No point in re-trying dependants since this resource was skipped
+                            self._timer_manager.remove_resource(dependant)
```
wouterdb, line 771: These are not added to concerned_resources?

src/inmanta/deploy/scheduler.py
```diff
@@ -744,6 +834,7 @@ def get_types_for_agent(self, agent: str) -> Collection[ResourceType]:
         return list(self._state.types_per_agent[agent])
 
     async def send_in_progress(self, action_id: UUID, resource_id: ResourceVersionIdStr) -> None:
+        self._timer_manager.remove_resource(Id.parse_id(resource_id).resource_str())
```
wouterdb, line 837: I would put this in 'get_resource_intent_for_deploy' .

`get_resource_intent_for_deploy` is the deploy start of the scheduler, this one is the external update to the DB

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,219 @@
...
+        else:
+            next_execute_time = periodic_repair_interval
+
+        if next_execute_time == sys.maxsize:
+            return
+
+        # We have to schedule the next deploy sooner than previously though:
+        # Cancel previous schedule and re-schedule
+        if next_execute_time < self.time_to_next_deploy:
+            self.cancel_next_schedule()
```
wouterdb, line 79: Any time difference leads to a re-schedule, I think? 

(i.e. if we do a deploy, we move all timers to a later place)

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        if self.next_schedule_handle is not None:
+            self.next_schedule_handle.cancel()
+
+    def cancel_repair_task(self) -> None:
+        """
+        Cancel the underlying repair task itself
+        """
+        if self.repair_handle is not None:
+            self.repair_handle.cancel()
```
wouterdb, line 103: Should we? Do the tasks support this?

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,219 @@
...
+        for timer in self.resource_timers.values():
+            timer.cancel_next_schedule()
+
+        for task in [self.global_periodic_repair_task, self.global_periodic_deploy_task]:
+            if task:
+                self._sched.remove(task)
+
+    def join(self) -> None:
+        for timer in self.resource_timers.values():
+            timer.cancel_repair_task()
```
wouterdb, line 142: join is expected to wait for the cancel to be done, which is not what this method does. (It is not async, so it clearly doesn't wait)

src/inmanta/deploy/scheduler.py
```diff
@@ -606,6 +626,21 @@ async def is_agent_running(self, name: str) -> bool:
         """
         return name in self._workers and self._workers[name].is_running()
 
+    async def repair_resource(self, resource: ResourceIdStr, repair_timer: int) -> None:
+        async with self._scheduler_lock:
```
wouterdb, line 630: it is somewhat interesting that the only async thing here is the lock.

I wonder, can we safely do a deploy, outside the lock, because we know we are sitting on the IOloop thread? 

It would simplify the timers somewhat I think

src/inmanta/deploy/scheduler.py
```diff
@@ -606,6 +626,21 @@ async def is_agent_running(self, name: str) -> bool:
         """
         return name in self._workers and self._workers[name].is_running()
 
+    async def repair_resource(self, resource: ResourceIdStr, repair_timer: int) -> None:
```
wouterdb, line 629: I would move this to the timer module as well, it is a bit hard to read when spread out. (imho)

src/inmanta/deploy/scheduler.py
wouterdb: I would add this explanation to the method definition. 

This method does a lot, and it has no docstring

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,219 @@
...
+            await self._resource_scheduler.repair(
+                reason=f"Global repair triggered because of set cron {cron_expression}",
+                priority=TaskPriority.INTERVAL_REPAIR,
+            )
+
+        task = self._sched.add_action(_action, cron_schedule)
+        assert isinstance(task, ScheduledTask)
+        return task
+
+    def update_resource(self, resource: ResourceIdStr, dirty: bool = False) -> None:
```
wouterdb, line 200: I would move this closer to the `ensure_timer`

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,219 @@
...
+        # Create if it is not known yet
+        if resource not in self.resource_timers:
+            self.resource_timers[resource] = ResourceTimer(resource)
+
+        # Check if timer needs updating
+        self.resource_timers[resource].ensure_timer(
+            periodic_deploy_interval=self.periodic_deploy_interval,
+            periodic_repair_interval=self.periodic_repair_interval,
+            is_dirty=dirty,
+            action_function=self._resource_scheduler.repair_resource,
```
wouterdb, line 210: if this is always the same method that is passed in, I would just call it there, not pass in the reference

src/inmanta/deploy/timers.py
wouterdb, line 200: I would make dirty a required argument, too important to leave defaulted

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,219 @@
...
+
+        # A deploy is already scheduled for this resource
+        if self.next_schedule_handle is not None:
+            return
+
+        def _create_repair_task() -> None:
+            self.repair_handle = asyncio.create_task(action_function(self.resource, next_execute_time))
+
+        self.time_to_next_deploy = next_execute_time
+        self.next_schedule_handle = asyncio.get_running_loop().call_later(next_execute_time, _create_repair_task)
```
wouterdb, line 89: shouldn't we have a different priorty/reason, based on what kind of trigger we use?

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: No, since in this case the dependency just finished its deploy and the result was `SKIPPED`, so I don't think we should deploy the dependants ?
Unless they are listening for event propagation, but this is handled below.

src/inmanta/deploy/scheduler.py
sanderr: > it has no docstring

It does in the ABC

src/inmanta/deploy/scheduler.py
sanderr: I agree in concept. This method did grow a lot more than I anticipated, so I'll need a more in-depth review to provide a better response.

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: Moved to `get_resource_intent_for_deploy`

src/inmanta/deploy/timers.py
Hugo-Inmanta: reworked to add proper repair/deploy methods with their respective priorities/reasons to the `ResourceTimer` class

src/inmanta/deploy/timers.py
Hugo-Inmanta: Refactored this a bit (see other comment)

src/inmanta/deploy/timers.py
Hugo-Inmanta: Refactored this a bit to the following:

Clarify in the `ResourceTimer` class between:
```
        # "Inner" asyncio task responsible for queuing a Deploy task
        self.trigger_deploy: asyncio.Task[None] | None = None
        # "Outer" handle to the soon-to-be-called trigger_deploy
        self.next_schedule_handle: asyncio.TimerHandle | None = None
```
        
- When `timer_manager` is stopped: 
  - cancel `next_schedule_handle` for each timer
- When `timer_manager` is joined:
  - await the `trigger_deploy` if any for each timer (i.e. wait for the enqueuing to be done)

src/inmanta/deploy/timers.py
Hugo-Inmanta: What do you mean ? They are in 2 distinct classes at the moment. ( `ResourceTimer.ensure_timer` and `TimerManager.update_resource` )

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: moved to timer module and split into a proper repair and deploy method

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: > I wonder, can we safely do a deploy, outside the lock, because we know we are sitting on the IOloop thread?

I don't know. Does this remark also apply to other places where we only do sync operations under the async lock ? e.g. regular client triggered [deploy](https://github.com/inmanta/inmanta-core/blob/dbf5c7ae7233a775375b50e77dc76cdb2e1bd68d/src/inmanta/deploy/scheduler.py#L310) ?

@sanderr do you know ?

src/inmanta/deploy/scheduler.py
```diff
@@ -490,9 +505,13 @@ async def _new_version(
                     elif details.attribute_hash != self._state.resources[resource].attribute_hash:
                         # The desired state has changed.
                         new_desired_state.add(resource)
+                    # Nothing need to be done for these resources' timers yet. They will be
+                    # picked up by the scheduler eventually, and only then, after deployment
+                    # action should be taken wrt timers.
                 else:
                     # It's a resource we don't know yet.
                     new_desired_state.add(resource)
+                    self._timer_manager.update_resource(resource, dirty=False)
```
wouterdb, line 514: Why is is marked as not dirty? It is new, so I would expect it to be dirty?

src/inmanta/deploy/scheduler.py
```diff
@@ -631,6 +651,8 @@ async def get_resource_intent(self, resource: ResourceIdStr) -> Optional[Resourc
             return ResourceIntent(model_version=self._state.version, details=resource_details, dependencies=None)
 
     async def get_resource_intent_for_deploy(self, resource: ResourceIdStr) -> Optional[ResourceIntent]:
```
wouterdb, line 653: I would rename this to deploy_start (unrelated to this ticket, now may not be the time, so eel free to ignore)

src/inmanta/deploy/scheduler.py
```diff
@@ -651,6 +673,9 @@ async def report_resource_state(
         if deployment_result is DeploymentResult.NEW:
```
wouterdb, line 673: I would rename this method to deploy_done (unrelated to this ticket, now may not be the time, so feel free to ignore)

src/inmanta/deploy/scheduler.py
```diff
@@ -675,37 +700,85 @@ async def report_resource_state(
...
+                        for dependant in provides:
+                            if not self._state.resources.get(dependant, None):
+                                continue
+                            if not (dependant_status := self._state.resource_state.get(dependant, None)):
+                                continue
+                            if dependant_status.blocked is BlockedStatus.NO:
+                                concerned_resources.add(dependant)
+                                # Remove timers for unblocked dependant resources because
+                                # they are marked for deployment below.
+                                self._timer_manager.remove_resource(dependant)
```
wouterdb, line 735: I wonder if we should remove them here. 

It is purely an optimization: (if it fires it won't change anything if priority is lower, if priority is higher, this remove is wrong.
 
So if we change the priorities, this optimization is a hard to find bug.

src/inmanta/deploy/scheduler.py
```diff
@@ -675,37 +700,85 @@ async def report_resource_state(
...
+                        # Also add back dependents to the dirty set
+                        # For now we add all dependents
+
+                        dependant_resources = {
+                            dependant for dependant in provides if self._state.resources.get(dependant, None) is not None
+                        }
+                        self._state.dirty.update(dependant_resources)
+                        for dependant in dependant_resources:
+                            # No point in re-trying dependants since this resource was skipped
+                            self._timer_manager.remove_resource(dependant)
```
wouterdb, line 754: This is a delicate one, wrt to @jptrindade PR, I think it is good like this for now, but we may want to add a comment linking to https://github.com/inmanta/inmanta-core/issues/8423?

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,249 @@
...
+        periodic_repair_interval: int | None,
+        is_dirty: bool,
+    ) -> None:
+        """
+        Make sure the underlying resource is marked for execution
+
+        :param periodic_deploy_interval: Per-resource deploy interval, assumed to be a positive int or None
+        :param periodic_repair_interval: Per-resource repair interval, assumed to be a positive int or None
+        :param is_dirty: Last known state of the resource at the timer install request time.
+        :param action_function: The function to execute
```
wouterdb, line 62:
```suggestion
```

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,249 @@
...
+
+        if periodic_repair_interval is None:
+            periodic_repair_interval = sys.maxsize
+        if periodic_deploy_interval is None:
+            periodic_deploy_interval = sys.maxsize
+
+        action_function: Callable[[ResourceIdStr, int], Coroutine[Any, Any, None]]
+        if is_dirty:
+            next_execute_time = min(periodic_deploy_interval, periodic_repair_interval)
+            action_function = self._trigger_deploy_for_resource
```
wouterdb, line 75:
```suggestion
            next_execute_time, action_function = min( 
                      (periodic_deploy_interval,  self._trigger_deploy_for_resource) , 
                      (periodic_repair_interval, self._trigger_repair_for_resource),  
                      key=lambda x:x[0])
```

src/inmanta/deploy/timers.py
wouterdb: What about if we go from dirty to clean by receiving an event? 

The deploy time hasn't fired, but it will and it shouldn't.

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,249 @@
...
+        else:
+            next_execute_time = periodic_repair_interval
+            action_function = self._trigger_repair_for_resource
+
+        if next_execute_time == sys.maxsize:
+            return
+
+        # A deploy is already scheduled for this resource
+        if self.next_schedule_handle is not None:
+            return
```
wouterdb, line 85: I don't think this is correct, see comment  on comment



What about if we go from dirty to clean by receiving an event?

The deploy time hasn't fired, but it will and it shouldn't.

wouterdb (changes requested):
Minor comments

src/inmanta/deploy/scheduler.py
wouterdb: it is very very specific to this code path

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: removed

src/inmanta/deploy/timers.py
Hugo-Inmanta: Updated this to cancel and reschedule if the interval changed and to early return otherwise

src/inmanta/deploy/scheduler.py
sanderr: It doesn't fit completely with how the method was intended, because I intentionally made it so it can work for e.g. a dry-run as well. But since it's not in use like that at the moment, and it's purely internal, I guess we can use the name for now.

src/inmanta/deploy/scheduler.py
```diff
@@ -490,9 +508,13 @@ async def _new_version(
                     elif details.attribute_hash != self._state.resources[resource].attribute_hash:
                         # The desired state has changed.
                         new_desired_state.add(resource)
+                    # Nothing need to be done for these resources' timers yet. They will be
+                    # picked up by the scheduler eventually, and only then, after deployment
+                    # action should be taken wrt timers.
                 else:
                     # It's a resource we don't know yet.
                     new_desired_state.add(resource)
+                    self._timer_manager.update_resource(resource, dirty=True)
```
sanderr, line 517: I don't think this is right: we're scheduling the resource *now*, so we don't need a timer for later.

sanderr (changes requested):
First batch of comments.

src/inmanta/deploy/timers.py
sanderr: I think that all transitions from / to dirty happen either when a new version comes in or when a resource finishes deploying. But I do think that overall it would be a lot simpler to reason on if whatever comes from the scheduler is considered the authority, relative to now, replacing what was previously there (even if there should never be anything there in practice).

src/inmanta/deploy/scheduler.py
sanderr: > Does this remark also apply to other places where we only do sync operations under the async lock ?

Only if **all** places would only have sync operations. i.e. if one place exists that does an `await` under the lock, then it might pass control to another async function. If that one at some point does a bunch of sync stuff that should be under the lock, it affects the integrity of the original location.

And I honestly think that that same argument applies here, doesn't it @wouterdb?

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,255 @@
...
+    Contact: code@inmanta.com
+"""
+
+import asyncio
+import logging
+import sys
+from collections.abc import Coroutine
+from typing import Any, Callable
+
+import inmanta.deploy.scheduler
```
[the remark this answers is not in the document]
sanderr: > I haven't looked a the rest of the file yet, so I may still be proven wrong.

I have by now, and as discussed in our meeting, I'm OK with it being passed as an argument, considering that the alternatives are of the same degree of ugliness. I do stand by the second comment though.

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,255 @@
...
+            self.next_schedule_handle.cancel()
+
+    async def join(self) -> None:
+        """
+        Wait for the deploy trigger task to finish
+        """
+        if self.trigger_deploy is not None:
+            await self.trigger_deploy
+
+    async def _trigger_repair_for_resource(self, resource: ResourceIdStr, timer: int) -> None:
```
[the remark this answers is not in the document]
Hugo-Inmanta: Moved to `Scheduler.trigger_deploy_for_resource`.

src/inmanta/deploy/scheduler.py
```diff
@@ -675,37 +703,83 @@ async def report_resource_state(
...
+                    priority = self._work.agent_queues.in_progress[task]
+                    self._work.deploy_with_context(
+                        concerned_resources,
+                        reason=f"Deploying because an event was received from {resource}",
+                        priority=priority,
+                        deploying=set(),
+                    )
+
+        # No matter the deployment result, schedule a re-deploy for this resource
+        self._timer_manager.update_resource(resource, dirty=is_dirty)
```
sanderr: Should also be under the lock. And perhaps we should simply read from `self._state.compliance_status` so we don't need `is_dirty`?

Hugo-Inmanta:
Separate PR to respond to setting changes: https://github.com/inmanta/inmanta-core/pull/8435

src/inmanta/deploy/scheduler.py
```diff
@@ -675,37 +703,83 @@ async def report_resource_state(
...
+
+                        # FIXME: stop informing resource skipped because of
+                        #  a custom handler skip via this mechanism
+
+                        for dependant in provides:
+                            if not self._state.resources.get(dependant, None):
+                                continue
+                            if not (dependant_status := self._state.resource_state.get(dependant, None)):
+                                continue
+                            if dependant_status.blocked is BlockedStatus.NO:
```
sanderr, line 761: This would have to become `TRANSIENT` when Joao's PR is in. Can you make sure that it gets updated, regardless of which PR is merged first? i.e. the FIXME a few lines up.

src/inmanta/deploy/scheduler.py
```diff
@@ -640,7 +665,7 @@ async def get_resource_intent_for_deploy(self, resource: ResourceIdStr) -> Optio
             self._deploying_latest.add(resource)
             return ResourceIntent(model_version=self._state.version, details=resource_details, dependencies=dependencies)
 
-    async def report_resource_state(
+    async def deploy_done(
```
sanderr, line 692: This method grew *a lot* in complexity. I think all pieces you added are indeed required, but I'd like to offer some suggestions to change the structure in order to get the complexity back to a readable level. Primarily, I think it gets complex now because some concerns are spread over the method. Concrete suggestions:

1. Make abstraction of the event propagation. Let's make it a private helper method that expects to be called under the lock. It really only needs to know the resource id, and whether this deploy recovered from a fail state (`send_recovery_events: bool`?). The method can then do the calculation of the concerned resources, and we lose a whole lot of lines in this method.
2. The current implementation is a mix of styles. This is a smaller issue than the other, and one I'm OK with to let slide (plus I think it will already improve by addressing 1), but it does feel a bit off to me that most of the method is quite functional, and then suddenly a mutable variable shows up that is updated in various unexpected places (`concerned_resources`). I think it can easily be inlined in the set comprehension for the events. Or if you don't like that, construct the two sets separately and simply `|` them.

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,295 @@
...
+        self.periodic_repair_interval = None
+        self.periodic_deploy_interval = None
+
+    def stop(self) -> None:
+        for timer in self.resource_timers.values():
+            timer.cancel()
+
+        for task in [self.global_periodic_repair_task, self.global_periodic_deploy_task]:
+            if task:
+                self._cron_scheduler.remove(task)
```
wouterdb, line 136: `self._cron_scheduler.stop()` ?

src/inmanta/deploy/scheduler.py
```diff
@@ -675,37 +703,83 @@ async def report_resource_state(
...
+
+                if concerned_resources:
+                    # do not pass deploying tasks because for event propagation we really want to start a new one,
+                    # even if the current intent is already being deployed
+                    task = Deploy(resource=resource)
+                    assert task in self._work.agent_queues.in_progress
+                    priority = self._work.agent_queues.in_progress[task]
+                    self._work.deploy_with_context(
+                        concerned_resources,
+                        reason=f"Deploying because an event was received from {resource}",
```
sanderr, line 804: Perhaps we could also make this a "recovery event" if the resource sending it went from a failed state to deployed? @wouterdb wdyt?

src/inmanta/deploy/scheduler.py
```diff
@@ -675,37 +703,83 @@ async def report_resource_state(
...
+                        # For now we add all dependents
+
+                        dependant_resources = {
+                            dependant for dependant in provides if self._state.resources.get(dependant, None) is not None
+                        }
+                        self._state.dirty.update(dependant_resources)
+                        for dependant in dependant_resources:
+                            # No point in re-trying dependants since this resource was skipped
+                            # FIXME as part of https://github.com/inmanta/inmanta-core/issues/8423
+                            self._timer_manager.remove_resource(dependant)
```
sanderr, line 782: I may be missing something (or have forgotten about a previous discussion) but I don't think this is correct / desirable. Could you give some more context as to why we'd want to do this?

src/inmanta/agent/config.py
sanderr, line 1: Nice doc updates here!

sanderr (changes requested):
Reviewed everything except the tests and the latest changes to the timers.

src/inmanta/deploy/scheduler.py
wouterdb: Just realized that requires-provides is mutable, so this is complicated as hell, comment below probably doesn't apply, you are right, lock it required 

---- old comment -> ignore --- 

Good point, so the question becomes: is any of the code below the lock affected by this.

Interestingly the  ScheduledWork is entirely sync, so it can be considered to be atomic wrt to asyncio. We never read from it (except de-queue). So data corruption is not a risk, nor is read-confusion, only re-ordering of operations.  All Operations are idempotent, associative and commutative **except** for de-queue, but that only happens in the agent worker (so the IOLoop).

We may be able to do without the lock, but that is perhaps not for now.

src/inmanta/deploy/scheduler.py
wouterdb: I like the suggestion

src/inmanta/deploy/scheduler.py
sanderr: > ---- old comment -> ignore ---

The part above or below this marker? EDIT: nevermind, the comment above references the other.

src/inmanta/deploy/timers.py
```diff
@@ -47,56 +54,39 @@ def __init__(self, resource: ResourceIdStr, scheduler: "inmanta.deploy.scheduler
...
-            # This timer is already in place
-            return
-
-        # Cancel previous schedule and re-schedule
-        self.cancel_next_schedule()
 
-        def _create_repair_task() -> None:
-            self.trigger_deploy = asyncio.create_task(action_function(self.resource, next_execute_time))
+        assert (
+            self.next_schedule_handle is None
```
sanderr, line 84: We don't seem to set it to `None` when we cancel the task.

src/inmanta/deploy/timers.py
```diff
@@ -47,56 +54,39 @@ def __init__(self, resource: ResourceIdStr, scheduler: "inmanta.deploy.scheduler
...
+        ), f"Per-resource timer set twice for resource {self.resource}, this should not happen"
 
-        self.time_to_next_deploy = next_execute_time
-        self.next_schedule_handle = asyncio.get_running_loop().call_later(next_execute_time, _create_repair_task)
+        self.next_schedule_handle = asyncio.get_running_loop().call_later(countdown, _create_repair_task)
```
sanderr, line 87: Can you remind me why we create a callback that creates a background task, rather than to immediately create a background task with `await`? I seem to recall there was a reason.

If we did, we'd have to shield the actual deploy from cancellation, but other than that it seems a lot simpler.

src/inmanta/deploy/timers.py
```diff
@@ -157,7 +117,7 @@ def __init__(self, resource_scheduler: "inmanta.deploy.scheduler.ResourceSchedul
 
         self._resource_scheduler = resource_scheduler
 
-        self._sched = Scheduler("Resource scheduler")
+        self._cron_scheduler = Scheduler("Resource scheduler")
```
sanderr, line 120:
```suggestion
        self._cron_scheduler = util.Scheduler("Resource scheduler")
```

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,295 @@
...
+            timer.cancel()
+
+        for task in [self.global_periodic_repair_task, self.global_periodic_deploy_task]:
+            if task:
+                self._cron_scheduler.remove(task)
+
+    async def join(self) -> None:
+        await asyncio.gather(*[timer.join() for timer in self.resource_timers.values()])
+
+    def _initialize(self) -> None:
```
sanderr, line 141: This is a private method but it's not called anywhere in this file?

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,295 @@
...
+        self.resource_timers[resource].set_timer(
+            countdown=countdown,
+            reason=reason,
+            priority=priority,
+        )
+
+    def remove_resource(self, resource: ResourceIdStr) -> None:
+        """
+        Stop managing the timer for the given resource and
+        cancel the associated re-deployment, if any.
```
sanderr, line 284: Didn't you propose on Slack that the timer instance would be kept in existence, only cancelling its timer? I'm fine either way tbh, this deviation just surprised me.

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,295 @@
...
+        """
+        if resource in self.resource_timers:
+            self.resource_timers[resource].cancel()
+            del self.resource_timers[resource]
+
+    def remove_resources(self, resources: Sequence[ResourceIdStr]) -> None:
```
sanderr, line 290: Probably more like a collection, since ordering has no meaning, and we'll probably get sets in practice.
```suggestion
    def remove_resources(self, resources: Collection[ResourceIdStr]) -> None:
```

src/inmanta/deploy/scheduler.py
```diff
@@ -360,6 +372,28 @@ async def get_facts(self, resource: dict[str, object]) -> None:
...
+        async with self._scheduler_lock:
+            if resource not in self._state.resource_state:
+                # The resource was removed from the model by the time this method was triggered
+                return
+
+            if self._state.resource_state[resource].blocked is BlockedStatus.YES:  # Can't deploy
+                return
+            to_deploy: set[ResourceIdStr] = {resource}
+
+            self._work.deploy_with_context(
```
sanderr, line 390: I think it would be good to also remove the timer in this method. In the current implementation the timer manager is the only user of this method, so it will be a NOOP (and admittedly a bit strange) except that it cleans up the timer instance. But if we ever call this from the outside, we want the timer to disappear.

tests/deploy/test_scheduler_agent.py
```diff
@@ -871,7 +886,6 @@ async def test_deploy_event_propagation(agent: TestAgent, make_resource_minimal)
 
     def make_resources(
         *,
-        version: int,
```
sanderr, line 874: Oh wow, I've written so many tests like this, and I never realized that the version wasn't used...

sanderr (changes requested):
Reviewed everything now. Next iteration is a final full sweep.

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
sanderr: 1. agreed
2. mostly fixed, superseded by other comments

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
sanderr: I'm leaving this one unresolved because I'm pretty sure it's linked in a few places and GitHub does not handle links to resolved discussions well.

src/inmanta/deploy/timers.py
```diff
@@ -0,0 +1,255 @@
...
+            timer.cancel_next_schedule()
+
+        for task in [self.global_periodic_repair_task, self.global_periodic_deploy_task]:
+            if task:
+                self._sched.remove(task)
+
+    async def join(self) -> None:
+        await asyncio.gather(*[timer.join() for timer in self.resource_timers.values()])
+
+    def initialize(self) -> None:
```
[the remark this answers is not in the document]
sanderr: #8435

sanderr:
Note to self for final sweep review: pay special attention to:
- https://inmanta.slack.com/archives/CKRF0C8R3/p1733237639448059?thread_ts=1732004866.610409&cid=CKRF0C8R3
- manipulation of dirty set, also see comments on  Joao's PR
- test coverage
- remove timers anywhere deploy is called.

src/inmanta/deploy/timers.py
Hugo-Inmanta: > immediately create a background task with await

What would this look like?

src/inmanta/deploy/scheduler.py
```diff
@@ -519,14 +541,14 @@ async def _new_version(
             async with self._scheduler_lock:
                 self._state.version = version
                 for resource in blocked_resources:
-                    self._state.block_resource(resource, resources[resource], is_transitive=False)
+                    self._state.block_resource(resource, resources[resource], is_transitive=False)  # Removes from the dirty set
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think this makes sense :+1:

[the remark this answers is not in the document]
sanderr: > Hmm, now that I think about it, the timer manager doesn't keep a set / dict of running timers but all timers.

Not sure if this statement is actually correct in the current implementation.

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: I think you're right. Here's the train of thought that lead to this:
I was only considering `repair` semantics and so since we know that the providing resource was just skipped, all dependents would fail their repair so we can cancel them right away. 
But when considering `deploy` semantic as well I don't think this is correct since this would destroy timers for deploy that could very well succeed.

sanderr: There's even more to it: while most dependents will skip as a result of this one skipping, that behavior is not guaranteed, so we have to give their handlers the opportunity to handle it how they see fit, i.e. we need to deploy them as planned.

To be clear, I was talking about this whole block, not just the timer part.

src/inmanta/deploy/timers.py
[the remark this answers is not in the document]
Hugo-Inmanta: Oops, didn't mean to close this.

> typing-only import

Do you mean like so ?:
```
if TYPE_CHECKING:
    from inmanta.deploy.scheduler import ResourceScheduler
```

> use inmanta.util as util instead of importing its objects?

:+1:

src/inmanta/deploy/timers.py
```diff
@@ -18,27 +18,34 @@
 
 import asyncio
 import logging
-import sys
-from collections.abc import Coroutine
-from typing import Any, Callable
+from typing import Sequence
```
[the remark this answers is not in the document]
Hugo-Inmanta: Damn it I looked at the one place in the code where we were still getting this from the typing module.
Fixed it there as well (`test_discovered_resources.py`)

src/inmanta/deploy/timers.py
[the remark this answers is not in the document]
sanderr: > Do you mean like so ?:

The `TYPE_CHECKING` part for sure. The unqualified import I don't care about too much, one way or the other.

src/inmanta/deploy/timers.py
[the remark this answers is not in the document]
sanderr: I noticed it there as well but didn't comment because it was pre-existing there. Thanks for fixing it both places.

src/inmanta/deploy/timers.py
Hugo-Inmanta: Right, I think I found this way easier to reason about at some point and then it stuck around ? But now I wonder about performance tradeoff between both approaches.

I think deleting+recreating has more overhead but then all operations on the resource_timers dict would be faster since the set is smaller ?

src/inmanta/deploy/timers.py
Hugo-Inmanta: fixed

src/inmanta/deploy/scheduler.py
```diff
@@ -476,12 +488,18 @@ async def _new_version(
             dropped_requires: dict[ResourceIdStr, Set[ResourceIdStr]] = {}
 
             for resource, details in up_to_date_resources.items():
-                self._state.add_up_to_date_resource(resource, details)
+                self._state.add_up_to_date_resource(resource, details)  # Removes from the dirty set
+                # Install timers for these resources. They are up-to-date now,
+                # but we want to make sure we periodically repair them.
+                self._timer_manager.update_resource(resource, dirty=False)
```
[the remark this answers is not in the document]
Hugo-Inmanta: I'm not super confident about this one.
Current implementation takes and releases the lock 3 times in a row. This doesn't feel right ? Should we acquire the lock once to have 'transactional' behavior instead ? But then we would keep the lock for a longer period of time at once.

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: I couldn't find a good reason for why this was added, removed.

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
sanderr: Indeed that doesn't look right. What I meant was to move this block of code further down to where we already hold the lock, not to acquire the lock here. But I think you should get that automatically once you implement [this comment](https://github.com/inmanta/inmanta-core/pull/8292#discussion_r1867507539).

src/inmanta/deploy/timers.py
sanderr: > all operations on the resource_timers dict would be faster since the set is smaller ?

I think that direct operations should have very little performance difference (could be wrong). Operations that only act on the *running* timers are a different matter, but we could easily keep a set of those.

That said, I think it would be OK to leave it as is, I'm not sure that the performance overhead would actually be significant. But I'd quickly run it by @wouterdb since it does deviate from what was concluded between the two of you on Slack.

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
Hugo-Inmanta: Right, of course :facepalm:

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: 1. refactored this into the `_propagate_events` method
2. Couldn't inline it (I think) because we want to add `concerned_resources` to the dirty set but not event listeners but still mark both sets for re-deployment.

Hugo-Inmanta:
@sanderr I added the bulk methods for the TimerManager add/remove resources
But I didn't add the capabilities to track "active" timers yet. The details + what we would gain are a bit blurry still. Do you think it would be ok to add this in a separate PR ?

sanderr:
> Do you think it would be ok to add this in a separate PR ?

Sounds good but we shouldn't lose track of it.

src/inmanta/deploy/scheduler.py
```diff
@@ -360,6 +374,28 @@ async def get_facts(self, resource: dict[str, object]) -> None:
...
+        """
+        async with self._scheduler_lock:
+            if resource not in self._state.resource_state:
+                # The resource was removed from the model by the time this method was triggered
+                return
+
+            if self._state.resource_state[resource].blocked is BlockedStatus.YES:  # Can't deploy
+                return
+            to_deploy: set[ResourceIdStr] = {resource}
+            self._timer_manager.remove_resource(resource)
```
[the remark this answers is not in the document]
sanderr: Oof, I hadn't considered priorities at all when it comes to the timers. And I think it's broader than this:
1. We only start one timer in the current implementation. I actually think this is fine if we simply take the most urgent priority (smallest value) and the smallest interval.
2. We currently assume that if a resource is scheduled we don't need a timer for it. But then we also will never bump to timer priority. In the current implementation the timers have the least priority, so in practice it's not too bad though.

So I think the issue you raise is OK, but number 2 is more of a grey area.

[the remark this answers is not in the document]
sanderr: I'm leaning towards considering priority bumping for timers out of scope. A deploy / repair interval declares that we'll schedule the resource for deploy. The priority is secondary. If a resource is already scheduled, it's scheduled.

How do you feel about that?

Hugo-Inmanta:
> > Do you think it would be ok to add this in a separate PR ?
> 
> Sounds good but we shouldn't lose track of it.

Ticketized here but it will need some refining: https://github.com/inmanta/inmanta-core/issues/8443

sanderr:
> Ticketized here

I'd honestly take it up together with the other timer manager refinement ticket (for the setting update).

sanderr:
Note to self: make sure the following race condition is covered. Suppose `a.requires = b`, then consider the following timeline:

1. a fails
2. b skips for dependencies
3. a gets picked up for deploy, sent to the executor
4. while a is still running, a new version for a is exported (does this unblock the resource?)
5. a (old version) deploys successfully
6. a's status is updated (should its blocked status be updated as well?) but no events are sent
7. a (new version) deploys successfully. Should send recovery events now.

src/inmanta/deploy/scheduler.py
```diff
@@ -697,15 +705,84 @@ async def deploy_done(
         status: ComplianceStatus,
         deployment_result: Optional[DeploymentResult] = None,
     ) -> None:
+
+        def _propagate_events(details: ResourceDetails, send_recovery_events: bool) -> None:
```
sanderr, line 709: Note to self: review this implementation

Hugo-Inmanta:
Superseded by https://github.com/inmanta/inmanta-core/pull/8450

PR #8293, opened 2024-10-30
Issue/pool reporting

### LINKED ISSUE #8228: Update pool metrics collector to also take into account the pool of the schedulers
We collect metrics on pool exhaustion and pool usage. 

We should make sure to add the pools of the schedulers as well 

Notify the FE as well and check the dashboard

wouterdb:
# Description

Update database statistics reporting 
1. towards the logs
2. towards the status enpoint
3. towards influxdb

closes #8228 

# TODO post merge:
1. [x]  inform FE that we broke the API
2. [ ] update grafana dashboard on grafana hub

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/8228-database_pool_report.yml
+++ b/changelogs/unreleased/8228-database_pool_report.yml
@@ -1,0 +1,6 @@
+description: Update database pool status reporting for the new scheduler
+issue-nr: 8228
+change-type: minor
+destination-branches: [master]
+sections:
+  minor-improvement: "{{description}}"
--- a/mypy-baseline.txt
+++ b/mypy-baseline.txt
@@ -101,7 +101,7 @@ src/inmanta/ast/type.py:0: error: No overload variant of "reversed" matches argu
 src/inmanta/ast/type.py:0: note: Possible overload variants:
 src/inmanta/ast/type.py:0: note: def [_T] __new__(cls, Reversible[_T], /) -> reversed[_T]
 src/inmanta/ast/type.py:0: note: def [_T] __new__(cls, SupportsLenAndGetItem[_T], /) -> reversed[_T]
-src/inmanta/server/services/databaseservice.py:0: error: Incompatible return value type (got "dict[str, object]", expected "dict[str, BaseModel | UUID | bool | int | float | datetime | str | Sequence[BaseModel | UUID | bool | int | float | datetime | str] | Mapping[str, BaseModel | UUID | bool | int | float | datetime | str]]")  [return-value]
+src/inmanta/server/services/metricservice.py:0: error: Incompatible types in assignment (expression has type "float", variable has type "int")  [assignment]
 src/inmanta/protocol/decorators.py:0: error: "FuncT" has no attribute "__protocol_method__"  [attr-defined]
 src/inmanta/protocol/decorators.py:0: error: "FuncT" has no attribute "__protocol_mapping__"  [attr-defined]
 src/inmanta/protocol/decorators.py:0: error: "FuncT" has no attribute "__api_version__"  [attr-defined]
@@ -156,6 +156,7 @@ src/inmanta/ast/attribute.py:0: error: Incompatible types in assignment (express
 src/inmanta/ast/attribute.py:0: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
 src/inmanta/ast/attribute.py:0: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
 src/inmanta/util/__init__.py:0: error: Argument "key" to "sorted" has incompatible type "Callable[[T], S]"; expected "Callable[[T], SupportsDunderLT[Any] | SupportsDunderGT[Any]]"  [arg-type]
+src/inmanta/util/__init__.py:0: error: Returning Any from function declared to return "ScheduledTask | None"  [no-any-return]
 src/inmanta/util/__init__.py:0: error: Returning Any from function declared to return "bool"  [no-any-return]
 src/inmanta/util/__init__.py:0: error: Incompatible return value type (got "bool | Awaitable[bool]", expected "bool")  [return-value]
 src/inmanta/resources.py:0: error: Function is missing a type annotation  [no-untyped-def]
@@ -226,6 +227,7 @@ src/inmanta/protocol/methods_v2.py:0: error: Missing return statement  [empty-bo
 src/inmanta/protocol/methods_v2.py:0: error: Module "inmanta.protocol.methods" does not explicitly export attribute "ArgOption"  [attr-defined]
 src/inmanta/protocol/methods_v2.py:0: error: Missing return statement  [empty-body]
 src/inmanta/protocol/methods_v2.py:0: error: Module "inmanta.protocol.methods" does not explicitly export attribute "ArgOption"  [attr-defined]
+src/inmanta/protocol/methods_v2.py:0: error: Missing return statement  [empty-body]
 src/inmanta/protocol/methods_v2.py:0: error: Missing return statement  [empty-body]
 src/inmanta/protocol/methods_v2.py:0: error: Missing return statement  [empty-body]
 src/inmanta/protocol/methods_v2.py:0: error: Missing return statement  [empty-body]
@@ -1101,6 +1103,7 @@ src/inmanta/server/services/orchestrationservice.py:0: error: Incompatible types
 src/inmanta/server/services/orchestrationservice.py:0: error: Incompatible types in assignment (expression has type "set[str | None]", variable has type "set[str]")  [assignment]
 src/inmanta/server/services/orchestrationservice.py:0: error: Argument "agents" to "_trigger_auto_deploy" of "OrchestrationService" has incompatible type "Collection[str]"; expected "Sequence[str] | None"  [arg-type]
 src/inmanta/server/services/orchestrationservice.py:0: error: Incompatible return value type (got "ReturnValueWithMeta[Sequence[DesiredStateVersion]]", expected "ReturnValue[list[DesiredStateVersion]]")  [return-value]
+src/inmanta/server/agentmanager.py:0: error: Value of type "dict[str, Any] | None" is not indexable  [index]
 src/inmanta/server/agentmanager.py:0: error: Argument 1 to "ensure_agent_registered" of "AgentManager" has incompatible type "Environment | None"; expected "Environment"  [arg-type]
 src/inmanta/agent/forking_executor.py:0: error: Argument 1 of "connection_made" is incompatible with supertype "BaseProtocol"; supertype defines the argument type as "BaseTransport"  [override]
 src/inmanta/agent/forking_executor.py:0: note: This violates the Liskov substitution principle
@@ -1115,7 +1118,6 @@ src/inmanta/server/services/paramservice.py:0: error: Unsupported operand types 
 src/inmanta/server/services/paramservice.py:0: note: Left operand is of type "datetime | None"
 src/inmanta/server/services/paramservice.py:0: error: Unsupported operand types for > ("timedelta" and "datetime")  [operator]
 src/inmanta/server/services/paramservice.py:0: note: Left operand is of type "datetime | timedelta"
-src/inmanta/server/services/metricservice.py:0: error: Incompatible types in assignment (expression has type "float", variable has type "int")  [assignment]
 src/inmanta/server/services/environment_metrics_service.py:0: error: Function is missing a return type annotation  [no-untyped-def]
 src/inmanta/server/services/environment_metrics_service.py:0: error: "Sequence[EnvironmentMetricsGauge]" has no attribute "append"  [attr-defined]
 src/inmanta/server/services/environment_metrics_service.py:0: error: Function is missing a return type annotation  [no-untyped-def]
--- a/mypy-baseline.txt
+++ b/mypy-baseline.txt
@@ -146,6 +146,10 @@ src/inmanta/execute/runtime.py:0: error: Argument 2 to "assign" of "AssignableNo
 src/inmanta/execute/runtime.py:0: error: Argument 1 to "AttributeException" has incompatible type "Instance"; expected "Locatable"  [arg-type]
 src/inmanta/execute/runtime.py:0: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
 src/inmanta/execute/runtime.py:0: error: Argument 1 to "RuntimeException" has incompatible type "Instance"; expected "Locatable | None"  [arg-type]
+src/inmanta/data/model.py:0: error: Return type "str" of "__getstate__" incompatible with return type "dict[Any, Any]" in supertype "BaseModel"  [override]
+src/inmanta/data/model.py:0: error: Argument 1 of "__setstate__" is incompatible with supertype "BaseModel"; supertype defines the argument type as "dict[Any, Any]"  [override]
+src/inmanta/data/model.py:0: note: This violates the Liskov substitution principle
+src/inmanta/data/model.py:0: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
 src/inmanta/data/model.py:0: error: Function is missing a type annotation  [no-untyped-def]
 src/inmanta/data/model.py:0: error: Unused "type: ignore" comment, use narrower [prop-decorator] instead of [misc] code  [unused-ignore]
 src/inmanta/ast/attribute.py:0: error: Incompatible types in assignment (expression has type "property", variable has type "Type")  [assignment]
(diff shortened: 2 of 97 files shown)


mypy-baseline.txt
```diff
@@ -101,7 +101,7 @@ src/inmanta/ast/type.py:0: error: No overload variant of "reversed" matches argu
 src/inmanta/ast/type.py:0: note: Possible overload variants:
 src/inmanta/ast/type.py:0: note: def [_T] __new__(cls, Reversible[_T], /) -> reversed[_T]
 src/inmanta/ast/type.py:0: note: def [_T] __new__(cls, SupportsLenAndGetItem[_T], /) -> reversed[_T]
-src/inmanta/server/services/databaseservice.py:0: error: Incompatible return value type (got "dict[str, object]", expected "dict[str, BaseModel | UUID | bool | int | float | datetime | str | Sequence[BaseModel | UUID | bool | int | float | datetime | str] | Mapping[str, BaseModel | UUID | bool | int | float | datetime | str]]")  [return-value]
+src/inmanta/server/services/metricservice.py:0: error: Incompatible types in assignment (expression has type "float", variable has type "int")  [assignment]
```
wouterdb, line 104: pre-existing issue, moved

mypy-baseline.txt
```diff
@@ -1101,6 +1103,7 @@ src/inmanta/server/services/orchestrationservice.py:0: error: Incompatible types
 src/inmanta/server/services/orchestrationservice.py:0: error: Incompatible types in assignment (expression has type "set[str | None]", variable has type "set[str]")  [assignment]
 src/inmanta/server/services/orchestrationservice.py:0: error: Argument "agents" to "_trigger_auto_deploy" of "OrchestrationService" has incompatible type "Collection[str]"; expected "Sequence[str] | None"  [arg-type]
 src/inmanta/server/services/orchestrationservice.py:0: error: Incompatible return value type (got "ReturnValueWithMeta[Sequence[DesiredStateVersion]]", expected "ReturnValue[list[DesiredStateVersion]]")  [return-value]
+src/inmanta/server/agentmanager.py:0: error: Value of type "dict[str, Any] | None" is not indexable  [index]
```
wouterdb, line 1106: decoding of api results, difficult to avoid

src/inmanta/agent/agent_new.py
```diff
@@ -84,6 +86,20 @@ def __init__(
...
+        # Make mypy happy
+        assert data.Resource._connection_pool is not None
+        self._db_monitor = DatabaseMonitor(
+            data.Resource._connection_pool,
+            opt.db_name.get(),
+            opt.db_host.get(),
+            "scheduler",
+            str(self.environment),
+        )
+        self._db_monitor.start()
```
wouterdb, line 99: Scheduler gets DB monitor as well

src/inmanta/agent/agent_new.py
```diff
@@ -323,6 +339,20 @@ async def get_facts(self, env: uuid.UUID, agent: str, resource: dict[str, Any])
...
+            return DataBaseReport(
+                connected=False,
+                database="",
+                host="",
+                max_pool=0,
+                open_connections=0,
+                free_connections=0,
+                pool_exhaustion_count=0,
+            )
+        return await self._db_monitor.get_status()
```
wouterdb, line 354: new status endpoint on the scheduler to get database status

src/inmanta/app.py
```diff
@@ -156,6 +157,9 @@ async def start() -> None:
             connection_pool_max_size=agent_config.scheduler_db_connection_pool_max_size.get(),
             connection_timeout=agent_config.scheduler_db_connection_timeout.get(),
         )
+        # also report metrics if this is relevant
+        metrics_reporter = MetricsService()
+        metrics_reporter.start_metric_reporters()
```
wouterdb, line 162: scheduler will now also report to influxdb

src/inmanta/data/model.py
```diff
@@ -89,7 +89,7 @@ class SliceStatus(BaseModel):
     """
 
     name: str
-    status: dict[str, ArgumentTypes]
+    status: Mapping[str, ArgumentTypes | Mapping[str, ArgumentTypes]]
```
wouterdb, line 92: This breaks the FE, but allow nicer presentation

src/inmanta/server/__init__.py
```diff
@@ -19,7 +19,7 @@
 # flake8: noqa: F401
 
 SLICE_SERVER = "core.server"
-SLICE_AGENT_MANAGER = "core.agentmanager"
+SLICE_AGENT_MANAGER = "core.scheduler_manager"
```
wouterdb, line 22: rename for FE, hopefully OK

src/inmanta/server/agentmanager.py
```diff
@@ -168,12 +169,63 @@ def __init__(self, closesessionsonstart: bool = True, fact_back_off: Optional[in
 
         self.closesessionsonstart: bool = closesessionsonstart
 
-    async def get_status(self) -> dict[str, ArgumentTypes]:
-        return {
+    async def get_status(self) -> Mapping[str, ArgumentTypes | Mapping[str, ArgumentTypes]]:
```
wouterdb, line 172: This method is very tricky, as it has to return withing the timeout of the status endpoint.

I chose to make it report what it has, even if that makes the format somewhat inconsistent. 
(e.g. if the DB is down, the environments will be reported by UUID instead of name)

src/inmanta/server/services/databaseservice.py
```diff
@@ -17,32 +17,144 @@
...
 from inmanta.types import ArgumentTypes
+from inmanta.util import Scheduler
 
 LOGGER = logging.getLogger(__name__)
 
+if typing.TYPE_CHECKING:
+    from inmanta.server import agentmanager
+
+
+class DatabaseMonitor:
```
wouterdb, line 41: Factored out!

src/inmanta/server/services/databaseservice.py
```diff
@@ -17,32 +17,144 @@
...
+            host=self.db_host,
+            max_pool=self._pool.get_max_size(),
+            open_connections=self._pool.get_size(),
+            free_connections=self._pool.get_idle_size(),
+            pool_exhaustion_count=self._db_pool_watcher._exhausted_pool_events_count,
+        )
+
+    def start_monitor(self) -> None:
+        """Attach to monitoring system"""
+        suffix = f",component={self.component}"
```
wouterdb, line 83: nasty way of adding tags, but it works

tests/test_influxdbreporting.py
```diff
@@ -182,20 +184,36 @@ async def test_available_metrics(server):
...
-    assert "db.free_connections" in metrics
+    assert metrics["db.connected,component=server"]["value"]
+    assert "db.max_pool,component=server" in metrics
+    assert "db.open_connections,component=server" in metrics
+    assert "db.free_connections,component=server" in metrics
     assert "self.spec.cpu" in metrics
 
     # ensure it doesn't crash when the server is down
     await server.stop()
     metrics = global_registry().dump_metrics()
```
Hugo-Inmanta, line 195:
```suggestion

```

stubs/pyformance/__init__.pyi
```diff
@@ -5,6 +5,9 @@ from pyformance.meters import Gauge
 
 
 class MetricsRegistry:
+
+    _gauges: dict[str, Gauge]
+
```
Hugo-Inmanta, line 10: What does this change bring ? I'm not familiar with `.pyi` files

src/inmanta/server/agentmanager.py
```diff
@@ -168,12 +169,66 @@ def __init__(self, closesessionsonstart: bool = True, fact_back_off: Optional[in
...
+        try:
+            # This can timeout, but not likely
+            env_mapping_result = await env_mapping
+            uuid_to_name = {env.id: env.name for env in env_mapping_result}
+        except TimeoutError:
+            # default to uuid's
+            uuid_to_name = {}
+
+        # Filter out timeouts
+        total: DataBaseReport = reduce(lambda x, y: x + y, (x[1] for x in results if not isinstance(x, BaseException)))
```
Hugo-Inmanta, line 205:
```suggestion
        total: DataBaseReport | None = reduce(lambda x, y: x + y, (x[1] for x in results if not isinstance(x, BaseException)))
```

Hugo-Inmanta (changes requested):

src/inmanta/data/model.py
```diff
@@ -859,3 +859,31 @@ def has_source(self) -> bool:
...
+
+    def __add__(self, other: "DataBaseReport") -> "DataBaseReport":
+        if not isinstance(other, DataBaseReport):
+            return NotImplemented
+        if other.database != self.database:
+            return NotImplemented
+        if other.host != self.host:
+            return NotImplemented
+        return DataBaseReport(
+            connected=self.connected,
```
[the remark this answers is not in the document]
wouterdb: good question, I should probably `and` them

stubs/pyformance/__init__.pyi
wouterdb: it is typing information: pyformance has no mypy type annotation, in this file, we add them.
So I made this field known to mypy by adding it here.

src/inmanta/server/agentmanager.py
wouterdb:
```suggestion
        total: DataBaseReport | None = reduce(lambda x, y: x + y, (x[1] for x in results if not isinstance(x, BaseException)), None)
```

wouterdb: good point, but not entirely correct either: 

```
>>> reduce(lambda x,y: x+y, [])
TypeError: reduce() of empty iterable with no initial value

```

src/inmanta/server/agentmanager.py
```diff
@@ -251,6 +252,12 @@ async def start(self) -> None:
             await self._expire_all_sessions_in_db()
 
         self.add_background_task(self._process_session_listener_actions())
+        # Schedule cleanup agentprocess and agentinstance tables
+        agent_process_purge_interval = opt.agent_process_purge_interval.get()
+        if agent_process_purge_interval > 0:
+            self.schedule(
+                self._purge_agent_processes, interval=agent_process_purge_interval, initial_delay=0, cancel_on_stop=False
+            )
```
Hugo-Inmanta, line 260: Why was this moved ?

Hugo-Inmanta (changes requested):

src/inmanta/server/agentmanager.py
wouterdb: Because I think it should have been in the agent manager in the first place: it is related to management of agents, so it should be in the agent manager.

Hugo-Inmanta (approved):

changelogs/unreleased/8228-database_pool_report.yml
```diff
@@ -0,0 +1,6 @@
+description: Update database pool status reporting for the new scheduler
+issue-nr: 8228
+change-type: minor
+destination-branches: [master]
+sections:
+  minor-improvement: "{{description}}"
```
sanderr, line 6: Does this deserve a change entry, considering that the new scheduler has not been publicly released? Or has it?

src/inmanta/server/agentmanager.py
```diff
@@ -168,12 +169,67 @@ def __init__(self, closesessionsonstart: bool = True, fact_back_off: Optional[in
...
             "sessions": len(self.sessions),
         }
 
+        # Try to get more info from scheduler, but make sure not to timeout
+        schedulers = self.get_all_schedulers()
+        deadline = 0.9 * Server.GET_SERVER_STATUS_TIMEOUT
+
+        async def get_report(env: uuid.UUID, session: protocol.Session) -> tuple[uuid.UUID, DataBaseReport]:
+            result = await asyncio.wait_for(session.client.get_db_status(), deadline)
+            assert result.code == 200
```
sanderr, line 186: Is assert appropriate here? It can realistically fail, can't it?

src/inmanta/server/agentmanager.py
```diff
@@ -168,12 +169,67 @@ def __init__(self, closesessionsonstart: bool = True, fact_back_off: Optional[in
...
+            # Mypy can't help here, ....
```
sanderr, line 187: Well,  `assert result.result is not None` would probably do it. Still far from perfect.

src/inmanta/util/__init__.py
```diff
@@ -392,6 +392,25 @@ def action_function() -> None:
         self._scheduled[task_spec] = handle
         return task_spec
 
+    def schedule(
```
sanderr, line 395: What does this method offer? I do prefer the name, but I feel like it's just a layer on top of `add_action` that makes a tiny streamline of the interface for one type of schedules, at the cost of dropping support for the other. So now we need to call either `schedule()` or `add_action()` depending on the use case.

src/inmanta/server/protocol.py
```diff
@@ -399,7 +390,7 @@ def get_extension_statuses(cls, slices: list["ServerSlice"]) -> list[ExtensionSt
                 result[ext_status.name] = ext_status
         return list(result.values())
 
-    async def get_status(self) -> dict[str, ArgumentTypes]:
+    async def get_status(self) -> Mapping[str, ArgumentTypes | Mapping[str, ArgumentTypes]]:
```
sanderr, line 393: Why change all these from `dict` to `Mapping`?

sanderr (changes requested):
I'm only vaguelly familiar with all of this tbh.

changelogs/unreleased/8228-database_pool_report.yml
wouterdb: It is a user visible change, but perhaps I should add more detail?

src/inmanta/server/agentmanager.py
wouterdb: problem is that you also has to assert the result is a dict, containing `data` which is also a dict, ....

src/inmanta/server/agentmanager.py
wouterdb: Good question, it is a bit a special case as the exception is handled right next to it and logged as debug, so it matters little, but I can improve if desired. 

This code is very very robust, making it less failure transparent.

src/inmanta/server/protocol.py
wouterdb: Due to invariance of dict, the Union is a bit hard to use otherwise  and it is inteded to be mapping all along

src/inmanta/util/__init__.py
sanderr: To me it seems like it adds more confusion than anything else. If I see a method `schedule` I'd assume that to be the main interface, but it isn't because it only supports one of the two schedule types.

So I prefer it without, but if you feel it's useful I won't stand in the way.

changelogs/unreleased/8228-database_pool_report.yml
sanderr: I mean the entire scheduler is new, so I would assume that to be one change entry rather than a separate one for each prt it affects.

src/inmanta/server/agentmanager.py
sanderr: True

src/inmanta/server/protocol.py
wouterdb: Left it like this, I think Mapping better reflects the intent.

changelogs/unreleased/8228-database_pool_report.yml
wouterdb: I don't want to lose the details, but we can probably collect them at the end

src/inmanta/server/protocol.py
sanderr: I don't necessarily agree as a general rule, but I can see your point for this case.

mypy-baseline.txt
```diff
@@ -740,6 +739,7 @@ src/inmanta/ast/variables.py:0: error: Missing type parameters for generic type
 src/inmanta/ast/variables.py:0: error: Missing type parameters for generic type "VariableReader"  [type-arg]
 src/inmanta/ast/variables.py:0: error: Argument 2 to "AttributeReference" has incompatible type "str"; expected "LocatableString"  [arg-type]
 src/inmanta/server/protocol.py:0: error: Argument 1 to "RESTServer" has incompatible type "SessionManager"; expected "SessionManagerInterface"  [arg-type]
+src/inmanta/server/protocol.py:0: error: Name "IntervalSchedule" is not defined  [name-defined]
```
sanderr, line 742: This looks like a missing import to me.

sanderr (approved):
One comment but I expect the tests will fail on it so I'm approving.

mypy-baseline.txt
wouterdb: it is

PR #8294, opened 2024-10-30
strip status

wouterdb:
# Description

Remove api capabilties that will become hard to maintain

# Self Check:

- [x] Changelog entry


sanderr (approved):

changelogs/unreleased/8196-remove_status.yml
```diff
@@ -0,0 +1,14 @@
...
+change-type: major
+destination-branches: [master]
+sections:
+  upgrade-note: >
+    All api endpoints reporting deployment status for specific versions of resources have been removed.
+
+    | api endpoint | change | alternative |
+    | ------------ | ------ | ------------|
+    | `/api/v1/resource/<id>` | removal of the status argument and removal of status field from the response | `/api/v2/resource/<id>/` |
+    | `/api/v1/version/<id>` | removal of `resources/status` field from the response | `/api/v2/resource` |
```
wouterdb, line 12: Is there an easy way to test this?

sanderr: I'm not sure it would be worth it. Even if it's still accidentally there (which would be a great surprise), we communicated its removal so we can consider it a bugfix if/when we remove it.

sanderr (approved):

changelogs/unreleased/8196-remove_status.yml
wouterdb: I mean the rendering of the changelog entry ;-)

sanderr: That makes more sense! Honestly I'd merge it and trigger the docs build, then check the changelog. If need be we can always update the file.

sanderr: And ideally we'd check it again when it's no longer the last in the list, to make sure that the table ends properly. Or we change the name so that it's no longer alphabetically last.

Hugo-Inmanta (approved):

PR #8299, opened 2024-10-31
More fixes for the serializer

### LINKED ISSUE #8267: Sanitize the changes dict prior to sending it through pickle
The changes to the executor reporting cause the changes dict to be pickled instead of json serialized, this can lead to problems when un-serializing. 

It should be sanitized (or force pickle to json serialize it)

wouterdb:
# Description

Make serializer more safe by switching to json when needed

dis-advantages:
1. doesn't reconstruct objects  (which is why it is applied only at the leaves)

closes #8267 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/mypy-baseline.txt
+++ b/mypy-baseline.txt
@@ -146,6 +146,10 @@ src/inmanta/execute/runtime.py:0: error: Argument 2 to "assign" of "AssignableNo
 src/inmanta/execute/runtime.py:0: error: Argument 1 to "AttributeException" has incompatible type "Instance"; expected "Locatable"  [arg-type]
 src/inmanta/execute/runtime.py:0: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
 src/inmanta/execute/runtime.py:0: error: Argument 1 to "RuntimeException" has incompatible type "Instance"; expected "Locatable | None"  [arg-type]
+src/inmanta/data/model.py:0: error: Return type "str" of "__getstate__" incompatible with return type "dict[Any, Any]" in supertype "BaseModel"  [override]
+src/inmanta/data/model.py:0: error: Argument 1 of "__setstate__" is incompatible with supertype "BaseModel"; supertype defines the argument type as "dict[Any, Any]"  [override]
+src/inmanta/data/model.py:0: note: This violates the Liskov substitution principle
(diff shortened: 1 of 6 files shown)


mypy-baseline.txt
```diff
@@ -146,6 +146,10 @@ src/inmanta/execute/runtime.py:0: error: Argument 2 to "assign" of "AssignableNo
 src/inmanta/execute/runtime.py:0: error: Argument 1 to "AttributeException" has incompatible type "Instance"; expected "Locatable"  [arg-type]
 src/inmanta/execute/runtime.py:0: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
 src/inmanta/execute/runtime.py:0: error: Argument 1 to "RuntimeException" has incompatible type "Instance"; expected "Locatable | None"  [arg-type]
+src/inmanta/data/model.py:0: error: Return type "str" of "__getstate__" incompatible with return type "dict[Any, Any]" in supertype "BaseModel"  [override]
+src/inmanta/data/model.py:0: error: Argument 1 of "__setstate__" is incompatible with supertype "BaseModel"; supertype defines the argument type as "dict[Any, Any]"  [override]
```
wouterdb, line 150: The typing in typeshed does not cover this case

src/inmanta/data/__init__.py
```diff
@@ -4518,7 +4530,7 @@ class Resource(BaseDocument):
 
     # Methods for backward compatibility
     @property
-    def resource_version_id(self):
+    def resource_version_id(self) -> ResourceVersionIdStr:
```
wouterdb, line 4533: accidental type cleanup

bartv (approved):

src/inmanta/data/model.py
```diff
@@ -242,6 +243,16 @@ def check_serializable(cls, v: Optional[object]) -> Optional[object]:
                 return str(v)
         return v
 
+    def __getstate__(self) -> str:
+        # make pickle use json to keep leaking stuff
```
sanderr, line 247:
```suggestion
        # make pickle use json to keep from leaking stuff
```

sanderr (approved):
Still gives me some minor shivers, but I fail to see a more satisfying solution.

PR #8301, opened 2024-10-31
Issue/8253 read resource state from scheduler

### LINKED ISSUE #8253: Read resource state from scheduler instead of DB
When doing a deploy_start, we read the state of all dependent resources from the DB. 
We could also read this from the scheduler itself. 

Once the new code is proven sufficiently stable, we should also make the switch here.

jptrindade:
# Description

Read resource state from scheduler instead of from DB. 

closes #8253 

# Self Check:

- [x] Changelog entry


src/inmanta/deploy/scheduler.py
```diff
@@ -628,6 +628,18 @@ async def report_resource_state(
...
+    ) -> dict[ResourceIdStr, const.ResourceState]:
+        """
+        Get resource state for every dependency of a give resource from the scheduler state
+
+        :param resource_id: The id of the resource to find the dependencies for
+        """
+        provides_view: Mapping[ResourceIdStr, Set[ResourceIdStr]] = self._state.requires
+        dependencies: Set[ResourceIdStr] = provides_view.get(resource_id, set())
+        return {id: const.ResourceState(self._state.resource_state[id].deployment_result) for id in dependencies}
+
```
jptrindade, line 642: the `const.ResourceState` does not match 1-to-1 to the `ResourceState` in `self._state.resource_state[id]`. The only states that are accounted for are `deployed` and `failed` (There is also the `new` state which was not defined on `const.ResourceState` and if it happens here it should crash).

How should this be handled? Do we only need to take into account the `deployed` and `failed` states?

src/inmanta/deploy/persistence.py
```diff
@@ -115,13 +119,14 @@ async def dryrun_update(self, env: UUID, dryrun_result: DryrunResult) -> None:
...
 
-    def __init__(self, client: Client, environment: UUID) -> None:
+    def __init__(self, client: Client, environment: UUID, scheduler: "ResourceScheduler") -> None:
         self.environment = environment
         # TODO: The client is only here temporarily while we fix the dryrun_update
         self.client = client
         # FIXME: We may want to move the writing of the log to the scheduler side as well,
         #  when all uses of this logger are moved
         self._resource_action_logger = logging.getLogger(const.NAME_RESOURCE_ACTION_LOGGER)
+        self.scheduler = scheduler
```
jptrindade, line 129: This will cause a circular import that is why I did the `TYPE_CHECKING` check. If you have another suggestion, let me know.

src/inmanta/deploy/scheduler.py
wouterdb: you need to convert the state from one form to the other. 

(the table is already in some testcase, but it may need some TLC)

wouterdb (changes requested):

src/inmanta/deploy/scheduler.py
```diff
@@ -628,6 +628,18 @@ async def report_resource_state(
                             deploying=set(),
                         )
 
+    def get_last_non_deploying_state_for_dependencies(
+        self, resource_id: ResourceIdStr
+    ) -> dict[ResourceIdStr, const.ResourceState]:
+        """
+        Get resource state for every dependency of a give resource from the scheduler state
```
Hugo-Inmanta, line 635:
```suggestion
        Get resource state for every dependency of a given resource from the scheduler state
```

src/inmanta/deploy/scheduler.py
```diff
@@ -628,6 +628,18 @@ async def report_resource_state(
...
+
+        :param resource_id: The id of the resource to find the dependencies for
+        """
+        provides_view: Mapping[ResourceIdStr, Set[ResourceIdStr]] = self._state.requires
```
Hugo-Inmanta, line 639: I think this variable name is confusing since these are the resources that this one requires and not the other way around (I think ?)

src/inmanta/deploy/persistence.py
sanderr: I feel like in this case it might actually point at a discouraged module structure, but I'll know more once I've gone through the rest of the changes.

wouterdb (approved):
The skipped state may cause trouble

src/inmanta/deploy/persistence.py
```diff
@@ -180,14 +185,10 @@ async def send_in_progress(
 
                 # FIXME: we may want to have this in the RPS table instead of Resource table, at some point
                 await resource.update_fields(connection=connection, status=const.ResourceState.deploying)
-
-            # FIXME: shortcut to use scheduler state
-            result = await data.Resource.get_last_non_deploying_state_for_dependencies(
-                environment=self.environment, resource_version_id=resource_id_parsed, connection=connection
+            return await self.scheduler.get_last_non_deploying_state_for_dependencies(
+                resource_id=resource_id_parsed.resource_str()
```
sanderr, line 189: I don't think this belongs here. It makes the flow a bit more convoluted than it needs to be: `scheduler -> task -> scheduler -> update manager -> scheduler` while it could simply be `scheduler -> task -> scheduler`. Most importantly, it adds a lock on yet another layer. I think we could easily make this part of `Scheduler.get_resource_intent` instead. It doesn't really have anything to do with persistence anymore.

sanderr: This will also solve the cyclic import issue.

src/inmanta/deploy/scheduler.py
```diff
@@ -628,6 +636,37 @@ async def report_resource_state(
                             deploying=set(),
                         )
 
+    async def get_last_non_deploying_state_for_dependencies(
```
sanderr, line 639:
```suggestion
    async def _get_last_non_deploying_state_for_dependencies(
```

src/inmanta/deploy/scheduler.py
```diff
@@ -628,6 +636,37 @@ async def report_resource_state(
...
+        """
+        Get resource state for every dependency of a given resource from the scheduler state.
+        The state is then converted to const.ResourceState.
+
+        Acquires the appropriate locks
+
+        :param resource_id: The id of the resource to find the dependencies for
+        """
+        async with self._scheduler_lock:
+            requires_view: Mapping[ResourceIdStr, Set[ResourceIdStr]] = self._state.requires.requires_view()
```
sanderr, line 651: Are you aware that `requires_view()` simply returns `self`? i.e. I don't think there's a need for this, you can simply use `self._state.requires` directly, unless this was a conscious decision?

sanderr (changes requested):
Still have to review the state mapping implementation (though I expect it will be fine given that we discussed it elaborately on Slack). But I have a meeting now so I'm releasing my other comments in the meantime.

src/inmanta/deploy/scheduler.py
jptrindade: Yes, I was aware. I find it more legible like this, but no strong opinion either way

src/inmanta/deploy/scheduler.py
```diff
@@ -559,24 +575,23 @@ async def is_agent_running(self, name: str) -> bool:
 
     # TaskManager interface
 
-    async def get_resource_intent(
-        self, resource: ResourceIdStr, *, for_deploy: bool = False
-    ) -> Optional[tuple[int, ResourceDetails]]:
+    async def get_resource_intent(self, resource_id: ResourceIdStr, *, for_deploy: bool = False) -> Optional[ResourceIntent]:
```
sanderr, line 578: I'd prefer to keep this `resource`, again for consistency with other methods.

src/inmanta/deploy/scheduler.py
```diff
@@ -45,6 +54,13 @@
 LOGGER = logging.getLogger(__name__)
 
 
+@dataclass
+class ResourceIntent:
+    state_version: int
+    resource: ResourceDetails
+    dependencies: dict[ResourceIdStr, const.ResourceState]
```
sanderr, line 61: Nice! One suggestion (optional). And a few naming suggestions to be consistent with the rest of the code:
```suggestion
@dataclass(frozen=True)
class ResourceIntent:
    model_version: int
    details: ResourceDetails
    dependencies: Mapping[ResourceIdStr, const.ResourceState]
```

src/inmanta/deploy/scheduler.py
```diff
@@ -628,13 +643,42 @@ async def report_resource_state(
                             deploying=set(),
                         )
 
+    async def _get_last_non_deploying_state_for_dependencies(
+        self, resource_id: ResourceIdStr
```
sanderr, line 647:
```suggestion
        self, resource: ResourceIdStr
```

src/inmanta/deploy/scheduler.py
```diff
@@ -628,13 +643,42 @@ async def report_resource_state(
...
+    ) -> dict[ResourceIdStr, const.ResourceState]:
+        """
+        Get resource state for every dependency of a given resource from the scheduler state.
+        The state is then converted to const.ResourceState.
+
+        Acquires the appropriate locks
```
sanderr, line 653:
```suggestion
        Should only be called under scheduler lock.
```

sanderr (changes requested):

sanderr:
I'm also rerequesting Wouter's review because I caused some conceptual changes to be made.

src/inmanta/deploy/scheduler.py
```diff
@@ -628,13 +643,42 @@ async def report_resource_state(
...
+
+        Should only be called under scheduler lock.
+
+        :param resource_id: The id of the resource to find the dependencies for
+        """
+        requires_view: Mapping[ResourceIdStr, Set[ResourceIdStr]] = self._state.requires.requires_view()
+        dependencies: Set[ResourceIdStr] = requires_view.get(resource_id, set())
+        dependencies_state = {}
+        for dep_id in dependencies:
+            new_state = self._state.resource_state[dep_id]
```
Hugo-Inmanta, line 661: I think we should type this. And I'm not too fond of the name `new_state` which implies a value change in the current method, what about `resource_state_object` / `resource_state_instance` ? Not a strong opinion either way, feel free to ignore.
```suggestion
            new_state: ResourceState = self._state.resource_state[dep_id]
```

jptrindade: I changed it to `resource_state_object` thanks for the suggestion

src/inmanta/agent/forking_executor.py
```diff
@@ -811,7 +811,7 @@ async def execute(
         gid: uuid.UUID,
         resource_details: "inmanta.agent.executor.ResourceDetails",
         reason: str,
-        requires: dict[ResourceIdStr, const.ResourceState],
```
sanderr, line 814: We typically use it as the unqualified `Mapping` nowadays. For new code we should follow that approach.

sanderr (approved):
I hope mypy accepts the Mapping changes now.

Hugo-Inmanta (approved):

src/inmanta/deploy/scheduler.py
wouterdb: @sanderr style and performance wise, I'm a bit apprehensive about this: 
1. we changed what this method does (it now also checks the state of dependencies)
2. it already had two modes


I wonder if we shouldn't split/copy this in two distinct methods, one to get the resource intent (for get_fact) and one to start deploying?

sanderr: https://inmanta.slack.com/archives/CKRF0C8R3/p1731414114134429

wouterdb (approved):

PR #8302, opened 2024-11-04
Fix logline

wouterdb:
# Description

Fix for bug in log-line deserialization 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/agent_executor_7_13.yml
+++ b/changelogs/unreleased/agent_executor_7_13.yml
@@ -1,0 +1,4 @@
+description: Fix serialization safety
+change-type: patch
+destination-branches: [master]
+
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4011,7 +4011,7 @@ class LogLine(DataDocument):
 
     def __setstate__(self, state: str) -> None:
         # This method exists only to keep IPC light compatible with the json based RPC
-        self.__dict__.update(json.loads(state))
+        self._data = json.loads(state)
 
 
 @stable_api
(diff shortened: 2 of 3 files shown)


edvgui (approved):
I don't understand how this works, I approved because I trust you do.  If you need a real review we should pull somebody else in

tests/conftest.py
```diff
@@ -749,8 +751,13 @@ async def server(server_pre_start) -> abc.AsyncIterator[Server]:
...
     except concurrent.futures.TimeoutError:
         logger.exception("Timeout during stop of the server in teardown")
-
     logger.info("Server clean up done")
+    tests_failed_during = request.session.testsfailed - tests_failed_before
+    if tests_failed_during and auto_start_agent:
+        for file in glob(log_dir.get() + "/*"):
+            if not os.path.isdir(file):
+                with open(file, "r") as fh:
+                    logger.debug("%s\n%s", file, fh.read())
```
wouterdb, line 760: this one really helps

tests/conftest.py
```diff
@@ -725,10 +727,10 @@ async def server_config(
 
 
 @pytest.fixture(scope="function")
-async def server(server_pre_start) -> abc.AsyncIterator[Server]:
+async def server(server_pre_start, request, auto_start_agent) -> abc.AsyncIterator[Server]:
     # fix for fact that pytest_tornado never set IOLoop._instance, the IOLoop of the main thread
     # causes handler failure
-
+    tests_failed_before = request.session.testsfailed
```
wouterdb, line 733: Only way I could find how to detect if the testcase has failed.

hlloreda (approved):

PR #8326, opened 2024-11-06
scheduler state fixes

sanderr:
# Description

Fixed two small bugs in scheduler state updates:
- when a deploy fails, make sure to mark it as dirty (e.g. when it was scheduled despite being in an assumed good state due to an event)
- when a deploy fails, leave its status what it was before, rather than to overwrite it with `HAS_UPDATE`. If the operational state previously satisfied the resource's desired state, a handler failure does not necessarily change that. We have the `deployment_result` to reflect that aspect of a resource's state.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/deploy/test_scheduler_agent.py
+++ b/tests/deploy/test_scheduler_agent.py
@@ -1119,8 +1119,8 @@ async def test_deploy_event_propagation(agent: TestAgent, make_resource_minimal)
     assert len(agent.scheduler._state.dirty) == 0
 
     # release change for r1, sending event to r2 when it finishes
-    resources = make_resources(version=16, r1_value=5, r2_value=0, r3_value=0)
-    await agent.scheduler._new_version(16, resources, make_requires(resources))
+    resources = make_resources(version=17, r1_value=5, r2_value=0, r3_value=0)
+    await agent.scheduler._new_version(17, resources, make_requires(resources))
     await retry_limited_fast(lambda: rid2 in executor2.deploys)
 
     # verify that r2 is still in an assumed good state, even though we're deploying it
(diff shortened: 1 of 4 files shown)


wouterdb (approved):

tests/deploy/test_scheduler_agent.py
```diff
@@ -1095,6 +1095,69 @@ def make_resources(
...
+    assert agent.executor_manager.executors["agent3"].execute_count == 0
+    assert agent.scheduler._state.resource_state[rid2] == state.ResourceState(
+        status=state.ResourceStatus.UP_TO_DATE,
+        deployment_result=state.DeploymentResult.DEPLOYED,
+        blocked=state.BlockedStatus.NO,
+    )
+    assert len(agent.scheduler._state.dirty) == 0
+
+    # release change for r1, sending event to r2 when it finishes
+    resources = make_resources(version=16, r1_value=5, r2_value=0, r3_value=0)
```
Hugo-Inmanta, line 1122: Is the version number not incremented on purpose ?

tests/deploy/test_scheduler_agent.py
```diff
@@ -1095,6 +1095,69 @@ def make_resources(
...
+    await agent.scheduler._new_version(16, resources, make_requires(resources))
```
sanderr, line 1123:
```suggestion
    resources = make_resources(version=17, r1_value=5, r2_value=0, r3_value=0)
    await agent.scheduler._new_version(17, resources, make_requires(resources))
```

Hugo-Inmanta (approved):

PR #8329, opened 2024-11-07
Stabilize traffic light - Second part

hlloreda:
# Description

- Added a diagnostic mechanism to retrieve useful information when compilations are taking too much time

# Self Check:

- [ ] Changelog entry


tests/server/test_workon.py
```diff
@@ -1193,3 +1294,160 @@ async def test_workon_sets_pip_config(
...
+            for i in range(nb_environments)
+        ]
+
+        for env in environments:
+            result: protocol.Result = await client.notify_change(env.id)
+            assert result.code == 200
+
+        try:
+            raise AssertionError
+        except AssertionError:
```
jptrindade, line 1389: I don't get this part. Is it so we have something written in the logs?

hlloreda: Yes, it was written like this to simulate a timeout but I think we could replace it with await asynctio.sleep(..). Wdyt?

tests/server/test_workon.py
```diff
@@ -1193,3 +1294,160 @@ async def test_workon_sets_pip_config(
             " the shell. This ensures the proper permission checks are performed."
         ),
     )
+
+
+@pytest.mark.parametrize("iteration", range(1000))
```
arnaudsjs, line 1299: This test case triggers a compile but it never waits until the compile is finished. This means that we might be branches in this test case that we never test. It might for example happen that none of the 1000 iterations contain a finished compile report. Maybe we have to introduce an additional parameter so that we wait for a compile to finish is some of these 1000 iterations.

arnaudsjs (changes requested):

tests/server/test_workon.py
hlloreda: Sorry, it was a leftover 🤦🏼‍♂️

jptrindade (approved):

arnaudsjs (approved):

PR #8334, opened 2024-11-07
Drop // syntax for comments

### LINKED ISSUE #6972: drop // syntax for comments
Turns out we have both `#` and `//` for comments, and noone was aware of the latter. The `//` syntax should be removed from iso8 onwards.

jptrindade:
# Description

Drop // syntax for comments

closes #6972 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/6972-drop-double-slash-comment-syntax.yml
+++ b/changelogs/unreleased/6972-drop-double-slash-comment-syntax.yml
@@ -1,0 +1,7 @@
+description: Drop // syntax for comments
+issue-nr: 6972
+change-type: major
+destination-branches: [master]
+sections:
+  deprecation-note: >
+    The // syntax for comments is no longer supported. Please use # for comments.
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1212,10 +1212,10 @@ end"""
 
 
 def test_lexer():
+    # // comments are no longer supported
     parse_code(
         """
 #test
-//test2
 a=0.5
 b=""
 """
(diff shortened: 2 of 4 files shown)


changelogs/unreleased/6972-drop-double-slash-comment-syntax.yml
```diff
@@ -0,0 +1,7 @@
+description: Drop // syntax for comments
+issue-nr: 6972
+change-type: major
+destination-branches: [master]
+sections:
+  deprecation-note: >
+    The // syntax for comments is no longer supported. Please use # for comments.
```
Hugo-Inmanta, line 7: Could you double check that this renders nicely in the changelog page ?

Hugo-Inmanta (commented):
I think we could add a small section about comments to the Inmanta language reference https://docs.inmanta.com/inmanta-service-orchestrator/7/language.html#language-reference

jptrindade:
I added a section in the documentation, but still couldn't figure out how to see if the changelog renders correctly

Hugo-Inmanta (approved):

tests/test_parser.py
```diff
@@ -1212,10 +1212,10 @@ def test_define_dict_attribute():
 
 
 def test_lexer():
+    # // comments are no longer supported
```
sanderr, line 1215: This message refers to a change in this PR, not to anything related to the resulting test. I think it should go, because it's meaning is unclear.

sanderr (changes requested):

sanderr (approved):

PR #8339, opened 2024-11-08
Track skipped resources

jptrindade:
# Description

In the current implementation we don't track skipped resources, and simply label them as failed. This PR changes that. In the future we will have a different state for resources that were skipped by a custom handler

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/deploy/tasks.py
+++ b/src/inmanta/deploy/tasks.py
@@ -130,13 +130,10 @@ class Deploy(Task):
         # The main difficulty off this code is exception handling
         # We collect state here to report back in the finally block
 
-        # Status of the deploy for the executor/requires/provides
-        success: bool = True
-
         # Full status of the deploy,
         # may be unset if we fail before signaling start to the server, will be set if we signaled start
         deploy_result: DeployResult | None = None
-        scheduler_deployment_result: state.DeploymentResult | None = None
+        scheduler_deployment_result: state.DeploymentResult
 
         try:
             # This try catch block ensures we report at the end of the task
@@ -148,7 +145,7 @@ class Deploy(Task):
                 )
             except Exception:
                 # Unrecoverable, can't reach server
-                success = False
+                scheduler_deployment_result = state.DeploymentResult.FAILED
                 LOGGER.error(
                     "Failed to report the start of the deployment to the server for %s",
                     resource_details.resource_id,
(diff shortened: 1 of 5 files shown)


src/inmanta/deploy/tasks.py
```diff
@@ -182,6 +183,11 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
             # Deploy
             try:
                 deploy_result = await my_executor.execute(action_id, gid, executor_resource_details, reason, requires)
+                # Translate deploy result status to the new deployment result state
+                if deploy_result.status == const.ResourceState.deployed:
+                    deploy_state = state.DeploymentResult.DEPLOYED
+                elif deploy_result.status == const.ResourceState.skipped:
+                    deploy_state = state.DeploymentResult.SKIPPED
```
sanderr, line 190: Purely a style preference, but I think this would be more readable as a conditional expression.

src/inmanta/deploy/tasks.py
```diff
@@ -136,6 +136,7 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
         # Full status of the deploy,
         # may be unset if we fail before signaling start to the server, will be set if we signaled start
         deploy_result: DeployResult | None = None
+        deploy_state: state.DeploymentResult = state.DeploymentResult.FAILED
```
sanderr, line 139: I have a strong preference for not setting a value here, and to require each branch to set it to the appropriate value. i.e. anywhere `success` is currently set.

Additionally, `deploy_state` is a strange name, since it's a deployment result. If we want to avoid ambiguity with `deploy_result`, I'd rather rename that one with more context, e.g. `executor_deploy_result`. Then, if you still find it ambiguous, name this one `scheduler_deployment_result`.

src/inmanta/deploy/tasks.py
```diff
@@ -217,7 +223,7 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
                 resource=self.resource,
                 attribute_hash=resource_details.attribute_hash,
                 status=state.ResourceStatus.UP_TO_DATE if success else None,
```
sanderr, line 225: I think we should get rid off `success` now and derive it from the deployment result.

sanderr (changes requested):
Concept looks good to me, some minor comments on implementation details.

src/inmanta/deploy/tasks.py
sanderr: (100% optional)

PR #8341, opened 2024-11-08
Added deploy timer

wouterdb:
# Description

1. I added the ability to report deploy time statistics
2. I moved the tags on component and environment from the DBMonitor to the process level. This is both cleaner and more consistent. (if this isn't done, the data will also be impossible to process)

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -158,7 +158,9 @@ def start_scheduler(options: argparse.Namespace) -> None:
             connection_timeout=agent_config.scheduler_db_connection_timeout.get(),
         )
         # also report metrics if this is relevant
-        metrics_reporter = MetricsService()
+        metrics_reporter = MetricsService(
+            extra_tags={"component": "scheduler", "environment": str(agent_config.environment.get())}
+        )
         metrics_reporter.start_metric_reporters()
         await a.start()
 
--- a/src/inmanta/deploy/tasks.py
+++ b/src/inmanta/deploy/tasks.py
@@ -24,6 +24,8 @@ import traceback
 import uuid
 from dataclasses import dataclass
 
+import pyformance
+
 from inmanta import const, data, resources
 from inmanta.agent import executor
 from inmanta.agent.executor import DeployResult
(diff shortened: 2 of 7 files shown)


src/inmanta/app.py
```diff
@@ -158,7 +158,9 @@ async def start() -> None:
             connection_timeout=agent_config.scheduler_db_connection_timeout.get(),
         )
         # also report metrics if this is relevant
-        metrics_reporter = MetricsService()
+        metrics_reporter = MetricsService(
+            extra_tags={"component": "scheduler", "environment": str(agent_config.environment.get())}
```
wouterdb, line 162: This is moved from the DatabaseMonitor

src/inmanta/server/services/metricservice.py
```diff
@@ -34,9 +34,10 @@
 class MetricsService(protocol.ServerSlice):
     """Slice managing metrics collector"""
 
-    def __init__(self) -> None:
+    def __init__(self, extra_tags: dict[str, str] = {"component": "server"}) -> None:
         super().__init__(SLICE_METRICS)
         self._influx_db_reporter: Optional[InfluxReporter] = None
+        self._extra_tags = extra_tags
```
arnaudsjs, line 40: Or we can also make it a mapping instead of a dict.
```suggestion
        self._extra_tags = dict(extra_tags)
```

tests/test_influxdbreporting.py
```diff
@@ -184,10 +184,10 @@ async def test_available_metrics(server):
                     base_types = [base_type]
                 types[key] = base_types
 
-    assert metrics["db.connected,component=server"]["value"]
-    assert "db.max_pool,component=server" in metrics
-    assert "db.open_connections,component=server" in metrics
-    assert "db.free_connections,component=server" in metrics
```
arnaudsjs, line 190: Why is the `component=server` part gone here?

src/inmanta/deploy/tasks.py
```diff
@@ -110,115 +112,115 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
 
 class Deploy(Task):
     async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reason: str | None = None) -> None:
+        with pyformance.timer("internal.deploy").time():
```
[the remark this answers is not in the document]
arnaudsjs: I am fine with `internal.deploy`.

arnaudsjs (changes requested):

tests/test_influxdbreporting.py
wouterdb: it is moved to the output reporter, so in the output, it is there, internally, it is not

arnaudsjs (approved):

src/inmanta/server/services/metricservice.py
```diff
@@ -34,9 +34,10 @@
 class MetricsService(protocol.ServerSlice):
     """Slice managing metrics collector"""
 
-    def __init__(self) -> None:
+    def __init__(self, extra_tags: dict[str, str] = {"component": "server"}) -> None:
```
sanderr, line 37:
```suggestion
    def __init__(self, extra_tags: Mapping[str, str] = {"component": "server"}) -> None:
```
Right?

sanderr, line 37: You have a mutable default here. If typed as `Mapping`, we have some confidence that it won't cause trouble, but even then it might be considered bad practice.

sanderr (changes requested):

sanderr (approved):

src/inmanta/server/services/metricservice.py
```diff
@@ -18,7 +18,7 @@
 
 import logging
 from time import perf_counter, time
-from typing import Optional
+from typing import Mapping, Optional
```
sanderr, line 21: Ideally we'd import this from `collections.abc`.

Issue #8315, opened 2024-11-05
move to Python 3.12


sanderr:
- [x] Make sure we build 3.12
- [x] Set default python on Jenkins runners to 3.12
- [ ] set minimum Python version on relevant Python projects
  - [x]  inmanta-service-orchestrator https://github.com/inmanta/inmanta-service-orchestrator/pull/536
  - [x] inmanta https://github.com/inmanta/inmanta/pull/140
  - [x] inmanta-license ->  joined head, can't do
  - [x] inmanta-support  -> joined head, can't do
  - [x] inmanta-lsm -> https://github.com/inmanta/inmanta-lsm/pull/1981
  - [x] inmanta-ui -> joined head, can't do
  - [ ] inmanta-core
- [x] update product config files if required
- [x] Make sure we test against 3.12
   - [x]  core https://github.com/inmanta/irt/pull/2271 
   - [x]  extensions
   - [x]  modules -> use default
      - [x]  std
      - [x]  lsm
      - [x]  ....

PR #8349, opened 2024-11-13
Fixes for python 3.12

wouterdb:
# Description

The root issue was pytest-coverage it seems

part of #8315 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -298,7 +298,7 @@ setup(name="test")
     expected = re.escape("Packages this-package-does-not-exist were not found. No indexes were used.")
     other_expected = r"Packages setuptools.*were not found\. No indexes were used\."
 
-    with pytest.raises(env.PackageNotFound, match=f"{expected}|{other_expected}" ):
+    with pytest.raises(env.PackageNotFound, match=f"{expected}|{other_expected}"):
         deactive_venv.install_for_config(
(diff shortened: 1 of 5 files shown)


tests/test_env.py
```diff
@@ -295,13 +295,15 @@ def test_process_env_install_no_index(tmpdir: py.path.local, monkeypatch, use_sy
     with open(setup_py_path, "w") as setup_file:
         setup_file.write(setup_py_content)
 
-    expected = "Packages this-package-does-not-exist were not found. No indexes were used."
+    expected = re.escape("Packages this-package-does-not-exist were not found. No indexes were used.")
+    other_expected = r"Packages setuptools.*were not found\. No indexes were used\."
 
-    with pytest.raises(env.PackageNotFound, match=re.escape(expected)):
+    with pytest.raises(env.PackageNotFound, match=f"{expected}|{other_expected}"):
```
arnaudsjs, line 301: This needs a docstring. Why do we expect one of these error messages?

arnaudsjs (changes requested):

tests/test_env.py
wouterdb: Because we install two things, they fail differently, but order is not deterministic

arnaudsjs (approved):

PR #8350, opened 2024-11-13
Issue/compiler venv

### LINKED ISSUE #7732: Remove compiler venvs with out-dated Python version
When the server has been upgraded to a new Python version, the compiler server should re-create the compiler venvs. Up to now this was done by [a script](https://github.com/inmanta/inmanta/blob/186cf7b2fc99827c81900408bfd65aa2d11c20cf/inmanta.spec#L265) included in the RPM that would move the out-dated venv into a `.rpmsave` folder. The migration to native docker containers requires this functionality to move into the server itself.

Tasks:

- [x] Add the above-mentioned functionality to the compiler service.
- [ ] Remove the script from the RPM

Question: 

Do we rename the venvs like the RPM does or do we actually remove them?

wouterdb:
# Description

Make compiler venv specific to python version 

closes #7732 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -747,7 +747,11 @@ class PythonEnvironment:
             self._write_pth_file()
 
     def can_activate(self) -> None:
-        """Can this venv be activated with this current python version?"""
+        """
+        Can this venv be activated with this current python version?
+
+        raises a VenvActivationFailedError exception if this is not the case
+        """
         # Make sure the venv hosts the same python version as the running process
         if sys.platform.startswith("linux"):
             # Check if the python binary exists in the environment's bin directory
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -728,35 +728,7 @@ class PythonEnvironment:
 
         # check if the virtual env exists
         if os.path.isdir(self.env_path) and os.listdir(self.env_path):
-            # Make sure the venv hosts the same python version as the running process
-            if sys.platform.startswith("linux"):
-                # Check if the python binary exists in the environment's bin directory
-                if not os.path.exists(self.python_path):
-                    raise VenvActivationFailedError(
-                        msg=f"Unable to use virtualenv at {self.env_path} as no Python installation exists."
-                    )
-                # On linux based systems, the python version is in the path to the site packages dir:
-                if not os.path.exists(self.site_packages_dir):
-                    raise VenvActivationFailedError(
-                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version "
-                        "is different from the Python version of this process."
-                    )
-            else:
-                # On other distributions a more costly check is required:
-                # Get version as a (major, minor) tuple for the venv and the running process
-                venv_python_version = (
-                    subprocess.check_output([self.python_path, "--version"]).decode("utf-8").strip().split()[1]
-                )
-                venv_python_version = tuple(map(int, venv_python_version.split(".")))[:2]
-
-                running_process_python_version = sys.version_info[:2]
-
-                if venv_python_version != running_process_python_version:
-                    raise VenvActivationFailedError(
-                        msg=f"Unable to use virtualenv at {self.env_path} because its Python version "
-                        "is different from the Python version of this process."
-                    )
-
+            self.can_activate()
         else:
             path = os.path.realpath(self.env_path)
             try:
@@ -773,6 +745,35 @@ class PythonEnvironment:
             # Venv was created using an older version of Inmanta -> Update pip binary and set sitecustomize.py file
             self._write_pip_binary()
             self._write_pth_file()
+
+    def can_activate(self) -> None:
+        """Can this venv be activated with this current python version?"""
+        # Make sure the venv hosts the same python version as the running process
+        if sys.platform.startswith("linux"):
+            # Check if the python binary exists in the environment's bin directory
+            if not os.path.exists(self.python_path):
+                raise VenvActivationFailedError(
+                    msg=f"Unable to use virtualenv at {self.env_path} as no Python installation exists."
+                )
+            # On linux based systems, the python version is in the path to the site packages dir:
+            if not os.path.exists(self.site_packages_dir):
+                raise VenvActivationFailedError(
+                    msg=f"Unable to use virtualenv at {self.env_path} because its Python version "
+                    "is different from the Python version of this process."
+                )
+        else:
+            # On other distributions a more costly check is required:
+            # Get version as a (major, minor) tuple for the venv and the running process
+            venv_python_version = subprocess.check_output([self.python_path, "--version"]).decode("utf-8").strip().split()[1]
+            venv_python_version = tuple(map(int, venv_python_version.split(".")))[:2]
+
+            running_process_python_version = sys.version_info[:2]
+
+            if venv_python_version != running_process_python_version:
+                raise VenvActivationFailedError(
+                    msg=f"Unable to use virtualenv at {self.env_path} because its Python version "
+                    "is different from the Python version of this process."
+                )
 
     def _write_pip_binary(self) -> None:
         """
@@ -1428,7 +1429,7 @@ class VirtualEnv(ActiveEnv):
         """
         Returns True iff the venv exists on disk.
         """
-        return os.path.exists(self.python_path) and os.path.exists(self._path_pth_file)
+        return os.path.exists(self.python_path)
 
     def is_using_virtual_env(self) -> bool:
         return self._using_venv
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -235,7 +235,7 @@ class CompileRun:
 
         It is built as
         1. a `.env` symlink to
-        2. a `.env-py3.12` versioned symlink
+        2. a `.env-py3.12` versioned directory
         """
         assert os.path.exists(self._project_dir)
         project_dir = self._project_dir
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -18,6 +18,7 @@
 
 import abc
 import asyncio
+import contextlib
 import datetime
 import json
 import logging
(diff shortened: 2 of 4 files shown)


src/inmanta/env.py
```diff
@@ -728,35 +728,7 @@ def init_env(self) -> None:
 
         # check if the virtual env exists
         if os.path.isdir(self.env_path) and os.listdir(self.env_path):
-            # Make sure the venv hosts the same python version as the running process
```
wouterdb, line 731: factored out

src/inmanta/env.py
```diff
@@ -1428,7 +1429,7 @@ def exists(self) -> bool:
         """
         Returns True iff the venv exists on disk.
         """
-        return os.path.exists(self.python_path) and os.path.exists(self._path_pth_file)
```
wouterdb, line 1431: only used on the code path I'm working on.

I suspect the second check was there to force pth file creation, this restores the semantics to be in-line with the name.

src/inmanta/server/services/compilerservice.py
```diff
@@ -226,6 +227,84 @@ async def _run_compile_stage(self, name: str, cmd: list[str], cwd: str, env: dic
                 # The process is still running, kill it
                 sub_process.kill()
 
+    async def ensure_compiler_venv(self) -> None:
```
wouterdb, line 230: to be able to log, this needs to be async

1. Do I need to run the file operations on a thread?  (they block the ioloop)
2. Do I need to run the venv creation on a thread? (blocks the ioloop)

src/inmanta/server/services/compilerservice.py
```diff
@@ -268,13 +347,9 @@ async def ensure_venv() -> Optional[data.Report]:
...
                 Ensure a venv is present at `venv_dir`.
                 """
-                virtual_env = VirtualEnv(venv_dir)
-                if virtual_env.exists():
-                    return None
-
                 await self._start_stage("Creating venv", command="")
                 try:
-                    virtual_env.init_env()
+                    await self.ensure_compiler_venv()
```
wouterdb, line 352: the stage is now always present

src/inmanta/server/services/compilerservice.py
```diff
@@ -226,6 +227,84 @@ async def _run_compile_stage(self, name: str, cmd: list[str], cwd: str, env: dic
...
+        # Versioned python dir
+        python_version = ".".join(platform.python_version_tuple()[0:2])
+        versioned_venv_dir = ".env-py" + python_version
+        versioned_venv_dir_full = os.path.join(project_dir, versioned_venv_dir)
+
+        # Determine the final action
+        # If we don't need to symlink, we return
+        do_create = False
+
+        if os.path.islink(venv_dir):
```
wouterdb, line 255: I don't like this monster if/except/... but this is the best I could come up with

src/inmanta/env.py
arnaudsjs: This was a mechanism to make sure that we recover if a venv creation failed half way. The creation of the pth file is the last step in the venv creation process. If that file exists, we can be pretty sure that the venv was created successfully. But maybe we have to replace this with a marker file. That would make the code more clear and it's also more correct since the creation of the marker file is an atomic operation. The creation of the pth file isn't.

wouterdb: Mhh, how did we previously handle broken venv's on this code path I wonder?

wouterdb: we would fall into this code path:

```python
        if not os.path.exists(self._path_pth_file):
        if not os.path.exists(self._path_pth_file):
            # Venv was created using an older version of Inmanta -> Update pip binary and set sitecustomize.py file
            # Venv was created using an older version of Inmanta -> Update pip binary and set sitecustomize.py file
            self._write_pip_binary()
            self._write_pip_binary()
            self._write_pth_file()
```

I wonder how to best handle this. 

I was now (for backward compat) considering this to be something the user has to fix.

src/inmanta/server/services/compilerservice.py
```diff
@@ -226,6 +227,84 @@ async def _run_compile_stage(self, name: str, cmd: list[str], cwd: str, env: dic
...
 
+    async def ensure_compiler_venv(self) -> None:
+        """ "
+        Ensure we have a compiler venv in the project
+
+        Expected to be in a running stage
+
+        It is built as
+        1. a `.env` symlink to
+        2. a `.env-py3.12` versioned symlink
```
arnaudsjs, line 238:
```suggestion
        2. a `.env-py3.12` versioned directory
```

src/inmanta/server/services/compilerservice.py
```diff
@@ -226,6 +227,84 @@ async def _run_compile_stage(self, name: str, cmd: list[str], cwd: str, env: dic
...
+            # We have a modern setup
+            try:
+                if os.path.samefile(venv_dir, versioned_venv_dir_full):
+                    # all good
+                    await self._info("Found existing venv")
+                    return None
+                else:
+                    # Wrong version
+                    os.unlink(venv_dir)
+                    await self._info(f"Creating new venv at {versioned_venv_dir_full}")
```
arnaudsjs, line 265: Shouldn't we add more context to the log line indicating that the Python version of the venv is out-of-date?

arnaudsjs (changes requested):

tests/server/test_compilerservice.py
```diff
@@ -450,10 +452,11 @@ def _compile_and_assert(env, export=True, meta={}, env_vars={}, update=False, ex
     # no export
     compile, stages = await _compile_and_assert(env=env, export=False)
     assert stages["Init"]["returncode"] == 0
+    assert stages["Creating venv"]["returncode"] == 0
```
arnaudsjs, line 455: Isn't it rather confusing to always have a compiler stage called  `creating venv`, even if no venv has to be created? Should we call it `Ensure venv` or something similar?

tests/server/test_compilerservice.py
```diff
@@ -2016,3 +2023,84 @@ async def test_overlapping_env_vars(mocked_compiler_service, server, client, env
...
+    log_contains(caplog, "inmanta.server.services.compilerservice", logging.INFO, "Creating new venv at")
+    caplog.clear()
+
+    # re-use
+    await run.ensure_compiler_venv()
+    log_contains(caplog, "inmanta.server.services.compilerservice", logging.INFO, "Found existing venv")
+    caplog.clear()
+
+
+async def test_venv_upgrade_capture(tmp_path, caplog):
```
arnaudsjs, line 2057: I don't really understand what the word capture mean in this name and that of the next test case

tests/server/test_compilerservice.py
```diff
@@ -2016,3 +2023,84 @@ async def test_overlapping_env_vars(mocked_compiler_service, server, client, env
             env_vars={"var": "val", "test": "123"},
             mergeable_env_vars={"var": "otherval", "somekey": "someval"},
         )
+
+
+class Dudreport:
```
arnaudsjs, line 2028: What does `Dud` mean?

src/inmanta/env.py
arnaudsjs: > I was now (for backward compat) considering this to be something the user has to fix.

Fine by me.

tests/server/test_compilerservice.py
wouterdb: I went for venv check, as ensure is not a super clear or common word I think.

tests/server/test_compilerservice.py
wouterdb: I assumed it to be a common word, perhaps it is not. 
(such a nice word though) 

I'll change it to mock. 

image

tests/server/test_compilerservice.py
wouterdb: improved, thx

src/inmanta/server/services/compilerservice.py
arnaudsjs: I see your point. Maybe it's better to keep it like this.

arnaudsjs (changes requested):
I think you forgot to push some changes, because the changes you announced in comments are not present in the code.

arnaudsjs (approved):

src/inmanta/server/services/compilerservice.py
```diff
@@ -226,6 +227,84 @@ async def _run_compile_stage(self, name: str, cmd: list[str], cwd: str, env: dic
...
+            virtual_env = VirtualEnv(versioned_venv_dir_full)
+            virtual_env.init_env()
+
+        # Relative symlink is a bit exotic in python
+        try:
+            dir_fd = os.open(project_dir, os.O_RDONLY)
+            os.symlink(versioned_venv_dir, ".env", dir_fd=dir_fd)
+        finally:
+            if dir_fd:
+                os.close(dir_fd)
```
sanderr, line 306: I think relative symlinks are supported even without `dir_fd`:

```
sander@bedevere:~$ ll ~/downloads/dirtest/
total 52K
drwxr-xr-x   4 sander sander 4.0K Nov 20 15:59 ./
drwx------ 147 sander sander  36K Nov 20 15:08 ../
drwxr-xr-x   2 sander sander 4.0K Nov 20 15:09 test/
drwxr-xr-x   2 sander sander 4.0K Nov 20 15:08 test_old/
sander@bedevere:~$ python
Python 3.12.7 (main, Oct  1 2024, 11:15:50) [GCC 14.2.1 20240910] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.symlink("test", "/home/sander/downloads/dirtest/sl")
>>>
sander@bedevere:~$ ll ~/downloads/dirtest/
total 52K
drwxr-xr-x   4 sander sander 4.0K Nov 20 15:59 ./
drwx------ 147 sander sander  36K Nov 20 15:08 ../
lrwxrwxrwx   1 sander sander    4 Nov 20 15:59 sl -> test/
drwxr-xr-x   2 sander sander 4.0K Nov 20 15:09 test/
drwxr-xr-x   2 sander sander 4.0K Nov 20 15:08 test_old/
sander@bedevere:~$
```

sanderr (changes requested):

src/inmanta/env.py
wouterdb: Yes, but it also requires it to work (which implies many, many more things), the pth file is a very particular addition

src/inmanta/server/services/compilerservice.py
```diff
@@ -226,6 +227,84 @@ async def _run_compile_stage(self, name: str, cmd: list[str], cwd: str, env: dic
...
+                    os.rename(venv_dir, versioned_venv_dir_full)
+                    await self._info(f"Moving existing venv from {venv_dir} to {versioned_venv_dir_full}")
+                except VenvActivationFailedError:
+                    # Version doesn't match, move to alternative location
+                    os.rename(venv_dir, venv_dir + "_old")
+                    await self._info(f"Discarding existing venv from {venv_dir} to {venv_dir}_old, Creating new")
+                    do_create = True
+            else:
+                # otherwise broken
+                os.rename(venv_dir, venv_dir + "_old")
```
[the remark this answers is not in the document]
wouterdb: Or fail?

src/inmanta/server/services/compilerservice.py
wouterdb: yes, but relative to what? (what if you are in a different folder?)

src/inmanta/server/services/compilerservice.py
```diff
@@ -248,62 +249,53 @@ async def ensure_compiler_venv(self) -> None:
...
             virtual_env.init_env()
 
-        # Relative symlink is a bit exotic in python
-        try:
-            dir_fd = os.open(project_dir, os.O_RDONLY)
-            os.symlink(versioned_venv_dir, ".env", dir_fd=dir_fd)
-        finally:
-            if dir_fd:
-                os.close(dir_fd)
+        await ensure_venv()
```
sanderr, line 278: Only a style preference, but this is hidden a bit squeezed in between the two method definitions. I'd prefer the two calls at the bottom, so you have definitions of behavior first, followed by the "script".

src/inmanta/server/services/compilerservice.py
```diff
@@ -248,62 +249,53 @@ async def ensure_compiler_venv(self) -> None:
...
+            with contextlib.suppress(FileNotFoundError):
+                if is_link and os.path.samefile(venv_dir, versioned_venv_dir_full):
+                    await self._info("Found existing venv")
+                    return
+            if os.path.exists(venv_dir):
+                if is_link:
+                    os.unlink(venv_dir)
+                else:
+                    # TODO: open question: what if backup target already exists?
+                    os.rename(venv_dir, f"{venv_dir}_old")
```
sanderr, line 294: Perhaps the open question isn't as relevant as I initially thought it would be. We control the venv, and we should only ever fall into this branch (or the one in the other function) once.

sanderr (changes requested):
LGTM. Requesting changes only for the TODO in the code.

PR #8363, opened 2024-11-15
Add new resource states to the database

### LINKED ISSUE #8187: Add new states to the database
- [ ] Add the new states, as designed [here](https://docs.google.com/presentation/d/1F3bFNy2BZtzZgAxQ3Vbvdw7BWI9dq0ty5c3EoLAtUUY/edit#slide=id.g292b508a90d_0_5) , to the database.

- [ ] Update either deploy_done or equivalent if #8189  is done. 
- [ ] Drop `const.ResourceState` usage from the scheduler and scheduler state, see https://github.com/inmanta/inmanta-core/pull/8165#discussion_r1794988785 and the comment regarding the mapping of states.

We expect to write both new and old state, to the resource persistent state table. 
This may affect the whole chain, from BaseResource.deploy up into the DB.

arnaudsjs:
# Description

Add new resource states to the database.

closes #8187 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4800,31 +4800,35 @@ class Resource(BaseDocument):
     async def reset_resource_state(
         cls,
         environment: uuid.UUID,
-        version: int,
         *,
         connection: Optional[asyncpg.connection.Connection] = None,
     ) -> None:
         """
-        Update resources on the latest version of the model stuck in "deploying" state. The status will be reset to the latest
-        non deploying status.
+        Update resources on the latest released version of the model stuck in "deploying" state.
+        The status will be reset to the latest non deploying status.
 
         :param environment: The environment impacted by this
-        :param version: The version of the model impacted by this
         :param connection: The connection to use
         """
+
         query = f"""
-            UPDATE resource AS updated_r
-            SET status=inconsistent_r.last_non_deploying_status::TEXT::resourcestate
-            FROM (
-                SELECT r.resource_id, r.status, r.model, ps.last_non_deploying_status
-                FROM {Resource.table_name()} r
-                INNER JOIN {ResourcePersistentState.table_name()} ps ON r.resource_id = ps.resource_id
-                AND r.environment = ps.environment
-                WHERE r.environment=$1 AND r.model = $2 AND r.status = 'deploying'
-            ) AS inconsistent_r
-            WHERE inconsistent_r.resource_id = updated_r.resource_id AND inconsistent_r.model = updated_r.model
+            UPDATE {Resource.table_name()} r
+            SET status=ps.last_non_deploying_status::TEXT::resourcestate
+            FROM {ResourcePersistentState.table_name()} ps
+            WHERE r.resource_id=ps.resource_id
+                AND r.environment=ps.environment
+                AND r.status='deploying'
+                AND r.environment=$1
+                AND r.model=(
+                    SELECT version
+                    FROM {ConfigurationModel.table_name()}
+                    WHERE environment=$1
+                        AND released=true
+                    ORDER BY version DESC
+                    LIMIT 1
+                )
         """
-        values = [cls._get_value(environment), cls._get_value(version)]
+        values = [cls._get_value(environment)]
         async with cls.get_connection(connection) as connection:
             await connection.execute(query, *values)
 
@@ -6013,14 +6017,14 @@ class ConfigurationModel(BaseDocument):
 
         outset: set[ResourceIdStr] = {res["resource_id"] for res in increment}
         original_provides: dict[str, list[ResourceIdStr]] = defaultdict(list)
-        send_events: list[ResourceIdStr] = []
+        send_events: set[ResourceIdStr] = set()
 
         # build lookup tables
         for res in resources:
             for req in res["requires"]:
                 original_provides[req].append(res["resource_id"])
             if res[const.RESOURCE_ATTRIBUTE_SEND_EVENTS]:
-                send_events.append(res["resource_id"])
+                send_events.add(res["resource_id"])
 
         # recursively include stuff potentially receiving events from nodes in the increment
         increment_work: list[ResourceIdStr] = list(outset)
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4391,6 +4391,11 @@ class ResourceAction(BaseDocument):
 
 
 class ResourcePersistentState(BaseDocument):
+    """
+    To avoid write contention, the `ComplianceStatus` is split up in different fields that are written from different code
+    paths. See get_compliance_status() for the associated logic.
+    """
+
     @classmethod
     def table_name(cls) -> str:
         return "resource_persistent_state"
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4436,33 +4436,21 @@ class ResourcePersistentState(BaseDocument):
     last_non_deploying_status: const.NonDeployingResourceState = const.NonDeployingResourceState.available
 
     @classmethod
-    async def update_orphan_state(cls, environment: uuid.UUID, connection: Optional[Connection] = None) -> None:
+    async def mark_as_orphan(
+        cls,
+        environment: UUID,
+        resource_ids: Set[ResourceIdStr],
+        connection: Optional[Connection] = None,
+    ) -> None:
         """
         Set the is_orphan column to True on all resources in the given environment that became an orphan.
         """
         query = f"""
-            WITH latest_released_version AS (
-                SELECT max(c.version) AS version
-                FROM {ConfigurationModel.table_name()} AS c
-                WHERE c.environment=$1 AND c.released
-            )
-            UPDATE {cls.table_name()} AS rps
-            SET
-                is_orphan=TRUE
-            WHERE
-                rps.environment=$1
-                AND NOT EXISTS(
-                    SELECT *
-                    FROM {Resource.table_name()} AS r
-                    WHERE (
-                        r.environment=$1
-                        AND r.model>=(SELECT version FROM latest_released_version)
-                        AND r.resource_id=rps.resource_id
-                    )
-                )
-                AND NOT rps.is_orphan
+            UPDATE {cls.table_name()}
+            SET is_orphan=TRUE
+            WHERE environment=$1 AND resource_id=ANY($2)
         """
-        await cls._execute_query(query, environment, connection=connection)
+        await cls._execute_query(query, environment, resource_ids, connection=connection)
 
     @classmethod
     async def update_resource_intent(
@@ -4518,6 +4506,57 @@ class ResourcePersistentState(BaseDocument):
             ) and rps.environment=$1
             """,
             environment,
+            connection=connection,
+        )
+
+    @classmethod
+    async def populate_for_version(
+        cls, environment: uuid.UUID, model_version: int, connection: Optional[Connection] = None
+    ) -> None:
+        """
+        Make sure that the resource_persistent_state table has a record for each resource present in the
+        given model version. This method assumes that the given model_version is the latest released version.
+        """
+        await cls._execute_query(
+            f"""
+            INSERT INTO {cls.table_name()} (
+                environment,
+                resource_id,
+                resource_type,
+                agent,
+                resource_id_value,
+                current_intent_attribute_hash,
+                is_undefined,
+                is_orphan,
+                deployment_result,
+                blocked_status
+            )
+            SELECT
+                r.environment,
+                r.resource_id,
+                r.resource_type,
+                r.agent,
+                r.resource_id_value,
+                r.attribute_hash,
+                r.status = 'undefined'::public.resourcestate,
+                FALSE,
+                'NEW',
+                CASE
+                    WHEN
+                        r.status = 'undefined'::public.resourcestate
+                        OR r.status = 'skipped_for_undefined'::public.resourcestate
+                    THEN 'YES'
+                    ELSE 'NO'
+                END
+            FROM {Resource.table_name()} AS r
+            WHERE r.environment=$1 AND r.model=$2 AND NOT EXISTS(
+                SELECT *
+                FROM {cls.table_name()} AS rps
+                WHERE rps.environment=r.environment AND rps.resource_id=r.resource_id
+            )
+            """,
+            environment,
+            model_version,
             connection=connection,
         )
 
@@ -4896,23 +4935,16 @@ class Resource(BaseDocument):
         This method generates a report of all resources in the given environment,
         with their latest version and when they are last deployed.
         """
-        query_resource_ids = f"""
-                SELECT resource_id, last_deployed_version as deployed_version, last_deploy
-                FROM {ResourcePersistentState.table_name()}
-                WHERE environment=$1
-        """
-        query_latest_version = f"""
-                SELECT resource_id, model AS latest_version
-                FROM {Resource.table_name()}
-                WHERE environment=$1 AND
-                      resource_id=r1.resource_id
-                ORDER BY model DESC
-                LIMIT 1
-        """
         query = f"""
-                SELECT r1.resource_id, r2.latest_version, r1.deployed_version, r1.last_deploy
-                FROM ({query_resource_ids}) AS r1 INNER JOIN LATERAL ({query_latest_version}) AS r2
-                      ON (r1.resource_id = r2.resource_id)
+        WITH latest_version_of_each_resource AS (
+            SELECT environment, max(model) AS model, resource_id
+            FROM {Resource.table_name()}
+            WHERE environment=$1
+            GROUP BY (environment, resource_id)
+        )
+        SELECT lver.resource_id, lver.model AS latest_version, rps.last_deployed_version AS deployed_version, rps.last_deploy
+        FROM latest_version_of_each_resource AS lver LEFT JOIN {ResourcePersistentState.table_name()} AS rps
+            ON lver.environment=rps.environment AND lver.resource_id=rps.resource_id
         """
         values = [cls._get_value(environment)]
         result = []
@@ -5096,7 +5128,9 @@ class Resource(BaseDocument):
         )
 
     @classmethod
-    async def get_resource_details(cls, env: uuid.UUID, resource_id: m.ResourceIdStr) -> Optional[m.ReleasedResourceDetails]:
+    async def get_released_resource_details(
+        cls, env: uuid.UUID, resource_id: m.ResourceIdStr
+    ) -> Optional[m.ReleasedResourceDetails]:
         def status_sub_query(resource_table_name: str) -> str:
             return f"""
             (CASE
@@ -5334,38 +5368,6 @@ class Resource(BaseDocument):
     async def insert(self, connection: Optional[asyncpg.connection.Connection] = None) -> None:
         self.make_hash()
         await super().insert(connection=connection)
-        # TODO: On conflict or is not exists or just make every update an upsert?
-        await self._execute_query(
-            """
-            INSERT INTO resource_persistent_state (
-                environment,
-                resource_id,
-                resource_type,
-                agent,
-                resource_id_value,
-                is_undefined,
-                is_orphan,
-                deployment_result,
-                blocked_status
-            )
-            VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
-            ON CONFLICT DO NOTHING
-            """,
-            self.environment,
-            self.resource_id,
-            self.resource_type,
-            self.agent,
-            self.resource_id_value,
-            self.status is const.ResourceState.undefined,
-            False,
-            state.DeploymentResult.NEW.name,
-            (
-                state.BlockedStatus.YES.name
-                if self.status in (const.ResourceState.undefined, const.ResourceState.skipped_for_undefined)
-                else state.BlockedStatus.NO.name
-            ),
-            connection=connection,
-        )
 
     @classmethod
     async def insert_many(
@@ -5374,38 +5376,6 @@ class Resource(BaseDocument):
         for doc in documents:
             doc.make_hash()
         # TODO performance?
-        for doc in documents:
-            await cls._execute_query(
-                """
-                INSERT INTO resource_persistent_state (
-                    environment,
-                    resource_id,
-                    resource_type,
-                    agent,
-                    resource_id_value,
-                    is_undefined,
-                    is_orphan,
-                    deployment_result,
-                    blocked_status
-                )
-                VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
-                ON CONFLICT DO NOTHING
-                """,
-                doc.environment,
-                doc.resource_id,
-                doc.resource_type,
-                doc.agent,
-                doc.resource_id_value,
-                doc.status is const.ResourceState.undefined,
-                False,
-                state.DeploymentResult.NEW.name,
-                (
-                    state.BlockedStatus.YES.name
-                    if doc.status in (const.ResourceState.undefined, const.ResourceState.skipped_for_undefined)
-                    else state.BlockedStatus.NO.name
-                ),
-                connection=connection,
-            )
         await super().insert_many(documents, connection=connection)
 
     async def update(self, connection: Optional[asyncpg.connection.Connection] = None, **kwargs: object) -> None:
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4526,7 +4526,11 @@ class ResourcePersistentState(BaseDocument):
             return state.ComplianceStatus.ORPHAN
         elif self.is_undefined:
             return state.ComplianceStatus.UNDEFINED
-        elif self.current_intent_attribute_hash != self.last_deployed_attribute_hash:
+        elif self.blocked_status is state.BlockedStatus.YES:
+            return state.ComplianceStatus.NON_COMPLIANT
+        elif (
+            self.last_deployed_attribute_hash is None or self.current_intent_attribute_hash != self.last_deployed_attribute_hash
+        ):
             return state.ComplianceStatus.HAS_UPDATE
         elif self.deployment_result is state.DeploymentResult.DEPLOYED:
             return state.ComplianceStatus.COMPLIANT
@@ -5363,7 +5367,7 @@ class Resource(BaseDocument):
             (
                 state.BlockedStatus.YES.name
                 if self.status in (const.ResourceState.undefined, const.ResourceState.skipped_for_undefined)
-                else state.BlockedStatus.NO
+                else state.BlockedStatus.NO.name
             ),
             connection=connection,
         )
@@ -5403,7 +5407,7 @@ class Resource(BaseDocument):
                 (
                     state.BlockedStatus.YES.name
                     if doc.status in (const.ResourceState.undefined, const.ResourceState.skipped_for_undefined)
-                    else state.BlockedStatus.NO
+                    else state.BlockedStatus.NO.name
                 ),
                 connection=connection,
             )
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -6488,6 +6488,21 @@ class Scheduler(BaseDocument):
 
     __primary_key__ = ("environment",)
 
+    @classmethod
+    async def set_last_processed_model_version(
+        cls, environment: uuid.UUID, version: int, connection: Optional[asyncpg.connection.Connection] = None
+    ) -> None:
+        await cls._execute_query(
+            f"""
+            UPDATE {cls.table_name()}
+            SET last_processed_model_version=$1
+            WHERE environment=$2
+            """,
+            version,
+            environment,
+            connection=connection,
+        )
+
 
 _classes = [
     Project,
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -20,7 +20,6 @@ import asyncio
 import copy
 import datetime
 import enum
-import hashlib
 import json
 import logging
 import re
@@ -62,7 +61,6 @@ from inmanta.data.model import (
     api_boundary_datetime_normalizer,
 )
 from inmanta.deploy import state
-from inmanta.protocol.common import custom_json_encoder
 from inmanta.protocol.exceptions import BadRequest, NotFound
 from inmanta.server import config
 from inmanta.stable_api import stable_api
@@ -4638,15 +4636,7 @@ class Resource(BaseDocument):
         return {r["resource_id"] + ",v=" + str(version): const.ResourceState(r["last_non_deploying_status"]) for r in result}
 
     def make_hash(self) -> None:
-        character = json.dumps(
-            {k: v for k, v in self.attributes.items() if k not in ["requires", "provides", "version"]},
-            default=custom_json_encoder,
-            sort_keys=True,  # sort the keys for stable hashes when using dicts, see #5306
-        )
-        m = hashlib.md5()
-        m.update(self.resource_id.encode("utf-8"))
-        m.update(character.encode("utf-8"))
-        self.attribute_hash = m.hexdigest()
+        self.attribute_hash = util.make_attribute_hash(self.resource_id, self.attributes)
 
     @classmethod
     async def get_resources(
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4519,6 +4519,9 @@ class ResourcePersistentState(BaseDocument):
         )
 
     def get_compliance_status(self) -> state.ComplianceStatus:
+        """
+        Return the ComplianceStatus associated with this resource_persistent_state.
+        """
         if self.is_orphan:
             return state.ComplianceStatus.ORPHAN
         elif self.is_undefined:
@@ -5357,7 +5360,11 @@ class Resource(BaseDocument):
             self.status is const.ResourceState.undefined,
             False,
             state.DeploymentResult.NEW.name,
-            state.BlockedStatus.NO.name,
+            (
+                state.BlockedStatus.YES.name
+                if self.status in (const.ResourceState.undefined, const.ResourceState.skipped_for_undefined)
+                else state.BlockedStatus.NO
+            ),
             connection=connection,
         )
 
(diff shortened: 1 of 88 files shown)


src/inmanta/data/__init__.py
```diff
@@ -5303,17 +5340,23 @@ async def insert(self, connection: Optional[asyncpg.connection.Connection] = Non
         self.make_hash()
         await super().insert(connection=connection)
         # TODO: On conflict or is not exists or just make every update an upsert?
+        resource_status = (
+            state.ResourceStatus.UNDEFINED if self.status is const.ResourceState.undefined else state.ResourceStatus.HAS_UPDATE
```
Hugo-Inmanta, line 5344: > else state.ResourceStatus.HAS_UPDATE

We set `resource_status` to `state.ResourceStatus.HAS_UPDATE` for all the following resource states ?
``` 
    unavailable = "unavailable"  # This state is set by the agent when no handler is available for the resource
    skipped = "skipped"  #
    dry = "dry"
    deployed = "deployed"
    failed = "failed"
    deploying = "deploying"
    available = "available"
    skipped_for_undefined = "skipped_for_undefined" 
```

arnaudsjs: This is the code path where a new resources are pushed to the server (put_version or put_partial) and only those that are not known by the server yet. So only the "initial" states are relevant here. All states that are the result of a deployment action cannot happen. So the relevant ones are: `available`, `undefined` and `skipped_for_undefined`. So it it's not an UNDEFINED resource, it must be HAS_UPDATE.

src/inmanta/deploy/scheduler.py
```diff
@@ -532,6 +545,9 @@ async def _new_version(
                 for resource in blocked_resources | transitively_blocked_resources:
                     self._work.delete_resource(resource)
 
+                # Write dirty state back to the database
+                await ResourcePersistentState.mark_as_has_update(self.environment, set(self._state.dirty))
```
sanderr, line 549: I'm hesitant about doing this under the scheduler lock, because it is a potentially slow IO operation, and everything under the lock is quite performance sensitive.

But doing it outside the lock also requires a lot of care. I wonder if we should consider bulk db writes every few seconds instead of on-the-fly writes. I'll come back to this.

src/inmanta/deploy/scheduler.py
```diff
@@ -360,19 +361,31 @@ async def _build_resource_mappings_from_db(
...
-            resource.resource_id: ResourceDetails(
-                resource_id=resource.resource_id,
-                attribute_hash=resource.attribute_hash,
-                attributes=resource.attributes,
-                status=resource.status,
+        resources_from_db = await data.Resource.get_resources_for_version_raw_with_persistent_state(
+            environment=self.environment,
+            version=version,
+            projection=None,
+            projection_presistent=None,
```
sanderr, line 368: typo? (presistent)

src/inmanta/deploy/scheduler.py
```diff
@@ -360,19 +361,31 @@ async def _build_resource_mappings_from_db(
 
         :return: resource_mapping {id -> resource details}
         """
```
sanderr, line 363: I feel like the changes to this method are mixing up responsibilities. I would say the scheduler is the authority on the resource state, and it ensures that the db comes to reflect that state. The db is the authority on the resource intent, which is what we should be fetching here.

I think we should return only resource id, attribute hash, attributes and either `const.ResourceStatus` or simply an `undefined` boolean. Whether we use the `ResourceDetails` dataclass, or a typed dict, or .... for it is an implementation detail. Either way, I do not think we should read anything other than resource intent from the db, nor return it, nor have a field for it in `ResourceDetails`.

sanderr: Exception: when we initialize the scheduler we could read the state as well, and use it to restore scheduler state (requires careful consideration of the scenario where the scheduler is stopped before it had a chance to process the latest version). But for now I think that the "increment calculation" we have going on will suffice.

src/inmanta/deploy/scheduler.py
sanderr, line 549: Not everything in the dirty set has an update. e.g. a repair puts everything in the dirty set, or resources that are `NON_COMPLIANT` (this last one didn't exist yet when you started this PR though).

src/inmanta/data/__init__.py
```diff
@@ -4501,6 +4504,40 @@ async def trim(cls, environment: UUID, connection: Optional[Connection] = None)
             connection=connection,
         )
 
+    @classmethod
+    async def mark_orphans(
```
sanderr, line 4508: Any reason to calculate the orphans rather than to pass them from the scheduler? It should have them in memory already, shouldn't it?

src/inmanta/data/__init__.py
sanderr, line 5344: These inserts into the resource persisted state table are the only ones where the server should actually write to it imo, all others should be through the scheduler.

sanderr (commented):
Seems like a tough ticket.

I went over the bulk of the pull request (skipped db migration, mypy diff and tests for now). Some of my comments are subjective and more design-level than typical review comments. So I'd appreciate your and @wouterdb's input on those.

src/inmanta/db/versions/v202411140.py
```diff
@@ -0,0 +1,100 @@
...
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    """
+    Add `resource_status` field to the resource_persistent_state table and populate it.
+    """
+    schema = """
+        CREATE TYPE public.new_resource_status AS ENUM ('UP_TO_DATE', 'HAS_UPDATE', 'UNDEFINED', 'ORPHAN');
```
wouterdb, line 27: I don't know if enum's in the database are super usefull? I.e. I think they are kind of messy to migrate and we already have the type checks?

(It may have advantages as well, I don't know)

src/inmanta/data/__init__.py
```diff
@@ -4482,6 +4483,8 @@ def table_name(cls) -> str:
     # Last produced an event. i.e. the end time of the last deploy where we had an effective change
     # (change is not None and change != Change.nochange)
 
+    resource_status: state.ResourceStatus
```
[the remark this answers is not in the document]
wouterdb: I would expect that as well

src/inmanta/deploy/persistence.py
```diff
@@ -305,7 +306,8 @@ def error_and_log(message: str, **context: Any) -> None:
                     last_deployed_version=resource_id_parsed.version,
                     last_deployed_attribute_hash=resource.attribute_hash,
                     last_non_deploying_status=const.NonDeployingResourceState(status),
-                    **extra_fields,
+                    resource_status=state.ResourceStatus.UP_TO_DATE if status is const.ResourceState.deployed else None,
```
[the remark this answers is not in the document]
wouterdb: I think, when we arrive here, the scheduler has already been updated, so we can just flush back the state we find there?

wouterdb (commented):
Overall:

1. I think it attaches in the expected places (deploy done, scheduler intialization and new version)
2. I think we are missing a concurrency mechanism, for scheduler to know what the last loaded version is, which would simplify a lot (e.g. orphans)
3. we may want to have a method to flush back state to the DB. All state updates (after initial creation) are then that method.

src/inmanta/deploy/persistence.py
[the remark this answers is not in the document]
sanderr: Conceptually yes, implementation-wise it's a bit of a puzzle because we're outside the scheduler domain here. But I still that that that's what we should do, I'm just not sure yet how best to achieve it. Perhaps we consider this method the compatibility layer, and add a new method to this class for the new fields, one that can be fully driven by the scheduler rather than the deploy task.

src/inmanta/server/services/resourceservice.py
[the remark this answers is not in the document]
arnaudsjs: The `ToServerUpdateManager` class was removed by the API cleanup PR.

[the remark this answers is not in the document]
sanderr: Great!

src/inmanta/data/__init__.py
arnaudsjs: I agree.

src/inmanta/db/versions/v202411140.py
arnaudsjs: I am also not a big fan of them. I remember I once said that we shouldn't use them anymore, but @sanderr disagreed to that. I don't recall the motivation. @sanderr Could you give your input on this.

src/inmanta/deploy/scheduler.py
```diff
@@ -641,7 +660,7 @@ async def report_resource_state(
         async with self._scheduler_lock:
             # refresh resource details for latest model state
             details: Optional[ResourceDetails] = self._state.resources.get(resource, None)
-            if details is None or details.attribute_hash != attribute_hash:
+            if details is None or details.status is ResourceStatus.UNDEFINED or details.attribute_hash != attribute_hash:
```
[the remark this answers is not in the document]
arnaudsjs: I think has to do with an issue I encountered in the tests. In the tests it's possible to have a resource in the undefined state that doesn't have an `<<undefined>>` attribute. So you can have a resource A of version X that is undefined, while the same resource in version X+1 is not undefined and has the same resource hash. Conclusion: I think you are right that the undefined check can be removed here, because the above-mentioned scenario cannot happen in practice.

src/inmanta/deploy/scheduler.py
arnaudsjs: This feels like a premature optimization. Like you say: _"it is a **potentially** slow IO operation"_. Wouldn't it be better to keep it like this an evaluate the performance. We can still optimize this later on, right? Or do you see issues with that?

src/inmanta/db/versions/v202411140.py
sanderr: > I don't recall the motivation.

Me neither, so I'll give some fresh input. ~~One potential risk I see is with evolving code: if we use enums in Python and ints in the database, we must make sure their int representations remain stable. We can mitigate that risk by not using `enum.auto()` and adding a comment to the enum definitions that the values must remain stable.~~ EDIT: We probably can't do ints.

Iirc the recommended (by Markus) approach, was to use (constrained?) ints, and then to have another table mapping ints to names (essentially defining the enum). I'm not sure how feasible that change would be, given that our data access framework is already in place. I think that this was one of my motivations for sticking with the enum: tough to migrate to the right approach, so might as well stick to the current approach instead of moving to another "not best-practice" one.

Other than that, and the minor inconvenience when working with the database directly, I don't have any other objections. And I do agree there are benefits as well.

Overall, if we can easily move to the best-practice approach, and we can integrate it cleanly, I think I'm in favor. If not, do I understand correctly you'd go for `varchar[]?` in the database? I'm more on the fence for that one, but I'm willing to give it a shot and see how we like it down the line.

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
sanderr: I see. I'm also not against keeping it and adding a comment that the scenario can happen in test cases. I'm typically against tailoring the code for the tests, but in this case the check is redundant, but still conceptually correct, so I'm more lenient.

src/inmanta/deploy/scheduler.py
sanderr: What I mean with "potentially" in this case is not "we don't know if it will be", but more "it occasionally *is*", e.g. waiting for a connection or a db lock. I'll try to give this some more thought today, but at the moment I think we should try to find a way where our db writes do not block scheduler operation, e.g. we read everything we want to flush inside the lock, but do the actual flushing outside of it. We then only have to make sure that our flushes can never be done out of order.

src/inmanta/data/__init__.py
```diff
@@ -4408,6 +4409,9 @@ def table_name(cls) -> str:
 
     # Field based on content from the resource actions
     last_deploy: Optional[datetime.datetime] = None
+    # When a resource is updated in a new model version, it might take some time until this update reaches the scheduler.
+    # This is the attribute hash that the scheduler considers the last released attribute hash for the given resource.
+    current_intent_attribute_hash: Optional[str] = None
```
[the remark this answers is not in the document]
wouterdb: HAS_UDPATE state, right. 


I would document that:
```
To avoid write contention, the `ComplianceStatus` is split up in different fields that are written from different code paths. see get_compliance_status for the associated logic 
```

src/inmanta/deploy/state.py
```diff
@@ -142,28 +202,110 @@ class ModelState:
...
+    dirty: set["ResourceIdStr"] = dataclasses.field(default_factory=set)
     # types per agent keeps track of which resource types live on which agent by doing a reference count
     # the dict is agent_name -> resource_type -> resource_count
-    types_per_agent: dict[str, dict[ResourceType, int]] = dataclasses.field(
+    types_per_agent: dict[str, dict["ResourceType", int]] = dataclasses.field(
         default_factory=lambda: defaultdict(lambda: defaultdict(lambda: 0))
     )
 
+    @classmethod
+    async def create_from_db(
```
wouterdb, line 220: This one scares mea bit, but I may not be seeing all ends. 

1. I does some rather delicate operations, manipulating a complex assembly of fields.
I would expect this kind of logic to be more isolated, but admittedly, every other method in this class is like this. 
There is no central 'update_state' method, instead, this entire class contains specific update methods. 
This may be for good reason, but it struct me as particular. 


2. For testing, it might be easier to split it in a DB related method that builds resource_state and details and the one that installs it. This way, the delicate loading logic can be tested without any DB related thing. (it could be less memory efficient, but I doubt it, as we load it all into memory )

src/inmanta/deploy/scheduler.py
```diff
@@ -267,17 +268,81 @@ async def _initialize(self) -> None:
...
-            # Get resources from the database
-            try:
-                version, resources, requires = await self._get_resources_in_latest_version(connection=con)
-            except KeyError:
-                # No model version has been released yet.
-                return
+        async with self._intent_lock:
+            async with self._scheduler_lock:
+                async with data.Scheduler.get_connection() as con:
+                    async with con.transaction():
```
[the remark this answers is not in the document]
sanderr: Might be a good case for multiple context managers in a single with statement though ([docs](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement)).

src/inmanta/deploy/persistence.py
```diff
@@ -65,6 +68,31 @@ async def dryrun_update(self, env: UUID, dryrun_result: DryrunResult) -> None:
...
+        pass
+
+    @abc.abstractmethod
+    async def set_last_processed_model_version(
+        self, environment: uuid.UUID, version: int, connection: Optional[Connection] = None
+    ) -> None:
+        """
+        Set the last model version that was processed by the scheduler.
+        """
+        await data.Scheduler.set_last_processed_model_version(environment, version, connection=connection)
```
wouterdb, line 94: marked as abstract, has implementation

src/inmanta/deploy/scheduler.py
```diff
@@ -54,6 +55,14 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class UnmanagedResource(Exception):
+    """
+    An exception that indicates that a resource is not managed by the orchestrator.
+    """
```
wouterdb, line 61: I would call it stale resource exception

src/inmanta/deploy/scheduler.py
```diff
@@ -357,17 +416,33 @@ def _should_deploy(resource: ResourceIdStr) -> bool:
...
+        result = []
+        for res in resources_in_version:
+            # Make mypy happy
+            assert res.attribute_hash is not None
+            result.append(
+                ResourceDetails(
+                    resource_id=res.resource_id,
+                    attribute_hash=res.attribute_hash,
+                    attributes=res.attributes,
+                    status=ComplianceStatus.UNDEFINED,  # TODO: FIX
```
wouterdb, line 430: todo alert

src/inmanta/deploy/state.py
```diff
@@ -87,6 +118,18 @@ class DeploymentResult(StrEnum):
...
 
+    @classmethod
+    def from_handler_resource_state(cls, handler_resource_state: const.HandlerResourceState) -> "DeploymentResult":
+        match handler_resource_state:
+            case const.HandlerResourceState.deployed:
+                return DeploymentResult.DEPLOYED
+            case const.HandlerResourceState.skipped | const.HandlerResourceState.skipped_for_dependency:
+                return DeploymentResult.SKIPPED
+            case const.HandlerResourceState.dry:
+                raise Exception(f"Unexpected handler_resource_state {const.HandlerResourceState.dry.value}")
```
wouterdb, line 129: I wonder if we should return a fail an log a warning?

i.e. will this bork the scheduler?

src/inmanta/deploy/state.py
```diff
@@ -139,25 +197,119 @@ class ModelState:
...
+                "current_intent_attribute_hash",
+                "last_deployed_attribute_hash",
+                "deployment_result",
+                "blocked_status",
+                "last_success",
+                "last_produced_events",
+            ],
+            project_attributes=["requires", const.RESOURCE_ATTRIBUTE_SEND_EVENTS],
+            connection=connection,
+        )
```
wouterdb, line 244: given the amount of data we need to massage below: why not use the DTO?

wouterdb (commented):
looks nice!

src/inmanta/data/__init__.py
```diff
@@ -4439,6 +4509,76 @@ async def trim(cls, environment: UUID, connection: Optional[Connection] = None)
...
+    def get_compliance_status(self) -> state.ComplianceStatus:
+        """
+        Return the ComplianceStatus associated with this resource_persistent_state.
+        """
+        if self.is_orphan:
+            return state.ComplianceStatus.ORPHAN
+        elif self.is_undefined:
+            return state.ComplianceStatus.UNDEFINED
+        elif self.blocked_status is state.BlockedStatus.YES:
+            return state.ComplianceStatus.NON_COMPLIANT
```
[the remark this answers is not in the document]
sanderr: Semantically, I'd say not. Implementation-wise, I'm not sure, I need to check. But if it's possible, I'd consider that a bug in the scheduler rather than in this function.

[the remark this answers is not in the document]
arnaudsjs: I agree with that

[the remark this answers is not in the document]
sanderr: > `BLOCKED <=> undefined`

This is not correct though (not in the current implementation, where `ComplianceStatus.UNDEFINED <=> intent undefined`). A resource can be blocked because *a dependency* is undefined.

Which makes me realize that I was missing that same case in my reply above. I think Wouter may be right that a resource can be both blocked and compliant, namely if it was in a good state, hasn't had any changes to its attributes since, and a dependency got into an undefined state since then. It then all boils down to what we consider `ComplianceStatus.UNDEFINED` to mean:
- if it means `ComplianceStatus.UNDEFINED <=> intent undefined`, then we can indeed be both blocked and compliant
- if on the other hand it means `ComplianceStatus.UNDEFINED <=> we don't know enough to judge compliance`, I think we should end up in that state when we get blocked, because we can't execute incremental deploys anymore, therefore our compliance data will go stale. In this case I think that we (@wouterdb perhaps in your PR?) should set the compliance status to undefined when we transitively block a resource. In this case this function is correct, as is Arnaud's latest statement.

I think the second definition makes a lot of sense, though we might consider changing the name of the `ComplianceStatus.UNDEFINED` value in that case, because it could be confusing that it is the same name we use for undefined intent, while meaning something broader.

Wdyt?

src/inmanta/data/__init__.py
```diff
@@ -5244,20 +5372,6 @@ async def get_resources_in_resource_sets(
...
-            VALUES ($1, $2, $3, $4, $5)
-            ON CONFLICT DO NOTHING
-            """,
-            self.environment,
-            self.resource_id,
-            self.resource_type,
-            self.agent,
-            self.resource_id_value,
-            connection=connection,
-        )
```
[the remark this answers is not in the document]
arnaudsjs: In the previous implementation the `resource_persistent_state` table was populated when the model version was created (`put_version()` and `put_partial()`). Now, this table is populated when a model version is processed by the scheduler. This method  is used by the `put_version()` / `put_partial()` code path so it doesn't need to write to the resource_persistent_state table anymore.

src/inmanta/deploy/state.py
arnaudsjs: The deploy task will catch this exception and log it, it will also mark the deployment result as FAILED. This is not going to break the compiler.

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
wouterdb: I wonder what the performance implications are...

src/inmanta/deploy/state.py
arnaudsjs: This method joins the resource table with the resource_persistent_state table and only selects very specific columns from the result. If we would go for the proposed approach of using the DTO we would:
1. Either have to make a lot more queries to the database.
2. Or, return all columns from the database.
Given that this method loads all the resources in the latest released version, I don't like option 1. Option 2 is possible if we accept the performance penalty. Wdyt?

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
arnaudsjs: I said it incorrectly. The resource_persistent_state table is populated by the server at release time (release_version()) and not by the scheduler. So there is no performance penalty in the scheduler is that is what you were wondering.

src/inmanta/deploy/scheduler.py
```diff
@@ -469,8 +556,17 @@ async def read_version(
...
+    async def _read_version(
+        self,
+        *,
+        intent_lock_acquired: bool = False,
+        scheduler_lock_acquired: bool = False,
+        connection: Optional[asyncpg.connection.Connection] = None,
+    ) -> None:
         try:
-            version, resources, requires = await self._get_resources_in_latest_version()
+            version, resources, requires = await self._get_resources_in_latest_version(connection=connection)
```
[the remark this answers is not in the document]
arnaudsjs: That's an argument for adding tests before we merge this in. But I will leave the final decision to you.

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
sanderr: > Why would another resources undefinedness cause uncertainty over our compliance?

Mainly because we lose the ability to verify it, so the compliance confidence will go down with time.

> we assume compliance remains unless we have a failure/skip or intent is updated. It is the fundamental assumption underlying increments.

Not entirely though: when a deploy fails we assume non-compliance even if it was compliant before, while this could mean all sort of things. So I would say the line is a bit more blurred than that.

> 2. ...

I'm not sure I follow 2, but I'm not sure it's worth getting into.

Overall I can get behind the first definition (we should probably document it). In which case this function needs to be updated accordingly.

sanderr (changes requested):
Not a full review yet, releasing intermediate comments.

src/inmanta/deploy/scheduler.py
```diff
@@ -298,25 +301,95 @@ async def _initialize(self) -> None:
...
+                    # Apply version from old to new
+                    await self._read_version(
+                        version=version,
+                        start_deployment=False,
+                        intent_lock_acquired=True,
+                        scheduler_lock_acquired=True,
+                        connection=con,
+                    )
+            # Start deploying everything in dirty set.
+            self._work.deploy_with_context(
```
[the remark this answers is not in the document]
sanderr: Additionally, `_new_version` assumes that timers are running for everything that is not scheduled when it's called, which is not the case if we restore scheduler state but not timer state.

Then again, #8453 will affect this as well. Proposal: we take the path of least resistance to get this merged, and I rework `read_version` in a follow-up PR to also do the iterative approach, which may unify the two flows again. For now, that means to start timers here (`timer_manager.update_timers()`) for all resources not in the dirty set.

src/inmanta/deploy/scheduler.py
```diff
@@ -644,6 +782,11 @@ async def _new_version(
         requires: Mapping[ResourceIdStr, Set[ResourceIdStr]],
         up_to_date_resources: Optional[Set[ResourceIdStr]] = None,
         reason: str = "Deploy was triggered because a new version has been released",
+        start_deployment: bool = True,
+        intent_lock_acquired: bool = False,
+        scheduler_lock_acquired: bool = False,
```
sanderr, line 787: I find that all these parameters make it a lot more difficult to reason on what's going on and how this may interact with other scheduler flows. But I do think much of these parameters can be dropped when we (I?) rework the non-initialize new-version flow to also use the iterative process, so I guess we can leave it be for now.

src/inmanta/deploy/scheduler.py
```diff
@@ -644,6 +782,11 @@ async def _new_version(
...
+        *,
+        connection: Optional[asyncpg.connection.Connection] = None,
```
sanderr, line 789: As it's implemented now, the scheduler lock lifetime might be interleaved with the transaction lifetime (as in the first db operation in the transaction is within the lock, but the lock is released before the transaction closes). I think this happens to be fine because of what we write to the database both here and at deploy done. But
1. I'm not sufficiently confident of this statement
2. It's a very big 

So I think we have to consider if it's worth it, or if perhaps we need to open a second transaction or move the other writes inside the scheduler lock. If we keep it as is I'd like a very explicit comment here and/or where we open the transaction.

This may also be affected by what happens above this method depending on which parameters are passed (any of the four new parameters). Which ties into the difficulty to reason about that I pointed out in another comment.

src/inmanta/deploy/scheduler.py
```diff
@@ -657,10 +800,13 @@ async def _new_version(
             successful deploy for the same intent. Should not include any blocked resources. Mostly intended for the very first
             version when the scheduler is started with a fresh state.
         :param reason: The reason to associate with any deploys caused by this newly released version.
+        :param intent_lock_acquired: True iff the self._intent_lock was already acquired before entering this method.
+        :param scheduler_lock_acquired: True iff the self._scheduler_lock was already acquired before entering this method.
         """
+        intent_lock = self._intent_lock if not intent_lock_acquired else asyncio.Lock()
```
sanderr, line 806: Why `asyncio.Lock()` rather than a simple async nullcontext?

sanderr (changes requested):
Some more comments. Still not through it all.

src/inmanta/deploy/scheduler.py
```diff
@@ -848,46 +1025,54 @@ async def get_resource_intent(self, resource: ResourceIdStr) -> Optional[Resourc
...
+    async def deploy_done(self, attribute_hash: str, result: DeployResult) -> None:
+        deployment_result: DeploymentResult = DeploymentResult.from_handler_resource_state(result.resource_state)
+        try:
+            await self.update_scheduler_state_for_finished_deploy(attribute_hash, result, deployment_result)
+        except StaleResource:
+            # The resource is no longer managed, no need to update the database state.
+            pass
+        else:
+            # Write deployment result to the database.
+            await self.send_deploy_done(attribute_hash, result, deployment_result)
```
sanderr, line 1053: I mentioned in the office that I thought there is a risk in writing this outside of the scheduler lock. But the motivation I had was related to `_task_done()`, which is in practice only called *after* the `deploy_done` call finishes. So it might be fine (and preferable) after all. Only reason I'm still in doubt is that I seem to recall my confidence that it had to be under the lock. Will get back to this.

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
sanderr: Path of least resistance is probably to revert the iteration part of this PR. Drop the associated extra parameters from `_read_version` etc, and simply call into the method as any `read_version` call would. Then we can dedicate a separate PR to the iterative approach, in a unified manner rather than to special-case it for the initializiation.

src/inmanta/deploy/persistence.py
```diff
@@ -253,7 +281,8 @@ def error_and_log(message: str, **context: Any) -> None:
                     last_deployed_version=resource_id_parsed.version,
                     last_deployed_attribute_hash=resource.attribute_hash,
                     last_non_deploying_status=const.NonDeployingResourceState(status),
-                    **extra_fields,
+                    deployment_result=deployment_result,
```
sanderr, line 284: Do I see correctly that the current implementation only writes deployment result, not compliance status or blocked status?

src/inmanta/deploy/state.py
wouterdb: Probably fine!

src/inmanta/deploy/persistence.py
wouterdb: I've been working on the code a bit to see where we are and I resolved this on a branch https://github.com/inmanta/inmanta-core/pull/8479/files

src/inmanta/deploy/scheduler.py
```diff
@@ -499,36 +588,51 @@ async def _build_discrepancy_map(
 
             return discrepancy_map
 
+        def report_model_version_mismatch(db_version: int | None) -> SchedulerStatusReport:
```
sanderr, line 599: Note to self: TODO: review these changes

src/inmanta/deploy/scheduler.py
```diff
@@ -1050,14 +1236,38 @@ async def _get_last_non_deploying_state_for_dependencies(
...
+    async def send_deploy_done(self, attribute_hash: str, result: DeployResult, deployment_result: DeploymentResult) -> None:
+        await self._state_update_delegate.send_deploy_done(attribute_hash, result, deployment_result)
 
     async def dryrun_update(self, env: uuid.UUID, dryrun_result: executor.DryrunResult) -> None:
         await self._state_update_delegate.dryrun_update(env, dryrun_result)
 
     async def set_parameters(self, fact_result: FactResult) -> None:
         await self._state_update_delegate.set_parameters(fact_result)
+
+    async def update_resource_intent(
```
sanderr, line 1271: Do we still need to implement the `StateUpdateManager` at all now that we lifted the authority of persisting things from the tasks to the scheduler? I'd think we could simply use composition for the persistence manager. I'd be glad to get rid of all these proxy methods and the extra layer of inheritance.

src/inmanta/deploy/scheduler.py
```diff
@@ -848,46 +1025,54 @@ async def get_resource_intent(self, resource: ResourceIdStr) -> Optional[Resourc
...
-
-        status = (
-            ComplianceStatus.COMPLIANT if deployment_result is DeploymentResult.DEPLOYED else ComplianceStatus.NON_COMPLIANT
-        )
+        """
+        Update the state of the scheduler based on the DeploymentResult of the given resource.
+
+        :raise StaleResource: This update is about a resource that is no longer managed by the server.
+        """
+        resource_id: ResourceIdStr = result.resource_id
```
sanderr, line 1075: Nice, I hadn't considered that this contains the resource id.

src/inmanta/deploy/scheduler.py
```diff
@@ -848,46 +1025,54 @@ async def get_resource_intent(self, resource: ResourceIdStr) -> Optional[Resourc
...
+            await self.update_scheduler_state_for_finished_deploy(attribute_hash, result, deployment_result)
+        except StaleResource:
+            # The resource is no longer managed, no need to update the database state.
+            pass
+        else:
+            # Write deployment result to the database.
+            await self.send_deploy_done(attribute_hash, result, deployment_result)
+
+    async def update_scheduler_state_for_finished_deploy(
+        self, attribute_hash: str, result: DeployResult, deployment_result: DeploymentResult
```
sanderr, line 1067: Any reason you added the `DeploymentResult` back? Merge conflict gone wrong? It was quite intentionally removed. Let me know if you'd like me to look into restoring this method to its intended state.

src/inmanta/deploy/scheduler.py
sanderr: Tiny preference to keep using `resource` for the name though. It's what the rest of this module does.

src/inmanta/deploy/scheduler.py
```diff
@@ -923,27 +1109,27 @@ async def deploy_done(
...
+                self._state.dirty.discard(resource_id)
             else:
                 # In most cases it will already be marked as dirty but in rare cases the deploy that just finished might
                 # have been triggered by an event, on a previously successful deployed resource. Either way, a failure
                 # (or skip) causes it to become dirty now.
-                self._state.dirty.add(resource)
+                self._state.dirty.add(resource_id)
 
             # propagate events
             self._send_events(details, recovered_from_failure=recovered_from_failure)
```
sanderr, line 1140: TODO: note to self: consider event timestamps for recovery

src/inmanta/deploy/scheduler.py
```diff
@@ -848,46 +1025,54 @@ async def get_resource_intent(self, resource: ResourceIdStr) -> Optional[Resourc
...
+        try:
+            await self.update_scheduler_state_for_finished_deploy(attribute_hash, result, deployment_result)
+        except StaleResource:
+            # The resource is no longer managed, no need to update the database state.
+            pass
+        else:
+            # Write deployment result to the database.
+            await self.send_deploy_done(attribute_hash, result, deployment_result)
+
+    async def update_scheduler_state_for_finished_deploy(
```
sanderr, line 1066:
```suggestion
    async def _update_scheduler_state_for_finished_deploy(
```

src/inmanta/deploy/scheduler.py
sanderr: Ok, I see why you did it. The reason it was removed was because it coupled the deployment result and the status (only certain combinations are actually valid). In this implementation the issue exists between result and deployment result. Since it is now an internal method it's not as bad, but could you add to the docstring that the method expects `result` and `deployment_result` to be consistent?

Or well, the conversion is very cheap, so perhaps we can simply do it both withing this method and within the `send_deploy_done`? I think I'd prefer that, but if you don't like it let's stick with documenting it.

src/inmanta/deploy/scheduler.py
sanderr: I think the initial motivation for doing it under the lock had to do with the write of the compliance status that might conflict with the write of a new scheduler state version. But iirc (I haven't reviewed that part yet), we discussed this and concluded that we'd write the last released and last deployed attribute hashes, from which the staleness of compliance status could be derived at recovery time. So if that is indeed the case, then I think that we can't conflict with new version and we can't conflict with other resources so then I don't think we need to do this under the lock.

sanderr (changes requested):

src/inmanta/deploy/scheduler.py
```diff
@@ -298,25 +301,95 @@ async def _initialize(self) -> None:
         Initialize the scheduler state and continue the deployment where we were before the server was shutdown.
         """
         self._timer_manager.initialize()
```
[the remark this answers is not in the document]
sanderr: d4f78b465

PR #8374, opened 2024-11-19
Fix resource state

wouterdb:
# Description

Fix state machine

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -44,7 +44,7 @@ from inmanta.deploy.state import (
     ModelState,
     ResourceDetails,
     ResourceState,
-    ResourceStatus,
+    ComplianceStatus,
 )
 from inmanta.deploy.tasks import Deploy, DryRun, RefreshFact
 from inmanta.deploy.work import PrioritizedTask, TaskPriority
@@ -107,7 +107,7 @@ class TaskManager(StateUpdateManager, abc.ABC):
         resource: ResourceIdStr,
         *,
         attribute_hash: str,
-        status: Optional[ResourceStatus] = None,
+        status: Optional[ComplianceStatus] = None,
         deployment_result: Optional[DeploymentResult] = None,
     ) -> None:
         """
@@ -472,7 +472,7 @@ class ResourceScheduler(TaskManager):
                     self._work.delete_resource(resource)
                 elif resource in self._state.resources:
                     # It's a resource we know.
-                    if self._state.resource_state[resource].status is ResourceStatus.UNDEFINED:
+                    if self._state.resource_state[resource].status is ComplianceStatus.UNDEFINED:
                         # The resource has been undeployable in previous versions, but not anymore.
                         unblocked_resources.add(resource)
                     elif details.attribute_hash != self._state.resources[resource].attribute_hash:
@@ -632,7 +632,7 @@ class ResourceScheduler(TaskManager):
         resource: ResourceIdStr,
         *,
         attribute_hash: str,
-        status: Optional[ResourceStatus] = None,
+        status: ComplianceStatus,
         deployment_result: Optional[DeploymentResult] = None,
     ) -> None:
         if deployment_result is DeploymentResult.NEW:
@@ -700,11 +700,11 @@ class ResourceScheduler(TaskManager):
         for dep_id in dependencies:
             resource_state_object: ResourceState = self._state.resource_state[dep_id]
             match resource_state_object:
-                case ResourceState(status=ResourceStatus.UNDEFINED):
+                case ResourceState(status=ComplianceStatus.UNDEFINED):
                     dependencies_state[dep_id] = const.ResourceState.undefined
                 case ResourceState(blocked=BlockedStatus.YES):
                     dependencies_state[dep_id] = const.ResourceState.skipped_for_undefined
-                case ResourceState(status=ResourceStatus.HAS_UPDATE):
+                case ResourceState(status=ComplianceStatus.HAS_UPDATE):
                     dependencies_state[dep_id] = const.ResourceState.available
                 case ResourceState(deployment_result=DeploymentResult.SKIPPED):
                     dependencies_state[dep_id] = const.ResourceState.skipped
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -641,11 +641,17 @@ class ResourceScheduler(TaskManager):
         async with self._scheduler_lock:
             # refresh resource details for latest model state
             details: Optional[ResourceDetails] = self._state.resources.get(resource, None)
-            if details is None or details.attribute_hash != attribute_hash:
-                # The reported resource state is for a stale resource and therefore no longer relevant for state updates.
-                # There is also no need to send out events because a newer version will have been scheduled.
-                return
             state: ResourceState = self._state.resource_state[resource]
+
+            if details is None or details.attribute_hash != attribute_hash:
+                # We are stale but still the last deploy
+                # We can update the deployment_result (which is about last deploy)
+                # We can't update status (which is about active state only)
+                # None of the event propagation or other update happen either for the same reason
+                if deployment_result is not None:
+                    state.deployment_result = deployment_result
+                return
+
             if status is not None:
                 state.status = status
             if deployment_result is not None:
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -641,9 +641,14 @@ class ResourceScheduler(TaskManager):
         async with self._scheduler_lock:
             # refresh resource details for latest model state
             details: Optional[ResourceDetails] = self._state.resources.get(resource, None)
+
+            if details is None:
+                # we are stale and removed
+                return
+
             state: ResourceState = self._state.resource_state[resource]
 
-            if details is None or details.attribute_hash != attribute_hash:
+            if details.attribute_hash != attribute_hash:
                 # We are stale but still the last deploy
                 # We can update the deployment_result (which is about last deploy)
                 # We can't update status (which is about active state only)
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -119,7 +119,7 @@ class TaskManager(StateUpdateManager, abc.ABC):
         :param resource: The resource to report state for.
         :param attribute_hash: The resource's attribute hash for which this state applies. No scheduler state is updated if the
             hash indicates the state information is stale.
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -2166,13 +2166,13 @@ class BadLoggingR(Resource):
...
-    @resource("test::Deploy", agent="agent", id_attribute="key")
-    class DeployR(Resource):
+    @resource("test::LSMLike", agent="agent", id_attribute="key")
+    class LsmLike(Resource):
         """
         Raise a SkipResource exception in the deploy() handler method.
         """
 
-        fields = ("key", "value", "set_state_to_deployed", "purged")
+        fields = ("key", "value", "purged", "purged")
```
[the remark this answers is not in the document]
wouterdb: thx!

wouterdb:
> I'm not sure how this fixes the issues that are occurring now. The translation between new state to the old remains the same.

@jptrindade I refined the state machine so the `HAS_UPDATE` no longer shadows the `skipped` state because skipped it now NON_COMPLIANT

src/inmanta/deploy/scheduler.py
```diff
@@ -641,11 +641,17 @@ async def report_resource_state(
...
+
             if details is None or details.attribute_hash != attribute_hash:
-                # The reported resource state is for a stale resource and therefore no longer relevant for state updates.
-                # There is also no need to send out events because a newer version will have been scheduled.
+                # We are stale but still the last deploy
+                # We can update the deployment_result (which is about last deploy)
+                # We can't update status (which is about active state only)
+                # None of the event propagation or other update happen either for the same reason
+                if deployment_result is not None:
+                    state.deployment_result = deployment_result
```
sanderr, line 657: We could lift this out of the condition to execute first (and remove the other one further down). But I can also see how you might want to have the comment tailored to this case.

src/inmanta/deploy/state.py
```diff
@@ -283,13 +283,13 @@ def update_blocked_status(resource: ResourceIdStr) -> bool:
                 return False
             # Unblock resource
             self.resource_state[resource].blocked = BlockedStatus.NO
-            if self.resource_state[resource].status is ResourceStatus.HAS_UPDATE:
+            if self.resource_state[resource].status is ComplianceStatus.HAS_UPDATE:
```
sanderr, line 286: I think we may want to include `NON_COMPLIANT` here now? It's probably relatively rare that that would happen, but it is possible e.g. when a resource that was failing becomes blocked as a result of an (undefined) update to one of its dependencies.

jptrindade (approved):

src/inmanta/deploy/scheduler.py
```diff
@@ -641,9 +641,14 @@ async def report_resource_state(
         async with self._scheduler_lock:
             # refresh resource details for latest model state
             details: Optional[ResourceDetails] = self._state.resources.get(resource, None)
+
+            if details is None:
+                # we are stale and removed
+                return
```
sanderr, line 647: Good catch

src/inmanta/deploy/tasks.py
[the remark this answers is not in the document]
sanderr: That is correct, but that is more a result of the fact that we change the semantics of the field. The core of the change is that we will now overwrite `HAS_UPDATE` with either `COMPLIANT` or `NON_COMPLIANT` as soon as we finish a deploy attempt. Previously we would only overwrite with `UP_TO_DATE` (equivalent of `COMPLIANT`), but we could not express that we'd tried a deploy and failed, so we kept it in `HAS_UPDATE` because it still has an update *relative to the operational state*. Now we express that part as non-compliant, and reserve `HAS_UPDATE` for an update *relative to the latest deploy attempt*.

This change then neatly makes `HAS_UPDATE | UNDEFINED` the sole qualifier to know whether the resource's deployment result represents the latest intent, or the result of some older deploy. Meaning we can safely use it in the match-case statement you added, because we step out of `HAS_UPDATE` as soon as we *attempt* to deploy.

Hope that's clear. It's tough to put into words. I repeat myself a bit above, but I hope that by phrasing it slightly differently I might get the point across.

tests/deploy/test_scheduler_agent.py
```diff
@@ -210,18 +210,20 @@ async def join(self, thread_pool_finalizer: list[ThreadPoolExecutor], timeout: f
...
     const.ResourceState.dry: (None, None, None),  # don't care
     const.ResourceState.deployed: (state.DeploymentResult.DEPLOYED, state.BlockedStatus.NO, None),
     const.ResourceState.failed: (state.DeploymentResult.FAILED, state.BlockedStatus.NO, None),
     const.ResourceState.deploying: (None, state.BlockedStatus.NO, None),
-    const.ResourceState.available: (None, state.BlockedStatus.NO, state.ResourceStatus.HAS_UPDATE),
-    const.ResourceState.undefined: (None, state.BlockedStatus.NO, state.ResourceStatus.UNDEFINED),
-    const.ResourceState.skipped_for_undefined: (None, state.BlockedStatus.YES, state.ResourceStatus.UNDEFINED),
+    const.ResourceState.available: (None, state.BlockedStatus.NO, state.ComplianceStatus.HAS_UPDATE),
+    const.ResourceState.undefined: (None, state.BlockedStatus.NO, state.ComplianceStatus.UNDEFINED),
+    const.ResourceState.skipped_for_undefined: (None, state.BlockedStatus.YES, state.ComplianceStatus.UNDEFINED),
```
sanderr, line 226: I think this is actually wrong (was before your PR), or at least overly specific.
```suggestion
    const.ResourceState.skipped_for_undefined: (None, state.BlockedStatus.YES, None),
```

tests/deploy/test_scheduler_agent.py
```diff
@@ -210,18 +210,20 @@ async def join(self, thread_pool_finalizer: list[ThreadPoolExecutor], timeout: f
...
 
-state_translation_table: dict[const.ResourceState, Tuple[state.DeploymentResult, state.BlockedStatus, state.ResourceStatus]] = {
+state_translation_table: dict[
+    const.ResourceState, Tuple[state.DeploymentResult, state.BlockedStatus, state.ComplianceStatus]
+] = {
     # A table to translate the old states into the new states
     # None means don't care, mostly used for values we can't derive from the old state
-    const.ResourceState.unavailable: (None, state.BlockedStatus.NO, state.ResourceStatus.UNDEFINED),
+    const.ResourceState.unavailable: (None, state.BlockedStatus.NO, state.ComplianceStatus.UNDEFINED),
     const.ResourceState.skipped: (None, state.BlockedStatus.YES, None),
```
sanderr, line 219: This one too
```suggestion
    const.ResourceState.skipped: (state.DeploymentResult.SKIPPED, None, None),
```

tests/deploy/test_scheduler_agent.py
```diff
@@ -210,18 +210,20 @@ async def join(self, thread_pool_finalizer: list[ThreadPoolExecutor], timeout: f
...
 
 
-state_translation_table: dict[const.ResourceState, Tuple[state.DeploymentResult, state.BlockedStatus, state.ResourceStatus]] = {
+state_translation_table: dict[
+    const.ResourceState, Tuple[state.DeploymentResult, state.BlockedStatus, state.ComplianceStatus]
+] = {
     # A table to translate the old states into the new states
     # None means don't care, mostly used for values we can't derive from the old state
-    const.ResourceState.unavailable: (None, state.BlockedStatus.NO, state.ResourceStatus.UNDEFINED),
+    const.ResourceState.unavailable: (None, state.BlockedStatus.NO, state.ComplianceStatus.UNDEFINED),
```
sanderr, line 218: I think this one too, but less sure.
```suggestion
    const.ResourceState.unavailable: (None, state.BlockedStatus.NO, state.ComplianceStatus.FAILED),
```

tests/deploy/test_scheduler_agent.py
```diff
@@ -210,18 +210,20 @@ async def join(self, thread_pool_finalizer: list[ThreadPoolExecutor], timeout: f
...
     const.ResourceState.skipped: (None, state.BlockedStatus.YES, None),
     const.ResourceState.dry: (None, None, None),  # don't care
     const.ResourceState.deployed: (state.DeploymentResult.DEPLOYED, state.BlockedStatus.NO, None),
     const.ResourceState.failed: (state.DeploymentResult.FAILED, state.BlockedStatus.NO, None),
     const.ResourceState.deploying: (None, state.BlockedStatus.NO, None),
-    const.ResourceState.available: (None, state.BlockedStatus.NO, state.ResourceStatus.HAS_UPDATE),
-    const.ResourceState.undefined: (None, state.BlockedStatus.NO, state.ResourceStatus.UNDEFINED),
-    const.ResourceState.skipped_for_undefined: (None, state.BlockedStatus.YES, state.ResourceStatus.UNDEFINED),
+    const.ResourceState.available: (None, state.BlockedStatus.NO, state.ComplianceStatus.HAS_UPDATE),
+    const.ResourceState.undefined: (None, state.BlockedStatus.NO, state.ComplianceStatus.UNDEFINED),
```
sanderr, line 225:
```suggestion
    const.ResourceState.undefined: (None, state.BlockedStatus.YES, state.ComplianceStatus.UNDEFINED),
```

tests/deploy/test_scheduler_agent.py
```diff
@@ -1139,8 +1141,8 @@ def make_resources(
     assert agent.executor_manager.executors["agent3"].execute_count == 0
     # verify that r2 is considered dirty now, despite being in an assumed good operational state
     assert agent.scheduler._state.resource_state[rid2] == state.ResourceState(
-        # we have no data indicating that operational state has deviated from desired state => still UP_TO_DATE
-        status=state.ResourceStatus.UP_TO_DATE,
+        # We are skipped, so not compliant
+        status=state.ComplianceStatus.NON_COMPLIANT,
```
sanderr, line 1145: I have to consider this change some more. It's possible that it now deviates from the intent of the test, and that we should set up a different scenario instead (if this edge case is still possible that is).

sanderr (changes requested):
Some more notes, still have to review e2e test but I've got to go now.

src/inmanta/deploy/state.py
wouterdb: thx, completely missed that.

tests/deploy/test_scheduler_agent.py
sanderr: The scenario is still relevant, just slightly less of an edge case. But I think we should update the comment two lines up that says "despite being in an assumed good operational state".

sanderr (changes requested):
One last tiny change request, no need to ask re-review.

sanderr (approved):

Issue #8185, opened 2024-10-09
Strip out code reloading 


wouterdb:
Remove the ability  of the agent to reload code

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/inmanta-core/8185?planning-poker) with Zenhub @arnaudsjs @hlloreda @Hugo-Inmanta @jptrindade

PR #8375, opened 2024-11-19
remove code reloading

Hugo-Inmanta:
# Description

* Short description here *

closes https://github.com/inmanta/inmanta-core/issues/8185

# Self Check:

- [x] Changelog entry


tests/test_loader.py
```diff
@@ -225,11 +201,14 @@ def test():
...
 
     with pytest.raises(ImportError):
-        import inmanta_bad_unit_test  # NOQA
+        import inmanta_plugins.inmanta_bad_unit_test  # NOQA
 
     cl.deploy_version([get_module_source("inmanta_plugins.inmanta_bad_unit_test", code)])
 
-    assert "ModuleNotFoundError: No module named 'badimmport'" in caplog.text
+    with pytest.raises(ModuleNotFoundError):
+        import inmanta_plugins.inmanta_bad_unit_test  # NOQA
```
Hugo-Inmanta, line 209: This was previously relying on the fact that the reloading would try importing, we have to explicitly call `import` now.

tests/test_loader.py
```diff
@@ -225,11 +212,14 @@ def test():
...
+
+        assert "ModuleNotFoundError: No module named 'badimmport'" in caplog.text
```
arnaudsjs, line 222: We have to do `caplog.clear()` before the second import I guess?

arnaudsjs (changes requested):

jptrindade (approved):

arnaudsjs (approved):

jptrindade (approved):

Issue #8340, opened 2024-11-08
Differentiate between skipped resources triggered by the default handler and skipped resources triggered by a custom handler


jptrindade:
Context: https://inmanta.slack.com/archives/CKRF0C8R3/p1730994402296499
# Eventual goal

- default handler skip would become a new exception type. This would be marked in the scheduler as (deployment_result=skipped, blocked=transient)
- custom handler skip would be raised as it is now. This would be marked in the scheduler as (deployment_result=skipped, blocked=no).

Why the difference? For the default handler skip we can make two assumptions:
- it does not make sense to retry until dependencies become successful
- it is possible that the resource may get deployed successfully later within this same model version (no new export required)

sanderr:
Discussed in planning meeting: the new type of skip exception should also be available to raise by custom handlers, with a clear indication of what it means (transient based on dependencies' state).

jptrindade:
Waiting for:
https://github.com/inmanta/inmanta-core/pull/8374
https://github.com/inmanta/inmanta-core/pull/8292
https://github.com/inmanta/inmanta-core/pull/8276

start with the part that is independent of the scheduler first (introduce the exception, raise it where appropriate, make sure it is correctly caught and propagated by the executors, i.e. that inmanta.deploy.tasks.Deploy.execute can see it.

PR #8383, opened 2024-11-20
Differentiate between custom and default handler skip

jptrindade:
# Description

Add a different kind of skip exception that marks the blocked status as transient
Adjust the blocked logic accordingly

closes: #8340 

# Self Check:

- [x] Changelog entry


src/inmanta/agent/handler.py
```diff
@@ -288,6 +299,10 @@ def action_id(self) -> uuid.UUID:
     def status(self) -> Optional[const.ResourceState]:
         return self._status
 
+    @property
+    def blocked(self) -> Optional[BlockedStatus]:
+        return self._blocked
+
```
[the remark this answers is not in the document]
sanderr: I think I see. Is it to be able to carry the differentiation over the wire to the scheduler? If so, I think it might be cleaner to simply add a new state to `const.ResourceState`. On the other hand, if we do that we have to carefully check where we expose values of that enum and whether we need to take action anywhere.

@wouterdb thoughts?

src/inmanta/deploy/state.py
```diff
@@ -194,13 +196,15 @@ def block_resource(self, resource: ResourceIdStr, details: ResourceDetails, is_t
         if resource in self.resource_state:
             if not is_transitive:
                 self.resource_state[resource].status = ComplianceStatus.UNDEFINED
-            self.resource_state[resource].blocked = BlockedStatus.YES
+                self.resource_state[resource].blocked = BlockedStatus.YES
```
sanderr, line 199: I don't think this is correct. Transient is not the same as transitive, and don't think we want to call this method at all for transiently blocked resources. We may want to rename it and/or clarify its docstring.

To clarify:
- transitive means we consider this resource blocked (`YES`) because it depends on another blocked (`YES`) resource. Somewhere at the head of that chain will be an `UNDEFINED` resource.
- transient means that the blocking might be short-lived, i.e. we can attempt to recover under certain conditions. This has nothing to do with `UNDEFINED`, and we do not transitively block resources because one of there dependencies is marked as transiently blocked (in a way we do, but we leave it up to the handler to drive the propagation).

src/inmanta/deploy/scheduler.py
```diff
@@ -634,6 +638,7 @@ async def report_resource_state(
         attribute_hash: str,
         status: ComplianceStatus,
         deployment_result: Optional[DeploymentResult] = None,
+        blocked_status: Optional[BlockedStatus] = None,
```
sanderr, line 641: I think it would be cleaner to derive this from `deployment_result` here. Related to my comment on `handler.py`.

src/inmanta/deploy/scheduler.py
```diff
@@ -672,7 +677,7 @@ async def report_resource_state(
                     # (or skip) causes it to become dirty now.
                     self._state.dirty.add(resource)
                 # propagate events
-                if details.attributes.get(const.RESOURCE_ATTRIBUTE_SEND_EVENTS, False):
+                if details.attributes.get(const.RESOURCE_ATTRIBUTE_SEND_EVENTS, False) and blocked_status != BlockedStatus.YES:
```
sanderr, line 680: Is this meant to check the target's blocked status instead? If so, I think @Hugo-Inmanta is incorperating that in his PR?

What we do need in this method is to set the resource's `BlockedStatus` depending on the deployment result -> either `NO` or `TRANSIENT`, but never overwriting `YES`.

sanderr (commented):
Had a very brief look at it and added some comments. One other note (haven't checked myself yet) is that we should have a critical look at `BlockedStatus.is_blocked()`, where it's used, and where direct comparisons on `BlockedStatus` are used. Some might care about `YES` while others really care about `not NO`.

src/inmanta/deploy/scheduler.py
jptrindade: It was related to this comment:
> force event propagation only for transiently blocked resources, not YES blocked ones

sanderr: Operative word here is "force". That's about what Hugo is working on. I could've been more explicit on that.

src/inmanta/deploy/state.py
jptrindade: Reverted these changes

src/inmanta/agent/handler.py
```diff
@@ -806,6 +815,13 @@ def execute(self, ctx: HandlerContext, resource: TResource, dry_run: bool = Fals
                     ctx.set_status(const.ResourceState.deployed)
             else:
                 ctx.set_status(const.ResourceState.dry)
+        except SkipResourceForDependencies as e:
+            ctx.set_status(const.ResourceState.transiently_skipped)
+            ctx.warning(
+                msg="Resource %(resource_id)s was skipped using the default exception: %(reason)s",
+                resource_id=resource.id,
+                reason=e.args,
+            )
```
wouterdb, line 824: I would use the same message as for the other one

src/inmanta/const.py
```diff
@@ -35,6 +35,7 @@ class ResourceState(str, Enum):
     cancelled = "cancelled"  # When a new version is pushed, in progress deploys are cancelled
     undefined = "undefined"  # The state of this resource is unknown at this moment in the orchestration process
     skipped_for_undefined = "skipped_for_undefined"  # This resource depends on an undefined resource
+    transiently_skipped = "transiently_skipped"  # This resource is transient and dependent on its dependencies state.
```
wouterdb, line 38: I don't like this name, every skip is transient,... (i.e. skip is a form of transient failure)

Perhaps `skipped_for_dependencies`

src/inmanta/agent/handler.py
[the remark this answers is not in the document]
wouterdb: We can't change const.ResourceState I think.

We are phasing it out, it leaks out of the api on all sides, I would not do that.

wouterdb (changes requested):

src/inmanta/agent/handler.py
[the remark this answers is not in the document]
sanderr: How do we see this handler context evolving over time? It wouldn't have the `const.ResourceState` anymore, so how would it communicate that it wants to skip? I ask because I'm hesitant to add a field now, one that feels out of place, unless it will come to fit into the future plans.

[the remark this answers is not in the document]
jptrindade: removed the new entry in `const.ResourceState` and created `ContextResourceState` to address this issue.

src/inmanta/agent/handler.py
```diff
@@ -288,15 +313,37 @@ def action_id(self) -> uuid.UUID:
...
+        Set the status of the handler operation and translate it to ContextResourceState
         """
         self._status = status
+        try:
+            self._resource_state = ContextResourceState(status)
+        except ValueError:
+            # How to proceed? What about unavailable?
+            self.logger.warning(
+                "Called set_status with status %s which is not directly translatable to NewResourceState", status
+            )
```
wouterdb, line 335: 1. This will become a failure, which is fine I think
2. I don't think unavailable can/should ever be set here, as it indicates we can't load the handler, so no ctx is constructed I think?

src/inmanta/deploy/tasks.py
```diff
@@ -192,11 +193,14 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
...
-                        case const.ResourceState.deployed:
+                    match deploy_result.resource_state:
+                        case ContextResourceState.deployed:
                             scheduler_deployment_result = state.DeploymentResult.DEPLOYED
-                        case const.ResourceState.skipped:
+                        case ContextResourceState.skipped:
+                            scheduler_deployment_result = state.DeploymentResult.SKIPPED
+                        case ContextResourceState.skipped_for_dependency:
                             scheduler_deployment_result = state.DeploymentResult.SKIPPED
+                            skipped_for_dependency = True
```
wouterdb, line 203: I wonder if we should encode this into the DeploymentResult as well?

(purely a question, I don't want to disturb this delicate machinery unduely)

wouterdb (approved):

wouterdb:
I'm not super confident on this, it is getting a bit subtle atm

src/inmanta/deploy/scheduler.py
```diff
@@ -670,6 +673,8 @@ async def report_resource_state(
                     state.deployment_result = deployment_result
                 return
 
+            if skipped_for_dependency:
+                state.blocked = BlockedStatus.TRANSIENT
```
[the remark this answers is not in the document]
sanderr: Finally, we should also drop it from `self._state.dirty` (already done below when the resource deploys successfully but should now also be done if we block the resource).

The drawback is that we also have to correctly unset the blocked status (back to `NO` and mark it as dirty again) when appropriate, because otherwise we'll never deploy it anymore. This part is missing from your PR atm.

I can think of two cases to consider:
1. new version comes in with one or more requires dropped
2. a dependency deploys successfully. We may choose to delegate this one to @Hugo-Inmanta's PR because I believe he already calculates this set. If we do, we should also delegate the "mark as dirty" so we don't get a stuck orchestrator in between both PRs. I'll leave it up to you to determine who does which part, considering double work and merge conflicts.

sanderr (changes requested):
General idea looks good to me now but some things are still missing in the scheduler part. Haven't reviewed tests yet.

src/inmanta/agent/handler.py
```diff
@@ -288,15 +313,37 @@ def action_id(self) -> uuid.UUID:
...
+        return self._resource_state
+
     @property
     def logs(self) -> list[data.LogLine]:
         return self._logs
 
     def set_status(self, status: const.ResourceState) -> None:
         """
-        Set the status of the handler operation.
+        Set the status of the handler operation and translate it to ContextResourceState
```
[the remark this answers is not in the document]
jptrindade: the old _status is the `const.ResourceState` while the new `_resource_state` is the `HandlerResourceState`.
`_resource_state` is the selection of states that we expect to see/set on the handler which is a subset of `const.ResourceState` with the inclusion of `skipped_for_dependency`. 
We translate 1:1  `_resource_state` to `status` (making `skipped_for_dependency`, `skipped`)
And we translate `_status` to `resource_state` when we can, otherwise we set `_resource_state` as `failed` because, to my understanding, we don't expect to see the other states in the handler.

Related to this discussion:
https://inmanta.slack.com/archives/CKRF0C8R3/p1732523218360489?thread_ts=1732522716.782239&cid=CKRF0C8R3

src/inmanta/agent/handler.py
```diff
@@ -221,6 +231,20 @@ def _log_msg(
         raise NotImplementedError
 
 
+class ContextResourceState(str, Enum):
+    """
+    Used by the HandlerContext to determine the state of the resource
+    Similar to const.ResourceState but with some states removed and with the added skipped_for_dependency state
```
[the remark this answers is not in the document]
jptrindade: Yes, that is my understanding of it. Thank you for pointing out this, I sometimes struggle to write proper docstrings. I'll start with the `is` instead of the `how` in the future

src/inmanta/agent/handler.py
```diff
@@ -221,6 +231,20 @@ def _log_msg(
...
+    """
+
+    skipped = "skipped"
+    deployed = "deployed"
+    failed = "failed"
+    dry = "dry"
+    skipped_for_dependency = "skipped_for_dependency"
+    unavailable = "unavailable"  # to be removed?
```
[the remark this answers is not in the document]
jptrindade: Yes, It was supposed to be a question to the reviewers. My thought process is that this will be caught and answered by the reviewers and then removed. But, like you mentioned, it is very likely that comments like this get lost and pollute the code. I'll start adding FIXMEs instead. (Wouter answered in another thread)

src/inmanta/agent/handler.py
```diff
@@ -83,6 +84,15 @@ class SkipResource(Exception):
     """
     A handler should raise this exception when a resource should be skipped. The resource will be marked as skipped
     instead of failed.
+    The resource will be skipped with a BlockedStatus of NO.
```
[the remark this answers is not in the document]
sanderr: I can't immediately think of any. It offers a broader range of cases where a handler developer could use it, but I'm also inclined to stick with your proposal. Mostly I think we have to make the decision (whatever it is) so that both sides know what to expect / provide.

src/inmanta/agent/handler.py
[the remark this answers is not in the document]
sanderr: Ok. Somehow I had completely missed that this method still set the old status field as well, that's what confused me. I think the current implementation is fine, with two small notes:
- I'd prefer to set the new status explicitly to failed, rather than to rely on what happens later
- remember to remove the comment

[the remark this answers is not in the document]
jptrindade: > I'd prefer to set the new status explicitly to failed, rather than to rely on what happens later

I'm not following this comment. What do you mean about explicitly setting the new status to failed and how does it differ from my approach?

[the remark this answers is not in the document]
sanderr: In the `except` block you now simply log a warning and stop there, leaving the new status at `None`.

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
jptrindade: I tried to follow this logic. I don't know if this what you have in mind. I'll add some comments

src/inmanta/deploy/scheduler.py
```diff
@@ -502,6 +504,11 @@ async def _new_version(
                     added_requires[resource] = added
                 if dropped:
                     dropped_requires[resource] = dropped
+                    if self._state.resource_state[resource].blocked is BlockedStatus.TRANSIENT:
+                        dependencies = await self._get_last_non_deploying_state_for_dependencies(resource=resource)
+                        if all(dependencies[requires_id].deployed for requires_id in new_requires):
+                            self._state.resource_state[resource].blocked = BlockedStatus.NO
+                            unblocked_resources.add(resource)
```
jptrindade, line 511: @sander Here is where we check the dropped requires. I make sure that all the requires are `deployed`, is this the correct check? Should we unblock them if they are in another non failing state (i.e. deploying or available)?

src/inmanta/deploy/scheduler.py
```diff
@@ -677,7 +685,16 @@ async def report_resource_state(
...
+                # Check if we need to mark a resource as transiently blocked
+                # Or if we can unblock a transiently blocked resource (its dependencies now succeed)
+                if skipped_for_dependency and state.blocked is not BlockedStatus.YES:
+                    dependencies = await self._get_last_non_deploying_state_for_dependencies(resource=resource)
+                    state.blocked = (
+                        BlockedStatus.NO
+                        if all(state is const.ResourceState.deployed for state in dependencies.values())
+                        else BlockedStatus.TRANSIENT
+                    )
+                if deployment_result is DeploymentResult.DEPLOYED or state.blocked is not BlockedStatus.NO:
```
jptrindade, line 697: and here is where we assign the `TRANSIENT` state when applicable and discard anything that is not unblocked

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
sanderr: We can also have a call about this if you prefer.

src/inmanta/agent/handler.py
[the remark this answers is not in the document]
jptrindade: https://github.com/inmanta/inmanta-core/pull/8383/files#diff-6fe603b0e2ead0d2899810275d0ed57935e40cd1daf59b0424fb3cc0d5c6e53dR323-R333

I now am also setting the status to failed.

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
jptrindade: Regarding this: 
>a dependency deploys successfully. We may choose to delegate this one to @Hugo-Inmanta's PR because I believe he already calculates this set. If we do, we should also delegate the "mark as dirty" so we don't get a stuck orchestrator in between both PRs. I'll leave it up to you to determine who does which part, considering double work and merge conflicts.

we should also delegate the part where we drop from the dirty set, right?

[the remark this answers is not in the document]
sanderr: Indeed. I see that I wasn't very clear on that.

src/inmanta/agent/handler.py
jptrindade: I'm starting to think that we need the unavailable state.
Here we can't find the handler but we have a context built: https://github.com/inmanta/inmanta-core/pull/8383/files#diff-f34628194310d50c9be6ca572bc1fccf6323eb8c0bf735df244e74b91e3e9b7eL163

And it is useful for when we call `DeployResult.undeployable` (assuming we want the state there to be `unavailable` instead of `failed`)

What do you think?

tests/test_server.py
```diff
@@ -1401,57 +1402,14 @@ async def test_send_deploy_done(server, client, environment, agent, caplog, meth
...
-                action_id=action_id,
-                status=const.ResourceState.deploying,
-                messages=[],
-                changes={"attr1": AttributeStateChange(current=None, desired="test")},
-                change=const.Change.created,
-            )
-        )
-    assert "No transient state can be used to mark a deployment as done" in str(exec_info.value)
-
-
```
jptrindade, line 1454: I removed this test because we no longer set the status of `DeployResult` like this. We use the `HandlerResourceState` which (I believe) does not have transient states

src/inmanta/deploy/scheduler.py
```diff
@@ -504,6 +504,11 @@ async def _new_version(
                     added_requires[resource] = added
                 if dropped:
                     dropped_requires[resource] = dropped
+                    if self._state.resource_state[resource].blocked is BlockedStatus.TRANSIENT:
+                        dependencies = await self._get_last_non_deploying_state_for_dependencies(resource=resource)
+                        if all(dependencies[requires_id].deployed for requires_id in new_requires):
+                            self._state.resource_state[resource].blocked = BlockedStatus.NO
+                            unblocked_resources.add(resource)
```
sanderr, line 511: `unblocked_resources` has a very different meaning, we must not add these. It's an understandable mistake to make though, given its name. So perhaps we should rename it. It made sense when there was only `YES` or `NO`, but now that we have `TRANSIENT`, the name should probably indicate that it has to do with definedness.

Perhaps `newly_defined` could be a good name? It sounds a bit clunky but it's the best I can come up with. The purpose of the list is to collect resources that were previously undefined but not anymore.

src/inmanta/deploy/scheduler.py
sanderr: Technically I think we want to wait for `ComplianceStatus.COMPLIANT` (not calling `_get_last_non_...()`), but I wonder if we might want to take the easy way out and simply unblock them (a bit too eagerly, which is a freedom we have: it *might* be unblocked, so we give it another shot) if `dropped` is not empty (i.e. at least one dependency was dropped). Main motivation: don't add even more complexity to this method for limited gains. Secondary motivation: performance while under scheduler lock, though this may be negligible (**important side note: we're not under scheduler lock here but we should be if we read from `self._state`, so if we keep the check this must be moved further down**).

I'm honestly not sure. The logic to do it accurately isn't too complex, but is it worth it, given that it might draw the focus from the important parts of this method? Perhaps we can get the best of both worlds if we incorporate it in `State.update_requires()`? But I'd only do it if it doesn't complicate that method too much. There's still the performance aspect but I think it should be acceptable.

What do you think?

src/inmanta/deploy/scheduler.py
```diff
@@ -673,15 +678,22 @@ async def report_resource_state(
...
             # We are not stale
             state.status = status
             if deployment_result is not None:
                 # first update state, then send out events
                 self._deploying_latest.remove(resource)
                 state.deployment_result = deployment_result
                 self._work.finished_deploy(resource)
+                # Check if we need to mark a resource as transiently blocked
+                # Or if we can unblock a transiently blocked resource (its dependencies now succeed)
+                if skipped_for_dependency and state.blocked is not BlockedStatus.YES:
```
sanderr, line 690: I was writing a comment about how I thought you were mixing up two scenarios here, until I realized that you're covering [this case](https://github.com/inmanta/inmanta-core/pull/8383#discussion_r1862339660) here. I think the implementation is good (with the same caveat as in `_new_version()` that I think we should check compliance instead), but I think it definitely needs a comment as to *why* we're doing this. i.e. if the task reports `skipped_for_dependency`, why don't we simply believe it?

src/inmanta/deploy/scheduler.py
```diff
@@ -673,15 +678,22 @@ async def report_resource_state(
...
+                    dependencies = await self._get_last_non_deploying_state_for_dependencies(resource=resource)
+                    state.blocked = (
+                        BlockedStatus.NO
+                        if all(state is const.ResourceState.deployed for state in dependencies.values())
+                        else BlockedStatus.TRANSIENT
+                    )
                 if deployment_result is DeploymentResult.DEPLOYED:
```
sanderr, line 697: If it does deploy successfully, we may want to unblock dependents, as mentioned in a previous comment. From the absence of that logic, do I understand correctly that you've passed on this part to Hugo's PR?

sanderr (changes requested):

src/inmanta/deploy/scheduler.py
jptrindade:
```
                if dropped:
                    dropped_requires[resource] = dropped
                    if self._state.resource_state[resource].blocked is BlockedStatus.TRANSIENT:
                            self._state.resource_state[resource].blocked = BlockedStatus.NO
```
If we do it like this, aren't we still reading from `self._state` and need to do it under `scheduler_lock`. Maybe we should do it next to update_requires, anyway.
I'm fine either way, leaning more towards just checking if all the dependencies are compliant.

sanderr: Yes, we need to do it under the lock either way, my performance remark was more that the less we do under the lock the faster we can release it again.

src/inmanta/deploy/scheduler.py
jptrindade: I think I can do it on my side easily.

jptrindade: I underestimated this. I'll talk with @Hugo-Inmanta tomorrow morning

sanderr: Let me know if you want me to join in.

src/inmanta/deploy/scheduler.py
```diff
@@ -496,20 +498,17 @@ async def _new_version(
...
                 new_requires: Set[ResourceIdStr] = requires.get(resource, set())
-                added: Set[ResourceIdStr] = new_requires - old_requires
-                dropped: Set[ResourceIdStr] = old_requires - new_requires
-                if added:
-                    added_requires[resource] = added
-                if dropped:
-                    dropped_requires[resource] = dropped
+                added_requires[resource] = new_requires - old_requires
+                dropped_requires[resource] = old_requires - new_requires
+
```
jptrindade, line 503: Is there maybe a reason that this was not done like this before?

src/inmanta/deploy/tasks.py
jptrindade: I did this in this commit: https://github.com/inmanta/inmanta-core/pull/8383/commits/a957c02cec8ced208a43ebd49a8841a944760dd7

I'm not sure if it changes the behaviour of `report_resource_state` in a way that we do not want to. What do you think?

src/inmanta/deploy/scheduler.py
```diff
@@ -736,10 +738,11 @@ async def _get_last_non_deploying_state_for_dependencies(
...
         """
         requires_view: Mapping[ResourceIdStr, Set[ResourceIdStr]] = self._state.requires.requires_view()
-        dependencies: Set[ResourceIdStr] = requires_view.get(resource, set())
+        dependencies: Mapping[ResourceIdStr, ResourceState] = {
+            dep_id: self._state.resource_state[dep_id] for dep_id in requires_view.get(resource, set())
+        }
         dependencies_state = {}
-        for dep_id in dependencies:
-            resource_state_object: ResourceState = self._state.resource_state[dep_id]
+        for dep_id, resource_state_object in dependencies.items():
```
sanderr, line 745: Why this change? I think it has a slight performance impact, so I'd only do it with good reason. Do you find it more readable?

src/inmanta/deploy/scheduler.py
```diff
@@ -677,7 +682,21 @@ async def report_resource_state(
...
+                if state.blocked is not BlockedStatus.YES:
+                    if (
+                        resource_state is const.HandlerResourceState.skipped_for_dependency
+                        or state.blocked is BlockedStatus.TRANSIENT
+                    ):
+                        state.blocked = (
+                            BlockedStatus.NO if self._state.resource_can_be_unblocked(resource) else BlockedStatus.TRANSIENT
+                        )
                 if deployment_result is DeploymentResult.DEPLOYED:
+                    # FIXME: Also discard blocked resources from the dirty set
```
sanderr, line 698:
```suggestion
                    # FIXME: Also discard blocked resources from the dirty set if we block it transiently
```

tests/test_server.py
sanderr: We didn't touch the update manager, did we?

sanderr (changes requested):
Final comments, all minor. I'm going to do one more quick sweep and a check of the test cases but I don't expect anything new to fall out.

Let's also have Wouter do one more pass for the latest changes.

tests/deploy/e2e/test_deploy.py
```diff
@@ -921,3 +921,45 @@ async def test_lsm_states(resource_container, server, client, clienthelper, envi
...
+            "purged": False,
+            "send_event": True,
+            "receive_events": False,
+        },
+    ]
+    await clienthelper.set_auto_deploy(True)
+    await clienthelper.put_version_simple(resources, version, wait_for_released=True)
+    await clienthelper.wait_for_deployed()
+    scheduler = agent.scheduler
+    assert scheduler._state.resource_state[rid2] == ResourceState(
```
sanderr, line 961: Can you also assert rid1 state to show the difference? i.e. I expect non compliant, skipped, not blocked.

sanderr (changes requested):
One more

tests/test_server.py
jptrindade: I believe we didn't. We just changed the definition of `DeployResult`, so this test does not make sense anymore

src/inmanta/deploy/scheduler.py
```diff
@@ -687,14 +684,19 @@ async def report_resource_state(
...
-                        BlockedStatus.NO
-                        if all(state is const.ResourceState.deployed for state in dependencies.values())
-                        else BlockedStatus.TRANSIENT
-                    )
+                # We might already be unblocked if a dependency succeeds on another agent
+                # so HandlerResourceState.skipped_for_dependency might be outdated
+                if state.blocked is not BlockedStatus.YES:
+                    if (
+                        resource_state is const.HandlerResourceState.skipped_for_dependency
+                        or state.blocked is BlockedStatus.TRANSIENT
```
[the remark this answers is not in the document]
jptrindade: (I've not removed it in this pass through, but I can still do it if you want)

tests/test_server.py
sanderr: Ah, you mean that it has a new field now? But as far as I can tell this method doesn't use the new field directly? Perhaps we should update it.

Either way, I think you're right that the test can go, but if we don't update the method it's built on an assumption that despite it being typed as the one enum, it will really only return what's in the other.

That said, if it proves difficult (I'm assuming a one-line change now), it's probably not worth it.

src/inmanta/deploy/scheduler.py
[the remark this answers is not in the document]
sanderr: Took this to [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1733397988252939).

src/inmanta/deploy/tasks.py
```diff
@@ -218,14 +207,14 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
...
-                    scheduler_deployment_result = state.DeploymentResult.FAILED
+                    failed_scheduler_deploy = True
             finally:
                 if deploy_result is not None:
                     # We signaled start, so we signal end
                     try:
                         await task_manager.send_deploy_done(deploy_result)
                     except Exception:
-                        scheduler_deployment_result = state.DeploymentResult.FAILED
+                        failed_scheduler_deploy = True
```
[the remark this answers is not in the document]
sanderr: It's subjective, but I personally find that a lot less readable: we perform an action, get a result back, and then to go and change that result object in one of the branches can be confusing. As I see it, that's the sort of flow where bugs easily hide in: you don't trivially expect it and when you glance over the code, it doesn't stand out, so you'll very easily miss it. Part of how I would define readable code is code that communicates its intent well. Communication means to put things out in the open, to offer no confusion, to have an easily traceable flow. A variable that is overridden somewhere in one specific branch creates an exception, a deviation from the standard. While adding a new variable makes the deviation part of the definition, i.e. we communicate that this case exists and that we handle it.

But, as I said above, it's very subjective. I don't like the style, but I also won't stand in the way.

sanderr (changes requested):
LGTM now. Requesting changes purely so that the final three in-progress threads don't get lost. Once we reach agreement there it's good to go in as far as I'm concerned.

tests/test_server.py
jptrindade: It has a new field (`resource_state`) which uses a different class (`HandlerResourceState`). We no longer support creating with the `status` field as it is read-only.
`HandlerResourceState`, by definition does not have transient states. So we can't actually call `send_deploy_done` with a transient state we would fail on creating `DeployResult` instead.
I'm not sure I understand this part:
> but if we don't update the method it's built on an assumption that despite it being typed as the one enum, it will really only return what's in the other.

src/inmanta/deploy/scheduler.py
```diff
@@ -673,16 +675,28 @@ async def report_resource_state(
...
+                # We might already be unblocked if a dependency succeeded on another agent, e.g. while waiting for the lock
+                # so HandlerResourceState.skipped_for_dependency might be outdated
+                if state.blocked is not BlockedStatus.YES:
+                    if (
+                        resource_state is const.HandlerResourceState.skipped_for_dependency
+                        or state.blocked is BlockedStatus.TRANSIENT
+                    ):
+                        state.blocked = (
+                            BlockedStatus.NO if self._state.are_dependencies_compliant(resource) else BlockedStatus.TRANSIENT
+                        )
```
wouterdb, line 696: I find this hard to understand.

What I think it means is 

```python
match state.blocked:
   case Yes: pass # blocked state can't be cleared here, I don't even think this code should be reachable?
   case Transient: 
        state.blocked = (
                            BlockedStatus.NO if self._state.are_dependencies_compliant(resource) else BlockedStatus.TRANSIENT
                        )
        # I don't understand how this is possible? the dependendencies would have updated us  no?
    case No:
           if resource_state is const.HandlerResourceState.skipped_for_dependency:
                      state.blocked = (
                            BlockedStatus.NO if self._state.are_dependencies_compliant(resource) else BlockedStatus.TRANSIENT
                        )
                     # I don't understand how self._state.are_dependencies_compliant(resource) can be false? Then we would have BlockedStatus.TRANSIENT?

```

src/inmanta/deploy/state.py
```diff
@@ -324,6 +336,17 @@ def update_requires(
         """
         Update the requires relation for a resource. Updates the reverse relation accordingly.
         """
+        # If the resource is blocked transiently, and we drop at least one of its requirements
+        # we check to see if the resource can now be unblocked
+        # i.e. all of its dependencies are now compliant with the desired state.
+        if (
+            self.resource_state[resource].blocked is BlockedStatus.TRANSIENT
+            and self.requires[resource] - requires
+            and self.are_dependencies_compliant(resource)
```
wouterdb, line 345: This is prior to the update of the requires, so it will include the removed ones, I think

wouterdb (changes requested):

src/inmanta/deploy/tasks.py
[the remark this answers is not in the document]
jptrindade: I added a `failed_to_send_deploy_done` variable to track this. I set it to true in this line and set it to false after the `task_manager.send_deploy_done(deploy_result)` (probably unneeded). Do you think I should declare it before the `if deploy_result is not None` ? I expected mypy to complain about it but it just accepted it.

Also, thank you for your thorough explanations and reasoning. I feel like they do make a difference in my thought process.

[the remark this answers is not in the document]
sanderr: mypy doesn't require it. I typically do so because I find the declaration is more explicit, but I don't feel too strongly about it.

src/inmanta/deploy/state.py
jptrindade: You are right. I updated this and added some more assertions to the test

src/inmanta/deploy/scheduler.py
jptrindade: I updated this check to remove the `or state.blocked is BlockedStatus.Transient` following the discussion with @sanderr .
Regarding your questions about the inconsistency between `resource_state` and `are_dependencies_compliant` they were motivated by this:
> this deploy might run concurrently with a successful deploy of a dependency (if it's on another agent). If the dependency gets the scheduler lock first, we're already unblocked again by the time we reach this point.

More context: https://github.com/inmanta/inmanta-core/pull/8383#discussion_r1862230622

tests/test_server.py
sanderr: I mean, the update manager still reads `status` instead of `resource_state`, right?

src/inmanta/deploy/state.py
sanderr: The fix is not necessarily correct either. I'm pretty sure the sets are mutable, so you're using the same set left and right of the `-` operator. I think you'll have to go for either:
- drop the method `are_dependencies_compliant` and inline it in both places because it's so simple
- two-stage it in this method: first check `dropped_requires: bool`, then update the requires, then check if dependencies are compliant and act on it.

tests/test_server.py
sanderr: So I'm arguing that it shouldn't. If it expects only the handler states, it should read only those. That wasn't possible before but this PR gives us the oppurtunity to clean that up.

sanderr (approved):
Could you hold off with the actual merge until tomorrow morning, so we can limit the fallout as discussed on Slack?

wouterdb (approved):

PR #8399, opened 2024-11-22
Dataclasses

wouterdb:
# Description

Adding dataclass support:

# What works
1. type validation
2. defining paired classes
3. using dataclass as return type

# TODO:

1. use as argument, probably requires a bit more magic to support round-trip (same object in and out)
2. clean exception
3. exception printing out both implementations
4. remove entity tracker (unrelated)
5. typing

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -15,10 +15,9 @@
 
     Contact: code@inmanta.com
 """
+
 import dataclasses
 import importlib
-# pylint: disable-msg=R0902,R0904
-
 from typing import Any, Dict, List, Optional, Set, Tuple, Union  # noqa: F401
 
 from inmanta.ast import (
@@ -37,6 +36,9 @@ from inmanta.ast.statements.generator import SubConstructor
 from inmanta.ast.type import Float, NamedType, NullableType, Type
 from inmanta.execute.runtime import Instance, QueueScheduler, Resolver, dataflow
 from inmanta.execute.util import AnyType, NoneValue
+
+# pylint: disable-msg=R0902,R0904
+
 
 try:
     from typing import TYPE_CHECKING
@@ -518,12 +520,11 @@ class Entity(NamedType, WithComment):
         # TODO: error reporting
         namespace = self.namespace.get_full_name()
 
-        dataclass_module = importlib.import_module("inmanta_plugins." + namespace.replace("::","."))
+        dataclass_module = importlib.import_module("inmanta_plugins." + namespace.replace("::", "."))
         dataclass = getattr(dataclass_module, self.name)
 
         assert dataclasses.is_dataclass(dataclass)
         assert dataclass.__dataclass_params__.frozen
-
 
         dc_fields = {f.name: f for f in dataclasses.fields(dataclass)}
 
@@ -545,12 +546,19 @@ class Entity(NamedType, WithComment):
                     assert py_type
                     assert issubclass(field.type, py_type)
 
-
-
         # Only std::none as implementation
+        for implement in self.get_implements():
+            for imp in implement.implementations:
+                assert imp.get_full_name() == "std::none"
 
+        # No index
 
-        #assert all(x.type == int for x in dataclasses.fields(dataclass))
+        assert not self.get_indices()
+
+        self._paired_dataclass = dataclass
+
+    def as_python_type(self) -> "typing.Type | None":
+        return self._paired_dataclass
 
 
 # Kept for backwards compatibility. May be dropped from iso7 onwards.
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -16,6 +16,7 @@
     Contact: code@inmanta.com
 """
 import dataclasses
+import importlib
 # pylint: disable-msg=R0902,R0904
 
 from typing import Any, Dict, List, Optional, Set, Tuple, Union  # noqa: F401
@@ -94,6 +95,7 @@ class Entity(NamedType, WithComment):
         self._paired_dataclass: Type[object] = None
 
     def normalize(self) -> None:
+        self.normalized = True
         for attribute in self.__default_values.values():
             if attribute.default is not None:
                 default_type: Type = attribute.type.get_type(self.namespace)
@@ -510,18 +512,45 @@ class Entity(NamedType, WithComment):
     def get_location(self) -> Location:
         return self.location
 
-    def pair_dataclass(self, dataclass: Type[object]) -> None:
+    def pair_dataclass(self) -> None:
         assert self.normalized
+
         # TODO: error reporting
+        namespace = self.namespace.get_full_name()
+
+        dataclass_module = importlib.import_module("inmanta_plugins." + namespace.replace("::","."))
+        dataclass = getattr(dataclass_module, self.name)
+
         assert dataclasses.is_dataclass(dataclass)
         assert dataclass.__dataclass_params__.frozen
 
-        # all own fields are primitive
-        # No relations
+
+        dc_fields = {f.name: f for f in dataclasses.fields(dataclass)}
+
+        for rel_or_attr_name in self.get_all_attribute_names():
+            rel_or_attr = self.get_attribute(rel_or_attr_name)
+            match rel_or_attr:
+                case inmanta.ast.attribute.RelationAttribute() as rel:
+                    # No relations except for requires and provides
+                    assert rel.entity.get_full_name() == "std::Entity", rel_or_attr_name
+                case inmanta.ast.attribute.Attribute() as attr:
+                    field = dc_fields.pop(rel_or_attr_name)
+                    inm_type = attr.get_type()
+
+                    # all own fields are primitive
+                    assert inm_type.is_primitive()
+
+                    # Type correspondence
+                    py_type = inm_type.as_python_type()
+                    assert py_type
+                    assert issubclass(field.type, py_type)
+
+
+
         # Only std::none as implementation
 
-        # Type correspondence
-        assert all(x.type == int for x in dataclasses.fields(test_data))
+
+        #assert all(x.type == int for x in dataclasses.fields(dataclass))
 
 
 # Kept for backwards compatibility. May be dropped from iso7 onwards.
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -15,7 +15,7 @@
 
     Contact: code@inmanta.com
 """
-
+import dataclasses
 # pylint: disable-msg=R0902,R0904
 
 from typing import Any, Dict, List, Optional, Set, Tuple, Union  # noqa: F401
@@ -90,6 +90,8 @@ class Entity(NamedType, WithComment):
         self.comment = comment
 
         self.normalized = False
+
+        self._paired_dataclass: Type[object] = None
 
     def normalize(self) -> None:
         for attribute in self.__default_values.values():
@@ -507,6 +509,19 @@ class Entity(NamedType, WithComment):
 
     def get_location(self) -> Location:
         return self.location
+
+    def pair_dataclass(self, dataclass: Type[object]) -> None:
+        assert self.normalized
+        # TODO: error reporting
+        assert dataclasses.is_dataclass(dataclass)
+        assert dataclass.__dataclass_params__.frozen
+
+        # all own fields are primitive
+        # No relations
+        # Only std::none as implementation
+
+        # Type correspondence
+        assert all(x.type == int for x in dataclasses.fields(test_data))
 
 
 # Kept for backwards compatibility. May be dropped from iso7 onwards.
--- a/src/inmanta/ast/type.py
+++ b/src/inmanta/ast/type.py
@@ -23,6 +23,7 @@ from collections.abc import Sequence
 from typing import Callable
 from typing import List as PythonList
 from typing import Optional
+import typing
 
 from inmanta.ast import DuplicateException, Locatable, LocatableString, Named, Namespace, NotFoundException, RuntimeException
 from inmanta.execute.util import AnyType, NoneValue, Unknown
@@ -107,6 +108,7 @@ class Type(Locatable):
         """
         Convert an instance of this type to its python form
         """
+
 
 class NamedType(Type, Named):
     def get_double_defined_exception(self, other: "NamedType") -> "DuplicateException":
@@ -419,6 +421,7 @@ class String(Primitive):
     def to_python(self, instance: object) -> "object":
         return str(instance)
 
+
 @stable_api
 class List(Type):
     """
@@ -456,6 +459,7 @@ class List(Type):
 
     def to_python(self, instance: object) -> "object":
         raise NotImplementedError()
+
 
 @stable_api
 class TypedList(List):
--- a/src/inmanta/ast/type.py
+++ b/src/inmanta/ast/type.py
@@ -96,6 +96,17 @@ class Type(Locatable):
         """
         return base_type
 
+    def as_python_type(self) -> "typing.Type | None":
+        """
+        Return a python type that can capture the values of this inmanta type
+
+        Returns None if this is not possible.
+        """
+
+    def to_python(self, instance: object) -> "object":
+        """
+        Convert an instance of this type to its python form
+        """
 
 class NamedType(Type, Named):
     def get_double_defined_exception(self, other: "NamedType") -> "DuplicateException":
@@ -145,6 +156,14 @@ class NullableType(Type):
         if not isinstance(other, NullableType):
             return NotImplemented
         return self.element_type == other.element_type
+
+    def as_python_type(self) -> "typing.Type | None":
+        return Optional[self.element_type.as_python_type()]
+
+    def to_python(self, instance: object) -> "object":
+        if instance is None:
+            return NoneValue()
+        return self.element_type.to_python(instance)
 
 
 @stable_api
@@ -229,6 +248,12 @@ class Number(Primitive):
     def type_string_internal(self) -> str:
         return self.type_string()
 
+    def as_python_type(self) -> "typing.Type | None":
+        return numbers.Number
+
+    def to_python(self, instance: object) -> "object":
+        return instance
+
 
 @stable_api
 class Float(Primitive):
@@ -265,6 +290,12 @@ class Float(Primitive):
     def type_string_internal(self) -> str:
         return self.type_string()
 
+    def as_python_type(self) -> "typing.Type | None":
+        return float
+
+    def to_python(self, instance: object) -> "object":
+        return instance
+
 
 @stable_api
 class Integer(Number):
@@ -290,6 +321,12 @@ class Integer(Number):
 
     def type_string(self) -> str:
         return "int"
+
+    def as_python_type(self) -> "typing.Type | None":
+        return int
+
+    def to_python(self, instance: object) -> "object":
+        return instance
 
 
 @stable_api
@@ -327,6 +364,12 @@ class Bool(Primitive):
 
     def get_location(self) -> None:
         return None
+
+    def as_python_type(self) -> "typing.Type | None":
+        return bool
+
+    def to_python(self, instance: object) -> "object":
+        return object
 
 
 @stable_api
@@ -370,6 +413,11 @@ class String(Primitive):
     def get_location(self) -> None:
         return None
 
+    def as_python_type(self) -> "typing.Type | None":
+        return str
+
+    def to_python(self, instance: object) -> "object":
+        return str(instance)
 
 @stable_api
 class List(Type):
@@ -403,6 +451,11 @@ class List(Type):
     def get_location(self) -> None:
         return None
 
+    def as_python_type(self) -> "typing.Type | None":
+        raise NotImplementedError()
+
+    def to_python(self, instance: object) -> "object":
+        raise NotImplementedError()
 
 @stable_api
 class TypedList(List):
@@ -452,6 +505,12 @@ class TypedList(List):
         if not isinstance(other, TypedList):
             return NotImplemented
         return self.element_type == other.element_type
+
+    def as_python_type(self) -> "typing.Type | None":
+        return list[self.element_type.as_python_type()]
+
+    def to_python(self, instance: object) -> "object":
+        return [self.element_type.to_python(instance) for element in instance]
 
 
 @stable_api
--- a/src/inmanta/compiler/__init__.py
+++ b/src/inmanta/compiler/__init__.py
@@ -247,8 +247,20 @@ class Compiler:
         )
         requires_rel.namespace = self.__root_ns.get_ns_from_string("std")
 
+        # Add std::DataClass
+        dataclass = DefineEntity(
+            ns,
+            LocatableString("Dataclass", nullrange, 0, ns),
+            LocatableString("The entity all other entities inherit from.", nullrange, 0, ns),
+            [
+                LocatableString("std::Entity", nullrange, 0, ns),
+            ],
+            [],
+        )
+
         statements.append(entity)
         statements.append(requires_rel)
+        statements.append(dataclass)
         return (statements, blocks)
 
     def export_data(self) -> None:
--- a/src/inmanta/plugins.py
+++ b/src/inmanta/plugins.py
@@ -826,3 +826,7 @@ def deprecated(
     if function is not None:
         return inner(function)
     return inner
+
+class Dataclass:
+
+    pass
--- a/src/inmanta/plugins.py
+++ b/src/inmanta/plugins.py
@@ -18,6 +18,7 @@
 
 import asyncio
 import collections.abc
+import dataclasses
 import inspect
 import os
 import subprocess
@@ -31,7 +32,7 @@ from inmanta.ast import LocatableString, Location, Namespace, Range, RuntimeExce
 from inmanta.ast.type import NamedType
 from inmanta.config import Config
 from inmanta.execute.proxy import DynamicProxy
-from inmanta.execute.runtime import QueueScheduler, Resolver, ResultVariable
+from inmanta.execute.runtime import Instance, QueueScheduler, Resolver, ResultVariable
 from inmanta.execute.util import NoneValue, Unknown
 from inmanta.stable_api import stable_api
 from inmanta.warnings import InmantaWarning
@@ -336,6 +337,26 @@ class PluginReturn(PluginValue):
 
     VALUE_TYPE = "returned value"
     VALUE_NAME = "return value"
+
+    def to_model_domain(self, value: object, resolver: Resolver, queue: QueueScheduler, location: Location) -> object:
+
+        if dataclasses.is_dataclass(value):
+            if self.resolved_type.as_python_type() == type(value):
+                instance = self.resolved_type.get_instance(
+                    value.__dict__,
+                    resolver,
+                    queue,
+                    location,
+                    None,
+                )
+                # generate an implementation
+                for stmt in self.resolved_type.get_sub_constructor():
+                    stmt.emit(instance, queue)
+                return instance
+            assert False
+
+        self.validate(value)
+        return value
 
 
 class Plugin(NamedType, WithComment, metaclass=PluginMeta):
@@ -682,6 +703,8 @@ class Plugin(NamedType, WithComment, metaclass=PluginMeta):
     def __call__(self, *args: object, **kwargs: object) -> object:
         """
         The function call itself
+
+        As a call, for backward compact
         """
         if self.deprecated:
             msg: str = f"Plugin '{self.get_full_name()}' is deprecated."
@@ -709,6 +732,42 @@ class Plugin(NamedType, WithComment, metaclass=PluginMeta):
         self.return_type.validate(value)
 
         return value
+
+    def call_in_context(
+        self,
+        args: Sequence[object],
+        kwargs: Mapping[str, object],
+        resolver: Resolver,
+        queue: QueueScheduler,
+        location: Location,
+    ) -> object:
+        """
+        The function call itself, with compiler context
+        """
+        if self.deprecated:
+            msg: str = f"Plugin '{self.get_full_name()}' is deprecated."
+            if self.replaced_by:
+                msg += f" It should be replaced by '{self.replaced_by}'."
+            warnings.warn(PluginDeprecationWarning(msg))
+        self.check_requirements()
+
+        def new_arg(arg: object) -> object:
+            if isinstance(arg, Context):
+                return arg
+            elif isinstance(arg, Unknown) and self.is_accept_unknowns():
+                return arg
+            else:
+                return DynamicProxy.return_value(arg)
+
+        new_args = [new_arg(arg) for arg in args]
+        new_kwargs = {k: new_arg(v) for k, v in kwargs.items()}
+
+        value = self.call(*new_args, **new_kwargs)
+
+        value = DynamicProxy.unwrap(value)
+
+        # Validate the returned value
+        return self.return_type.to_model_domain(value, resolver, queue, location)
 
     def get_full_name(self) -> str:
         return f"{self.ns.get_full_name()}::{self.__class__.__function_name__}"
--- a/src/inmanta/plugins.py
+++ b/src/inmanta/plugins.py
@@ -826,7 +826,3 @@ def deprecated(
     if function is not None:
         return inner(function)
     return inner
-
-class Dataclass:
-
-    pass
--- a/src/inmanta/ast/statements/call.py
+++ b/src/inmanta/ast/statements/call.py
@@ -262,7 +262,7 @@ class PluginFunction(Function):
             self.plugin(*args, **kwargs)
         else:
             try:
-                value = self.plugin(*args, **kwargs)
+                value = self.plugin.call_in_context(args, kwargs, resolver, queue, self.ast_node.location)
                 result.set_value(value if value is not None else NoneValue(), self.ast_node.location)
             except UnknownException as e:
                 result.set_value(e.unknown, self.ast_node.location)
--- a/src/inmanta/execute/scheduler.py
+++ b/src/inmanta/execute/scheduler.py
@@ -265,6 +265,11 @@ class Scheduler:
 
         self.types = {k: v for k, v in types_and_impl.items() if isinstance(v, Type)}
 
+        # Dataclass validation
+        data_class_root = self.types["std::Dataclass"]
+        for dataclass in data_class_root.get_all_child_entities():
+            dataclass.pair_dataclass()
+
     def get_anchormap(
         self, compiler: "Compiler", statements: Sequence["Statement"], blocks: Sequence["BasicBlock"]
     ) -> Sequence[tuple[Location, AnchorTarget]]:
--- a/tests/compiler/test_dataclasses.py
+++ b/tests/compiler/test_dataclasses.py
@@ -1,0 +1,16 @@
+import dataclasses
+
+from inmanta import compiler
+
+
+def test_dataclass_load(snippetcompiler):
+    snippetcompiler.setup_for_snippet(
+        """
+import dataclasses
+""",
+        ministd=True,
+    )
+    compiler.do_compile()
+
+
+dataclasses
--- a/tests/compiler/test_dataclasses.py
+++ b/tests/compiler/test_dataclasses.py
@@ -13,4 +13,15 @@ import dataclasses
     compiler.do_compile()
 
 
-dataclasses
+
+def test_dataclass_load_bad(snippetcompiler):
+    snippetcompiler.setup_for_snippet(
+        """
+import dataclasses::bad_sub
+""",
+        ministd=True,
+    )
+    compiler.do_compile()
+
+
+
--- a/tests/compiler/test_dataclasses.py
+++ b/tests/compiler/test_dataclasses.py
@@ -1,5 +1,7 @@
 import dataclasses
 
+import pytest
+
 from inmanta import compiler
 
 
@@ -13,7 +15,6 @@ import dataclasses
     compiler.do_compile()
 
 
-
 def test_dataclass_load_bad(snippetcompiler):
     snippetcompiler.setup_for_snippet(
         """
@@ -21,7 +22,5 @@ import dataclasses::bad_sub
 """,
         ministd=True,
     )
-    compiler.do_compile()
-
-
-
+    with pytest.raises(Exception):
+        compiler.do_compile()
--- a/tests/data/modules/dataclasses/model/_init.cf
+++ b/tests/data/modules/dataclasses/model/_init.cf
@@ -9,3 +9,9 @@ end
 implement Virtualmachine using std::none
 
 Virtualmachine(name="A", os="B", ram=1, cpus=2, disk=10)
+
+one=make_virtual_machine()
+
+one.requires = one
+
+std::print(one)
--- a/tests/data/modules/dataclasses/model/_init.cf
+++ b/tests/data/modules/dataclasses/model/_init.cf
@@ -1,0 +1,11 @@
+entity Virtualmachine extends std::Dataclass:
+    string name
+    string os
+    int ram
+    int cpus
+    int disk
+end
+
+implement Virtualmachine using std::none
+
+Virtualmachine(name="A", os="B", ram=1, cpus=2, disk=10)
--- a/tests/data/modules/dataclasses/model/bad_sub.cf
+++ b/tests/data/modules/dataclasses/model/bad_sub.cf
@@ -1,0 +1,11 @@
+entity Virtualmachine extends std::Dataclass:
+    string name
+    string os
+    int ram
+    int cpus
+    int disk
+end
+
+implement Virtualmachine using std::none
+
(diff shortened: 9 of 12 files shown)


src/inmanta/compiler/__init__.py
```diff
@@ -247,8 +247,20 @@ def compile(self) -> tuple[list["Statement"], list["BasicBlock"]]:
         )
         requires_rel.namespace = self.__root_ns.get_ns_from_string("std")
 
+        # Add std::DataClass
+        dataclass = DefineEntity(
```
wouterdb, line 251: to std

src/inmanta/ast/type.py
```diff
@@ -544,6 +713,33 @@ def type_string_internal(self) -> str:
...
+            return False
+        return self.element_type.corresponds_to(args[1])
+
+    def as_python_type_string(self) -> "str | None":
+        return f"dict[str, {self.element_type.as_python_type_string()}]"
+
+    def has_custom_to_python(self) -> bool:
+        return self.element_type.has_custom_to_python()
+
+    def to_python(self, instance: object) -> "object":
```
wouterdb, line 738: This can't be expressed currently at the plugin boundary

src/inmanta/plugins.py
```diff
@@ -337,6 +438,47 @@ class PluginReturn(PluginValue):
...
+            if isinstance(outer_type, inmanta_type.NullableType):
+                if isinstance(value, NoneValue):
+                    return value
+                outer_type = outer_type.element_type
+
+            if isinstance(outer_type, inmanta_type.List):
+                if isinstance(value, Sequence):
+                    return [make_dc(v) for v in value]
+                else:
+                    raise RuntimeException(None, f"Invalid value '{value}', expected {self.resolved_type.type_string()}")
```
wouterdb, line 467: I don't like this bit, suggestions welcome

bartv (approved):

src/inmanta/plugins.py
```diff
@@ -201,16 +216,99 @@ def type_string(self) -> str:
...
+    expected_type.validate(value)
+
+    if isinstance(value, NoneValue):
+        # if the type is not nullable, it will fail when validating
+        # if the value is None, it becomes None
+        return None
+
+    base_type = expected_type.get_base_type()
+    if base_type.has_custom_to_python():
+        return expected_type.to_python(value)
```
bartv, line 259: This is probably too strict and will break existing code that is not typed correctly

sanderr (approved):
To my regret I can't spare the time for a proper review. Added some comments that may or may not be relevant anymore.

src/inmanta/plugins.py
wouterdb: It was a stupid mistake elsewhere I think

src/inmanta/plugins.py
```diff
@@ -627,15 +769,23 @@ def check_args(self, args: Sequence[object], kwargs: Mapping[str, object]) -> bo
...
         for position, value in enumerate(args):
             # (1) Get the corresponding argument, fails if we don't have one
             arg = self.get_arg(position)
+            result: object
+            if isinstance(value, Unknown):
+                result = value
+                is_unknown = True
+            else:
+                # (4) Validate the input value
+                result = validate_and_convert_to_python_domain(arg.resolved_type, value)
```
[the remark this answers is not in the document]
wouterdb: I moved it the other way around

src/inmanta/plugins.py
wouterdb: https://github.com/inmanta/inmanta-core/pull/8399/commits/e26b82d7780e73edc6add0e33d316b7bd5738da6

src/inmanta/ast/entity.py
```diff
@@ -508,9 +525,204 @@ def get_double_defined_exception(self, other: "Namespaced") -> "DuplicateExcepti
...
+            index_locations = ",".join([str(i.location) for i in self._indexes])
+            raise DataClassException(
+                self,
+                f"The dataclass {self.get_full_name()} defined at {self.location} has an indexes "
+                f"defined at {index_locations}. Dataclasses can not have any indexes.",
+            )
+
+        self._paired_dataclass = dataclass
+
+    def get_paired_dataclass(self) -> Optional[typing.Type[object]]:
```
bartv, line 652: you should use `type`

sanderr (changes requested):
Some comments. Still a lot I didn't get to. A lot of things popped up today.

src/inmanta/types.py
```diff
@@ -33,6 +33,15 @@
     from inmanta.protocol.common import ReturnValue  # noqa: F401
 
 
+# Typing of dataclass.* methods relies entirely on the definition in typeshed that only exists during typechecking.
+# This ensures that our code works during typechecking and at runtime.
+if not TYPE_CHECKING:
+    DataclassProtocol = object
+else:
+    import _typeshed
```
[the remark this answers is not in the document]
wouterdb: it is not implied by mypy? (i.e. it always seems to have it available)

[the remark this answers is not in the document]
sanderr:
```
sander@bedevere:~/.virtualenvs$ mktmpenv -p 312
created virtual environment CPython3.12.8.final.0-64 in 79ms
  creator CPython3Posix(dest=/home/sander/.virtualenvs/tmp-639a0b6b7a3f85, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, via=copy, app_data_dir=/home/sander/.local/share/virtualenv)
    added seed packages: pip==24.3.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
virtualenvwrapper.user_scripts creating /home/sander/.virtualenvs/tmp-639a0b6b7a3f85/bin/predeactivate
virtualenvwrapper.user_scripts creating /home/sander/.virtualenvs/tmp-639a0b6b7a3f85/bin/postdeactivate
virtualenvwrapper.user_scripts creating /home/sander/.virtualenvs/tmp-639a0b6b7a3f85/bin/preactivate
virtualenvwrapper.user_scripts creating /home/sander/.virtualenvs/tmp-639a0b6b7a3f85/bin/postactivate
virtualenvwrapper.user_scripts creating /home/sander/.virtualenvs/tmp-639a0b6b7a3f85/bin/get_env_details
This is a temporary environment. It will be deleted when you run 'deactivate'.
(tmp-639a0b6b7a3f85) sander@bedevere:~/.virtualenvs/tmp-639a0b6b7a3f85$ pip install --quiet mypy
py
[notice] A new release of pip is available: 24.3.1 -> 25.0
[notice] To update, run: pip install --upgrade pip
(tmp-639a0b6b7a3f85) sander@bedevere:~/.virtualenvs/tmp-639a0b6b7a3f85$ python -c 'import _typeshed'
ModuleNotFoundError: No module named '_typeshed'
```

src/inmanta/const.py
```diff
@@ -391,6 +391,11 @@ class NotificationSeverity(str, Enum):
 RESOURCE_ATTRIBUTE_RECEIVE_EVENTS = "receive_events"
 
 
+# Field in an instance to link it to the associated dataclass
+# Because it starts with a space, it can never collide with a valid attribute defined in the model
+DATACLASS_SELF_FIELD = " dataclass"
```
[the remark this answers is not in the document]
wouterdb: I didn't really consider that one. 

I saves a bit of memory (potentially on a LOT of objects).

I have no hard preference, but updating the potentially millions of instances to be a bit larger may have an impact.

src/inmanta/types.py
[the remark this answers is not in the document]
sanderr: It looks like it vendors it (`mypy.typeshed`)

src/inmanta/execute/proxy.py
```diff
@@ -75,10 +95,18 @@ def recurse_dict_item(key_value: tuple[object, object]) -> tuple[object, object]
                     raise RuntimeException(
                         None, f"dict keys should be strings, got {key} of type {type(key)} with dict value {value}"
                     )
-                return (key, cls.unwrap(value))
+                return (key, cls.unwrap(value, dynamic_context=dynamic_context))
 
             return dict(map(recurse_dict_item, item.items()))
 
+        if is_dataclass(item) and not isinstance(item, type) and dynamic_context is not None:
```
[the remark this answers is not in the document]
wouterdb: It would restrict the meaning of a declared `any`. This runs prior to type checking. Such errors are expected to happen when type checking.

src/inmanta/execute/proxy.py
```diff
@@ -37,7 +39,23 @@
...
+@dataclasses.dataclass(frozen=True, slots=True)
+class DynamicUnwrapContext:
+    """A set of context information that allows dynamic proxy unwrapping to construct instances"""
+
+    resolver: "Resolver"
+    queue: "QueueScheduler"
+    location: Location
+
+
+# this is here to avoid import loops
```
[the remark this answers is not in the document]
wouterdb: It would be cleaner as a class method on Entity, but then importing this would pull in the entire compiler

src/inmanta/execute/proxy.py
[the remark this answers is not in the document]
wouterdb: updated

src/inmanta/const.py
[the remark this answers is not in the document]
wouterdb: updated

src/inmanta/ast/type.py
```diff
@@ -199,6 +275,10 @@ def __eq__(self, other: object) -> bool:
             return NotImplemented
         return True
 
+    def has_custom_to_python(self) -> bool:
+        # All primitives can be trivially converted
+        return True
```
sanderr, line 280: This sounds risky. Wouldn't this cause e.g. primitive lists to be converted instead of being dynamic proxies? I find it tough to judge the impact, but I think this could have a real effect on (possibly rare) models, because object identity might change.

src/inmanta/ast/type.py
```diff
@@ -161,6 +200,43 @@ def __eq__(self, other: object) -> bool:
             return NotImplemented
         return self.element_type == other.element_type
 
+    def is_primitive(self) -> bool:
+        return self.element_type.is_primitive()
```
sanderr, line 204: This does not correspond with the docstring. As far as I can tell, this was an orphaned method before this PR, so I think we have some freedom, but I think we need to define the semantics of "primitive" properly in the docstring.

src/inmanta/ast/type.py
```diff
@@ -161,6 +200,43 @@ def __eq__(self, other: object) -> bool:
...
+        return self.element_type.has_custom_to_python()
+
+
+class Any(Type):
+
+    def corresponds_to(self, pytype: type[object]) -> bool:
+        return True
+
+    def as_python_type_string(self) -> "str | None":
+        return "object"
```
sanderr, line 232: This isn't right, is it?

src/inmanta/ast/type.py
```diff
@@ -161,6 +200,43 @@ def __eq__(self, other: object) -> bool:
...
+    def to_python(self, instance: object) -> "object":
+        if isinstance(instance, NoneValue):
+            return None
+        return self.element_type.to_python(instance)
+
+    def has_custom_to_python(self) -> bool:
+        return self.element_type.has_custom_to_python()
+
+
+class Any(Type):
```
sanderr, line 226: If we keep this, I'd like a docstring that clearly states that while this type represents the any type, similar to Python's, the `Any` class itself is neither the top nor the bottom type in the type hierarchy.

Turns out it's the same for `typing.Any`, but still I think it's good to point out.

src/inmanta/ast/type.py
```diff
@@ -94,6 +99,40 @@ def with_base_type(self, base_type: "Type") -> "Type":
         """
         return base_type
 
+    def corresponds_to(self, pytype: type[object]) -> bool:
+        """
+        Return if the python type corresponds to this inmanta type
+
+        Used only on the plugin boundary
```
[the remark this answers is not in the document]
sanderr: And if we keep it, I think we should rename it (and change the docstring) to `is_compatible_with` or `covers` or something. It's not currently clear that we allow one to be broader than the other.

src/inmanta/ast/type.py
```diff
@@ -421,6 +546,23 @@ def type_string_internal(self) -> str:
...
+            return False
+
+        if not typing_inspect.is_generic_type(pytype):
+            return issubclass(pytype, Sequence)
+
+        origin = typing_inspect.get_origin(pytype)
+        return issubclass(origin, Sequence)
+
+    def as_python_type_string(self) -> "str | None":
+        return "list[object]"
```
sanderr, line 560:
```suggestion
        return "Sequence[object]"
```
Right?

sanderr: Same question for dict

src/inmanta/ast/type.py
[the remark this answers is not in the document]
sanderr: For now, I didn't check all `corresponds_to` implementations in detail.

src/inmanta/ast/type.py
```diff
@@ -474,6 +616,35 @@ def __eq__(self, other: object) -> bool:
...
+        args = typing_inspect.get_args(pytype, True)
+        if len(args) != 1:
+            return False
+        return self.element_type.corresponds_to(args[0])
+
+    def as_python_type_string(self) -> "str | None":
+        return f"list[{self.element_type.as_python_type_string()}]"
+
+    def to_python(self, instance: object) -> "object":
+        assert isinstance(instance, Sequence), f"This method can only be called on iterables, not on {type(instance)}"
```
sanderr, line 642: Shouldn't these be `ValueError` instead of `assert`?

src/inmanta/ast/type.py
```diff
@@ -704,6 +968,23 @@ def get_namespace(self) -> "Namespace":
     def get_double_defined_exception(self, other: "NamedType") -> DuplicateException:
         return DuplicateException(self, other, "TypeConstraint %s is already defined" % (self.get_full_name()))
 
+    def get_base_type(self) -> "Type":
+        assert self.basetype is not None
+        return self.basetype
```
sanderr, line 973: Also doesn't seem to completely match the docstring (though it's somewhat ambiguous here). I notice it's used in the lsm module `__init__.py` and `typedef.py`. Have you checked that those are compatible?

sanderr (changes requested):
Releasing first batch of comments

src/inmanta/ast/type.py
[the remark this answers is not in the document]
wouterdb: It is specifically there to allow some freedom when the type one either side is can't be expressed exactly on the other side, so we can't check exactly. 

e.g. ConstraintType corresponds to the base type in the python domain


(Union is still empty, I wonder how I should handle that)

If you want to remove this method, how exactly would you do that?

src/inmanta/ast/type.py
wouterdb: works fine without

src/inmanta/ast/type.py
wouterdb: is it not?  I think object is the most precise representation of our 'Any'

OTOH, it can't be written in the model, so this code is unreachable. I can raise an exception here as well?

what do you prefer?

src/inmanta/ast/type.py
wouterdb: ostensibly yes, but then we also need the import or use is fully qualified.

it is a story of readability/clarity for less sophisticated users Vs precision for sophisticated users.

How would you solve the import problem? always print that in the example?

src/inmanta/ast/type.py
wouterdb: Good question

I think either we consider this a pre-condition and the assertion could be considered appropriate.

I don't think value error is correct: 
```
exception ValueError

    Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).
```

But type error is probably most specific. I'll change it to that

src/inmanta/ast/type.py
wouterdb: it is indeed wrong.

src/inmanta/ast/type.py
wouterdb: updated

src/inmanta/ast/type.py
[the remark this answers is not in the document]
sanderr: I had not considered that. This is more difficult than I realized. Still, if we can (though it's a bigger "if" than I assumed), I think it would be good to centralize the domain conversion, and to decouple it from the compatibility check.

Assuming we validate attribute types when we cross the Python->DSL boundary (haven't reviewed that part yet), it would be safe to allow wider types on the Python side. Then how about we do a two-stage approach?

1. Convert the Python type annotation to the DSL type through `to_dsl_type()`, ensuring we get the exact same mapping as for plugins. This would then also allow `Annotated` without further effort. We'd have to make sure to construct our DSL types as unambiguously as possible (e.g. flattened unions etc), but I think we already make that effort.
2. Introduce a method similar to the one you have, but between two inmanta types. Something like Python's `issubclass`, though in our case it would have little to do with classes. e.g. `assert Integer().wider_than(my_int_constrained_type)`. False positives (i.e. reject valid pairs) would be acceptable for complex scenarios, though we'd try to avoid them.

I haven't gone over all types in my mind, but I think this should be doable. Wdyt?

src/inmanta/ast/type.py
sanderr: On second thought, I think I agree with you that it matches.

src/inmanta/ast/type.py
sanderr: I now realized the same actually goes for e.g. typedefs / unions / ..., so it was a bit naive of me to think that this might be the case in the first place. Docstring doesn't hurt, but I no longer think it required.

src/inmanta/ast/type.py
sanderr: > it is a story of readability/clarity for less sophisticated users Vs precision for sophisticated users.

In part yes. But there's also the aspect that a user might actually `append` elements to the list, which (I hope) would have no effect in the DSL. This would be an even worse experience for less sophisticated users imo.

> How would you solve the import problem? always print that in the example?

This is a good point, and I don't have a satisfying answer. We could check `if "Sequence" in dataclass_repr` etc, but that's not great either.

src/inmanta/ast/type.py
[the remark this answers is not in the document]
wouterdb: the problem is that it is not 'wider' per se. 
1. Dataclasses are both input and output, so we expect invariance. 
2. but the same types can't be expressed equally well on both sides. 

I think first converting to inmanta domain prior to comparison will be difficult, as we can't/don't represent the difference in representation in both domains. 
If we assume that inmanta language is strictly more expressive, it should work, but the resulting method will be somewhat awkward, as it represent the difference between python and dsl domain in the dsl domain. 

(even if having the annotated types is really nice)

Do you thin Inmanta is strictly more powerfull?

src/inmanta/ast/type.py
wouterdb: Declaring it as Sequence will not stop anyone from appending to it, unless they have mypy configured and they pay attention to it. So while essentially a correct observation, I don't think it is a very strong argument.

src/inmanta/ast/type.py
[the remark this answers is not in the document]
sanderr: 1. While I agree statically, the fact that we have runtime validation on top of it gives us some leeway. I would honestly prefer invariance, as you put it, but as you say I don't think that's feasible, in which case I'd prefer a predictable widening on the side that gets validated anyway, over a semi-arbitrary 1-1 mapping, which sounds like it would be difficult to evolve.

> I think first converting to inmanta domain prior to comparison will be difficult, as we can't/don't represent the difference in representation in both domains. If we assume that inmanta language is strictly more expressive, it should work, but the resulting method will be somewhat awkward, as it represent the difference between python and dsl domain in the dsl domain.

I don't follow. Can you elaborate?

> Do you thin Inmanta is strictly more powerfull?

Definitely not, the other way around if anything. Our typing is better than nothing, but it's not as well structured as Python's.

src/inmanta/ast/type.py
sanderr: Ah yes, I was going off the assumption that mypy (and perhaps even mypy-baseline) would become part of the recommended (and thus conventional) development flow. Would that not be the case? If indeed my assumption is incorrect, then I have to agree that it's a weak argument, and that yours carry more weight.

I do want to make sure that we *allow* `Sequence`, but I think we already do.

wouterdb: We definitely allow it. 

And even if we suggest using it, that is still a long shot from it being used.

src/inmanta/plugins.py
```diff
@@ -341,6 +353,63 @@ def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
...
+        # if the value is None, it becomes None
+        return None
+
+    if expected_type.has_custom_to_python():
+        return expected_type.to_python(value)
+
+    return DynamicProxy.return_value(value)
+
+
+def primitive_python_type_to_model_domain(intype: type) -> inmanta_type.Type:
```
sanderr, line 375: Any reason for this custom implementation rather than `to_dsl_type(intype).type_string()`, raising an error if `type_string()` returns None?

src/inmanta/plugins.py
```diff
@@ -476,6 +545,14 @@ class PluginReturn(PluginValue):
     VALUE_NAME = "return value"
 
 
+@dataclasses.dataclass
+class CheckedArgs:
+
+    args: list[object]
```
sanderr, line 551:
```suggestion
    args: Sequence[object]
```

src/inmanta/plugins.py
```diff
@@ -476,6 +545,14 @@ class PluginReturn(PluginValue):
...
+    kwargs: Mapping[str, object]
+    unknows: bool
```
sanderr, line 553: typo alert

src/inmanta/plugins.py
```diff
@@ -852,24 +946,15 @@ def __call__(self, *args: object, **kwargs: object) -> object:
...
-        def new_arg(arg: object) -> object:
-            if isinstance(arg, Context):
-                return arg
-            elif isinstance(arg, Unknown) and self.is_accept_unknowns():
-                return arg
-            else:
-                return DynamicProxy.return_value(arg)
-
-        new_args = [new_arg(arg) for arg in args]
-        new_kwargs = {k: new_arg(v) for k, v in kwargs.items()}
```
sanderr, line 864: Why is it safe to remove this?

src/inmanta/ast/entity.py
```diff
@@ -508,9 +525,199 @@ def get_double_defined_exception(self, other: "Namespaced") -> "DuplicateExcepti
...
+        dc_fields = {f.name: f for f in dataclasses.fields(dataclass)}
+        dc_types = typing.get_type_hints(dataclass)
+        failures = []
+
+        for rel_or_attr_name in self.get_all_attribute_names():
+            rel_or_attr = self.get_attribute(rel_or_attr_name)
+            match rel_or_attr:
+                case inmanta.ast.attribute.RelationAttribute() as rel:
+                    # No relations except for requires and provides
+                    if not rel.entity.get_full_name() == "std::Entity":
```
sanderr, line 584: Isn't this too accepting? e.g. `MyCollector.all [0:] -- std::Entity` would be accepted here, wouldn't it?

sanderr (commented):
Reviewed everything now

src/inmanta/ast/type.py
```diff
@@ -170,35 +263,45 @@ class Primitive(Type):
 
     def __init__(self) -> None:
         Type.__init__(self)
-        self.try_cast_functions: Sequence[Callable[[Optional[object]], object]] = []
+        self.cast_function: Callable[[Optional[object]], object] = cast_not_implemented
```
wouterdb, line 266: I simplified this a bit

src/inmanta/ast/type.py
```diff
@@ -280,16 +386,22 @@ def type_string(self) -> str:
...
+    def as_python_type_string(self) -> "str | None":
+        return "float"
+
+    def to_python(self, instance: object) -> "object":
+        return instance
+
 
 @stable_api
-class Integer(Number):
+class Integer(Primitive):
```
wouterdb, line 397: This was irregular (Integrer is a number, float is not)

I updated it, but perhaps I shouldn't?

@sanderr wdyt?

src/inmanta/ast/type.py
```diff
@@ -329,6 +450,7 @@ def validate(self, value: Optional[object]) -> bool:
         raise RuntimeException(None, f"Invalid value '{value}', expected {self.type_string()}")
 
     def cast(self, value: Optional[object]) -> object:
+        # TODO: this is a bit odd, in that is accepts None?
```
wouterdb, line 453: @sanderr what do you think of this?

src/inmanta/ast/type.py
```diff
@@ -460,8 +601,8 @@ def type_string(self) -> Optional[str]:
     def type_string_internal(self) -> str:
         return self._wrap_type_string(self.element_type.type_string_internal())
 
-    def get_location(self) -> None:
-        return None
+    def get_location(self) -> Location | None:
+        return self.element_type.get_location()
```
wouterdb, line 605: changed this as well, but I think it is dead code now.

src/inmanta/ast/entity.py
sanderr: Ah, I (wrongfully) assumed that `relation.entity` was the other side of the relation, but it's the one the relation is defined on. So I agree, it could only happen in the scenario you describe. But we should probably still cover it, even if I don't expect anyone would ever really do something like that.

src/inmanta/plugins.py
wouterdb: that is indeed much better.

src/inmanta/plugins.py
wouterdb: thx

src/inmanta/ast/type.py
sanderr: I think I was vaguely aware of this, but never dared touch it. I think it will be fine, and I think we have sufficient test coverage now that we'll notice if it isn't.

src/inmanta/ast/type.py
sanderr: I mean, it can really get any value, we don't do any validation before we call this. And I think that's good for a cast function. Whether the explicit `Optional` annotation is useful here is a different matter. I often use it, even around `object` to draw explicit attention to the fact that a value might be `None`, but in this case I'm not sure that there's value to that. I'd be fine with making it just `object`.

src/inmanta/plugins.py
wouterdb: the jinja integration also uses it

https://github.com/inmanta/std/blob/master/plugins/__init__.py#L175

I'll put it back

src/inmanta/ast/type.py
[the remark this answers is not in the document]
wouterdb: I think it is much better now

src/inmanta/plugins.py
wouterdb: I clarified that one: it get the pre-processed args from the higher level

src/inmanta/ast/entity.py
wouterdb: I updated it, but it will break of you define ANY relation on std::entity, so a bit of a tough one

src/inmanta/ast/type.py
wouterdb: Sorry, I mean the implementation  one line below

src/inmanta/ast/type.py
```diff
@@ -94,6 +106,46 @@ def with_base_type(self, base_type: "Type") -> "Type":
         """
         return base_type
 
+    def corresponds_to(self, type: "Type") -> bool:
+        """
+        Determine if the given 'type' is a good approximation to `self` given that `type` is derived from a python type
+
+        Intended specifically for type correspondence for dataclasses.
+        This brings specific assumptions
```
sanderr, line 114: I like that you made the coupling explicit here.

src/inmanta/ast/type.py
```diff
@@ -94,6 +106,46 @@ def with_base_type(self, base_type: "Type") -> "Type":
...
+        - `self` is the native inmanta type
+        - `type` is translated form the python domain
+        - 'Any' means: trust me, it will be fine (i.e. always corresponds)
```
sanderr, line 117: May become clear from the rest of the review, but does this mean that we also accept python's `object` without question? Iirc we translate it to `Any` internally.

src/inmanta/ast/type.py
```diff
@@ -77,7 +89,7 @@ def normalize(self) -> None:
 
     def is_primitive(self) -> bool:
         """
-        Returns true iff this type is a primitive type, i.e. number, string, bool.
+        Returns true iff this type is a primitive type, i.e. number, string, bool or a collection of only primitives.
```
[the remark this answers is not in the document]
sanderr: Also, given that we currently only use it for dataclasses, and that we may want to extend support for that in the future, I wonder if we might just have to bite the bullet and make the method dataclass-tailored (and -named)?

src/inmanta/ast/type.py
sanderr: We used to have more than one here, I hadn't realized that that was no longer the case. Nice! I like how it makes the `cast` method implementation so much simpler. As to the not implemented path, I wonder if we can simply go for an abstract property instead. Iirc you can override that with a simple attribute in the child classes, but it's been a while since I've tried.

src/inmanta/ast/type.py
```diff
@@ -170,35 +267,45 @@ class Primitive(Type):
...
+            return True
+        return type == self
+
     def __eq__(self, other: object) -> bool:
         if other.__class__ != self.__class__:
             return NotImplemented
         return True
 
+    def has_custom_to_python(self) -> bool:
+        # All primitives can be trivially converted
```
sanderr, line 299: This comment doesn't apply anymore

src/inmanta/ast/type.py
```diff
@@ -161,6 +213,51 @@ def __eq__(self, other: object) -> bool:
...
+    """
+
+    def corresponds_to(self, type: Type) -> bool:
+        return True
+
+    def as_python_type_string(self) -> "str | None":
+        return "object"
+
+    def has_custom_to_python(self) -> bool:
+        return False
```
sanderr, line 252: Do we need all these explicit `return False` implementations (here, `Primitive`, ...)? The `Type` implementation should cover that.

src/inmanta/ast/type.py
```diff
@@ -244,6 +348,15 @@ def type_string(self) -> str:
     def type_string_internal(self) -> str:
         return self.type_string()
 
+    def as_python_type_string(self) -> "str | None":
+        return "numbers.Number"
+
+    def to_python(self, instance: object) -> "object":
+        return instance
```
sanderr, line 355: Never called anymore since this doesn't override `has_custom_to_python`?

src/inmanta/ast/type.py
```diff
@@ -418,9 +550,22 @@ def type_string(self) -> str:
...
 
-    def get_location(self) -> None:
+    def get_location(self) -> None | Location:
         return None
 
+    def corresponds_to(self, type: Type) -> bool:
+        if isinstance(type, Any):
+            return True
+
+        return isinstance(type, List)
```
sanderr, line 560: Do we want to allow children here (i.e. `TypedList`)? I would expect `return builtins.type(type) is self.__class__ and self.base_type() is self or self.base_type().corresponds_to(type.base_type())` or something (everything from the first `and` on is to cover the child implementation, might not be appropriate here)

src/inmanta/ast/type.py
```diff
@@ -474,6 +619,35 @@ def __eq__(self, other: object) -> bool:
...
+        if isinstance(type, Any):
+            return True
+
+        if not isinstance(type, List):
+            # Not a list at all
+            return False
+
+        if not isinstance(type, TypedList):
+            # The other list is untyped, so we are equivalent
+            return True
```
sanderr, line 635: Is this correct? When would this happen? Unparametrized `list` in Python would become `List(Any())` in the inmanta domain afaik.

Same question for dict.

src/inmanta/ast/type.py
```diff
@@ -612,11 +821,34 @@ def validate(self, value: object) -> bool:
...
+    def corresponds_to(self, type: Type) -> bool:
+        raise NotImplementedError("No unions can be specified as attributes, unexpected usage")
+
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, Union):
             return NotImplemented
         return self.types == other.types
 
+    def is_primitive(self) -> bool:
+        return all(tp.is_primitive() for tp in self.types)
```
sanderr, line 846: This makes types like `int | list[int | str] | dict[str, float | str]` (python syntax) primitive as well. Is that intentional?

sanderr (commented):
I wanted to do a final full sweep but I notice I can't keep focused at the moment. Still need to process explainer, proxy and plugins.

docs/model_developers/plugins.rst
```diff
@@ -112,6 +112,65 @@ see :ref:`moddev-module`.
...
+        int ram
+        int cpus
+    end
+
+    implement Virtualmachine using std::none
+
+    vm = make_virtual_machine()
+    std::print(vm.name)
+
+When using dataclasses, the object can be passed around freely into and out of plugins.
```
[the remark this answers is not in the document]
wouterdb: it that relevant? I.e. it has no observable associated behavior?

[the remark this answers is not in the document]
sanderr: Right, no indexes, no relations and no implementations => no observable behavior.

src/inmanta/ast/type.py
wouterdb: I wonder if type shouldn't be an ABC (well, that ship has sailed).

Its implementation of has_custom_to_python makes no sense in general.

wouterdb: I cleaned it up some more

src/inmanta/ast/type.py
wouterdb: The model can't produce this type, so it is a moot point. I could just assert False here I think. 

Would it be better to make it raise an exception when called?

src/inmanta/ast/type.py
wouterdb: I agree, this is dead code. Type could/should be an ABC.

Same as above: do I raise an exception here?

src/inmanta/ast/type.py
wouterdb: yes, under this definition of primitive.

src/inmanta/ast/type.py
sanderr: I have no strong objection. It's just very wide on the Python side. But there is always still the runtime validation (right?) so it's not too bad.

src/inmanta/ast/type.py
sanderr: Right, when I typed it I was thinking it was the DSL's `list` type, but that's `TypedList(Literal())`. I guess it's fine like this then. Perhaps add a comment that it's currently not expected to be called?

src/inmanta/ast/type.py
wouterdb: on the model side: yes

src/inmanta/ast/type.py
sanderr: > that ship has sailed

Many ships, when it comes to this module.

src/inmanta/ast/type.py
sanderr: But we don't allow them for dataclasses, purely because they can't be expressed in the DSL?

src/inmanta/ast/type.py
sanderr: I'd honestly just return `False` if it's not a typed list, no?

src/inmanta/ast/entity.py
```diff
@@ -508,9 +528,210 @@ def get_double_defined_exception(self, other: "Namespaced") -> "DuplicateExcepti
...
+         i.e. instances of the associated dataclass
+        """
+        assert self._paired_dataclass is not None  # make mypy happy
+        assert isinstance(value, self._paired_dataclass)
+
+        def convert_none(x: object | None) -> object:
+            return x if x is not None else NoneValue()
+
+        instance = self.get_instance(
+            {k.name: convert_none(getattr(value, k.name)) for k in dataclasses.fields(value)},
```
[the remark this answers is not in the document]
sanderr: Fair enough

src/inmanta/ast/type.py
wouterdb: it does! It just means the opposite!

src/inmanta/ast/type.py
sanderr: Hmm, I agree it's strange. But I don't think we can change this on a minor. Currently `bool(null)` is simply accepted and becomes `false`. And perhaps there is something to say about it.

src/inmanta/ast/entity.py
sanderr: > it will break of you define ANY relation on std::entity

Isn't that what we want? At least for now.

src/inmanta/types.py
sanderr, line 41: Drawing attention to [this comment](https://github.com/inmanta/inmanta-core/pull/8399/files#r1928841910) once more (it gets lost in GitHub's thread collapsing).

sanderr (changes requested):
I resolved all threads that are no longer relevant. Only a few remaining, some of which are simply open discussions or questions from my side.

src/inmanta/ast/entity.py
wouterdb: I think so

src/inmanta/ast/type.py
wouterdb: I would like something abc-like, but that ship,...

src/inmanta/types.py
wouterdb: What is the actionable point?

src/inmanta/ast/type.py
sanderr: Well, this could still become one, but I see your point. Ok, we leave this as is then.

src/inmanta/ast/type.py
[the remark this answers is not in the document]
wouterdb: Could you make a suggestion?

src/inmanta/types.py
sanderr: Make sure we either include typeshed as a dev dependency, or that we use the typeshed that's vendored in mypy.

src/inmanta/ast/type.py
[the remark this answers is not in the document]
sanderr: If you agree that the only use case is whether it's suited for dataclasses atm, something like `is_dataclass_compatible` or `suited_for_dataclass` and then mention in the docstring that we mean inmanta dataclasses? I think I prefer the first.

Now that I think about it, we could have made our lives easier (but it's probably too late for such a change at this point) by combining it with the other dataclass specific methods into one, that simply returns an `Optional[DataclassAttributeType]`, which has fields for the Python type string, a method to validate, ... Then we wouldn't have had any of the issues we had now with having to implement methods in a sane manner, while we know they will never be called, except that perhaps in the future we forget to double check them if/when we include those types as dataclass attributes.

Anyway, that pondering aside, I'd already be happy with a name change such as in the first paragraph.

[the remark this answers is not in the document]
sanderr: Do you mean in `Literal.corresponds_to()`? I think we could safely make that an `isinstance(..., Primitive)` check, because we don't actually support recursion on the attributes. But I'm not 100% confident.

Either way, I think calling `is_dataclass_compatible()` from `Literal.corresponds_to()` is definitely the lesser of two evils compared to having a method whose name says one thing but whose semantics are completely different, and a lot less stable towards the future than the name implies.

src/inmanta/ast/type.py
```diff
@@ -610,8 +610,8 @@ def corresponds_to(self, type: Type) -> bool:
             return False
 
         if not isinstance(type, TypedList):
-            # The other list is untyped, so we are equivalent
-            return True
+            # The other list is untyped, so we are not equivalent
+            return False
```
sanderr, line 614: This makes the first `isinstance` check redundant.

sanderr (changes requested):

sanderr (changes requested):
@wouterdb I really think something needs to be done about the `is_primitive` still. Even if it's just a note in its docstring stating that it is considered a part of the dataclass functionality, and subject to changes when the dataclass requirements change. Just anything that will make sure we don't start counting on its behavior, and that we know in the future which liberties we have with it.

src/inmanta/ast/type.py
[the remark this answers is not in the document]
wouterdb: I made it into `is_attribute_type` to indicate the type can be used as a type for an attribute, which is what we use it for

sanderr (approved):

PR #8419, opened 2024-11-27
clarify atomicity of deploy start

### LINKED ISSUE #8325: Deploy.execute does not report resource state when deploy result is None
It has an `if deploy_result is not None` condition. ~~However, this violates the contract of the `TaskManager`. The method must always be called. If need be the method's contract must be updated so that the notification becomes a NOOP, but I actually think it would be more appropriate to simply report a failure.~~ EDIT: the mandatory part is already outside of the if body. The relevant part is described below.

A more delicate instance of this problem exists [here](https://github.com/inmanta/inmanta-core/blame/652e14fc50e0f507cd51d15bfbf933e4beabfed4/src/inmanta/deploy/tasks.py#L148). The deploy start notification resulted in an error. Do we need to report that it finished? One observed case had the server register the resource as "deploying", with the result that it was stuck in this state.

wouterdb:
# Description

Ensure we end the deploy if we start it, all part where in place, this is mostly docs 

closes #8325 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/persistence.py
+++ b/src/inmanta/deploy/persistence.py
@@ -44,6 +44,12 @@ class StateUpdateManager(abc.ABC):
 
     @abc.abstractmethod
     async def send_in_progress(self, action_id: UUID, resource_id: ResourceVersionIdStr) -> None:
+        """
+        This method sets the state to in_progress.
+
+        It is important that this method is atomic: if it fails, we assume the state to be not set and will not re-set it
+
+        """
         # FIXME: get rid of version in the id
         pass
 
--- a/src/inmanta/deploy/tasks.py
+++ b/src/inmanta/deploy/tasks.py
@@ -120,17 +120,7 @@ class Deploy(Task):
             if intent is None:
                 # Stale resource, can simply be dropped.
                 return
-
-            # Dependencies are always set when calling get_resource_intent_for_deploy
-            assert intent.dependencies is not None
-            # Resolve to exector form
-            version = intent.model_version
-            resource_details = intent.details
-            executor_resource_details: executor.ResourceDetails = self.get_executor_resource_details(version, resource_details)
-
-            # Make id's
-            gid = uuid.uuid4()
-            action_id = uuid.uuid4()  # can this be gid ?
+            # From this point on, we HAVE to call report_resource_state to make the scheduler propagate state
 
             # The main difficulty off this code is exception handling
             # We collect state here to report back in the finally block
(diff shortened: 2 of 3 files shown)


src/inmanta/deploy/tasks.py
```diff
@@ -120,41 +120,47 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
...
+            # From this point on, we HAVE to call report_resource_state to make the scheduler propagate state
 
             # The main difficulty off this code is exception handling
             # We collect state here to report back in the finally block
 
             # Full status of the deploy,
             # may be unset if we fail before signaling start to the server, will be set if we signaled start
             deploy_result: DeployResult | None = None
-            scheduler_deployment_result: state.DeploymentResult
+            scheduler_deployment_result: state.DeploymentResult = state.DeploymentResult.FAILED
```
jptrindade, line 131: I remember @sanderr not wanting to initialize this value and to explicitly set it if the deploy were to fail. Like in line 154

wouterdb: The question is: where do you catch that failure?

We caught all the ones we expect, but if something unexpected happens, do we prefer to crash the scheduler or report failure and continue

sanderr: I think João may be referring to my preference for style a over style b in the snippet below
```python
# style a
x: int
try:
  do_stuff()
  x = 0
except Exception:
  x = 1

# style b
y: int = 1
try:
  do_stuff()
  y = 0
except Exception:
  pass
```

This is mostly a stylistic preference (semi-functional over updating the value), with a small part of confidence that we can't forget to set it in any of the branches.

But it remains mostly stylistic, and thus subjective, so I won't raise a fuss.

sanderr (changes requested):

src/inmanta/deploy/persistence.py
```diff
@@ -44,6 +44,12 @@ class StateUpdateManager(abc.ABC):
 
     @abc.abstractmethod
     async def send_in_progress(self, action_id: UUID, resource_id: ResourceVersionIdStr) -> None:
+        """
+        This method sets the state to in_progress.
+
+        It is important that this method is atomic: if it fails, we assume the state to be not set and will not re-set it
```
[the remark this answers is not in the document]
wouterdb: (the to_server version definitely was not atomic, this one is)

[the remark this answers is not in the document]
sanderr: Right, it's not to_server anymore

sanderr (approved):

Issue #8321, opened 2024-11-06
Add api end-point for scheduler self-check


wouterdb:
1. Add an api endpoint to the server
2. wire it through to the scheduler
3. make it verify the scheduler state against the DB state and report on any discrepancies
4. make it return the full internal state of the scheduler

This is intended to assist in diagnosing scheduler problems

wouterdb:
The DummyStateManager class already contains part of the checks

PR #8446, opened 2024-12-05
Issue/8321 add api endpoint for scheduler self check

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/8321

# Self Check:

- [x] Changelog entry


src/inmanta/agent/agent_new.py
```diff
@@ -28,17 +28,17 @@
...
+from inmanta.protocol import SessionEndpoint, TypedClient, methods, methods_v2
 from inmanta.server.services.databaseservice import DatabaseMonitor
 from inmanta.types import Apireturn
 from inmanta.util import ensure_directory_exist, join_threadpools
 
 LOGGER = logging.getLogger("inmanta.scheduler")
 
 
-class Agent(SessionEndpoint):
+class Agent(SessionEndpoint, TypedClient):
```
wouterdb, line 41: 1. can you do this? 
2. What does this mean?

src/inmanta/data/model.py
```diff
@@ -898,6 +898,19 @@ def has_source(self) -> bool:
...
+    """
+    Status report for the scheduler self-check
+
+    :param resource_state: In-memory representation of the resources in the scheduler
+    :param discrepancies: Discrepancies between the in-memory representation of the resources
+        and their state in the database.
+    """
+
+    resource_state: dict[ResourceIdStr, typing.Any]
+    discrepancies: dict[ResourceIdStr, typing.Any]
```
wouterdb, line 911: It would be nicer if the second arguments are typed, not crucial

src/inmanta/deploy/scheduler.py
```diff
@@ -372,6 +382,107 @@ async def get_facts(self, resource: dict[str, object]) -> None:
...
+        """
+        Check that the state of the resources in the DB corresponds
+        to the internal state of the scheduler and return a SchedulerStatusReport
+        object containing the internal state and any discrepancies between this
+        state and that of the DB.
+        """
+
+        resources_in_db: Mapping[ResourceIdStr, ResourceDetails]
+
+        try:
```
wouterdb, line 395: We may want to grab some lock to make sure we read one version of the data?

src/inmanta/deploy/scheduler.py
```diff
@@ -372,6 +382,107 @@ async def get_facts(self, resource: dict[str, object]) -> None:
...
+            # Resources only present in the DB but missing from the scheduler
+            only_in_db = resources_in_db.keys() - self._state.resource_state.keys()
+            for rid in only_in_db:
+                discrepancy_map[rid] = [
+                    Discrepancy(
+                        expected=(
+                            f"Resource {rid} is present in the DB (model version {latest_version}). "
+                            "It is expected in the scheduler's resource_state map."
+                        ),
+                        actual=f"Resource {rid} is missing from the scheduler's resource_state map.",
```
wouterdb, line 437: I wonder if we can split the context into different fields in the `Discrepancy` i.e. such that is has a field for 'rid', an optional for 'field' (which field is it about) and then expected actual

src/inmanta/deploy/scheduler.py
```diff
@@ -372,6 +382,107 @@ async def get_facts(self, resource: dict[str, object]) -> None:
...
+                            Discrepancy(expected=db_compliance_status, actual=scheduler_resource_state.status)
+                        )
+
+                if resource_discrepancies:
+                    discrepancy_map[rid] = resource_discrepancies
+
+            return discrepancy_map
+
+        discrepancy_map = _build_discrepancy_map(resources_in_db)
+        return SchedulerStatusReport(resource_state=self._state.resource_state, discrepancies=discrepancy_map)
```
wouterdb, line 484: We may want to pack the DB state in here as well.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1305,6 +1306,20 @@ async def get_diff_of_versions(
...
+        env: data.Environment,
+    ) -> SchedulerStatusReport:
+        await self.autostarted_agent_manager._ensure_scheduler(env.id)
+        client = self.agentmanager_service.get_agent_client(env.id, const.AGENT_SCHEDULER_ID)
+        assert client is not None
+        status = await client.trigger_get_status(env.id)
+        assert status.code == 200
+        assert status.result is not None
+        resp = SchedulerStatusReport.model_validate(status.result["data"])
+        return resp
```
wouterdb, line 1321: There assert should not be asserts, but raise proper errors with a proper http code

wouterdb (commented):
Looks good overall

src/inmanta/agent/agent_new.py
Hugo-Inmanta: Removed, it was an artifact from experimentation with `TypedClient`

src/inmanta/deploy/scheduler.py
Hugo-Inmanta: How much of it should we pack ? I plopped everything from the `ResourceDetails` in there but this is maybe overkill ?

src/inmanta/data/model.py
Hugo-Inmanta: I don't get how to type these, there's probably something obvious I'm missing. 
e.g. for `resource_state` 
Should we import ResourceState from inmanta.deploy.state and type it as such here ? I tried but I was running into circular imports but I can have another look if that's the way to go.

`    resource_state: dict[ResourceIdStr, ResourceState]`

src/inmanta/deploy/scheduler.py
```diff
@@ -372,6 +372,139 @@ async def get_facts(self, resource: dict[str, object]) -> None:
...
+            except KeyError:
+                return SchedulerStatusReport(scheduler_state={}, db_state={}, discrepancies={})
+
+            if latest_version != self._state.version:
+                LOGGER.info(
+                    "Cannot compare scheduler status and database status because of "
+                    "model version mismatch (%s and %s respectively). "
+                    "The scheduler might not have processed this new version yet.",
+                    self._state.version,
+                    latest_version,
```
wouterdb, line 499: This message should go in the report as well.

src/inmanta/deploy/scheduler.py
```diff
@@ -372,6 +372,139 @@ async def get_facts(self, resource: dict[str, object]) -> None:
...
+        """
+        Check that the state of the resources in the DB corresponds
+        to the internal state of the scheduler and return a SchedulerStatusReport
+        object containing the internal state and any discrepancies between this
+        state and that of the DB.
+        """
+
+        resources_in_db: Mapping[ResourceIdStr, ResourceDetails]
+
+        def _build_discrepancy_map(
```
wouterdb, line 385: This method may take a long time. I wonder if we should yield to the ioloop ('await sleep(0)`) every now and them to keep the session alive ? (e.g. every 1000 loop iterations) or run this on a thread

src/inmanta/data/model.py
```diff
@@ -898,6 +898,43 @@ def has_source(self) -> bool:
...
+    Status report for the scheduler self-check
+
+    :param scheduler_state: In-memory representation of the resources in the scheduler
+    :param db_state: Desired state of the resources as persisted in the database
+    :param discrepancies: Discrepancies between the in-memory representation of the resources
+        and their state in the database.
+    """
+
+    scheduler_state: Mapping[ResourceIdStr, typing.Any]
+    db_state: Mapping[ResourceIdStr, typing.Any]
```
wouterdb, line 934: Would it be difficult to have these typed as well? i.e. if easy, I would, if the existing types don't work, I would document why it is any.

wouterdb (approved):
Cool!

src/inmanta/data/model.py
```diff
@@ -898,6 +898,44 @@ def has_source(self) -> bool:
...
 
+class Discrepancy(BaseModel):
+    """
+    Records a discrepancy between the state as persisted in the database and
+    the in-memory state in the scheduler.
+
+    :param rid: If set, this discrepancy is specific to this resource.
+    :param field: If set, specifies on which field this discrepancy was detected.
+        If left unset, and a rid is specified, the discrepancy was detected on the
+        resource level i.e. it is missing from either the db or the scheduler.
```
jptrindade, line 909: You could also mention the case of `model_version`

src/inmanta/protocol/methods_v2.py
```diff
@@ -392,6 +392,39 @@ def get_api_docs(format: Optional[ApiDocsFormat] = ApiDocsFormat.swagger) -> Ret
...
+    arg_options=methods.ENV_OPTS,
+    client_types=[],
+    api_version=2,
+)
+def trigger_get_status(tid: uuid.UUID) -> model.SchedulerStatusReport:
+    """
+    Get a snapshot of the scheduler state
+
+    :param tid: The id of the environment.
+    """
```
jptrindade, line 425: I don't quite get the difference between these two endpoints

jptrindade (changes requested):
Is it possible (maybe even in another PR) to make this call at the end of every test case and assert that `discrepancies` is empty? Making it only run in the nightly if it is too heavy

src/inmanta/protocol/methods_v2.py
Hugo-Inmanta: `get_scheduler_status` is the outer one: api on the server to request the status, (the orchestration service responds to this one) and it in turns calls the inner `trigger_get_status` to retrieve the status (the new agent responds to that one)

src/inmanta/data/model.py
Hugo-Inmanta: Updated docstring.

src/inmanta/data/model.py
Hugo-Inmanta: As discussed, typed as object + added a comment

jptrindade (approved):

PR #8459, opened 2024-12-09
Fix bugs and inefficiencies in report resource state query

### LINKED ISSUE #8324: bugs and inefficiencies in new query?
```python
diff --git a/src/inmanta/data/__init__.py b/src/inmanta/data/__init__.py
index 7430ea1c..d2578c19 100644
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4742,6 +4742,9 @@ class Resource(BaseDocument):
         :param version: The version of the model impacted by this
         :param connection: The connection to use
         """
+        # TODO: function takes version argument but only works for latest version because it writes latest status
+        # TODO: missing environment= in WHERE clause?!!!!!
+        # TODO: convoluted query? FROM clause really only needs rps? Current implementation is a self join + rps join while only rps join is needed
         query = f"""
             UPDATE resource AS updated_r
             SET status=inconsistent_r.last_non_deploying_status::TEXT::resourcestate
```

jptrindade:
# Description

`report_resource_state` query now fetches  the latest released version of the model directly instead of it being provided by an argument

closes #8324 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/8324-fix-bugs-in-report-resource-state-query.yml
+++ b/changelogs/unreleased/8324-fix-bugs-in-report-resource-state-query.yml
@@ -1,0 +1,4 @@
+description: Fix bugs and inefficiencies in report resource state query
+issue-nr: 8324
+change-type: patch
+destination-branches: [master]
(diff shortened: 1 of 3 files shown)


sanderr (approved):
Nice

PR #8461, opened 2024-12-09
Issue/8290 race

### LINKED ISSUE #8290: potential races in scheduler
Some potential races related to starting and stopping the scheduler were uncovered when investigating a strange behavior (that turned out to be unrelated in the end). These scenarios should be investigated: are they actually realistic, do they present a problem? And to fix them if deemed relevant.

Relevant Slack thread [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1730110283566659).

# agent queue operations

In the implementation at the time of writing, `_run_for_agent` might still pick up one task after the scheduler has been instructed to finish. Suppose `start()` is called shortly after `stop()`, the agent queues might have been reset already, meaning that the in-progress operations run on a stale queue. This mostly affects `queue_get()` and potentially `task_done()`. See the PoC diff below.

Alternatively, the race between stop and start could be made more robust. Then the agent queues' reset could have the invariant that all queue operations must be fully wrapped up before queue operations may be resumed. To fix the race, we'd probably need to have `start()` wait until `stop()` has fully wrapped up, before doing anything.

Additionally, `_in_progress` is never cleared.

**Important:** at the time of writing a heavy rework of the scheduler halting is in progress. It is also adding per-agent halting. Depending on the implementation it may address these issues already, or it may simply make them more broadly relevant.

```diff
diff --git a/src/inmanta/deploy/work.py b/src/inmanta/deploy/work.py
index d6ee7649d..43b050cb5 100644
--- a/src/inmanta/deploy/work.py
+++ b/src/inmanta/deploy/work.py
@@ -133,6 +133,7 @@ class AgentQueues:
     def reset(self) -> None:
         self._agent_queues.clear()
         self._tasks_by_resource.clear()
+        self._in_progress.clear()
         self._entry_count = 0

     def _get_queue(self, agent_name: str) -> asyncio.PriorityQueue[TaskQueueItem]:
@@ -236,9 +237,13 @@ class AgentQueues:
         """
         Consume a task from an agent's queue. If the queue is empty, blocks until a task becomes available.
         """
-        queue: asyncio.PriorityQueue[TaskQueueItem] = self._agent_queues[agent]
         while True:
+            queue: asyncio.PriorityQueue[TaskQueueItem] = self._agent_queues[agent]
             item: TaskQueueItem = await queue.get()
+            if self._agent_queues[agent] is not queue:
+                # reset() was called while waiting for the next item: try again with the fresh queue
+                # TODO: this agent might not exist anymore -> what to return then? virtual poison pill? But that might break task_done?
+                continue
             if item.deleted:
                 # task was marked as removed, ignore it and consume the next item from the queue
                 queue.task_done()
@@ -256,6 +261,7 @@ class AgentQueues:
         Used by queue consumers. For each get() used to fetch a task, a subsequent call to task_done() tells the corresponding
         agent queue that the processing on the task is complete.
         """
+        # TODO: can this be received out of order?
         del self._in_progress[task]
         self._agent_queues[agent].task_done()
```

# better / earlier error for empty code

If for any reason (e.g. due to a race), `types_per_agent` is empty for a given agent, the current implementation just passes that empty list on until we reach a `ValueError` relatively deep in the stack. This is not user friendly because it doesn't inform the user as to what's wrong. We should probably raise the error at the highest level where it's relevant. In practice, I don't think the list is ever supposed to be empty, it likely always indicates a bug in the agent type tracking.

wouterdb:
# Description

Fix shutdown/restart race

closes #8290 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -277,7 +277,6 @@ class ResourceScheduler(TaskManager):
         await self._initialize()
         self._running = True
 
-
     async def stop(self) -> None:
         if not self._running:
             return
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -133,15 +133,21 @@ class TaskRunner:
         self._task: typing.Optional[asyncio.Task[None]] = None
         self._notify_task: typing.Optional[asyncio.Task[None]] = None
 
-    async def _start(self) -> None:
+    async def start(self) -> None:
         self.status = AgentStatus.STARTED
         assert (
             self._task is None or self._task.done()
         ), f"Task Runner {self.endpoint} is trying to start twice, this should not happen"
         self._task = asyncio.create_task(self._run())
 
-    async def _stop(self) -> None:
+    async def stop(self) -> None:
         self.status = AgentStatus.STOPPING
+
+    async def join(self) -> None:
+        assert not self.is_running(), "Joining worker that is not stopped"
+        if self._task is None or self._task.done():
+            return
+        await self._task
 
     async def notify(self) -> None:
         """
@@ -155,9 +161,9 @@ class TaskRunner:
 
         match self.status:
             case AgentStatus.STARTED if not should_be_running:
-                await self._stop()
+                await self.stop()
             case AgentStatus.STOPPED if should_be_running:
-                await self._start()
+                await self.start()
             case AgentStatus.STOPPING if should_be_running:
                 self.status = AgentStatus.STARTED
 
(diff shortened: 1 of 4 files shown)


src/inmanta/deploy/scheduler.py
```diff
@@ -256,19 +257,42 @@ async def reset(self) -> None:
...
 
+    async def stop(self) -> None:
+        if not self._running:
+            return
+        self._running = False
+        await self._timer_manager.stop()
+        # Ensure workers go down
+        # First stop them
+        for worker in self._workers.values():
+            await worker.stop()
```
wouterdb, line 288: this was missing entirely, making the race a leak as well.

arnaudsjs (approved):
I did a high level review. I'll leave the in depth review to Sander. I just don't have the time at the moment.

sanderr (approved):

PR #8465, opened 2024-12-10
added fixes for tranistive blocking

### LINKED ISSUE #8284: update_desired_state for transitively blocked resources
There are a number of small bugs in `update_desired_state` and `drop` where they don't take the possibility of transitively blocked resources into account.

- [x] Fix `update_desired_state`
- [x] Fix `drop`
- [x] When requires are changed (might fix the other two for free)

# update_desired_state

It's easiest to explain with the code, so below is the current implementation, with FIXME comments attached.

```python
    def update_desired_state(
        self,
        resource: ResourceIdStr,
        details: ResourceDetails,
    ) -> None:
        """
        Register a new desired state for a resource.

        For blocked resources, block_resource() should be called instead.
        """
        self.resources[resource] = details
            self.resource_state[resource].status = ResourceStatus.HAS_UPDATE
            # FIXME: it's possible that this resource was previously transitively blocked, in which case it should remain so.
            #       If the cause becomes unblocked, it will recurse along its provides, so it will resolve itself.
            #       => if the ResourceStatus was previously UNDEFINED, we need to recalculate blocked status, but in that case
            #           we don't end up here, instead calling mark_as_defined. So no need to do anything here, unless the two
            #           methods are unified.
            #       => otherwise NO remains safe because newly undefined resources will recurse when they are marked as such.
            self.resource_state[resource].blocked = BlockedStatus.NO
        else:
            self.resource_state[resource] = ResourceState(
                status=ResourceStatus.HAS_UPDATE,
                deployment_result=DeploymentResult.NEW,
                # FIXME: We need to calculate the appropriate blocked status here and recurse along provides iff we set it to blocked.
                #       => +- mark_as_defined minus the marking as defined
                blocked=BlockedStatus.NO,
            )
            self.types_per_agent[details.id.agent_name][details.id.entity_type] += 1
        # FIXME: only if not blocked
        self.dirty.add(resource)
```

# drop

When a resource is dropped from the model, it should have the same effect as `mark_as_defined` with respect to its provides.

wouterdb:
# Description

1. make a single calculation for blocked updates, because requires-provides updates can move nodes both ways. If I could not cleanly separate them, I could benefit from shared cache
2. move all transitive update out of all other code 

- [ ] ticketize: add self check that all blocks and transient blocks are well founded 

closes #8284 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -489,10 +489,6 @@ class ResourceScheduler(TaskManager):
         up_to_date_resources: Optional[Mapping[ResourceIdStr, ResourceDetails]] = None,
         reason: str = "Deploy was triggered because a new version has been released",
     ) -> None:
-        # TODO:
-        #   1. update all immediate state
-        #   2. propagate all transients (first the blocked: yes, then the blocked: No, no node is ever visited twice)
-
         up_to_date_resources = {} if up_to_date_resources is None else up_to_date_resources
         async with self._intent_lock:
             # Inspect new state and compare it with the old one before acquiring the scheduler lock.
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -489,6 +489,10 @@ class ResourceScheduler(TaskManager):
         up_to_date_resources: Optional[Mapping[ResourceIdStr, ResourceDetails]] = None,
         reason: str = "Deploy was triggered because a new version has been released",
     ) -> None:
+        # TODO:
+        #   1. update all immediate state
+        #   2. propagate all transients (first the blocked: yes, then the blocked: No, no node is ever visited twice)
+
         up_to_date_resources = {} if up_to_date_resources is None else up_to_date_resources
         async with self._intent_lock:
             # Inspect new state and compare it with the old one before acquiring the scheduler lock.
@@ -502,7 +506,7 @@ class ResourceScheduler(TaskManager):
             # Only contains the direct undeployable resources, not the transitive ones.
             undefined: set[ResourceIdStr] = set()
             # Resources that were undeployable in a previous model version, but got unblocked. Not the transitive ones.
-            newly_defined: set[ResourceIdStr] = set()  # resources that were previously undefined but not anymore
+            now_defined: set[ResourceIdStr] = set()  # resources that were previously undefined but not anymore
 
             # Track potential changes in requires per resource
             added_requires: dict[ResourceIdStr, Set[ResourceIdStr]] = {}
@@ -519,7 +523,9 @@ class ResourceScheduler(TaskManager):
                     # It's a resource we know.
                     if self._state.resource_state[resource].status is ComplianceStatus.UNDEFINED:
                         # The resource has been undeployable in previous versions, but not anymore.
-                        newly_defined.add(resource)
+                        now_defined.add(resource)
+                        if details.attribute_hash == self._state.resources[resource].attribute_hash:
+                            LOGGER.warning("The resource with id %s has become defined, but the hash has not changed", resource)
                     elif details.attribute_hash != self._state.resources[resource].attribute_hash:
                         # The desired state has changed.
                         new_desired_state.add(resource)
@@ -528,16 +534,20 @@ class ResourceScheduler(TaskManager):
 
                 old_requires: Set[ResourceIdStr] = self._state.requires.get(resource, set())
                 new_requires: Set[ResourceIdStr] = requires.get(resource, set())
-                added_requires[resource] = new_requires - old_requires
-                dropped_requires[resource] = old_requires - new_requires
+                my_added = new_requires - old_requires
+                if my_added:
+                    added_requires[resource] = my_added
+                my_removed = old_requires - new_requires
+                if my_removed:
+                    dropped_requires[resource] = my_removed
 
                 # this loop is race-free, potentially slow, and completely synchronous
                 # => regularly pass control to the event loop to not block scheduler operation during update prep
                 await asyncio.sleep(0)
 
             # A resource should not be present in more than one of these resource sets
-            assert len(new_desired_state | undefined | newly_defined) == (
-                len(new_desired_state) + len(undefined) + len(newly_defined)
+            assert len(new_desired_state | undefined | now_defined) == (
+                len(new_desired_state) + len(undefined) + len(now_defined)
             )
 
             # in the current implementation everything below the lock is synchronous, so it's not technically required. It is
@@ -548,24 +558,29 @@ class ResourceScheduler(TaskManager):
             async with self._scheduler_lock:
                 self._state.version = version
                 for resource in undefined:
-                    self._state.block_resource(resource, resources[resource], is_transitive=False)  # Removes from the dirty set
+                    self._state.block_and_update_resource(resource, resources[resource])  # Removes from the dirty set
+                for resource in now_defined:
+                    # the resources moving out of undefined, normal update
+                    self._state.update_desired_state(resource, resources[resource])  # Updates the dirty set
                 for resource in new_desired_state:
                     self._state.update_desired_state(resource, resources[resource])  # Updates the dirty set
                 for resource in added_requires.keys() | dropped_requires.keys():
                     self._state.update_requires(resource, requires[resource])
-                blocked_for_undefined_dep: Set[ResourceIdStr] = self._state.block_provides(resources=undefined)
-                for resource in newly_defined:
-                    self._state.mark_as_defined(resource, resources[resource])  # Updates the dirty set
-                # Update set of in-progress non-stale deploys by trimming resources with new state
-                self._deploying_latest.difference_update(
-                    new_desired_state, deleted_resources, undefined, blocked_for_undefined_dep
+
+                t_unblocked, t_blocked = self._state.update_transitive_state(
+                    undefined, added_requires.keys(), now_defined | dropped_requires.keys()
                 )
+
+                # blocked_for_undefined_dep: Set[ResourceIdStr] = self._state.block_provides(resources=undefined)
+
+                # Update set of in-progress non-stale deploys by trimming resources with new state
+                self._deploying_latest.difference_update(new_desired_state, deleted_resources, undefined, t_blocked)
 
                 # Remove timers for resources that are:
                 #    - in the dirty set (because they will be picked up by the scheduler eventually)
                 #    - blocked: must not be deployed
                 #    - deleted from the model
-                self._timer_manager.stop_timers(self._state.dirty | undefined | blocked_for_undefined_dep)
+                self._timer_manager.stop_timers(self._state.dirty | undefined | t_blocked)
                 self._timer_manager.remove_timers(deleted_resources)
                 # Install timers for initial up-to-date resources. They are up-to-date now,
                 # but we want to make sure we periodically repair them.
@@ -582,7 +597,7 @@ class ResourceScheduler(TaskManager):
                 )
                 for resource in deleted_resources:
                     self._state.drop(resource)  # Removes from the dirty set
-                for resource in undefined | blocked_for_undefined_dep:
+                for resource in undefined | t_blocked:
                     self._work.delete_resource(resource)
 
             # Once more, drop all resources that do not exist in this version from the scheduled work, in case they got added
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -554,7 +554,7 @@ class ResourceScheduler(TaskManager):
             async with self._scheduler_lock:
                 self._state.version = version
                 for resource in undefined:
-                    self._state.block_and_update_resource(resource, resources[resource])  # Removes from the dirty set
+                    self._state.update_resource_to_undefined(resource, resources[resource])  # Removes from the dirty set
                 for resource in now_defined:
                     # the resources moving out of undefined, normal update
                     self._state.update_desired_state(resource, resources[resource])  # Updates the dirty set
@@ -566,8 +566,6 @@ class ResourceScheduler(TaskManager):
                 t_unblocked, t_blocked = self._state.update_transitive_state(
                     undefined, added_requires.keys(), now_defined | dropped_requires.keys()
                 )
-
-                # blocked_for_undefined_dep: Set[ResourceIdStr] = self._state.block_provides(resources=undefined)
 
                 # Update set of in-progress non-stale deploys by trimming resources with new state
                 self._deploying_latest.difference_update(new_desired_state, deleted_resources, undefined, t_blocked)
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -24,7 +24,7 @@ import uuid
 from abc import abstractmethod
 from collections.abc import Collection, Mapping, Set
 from dataclasses import dataclass
-from typing import Optional
+from typing import Optional, Tuple
 from uuid import UUID
 
 import asyncpg
@@ -34,7 +34,7 @@ from inmanta.agent import executor
 from inmanta.agent.code_manager import CodeManager
 from inmanta.agent.executor import DeployResult, FactResult
 from inmanta.data import ConfigurationModel, Environment
-from inmanta.data.model import ResourceIdStr, ResourceType, ResourceVersionIdStr
+from inmanta.data.model import Discrepancy, ResourceIdStr, ResourceType, ResourceVersionIdStr, SchedulerStatusReport
 from inmanta.deploy import timers, work
 from inmanta.deploy.persistence import StateUpdateManager, ToDbUpdateManager
 from inmanta.deploy.state import (
@@ -46,8 +46,8 @@ from inmanta.deploy.state import (
     ResourceDetails,
     ResourceState,
 )
-from inmanta.deploy.tasks import Deploy, DryRun, RefreshFact
-from inmanta.deploy.work import PrioritizedTask, TaskPriority
+from inmanta.deploy.tasks import Deploy, DryRun, RefreshFact, Task
+from inmanta.deploy.work import TaskPriority
 from inmanta.protocol import Client
 from inmanta.resources import Id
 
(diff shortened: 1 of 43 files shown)


src/inmanta/deploy/state.py
```diff
@@ -210,66 +215,193 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
...
+        #  - a worklist of potential unblocked nodes
+
+        #  - a cache of blocker nodes. When we find a node B blocking a potentially unblocked node A, we cache this.
+        #    If we hit node A again and we are not coming from B, we know it is still blocked.
+
+        out_blocked: set[ResourceIdStr] = set()
+        out_unblocked: set[ResourceIdStr] = set()
+
+        # a set of definitely blocked nodes
+        is_blocked: set[ResourceIdStr] = set(now_blocked)
```
wouterdb, line 266: do we need to to keep this optimization?

src/inmanta/deploy/state.py
```diff
@@ -329,6 +466,9 @@ def update_requires(self, resource: ResourceIdStr, requires: Set[ResourceIdStr])
         """
         Update the requires relation for a resource. Updates the reverse relation accordingly.
         """
+
+        # TODO: verify this is correct and needed
```
wouterdb, line 470: @sanderr I don't know about this bit very well.

tests/deploy/test_scheduler_agent.py
```diff
@@ -105,11 +106,14 @@ async def execute(
         # The actual reasons are of the form `action because of reason`
         assert ("because" in reason) or ("Test" in reason)
         self.execute_count += 1
-        result = (
-            const.HandlerResourceState.failed
-            if resource_details.attributes.get(FAIL_DEPLOY, False) is True
-            else const.HandlerResourceState.deployed
-        )
+        if any(status != const.HandlerResourceState.deployed for status in requires.values()):
+            result = const.HandlerResourceState.skipped_for_dependency
```
wouterdb, line 110: Propagate skip

tests/deploy/test_scheduler_agent.py
```diff
@@ -1838,3 +1842,208 @@ async def test_state_of_skipped_resources_for_dependencies(agent: TestAgent, mak
...
+    def is_new_blocked(rid: ResourceIdStr):
+        assert agent.scheduler._state.resource_state[rid].deployment_result == DeploymentResult.NEW
+        assert agent.scheduler._state.resource_state[rid].status == ComplianceStatus.HAS_UPDATE
+        assert agent.scheduler._state.resource_state[rid].blocked == BlockedStatus.YES
+
+    # Chain of 3
+    # 3 -> 1 -> 2
+    # 1 is undefined
+    # 2 is free to deploy
+    # 3 is undefined
```
jptrindade, line 1918: 3 is blocked, no?

tests/deploy/test_scheduler_agent.py
```diff
@@ -1838,3 +1842,208 @@ async def test_state_of_skipped_resources_for_dependencies(agent: TestAgent, mak
...
+    is_blocked(rid2)
+    is_blocked(rid3)
+    is_blocked(rid5)
+    is_deployed(rid4)
+    is_deployed(rid7)
+    is_deployed(rid8)
+    for rid in range(9, 14):
+        is_deployed(f"test::Resource[agent1,name={rid + 1}]")
+
+    # Also unblock
```
jptrindade, line 2030:
```suggestion
    # Also unblock 9
```

src/inmanta/deploy/state.py
```diff
@@ -210,66 +215,193 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
...
+            if blocked_dependency:
+                # Resource is blocked, because a dependency is blocked.
+                known_blockers_cache[resource] = blocked_dependency
+
+            if blocked_dependency in is_blocked:
+                # We know the root is undefined, hard block
+                is_blocked.add(resource)
+
+            # Unblock resource
+            self._block_resource_transitive(resource)
```
jptrindade, line 353: I assume the comment is wrong

src/inmanta/deploy/state.py
```diff
@@ -210,66 +215,193 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
...
+            out_blocked.add(resource)
+            return True
+
+        for resource_id in verify_blocked:
+            # verify all potential blockers
+            blocked = update_blocked_status_to_blocked(resource_id)
+            # Keep track of this that have been blocked
```
jptrindade, line 360: This comment could be more clear

jptrindade (changes requested):
The logic on state.py was a bit confusing to me, tomorrow morning I'll check it out locally and see if I can make sense of it.

src/inmanta/deploy/scheduler.py
```diff
@@ -528,16 +530,20 @@ async def _new_version(
...
                 old_requires: Set[ResourceIdStr] = self._state.requires.get(resource, set())
                 new_requires: Set[ResourceIdStr] = requires.get(resource, set())
-                added_requires[resource] = new_requires - old_requires
-                dropped_requires[resource] = old_requires - new_requires
+                my_added = new_requires - old_requires
+                if my_added:
+                    added_requires[resource] = my_added
+                my_removed = old_requires - new_requires
+                if my_removed:
+                    dropped_requires[resource] = my_removed
```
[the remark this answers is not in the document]
wouterdb: we added it to the dict even when empty, and then use the keys of the dict to see what has changes, so that was wrong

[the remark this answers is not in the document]
jptrindade: Hum I see now

src/inmanta/deploy/state.py
```diff
@@ -174,31 +175,35 @@ def add_up_to_date_resource(self, resource: ResourceIdStr, details: ResourceDeta
             self.types_per_agent[details.id.agent_name][details.id.entity_type] += 1
         self.dirty.discard(resource)
 
-    def block_resource(self, resource: ResourceIdStr, details: ResourceDetails, is_transitive: bool) -> None:
+    def block_and_update_resource(self, resource: ResourceIdStr, details: ResourceDetails) -> None:
```
jptrindade, line 178: I think the name of this method should indicate that we set the status to undefined. When I read it I thought it only put blocked to `YES`

src/inmanta/deploy/state.py
```diff
@@ -210,66 +215,193 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
 
         return all(self.resource_state[dep_id].status is ComplianceStatus.COMPLIANT for dep_id in dependencies)
 
-    def block_provides(self, resources: Set[ResourceIdStr]) -> Set[ResourceIdStr]:
+    def update_transitive_state(
+        self,
+        now_blocked: AbstractSet[ResourceIdStr],
```
jptrindade, line 220: should also mention undefined in the argument name, imo

src/inmanta/deploy/state.py
```diff
@@ -210,66 +215,193 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
...
+        for resource_id in verify_blocked:
+            # verify all potential blockers
+            blocked = update_blocked_status_to_blocked(resource_id)
+            # Keep track of this that have been blocked
+            if blocked:
+                propagate_blocked_work.append(resource_id)
+
+        # Propagate blockers once more
+        # We iterate differently, because the parents blocked status is relevant
+        propagate_blocked_work = list(verify_blocked)
```
jptrindade, line 366: This doesn't make sense to me, why are we appending the blocked resources to `propagate_blocked_work` if we are just going to overwrite it?

src/inmanta/deploy/state.py
```diff
@@ -210,66 +215,193 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
...
-        are populated in the resources dictionary.
+        Update transitive states
 
-        :param resources: The set of resources for which the provides have to be blocked.
-        :return: The set of dependent resources that were newly marked as blocked (transitively).
+        :param now_blocked: resources that have become undefined
+        :param verify_blocked: resources that have added requires
+        :param verify_unblocked: resources that have lost requires relations
+
+        returns: <blocked, unlocked> resources
```
jptrindade, line 232: This returns is in the wrong order. And I assume that we mean `unblocked` 
```suggestion
        returns: <unblocked, blocked> resources
```

src/inmanta/deploy/state.py
```diff
@@ -210,66 +215,193 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
...
 
             :return: True iff the given resource was unblocked.
             """
-            if self.resource_state[resource].blocked is not BlockedStatus.YES:
+            if resource in is_blocked:
+                # Is blocked for a know root
+                return False
+
+            my_state = self.resource_state[resource]
+            if my_state.blocked is not BlockedStatus.YES:
```
jptrindade, line 395: Do we also want to unblock `BlockedStatus.TRANSIENT` ?

src/inmanta/deploy/state.py
jptrindade: I don't think this is needed anymore. But you also have to check for `BlockedStatus.TRANSIENT`  when unblocking.

jptrindade (changes requested):

src/inmanta/deploy/state.py
wouterdb: worklist algorthm! 

1. We consume form the list until it is empty
2. Here we initialize it with initial work
3. every iteration adds new work to the list (the code is defined above, but execute after this statement)


It is a way if doing (tail) recursion without ever running out of stack

src/inmanta/deploy/state.py
wouterdb: @sanderr I don't think so?

src/inmanta/deploy/state.py
```diff
@@ -204,6 +204,15 @@ def _block_resource_transitive(self, resource: ResourceIdStr) -> None:
         self.resource_state[resource].blocked = BlockedStatus.YES
         self.dirty.discard(resource)
 
+    def _unblock_resource(self, resource: ResourceIdStr) -> None:
+        """
+        Mark the given resource as transitively blocked, i.e. it's not deployable.
```
jptrindade, line 209: This docstring does not match the function

src/inmanta/deploy/state.py
jptrindade: I don't get this part before we overwrite the `propagate_blocked_work` and start consuming from the list:
```
        for resource_id in verify_blocked:
            # verify all potential blockers
            blocked = update_blocked_status_to_blocked(resource_id)
            # Keep track of this that have been blocked
            if blocked: <------- Specifically this part and the line below
                propagate_blocked_work.append(resource_id)
```

jptrindade (changes requested):
Some clarifications on the open discussion and a docstring that requires fixing

src/inmanta/deploy/state.py
wouterdb: right!

jptrindade (approved):

src/inmanta/deploy/state.py
sanderr: 1. We must set `TRANSIENT -> YES` when applicable
2. Ideally, `TRANSIENT` remains `TRANSIENT` when 1 does not apply. The current implementation might occasionally set `TRANSIENT -> YES -> NO` e.g. if a resource got a new dependency that happens to be in `verify_unblocked`. This is however acceptable imo.

src/inmanta/deploy/state.py
```diff
@@ -210,66 +224,192 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
...
+        for resource_id in verify_blocked:
+            # verify all potential blockers
+            blocked = update_blocked_status_to_blocked(resource_id)
+            # if we did block it, add it to the work
+            if blocked:
+                propagate_blocked_work.append(resource_id)
+
+        # Propagate blockers once more
+        # We iterate differently, because the parents blocked status is relevant
+        while propagate_blocked_work:
```
sanderr, line 375: Note to self: continue review from here.

sanderr (changes requested):
I had a realization while reviewing. I'll take that to Slack, might be best to resolve that before processing my comments. I'm releasing them in the meantime anyway. I did stop halfway through, because the realization might lead to major changes.

src/inmanta/deploy/scheduler.py
```diff
@@ -548,24 +554,27 @@ async def _new_version(
...
                 for resource in added_requires.keys() | dropped_requires.keys():
                     self._state.update_requires(resource, requires[resource])
-                blocked_for_undefined_dep: Set[ResourceIdStr] = self._state.block_provides(resources=undefined)
-                for resource in newly_defined:
-                    self._state.mark_as_defined(resource, resources[resource])  # Updates the dirty set
-                # Update set of in-progress non-stale deploys by trimming resources with new state
-                self._deploying_latest.difference_update(
-                    new_desired_state, deleted_resources, undefined, blocked_for_undefined_dep
+
+                t_unblocked, t_blocked = self._state.update_transitive_state(
```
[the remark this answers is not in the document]
wouterdb: transitive, but probably not good, convention in C is t_ for types...

tests/deploy/test_scheduler_agent.py
```diff
@@ -57,7 +58,7 @@
 
 async def retry_limited_fast(
     fun: Callable[..., bool] | Callable[..., Awaitable[bool]],
-    timeout: float = 0.1,
+    timeout: float = 15,
```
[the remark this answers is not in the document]
wouterdb: debugging, should be reverted

src/inmanta/deploy/scheduler.py
```diff
@@ -548,24 +554,27 @@ async def _new_version(
             async with self._scheduler_lock:
                 self._state.version = version
                 for resource in undefined:
-                    self._state.block_resource(resource, resources[resource], is_transitive=False)  # Removes from the dirty set
+                    self._state.update_resource_to_undefined(resource, resources[resource])  # Removes from the dirty set
+                for resource in now_defined:
+                    # the resources moving out of undefined, normal update
+                    self._state.update_desired_state(resource, resources[resource])  # Updates the dirty set
```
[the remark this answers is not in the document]
wouterdb: nice!

src/inmanta/deploy/scheduler.py
```diff
@@ -502,7 +502,7 @@ async def _new_version(
             # Only contains the direct undeployable resources, not the transitive ones.
             undefined: set[ResourceIdStr] = set()
             # Resources that were undeployable in a previous model version, but got unblocked. Not the transitive ones.
-            newly_defined: set[ResourceIdStr] = set()  # resources that were previously undefined but not anymore
+            now_defined: set[ResourceIdStr] = set()  # resources that were previously undefined but not anymore
```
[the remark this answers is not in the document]
wouterdb: I know, this is the third attempt on my side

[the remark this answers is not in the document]
sanderr: `newly_defined` was the third on my side!

src/inmanta/deploy/state.py
```diff
@@ -210,66 +224,192 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
 
         return all(self.resource_state[dep_id].status is ComplianceStatus.COMPLIANT for dep_id in dependencies)
 
-    def block_provides(self, resources: Set[ResourceIdStr]) -> Set[ResourceIdStr]:
+    def update_transitive_state(
+        self,
+        new_undefined: AbstractSet[ResourceIdStr],
+        verify_blocked: AbstractSet[ResourceIdStr],
```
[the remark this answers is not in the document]
wouterdb: dang, Set vs set, I knew there had to be a beter solution. thx

src/inmanta/deploy/state.py
```diff
@@ -210,66 +224,192 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
...
+            # one blocked requires, evidence we are blocked
+
+            if resource in known_blockers_cache:
+                # First check the blocked status of the cached known blocker for improved performance.
+                known_blocker: ResourceIdStr = known_blockers_cache[resource]
+                if self.resource_state[known_blocker].blocked is BlockedStatus.YES:
+                    blocked_dependency = known_blocker
+                else:
+                    # Cache is out of date. Clear cache item.
+                    del known_blockers_cache[resource]
```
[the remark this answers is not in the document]
wouterdb: I think this is correct: 
1. `known_blockers_cache` contains, for every resource A, a resource B that is seen to be blocking the resource A (by definition a direct dependency of A)
2. if this witness it no longer blocked, we drop the cache entry and continue 

I think this is fairly efficient, as it avoids a reverse cache lookup.

src/inmanta/deploy/state.py
```diff
@@ -210,66 +224,192 @@ def are_dependencies_compliant(self, resource: ResourceIdStr) -> bool:
...
+            if blocked_dependency is None:
+                return False
+
+            if blocked_dependency:
+                # Resource is blocked, because a dependency is blocked.
+                known_blockers_cache[resource] = blocked_dependency
+
+            if blocked_dependency in is_blocked:
+                # We know the root is undefined, hard block
+                is_blocked.add(resource)
```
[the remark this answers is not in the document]
wouterdb: So you would remove it?

src/inmanta/deploy/state.py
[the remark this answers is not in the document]
sanderr: Oh wow, it was clearly the end of a long review day. I read this block as `for ... in ...` rather than `if ... in ...` This is exactly what I would expect.

src/inmanta/deploy/state.py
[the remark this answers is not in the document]
sanderr: I think so, but I need to reconsider with the new context I earned yesterday (through my shortsighted alternative proposal).

sanderr (commented):
Responded to threads, somehow got marked as review comments.

sanderr (changes requested):
Final comments, one of which is an actual bug

sanderr (approved):

PR #8466, opened 2024-12-11
Issue/type untangling

wouterdb:
# Description

Make type tangle less bad:
1. move resource id string type to types module 
2. make resources.ID less prone to loops by making resources depend on less
   - make `inmanta.ast` contain all exceptions of the compiler
   - make `inmanta.ast` import very little else.
   
   
   
My think towards general rules:
1. a package can offer an interface module (called now something like `model`/`types`/`__init__`)..... 
   - This module contains the external interface of the module for other module to consume
      - superclasses
      - exceptions
      - interface types (executor.executor e.g), 
      - datatypes,...
   - This module should import the least stuff possible
   - It should never import any non-interface modules
 2. typing only imports can do what they want 
 3. Packages should roughly form a tree, 
   - where only composing modules import the non-interface modules of other packages.
   - and interface modules should also only import interface modules of composing modules  
   
 
 e.g. consider executor and deploy packages and the new_agent
 
 1. `executor.executor` is the interface package
 2. `deploy` imports `executor.executor` 
 3. `new_agent` import both the interface and concrete implementations of both `executor` and `deploy` (top level)
 4. `deploy` doesn't import the non-interface parts of `executor` because it never constructs executors (it is constructed with a reference to a executor manager, which serves as a factory)
 5. `executor` doesn't know about `deploy` at all 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/agent_new.py
+++ b/src/inmanta/agent/agent_new.py
@@ -22,7 +22,6 @@ import uuid
 from concurrent.futures.thread import ThreadPoolExecutor
 from typing import Any, Optional
 
-import inmanta.protocol  # noqa: F401 prevent import loop
 import inmanta.server.config as opt
 from inmanta import config, const, data, protocol
 from inmanta.agent import config as cfg
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -29,7 +29,6 @@ from uuid import UUID
 
 import asyncpg
 
-import inmanta.protocol  # noqa: F401 prevent import loop
 from inmanta import const, data
 from inmanta.agent import executor
 from inmanta.agent.code_manager import CodeManager


src/inmanta/data/model.py
```diff
@@ -202,22 +201,6 @@ class CompileDetails(CompileReport):
     reports: Optional[list[CompileRunReport]] = None
 
 
-ResourceVersionIdStr = NewType("ResourceVersionIdStr", str)  # Part of the stable API
```
wouterdb, line 205: Real change, check bottom of file

src/inmanta/execute/proxy.py
```diff
@@ -32,48 +33,10 @@
     TYPE_CHECKING = False
 
 if TYPE_CHECKING:
-    from inmanta.ast.attribute import Attribute
     from inmanta.ast.entity import Entity
     from inmanta.execute.runtime import Instance
 
 
-class UnsetException(RuntimeException):
```
wouterdb, line 40: Real change, check top of file

bartv (approved):
A PITA to review but a nice improvement

src/inmanta/ast/__init__.py
```diff
@@ -996,3 +995,50 @@ def format_trace(self, indent: str = "", indent_level: int = 0) -> str:
...
+    depending on this value by return an instance of Unknown as well.
+    """
+
+    def __init__(self, unknown: "Unknown"):
+        super().__init__()
+        self.unknown = unknown
+
+
+@stable_api
+class PluginException(Exception):
```
sanderr, line 1038: I don't like this location for the stable API though. Have to give this one some more thought.

src/inmanta/agent/agent_new.py
```diff
@@ -22,6 +22,7 @@
 from concurrent.futures.thread import ThreadPoolExecutor
 from typing import Any, Optional
 
+import inmanta.protocol  # noqa: F401 prevent import loop
```
sanderr, line 25: What's this?

src/inmanta/ast/__init__.py
```diff
@@ -38,8 +38,7 @@
     from inmanta.ast.statements.define import DefineEntity, DefineImport  # noqa: F401
     from inmanta.ast.type import NamedType, Type  # noqa: F401
     from inmanta.compiler import Compiler
-    from inmanta.execute.runtime import DelayedResultVariable, ExecutionContext, Instance, ResultVariable  # noqa: F401
-    from inmanta.plugins import PluginException
+    from inmanta.execute.runtime import DelayedResultVariable, ExecutionContext, Instance, ResultVariable, Unknown  # noqa: F401
```
sanderr, line 41: Do you know why we have the noqa here?

src/inmanta/deploy/scheduler.py
```diff
@@ -29,12 +29,12 @@
 
 import asyncpg
 
+import inmanta.protocol  # noqa: F401 prevent import loop
```
sanderr, line 32: Same question, what's this?

src/inmanta/execute/runtime.py
```diff
@@ -20,7 +20,9 @@
 from collections.abc import Hashable, Sequence, Set
 from typing import TYPE_CHECKING, Deque, Generic, List, Literal, NewType, Optional, TypeVar, Union, cast
 
+import inmanta.ast
```
sanderr, line 23: We already use this one partially qualified as `ast` here (three lines down). We should stay consistent.

src/inmanta/execute/runtime.py
```diff
@@ -20,7 +20,9 @@
...
 import inmanta.ast.attribute  # noqa: F401 (pyflakes does not recognize partially qualified access ast.attribute)
+import inmanta.execute
```
sanderr, line 25: Why do we need this?

sanderr (commented):
Some notes. Overall looks like a nice improvement. Didn't review the tests.

src/inmanta/ast/__init__.py
wouterdb: I agree to that

src/inmanta/agent/agent_new.py
wouterdb: it prevents an import loop (nasty one we already have, but we always happen to load protocol before inmanta.data.model)

src/inmanta/ast/__init__.py
wouterdb: I added it so pep8 doesn't complain the import is not used (but the import is there to keep the symbols in place)

src/inmanta/execute/runtime.py
wouterdb: mhhh, I don't think we do, good carch

src/inmanta/ast/__init__.py
sanderr: But this one was already here.

src/inmanta/agent/agent_new.py
sanderr: Ah no, you didn't fix it, just worked around it here?

sanderr: I also can't seem to reproduce it. Can you tell me more?

wouterdb: I managed to remove two already

wouterdb:
```
conftest.py:31: in <module>
    from inmanta.agent.handler import CRUDHandler, HandlerContext, ResourceHandler, SkipResource, TResource, provider
../src/inmanta/agent/__init__.py:19: in <module>
    from inmanta.agent.agent_new import Agent
../src/inmanta/agent/agent_new.py:26: in <module>
    from inmanta import config, const, data, protocol
../src/inmanta/data/__init__.py:55: in <module>
    from inmanta.data import model as m
../src/inmanta/data/model.py:34: in <module>
    import inmanta.protocol
../src/inmanta/protocol/__init__.py:56: in <module>
    from . import methods, methods_v2
../src/inmanta/protocol/methods.py:30: in <module>
    from inmanta.protocol.common import ArgOption
../src/inmanta/protocol/common.py:46: in <module>
    from inmanta.data.model import BaseModel, DateTimeNormalizerModel
E   ImportError: cannot import name 'BaseModel' from partially initialized module 'inmanta.data.model' (most likely due to a circular import) (/home/wouter/projects/inmanta-core/src/inmanta/data/model.py)
```

arnaudsjs (approved):
I don't have time for a review this size, so I am just going to approve this PR without review.

PR #8468, opened 2024-12-12
Issue/1597 drop inmanta module commit command

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/irt/issues/1597

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/1597-drop-inmanta-module-commit-command.yml
+++ b/changelogs/unreleased/1597-drop-inmanta-module-commit-command.yml
@@ -1,0 +1,10 @@
+description: >
+  Remove deprecated 'inmanta module commit' command.
+  Please use the 'inmanta module release' command instead to perform
+  a release for a module.
+issue-nr: 1597
+change-type: minor
+destination-branches: [master]
+sections:
+  deprecation-note: "{{description}}"
+
--- a/changelogs/unreleased/1597-drop-inmanta-module-commit-command.yml
+++ b/changelogs/unreleased/1597-drop-inmanta-module-commit-command.yml
@@ -3,6 +3,7 @@ description: >
   Please use the 'inmanta module release' command instead to perform
   a release for a module.
 issue-nr: 1597
+issue-repo: irt
 change-type: minor
 destination-branches: [master]
 sections:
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/1597-drop-inmanta-module-commit-command.yml
```diff
@@ -0,0 +1,11 @@
+description: >
+  Remove deprecated 'inmanta module commit' command.
+  Please use the 'inmanta module release' command instead to perform
+  a release for a module.
+issue-nr: 1597
+issue-repo: irt
+change-type: minor
```
sanderr, line 7:
```suggestion
change-type: major
```

sanderr (approved):

Hugo-Inmanta:
I initially removed the underlying logic (i.e. the [commit method](https://github.com/inmanta/inmanta-core/blob/f586d461e5fe9d4af02943d234aff6777faa827a/src/inmanta/moduletool.py#L1022) itself) but decided against it since it is still used to set up test cases e.g. https://github.com/inmanta/inmanta-core/blob/d048e728d2a0a518801b39a76d8bd039e8a8195b/tests/moduletool/conftest.py#L124

sanderr (changes requested):
> but decided against it since it is still used to set up test cases

Ugh, hadn't considered that. Probably a good call to keep it for now. Still, could you add a comment / add to the docstring that it's kept purely for testing purposes?

sanderr:
@wouterdb I'd like to draw your attention to [this comment](https://github.com/inmanta/inmanta-core/pull/8468#issuecomment-2538435887)

src/inmanta/moduletool.py
```diff
@@ -620,25 +620,6 @@ def modules_parser_config(cls, parser: ArgumentParser, parent_parsers: abc.Seque
             parents=parent_parsers,
         )
 
-        commit = subparser.add_parser("commit", help="Commit all changes in the current module.")
```
wouterdb, line 623: This is not sufficient, you need to remove the `commit` method as well (and any dead code it causes)

wouterdb (changes requested):

wouterdb:
Can't the tests move to the release command?

sanderr (approved):
The removal of the command looks good to me. But I'm afraid I can't spare the time at the moment to review the changes to the tests in detail. I glanced them over and they look OK. I'll leave the rest to @wouterdb.

wouterdb (approved):

Issue #8304, opened 2024-11-04
Update installation instructions native docker container


arnaudsjs:
The old docker container and the new native docker container are not entirely backwards compatible. For example, the new container doesn't put any config files in place, which makes it fallback to the default configuration options. 

Tasks:

- [ ] Verify that the installation instructions work for the new native container
- [ ] Mention that configuration options can be provided via environment variables. Refer to: https://github.com/inmanta/inmanta-core/issues/8303
- [ ] Add an upgrade note to the product repos describing the breaking changes with respect to the non-native docker container.
     - [ ]  ISO
     - [ ] OSS
- [ ] Ensure the oss docs are also updated

sanderr:
-> in inmanta-core repo, for OSS and iso, on master

Issue #8421, opened 2024-11-27
Document ssh sidecar setup (docker install)


edvgui:
Starting from iso8, the ssh access inside the orchestrtaor container has changed, it requires the usage of ssh sidecar, as illustrated here: https://github.com/inmanta/inmanta-docker?tab=readme-ov-file#deploy-the-service-orchestrator-with-an-ssh-sidecar

This should be properly documented.  
Some first step was done here but had to be dropped (priority prompting): https://github.com/inmanta/inmanta-core/pull/8420

PR #8476, opened 2024-12-13
Update installation instructions native docker container

DariushMoh:
# Description

* Short description here *

closes https://github.com/inmanta/inmanta-core/issues/8304#issue-2632714023

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/8304-update-installation-instructions-native-docker-container.yml
+++ b/changelogs/unreleased/8304-update-installation-instructions-native-docker-container.yml
@@ -1,0 +1,5 @@
+description: >
+  Update installation instructions for new native docker container
+issue-nr: 8303
+change-type: patch
+destination-branches: [master]
--- a/changelogs/unreleased/8304-update-installation-instructions-native-docker-container.yml
+++ b/changelogs/unreleased/8304-update-installation-instructions-native-docker-container.yml
@@ -1,5 +1,4 @@
-description: >
-  Update installation instructions for new native docker container
-issue-nr: 8303
+description: Update installation instructions for new native docker container
+issue-nr: 8304
 change-type: patch
 destination-branches: [master]
--- a/docs/install/2-install-server-with-docker.rst
+++ b/docs/install/2-install-server-with-docker.rst
@@ -268,7 +268,7 @@ There are two ways you can achieve this:
                 - ./resources/my-server-conf.cfg:/etc/inmanta/inmanta.cfg
 
 Mounting files/directories
-########################
+##########################
 
 The recommended way to mount files and directories is to use docker volumes:
 
--- a/docs/install/2-install-server-with-docker.rst
+++ b/docs/install/2-install-server-with-docker.rst
@@ -152,7 +152,7 @@ Here is a minimalistic docker-compose file content that can be used to deploy th
         volumes:
             pgdata:
 
-    You can paste this yaml snippet in a file named `docker-compose.yml` and ensure you have you license files available.
+    You can paste this yaml snippet in a file named `docker-compose.yml` and ensure you have your license files available.
     With the proposed config, they should be located in a ``resources/`` folder on the side of the docker-compose file you create,
     and the license files should be named ``com.inmanta.license`` and ``com.inmanta.jwe``. You can of course update the content
     of the docker-compose file to match your current configuration.
@@ -168,8 +168,8 @@ The PostgreSQL server started by the above-mentioned docker-compose file has a n
 means that no data will be lost when the PostgreSQL container restarts. Pass the ``-v`` option to the
 ``docker-compose down`` to remove the volume.
 
-The default server config included in the container images assumes by default that the orchestrator can reach a database server on localhost.
-To change this behavior you can use the env variables as shown in the snippet above.
+The default server config assumes that the orchestrator can reach the database server on localhost.
+To change this behavior you can use the environment variables as shown in the snippet above.
 When using a different setup than the one mentioned above, you should overwrite the server config with one
 matching your needs.  You can find more instructions for overwriting the config in a following section,
 :ref:`here<docker_overwrite_server_conf>`.
@@ -185,10 +185,10 @@ matching your needs.  You can find more instructions for overwriting the config 
 Overwrite default server configuration
 ######################################
 
-If you want to change default server configuration, the recommended way is to provide the server
+If you want to change the default server configuration, the recommended way is to provide the server
 config options via environment variables as done in the above example.
-All the different options and associated env variables are described :ref:`here<config_reference>`.
-It is also possible to provide the configuration file, make sure to mount it in ``/etc/inmanta/inmanta.cfg``.
+All the different options and associated environment variables are described :ref:`here<config_reference>`.
+It is also possible to provide a configuration file. Make sure to mount it in ``/etc/inmanta/inmanta.cfg``.
 
 Starting the ssh server
 #######################
@@ -284,8 +284,8 @@ And then you can use it in docker-compose file:
         - mydockervolume:/etc/inmanta/myfolder
 
 
-However if you really need to mount from the host, you can use bind mounts. You just need to make sure to change the ownership of
-the file/directory you want to mount to make sure it has same uid/gid as the inmanta user. To find them, in the container, you can use ``id`` command.
+However if you really need to mount a file or directory from the host, you can use bind mounts. You just need to make sure to change the ownership of
+the file/directory you want to mount to make sure it has same uid/gid as the inmanta user inside the container. To find them, in the container, you can use ``id`` command.
 By default, currently, inmanta user ``uid`` is 997 and ``gid`` is 995. On your host you can easily change ownership of your file/directory with these values:
 
 .. code-block:: sh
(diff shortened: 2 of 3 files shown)


arnaudsjs:
> I'm a bit confused because the ticket mention that the new native docker container is not backward compatible, which should causes some issues. However, I managed to deploy the orchestrator (both oss and iso) without changing anything in docker-compose setup (the orchestrator was up and running. But I mounted the inmanta.cfg in the container and the options I modified do not seem to be taken into account in environment settings (not sure if it should be the case).

Did you pull the latest (dev) image for OSS and ISO8? It should not be possible that https://docs.inmanta.com/inmanta-service-orchestrator-dev/8/install/2-install-server-with-docker.html works. For example, the section "Starting the ssh server" should not work because the `server-with-ssh` command is no longer available in the new image.

arnaudsjs (changes requested):
See comments.

docs/install/2-install-server-with-docker.rst
```diff
@@ -131,6 +138,13 @@ Here is a minimalistic docker-compose file content that can be used to deploy th
                 volumes:
                     - ./resources/com.inmanta.license:/etc/inmanta/license/com.inmanta.license
                     - ./resources/com.inmanta.jwe:/etc/inmanta/license/com.inmanta.jwe
```
DariushMoh, line 140: Unfortunately I cannot use env variables for that in the example. I guess this will come later.

docs/install/2-install-server-with-docker.rst
```diff
@@ -181,39 +195,11 @@ Overwrite default server configuration
...
-        inmanta-server:
-            container_name: inmanta_orchestrator
-            image: ghcr.io/inmanta/orchestrator:latest
-            ports:
-                - 8888:8888
-            volumes:
-                - ./resources/my-server-conf.cfg:/etc/inmanta/inmanta.cfg
-
 
 Starting the ssh server
```
DariushMoh, line 204: This section will be addressed in this ticket: https://github.com/inmanta/inmanta-core/issues/8421

docs/install/2-install-server-with-docker.rst
```diff
@@ -291,9 +275,16 @@ There are two ways you can achieve this:
             image: ghcr.io/inmanta/orchestrator:latest
             ports:
                 - 8888:8888
+            env_file: ./resources/my-env-file
+            environment:
+                INMANTA_DATABASE_HOST: inmanta_db
```
DariushMoh, line 280: I did add both ways to do it in the example, they can actually be combined and this clearly show how to use both of them. Not sure if this is alright.

docs/install/2-install-server-with-docker.rst
```diff
@@ -305,11 +296,18 @@ There are two ways you can achieve this:
...
+                INMANTA_DATABASE_USERNAME: inmanta
+                INMANTA_DATABASE_PASSWORD: inmanta
+                INMANTA_CONFIG_AGENT_DEPLOY_INTERVAL: 800
+                INMANTA_CONFIG_AGENT_REPAIR_INTERVAL: 86400
+                INMANTA_SERVER_BIND_PORT: 8888
```
arnaudsjs, line 306: * The log rotation section talks about the logs of the SSH server. That's not required anymore, because the container doesn't have an SSH server anymore.
* The handling of the `INMANTA_UID` and `INMANTA_GID` environment variables is also something that was done by the old entrypoint, right?

arnaudsjs (changes requested):
Did you also verify whether [this page](https://docs.inmanta.com/inmanta-service-orchestrator-dev/8/install/2-install-server-with-podman-and-systemd.html) is still correct with the new/native container?

docs/install/2-install-server-with-docker.rst
```diff
@@ -181,39 +195,11 @@ Overwrite default server configuration
 ######################################
 
 By default the server will use the file located in the image at ``/etc/inmanta/inmanta.cfg``.
-If you want to change it, you can copy this file, edit it, then mount it in the container,
+If you want to change it, the recommended way is to provide the server config options via environment variables as done in the above example.
```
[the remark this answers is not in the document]
DariushMoh: Right I forgot to update this part.

changelogs/unreleased/8304-update-installation-instructions-native-docker-container.yml
```diff
@@ -0,0 +1,5 @@
+description: >
+  Update installation instructions for new native docker container
+issue-nr: 8303
+change-type: patch
+destination-branches: [master]
```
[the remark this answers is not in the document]
arnaudsjs: True. That's better.

docs/install/2-install-server-with-docker.rst
```diff
@@ -85,6 +85,13 @@ Here is a minimalistic docker-compose file content that can be used to deploy th
                 image: ghcr.io/inmanta/orchestrator:latest
                 ports:
                     - 8888:8888
+                environment:
+                    INMANTA_DATABASE_HOST: inmanta_db
+                    INMANTA_DATABASE_USERNAME: inmanta
+                    INMANTA_DATABASE_PASSWORD: inmanta
+                    INMANTA_CONFIG_AGENT_DEPLOY_INTERVAL: 800
+                    INMANTA_CONFIG_AGENT_REPAIR_INTERVAL: 86400
+                    INMANTA_SERVER_BIND_PORT: 8888
```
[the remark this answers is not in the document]
DariushMoh: It was just to give an example of what can be provided (even if the provided value is the same as default).

docs/install/2-install-server-with-docker.rst
DariushMoh: I'm not quite sure about the UID and GID. I never had to deal with them when deploying the orchestrator with container.

DariushMoh: After some more testing, uid and gid env do not work anymore.

DariushMoh:
I added upgrade notes [here](https://github.com/inmanta/inmanta/pull/143) and [here](https://github.com/inmanta/inmanta-service-orchestrator/pull/541)

docs/install/2-install-server-with-docker.rst
```diff
@@ -305,42 +257,40 @@ There are two ways you can achieve this:
...
-functionality.
-The inmanta container allows you to change the user and group id of the inmanta user inside the
-container when starting the container to match the user on the host, getting rid that way of any
-conflict in the files ownership.
+.. code-block:: sh
 
-This can be done easily by simply setting those environment variables:
- - ``INMANTA_UID``: Will change, when starting the container, the id of the inmanta user.
- - ``INMANTA_GID``: Will change, when starting the container, the id of the inmanta group.
+    docker volume create mydockervolume
```
arnaudsjs, line 277: I would assume that docker-compose has support to create volumes itself, so that you don't have to execute this command. I know it has support to delete volumes, so then it would surprise me if it wouldn't be able to create them.

docs/install/2-install-server-with-docker.rst
```diff
@@ -305,42 +257,40 @@ There are two ways you can achieve this:
...
-        container_name: inmanta_orchestrator
-        ...
-        environment:
-            INMANTA_UID: 1000
-            INMANTA_GID: 1000
+    volumes:
+        - mydockervolume:/etc/inmanta/myfolder
+
+
+However if you really need to mount from the host, you can use bind mounts. You just need to make sure to change the ownership of
```
arnaudsjs, line 287:
```suggestion
However if you really need to mount a file or directory from the host, you can use bind mounts. You just need to make sure to change the ownership of
```

docs/install/2-install-server-with-docker.rst
```diff
@@ -305,42 +257,40 @@ There are two ways you can achieve this:
...
+the file/directory you want to mount to make sure it has same uid/gid as the inmanta user. To find them, in the container, you can use ``id`` command.
```
arnaudsjs, line 288:
```suggestion
the file/directory you want to mount to make sure it has same uid/gid as the inmanta user inside the container. To find them, in the container, you can use ``id`` command.
```

arnaudsjs (approved):
Looks good. Only minor formulation improvements.

docs/install/2-install-server-with-docker.rst
DariushMoh: Indeed I forgot about it. I will modify this

docs/install/2-install-server-with-docker.rst
```diff
@@ -305,42 +257,40 @@ There are two ways you can achieve this:
...
+                INMANTA_DATABASE_USERNAME: inmanta
+                INMANTA_DATABASE_PASSWORD: inmanta
             volumes:
                 - ./resources/com.inmanta.license:/etc/inmanta/license/com.inmanta.license
                 - ./resources/com.inmanta.jwe:/etc/inmanta/license/com.inmanta.jwe
                 - ./resources/my-server-conf.cfg:/etc/inmanta/inmanta.cfg
-                - ./resources/my-env-file:/etc/inmanta/env
 
+Mounting files/directories
+########################
```
Hugo-Inmanta, line 271: Sphinx might complain (not sure)
```suggestion
Mounting files/directories
##########################
```

Hugo-Inmanta (commented):
LGTM as far as I can see, one minor sphinx-related comment, approving already

Hugo-Inmanta (approved):

PR #8479, opened 2024-12-16
Active comments

wouterdb:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5480,7 +5480,7 @@ class Resource(BaseDocument):
         last_produced_events: Optional[datetime.datetime] = None,
         last_deployed_attribute_hash: Optional[str] = None,
         connection: Optional[asyncpg.connection.Connection] = None,
-        deployment_result: Optional[state.DeploymentResult] = None,
+        state: Optional[state.ResourceState] = None,
     ) -> None:
         """Update the data in the resource_persistent_state table"""
         args = ArgumentCollector(2)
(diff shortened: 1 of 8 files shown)


src/inmanta/data/__init__.py
```diff
@@ -5494,8 +5494,10 @@ async def update_persistent_state(
             "last_deployed_version": last_deployed_version,
         }
         query_parts = [f"{k}={args(v)}" for k, v in invalues.items() if v is not None]
-        if deployment_result:
-            query_parts.append(f"deployment_result={args(deployment_result.name)}")
+        if state:
+            query_parts.append(f"deployment_result={args(state.deployment_result.name)}")
+            # TODO: split blocked status field to make raceless
```
wouterdb, line 5499: There are comments the blocked field should only be updated under lock as both release and deplot path write it. 

We could get rid of that by splitting this field into 'blocked' and 'blocked_transient' booleans. These encode the 3 states without write conflict, then again, it may not be worth it, as the lock is available. .

tests/conftest.py
```diff
@@ -865,6 +865,17 @@ async def create(environment: uuid.UUID) -> Agent:
...
+    for agent in agents:
+        the_state = dict(agent.scheduler._state.resource_state)
+        for r,state in the_state.items():
+            print(r, state)
+        await agent.stop_working()
+        agent.scheduler._work.agent_queues._new_agent_notify = lambda x: x
+        await agent.start_working()
+
+        new_state = dict(agent.scheduler._state.resource_state)
+        assert the_state == new_state
```
wouterdb, line 877: magic test

PR #8481, opened 2024-12-17
Add test case for skip recovery deploys

### LINKED ISSUE #8452: Add test case for skip recovery deploys
#8292 / #8450 added per-resource deploy and repair timers, plus event propagation for recovery deploys to unblock previously skipped-for-dependencies resources. The normal timer behavior is tested in `tests/deploy/e2e/test_deploy_trigger.py` (`test_spontaneous_deploy` and `test_spontaneous_repair`), but there is no test yet for the recovery. This test should be added, i.e. a test where a resource skips because of a failed dependency, and it gets deployed shortly after a successful redeploy of the dependency, even if no timers are involved / the interval is set really high.

jptrindade:
# Description

Switched FAIL_DEPLOY for FAIL_COUNT in order to have a previously failing deploy succeed without creating resources and releasing a new verison.

closes #8452 

# Self Check:

- [x] Changelog entry


tests/deploy/test_scheduler_agent.py
```diff
@@ -107,7 +108,7 @@ async def execute(
         self.execute_count += 1
         result = (
             const.HandlerResourceState.failed
-            if resource_details.attributes.get(FAIL_DEPLOY, False) is True
+            if int(str(resource_details.attributes.get(FAIL_COUNT, 0))) >= self.execute_count
```
sanderr, line 111: Why the `str()` cast?

tests/deploy/test_scheduler_agent.py
```diff
@@ -76,7 +76,8 @@ class DummyExecutor(executor.Executor):
     A dummy executor:
         * It doesn't actually do any deploys, but instead keeps track of the actions
           (execute, dryrun, get_facts, etc.) that were requested on it.
-        * It reports a deploy as failed if the resource has an attribute with the value of the `FAIL_DEPLOY` variable.
+        * It reports a deploy as failed if the resource has an attribute with the value of the `FAIL_COUNT` variable with
```
sanderr, line 79: I like the concept, but I think it should be at the resource level rather than at the executor level. If that proves difficult I'd rather revert to the old behavior.

sanderr (changes requested):
Added two comments for something that stood out. I'm afraid the rest of the review will have to wait for other reviews.

tests/deploy/test_scheduler_agent.py
jptrindade: Pycharm was complaining (and I thought mypy would complain to) about casting an object to int. It turns out mypy did not care

tests/deploy/test_scheduler_agent.py
jptrindade: Reverted to the old behaviour and tried a workaround. What do you think?

tests/deploy/test_scheduler_agent.py
```diff
@@ -77,7 +77,7 @@ class DummyExecutor(executor.Executor):
         * It doesn't actually do any deploys, but instead keeps track of the actions
           (execute, dryrun, get_facts, etc.) that were requested on it.
         * It reports a deploy as failed if the resource has an attribute with the value of the `FAIL_DEPLOY` variable.
-          Otherwise, the deploy is reported as successful.
+
```
sanderr, line 80:
```suggestion
```

tests/deploy/test_scheduler_agent.py
```diff
@@ -977,8 +977,8 @@ def make_resources(
             )
             for rid, value, send_event, fail in [
                 (rid1, r1_value, r1_send_event, r1_fail),
-                (rid2, r2_value, r2_send_event, False),
-                (rid3, r3_value, False, False),
+                (rid2, r2_value, r2_send_event, 0),
+                (rid3, r3_value, False, 0),
```
sanderr, line 981: Is this a leftover from the counter attempt?

tests/deploy/test_scheduler_agent.py
```diff
@@ -1232,6 +1215,67 @@ def make_resources(
     assert agent.executor_manager.executors["agent2"].execute_count == 3
     assert agent.executor_manager.executors["agent3"].execute_count == 0
 
+    #########################
+    # Verify failure events #
+    #########################
+    agent.executor_manager.reset_executor_counters()
+    # make agent1's executor also managed to make r1 fail and succeed as we please
+    executor1: ManagedExecutor = ManagedExecutor()
```
sanderr, line 1223: I've been very careful to avoid this in the current implementation, because it could be very confusing if any scenarios are ever added after this one. I would consider how much effort it would be to make it a separate test case. If you decide against that, I'd like an explicit comment near the top of this one, where we make executor2 a managed one, stating that later on we'll do the same for this one.

sanderr: That or revert it at the end of this scenario (e.g. with a [monkeypatch context manager](https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.context)), whichever you prefer.

tests/deploy/test_scheduler_agent.py
```diff
@@ -1232,6 +1215,67 @@ def make_resources(
...
+    agent.executor_manager.register_managed_executor("agent1", executor1)
+
+    # release change to r1, and make its deploy fail
+    resources = make_resources(r1_value=2, r2_value=1, r3_value=0)
```
sanderr, line 1227: I'd keep the change restricted to what you actually want to change
```suggestion
    resources = make_resources(r1_value=6, r2_value=0, r3_value=0)
```
(previous "export" was 5, 0, 0).

I admit this incremental approach is not great. At some point I tried to make each of the scenarios more standalone like I did with the resetting of the counters, but it didn't work out.

tests/deploy/test_scheduler_agent.py
```diff
@@ -1232,6 +1215,67 @@ def make_resources(
...
+    await agent.scheduler._new_version(18, resources, make_requires(resources))
+    await retry_limited_fast(lambda: rid1 in executor1.deploys)
+    executor1.deploys[rid1].set_result(const.HandlerResourceState.failed)
+    # assert that r2 got skipped for its dependencies through event propagation
```
sanderr, line 1231:
```suggestion
    # assert that r2 got scheduled for deploy through event propagation
```
We don't assert that it skips, we *make* it skip. At this point, the part we care about is simply that it gets the event.

tests/deploy/test_scheduler_agent.py
```diff
@@ -1232,6 +1215,67 @@ def make_resources(
...
+    await retry_limited_fast(lambda: rid2 in executor2.deploys)
+    executor2.deploys[rid2].set_result(const.HandlerResourceState.skipped_for_dependency)
```
sanderr, line 1233: Here we should add a comment that we make it skip for dependencies to set up for verifying its recovery later on.

tests/deploy/test_scheduler_agent.py
```diff
@@ -1232,6 +1215,67 @@ def make_resources(
...
+    executor1: ManagedExecutor = ManagedExecutor()
+    agent.executor_manager.register_managed_executor("agent1", executor1)
+
+    # release change to r1, and make its deploy fail
+    resources = make_resources(r1_value=2, r2_value=1, r3_value=0)
+    await agent.scheduler._new_version(18, resources, make_requires(resources))
+    await retry_limited_fast(lambda: rid1 in executor1.deploys)
+    executor1.deploys[rid1].set_result(const.HandlerResourceState.failed)
+    # assert that r2 got skipped for its dependencies through event propagation
+    await retry_limited_fast(lambda: rid2 in executor2.deploys)
```
sanderr, line 1232: This one succeeds now, but I'm pretty sure it would even without the event propagation because you released a diff for r2. This is the reason for my comment about the diff higher up.

tests/deploy/test_scheduler_agent.py
```diff
@@ -1232,6 +1215,67 @@ def make_resources(
     assert agent.executor_manager.executors["agent2"].execute_count == 3
     assert agent.executor_manager.executors["agent3"].execute_count == 0
 
+    #########################
+    # Verify failure events #
+    #########################
```
sanderr, line 1220: I see why you moved this, but honestly I don't think you can contract these two scenarios.

The existing one verifies that we send failure events when event propagation is *enabled*.

The new one verifies that we send recovery events *even if event propagation is disabled*. I'd expect to start this one with the exact opposite assertion as for the first test, i.e. set up r1 to deploy with events disabled and assert that r2 does *not* get deployed. Then deploy it, recover r1, and verify that now r2 does get deployed for the recovery.

sanderr: I see that at the bottom of this file you've got a similar scenario. If you make sure that event propagation is disabled there, that could suffice. Do you see a reason to have a scenario in this test case as well?

tests/deploy/test_scheduler_agent.py
sanderr: Looks OK to me, with one caveat (see comment there). Though I wonder why you need it tbh. You release a new version anyway, so couldn't you simply set the `r1_fail` flag?

tests/deploy/e2e/test_deploy.py
```diff
@@ -1068,3 +1068,78 @@ async def test_skipped_for_dependency(resource_container, server, client, client
...
+
+    resources = [
+        {
+            "key": "key",
+            "value": "value",
+            "id": f"{rid1},v={version}",
+            "requires": [],
+            "purged": False,
+            "send_event": True,
+            "receive_events": False,
```
sanderr, line 1096: This is about a special event regardless of event settings, so I'd set it to the strictest mode: disable all events. Then we can verify that even then, we get the recovery.
```suggestion
            "send_event": False,
            "receive_events": False,
```

tests/deploy/e2e/test_deploy.py
```diff
@@ -1068,3 +1068,78 @@ async def test_skipped_for_dependency(resource_container, server, client, client
...
+        },
+        {
+            "key": "key2",
+            "value": "value",
+            "id": f"{rid2},v={version}",
+            "requires": [rid1],
+            "purged": False,
+            "send_event": True,
+            "receive_events": True,
```
sanderr, line 1105:
```suggestion
            "send_event": False,
            "receive_events": False,
```

tests/deploy/e2e/test_deploy.py
```diff
@@ -1068,3 +1068,78 @@ async def test_skipped_for_dependency(resource_container, server, client, client
...
+    assert scheduler._state.resource_state[rid2] == ResourceState(
+        status=ComplianceStatus.NON_COMPLIANT,
+        deployment_result=DeploymentResult.SKIPPED,
+        blocked=BlockedStatus.TRANSIENT,
+    )
+    assert scheduler._state.resource_state[rid1] == ResourceState(
+        status=ComplianceStatus.NON_COMPLIANT,
+        deployment_result=DeploymentResult.FAILED,
+        blocked=BlockedStatus.NO,
+    )
```
sanderr, line 1122: I'm not following with how you make this fail. And how you make it succeed later on.

tests/deploy/e2e/test_deploy.py
```diff
@@ -1068,3 +1068,78 @@ async def test_skipped_for_dependency(resource_container, server, client, client
...
+
+    # Trigger deploy without incrementing version
+    result = await client.deploy(environment)
```
sanderr, line 1125: This is a tough one. I think that with the current implementation of `deploy` this happens to assert what we want it to. But only as a result of an implementation detail, namely that deploy does not schedule skipped-for-dependencies resources.

What we're really looking to verify is that even if we trigger a deploy *only* for r1, does r2 still get deployed as a result? So I think we need to trigger this more fine-grained deploy. Alas, we don't (yet) expose it on the client. But you already have a scheduler object here, so I'd simply call its `deploy_resource` method.

sanderr (changes requested):

tests/deploy/test_scheduler_agent.py
```diff
@@ -1286,6 +1285,72 @@ def make_resources(
     assert agent.executor_manager.executors["agent3"].execute_count == 0
 
 
+async def test_skipped_for_dependencies_with_event_propagation(agent: TestAgent, make_resource_minimal):
+    """
+    Ensure that a resource that was skipped for its dependencies gets scheduled when they succeed
```
sanderr, line 1290:
```suggestion
    Ensure that a resource that was skipped for its dependencies gets scheduled when the dependency
    succeeds, even when normal event propagation is disabled.
```

tests/deploy/test_scheduler_agent.py
```diff
@@ -1286,6 +1285,72 @@ def make_resources(
...
+    executor2: ManagedExecutor = ManagedExecutor()
+    agent.executor_manager.register_managed_executor("agent2", executor2)
+
+    resources = {
+        rid1: make_resource_minimal(rid=rid1, values={"value": "r1_value"}, requires=[]),
+        rid2: make_resource_minimal(rid=rid2, values={"value": "r2_value"}, requires=[rid1]),
+    }
+    await agent.scheduler._new_version(version=1, resources=resources, requires=make_requires(resources))
+
+    # We set rid1 to fail so that rid2 has a failing dependency
```
sanderr, line 1309: You immediately set up the failure scenario, but you don't verify the assumptions of the start condition, i.e. that event propagation is disabled. I think you should first deploy both successfully, then trigger a deploy for rid1, asserting that rid2 does not get deployed (proving that no events are sent), and only then set up the failure scenario so that the final assertion becomes meaningful.

tests/deploy/e2e/test_deploy.py
sanderr: This is still unclear to me. What am I missing?

tests/deploy/test_scheduler_agent.py
```diff
@@ -1286,6 +1285,72 @@ def make_resources(
...
+    )
+
+    assert agent.scheduler._state.resource_state[rid2] == state.ResourceState(
+        # We are skipped, so not compliant
+        status=state.ComplianceStatus.NON_COMPLIANT,
+        deployment_result=state.DeploymentResult.SKIPPED,
+        blocked=state.BlockedStatus.TRANSIENT,
+    )
+
+    agent.executor_manager.reset_executor_counters()
```
sanderr, line 1331: First assert current counters and that nothing is queued anymore.

tests/deploy/test_scheduler_agent.py
```diff
@@ -1286,6 +1285,72 @@ def make_resources(
...
+    await agent.scheduler.deploy_resource(rid1, reason="Deploy rid1", priority=TaskPriority.USER_DEPLOY)
+
+    await retry_limited_fast(lambda: rid1 in executor1.deploys)
+    executor1.deploys[rid1].set_result(const.HandlerResourceState.deployed)
+
+    await retry_limited_fast(lambda: rid2 in executor2.deploys)
```
sanderr, line 1337: I'd add a comment here, because this is the main assertion of this test case
```suggestion
    # verify that rid2 gets scheduled due to the recovery event
    await retry_limited_fast(lambda: rid2 in executor2.deploys)
```

tests/deploy/test_scheduler_agent.py
```diff
@@ -2263,12 +2328,40 @@ async def test_state_of_skipped_resources_for_dependencies(agent: TestAgent, mak
...
+        blocked=state.BlockedStatus.NO,
+    )
+
     assert agent.scheduler._state.resource_state[rid2] == state.ResourceState(
         # We are skipped, so not compliant
         status=state.ComplianceStatus.NON_COMPLIANT,
         deployment_result=state.DeploymentResult.SKIPPED,
         blocked=state.BlockedStatus.TRANSIENT,
     )
+    # Release a new version where r1 succeeds and assert that r2 also gets deployed
```
sanderr, line 2343: Aren't the changes to this test case completely redundant? As in, covered by the new test case you added?

sanderr (changes requested):

sanderr (approved):

PR #8490, opened 2024-12-17
Refine error handling in dryrun method

### LINKED ISSUE #8206: Refine error in dryrun
https://github.com/inmanta/inmanta-core/blob/087ed07ceb68aba50f44cdbcd44759c121b8b3d3/src/inmanta/deploy/tasks.py#L195

jptrindade:
# Description

If something goes wrong with the dryrun, we expect the exception to be dealt with there, and this way, if something unexpected does happen, we will have more visibility on the error.

closes #8206 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/tasks.py
+++ b/src/inmanta/deploy/tasks.py
@@ -249,21 +249,13 @@ class DryRun(Task):
             my_executor: executor.Executor = await self.get_executor(
                 task_manager, agent, executor_resource_details.id.entity_type, self.version
             )
-
-            dryrun_result: executor.DryrunResult = await my_executor.dry_run(executor_resource_details, self.dry_run_id)
-            await task_manager.dryrun_update(
-                env=task_manager.environment,
-                dryrun_result=dryrun_result,
-            )
-
         except Exception:
-            # FIXME: seems weird to conclude undeployable state from generic Exception on either of two method calls
             logger_for_agent(agent).error(
                 "Skipping dryrun for resource %s because it is in undeployable state",
                 executor_resource_details.rvid,
                 exc_info=True,
             )
-            result = executor.DryrunResult(
+            dryrun_result = executor.DryrunResult(
                 rvid=executor_resource_details.rvid,
                 dryrun_id=self.dry_run_id,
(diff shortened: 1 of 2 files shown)


src/inmanta/deploy/tasks.py
```diff
@@ -249,29 +249,24 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
...
                 rvid=executor_resource_details.rvid,
                 dryrun_id=self.dry_run_id,
                 changes={"handler": AttributeStateChange(current="FAILED", desired="Resource is in an undeployable state")},
                 started=started,
                 finished=datetime.datetime.now().astimezone(),
                 messages=[],
             )
-            await task_manager.dryrun_update(env=task_manager.environment, dryrun_result=result)
+        else:
+            dryrun_result = await my_executor.dry_run(executor_resource_details, self.dry_run_id)
```
wouterdb, line 267: I think this code needs a try except as well, to make sure all exceptions are handled

wouterdb (changes requested):

src/inmanta/deploy/tasks.py
jptrindade: Added try except here and changed the error description for the one above. What do you think? I did not want to have the same error log in both cases

src/inmanta/deploy/tasks.py
```diff
@@ -249,29 +249,42 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
...
                 exc_info=True,
             )
-            result = executor.DryrunResult(
+            dryrun_result = executor.DryrunResult(
                 rvid=executor_resource_details.rvid,
                 dryrun_id=self.dry_run_id,
-                changes={"handler": AttributeStateChange(current="FAILED", desired="Resource is in an undeployable state")},
+                changes={
+                    "handler": AttributeStateChange(
+                        current="FAILED", desired="Unable to generate an appropriate executor for this resource"
```
wouterdb, line 263:
```suggestion
                        current="FAILED", desired="Unable to construct an executor for this resource"
```

wouterdb (approved):

Hugo-Inmanta (approved):

PR #8499, opened 2024-12-19
Issue/type untangling 2

wouterdb:
# Description

Make type tangle less bad:

-   move resource id string type to types module
-  make resources.ID less prone to loops by making resources depend on less
   -  make inmanta.ast contain all exceptions of the compiler
   -  make inmanta.ast import very little else.

Successor of #8466 

# Self Check:

- [x] Changelog entry
- [X] Type annotations are present
- [X] Correct, in line with design


src/inmanta/execute/proxy.py
```diff
@@ -18,9 +18,10 @@
 
 from collections.abc import Iterable, Mapping, Sequence
 from copy import copy
-from typing import Callable, Optional, Union
+from typing import Callable, Union
 
-from inmanta.ast import NotFoundException, RuntimeException
+# Keep UnsetException, UnknownException and AttributeNotFound in place for backward compat with <iso8
+from inmanta.ast import AttributeNotFound, NotFoundException, RuntimeException, UnknownException
```
Hugo-Inmanta, line 24: Probably need a noqa in here as well.

src/inmanta/data/__init__.py
```diff
@@ -46,25 +46,19 @@
 from asyncpg.protocol import Record
 
 import inmanta.db.versions
+import inmanta.protocol
+import inmanta.types
```
Hugo-Inmanta, line 50: 1. Why do we need protocol in here ?
2. I think we can remove `import inmanta.types` if we import `ResourceType` from `inmanta.types` + change where we use it (only 1 place I think)

tests/server/test_incremental_deploy.py
```diff
@@ -27,16 +27,17 @@
 
 import pytest
 
+import inmanta.types
```
Hugo-Inmanta, line 30: I think this isn't required
```suggestion

```

Hugo-Inmanta (changes requested):

Hugo-Inmanta:
Just realized inmanta-lsm will need addressing as well at some point, but should work just fine because we kept things backwards compatible.

src/inmanta/execute/proxy.py
wouterdb: they are actually used, if it passed, it is fine

sanderr (approved):

src/inmanta/agent/forking_executor.py
```diff
@@ -97,6 +97,7 @@
 from inmanta.agent import executor, resourcepool
 from inmanta.agent.executor import DeployResult, FactResult
 from inmanta.agent.resourcepool import PoolManager, PoolMember
+from inmanta.const import LOGGER_NAME_EXECUTOR
```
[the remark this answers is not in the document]
wouterdb: Very much required,.... (i.e. not a change if you look at all commits)

src/inmanta/agent/in_process_executor.py
```diff
@@ -30,7 +30,7 @@
 from inmanta.agent import executor, handler
 from inmanta.agent.executor import DeployResult, DryrunResult, FactResult, FailedResources, ResourceDetails
 from inmanta.agent.handler import HandlerAPI, SkipResource, SkipResourceForDependencies
-from inmanta.const import ParameterSource
+from inmanta.const import NAME_RESOURCE_ACTION_LOGGER, ParameterSource
```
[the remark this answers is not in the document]
wouterdb: Very much required,.... (i.e. not a change if you look at all commits)
 (merge from master)

src/inmanta/data/__init__.py
```diff
@@ -60,7 +60,14 @@
 )
 from inmanta.data import model as m
 from inmanta.data import schema
-from inmanta.data.model import AuthMethod, BaseModel, PagingBoundaries, PipConfig, api_boundary_datetime_normalizer
+from inmanta.data.model import (
+    AuthMethod,
+    BaseModel,
+    PagingBoundaries,
+    PipConfig,
+    ResourceType,
```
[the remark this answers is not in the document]
wouterdb: doh!

Hugo-Inmanta (approved):

Issue #8485, opened 2024-12-17
logging: Some methods in the stable api of the logging are deprecated


wouterdb:
drop them

PR #8509, opened 2024-12-20
drop deprecated methods

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/8485

# Self Check:

- [x] Changelog entry


changelogs/unreleased/8485-drop-deprecated-logging-api-methods.yml
```diff
@@ -0,0 +1,10 @@
+description: >
+  Drop deprecated methods: `set_log_level`, `set_log_formatter`, `set_logfile_location` and `get_handler`
+  from the stable api of the logging.
```
Hugo-Inmanta, line 3: Not sure about the wording and relevance for end-user

wouterdb: I would also mentions the class we removed it from:

```suggestion
  Dropped deprecated methods `set_log_level`, `set_log_formatter`, `set_logfile_location` and `get_handler`
  from `inmanta.logging.InmantaLoggerConfig`. Use the `apply_options` method or a log config file instead. 
```

src/inmanta/logging.py
```diff
@@ -667,61 +667,6 @@ def _apply_logging_config(self, logging_config: FullLoggingConfig) -> abc.Sequen
...
-        [DEPRECATED] Set the logging level. A handler should have been created before.
-        The possible inmanta log levels and their associated python log level
-        are defined in the inmanta.logging.log_levels dictionary.
-
-        :param inmanta_log_level: The inmanta logging level
-        :param cli: True if the logs will be outputted to the CLI.
-        """
-        # ToDO: remove?
-        python_log_level = convert_inmanta_log_level(inmanta_log_level, cli)
-        for handler in self._handlers:
```
wouterdb, line 682: Could we also lose the _handler field?

Hugo-Inmanta: As far as I can see, it is still used in the `InmantaLoggerConfig.get_instance` method, but I'm not sure if it is still relevant there or not.

wouterdb (approved):

PR #8512, opened 2024-12-20
Added the validate-logging-config command

### LINKED ISSUE #8482: Logging: add a way to validate config files for specific tools (like scheduler)
Add a tool to load a config file exactly like e.g. the scheduler would and print some standard log lines that are expected.
(e.g. resource action log, callback log, access log, tornado client pool, ....)

The difficulty is that the server starts the scheduler, so it take no user controllable arguments, while server and compiler do.

arnaudsjs:
# Description

Added the `validate-logging-config` command.

closes #8482 

# Self Check:

- [x] Changelog entry


src/inmanta/app.py
```diff
@@ -620,6 +621,74 @@ def export(options: argparse.Namespace) -> None:
...
+def validate_logging_config_parser_config(
+    parser: argparse.ArgumentParser, parent_parsers: abc.Sequence[ArgumentParser]
+) -> None:
+    """
+    Config parser for the validate-logging-config command.
+    """
+    parser.add_argument(
+        "filename",
+        help="Path to the logging configuration file.",
+    )
```
wouterdb, line 633: shouldn't we use the `--logging-config`

arnaudsjs: Yes, we could. But then we need to document clearly how the command gets its input, because `--help` will otherwise not show any useful information.

src/inmanta/app.py
```diff
@@ -620,6 +621,74 @@ def export(options: argparse.Namespace) -> None:
...
+    parser.add_argument(
+        "-e",
+        dest="environment",
+        help="The environment id to be used as context variable in logging config templates.",
+    )
```
wouterdb, line 638: I wonder if we should default to random uuid?

src/inmanta/app.py
```diff
@@ -620,6 +621,74 @@ def export(options: argparse.Namespace) -> None:
     summary_reporter.print_summary_and_exit(show_stack_traces=options.errors)
 
 
+def validate_logging_config_parser_config(
```
wouterdb, line 624: I wonder if we should define subcommands for all components, so you can use this exactly like the process itself....

wouterdb (changes requested):

src/inmanta/app.py
```diff
@@ -965,18 +1045,24 @@ def app() -> None:
...
-    if env:
-        log_context[LOG_CONTEXT_VAR_ENVIRONMENT] = env
+    if hasattr(options, "func") and options.func == validate_logging_config:
+        # By-pass all other logging setup code when validating logging config.
+        try:
+            options.func(options)
+        except Exception as e:
+            print(str(e), file=sys.stderr)
+            sys.exit(1)
+        return
```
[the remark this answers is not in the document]
arnaudsjs: I thought you told me explicitly to do that. But maybe I misunderstood.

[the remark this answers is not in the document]
wouterdb: Might be, but even if I came up with the idea: I don't like it.

jptrindade (approved):

src/inmanta/app.py
```diff
@@ -620,6 +620,74 @@ def export(options: argparse.Namespace) -> None:
...
+) -> None:
+    """
+    Config parser for the validate-logging-config command.
+    """
+    parser.add_argument(
+        "-e",
+        dest="environment",
+        help="The environment id to be used as context variable in logging config templates. If not specified,"
+        " 0c111d30-feaf-4f5b-b2d6-83d589480a4a will be used.",
+        default="0c111d30-feaf-4f5b-b2d6-83d589480a4a",
```
sanderr, line 634: This feels very arbitrary. Can't we use something like "<env-id>", or does it need to be a valid uuid?

src/inmanta/app.py
```diff
@@ -620,6 +620,74 @@ def export(options: argparse.Namespace) -> None:
...
+    if logging_config.loaded_config_file is None:
+        raise Exception("No logging configuration file found.")
+    print(f"Using logging config file: {logging_config.loaded_config_file}", file=sys.stderr)
+    env_id = options.environment
+    logger_and_message = [
+        (logging.getLogger("inmanta.protocol.rest.server"), "Log line from Inmanta server"),
+        (logging.getLogger("inmanta.server.services.compilerservice"), "Log line from compiler service"),
+        (logging.getLogger(const.NAME_RESOURCE_ACTION_LOGGER).getChild(env_id), "Log line for resource action log"),
+        (logging.getLogger("inmanta_lsm.callback"), "Log line from callback"),
+        (logging.getLogger("inmanta.scheduler"), "Log line from the scheduler"),
```
sanderr, line 664:
```suggestion
        (logging.getLogger("inmanta.scheduler"), "Log line from the resource scheduler"),
```

src/inmanta/app.py
```diff
@@ -620,6 +620,74 @@ def export(options: argparse.Namespace) -> None:
...
+@command(
+    "validate-logging-config",
+    help_msg="This command loads a logging config file, taking into account the precedence rules for logging config files,"
+    " and produces log lines. It serves as a tool to validate whether a logging config file is syntactically correct"
+    " and behaves as expected. Optionally, a sub-command can be specified to indicate the component for which the"
+    " logging config file should be loaded.",
+    parser_config=validate_logging_config_parser_config,
+)
+def validate_logging_config(options: argparse.Namespace) -> None:
+    logging_config = InmantaLoggerConfig.get_current_instance()
```
sanderr, line 654: Doesn't this raise an exception if the config isn't loaded yet?

sanderr (changes requested):
LGTM overall, a few minor questions / comments

src/inmanta/app.py
arnaudsjs: This is based on a suggestion that Wouter made (see comment above). I think it's a good to have a uuid by default, because like this the name of the logger will be as it is in reality. It's indeed a very arbitrary default value, but it has two main advantages:

* The user doesn't need to specify a uuid him-/herself.
* The uuid is stable.

src/inmanta/app.py
arnaudsjs: Yes. But the config is loaded in the `app.app()` function before it calls into this method.

PR #8514, opened 2024-12-20
bugfix in event propagation

sanderr:
# Description

Bugfix in event propagation. We had a special case for sending events where we set `deploying=set()` in order to force the receiving resource to be scheduled, even if it happens to already be running. But this also hid any other resources that are running. In other words, if another dependency of the receiving resource happens to be running, it becomes invisible and the receiving resource will not wait for it. It will then see that resource in the available state and skip.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -1233,9 +1233,9 @@ class ResourceScheduler(TaskManager):
                     else f"Deploying because an event was received from {resource_id}"
                 ),
                 priority=priority,
-                # report no ongoing deploys because ongoing deploys can not capture the event. We desire new deploys
-                # to be scheduled, even if any are ongoing for the same intent.
-                deploying=set(),
+                # report no ongoing deploys for the given resources because ongoing deploys can not capture the event.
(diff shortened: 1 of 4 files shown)


wouterdb (approved):

src/inmanta/deploy/scheduler.py
```diff
@@ -1233,9 +1233,9 @@ def _send_events(
                     else f"Deploying because an event was received from {resource_id}"
                 ),
                 priority=priority,
-                # report no ongoing deploys because ongoing deploys can not capture the event. We desire new deploys
-                # to be scheduled, even if any are ongoing for the same intent.
-                deploying=set(),
+                # report no ongoing deploys for the given resources because ongoing deploys can not capture the event.
+                # We desire new deploys to be scheduled, even if any are ongoing for the same intent.
+                deploying=self._deploying_latest,
```
Hugo-Inmanta, line 1236: Shouldn't we set force_deploy here ?
```suggestion
                deploying=self._deploying_latest,
                force_deploy = True,
```

Hugo-Inmanta (approved):

PR #8530, opened 2024-12-24
changed default for InmantaBootloader's logging config argument

sanderr:
# Description

This new parameter was breaking all extension pipelines because they call the bootloader without arguments. The current implementation then breaks later on when `get_current_instance()` is called.

I'm not sure if this is the proper solution, but it's the cheapest to get the builds unstuck until it can be properly considered after the holidays.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/server/bootloader.py
+++ b/src/inmanta/server/bootloader.py
@@ -82,10 +82,11 @@ class InmantaBootloader:
     # Cache field for available extensions
     AVAILABLE_EXTENSIONS: Optional[dict[str, str]] = None
 
-    def __init__(self, configure_logging: bool = False) -> None:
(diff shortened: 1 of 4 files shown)


src/inmanta/server/bootloader.py
```diff
@@ -82,10 +82,11 @@ class InmantaBootloader:
     # Cache field for available extensions
     AVAILABLE_EXTENSIONS: Optional[dict[str, str]] = None
 
-    def __init__(self, configure_logging: bool = False) -> None:
+    def __init__(self, *, configure_logging: bool = True) -> None:
```
sanderr, line 85: This is the main change

PR #8540, opened 2024-12-31
Logging force tty when generating default config

### LINKED ISSUE #8484: logging Force TTY to be abled when generating default config
When generating default logging config, piping to file or not produces different output, because the logger is tty aware

Make the tool force TTY config

jptrindade:
# Description

Added the option to `force_tty` on `_get_multiline_formatter_config` so that we output the same config when printing on the console or writing the config to a file

closes #8484 

# Self Check:

- [x] Changelog entry


jptrindade:
@sanderr I added you as a reviewer just in case you were curious about the test case.

src/inmanta/app.py
```diff
@@ -964,7 +964,17 @@ def default_logging_config(options: argparse.Namespace) -> None:
...
+    original_force_tty = os.environ.get(const.ENVIRON_FORCE_TTY, None)
+    os.environ[const.ENVIRON_FORCE_TTY] = original_force_tty if original_force_tty else "yes"
+    try:
+        component_config.apply_options(options, options.cmd, context)
+    except Exception as e:
+        raise e
+    finally:
+        # Revert this env var back to its original state
+        if original_force_tty is None:
+            del os.environ[const.ENVIRON_FORCE_TTY]
```
arnaudsjs, line 977: * The except clause is redundant.
* If `original_force_tty` is set to an empty string, we will overwrite the value with "yes". This makes the comment "Revert this env var back to its original state" invalid.

```suggestion
    # Force TTY so that this command outputs the same config when piping to a file
    original_force_tty = os.environ.get(const.ENVIRON_FORCE_TTY, None)
    if original_force_tty is None:
        os.environ[const.ENVIRON_FORCE_TTY] = "yes"
    try:
        component_config.apply_options(options, options.cmd, context)
    finally:
        # Revert this env var back to its original state
        if original_force_tty is None:
            del os.environ[const.ENVIRON_FORCE_TTY]
```

src/inmanta/app.py
```diff
@@ -964,7 +964,17 @@ def default_logging_config(options: argparse.Namespace) -> None:
     context = {var: f"{place_holder}{var}{place_holder}" for var in context_vars}
 
     component_config = InmantaLoggerConfig(stream=sys.stdout, no_install=True)
-    component_config.apply_options(options, options.cmd, context)
+    # Force TTY so that this command outputs the same config when piping to a file
```
arnaudsjs, line 967: The call to `start_loggers_for_extensions()` also extends the logging config, so we need to make sure that the environment variable is also set for that call. Why not set is for this entire method?

tests/test_logging.py
```diff
@@ -409,3 +410,48 @@ def test_server_documentation_conformance(inmanta_config, monkeypatch):
...
+
+        tty_config_stdout = stdout.decode("utf-8")
+        # Assert that TTY was present
+        assert "no_color: false" in tty_config_stdout
+        assert "reset: true" in tty_config_stdout
+        assert "log_colors: null" not in tty_config_stdout
+
+        # Output the logging config on the CLI with TTY unset
+        # This is the same as piping to a file
+        process = await subprocess.create_subprocess_exec(*args, stdout=subprocess.PIPE)
```
arnaudsjs, line 445:
```suggestion
        # This is the same as piping to a file
        assert "ENVIRON_FORCE_TTY" not in os.environ
        process = await subprocess.create_subprocess_exec(*args, stdout=subprocess.PIPE)
```

tests/test_logging.py
```diff
@@ -409,3 +410,48 @@ def test_server_documentation_conformance(inmanta_config, monkeypatch):
...
+    components = ["scheduler", "server", "compiler"]
+    for component in components:
+        args = [sys.executable, "-m", "inmanta.app", "print-default-logging-config", component]
+
+        # Output the logging config on the CLI.
+        # Here we force ENVIRON_FORCE_TTY to be set to simulate that we are on a TTY
+        process = await subprocess.create_subprocess_exec(
+            *args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env={ENVIRON_FORCE_TTY: "yes"}
+        )
+        await process.wait()
```
arnaudsjs, line 428: As the documentation mentions, `process.wait()` can deadlock when `subprocess.PIPE` is used. I don't really see why this call is required, given that we already call into `process.communicate()`.

tests/test_logging.py
```diff
@@ -409,3 +410,48 @@ def test_server_documentation_conformance(inmanta_config, monkeypatch):
...
+    Test that piping to file does not change the logging config
+    """
+    components = ["scheduler", "server", "compiler"]
+    for component in components:
+        args = [sys.executable, "-m", "inmanta.app", "print-default-logging-config", component]
+
+        # Output the logging config on the CLI.
+        # Here we force ENVIRON_FORCE_TTY to be set to simulate that we are on a TTY
+        process = await subprocess.create_subprocess_exec(
+            *args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env={ENVIRON_FORCE_TTY: "yes"}
```
arnaudsjs, line 426:
```suggestion
        process = await subprocess.create_subprocess_exec(
            *args, stdout=subprocess.PIPE, env={ENVIRON_FORCE_TTY: "yes"}
```

arnaudsjs (changes requested):

src/inmanta/app.py
jptrindade: I'm not sure it uses the `_is_on_tty` which, I believe, is the only real relevant reason why we set the environment variable. I set the environment variable to the whole method anyway, just to avoid some errors in the future if we end up using the `_is_on_tty` in other parts of this code

arnaudsjs: I know. My main concern here was to make it future proof.

arnaudsjs (approved):

wouterdb (approved):

PR #8547, opened 2025-01-02
Issue/8511 better warning message when handler skips

### LINKED ISSUE #8511: better warning message when handler skips for dependency in available state
Motivation: https://inmanta.slack.com/archives/CKRF0C8R3/p1734695975619039

1. is WARNING an appropriate log level for this skip, no matter the dependency state? Probably not
2. When a dependency is in an available state, we should clarify that we skip because the dependency got a new desired state while we were preparing for deploy, and that we will retry when it gets deployed successfully. Consider other states: which are expected and therefore clear to the end user when reported, which require a more tailored message (or are never expected at all)?

jptrindade:
# Description

Refined warning message so that no warning is shown when a dependency gets into a deploying/available state.

closes #8511

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/handler.py
+++ b/src/inmanta/agent/handler.py
@@ -552,7 +552,7 @@ class HandlerAPI(ABC, Generic[TResource]):
             return result.result["data"]
 
         def filter_resources_by_state(
-            reqs: abc.Mapping[ResourceIdStr, ResourceState], states: typing.Set[const.ResourceState]
+            reqs: abc.Mapping[ResourceIdStr, ResourceState], states: typing.Set[ResourceState]
         ) -> abc.Mapping[ResourceIdStr, ResourceState]:
             """
             Return a sub-dictionary of dependencies of this resource.
@@ -591,7 +591,7 @@ class HandlerAPI(ABC, Generic[TResource]):
             ctx.info(
                 "Resource %(resource)s skipped because some dependencies %(reqs)s "
                 "got a new desired state while we were preparing to deploy."
-                "We will retry when all dependencies get deployed successfully",
+                " We will retry when all dependencies get deployed successfully",
                 resource=resource.id.resource_version_str(),
                 reqs=str({rid for rid in dependencies_waiting_to_be_deployed.keys()}),
             )
--- a/src/inmanta/agent/handler.py
+++ b/src/inmanta/agent/handler.py
@@ -551,28 +551,49 @@ class HandlerAPI(ABC, Generic[TResource]):
                 raise Exception(f"Failed to determine whether resource should reload{error_msg_from_server}")
             return result.result["data"]
 
-        def filter_resources_in_unexpected_state(
-            reqs: abc.Mapping[ResourceIdStr, ResourceState]
+        def filter_resources_by_state(
+            reqs: abc.Mapping[ResourceIdStr, ResourceState], states: typing.Set[const.ResourceState]
         ) -> abc.Mapping[ResourceIdStr, ResourceState]:
             """
-            Return a sub-dictionary of reqs with only those resources that are in an unexpected state.
+            Return a sub-dictionary of dependencies of this resource.
+            Only keeping dependencies that are in a state that is in the provided set.
+
+            :param reqs: The list of requirements of this resource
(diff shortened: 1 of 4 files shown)


src/inmanta/agent/handler.py
```diff
@@ -551,28 +551,49 @@ def _should_reload() -> bool:
...
                 const.ResourceState.dry,
                 const.ResourceState.undefined,
                 const.ResourceState.skipped_for_undefined,
-                const.ResourceState.deploying,
-            }
-            return {rid: state for rid, state in reqs.items() if state in unexpected_states}
-
-        resources_in_unexpected_state = filter_resources_in_unexpected_state(requires)
-        if resources_in_unexpected_state:
+            },
```
jptrindade, line 574: Are the other states, by definition, unexpected? 
i.e. unavailable / skipped / failed / cancelled ?

jptrindade:
Is this the only distinction between unexpected states that we want?

src/inmanta/agent/handler.py
```diff
@@ -551,28 +551,49 @@ def _should_reload() -> bool:
...
+        # Check if any dependencies got a new desired state while this resource was waiting to deploy
+        dependencies_waiting_to_be_deployed = filter_resources_by_state(
+            requires, {const.ResourceState.available, const.ResourceState.deploying}
+        )
+        if dependencies_waiting_to_be_deployed:
+            ctx.set_resource_state(const.HandlerResourceState.skipped_for_dependency)
+            ctx.debug(
+                "Resource %(resource)s skipped because some dependencies %(reqs)s "
+                "got a new desired state while we were preparing to deploy."
+                "We will retry when all dependencies get deployed successfully",
```
arnaudsjs, line 594:
```suggestion
                " We will retry when all dependencies get deployed successfully",
```

src/inmanta/agent/handler.py
```diff
@@ -551,28 +551,49 @@ def _should_reload() -> bool:
                 raise Exception(f"Failed to determine whether resource should reload{error_msg_from_server}")
             return result.result["data"]
 
-        def filter_resources_in_unexpected_state(
-            reqs: abc.Mapping[ResourceIdStr, ResourceState]
+        def filter_resources_by_state(
+            reqs: abc.Mapping[ResourceIdStr, ResourceState], states: typing.Set[const.ResourceState]
```
arnaudsjs, line 555: Let's make the typing consistent with the `reqs` parameter.
```suggestion
            reqs: abc.Mapping[ResourceIdStr, ResourceState], states: typing.Set[ResourceState]
```

src/inmanta/agent/handler.py
```diff
@@ -551,28 +551,49 @@ def _should_reload() -> bool:
...
+            )
+            return
+
+        # Check if any dependencies got a new desired state while this resource was waiting to deploy
+        dependencies_waiting_to_be_deployed = filter_resources_by_state(
+            requires, {const.ResourceState.available, const.ResourceState.deploying}
+        )
+        if dependencies_waiting_to_be_deployed:
+            ctx.set_resource_state(const.HandlerResourceState.skipped_for_dependency)
+            ctx.debug(
```
arnaudsjs, line 591: Why debug? Wouldn't it be better to use `info()` here?

arnaudsjs (changes requested):

src/inmanta/agent/handler.py
jptrindade: I was on the fence between these 2 log types. I saw that we indicate the start of the run/deploy in `DEBUG`  ( Based on this screenshot from Guillaume) so I figured it should be debug instead of info. How do you differentiate between info and debug?
image

arnaudsjs (approved):

src/inmanta/agent/handler.py
arnaudsjs: These states should be processed by the handler, because they are deploy-time failures states. The lsm transfer resource will inspect the states of its dependencies to determine whether it should move the service into a success or a failed state.

wouterdb (approved):

Issue #8513, opened 2024-12-20
logging: make compiler load in the log config as specified in the server config


wouterdb:
Currently, we can configure the log config file in the config file of the server, but nothing forces the compiler to load the config file of the server (except when it is in `/etc/inmanta`) . 

If the server starts the compiler, and a log config file is set for the compiler in the server config, it should pass the options along to the compiler on the cli

Should be added to the docs, when discussing the `logging.compiler` option, that it affect the output of all compiles done by the server and can be used to finetune the output as seen in the dashboard.

PR #8558, opened 2025-01-06
Issue/8513 make compiler load log config specified in server config

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/8513

# Self Check:

- [x] Changelog entry


src/inmanta/server/services/compilerservice.py
```diff
@@ -448,8 +449,15 @@ async def setup() -> AsyncIterator[Awaitable[Optional[data.Report]]]:
...
             server_port = opt.server_bind_port.get()
-            cmd = [
-                "-vvv",
+
+            app_cli_args = ["-vvv"]
+            # Pass down the logging config to the compiler
+            compiler_log_config_file: str | None = compiler_log_config.get()
+            if compiler_log_config_file:
+                app_cli_args.append("--logging-config")
+                app_cli_args.append(os.path.abspath(compiler_log_config_file))
```
wouterdb, line 458: I wonder if it would not be better to pass along the config file and folder of the server in full? 

i.e. the config folder defaults to `/etc/inmanta/inmanta.d` so if we don't pass it along, we will read `/etc/inmanta/inmanta.d` wether the server uses it or not.

tests/test_logging.py
```diff
@@ -611,3 +616,75 @@ async def test_print_default_logging_cmd(inmanta_config, tmp_path):
...
+    report = await client.get_report(uuid.UUID(compile_id))
+    assert report.code == 200
+
+    # Get the compile outstream
+    for report in report.result["report"]["reports"]:
+        if report["name"] == "Recompiling configuration model":
+            compile_outstream = report["outstream"]
+            break
+
+    assert "COMPILER_CONFIG_FLAG -- Starting compile" in compile_outstream
```
arnaudsjs, line 690: A bit more elegant
```suggestion
    for report in report.result["report"]["reports"]:
        if report["name"] == "Recompiling configuration model":
            compile_outstream = report["outstream"]
            assert "COMPILER_CONFIG_FLAG -- Starting compile" in compile_outstream
            break
    else:
        assert False, "Compile report doesn't contain a 'Recompiling configuration model' section."
```

arnaudsjs (changes requested):

src/inmanta/server/services/compilerservice.py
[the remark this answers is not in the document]
Hugo-Inmanta: I think if these are set, they should already get passed down [here](https://github.com/inmanta/inmanta-core/blob/67df4d0b541f5c4cc3a6a789586f3d32ad55e661/src/inmanta/server/services/compilerservice.py#L521), or do you mean something else ?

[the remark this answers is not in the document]
arnaudsjs: The problem is that the `--logging-config` takes precedence over the environment variables. So the environment variables will be ignored even if you pass them along, while the `--logging-config` option is set.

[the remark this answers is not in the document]
arnaudsjs: The `_get_logging_config_source()` method in the `InmantaLoggerConfig` class contains the logic about the precedence rules between the different way to provide logging config.

tests/test_logging.py
```diff
@@ -611,3 +616,86 @@ async def test_print_default_logging_cmd(inmanta_config, tmp_path):
...
+        fh.write(compiler_logging_config)
+    config = os.path.join(tmpdir, "logging_config.yml")
+    with open(config, "w") as fh:
+        fh.write(
+            f"""
+[logging]
+compiler = {os.path.abspath(compiler_logging_config_file)}
+        """
+        )
+    Config.load_config(min_c_config_file=config)
```
Hugo-Inmanta, line 654: Bit hackish to load the config twice ? Once here and once in the `inmanta_config` fixture ? Is there a better alternative ?

src/inmanta/config.py
```diff
@@ -95,6 +96,8 @@ def load_config(
         files: list[str]
         if min_c_config_file is not None:
             files = [main_cfg_file] + cfg_files_in_config_dir + local_dot_inmanta_cfg_files + [min_c_config_file]
+            cls._min_c_config_file = min_c_config_file
```
arnaudsjs, line 99: We must reset this value in the `reset()` method.

arnaudsjs: We should also set this value explicitly to `None` in the else-block. For the situation where `reset()` wasn't called first.

tests/test_logging.py
arnaudsjs: I think this question is no longer relevant, because with the above-mentioned issues resolved, I don't think the test case will still work. The config will be overridden by the start of the server.

arnaudsjs (changes requested):

tests/test_logging.py
arnaudsjs: Maybe you could overwrite the `server_config` fixture to make it load the right config.

Hugo-Inmanta: Overwrote the `inmanta_config` fixture.

Hugo-Inmanta: I had to move the test to its own file, I couldn't get it to play nice with the other tests calling into the `inmanta_config` fixture.

arnaudsjs (approved):

PR #8559, opened 2025-01-06
Native python types

wouterdb:
Basic native python types

# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/ast/type.py
+++ b/src/inmanta/ast/type.py
@@ -371,6 +371,9 @@ class String(Primitive):
         return None
 
 
+    def __eq__(self, other):
+        return type(self) == type(other)
+
 @stable_api
 class List(Type):
     """
@@ -543,6 +546,11 @@ class TypedDict(Dict):
 
     def get_location(self) -> None:
         return None
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, TypedDict):
+            return NotImplemented
+        return self.element_type == other.element_type
 
 
 @stable_api
--- a/src/inmanta/plugins.py
+++ b/src/inmanta/plugins.py
@@ -19,13 +19,16 @@
 import asyncio
 import collections.abc
 import inspect
+import numbers
 import os
 import subprocess
+import typing
 import warnings
 from collections import abc
-from typing import TYPE_CHECKING, Any, Callable, Literal, Mapping, Optional, Sequence, Type, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Literal, Mapping, Optional, Sequence, Type, TypeVar, Dict
 
 import inmanta.ast.type as inmanta_type
+import typing_inspect
 from inmanta import const, protocol, util
 from inmanta.ast import (  # noqa: F401 Plugin exception is part of the stable api
     LocatableString,
@@ -35,7 +38,7 @@ from inmanta.ast import (  # noqa: F401 Plugin exception is part of the stable a
     Range,
     RuntimeException,
     TypeNotFoundException,
-    WithComment,
+    WithComment, TypingException,
 )
 from inmanta.ast.type import NamedType
 from inmanta.config import Config
@@ -210,6 +213,9 @@ class Null(inmanta_type.Type):
     def type_string_internal(self) -> str:
         return self.type_string()
 
+    def __eq__(self, other):
+        return type(self) == type(other)
+
 
 # Define some types which are used in the context of plugins.
 PLUGIN_TYPES = {
@@ -218,6 +224,78 @@ PLUGIN_TYPES = {
     "null": Null(),  # Only NoneValue will pass validation
     None: Null(),  # Only NoneValue will pass validation
 }
+
+python_to_model = {
+    str: inmanta_type.String(),
+    float: inmanta_type.Float(),
+    numbers.Number: inmanta_type.Number(),
+    int: inmanta_type.Integer(),
+    bool: inmanta_type.Bool(),
+    dict: inmanta_type.LiteralDict(),
+    list: inmanta_type.LiteralList(),
+}
+
+
+def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
+    """
+    :param python_type: The evaluated python type as provided in the Python type annotation.
+    """
+    # Any to any
+    if python_type is typing.Any:
+        return inmanta_type.AnyType()
+
+    if python_type is type(None):
+        return Null()
+
+    if typing_inspect.is_union_type(python_type):
+        if any(typing_inspect.is_optional_type(tt) for tt in typing.get_args(python_type)):
+            # Optional type
+            other_types = [tt for tt in typing.get_args(python_type) if not typing_inspect.is_optional_type(tt)]
+            if len(other_types) == 0:
+                # Probably not possible
+                return Null()
+            if len(other_types) == 1:
+                return inmanta_type.NullableType(to_dsl_type(other_types[0]))
+            # TODO: optional unions
+            return inmanta_type.AnyType()
+        else:
+            return inmanta_type.AnyType()
+
+    # TODO: unions
+    # bases: Sequence[inmanta.ast.type.Type] = [to_dsl_type(arg) for arg in typing.get_args(python_type)]
+    # return inmanta.ast.type.Union(bases)
+
+    if typing_inspect.is_generic_type(python_type):
+        origin = typing.get_origin(python_type)
+        if  origin is Sequence:
+            # Can this fail?
+            base: inmanta.ast.type.Type = typing.get_args(python_type)[0]
+            return inmanta.ast.type.TypedList(to_dsl_type(base))
+
+        if issubclass(origin, dict) or origin is collections.abc.Mapping:
+            args = typing_inspect.get_args(python_type)
+            assert len(args) == 2
+            if not issubclass(args[0], str):
+                raise TypingException(None, f"invalid type {python_type}, the keys of any dict should be 'str', got {args[0]} instead")
+            return inmanta_type.TypedDict(to_dsl_type(args[1]))
+
+
+    # TODO annotated types
+    # if typing.get_origin(t) is typing.Annotated:
+    #     args: Sequence[object] = typing.get_args(python_type)
+    #     inmanta_types: Sequence[plugin_typing.InmantaType] = [arg if isinstance(arg, plugin_typing.InmantaType) for arg in args]
+    #     if inmanta_types:
+    #         if len(inmanta_types) > 1:
+    #             # TODO
+    #             raise Exception()
+    #         # TODO
+    #         return parse_dsl_type(inmanta_types[0].dsl_type)
+    #     # the annotation doesn't concern us => use base type
+    #     return to_dsl_type(args[0])
+    if python_type in python_to_model:
+        return python_to_model[python_type]
+
+    return inmanta_type.AnyType()
 
 
 class PluginValue:
@@ -267,11 +345,13 @@ class PluginValue:
             return self._resolved_type
 
         if not isinstance(self.type_expression, str):
-            raise RuntimeException(
-                stmt=None,
-                msg="Bad annotation in plugin %s for %s, expected str but got %s (%s)"
-                % (plugin.get_full_name(), self.VALUE_NAME, type(self.type_expression).__name__, self.type_expression),
-            )
+            if not isinstance(python_type, type) and typing.get_origin(python_type) is None:
+                raise RuntimeException(
+                    stmt=None,
+                    msg="Bad annotation in plugin %s for %s, expected str or python type but got %s (%s)"
+                        % (plugin.get_full_name(), self.VALUE_NAME, type(self.type_expression).__name__, self.type_expression),
+                )
+            self._resolved_type = to_dsl_type(self.type_expression)
 
         plugin_line: Range = Range(plugin.location.file, plugin.location.lnr, 1, plugin.location.lnr + 1, 1)
         locatable_type: LocatableString = LocatableString(self.type_expression, plugin_line, 0, resolver)
--- a/tests/compiler/test_plugin_types.py
+++ b/tests/compiler/test_plugin_types.py
@@ -1,0 +1,25 @@
+from typing import Mapping, Any, Union
+
+import pytest
+from inmanta.ast import RuntimeException
+from inmanta.plugins import to_dsl_type, Null
+import inmanta.ast.type as inmanta_type
+import collections.abc
+
+def test_conversion():
+    assert inmanta_type.Integer() == to_dsl_type(int)
+    assert inmanta_type.Float() == to_dsl_type(float)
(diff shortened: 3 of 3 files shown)


src/inmanta/plugins.py
```diff
@@ -219,6 +225,78 @@ def type_string_internal(self) -> str:
...
+            return inmanta_type.AnyType()
+
+    # TODO: unions
+    # bases: Sequence[inmanta.ast.type.Type] = [to_dsl_type(arg) for arg in typing.get_args(python_type)]
+    # return inmanta.ast.type.Union(bases)
+
+    if typing_inspect.is_generic_type(python_type):
+        origin = typing.get_origin(python_type)
+        if  origin is Sequence:
+            # Can this fail?
```
wouterdb, line 271: Question to reviewers

mypy-baseline.txt
```diff
@@ -791,7 +792,12 @@ src/inmanta/protocol/rest/client.py:0: error: Argument "result" to "Result" has
 src/inmanta/plugins.py:0: error: Missing type parameters for generic type "ResultVariable"  [type-arg]
 src/inmanta/plugins.py:0: error: Argument 1 to "add_function" of "PluginMeta" has incompatible type "PluginMeta"; expected "type[Plugin]"  [arg-type]
 src/inmanta/plugins.py:0: error: "type[Plugin]" has no attribute "__fq_plugin_name__"  [attr-defined]
+src/inmanta/plugins.py:0: error: Function is missing a type annotation  [no-untyped-def]
```
wouterdb, line 795: pre-existing

src/inmanta/plugins.py
```diff
@@ -219,6 +225,78 @@ def type_string_internal(self) -> str:
...
+            base: inmanta.ast.type.Type = typing.get_args(python_type)[0]
+            return inmanta.ast.type.TypedList(to_dsl_type(base))
+
+        if issubclass(origin, dict) or origin is collections.abc.Mapping:
+            args = typing_inspect.get_args(python_type)
+            assert len(args) == 2
```
sanderr, line 283: I don't think an `assert` is appropriate here.

```
>>> typing.get_args(Mapping)
()
>>> typing.get_args(Mapping[str])
(<class 'str'>,)
```

We should treat missing values as `Any`.

src/inmanta/plugins.py
```diff
@@ -219,6 +225,78 @@ def type_string_internal(self) -> str:
...
+    float: inmanta_type.Float(),
+    numbers.Number: inmanta_type.Number(),
+    int: inmanta_type.Integer(),
+    bool: inmanta_type.Bool(),
+    dict: inmanta_type.LiteralDict(),
+    list: inmanta_type.LiteralList(),
+}
+
+
+def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
```
sanderr, line 241: How about `Unknown`? Do we want to support that already? i.e. `int | Unknown` rather than `int` with `allow_unknowns=True`. Or is that fine for iso8-only. I expect we only use that in std, where we have already branched off anyway?

sanderr, line 241: How about the opaque values? Do we need those for iso7?

sanderr (changes requested):

src/inmanta/plugins.py
```diff
@@ -219,6 +227,83 @@ def type_string_internal(self) -> str:
...
+    """
+    :param python_type: The evaluated python type as provided in the Python type annotation.
```
bartv, line 243:
```suggestion
    Convert a python type annotation to an Inmanta DSL type annotation.

    :param python_type: The evaluated python type as provided in the Python type annotation.
```

src/inmanta/plugins.py
bartv: Is it the same at collections.abc.Sequence? And does this support `list[str]`?

src/inmanta/plugins.py
```diff
@@ -219,6 +225,78 @@ def type_string_internal(self) -> str:
...
 }
 
+python_to_model = {
+    str: inmanta_type.String(),
+    float: inmanta_type.Float(),
+    numbers.Number: inmanta_type.Number(),
+    int: inmanta_type.Integer(),
+    bool: inmanta_type.Bool(),
+    dict: inmanta_type.LiteralDict(),
+    list: inmanta_type.LiteralList(),
```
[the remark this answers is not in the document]
wouterdb: Upon closer consideration, what I had here was even wrong.

`inmanta.ast.type.TYPES` is the internal attribute types, which are subtly different from what is here, mapping it twice won't be correct.

src/inmanta/plugins.py
wouterdb: That is for later, but good remark

wouterdb: They don't exist there, so no.

src/inmanta/plugins.py
```diff
@@ -219,6 +225,78 @@ def type_string_internal(self) -> str:
...
+
+def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
+    """
+    :param python_type: The evaluated python type as provided in the Python type annotation.
+    """
+    # Any to any
+    if python_type is typing.Any:
+        return inmanta_type.AnyType()
+
+    if python_type is type(None):
```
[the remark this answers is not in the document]
wouterdb: oh my

src/inmanta/plugins.py
```diff
@@ -219,6 +225,78 @@ def type_string_internal(self) -> str:
...
+        return Null()
+
+    if typing_inspect.is_union_type(python_type):
+        if any(typing_inspect.is_optional_type(tt) for tt in typing.get_args(python_type)):
+            # Optional type
+            other_types = [tt for tt in typing.get_args(python_type) if not typing_inspect.is_optional_type(tt)]
```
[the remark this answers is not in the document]
wouterdb: Why?  That seems entirely as expected?

src/inmanta/plugins.py
wouterdb: thx

src/inmanta/plugins.py
sanderr: @wouterdb as to your question: I think it can indeed fail for implicit `Any`. See below
```
>>> from collections.abc import Sequence
>>> import typing
>>> typing.get_args(Sequence)
()
```

> Is it the same at collections.abc.Sequence?

It should be that one really. The import from `typing` is deprecated.

> And does this support list[str]?

I don't think so. I also had a comment about this.

src/inmanta/plugins.py
```diff
@@ -267,11 +345,13 @@ def resolve_type(self, plugin: "Plugin", resolver: Namespace) -> inmanta_type.Ty
             return self._resolved_type
 
         if not isinstance(self.type_expression, str):
```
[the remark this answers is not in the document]
wouterdb: I had understood that every string is considered to be an inmanta type

src/inmanta/plugins.py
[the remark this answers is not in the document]
sanderr: When you write it all without quotes, Python probably collapses them together so you can't get in this scenario. But then we have to be very careful about this case when we add support for quoted types.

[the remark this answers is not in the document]
sanderr: Then again, `typing_inspect.is_optional_type` doesn't seem to work at all on `typing.Annotated`.

src/inmanta/plugins.py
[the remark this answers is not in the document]
sanderr: I think I would prefer to also support python type strings, because that's what you would expect when writing Python. But that doesn't have to be now necessarily. What I had in mind was to
1. try to evaluate the annotation as a python type expression (converting any strings, even nested ones, to types)
2. convert python type expression to dsl type expression

If 1 fails -> fall back to dsl type parsing.

But I don't feel too strongly about this.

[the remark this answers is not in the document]
wouterdb: I don't want that. That would create the same type of problem that yaml has (is `00:00:00:00` a number?), that parsing outcome depends on context you (often) can't know. if a type 'string' is somehow in scope, the typing changes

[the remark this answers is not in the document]
sanderr: Perhaps a key element in my reasoning is that I approach it from the idea that the pure dsl annotations would eventually disappear, leaving only the `typing.Annotated` when you really need them. With that reasoning, the Python parsing would always be the primary, and the DSL parsing would become only a backwards compatibility fallback.

But if that's not how you see it evolve, then perhaps I agree with you, given your argument above. I still think it can have unintuitive aspects of its own, but the `typing.Annotated` should give us a way to work around any we encounter to give ourselves time to discover where we really want to go.

sanderr (changes requested):

src/inmanta/plugins.py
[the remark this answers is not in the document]
sanderr: I'm not following with this. From your comment I understood that you agreed we should use `inmanta.ast.type.TYPES`, but from the implementation I think I misunderstood? Why should we not use it?

src/inmanta/plugins.py
```diff
@@ -219,6 +225,78 @@ def type_string_internal(self) -> str:
...
+        else:
+            return inmanta_type.AnyType()
+
+    # TODO: unions
+    # bases: Sequence[inmanta.ast.type.Type] = [to_dsl_type(arg) for arg in typing.get_args(python_type)]
+    # return inmanta.ast.type.Union(bases)
+
+    if typing_inspect.is_generic_type(python_type):
+        origin = typing.get_origin(python_type)
+        if  origin is Sequence:
```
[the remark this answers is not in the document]
sanderr: I think this is still missing.

src/inmanta/ast/type.py
```diff
@@ -111,6 +114,14 @@ def get_double_defined_exception(self, other: "NamedType") -> "DuplicateExceptio
     def type_string(self) -> str:
         return self.get_full_name()
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, NamedType):
+            return False
+        return self.get_full_name() == other
```
sanderr, line 120:
```suggestion
        return self.get_full_name() == other.get_full_name()
```
Right?

src/inmanta/plugins.py
sanderr, line 1: I just realized that `typing_inspect.is_generic_type(Mapping) == False` (non-parametrized). Perhaps we have to be a bit smarter about it?

```python
origin: type = typing.get_origin(python_type) if typing_inspect.is_generic_type(python_type) else python_type
```

sanderr (changes requested):

src/inmanta/plugins.py
```diff
@@ -234,7 +234,11 @@ def __eq__(self, other: object) -> bool:
     int: inmanta_type.Integer(),
     bool: inmanta_type.Bool(),
     dict: inmanta_type.TypedDict(inmanta_type.Type()),
+    Mapping: inmanta_type.TypedDict(inmanta_type.Type()),
+    collections.abc.Mapping: inmanta_type.TypedDict(inmanta_type.Type()),
```
sanderr, line 238: Something weird is going on here. `Mapping` should be from `collections.abc`. I think it's good to support both, but then the qualification must be the other way around (or both qualified if you prefer). Because it's very likely that the import will be updated at some point.

sanderr: Same for sequence.

src/inmanta/plugins.py
```diff
@@ -277,26 +281,32 @@ def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
...
+            else:
+                raise TypingException(None, f"invalid type {python_type}, dictionary types should be Mapping or dict")
 
         # List
         if issubclass(origin, collections.abc.Sequence):
-            args = typing.get_args(python_type)
-            if not args:
-                return inmanta_type.List()
-            return inmanta_type.TypedList(to_dsl_type(args[0]))
+            if origin in [collections.abc.Sequence, list, typing.Sequence]:
```
sanderr, line 303: Not Collection?

src/inmanta/plugins.py
```diff
@@ -277,26 +281,32 @@ def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
...
 
-            if len(args) == 1:
-                return inmanta_type.TypedDict(inmanta_type.Type())
+                if len(args) == 1:
+                    return inmanta_type.TypedDict(inmanta_type.Type())
 
-            return inmanta_type.TypedDict(to_dsl_type(args[1]))
+                return inmanta_type.TypedDict(to_dsl_type(args[1]))
+            else:
+                raise TypingException(None, f"invalid type {python_type}, dictionary types should be Mapping or dict")
```
sanderr, line 299: Nice. I hadn't considered this, but it's good to have a clear message!

sanderr, line 299: Nice. I hadn't considered this, but it's good to have a clear message!

One suggestion
```suggestion
                raise TypingException(None, f"invalid type {python_type}, dictionary types should be Mapping (preferred) or dict")
```

sanderr (changes requested):

src/inmanta/plugins.py
wouterdb: Collection also contains Mapping and Set, so I think that is a bit too wide?

sanderr: But we're not accepting subclasses anymore, only those exact annotations. A list is a collection, so that would be acceptable for input args. And we can easily construct lists from it so it's acceptable as return type.

But I'm not too attached to it, it's not that common.

src/inmanta/plugins.py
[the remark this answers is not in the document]
wouterdb: I don't think we should use `inmanta.ast.type.TYPES` because:
1. I consider the type object the interface, so we should decouple on that
2. `inmanta.ast.type.TYPES` maps attribute type string to inmanta types, but this is subtly different from the type mapping at the plugin boundary. E.g. ` "dict": LiteralDict(),` Vs  `    dict: inmanta_type.TypedDict(inmanta_type.Type()),` And this is pre-existing condition

src/inmanta/plugins.py
```diff
@@ -219,6 +225,78 @@ def type_string_internal(self) -> str:
...
+        if any(typing_inspect.is_optional_type(tt) for tt in typing.get_args(python_type)):
+            # Optional type
+            other_types = [tt for tt in typing.get_args(python_type) if not typing_inspect.is_optional_type(tt)]
+            if len(other_types) == 0:
+                # Probably not possible
+                return Null()
+            if len(other_types) == 1:
+                return inmanta_type.NullableType(to_dsl_type(other_types[0]))
+            # TODO: optional unions
+            return inmanta_type.AnyType()
```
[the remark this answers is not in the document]
wouterdb: I will ticketize and we'll see when we get them.

(e.g. the unions could be easy, but then again, we don't use unions anywhere yet, so they could also be very difficult)

PR #8568, opened 2025-01-07
Fix bug in get_agents() endpoint

arnaudsjs:
# Description

This PR fixes two bugs that make the `get_agents()` endpoint behave incorrectly. More concretely the `get_agents()` endpoint will return an empty list if there is no agents record for the scheduler present in the database. The following fixes are done:

* Make sure that the background job, that cleans up old agent records from the database, doesn't clean up the record for the scheduler.
* Fix/simplify the query done by the `get_agents()` endpoint. The inner and the outer query were doing inconsistent filtering on the `name` column of the agent table.

# Self Check:

- [x] Changelog entry


wouterdb (approved):

src/inmanta/data/dataview.py
```diff
@@ -1240,26 +1240,17 @@ def get_base_query(self) -> SimpleQueryBuilder:
...
-                                    sa.name,
-                                    sa.id_primary,
-                                    ap.hostname,
-                                    ai.process
-                                FROM {Agent.table_name()} sa
-                                LEFT JOIN {AgentInstance.table_name()} ai ON sa.id_primary=ai.id
-                                LEFT JOIN {AgentProcess.table_name()} ap ON ai.process = ap.sid
-                                WHERE sa.environment = $1 AND sa.name = $2
-                            ) sa_join
+                            LEFT JOIN {AgentInstance.table_name()} ai ON a.id_primary=ai.id
```
jptrindade, line 1252:
```suggestion
                            LEFT JOIN {AgentInstance.table_name()} ai ON a.id_primary = ai.id
```

jptrindade (approved):

wouterdb (approved):

PR #8570, opened 2025-01-07
Add output option to print-default-logging-config command

arnaudsjs:
# Description

Add the --output option to the print-default-logging-config command.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -931,7 +931,7 @@ def default_log_config_parser(parser: ArgumentParser, parent_parsers: abc.Sequen
     parser.add_argument(
         "output_file",
         help="The file where the logging config should be saved. For the scheduler component, this file must end with a .tmpl"
-             " suffix, because a logging configuration template will be generated.",
+        " suffix, because a logging configuration template will be generated.",
     )
 
 
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -930,7 +930,8 @@ def default_log_config_parser(parser: ArgumentParser, parent_parsers: abc.Sequen
     )
     parser.add_argument(
         "output_file",
-        help="The file where the logging config should be saved.",
+        help="The file where the logging config should be saved. For the scheduler component, this file must end with a .tmpl"
+             " suffix, because a logging configuration template will be generated.",
     )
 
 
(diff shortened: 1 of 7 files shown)


jptrindade (approved):

wouterdb (changes requested):
You will have to update the docs as well (and the files in misc also have the command that created them on top)

src/inmanta/app.py
```diff
@@ -989,7 +998,11 @@ def default_logging_config(options: argparse.Namespace) -> None:
             for context_var in context_vars:
                 raw_dump = raw_dump.replace(f"{place_holder}{context_var}{place_holder}", "{" + context_var + "}")
 
-        print(raw_dump)
+        if options.file is not None:
```
wouterdb, line 1001: We may want to force the tmpl  suffix

src/inmanta/app.py
```diff
@@ -916,32 +916,40 @@ def cmd_parser() -> argparse.ArgumentParser:
...
 )
 def default_logging_config(options: argparse.Namespace) -> None:
+    if os.path.exists(options.output_file):
+        raise Exception(f"The requested output location already exists: {options.output_file}")
+    if options.component == "scheduler" and not options.output_file.endswith(".tmpl"):
+        raise Exception(
+            "The config being generated will be a template, but the given filename doesn't end with the .tmpl suffix."
+        )
+    if options.component == "scheduler" and not options.environment:
+        raise Exception("The -e option must be set when generating the config for the scheduler component.")
```
wouterdb, line 951: Why? It can make the template anyways?

src/inmanta/app.py
```diff
@@ -1054,7 +1063,12 @@ def app() -> None:
 
     # Log config
     component = options.component if hasattr(options, "component") else None
-    log_config.apply_options(options, component, log_context)
+    if hasattr(options, "func") and options.func.__name__ == "default_logging_config":
+        # The logging config related CLI options, passed to the output-default-logging-config command, are intended for
+        # the generated logging config and not for the logging of the execution of the command itself.
+        log_config.apply_options(Options(), component=None, context=log_context)
```
wouterdb, line 1069: I really don't like this. I would change name of the internal/hidden option called 'component' to not have to do this. 

i.e. before, the component option was not set on the `default_logging_config` because it itself is not that component, I would like to keep this this way.

wouterdb (changes requested):

wouterdb (approved):

PR #8582, opened 2025-01-08
implement conservative repairs for resources skipped for dependencies

### LINKED ISSUE #8423: implement conservative repairs of Blocked.TRANSIENT resources
The invariants of the scheduler make it safe not to schedule `Blocked.TRANSIENT` (`SkippedForDependencies()`) resources when a repair is triggered (repair button or per-resource timers) because the semantics of that state mean that we expect them to result in the exact same behavior, unless they got an event from their dependencies.

However, since this is all very new, we'd like to temporarily increase the confidence by implementing a conservative (as in fewer optimizations) mode, where we do in fact schedule them, but raise a warning when they finish a deploy with a result other than the expected one.

This would serve a dual purpose:
- allow us to detect bugs in the scheduler related to this state
- inform users of the `SkippedForDependencies()` exception of contract violations

# Implementation

- [ ] make sure repairs schedule `Blocked.TRANSIENT` resources, where currently only `Blocked.NO` are scheduled.
- [ ]  where the scheduler gets a notification of a successful deploy, check if we had expected it to skip instead (TODO: how exactly?) and log a clear warning if this is the case. The warning should explain the purpose of the `SkippedForDependencies()` exception and ask to check the resource's handler code. Additionally, it should mention that if the handler does use the exception correctly, this is likely a bug in the inmanta resource scheduler, and to please report it. Add a comment with a short motivation and pointing to this issue.
- [ ]  In both locations add a comment with a short motivation and pointing to this issue, so that the code tracks why it's there and that we have the freedom to migrate away from it when we get more confidence.

sanderr:
# Description

Make sure that repairs do still schedule TRANSIENT resources just in case. Log a warning to resource action log (db for web console + scheduler log) when this new deploy results in an unexpected success (handler should keep on skipping for dependencies until all are in good state).

closes #8423

# Self Check:

- [x] Changelog entry


src/inmanta/deploy/scheduler.py
```diff
@@ -1161,6 +1165,24 @@ async def _update_scheduler_state_for_finished_deploy(
...
                 self._state.dirty.discard(resource)
+                if state.blocked is BlockedStatus.TRANSIENT:
+                    # For now, we make sure to schedule even TRANSIENT resources for repair, just in case we have made
+                    # incorrect assumptions. This warning will trigger if that is the case. (#8580)
+                    LOGGER.warning(
+                        (
+                            # TODO: question for reviewers: is this message good? And is it sufficiently visible
+                            #   (should it be resource action log)?
+                            "Successfully deployed resource %s that was expected to skip for dependencies."
+                            " This likely indicates that its handler raised a SkipResourceForDependencies exception where it"
```
sanderr, line 1176:
```suggestion
                            " This likely indicates that its handler previously raised a SkipResourceForDependencies exception where it"
```

src/inmanta/deploy/scheduler.py
```diff
@@ -1161,6 +1165,24 @@ async def _update_scheduler_state_for_finished_deploy(
...
+                            " should have been a plain SkipResource exception. Please double check the handler implementation"
+                            " and the exception semantics."
+                            " If the raised exception is appropriate after all, this behavior may indicata a (non-critical)"
```
sanderr, line 1179:
```suggestion
                            " If the raised exception is appropriate after all, this behavior may indicate a (non-critical)"
```

src/inmanta/deploy/scheduler.py
```diff
@@ -1161,6 +1167,57 @@ async def _update_scheduler_state_for_finished_deploy(
...
+                        # TODO: test case
+                        log_line = data.LogLine.log(
+                            logging.WARNING,
+                            (
+                                "Inconsistent internal state for resource %(resource)s. The inmanta resource scheduler"
+                                " assumed it was still blocked, pending a successful deploy of at least one of its requires."
+                                " However, all dependencies are already in a deployed state. This was expected to be impossible"
+                                " and it indicates a (non-critical) bug in the inmanta resource scheduler. Please report this"
+                                " incident."
+                                " In the meantime, if you encounter any resources stuck in the skipped state, triggering"
```
bartv, line 1215: Trigger a repair as a workaround to force a deploy.

src/inmanta/deploy/scheduler.py
```diff
@@ -1199,9 +1205,23 @@ async def _update_scheduler_state_for_finished_deploy(
                 state.blocked = BlockedStatus.TRANSIENT
                 # Remove this resource from the dirty set when we block it
                 self._state.dirty.discard(resource)
+            # TODO: consider whether we also want to log on a failed deploy
             elif deployment_result is DeploymentResult.DEPLOYED:
```
sanderr, line 1209: Question for reviewers: do we also want this warning when a resource *fails* while we expect it to skip for dependencies. I seem to recall we discussed this before and that the conclusion was to only raise it on success, because a failure could be caused by any number of things, and could in fact happen before the dependencies are even checked, which would be valid.

However, @edvgui seemed to wish it to be emitted also for failure, and I promised we'd consider it. I'm personally still in favor of leaving it as is, for the reasons outlined above, and that a use case where the message might be helpful to the user in a failure case all seem like extreme edge cases to me.

src/inmanta/deploy/scheduler.py
```diff
@@ -1284,6 +1303,56 @@ def _send_events(
...
+                    " execution, indicating that it would only be able to progress once all requires reached a"
+                    " deployed state. Some requires (%(dependencies)s) are still in a non-deployed state,"
+                    " therefore it was assumed that a new deploy attempt would have no effect."
+                    " This indicates incorrect usage of the exception, and it will lead to resources getting stuck"
+                    " in the skipped state. While this can be worked around by triggering a repair for now, that"
+                    " may not work in the future."
+                    " Please check your handler implementation, and make sure to raise the generic SkipResource"
+                    " rather than SkipResourceForDependencies if you wish to skip a deploy for any other reason"
+                    " than to wait until all requires are in a good state."
+                    " Please report this incident if you believe your handler implementation is correct after all."
```
wouterdb, line 1336:
```suggestion
                    " Please contact support if you believe your handler implementation is correct after all."
```

src/inmanta/deploy/scheduler.py
```diff
@@ -1284,6 +1303,56 @@ def _send_events(
...
+            )
+        else:
+            return data.LogLine.log(
+                logging.WARNING,
+                (
+                    "Inconsistent internal state for resource %(resource)s. The inmanta resource scheduler"
+                    " assumed it was still blocked, pending a successful deploy of at least one of its requires."
+                    " However, all dependencies are already in a deployed state. This was expected to be impossible"
+                    " and it indicates a (non-critical) bug in the inmanta resource scheduler. Please report this"
+                    " incident."
```
wouterdb, line 1349:
```suggestion
                    " bug."
```

wouterdb (approved):
I didn't review the full testcase, only its description

src/inmanta/deploy/scheduler.py
sanderr: (failure can also be failure to communicate with the executor, to load the code, ...)

jptrindade: I'm in favor of communicating only on success.

jptrindade (approved):

Issue #8541, opened 2024-12-31
race conditions on blocked state in db


sanderr:
There are a few issues with the current implementation of the persisting of the blocked status of a resource. The following checklist presents these. There are multiple approaches we could take, which may end up with vastly different results. But whichever we choose, we must make sure that these checkboxes' conditions are met.

- [ ] when a recovery event unblocks (TRANSIENT -> NO) a dependent, it should be persisted in the database. The current implementation only persists the resource state for the resource that finished its deploy.
- [ ] the three paths that write to the blocked state should all be race-free, either by writing to different columns, or through proper locking

The three paths that write blocked state are
1. new version: may set it to NO or to YES
2. deploy done for resource that finished: may set it to TRANSIENT
3. deploy done for dependent resources: may set it to NO

TODO: continue the ticket description

sanderr:
Overall, I'm very much in favor of option 1 for now, to give us time for a proper solution.

sanderr:
Conclusion: for now we stop writing the transient state. I will leave todo notes linked to this ticket in the locations where they should be written. Then, for 8.1 we give this problem proper thought: can we solve it cleanly? Is the transient status sufficiently important to go through the trouble? ...

PR #8585, opened 2025-01-08
make sure transaction does not wrap scheduler lock

sanderr:
# Description

This PR addresses a concurrency issue that I forgot about in #8560: the transaction must not wrap the scheduler lock, because then parts that have been written explicitly under the lock, would only really be persisted when the transaction commits later, defeating the purpose of writing it under the lock in the first place.

Now that (since #8560) we no longer write the transient state, we can afford to simply write all state outside of the lock (which was part of the motivation for #8560).

Since the diff is quite large for such a conceptually small change, I'll make sure to mark all effective changes with a comment on the diff. You can safely assume all else has remained unchanged, apart from indentation.

part of #8541 (first stage)

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -864,98 +864,93 @@ class ResourceScheduler(TaskManager):
         #   with awaits in other operations under lock.
         #   and has to be even apart from motivations 2-3, it may interleave with
         # 2. clarity: it clearly signifies that this is the atomic and performance-sensitive part
+        async with self._scheduler_lock:
+            # update model version
+            self._state.version = model.version
+            # update resource details
+            for resource in up_to_date_resources:
+                # Registers resource and removes from the dirty set
+                self._state.update_resource(model.resources[resource], known_compliant=True)
+            for resource in new | updated:
+                # update resource state and dirty set
+                self._state.update_resource(
+                    model.resources[resource],
+                    force_new=intent_changes[resource] is ResourceIntentChange.NEW,
+                    undefined=resource in model.undefined,
+                )
+            # update requires
+            for resource in added_requires.keys() | dropped_requires.keys():
+                self._state.update_requires(resource, model.requires[resource])
+
+            # update transitive state
+            transitive_unblocked, transitive_blocked = self._state.update_transitive_state(
+                new_undefined=became_undefined,
+                verify_blocked=added_requires.keys(),
+                verify_unblocked=became_defined | dropped_requires.keys(),
+            )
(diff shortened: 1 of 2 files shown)


src/inmanta/deploy/scheduler.py
```diff
@@ -864,98 +864,93 @@ async def _new_version(
         #   with awaits in other operations under lock.
         #   and has to be even apart from motivations 2-3, it may interleave with
         # 2. clarity: it clearly signifies that this is the atomic and performance-sensitive part
-        async with self.state_update_manager.get_connection(connection=connection) as con, con.transaction():
```
sanderr, line 867: I dropped this line here

src/inmanta/deploy/scheduler.py
```diff
@@ -864,98 +864,93 @@ async def _new_version(
...
 
-                await self.state_update_manager.update_resource_intent(
-                    self.environment,
-                    intent={
-                        rid: (self._state.resource_state[rid], self._state.resources[rid])
-                        for rid in resources_with_updated_blocked_state
-                    },
-                    update_blocked_state=True,
-                    connection=con,
-                )
```
sanderr, line 950: I dropped this block, which was under the lock

src/inmanta/deploy/scheduler.py
```diff
@@ -864,98 +864,93 @@ async def _new_version(
...
+            for resource in deleted | became_undefined | transitive_blocked:
+                self._work.delete_resource(resource)
+
+            # Updating the blocked state should be done under the scheduler lock, because this state is written
+            # by both the deploy and the new version code path.
+            resources_with_updated_blocked_state: Set[ResourceIdStr] = (
+                became_undefined | transitive_blocked | transitive_unblocked
+            )
 
+        async with self.state_update_manager.get_connection(connection=connection) as con, con.transaction():
```
sanderr, line 941: Everything below this line are effective changes.

wouterdb, line 941: So the intent lock keeps this neatly sequential.

wouterdb (approved):

src/inmanta/deploy/scheduler.py
sanderr: Well, not quite, because who knows what the IO loop may do in the meantime

sanderr: Oh, intent lock... Yes, that's the idea.

arnaudsjs (approved):

PR #8586, opened 2025-01-08
Issue/8453 timers

### LINKED ISSUE #8453: rework timers to be relative to last deploy
Scheduler should keep last deploy time in state. Should be synced with persisted state table. Timers should always be relative to that time. We no longer need to cancel timers when we add something to the scheduled work. Instead, when they fire, they check the last deploy time again and either fire or reschedule themselves. There may be one repair and one deploy timer running for a resource, each with their own priority. This will then ensure proper priority bumping.

wouterdb:
# Description

Update timers to be relative to last deploy

closes #8453 

# Self Check:

- [x] Changelog entry


src/inmanta/deploy/timers.py
```diff
@@ -33,6 +37,14 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class TimerState(enum.Enum):
+
+    IDLE = 0
+    WAITING = 1
+    DEPLOY_REQUESTED = 2
+    REPAIR_REQUESTED = 3
```
Hugo-Inmanta, line 45: This isn't used afaics

src/inmanta/deploy/timers.py
```diff
@@ -266,28 +304,26 @@ def _setup_deploy(deploy_interval: int) -> tuple[int, str, TaskPriority]:
...
-
-    def update_timers(self, resources: Collection[ResourceIdStr], *, are_compliant: bool) -> None:
+    def update_timers(self, resources: Collection[ResourceIdStr]) -> None:
         """
         Make sure the given resources are marked for eventual re-deployment in the future.
+
+        Must be called under scheduler lock
+        Should not be called with resources that are already scheduled
+         -> this implies that every resource is expected to have a last_deployed time already,
+            otherwise it would be in the queue
```
Hugo-Inmanta, line 323: I'm not sure this is True for transitive_unblocked resources that were never deployed before ? https://github.com/inmanta/inmanta-core/blob/fee8e60eff44ebcc7e85278c3fa69dad24bf5466/src/inmanta/deploy/scheduler.py#L923

Hugo-Inmanta (changes requested):

src/inmanta/deploy/timers.py
```diff
@@ -200,64 +233,69 @@ async def _action() -> None:
...
-
-            return (countdown, reason, priority)
+        match state.blocked:
+            case BlockedStatus.YES:
+                self.resource_timers[resource].cancel()
+                return
+            case BlockedStatus.NO:
+                repair_only = state.status == ComplianceStatus.COMPLIANT
+            case BlockedStatus.YES:
+                repair_only = True
```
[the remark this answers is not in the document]
wouterdb: has to be transient ... :facepalm:

src/inmanta/deploy/timers.py
wouterdb: I would assume those to be in the dirty set?

src/inmanta/deploy/scheduler.py
```diff
@@ -1121,7 +1126,6 @@ async def _update_scheduler_state_for_finished_deploy(
                 DeploymentResult.NEW,
             )
 
-            # TODO[#8453]: make sure that timer is running even if we return early
             if details.attribute_hash != attribute_hash or state.status is ComplianceStatus.UNDEFINED:
```
wouterdb, line 1129: @sanderr isn't the second part (`state.status is ComplianceStatus.UNDEFINED`) implied by the first in all practical cases?

src/inmanta/deploy/scheduler.py
```diff
@@ -1131,8 +1135,10 @@ async def _update_scheduler_state_for_finished_deploy(
                 # because we might otherwise miss the recovery (in the sense that the next deploy wouldn't be a transition
                 # from a bad to a good state, since we're transitioning to that good state now).
                 state.deployment_result = deployment_result
+                state.last_deployed = finished
```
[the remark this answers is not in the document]
wouterdb: thx

src/inmanta/deploy/timers.py
wouterdb: thx, I had started working on a way to handle priority bumping when the repair interval is shorter than the deploy interval, but it was just too complicated to be worth it.

src/inmanta/deploy/scheduler.py
sanderr: It would be, except for the thing where we don't enforce the hash diff. We emit a warning if we observe this, but that still doesn't prevent it. So it is, for the 99% case, but I wanted to cover the other case.
I seem to recall that Arnaud added this at some point as well, and I made the same remark you did, so it got reverted. It was only when I was working on it myself that I realized that it was actually a good safeguard.

src/inmanta/deploy/timers.py
sanderr: Just call `deploy_resource()` again and it should get magically taken care of?

wouterdb: The issue is: call it when and with which parameters. 

when assuming the shortest timer has the highest priority, everything is simple: only the shortest one matters.

If this is not true, the complication is what state are we in?  Did we already schedule the shortest one or not? 
1. So either we run two timers (which has its own complications)
2. We keep a state machine to determine if we are past the shortest (but depending on the config and resource state, we traverse it differently, which is complicated)
3. We work based of time (are we past the shortest time now), which forces one to consider the non-monotonous nature of datetime

For a case that is not expected to happen (and where we can tell the user DUH!), I'm not going to invite this complexity in

src/inmanta/deploy/scheduler.py
```diff
@@ -765,6 +765,7 @@ async def _new_version(
         new_versions: Sequence[ModelVersion],
         *,
         up_to_date_resources: Optional[Set[ResourceIdStr]] = None,
+        last_deploy_time: dict[ResourceIdStr, datetime.datetime] | None = None,
```
sanderr, line 768: I think we should add to the docstring, similar to that of `up_to_date_resources` that "it is only expected for the very first version when the scheduler is started with a fresh state".

sanderr (changes requested):
Haven't reviewed in detail yet all the locations where timers are updated, pending [Slack discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1736421896604269).

src/inmanta/deploy/state.py
```diff
@@ -374,6 +386,7 @@ def update_resource(
                 status=compliance_status,
                 deployment_result=DeploymentResult.DEPLOYED if known_compliant else DeploymentResult.NEW,
                 blocked=blocked,
+                last_deployed=last_deployed if known_compliant else None,
```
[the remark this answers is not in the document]
wouterdb: right

src/inmanta/deploy/timers.py
```diff
@@ -55,12 +59,22 @@ def __init__(self, resource: ResourceIdStr, scheduler: "ResourceScheduler"):
...
         self.next_schedule_handle: asyncio.TimerHandle | None = None
+        self.next_scheduled_time: datetime | None = None
+        self.reason: str | None = None
+        self.priority: TaskPriority | None = None
 
         self._resource_scheduler = scheduler
 
+    def _activate(self) -> None:
+        assert self.reason is not None
+        assert self.priority is not None
```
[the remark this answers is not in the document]
wouterdb: it would complicate the situation where the time remains the same, but the priority changes

src/inmanta/deploy/state.py
```diff
@@ -259,6 +265,7 @@ async def create_from_db(
             # Populate resource_state
 
             compliance_status: ComplianceStatus
+            last_deployed = None if res["last_deploy"] is None else res["last_deploy"].astimezone()
```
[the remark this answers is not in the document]
wouterdb: For some reason they come out as UTC, not CET. So they have a timezone (they represent the same time), but they end up not equal

src/inmanta/deploy/scheduler.py
wouterdb: added to code

src/inmanta/deploy/timers.py
```diff
@@ -200,64 +224,69 @@ async def _action() -> None:
...
         to decide when re-deployment will happen and whether this timer is global (i.e. pertains
         to all resources) or specific to this resource.
 
-        The is_compliant parameter lets the caller (i.e. the resource scheduler sitting on top
-        of this TimerManager) give information regarding the last known state of the resource
-        so that this method can decide whether a repair or a deploy should be scheduled.
-
-        The scheduler is considered the authority. If a previous timer is already in place
-        for the given resource, (which should not happen) it will be canceled first and a
-        new one will be re-scheduled.
```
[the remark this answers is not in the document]
wouterdb: tried to improve

src/inmanta/deploy/state.py
```diff
@@ -163,13 +164,17 @@ def db_value(self: "BlockedStatus") -> "BlockedStatus":
 class ResourceState:
     """
     State of a resource. Consists of multiple independent (mostly) state vectors that make up the final state.
+
+    :param last_deployed: when was this resource last deployed with the latest hash?
+        Should only be set in the Compliant or Non Compliant statuses
```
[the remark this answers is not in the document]
sanderr: That's a good point. Though we practically always schedule those immediately, so the problem would quickly be resolved.

src/inmanta/deploy/state.py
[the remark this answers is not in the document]
sanderr: That really shouldn't matter though:
```
>>> now.astimezone(datetime.timezone.utc) == now.astimezone()
True
```

src/inmanta/deploy/scheduler.py
```diff
@@ -1129,7 +1131,13 @@ async def _update_scheduler_state_for_finished_deploy(
                 # except for the event to notify dependents of failure recovery (to unblock skipped for dependencies)
                 # because we might otherwise miss the recovery (in the sense that the next deploy wouldn't be a transition
                 # from a bad to a good state, since we're transitioning to that good state now).
+
+                # The second part of the or would not be required because is implied by the first,
+                # except that we don't enforce the hash diff.
+                # We emit a warning if we observe this, but that still doesn't prevent it.
+                # While it should not happen it can
```
sanderr, line 1138: Good addition, but not clear what it's attached to. Perhaps add it right above the condition?

src/inmanta/deploy/scheduler.py
```diff
@@ -914,9 +921,7 @@ async def _new_version(
             self._timer_manager.remove_timers(deleted)
             # Install timers for initial up-to-date resources. They are up-to-date now,
             # but we want to make sure we periodically repair them.
-            self._timer_manager.update_timers(
-                up_to_date_resources | (transitive_unblocked - self._state.dirty), are_compliant=True
-            )
+            self._timer_manager.update_timers(up_to_date_resources | (transitive_unblocked - self._state.dirty))
```
sanderr, line 924: I think we'll have to include `resources_with_reset_requires` here (or at least the subset that we actually updated) to cover transitions from `TRANSIENT` to `NO` (ensure repair timer is replaced with deploy timer if not compliant).
```suggestion
            self._timer_manager.update_timers(up_to_date_resources | (transitive_unblocked | resources_with_reset_requires - self._state.dirty))
```

That or add them to the dirty set when we unblock them (line 909).

src/inmanta/deploy/scheduler.py
```diff
@@ -1125,9 +1131,16 @@ async def _update_scheduler_state_for_finished_deploy(
...
+
+                # The second part of the or would not be required because is implied by the first,
+                # except that we don't enforce the hash diff.
+                # We emit a warning if we observe this, but that still doesn't prevent it.
+                # While it should not happen it can
                 state.deployment_result = deployment_result
+                state.last_deployed = finished
                 if recovered_from_failure:
                     self._send_events(details, stale_deploy=True, recovered_from_failure=True)
+                self._timer_manager.update_timer(resource, state=state)
```
sanderr, line 1143: I wonder if we might want to move those to `deploy_done()` under the `finally` for robustness. Wdyt?

Speaking of, you raised the issue of `work.finished_deploy()` needing to be in a `finally` block. I assumed that meant you did it in this PR but I don't see it. Do you plan to add it or do we create a separate PR? Just so we don't lose track of it.

src/inmanta/deploy/scheduler.py
sanderr, line 924: This suffices under the assumption that timers are already running for all other managed resources (not in the dirty set). This property is expected to hold during normal operation, but it might not when we restore state from a previous scheduler. i.e. we restore state, but we don't currently restore the timers. By restoring state, we cause some resources to be considered unchanged, i.e. we don't update their timer, hence we don't start it.

Concretely, I think we need to start them in the `_initialize` method. After calling `read_version()`, we could simply acquire the scheduler lock again, and call `update_timers(self._state.resources.keys() - self._state.dirty)`. I think that should suffice, but I'd appreciate if you could check the reasoning.

sanderr (changes requested):

src/inmanta/deploy/timers.py
sanderr: Makes sense. Now that we document that deploy is supposed to be shorter, this is a moot point.

src/inmanta/deploy/timers.py
```diff
@@ -73,19 +87,25 @@ def set_timer(
 
         Should not be called if the timer is already active.
```
Hugo-Inmanta, line 88: No longer accurate I think?

src/inmanta/deploy/state.py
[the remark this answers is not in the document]
wouterdb: indeed, it was due to unclear reporting by pycharm

src/inmanta/deploy/timers.py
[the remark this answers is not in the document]
wouterdb: updated that

src/inmanta/deploy/scheduler.py
wouterdb: done both

src/inmanta/deploy/scheduler.py
wouterdb: I think you are right. 

On the other initialization path, I added the code to read_version

src/inmanta/deploy/scheduler.py
```diff
@@ -449,7 +449,10 @@ async def _initialize(self) -> None:
                 # Set running flag because we're ready to start accepting tasks.
                 # Set before scheduling first tasks because many methods (e.g. read_version) skip silently when not running
                 self._running = True
+                # All resources get a timer
                 await self.read_version(connection=con)
+                self._timer_manager.update_timers(restored_state.resources.keys() - self._state.dirty)
```
sanderr, line 454: We should probably do this under lock though:
```suggestion
                with self._scheduler_lock:
                    self._timer_manager.update_timers(self._state.resources.keys() - self._state.dirty)
```

src/inmanta/deploy/scheduler.py
```diff
@@ -921,7 +927,9 @@ async def _new_version(
             self._timer_manager.remove_timers(deleted)
             # Install timers for initial up-to-date resources. They are up-to-date now,
             # but we want to make sure we periodically repair them.
-            self._timer_manager.update_timers(up_to_date_resources | (transitive_unblocked - self._state.dirty))
+            self._timer_manager.update_timers(
+                (last_deploy_time.keys() | transitive_unblocked | resources_with_reset_requires) - self._state.dirty
```
sanderr, line 931: I think `last_deploy_time.keys() - self._state.dirty` is equivalent to `up_to_date_resources`, but I have no objections. Except that if we do it like this, we should probably document above that either both should be set or both should be None.

src/inmanta/deploy/scheduler.py
```diff
@@ -1064,29 +1072,34 @@ async def deploy_start(self, action_id: uuid.UUID, resource: ResourceIdStr) -> O
...
+                result=result,
+                state=state,
+                started=intent.deploy_start,
+                finished=finished,
+            )
+        finally:
+            # Always do this, even if the DB is broken
+            async with self._scheduler_lock:
+                if state is not None:
+                    self._timer_manager.update_timer(intent.details.resource_id, state=state)
```
sanderr, line 1100: I think we probably want to fetch the latest state here, to make sure the timer is accurate.

In which case, the `state` declaration can move into the first try.

sanderr (changes requested):

Hugo-Inmanta (approved):

sanderr (approved):

PR #8594, opened 2025-01-10
Issue/8312 drop logging config from iso7

### LINKED ISSUE #8312: Drop the --logging-config option from iso7
Configurable logging will not be supported in iso7, as such this ticket should remove the `--logging-config` CLI option from iso7.

jptrindade:
# Description

Looked mostly to revert this PR: https://github.com/inmanta/inmanta-core/pull/7586

Let me know if I missed something

closes #8312

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/logging.py
+++ b/src/inmanta/logging.py
@@ -30,7 +30,6 @@ from logging import handlers
 from typing import Optional, TextIO
 
 import colorlog
-import yaml
 from colorlog.formatter import LogColors
 
 from inmanta import config, const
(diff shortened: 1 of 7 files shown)


src/inmanta/logging.py
```diff
@@ -593,35 +574,7 @@ def apply_options(self, options: Options) -> None:
         if self._options_applied:
             raise Exception("Options can only be applied once to a handler.")
 
-        logging_config_file: Optional[str] = self._get_path_to_logging_config_file(options)
-        if logging_config_file:
-            self._apply_logging_config_from_file(logging_config_file)
-        else:
-            self._apply_logging_config_from_options(options)
```
arnaudsjs, line 600: We should set the `_options_applied` flag at the end of this method. Otherwise we might get into an inconsistent state if an exception occurs.

arnaudsjs (changes requested):

Hugo-Inmanta (approved):

arnaudsjs (approved):

PR #8609, opened 2025-01-13
Don't refresh old facts

### LINKED ISSUE #8456: Fact refresh deploying resource that no longer exist
# Problem

1.if a resource has an UNKNOWN parameter
2. we keep refreshing that on the latest version of the resource
   - i.e. not the latest released version
   - and we keep refreshing if the resource doesn't exist in the latest version
      -  which causes code reloading issues
3. and we never remove the UNKNOWN (as long as its assoicated version exists)

arnaudsjs:
# Description

Don't refresh facts that no longer exist in the latest released model version.

closes #8456 

# Self Check:

- [x] Changelog entry


wouterdb (approved):

tests/server/test_parameters_api.py
```diff
@@ -258,3 +258,101 @@ async def test_get_set_param(environment, client, server):
...
+    monkeypatch.setattr(agent_manager_slice, "request_parameter", request_parameter_mock)
+
+    resource_id1 = "std::testing::NullResource[vm1.dev.inmanta.com,name=test1]"
+    resource_id2 = "std::testing::NullResource[vm1.dev.inmanta.com,name=test2]"
+    version = await clienthelper.get_version()
+    resources = [
+        {
+            "id": f"{resource_id1},v={version}",
+            "name": "test1",
+            "param": "val1",
```
Hugo-Inmanta, line 285: What is this "param" field ? Is this just an arbitrary field ?

arnaudsjs: It's a resource so it can have an arbitrary number of attributes. It doesn't serve a purpose in this test case. Should I remove it?

Hugo-Inmanta: > It doesn't serve a purpose in this test case. Should I remove it?

No it's fine either way

> It's a resource so it can have an arbitrary number of attributes

Is this True for all resources or only for this `testing::NullResource` ?
I'm just curious/surprised and I might be missing something obvious, but I would expect resources to have a well-defined schema.

arnaudsjs: Ah. Right. You are pointing to the fact that the attributes of the resource are not in-line with those defined in the resource definition. That's something I should fix. I didn't realize we also have a handler for that resource.

Hugo-Inmanta (approved):

PR #8610, opened 2025-01-13
Issue/8601 agent restart setting

### LINKED ISSUE #8601: add fast-restart option for scheduler
Add an environment setting to toggle the initialization behavior of the scheduler. Typically, the scheduler restores previous state from the database and builds from there. However, in rare cases (e.g. a support call) it might be useful to be able to start fresh: start with an empty state, call into the increment calculation flow and go from there.

sanderr:
# Description

Added environment setting to toggle whether the scheduler should restore its state from the database (or alternatively fall back to increment calculation). This setting is introduced exclusively as an easy fallback mechanism in case of support calls related to scheduler state inconsistencies, so we don't keep building on them. We are confident in the scheduler correctness, but we'd rather be prepared just in case.

closes #8601

# Self Check:

- [ ] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -2335,6 +2335,7 @@ def validate_cron(value: str, allow_empty: bool = True) -> str:
 AUTOSTART_AGENT_DEPLOY_SPLAY_TIME = "autostart_agent_deploy_splay_time"
 AUTOSTART_AGENT_REPAIR_INTERVAL = "autostart_agent_repair_interval"
 AUTOSTART_AGENT_REPAIR_SPLAY_TIME = "autostart_agent_repair_splay_time"
```
sanderr, line 2337: @wouterdb do you know if this option is still relevant?

src/inmanta/data/__init__.py
```diff
@@ -2335,6 +2335,7 @@ def validate_cron(value: str, allow_empty: bool = True) -> str:
...
+AUTOSTART_AGENT_RESTORE_STATE = "autostart_agent_restore_state"
```
sanderr, line 2338: I would like some input on the name, especially the prefix. "autostart" isn't all that relevant anymore, because it's the only kind of agent at the moment. But I thought it best to stick with it, simply for consistency with the other option names.

src/inmanta/data/__init__.py
```diff
@@ -2523,6 +2524,20 @@ def to_dto(self) -> m.Environment:
...
+            name=AUTOSTART_AGENT_RESTORE_STATE,
+            typ="bool",
+            default=True,
+            doc=(
+                "Restore state of the resource scheduler when the agent (re)starts to continue (close to) where it left of."
+                " This is the recommended behavior as it results in more efficient deploy scheduling at startup."
+                " The alternative is to discard all but the most basic resource state, and to fall back to deploy increment"
+                " calculation as was the norm in older versions of the Inmanta orchestrator (<= iso7/2024)."
+                " We recommend to disable this setting only in the unlikely event of certain state inconsistencies in order"
+                " to start the agent with a clean slate."
```
sanderr, line 2537: Is this clear?

src/inmanta/data/__init__.py
```diff
@@ -5985,7 +6000,7 @@ async def get_last_deployed_and_neg_increment(
         cls, environment: uuid.UUID, version: int, *, connection: Optional[Connection] = None
     ) -> tuple[set[ResourceIdStr], dict[ResourceIdStr, datetime.datetime]]:
         outset, negative, last_deployed = await cls.get_increment_and_last_deployed(environment, version, connection=connection)
-        return outset, last_deployed
+        return negative, last_deployed
```
sanderr, line 6003: Bugfix!

src/inmanta/data/__init__.py
sanderr: Because afaik we don't really expose the name / concept of "scheduler" anywhere at the moment. I see it as an internal change: it's a component of what the user knows as "the agent". But that's just how I see it, we can change that, but then I think we need to add a bunch of documentation as well.

src/inmanta/data/__init__.py
```diff
@@ -2523,6 +2524,20 @@ def to_dto(self) -> m.Environment:
             validator=convert_int,
             agent_restart=True,
         ),
+        AUTOSTART_AGENT_RESTORE_STATE: Setting(
+            name=AUTOSTART_AGENT_RESTORE_STATE,
+            typ="bool",
+            default=True,
+            doc=(
+                "Restore state of the resource scheduler when the agent (re)starts to continue (close to) where it left of."
```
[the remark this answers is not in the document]
sanderr: Related to [this comment](https://github.com/inmanta/inmanta-core/pull/8610#discussion_r1913260526). In this context, when communicating to the end user, I'd say the agent is the component that deploys resources, in its broadest sense.

src/inmanta/data/__init__.py
Hugo-Inmanta: We expose it on iso8 dev: https://docs.inmanta.com/inmanta-service-orchestrator-dev/8/architecture.html

src/inmanta/data/__init__.py
wouterdb: it isn't

src/inmanta/data/__init__.py
wouterdb: I like scheduler fast start.

src/inmanta/data/__init__.py
```diff
@@ -2523,6 +2524,20 @@ def to_dto(self) -> m.Environment:
...
+            name=AUTOSTART_AGENT_RESTORE_STATE,
+            typ="bool",
+            default=True,
+            doc=(
+                "Restore state of the resource scheduler when the agent (re)starts to continue (close to) where it left off."
+                " This is the recommended behavior as it results in more efficient deploy scheduling at startup."
+                " The alternative is to discard all but the most basic resource state, and to fall back to deploy increment"
+                " calculation as was the norm in older versions of the Inmanta orchestrator (<= iso7/2024)."
+                " We recommend to disable this setting only in the unlikely event of certain state inconsistencies in order"
+                " to start the agent with a clean slate."
```
wouterdb, line 2537: Fast restart allows the scheduler to restart faster. When disabled, the restart is more robust. Disabling this option can help to resolve scheduler problems.

src/inmanta/data/__init__.py
sanderr: I moved away from that name for two reasons:
1. I'm not sure it's always faster. Depends on how many versions we have to walk.
2. It feels a bit too vague to me. At least this gives the user an impression of what it does, and it's tied to the help string and the motivation for disabling it. I mean, when you consider a user facing the problem of "scheduler has inconsistent state", "restore-state" immediately hints at its relevance, while "fast-start" is a lot more far fetched.

Regarding AUTOSTART_AGENT vs SCHEDULER, if we do in fact document it, I think I agree that SCHEDULER is more appropriate. The only part that bothers me a bit is the inconsistency with the other options, though they do act on a slightly different level, so overall I think it's for the best.

sanderr:
@bartv requesting your review, as stated on [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1736785853581339)

src/inmanta/data/__init__.py
```diff
@@ -2523,6 +2524,19 @@ def to_dto(self) -> m.Environment:
             validator=convert_int,
             agent_restart=True,
         ),
+        RESET_DEPLOY_PROGRESS_ON_START: Setting(
+            name=RESET_DEPLOY_PROGRESS_ON_START,
+            typ="bool",
+            default=False,
+            doc=(
+                "By default the orchestrator picks up the deployment process where it was when it restarted (or halted)."
+                " When this option is enabled, the orchestrator we restart the deployment based on the last known deployment"
```
sanderr, line 2533:
```suggestion
                " When this option is enabled, the orchestrator restarts the deployment based on the last known deployment"
```

bartv (approved):

sanderr:
Requesting re-review for changes made after discussion [on Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1736785853581339).

Hugo-Inmanta (approved):

wouterdb (approved):

PR #8611, opened 2025-01-13
Added tests processing multiple versions in scheduler

### LINKED ISSUE #8569: add more tests for processing multiple versions in sequence
See [TODO note here](https://github.com/inmanta/inmanta-core/blob/4bb74bea3a516ee4376c983ec87ad3dd3fd901da/tests/deploy/test_scheduler_agent.py#L3118).

Ask sander for detailed instructions

arnaudsjs:
# Description

Added tests regarding the scheduler processing multiple versions at once.

closes #8569 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/state.py
+++ b/src/inmanta/deploy/state.py
@@ -334,6 +334,7 @@ class ModelState:
         self.requires.clear()
         self.resource_state.clear()
         self.types_per_agent.clear()
+        self.dirty = set()
 
     def update_resource(
         self,
--- a/tests/deploy/test_scheduler_agent.py
+++ b/tests/deploy/test_scheduler_agent.py
@@ -3172,7 +3172,11 @@ async def test_multiple_versions_intent_changes(agent: TestAgent, make_resource_
     assert scheduler._state.resources == all_models[-1].resources
     for rid in (rid1, rid2, rid4):
         assert scheduler._state.resource_state[rid].status is ComplianceStatus.UNDEFINED
-        assert scheduler._state.resource_state[rid].deployment_result is DeploymentResult.DEPLOYED if rid != rid4 else DeploymentResult.NEW
+        assert (
+            scheduler._state.resource_state[rid].deployment_result is DeploymentResult.DEPLOYED
+            if rid != rid4
+            else DeploymentResult.NEW
+        )
         assert scheduler._state.resource_state[rid].blocked is BlockedStatus.YES
 
     await restore_baseline_state()
--- a/tests/deploy/test_scheduler_agent.py
+++ b/tests/deploy/test_scheduler_agent.py
@@ -3182,6 +3182,8 @@ async def test_multiple_versions_intent_changes(agent: TestAgent, make_resource_
     await restore_baseline_state()
 
     # Resource becomes defined
+
+    # Make sure resources are initialized as undefined
     await scheduler._new_version(
         [
             model(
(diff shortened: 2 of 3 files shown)


src/inmanta/deploy/state.py
```diff
@@ -334,6 +334,7 @@ def reset(self) -> None:
         self.requires.clear()
         self.resource_state.clear()
         self.types_per_agent.clear()
+        self.dirty = set()
```
sanderr, line 337: Good catch!

tests/deploy/test_scheduler_agent.py
```diff
@@ -3056,6 +3055,8 @@ async def restore_baseline_state() -> None:
             assert scheduler._state.resource_state[rid].deployment_result is DeploymentResult.DEPLOYED
             assert scheduler._state.resource_state[rid].blocked is BlockedStatus.NO
         assert len(scheduler._state.dirty) == 0
+        # Make sure that _new_version() calls done by the test case do not get deployed.
```
sanderr, line 3058:
```suggestion
        # Make sure that _new_version() calls done by the test case do not get deployed to prevent races on state assertions.
```

tests/deploy/test_scheduler_agent.py
```diff
@@ -3137,12 +3138,199 @@ async def restore_baseline_state() -> None:
...
+                    rid2: all_models[-1].resources[rid2],
+                    rid4: make_resource_minimal(rid4, values={"new": "unknown"}, requires=[]),
+                },
+                undefined={rid4},
+            ),
+            model(
+                {
+                    rid1: make_resource_minimal(rid1, values={"changed": "unknown"}, requires=[]),
+                    rid2: all_models[-1].resources[rid2],
+                    rid4: make_resource_minimal(rid4, values={"new": "unknown"}, requires=[]),
```
sanderr, line 3158: You could simply use `all_models[-1].resources[rid4]` here. The purpose of that list is to keep the focus on what changes each version, rather than the full verbose thing.

That said, if you prefer it like this, I won't stand in the way.

tests/deploy/test_scheduler_agent.py
```diff
@@ -3137,12 +3138,199 @@ async def restore_baseline_state() -> None:
...
+        assert (
+            scheduler._state.resource_state[rid].deployment_result is DeploymentResult.DEPLOYED
+            if rid != rid4
+            else DeploymentResult.NEW
+        )
+        assert scheduler._state.resource_state[rid].blocked is BlockedStatus.YES
+
+    await restore_baseline_state()
+
+    # Resource becomes defined
```
sanderr, line 3184: This one is now more or less the same as the next scenario ("Resource becomes undefined, defined and then undefined again"). I think it would be good to have at least one resource that starts as undefined (i.e. from a previous `_new_version`), to make sure we cover that scenario as well. e.g. through a separate `_new_version()` call to setup the initial scenario, and only then have the multi-version `_new_version()` call.

tests/deploy/test_scheduler_agent.py
```diff
@@ -3137,12 +3138,199 @@ async def restore_baseline_state() -> None:
...
+        ]
+    )
+    assert scheduler._state.resources == all_models[-1].resources
+    assert scheduler._state.resource_state[rid1].status is ComplianceStatus.UNDEFINED
+    assert scheduler._state.resource_state[rid1].deployment_result is DeploymentResult.DEPLOYED
+    assert scheduler._state.resource_state[rid1].blocked is BlockedStatus.YES
+
+    await restore_baseline_state()
+
+    # Resource becomes defined, undefined and then defined again
```
sanderr, line 3257: I like it, but I think this one requires a setup stage as well, because the resource doesn't "become defined" in the first version, it "remains defined" because `restore_baseline_state` makes it defined.

sanderr (changes requested):
Looks good, only minor comments.

jptrindade (approved):

sanderr (approved):

PR #8613, opened 2025-01-13
renames

sanderr:
# Description

Requesting conceptual review. Context on [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1736783108439729)

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -75,7 +75,8 @@ class DryrunResult:
     messages: list[LogLine]
 
 
-class ResourceDetails:
+# rename 2
+class ResourceSpec:
     """
     In memory representation of the desired state of a resource
     """
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -75,7 +75,6 @@ class DryrunResult:
     messages: list[LogLine]
 
 
-# rename 2
 class ResourceSpec:
     """
     In memory representation of the desired state of a resource
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -68,17 +68,17 @@ class StaleResource(Exception):
 
 
 @dataclass(frozen=True)
-class ResourceIntent:
+class ResourceVersion:
     """
     Resource intent for a single resource.
     """
 
     model_version: int
-    details: ResourceDetails
+    intent: ResourceIntent
 
 
 @dataclass(frozen=True)
-class DeployIntent(ResourceIntent):
+class DeployIntent(ResourceVersion):
     """
     Deploy intent for a single resource. Includes dependency state to provide to the resource handler.
     """
@@ -302,6 +302,7 @@ class TaskRunner:
         return self.status == AgentStatus.STARTED
 
 
+# TODO: review full implementation
 class ResourceScheduler(TaskManager):
     """
     Scheduler for resource actions. Reads resource state from the database and accepts deploy, dry-run, ... requests from the
--- a/src/inmanta/deploy/state.py
+++ b/src/inmanta/deploy/state.py
@@ -54,7 +54,7 @@ class RequiresProvidesMapping(BidirectionalManyMapping["ResourceIdStr", "Resourc
         return self.reverse_mapping()
 
 
-class ComplianceStatus(StrEnum):
+class Compliance(StrEnum):
     """
     Status of a resource's operational status with respect to its latest desired state, to the best of our knowledge.
     COMPLIANT: The operational state complies to latest resource intent as far as we know.
@@ -80,7 +80,7 @@ class ComplianceStatus(StrEnum):
 
 
 @dataclass(frozen=True)
-class ResourceDetails:
+class ResourceIntent:
     resource_id: "ResourceIdStr"
     attribute_hash: str
     attributes: Mapping[str, object] = dataclasses.field(hash=False)
@@ -92,7 +92,8 @@ class ResourceDetails:
         object.__setattr__(self, "id", resources.Id.parse_id(self.resource_id))
 
 
-class DeploymentResult(StrEnum):
+# TODO: this vs executor.DeployResult
+class DeployStatus(StrEnum):
     """
     The result of a resource's last (finished) deploy. This result may be for an older version than the latest desired state.
     See ComplianceStatus for a resource's operational status with respect to its latest desired state.
@@ -135,7 +136,7 @@ class AgentStatus(StrEnum):
     STOPPED = enum.auto()
 
 
-class BlockedStatus(StrEnum):
+class Blocked(StrEnum):
     """
     YES: The resource will retain its blocked status within this model version. For example: A resource that has unknowns
          or depends on a resource with unknowns.
@@ -144,8 +145,8 @@ class BlockedStatus(StrEnum):
         Concretely it is waiting for its dependencies to deploy successfully.
     """
 
-    YES = enum.auto()
-    NO = enum.auto()
+    BLOCKED = enum.auto()
+    FREE = enum.auto()
     TRANSIENT = enum.auto()
 
     def db_value(self: "BlockedStatus") -> "BlockedStatus":
@@ -170,9 +171,9 @@ class ResourceState:
 
     # FIXME: review / finalize resource state. Based on draft design in
     #   https://docs.google.com/presentation/d/1F3bFNy2BZtzZgAxQ3Vbvdw7BWI9dq0ty5c3EoLAtUUY/edit#slide=id.g292b508a90d_0_5
-    status: ComplianceStatus
-    deployment_result: DeploymentResult
-    blocked: BlockedStatus
+    compliance: Compliance
+    deploy_status: DeployStatus
+    blocked: Blocked
     last_deployed: datetime.datetime | None
 
     def is_dirty(self) -> bool:
@@ -198,9 +199,9 @@ class ModelState:
     """
 
     version: int
-    resources: dict["ResourceIdStr", ResourceDetails] = dataclasses.field(default_factory=dict)
+    intent: dict["ResourceIdStr", ResourceIntent] = dataclasses.field(default_factory=dict)
     requires: RequiresProvidesMapping = dataclasses.field(default_factory=RequiresProvidesMapping)
-    resource_state: dict["ResourceIdStr", ResourceState] = dataclasses.field(default_factory=dict)
+    state: dict["ResourceIdStr", ResourceState] = dataclasses.field(default_factory=dict)
     # resources with a known or assumed difference between intent and actual state
     dirty: set["ResourceIdStr"] = dataclasses.field(default_factory=set)
     # types per agent keeps track of which resource types live on which agent by doing a reference count
--- a/src/inmanta/deploy/state.py
+++ b/src/inmanta/deploy/state.py
@@ -92,7 +92,6 @@ class ResourceIntent:
         object.__setattr__(self, "id", resources.Id.parse_id(self.resource_id))
 
 
-# TODO: this vs executor.DeployResult
 class DeployStatus(StrEnum):
     """
     The result of a resource's last (finished) deploy. This result may be for an older version than the latest desired state.


src/inmanta/agent/executor.py
```diff
@@ -75,7 +75,7 @@ class DryrunResult:
     messages: list[LogLine]
 
 
-class ResourceDetails:
+class ResourceSpec:
```
sanderr, line 78: rename 1.

motivation: conflict with `deploy.state.ResourceDetails`.

notes:
- feels somewhat clunky. Then again, so does `ResourceDetails`.
- I also renamed `deploy.state.ResourceDetails`, so this rename here is not strictly necessary. Let's only keep it if we consider it an improvement even apart from the name conflict.

src/inmanta/deploy/scheduler.py
```diff
@@ -68,17 +68,17 @@ class StaleResource(Exception):
 
 
 @dataclass(frozen=True)
-class ResourceIntent:
+class ResourceVersion:
```
sanderr, line 71: rename 2.

motivation: conflict with proposed rename of `deploy.state.ResourceDetails` -> `deploy.state.ResourceIntent`

src/inmanta/deploy/scheduler.py
```diff
@@ -302,6 +302,7 @@ def is_running(self) -> bool:
         return self.status == AgentStatus.STARTED
 
 
+# TODO: review full implementation
```
sanderr, line 305: note to self, not relevant for reviewers.

src/inmanta/deploy/state.py
```diff
@@ -54,7 +54,7 @@ def provides_view(self) -> Mapping["ResourceIdStr", Set["ResourceIdStr"]]:
         return self.reverse_mapping()
 
 
-class ComplianceStatus(StrEnum):
+class Compliance(StrEnum):
```
sanderr, line 57: rename 3.

motivation: reduce verbosity.

notes:
- not required for any name conflicts, can be dropped if not considered an improvement.
- rename of `ResourceState.status` -> `ResourceState.compliance` will require a db migration script.

src/inmanta/deploy/state.py
```diff
@@ -80,7 +80,7 @@ def is_dirty(self) -> bool:
 
 
 @dataclass(frozen=True)
-class ResourceDetails:
+class ResourceIntent:
```
sanderr, line 83: rename 4.

motivation: conflict with `executor.ResourceDetails` + `ResourceDetails` simply isnt't very revealing. `ResourceIntent` more clearly indicates what the object represents.

src/inmanta/deploy/state.py
```diff
@@ -92,7 +92,7 @@ def __post_init__(self) -> None:
         object.__setattr__(self, "id", resources.Id.parse_id(self.resource_id))
 
 
-class DeploymentResult(StrEnum):
+class DeployStatus(StrEnum):
```
sanderr, line 95: rename 5.

motivation: conflict with `executor.DeployResult`. The executor one is aptly named imo.

notes:
- rename of `ResourceState.deployment_result` -> `ResourceState.deploy_status` will require db migration script

personal opinion: ambiguity improvement => make the change.

src/inmanta/deploy/state.py
```diff
@@ -135,7 +135,7 @@ class AgentStatus(StrEnum):
     STOPPED = enum.auto()
 
 
-class BlockedStatus(StrEnum):
+class Blocked(StrEnum):
```
sanderr, line 138: rename 6.

motivation: reduce verbosity + make consistent with field name `ResourceState.blocked`

personal opinion: tiny readability improvement => make the change.

src/inmanta/deploy/state.py
```diff
@@ -144,8 +144,8 @@ class BlockedStatus(StrEnum):
         Concretely it is waiting for its dependencies to deploy successfully.
     """
 
-    YES = enum.auto()
-    NO = enum.auto()
+    BLOCKED = enum.auto()
+    FREE = enum.auto()
```
sanderr, line 148: rename 7.

motivation: YES/NO always felt clunky to me.

notes:
- not sure about `FREE`. Could also be `NOT_BLOCKED`
- will require db migration script

personal opinion: on the fence, input welcome.

src/inmanta/deploy/state.py
```diff
@@ -170,9 +170,9 @@ class ResourceState:
 
     # FIXME: review / finalize resource state. Based on draft design in
     #   https://docs.google.com/presentation/d/1F3bFNy2BZtzZgAxQ3Vbvdw7BWI9dq0ty5c3EoLAtUUY/edit#slide=id.g292b508a90d_0_5
-    status: ComplianceStatus
-    deployment_result: DeploymentResult
-    blocked: BlockedStatus
+    compliance: Compliance
+    deploy_status: DeployStatus
```
sanderr, line 174: renames 3b & 5b. No comments needed, see respective threads for details.

src/inmanta/deploy/state.py
```diff
@@ -198,9 +198,9 @@ class ModelState:
     """
 
     version: int
-    resources: dict["ResourceIdStr", ResourceDetails] = dataclasses.field(default_factory=dict)
+    intent: dict["ResourceIdStr", ResourceIntent] = dataclasses.field(default_factory=dict)
     requires: RequiresProvidesMapping = dataclasses.field(default_factory=RequiresProvidesMapping)
-    resource_state: dict["ResourceIdStr", ResourceState] = dataclasses.field(default_factory=dict)
+    state: dict["ResourceIdStr", ResourceState] = dataclasses.field(default_factory=dict)
```
sanderr, line 203: rename 8.

motivation: verbosity + more explicitly differentiate state from intent (important for locking etc).

notes:
- I have to double check but I don't think this requires a db migration script.

personal opinion: small readability improvement => make the change.

src/inmanta/agent/executor.py
wouterdb: It is the same as the ResourceIntent, but in a different domain. 

We don't really have a convention for it, but it is not weird for the same concept to have the same name.

src/inmanta/deploy/state.py
wouterdb: I like this a lot

src/inmanta/deploy/state.py
wouterdb: I also didn't like the YES/NO.

src/inmanta/deploy/state.py
wouterdb: I like that `result` makes clear this is about the outcome of a specific event, not some aggregated status. 
So I'm a bit on the fence about this one.

src/inmanta/deploy/state.py
wouterdb: It is a bit weird that we have `compliance` (not `compliance_status`) and `deploy_status` not `deploy` 

I would go for 

`compliance`, `deploy_result` and `blocked`

src/inmanta/agent/executor.py
arnaudsjs: I think we have to add a prefix or suffix to indicate to indicate the domain: SchedulerResourceIntent and ExecutorResourceIntent. Then it's clear they represent something similar and at the same time it's easy to keep them apart.

src/inmanta/deploy/scheduler.py
arnaudsjs: I find `ResourceVersion` a strange name, because it represents a log more than just a version. What about `ResourceVersionIntent` in analogy with `ResourceId` and `ResourceVersionId`?

src/inmanta/deploy/state.py
arnaudsjs: I like it.

src/inmanta/deploy/state.py
arnaudsjs: I like it.

src/inmanta/deploy/state.py
arnaudsjs: * If we are going to change this, I wonder whether we should use `PERMANENTLY_BLOCKED` instead of `BLOCKED`. It would make more clear what it represents.
* I am not convinced that `FREE` is any better than `NO`, but I have no objections against the rename.

src/inmanta/agent/executor.py
sanderr: > It is the same as the ResourceIntent, but in a different domain.

I partially agree, but on the other hand it also contains a bunch of metadata that could be argued to be outside of the resource's own intent, like the model version and requires. That's why I don't find intent very appropriate here, while I find it perfectly appropriate for the other one.

> I think we have to add a prefix or suffix to indicate the domain: SchedulerResourceIntent and ExecutorResourceIntent. Then it's clear they represent something similar and at the same time it's easy to keep them apart.

I personally don't like this. I agree that context is occasionally important. But when it is, I'm strongly of the opinion that it should be provided by qualified imports, not name prefixing.

src/inmanta/deploy/scheduler.py
sanderr: I'm fine with that. Part of the reason for leaving the `Intent` part off was verbosity, part was to more clearly differentiate between this class and its `intent` attribute. But neither of those arguments are very strong.

src/inmanta/deploy/state.py
sanderr: For me, `executor.DeploymentResult` is very clearly the outcome of an action. This one slightly less so. But, now that I think about it, I do have to admit that it's very close.

I do think the name similarity remains confusing.

src/inmanta/deploy/state.py
sanderr: But do you prefer the new proposal?

src/inmanta/deploy/state.py
sanderr: `deploy_result: DeployResult` or `deployment_result: DeploymentResult`? It was previously the latter.

src/inmanta/agent/executor.py
arnaudsjs: > I'm strongly of the opinion that it should be provided by qualified imports, not name prefixing.

The problem with using qualified imports is that it doesn't make you actively aware about the fact that there is another class with exactly the same name somewhere in the code base. So using qualified imports to make the distinction doesn't make it easier for someone to enter the code base.

src/inmanta/deploy/state.py
sanderr: > If we are going to change this, I wonder whether we should use PERMANENTLY_BLOCKED instead of BLOCKED. It would make more clear what it represents.

I like the suggestion, though I think "permanent" is too... well, permanent. I still prefer the simple `BLOCKED` overall, but perhaps we can find a middle ground.

> I am not convinced that FREE is any better than NO

Me neither tbh. Wdyt of `NOT_BLOCKED` versus `FREE` and `NO`?

arnaudsjs: I think I like `FREE` the least of all. I don't have a strong preference amongst. `NON_BLOCKED` or `NO`. Maybe a slight preference for `NON_BLOCKED`, because it's consistent with `BLOCKED`.

arnaudsjs: > I like the suggestion, though I think "permanent" is too... well, permanent.

I see your point. I can't immediately think of a better name so maybe we have to stick with `BLOCKED` after all.

src/inmanta/deploy/scheduler.py
sanderr: @wouterdb I'm siding with @arnaudsjs here. Do you agree with `ResourceVersionIntent` for this one?

src/inmanta/deploy/state.py
arnaudsjs: I think we have to stress here that this represents in the outcome of the last deploy. So what about `StatusLastDeploy`?

src/inmanta/deploy/state.py
sanderr: What would you think of `TRANSIENT -> TEMPORARILY_BLOCKED`? It's more verbose than I'd like, but it's consistent, and perhaps more clear than transient (especially considering that transitive sounds so similar and is also an important part of the blocking).

arnaudsjs: > What would you think of TRANSIENT -> TEMPORARILY_BLOCKED?

I like it a lot. I also agree on the verbosity remark, but I am not afraid of using longer names if it makes things more clear.

src/inmanta/deploy/state.py
sanderr: Or perhaps, combining your and @wouterdb's concerns, `LastDeployResult`? But, while I can see that for the field name (`last_deploy_result`), I don't feel like it belongs in the enum name. The fact that it represents the last deploy is a property of the `ResourceState` field, not of the enum itself, which represents nothing more than "the result of a deploy"

src/inmanta/deploy/state.py
sanderr: Ok, new consolidated proposal for this one:
```python
class Blocked(Enum):
    BLOCKED
    NOT_BLOCKED
    TEMPORARILY_BLOCKED
```

Wdyt? (@wouterdb as well)

src/inmanta/agent/executor.py
sanderr: I'm still inclined to leave this one at `ResourceDetails`. It's not perfect, but it does describe what it is, i.e. all details required for a resource action on a given resource.

The scheduler on the other hand deals a lot more directly with intent, irrespective of model version or requires-provides. The name change that we agreed in in that domain should address Arnaud's naming concerns if we leave this one as is.

@wouterdb how strongly do you feel we should rename this one?

arnaudsjs: The problem with the qualified imports approach is that I bet that someone in the team will introduce a PR in the coming 6 months (if it even takes that long) that is not using the qualified imports for this case. It's an approach that works if you work alone on a project but it doesn't work when there is a team of developers involved.

src/inmanta/deploy/state.py
sanderr: I'm inclined to leave this one as is if we can't find something that we all agree is better. Optionally with a field rename to prefix it with `last_`.

But, that would mean that we are stuck with the name similarity between this and `executor.DeployResult`. And I think that one is definitely a good name. So, that's definitely unfortunate. But perhaps acceptable, assuming we use qualified imports (we don't everywhere atm), and the fact that we simply don't see anything that's better per se?

src/inmanta/agent/executor.py
sanderr: I agree there's a risk to that, but I also think we should simply pay more attention to imports, and we have been growing better in that regard over the past year or so. I'd love to continue this discussion, in part *because* it's something I've been thinking about a lot recently (and discussing with various people at the office), but perhaps this is not the time or place, considering that we have a proposal that gets rid of the ambiguity here? Unless you don't like that proposal?

wouterdb: I don't feel very strong about this at all.

src/inmanta/deploy/state.py
wouterdb: no strong preference

wouterdb (commented):
Replied to the comments

src/inmanta/agent/executor.py
arnaudsjs: It's not perfect. But it's better than two classes with the same name. So it's fine by me.

src/inmanta/deploy/state.py
sanderr: Discussed in person, agreed on that latest proposal

sanderr: Discussed in person again: addendum: change `executor.DeployResult` -> `executor.DeployReport` and similar for the other resource action results/reports.

src/inmanta/deploy/state.py
sanderr, line 57: Note to self: db migration script

sanderr: `status -> compliance`

arnaudsjs (approved):

src/inmanta/deploy/state.py
sanderr: `deployment_result` -> `last_deploy_result`

sanderr: `YES/NO/TRANSIENT` -> ...

sanderr: check `inmanta.data` as well

sanderr, line 57: Note to self: db dump

sanderr:
Superseded by #8630

PR #8614, opened 2025-01-13
Issue/fix outdated iso8 documentation

Hugo-Inmanta:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/docs/model_developers/modules.rst
+++ b/docs/model_developers/modules.rst
@@ -252,9 +252,9 @@ The Inmanta Python SDK offers several extension mechanism:
 * Dependency managers
 
 Only the compiler and executors load code included in modules (See :doc:`/architecture`). A module can include external
-dependencies. Both the compiler and the executors will install this dependencies with ``pip install``, each in a
+dependencies. Both the compiler and the executors will install these dependencies with ``pip install``, each in a
 dedicated virtual environment. By default this is in `.env` of the project for the compiler and in
-`/var/lib/inmanta/agent/env` for the agent.
+`/var/lib/inmanta/server/<environment>/executors/venvs/` for the executors.
 
 Inmanta uses a special format of requirements that was defined in python PEP440 but never fully
 implemented in all python tools (setuptools and pip). Inmanta rewrites this to the syntax pip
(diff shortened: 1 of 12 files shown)


docs/administrators/operational_procedures.rst
Hugo-Inmanta, line 1: These settings were removed in https://github.com/inmanta/inmanta-core/commit/fce9311ce805f075430411bd9f88d68ef45ba1b9

wouterdb (approved):

docs/install/1-install-server.rst
```diff
@@ -376,34 +376,14 @@ Replace <password> in the above-mentioned snippet with the password of the inman
...
-in the configuration file stored at ``/etc/inmanta/inmanta.d/server.cfg``.
-
-.. code-block:: text
-
-  [server]
-  server-address=<server-ip-address-or-hostname>
-
-.. note:: If you deploy configuration models that modify resolver configuration it is recommended to use the IP address instead
-  of the hostname.
-
```
arnaudsjs, line 398: This cannot be dropped, can it? The scheduler needs this option to know how to connect back to the server.

docs/install/1-install-server.rst
```diff
@@ -433,11 +413,10 @@ ssh and if the remote agent is used.
 
     sudo chown inmanta:inmanta /var/lib/inmanta/.ssh/config
 
-3. Add the public key to any git repositories and save it to include in configuration models that require remote agents.
```
arnaudsjs, line 436: Why drop 3? The git repository part is still relevant, right?

docs/install/1-install-server.rst
```diff
@@ -433,11 +413,10 @@ ssh and if the remote agent is used.
...
-4. Test if you can login into a machine that has the public key and make sure ssh does not show you any prompts to store
+3. Test if you can login into a machine that has the public key and make sure ssh does not show you any prompts to store
```
arnaudsjs, line 416: Shouldn't we say here "try to clone git repos", instead of "try to login to machines"?

docs/model_developers/images/handler_flow.svg
arnaudsjs, line 1: I think that "executor" needs both `resource(python)` and `handler(python)`.

docs/model_developers/modules.rst
```diff
@@ -357,19 +357,19 @@ By following the previous steps, the Inmanta server will be able to install modu
 Inter-module dependencies
 #########################
 
-The plugins code of a module mod-a can have a dependency on the plugins code of another V2 module mod-b. When doing this,
+The plugins code of a module ``mod-a`` can have a dependency on the plugins code of another V2 module ``mod-b``. When doing this,
```
arnaudsjs, line 360: I think we can remove this entire section. We can no longer run into this issue because we create a new executor process when the code for that executor changes.

docs/reference/projectyml.rst
```diff
@@ -95,8 +95,6 @@ Setting this to ``true`` will have the following consequences:
 - If ``pre`` is set, it will supersede pip's ``pre`` option set by the ``PIP_PRE`` environment variable or in pip
 config file. When true, pre-release versions are allowed when installing v2 modules or v1 modules' dependencies.
 
-- Auto-started agents live on the same host as the server, and so they will share the pip config at the system level.
-
```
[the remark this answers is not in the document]
arnaudsjs: Why not just replace `Auto-started agents` with `Executors`?

arnaudsjs (changes requested):

docs/install/1-install-server.rst
Hugo-Inmanta: oof, right

docs/install/1-install-server.rst
Hugo-Inmanta: Aah, do I understand correctly that this is to set up repos where developers would store .cf files. This has nothing to do with v1 modules, right ?

arnaudsjs: This is required for two reasons:

* For cloning the Inmanta project. This is always a git repo.
* For cloning V1 modules (if used).

docs/model_developers/images/handler_flow.svg
Hugo-Inmanta: Improved the picture in the scheduler as well to show ordering

arnaudsjs (approved):

PR #8616, opened 2025-01-13
Cleanup handling of the unavailable state

### LINKED ISSUE #8491: Cleanup handling of the unvailable state
in_process_executor.deserialize reports the unavailable state back to the server, not the scheduler. 

This should be cleaned up

jptrindade:
# Description

Changed handling of the unavailable state to be coherent with the rest of the code. We now return a `FactResult/DryrunResult` with additional information, so that the scheduler can handle it accordingly.

closes #8491 

# Self Check:

- [x] Changelog entry


tests/deploy/e2e/test_dryrun.py
```diff
@@ -243,6 +243,7 @@ def assert_handler_failed(resource, msg):
     response = await client.get_resource(environment, "test::DoesNotExist[agent1,key=key2],v=%d" % version, logs=True)
     assert response.code == 200
 
+    # TODO: Do we want this behaviour?
     # resource stays available but an unavailable state is logged because of the failed dryrun
```
jptrindade, line 247: Was this behaviour intentional?

tests/deploy/e2e/test_dryrun.py
```diff
@@ -243,6 +243,7 @@ def assert_handler_failed(resource, msg):
     response = await client.get_resource(environment, "test::DoesNotExist[agent1,key=key2],v=%d" % version, logs=True)
     assert response.code == 200
 
+    # TODO: Do we want this behaviour?
```
jptrindade, line 246: Was this behaviour intentional?

i.e. We log an `unavailable` resource action as well as a `dryrun` action. And we log them in such a way that the unavailable is the first to be returned when we fetch the logs?

src/inmanta/deploy/persistence.py
```diff
@@ -302,16 +302,33 @@ async def dryrun_update(self, env: UUID, dryrun_result: executor.DryrunResult) -
...
+                await self._write_resource_action(
+                    env,
+                    dryrun_result.rvid,
+                    const.ResourceAction.dryrun,
+                    uuid.uuid4(),
+                    const.ResourceState.unavailable,
+                    # Small hack so that when we return the logs ordered by started time,
+                    # we don't get inconsistencies in the ordering by having 2 logs with the same started time.
+                    dryrun_result.started + datetime.timedelta(milliseconds=1),
+                    dryrun_result.finished + datetime.timedelta(milliseconds=1),
```
jptrindade, line 328: I added this because the test in which I added a comment, was flaky.
It was not flaky before because we had different start times

wouterdb: Why write two resource actions?

src/inmanta/agent/in_process_executor.py
```diff
@@ -286,15 +261,23 @@ async def dry_run(
             with self._cache:
                 started = datetime.datetime.now().astimezone()
                 try:
-                    resource_obj: Resource | None = await self.deserialize(resource, const.ResourceAction.dryrun)
-                except Exception:
+                    resource_obj: Resource = Resource.deserialize(resource.attributes)
+                except Exception as e:
+                    msg = data.LogLine.log(
```
wouterdb, line 266: I think you have to flush this to the logger (it has a method for that)

src/inmanta/agent/in_process_executor.py
```diff
@@ -389,17 +372,25 @@ async def get_facts(self, resource: ResourceDetails) -> FactResult:
         started = datetime.datetime.now().astimezone()
         try:
             try:
-                resource_obj: Resource | None = await self.deserialize(resource, const.ResourceAction.getfact)
-            except Exception:
+                resource_obj: Resource = Resource.deserialize(resource.attributes)
+            except Exception as e:
+                msg = data.LogLine.log(
```
wouterdb, line 377: I think you have to flush this to the logger (it has a method for that)

tests/deploy/e2e/test_dryrun.py
wouterdb: I don't think it is very valuable to have two log lines for the same thing

wouterdb (changes requested):

tests/deploy/e2e/test_dryrun.py
jptrindade: One of them has the status as `dry` and the other as `unavailable`. It is the behaviour that was present before my changes.

Should we keep only the `dry` or the `unavailable`?

src/inmanta/deploy/persistence.py
jptrindade: One of them has the status as `dry` and the other as `unavailable`. It is the behaviour that was present before my changes.

Should we keep only the dry` or the `unavailable`?

tests/deploy/e2e/test_dryrun.py
jptrindade: I agree, which do you want to keep?

I'm inclined to keep the `unavailable` one. I think that makes the most sense

src/inmanta/agent/in_process_executor.py
jptrindade: Added the `_log_deserialization_error` method that I believe does this

src/inmanta/deploy/persistence.py
wouterdb: the unavailable indeed

wouterdb (approved):

wouterdb (approved):

PR #8622, opened 2025-01-14
Remove splaying

wouterdb:
# Description

Remove splay env options

1. the config values were already removed
2. I removed the option, but it will keep on existing in the DB if set. I think this is OK because it is small and the api doesn't return it. 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/deploy/e2e/test_timers.py
+++ b/tests/deploy/e2e/test_timers.py
@@ -72,10 +72,7 @@ async def test_notification_mechanism(agent, environment, clienthelper, client, 
 
     def is_approx(rid: str, seconds: int) -> None:
         time = tm.resource_timers[rid].next_scheduled_time
-        assert abs(time - last_deploy_time_approx - timedelta(seconds=seconds)) < timedelta(milliseconds=100)
-
-    def is_disabled(rid) -> None:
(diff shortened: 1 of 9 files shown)


tests/deploy/e2e/test_timers.py
```diff
@@ -72,10 +72,7 @@ async def test_notification_mechanism(agent, environment, clienthelper, client,
 
     def is_approx(rid: str, seconds: int) -> None:
         time = tm.resource_timers[rid].next_scheduled_time
-        assert abs(time - last_deploy_time_approx - timedelta(seconds=seconds)) < timedelta(milliseconds=100)
-
-    def is_disabled(rid) -> None:
-        assert rid not in tm.resource_timers or tm.resource_timers[rid].when is None
+        assert abs(time - last_deploy_time_approx - timedelta(seconds=seconds)) < timedelta(milliseconds=200)
```
Hugo-Inmanta, line 75: Was `timedelta(milliseconds=100)` causing too much flakiness ?

wouterdb: on jenkins it was

sanderr (approved):

Hugo-Inmanta (approved):

PR #8624, opened 2025-01-14
Fixed test

wouterdb:
# Description

Fix existing test

also fix broken and untested and very old api endpoint

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/agentmanager.py
+++ b/src/inmanta/server/agentmanager.py
@@ -844,7 +844,7 @@ class AgentManager(ServerSlice, SessionListener):
     @handle(methods.list_agents, env="tid")
     async def list_agents(
         self,
-        env: Optional[data.Environment],
+        env: data.Environment,
         start: Optional[str] = None,
         end: Optional[str] = None,
         limit: Optional[int] = None,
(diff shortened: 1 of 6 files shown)


src/inmanta/server/agentmanager.py
```diff
@@ -858,30 +858,16 @@ async def list_agents(
         :raises BadRequest: Limit, start and end can not be set together
         :raises BadRequest: Limit parameter can not exceed 1000
         """
-        query = {}
-        if env is not None:
-            query["environment"] = env.id
+        new_agent_endpoint = await self.get_agents(env, limit, start, end, start, end)
```
wouterdb, line 861: Not the nicest fix, but is does work (insofar this is tested)

tests/test_agent_manager.py
```diff
@@ -446,46 +446,19 @@ async def test_api(init_dataclasses_and_load_schema):
...
-                "primary": UNKWN,
-                "environment": env4.id,
-                "state": "up",
-            },
-            {"name": "agent3", "paused": False, "last_failover": "", "primary": "", "environment": env4.id, "state": "down"},
-            {"name": "agent4", "paused": False, "last_failover": "", "primary": "", "environment": env2.id, "state": "down"},
-        ]
-    }
-    assert_equal_ish(shouldbe, all_agents, sortby=["environment", "name"])
+    assert code == 404
```
wouterdb, line 449: This is a behavior change: I don't allow the environment to be None, but I suspect the API library would not allow it either

arnaudsjs (approved):

Hugo-Inmanta (approved):

PR #8625, opened 2025-01-14
Log warning if old logging configuration option is ignored

### LINKED ISSUE #8338: Log warning if old/ignored logging configuration option is used
Tasks:

- [ ] Log warnings
- [x] ~~Add deprecation notice in the changelog~~
- [x] ~~Create a ticket to remove the options in ISO9~~

jptrindade:
# Description

Log warning if we successfully found a logging config but have any of these options set:
- log_file
- log_file_level

closes #8338 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/logging.py
+++ b/src/inmanta/logging.py
@@ -803,7 +803,7 @@ class InmantaLoggerConfig:
                 "keep_logger_names": "--keep-logger-names",
             }
             ignored_options = [
-                f"{value} {options.__getattribute__(key)}" for key, value in option_to_cli.items() if key in options
+                f"{value} {getattr(options, key)}" for key, value in option_to_cli.items() if key in options
             ]
             if len(ignored_options) != 0:
                 return ", ".join(ignored_options)
--- a/src/inmanta/logging.py
+++ b/src/inmanta/logging.py
@@ -792,12 +792,25 @@ class InmantaLoggerConfig:
         :param context: context variables to use if the config file is a template
         """
 
-        def user_defined_options() -> dict[str, bool]:
+        def user_defined_options() -> Optional[str]:
             """
-            Returns a dictionary with the options (excluding "verbose" and "logging_config") that were set by the user
+            Returns a string with the options (excluding "--verbose" and "--logging-config") that were set by the user
             """
             ignore_keys = ["verbose", "logging_config"]
-            return {key: value for key, value in options.__dict__.items() if key not in ignore_keys and key in options}
+
+            option_to_cli = {
+                "log_file": "--log-file",
+                "log_file_level": "--log-file-level",
+                "timed": "--timed-logs",
+                "keep_logger_names": "--keep-logger-names",
+            }
+            ignored_options = [
+                f"{option_to_cli[key]} {value}"
+                for key, value in options.__dict__.items()
+                if key not in ignore_keys and key in options
+            ]
+            if len(ignored_options) != 0:
+                return ", ".join(ignored_options)
 
         if self._options_applied:
             raise Exception("Options can only be applied once to a handler.")
--- a/src/inmanta/logging.py
+++ b/src/inmanta/logging.py
@@ -285,10 +285,10 @@ class Options(Namespace):
     """
 
     log_file: Optional[str] = None
-    log_file_level: Optional[str] = None
+    log_file_level: str = "INFO"
     verbose: int = 1
-    timed: Optional[bool] = None
-    keep_logger_names: Optional[bool] = None
+    timed: bool = False
+    keep_logger_names: bool = False
     logging_config: Optional[str] = None
 
 
@@ -343,15 +343,6 @@ class LoggingConfigBuilder:
         )
         return logging_config_core
 
-    def set_defaults_for_options(self, options: Options) -> Options:
-        """
-        Sets the default value of fields that were not set by the user and have a default value
-        """
-        options.log_file_level = options.log_file_level if options.log_file_level is not None else "INFO"
-        options.timed = options.timed if options.timed is not None else False
-        options.keep_logger_names = options.keep_logger_names if options.keep_logger_names is not None else False
-        return options
-
     def get_logging_config_from_options(
         self,
         stream: TextIO,
@@ -373,8 +364,6 @@ class LoggingConfigBuilder:
         handler_root_logger: str
         log_level: int
 
-        # Set the default values for fields that were not set by the user
-        options = self.set_defaults_for_options(options)
         log_file_cli_option = options.log_file
 
         short_names = False
@@ -396,7 +385,6 @@ class LoggingConfigBuilder:
 
         # Shared config
         if log_file_cli_option:
-            assert options.log_file_level is not None  # make mypy happy, always set in set_defaults_for_options()
             log_level = convert_inmanta_log_level(options.log_file_level)
             handler_root_logger = f"{component}_handler" if component is not None else "root_handler"
             handlers[handler_root_logger] = {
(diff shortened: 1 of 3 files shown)


arnaudsjs (changes requested):
Why don't we log a warning for the other log-related configuration options?

* `--timed-logs`
* `--keep-logger-names`

tests/test_logging.py
```diff
@@ -178,28 +178,36 @@ def setup_logging_config(cli_options: Options, file_option_value: Optional[str]
...
     path_logging_config_file1 = os.path.join(tmpdir, "logging_config1.yml")
     path_logging_config_file2 = os.path.join(tmpdir, "logging_config2.yml")
     path_logging_config_file3 = os.path.join(tmpdir, "logging_config3.yml")
 
     # Set --logging-config option on CLI only
     write_logging_config_file(path=path_logging_config_file1, formatter="AAA %(message)s")
     # Also assert that the other config options are ignored when logging_config is set.
-    setup_logging_config(cli_options=Options(logging_config=path_logging_config_file1, verbose=1), file_option_value=None)
+    setup_logging_config(
+        cli_options=Options(logging_config=path_logging_config_file1, verbose=1, log_file=old_log_file, log_file_level="4"),
```
Hugo-Inmanta, line 192: Could we also test that the warning was properly emitted ?

jptrindade: @arnaudsjs the discussion is related to this comment.
I want to keep all the assertions in the code and also assert that warnings are properly emitted

src/inmanta/logging.py
```diff
@@ -340,6 +343,15 @@ def get_bootstrap_logging_config(
...
 
+    def set_defaults_for_options(self, options: Options) -> Options:
+        """
+        Sets the default value of fields that were not set by the user and have a default value
+        """
+        options.log_file_level = options.log_file_level if options.log_file_level is not None else "INFO"
+        options.timed = options.timed if options.timed is not None else False
+        options.keep_logger_names = options.keep_logger_names if options.keep_logger_names is not None else False
+        return options
+
```
arnaudsjs, line 354: This is rather verbose/complex. There is an easier way to determine whether a value is provided on the CLI and that's by relying on the `in` operator. You can for example do `"log_file_level" in options` to know whether the `--log-file-level` option was set.

arnaudsjs (changes requested):

src/inmanta/logging.py
jptrindade: Hum, I was not aware that it worked like that. Your solution is much better, will push the changes now

tests/test_logging.py
```diff
@@ -178,59 +178,104 @@ def setup_logging_config(cli_options: Options, file_option_value: Optional[str]
...
         setup_logging_config(
-            cli_options=Options(),
+            cli_options=Options(**options3),
             file_option_value=path_logging_config_file2,
         )
         logger.info("test")
-        assert "EEE test" in stream.getvalue()
+        output = get_output_and_clear()
+        assert "EEE test" in output
+        assert f"{options3} options were ignored. Using logging config from file {path_logging_config_file1}" in output
```
arnaudsjs, line 245: This message is not in line with the comment above saying "The environment variable takes precedence".

tests/test_logging.py
```diff
@@ -178,59 +178,104 @@ def setup_logging_config(cli_options: Options, file_option_value: Optional[str]
...
+    old_log_file = os.path.join(tmpdir, "old_log_file.txt")
     path_logging_config_file1 = os.path.join(tmpdir, "logging_config1.yml")
     path_logging_config_file2 = os.path.join(tmpdir, "logging_config2.yml")
     path_logging_config_file3 = os.path.join(tmpdir, "logging_config3.yml")
 
     # Set --logging-config option on CLI only
     write_logging_config_file(path=path_logging_config_file1, formatter="AAA %(message)s")
     # Also assert that the other config options are ignored when logging_config is set.
-    setup_logging_config(cli_options=Options(logging_config=path_logging_config_file1, verbose=1), file_option_value=None)
+    options1 = {"log_file": old_log_file, "log_file_level": "4", "keep_logger_names": False}
```
arnaudsjs, line 197: I think we need to provide a better warning to the user than a dictionary of an internal data structure. Why not just show the actual options names, i.e. `--log-file`, `--log-file-level`, etc.

arnaudsjs (dismissed):

tests/test_logging.py
jptrindade: What do you mean? Don't we set the env variable above with:
`m.setenv("INMANTA_CONFIG_LOGGING_CONFIG", path_logging_config_file1)`

Hugo-Inmanta (approved):

src/inmanta/logging.py
```diff
@@ -788,6 +791,27 @@ def apply_options(self, options: Options, component: str | None = None, context:
...
+            Returns a string with the options (excluding "--verbose" and "--logging-config") that were set by the user
+            """
+            ignore_keys = ["verbose", "logging_config"]
+
+            option_to_cli = {
+                "log_file": "--log-file",
+                "log_file_level": "--log-file-level",
+                "timed": "--timed-logs",
+                "keep_logger_names": "--keep-logger-names",
+            }
```
jptrindade, line 806: @arnaudsjs I couldn't find a way to figure out the cli name of the options from the Namespace option so I figured I would just do this. What do you think? And should I move this to const since I use it on the test or do you think this is fine?

src/inmanta/logging.py
```diff
@@ -788,6 +791,27 @@ def apply_options(self, options: Options, component: str | None = None, context:
...
+            ignored_options = [
+                f"{option_to_cli[key]} {value}"
+                for key, value in options.__dict__.items()
+                if key not in ignore_keys and key in options
```
sanderr, line 810:
```suggestion
                for key, value in options.__dict__.items()
                if key not in ignore_keys
```
Isn't this redundant?

sanderr: Ah no, `options.__dict__` are all possible options while `options` are the defined ones?

src/inmanta/logging.py
wouterdb: Not a big fan of this, but it does work

wouterdb (approved):

src/inmanta/logging.py
```diff
@@ -788,6 +791,27 @@ def apply_options(self, options: Options, component: str | None = None, context:
...
+            ignore_keys = ["verbose", "logging_config"]
+
+            option_to_cli = {
+                "log_file": "--log-file",
+                "log_file_level": "--log-file-level",
+                "timed": "--timed-logs",
+                "keep_logger_names": "--keep-logger-names",
+            }
+            ignored_options = [
+                f"{option_to_cli[key]} {value}"
```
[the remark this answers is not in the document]
jptrindade: Not so trivial, because `timed` does not respect this

src/inmanta/logging.py
jptrindade: Yes, I'm reworking the logic a bit to be (in my opinion) simpler

src/inmanta/logging.py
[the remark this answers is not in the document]
jptrindade: `key.replace('_', '-') if key is not "timed" else "timed-logs"` also does not appeal that much to me.
The other option is to switch `timed` with `timed_logs` for the destination of the parser argument. What do you think @sanderr ?

src/inmanta/logging.py
```diff
@@ -796,18 +796,14 @@ def user_defined_options() -> Optional[str]:
...
                 "log_file": "--log-file",
                 "log_file_level": "--log-file-level",
                 "timed": "--timed-logs",
                 "keep_logger_names": "--keep-logger-names",
             }
             ignored_options = [
-                f"{option_to_cli[key]} {value}"
-                for key, value in options.__dict__.items()
-                if key not in ignore_keys and key in options
+                f"{value} {options.__getattribute__(key)}" for key, value in option_to_cli.items() if key in options
```
sanderr, line 806:
```suggestion
                f"{value} {getattr(options, key)}" for key, value in option_to_cli.items() if key in options
```
Does that work as well? It's more conventional.

sanderr (changes requested):

src/inmanta/logging.py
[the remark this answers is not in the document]
sanderr: I agree on all counts. I think your initial implementation is the least of evils here.

sanderr (approved):

Issue #8612, opened 2025-01-13
Fix skipped tests


arnaudsjs:
The following tests are currently skipped and need to be reconsidered. This ticket should not be picked up in its entirety by anyone, instead it should be picked up test-by-test to spread the load across the team. Edit the ticket and put your name in front of a test to indicate you are working on that specific test case. Once done, mark the checkbox.

- [x]  @wouterdb  tests/test_agent_manager.py::test_agent_actions -> replaced by other tests, to be removed, https://github.com/inmanta/inmanta-core/pull/8620
- [x] @arnaudsjs tests/test_deploy_command.py::test_deploy
- [x] @arnaudsjs tests/test_deploy_command.py::test_deploy_with_non_default_config
- [x] @arnaudsjs tests/deploy/e2e/test_deploy.py::test_deploy_with_undefined
- [x] @wouterdb tests/test_server.py::test_get_resource_actions -> https://github.com/inmanta/inmanta-core/pull/8623
- [x] @wouterdb tests/deploy/e2e/test_autostarted.py::test_stop_autostarted_agents_on_environment_removal -> https://github.com/inmanta/inmanta-core/pull/8624
- [x] @wouterdb tests/deploy/e2e/test_deploy_trigger.py::test_deploy_trigger ->https://github.com/inmanta/inmanta-core/pull/8636
- [x]  @wouterdb  tests/deploy/e2e/test_dryrun.py::test_dryrun_v2 
- [x]  @wouterdb tests/deploy/e2e/test_pause_and_halt.py::test_agent_stop_deploying_when_paused -> https://github.com/inmanta/inmanta-core/pull/8627
- [x] @wouterdb tests/deploy/e2e/test_pause_and_halt.py::test_agentinstance_stops_deploying_when_stopped -> covered by the above
- [x] @wouterdb  tests/server/test_agents_api.py::test_agent_list_filters -> https://github.com/inmanta/inmanta-core/pull/8631
- [x] @wouterdb   tests/server/test_databaseservice.py::test_agent_process_cleanup -> https://github.com/inmanta/inmanta-core/pull/8632
- [x] @arnaudsjs tests/server/test_environment_metrics.py::test_agent_count_metric
- [x] @jptrindade tests/test_data.py::test_get_current_resource_state https://github.com/inmanta/inmanta-core/pull/8620

# topics
- [x] dry-run and halting -> @wouterdb

PR #8626, opened 2025-01-14
Do not skip test_get_current_resource_state

jptrindade:
# Description

I believe loading std is no longer broken, right? This should be safe to revert

related to #8612 

# Self Check:

- [x] Changelog entry


jptrindade:
Hum... This was passing on my machine. Maybe it is one of those flaky tests

arnaudsjs (changes requested):
The test failed so there is still something wrong with the test case.

PR #8630, opened 2025-01-14
Issue/8444 renames full

sanderr:
# Description

Scheduler renames as proposed and discussed in #8613

Review request limited to the parts I've explicitly attached a comment to below. The rest is simply as proposed in #8613.

# Self Check:

- [ ] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -4475,7 +4475,7 @@ async def mark_as_orphan(
     async def update_resource_intent(
         cls,
         environment: uuid.UUID,
-        intent: dict[ResourceIdStr, tuple[state.ResourceState, state.ResourceDetails]],
+        intent: dict[ResourceIdStr, tuple[state.ResourceState, state.ResourceIntent]],
```
sanderr, line 4478: note to self: import as `deploy.ResourceIntent`?

src/inmanta/deploy/scheduler.py
```diff
@@ -922,20 +922,20 @@ async def _new_version(
...
                 itertools.chain.from_iterable(
                     self._state.requires.provides_view().get(resource, set()) for resource in force_new
                 )
             )
             for resource in resources_with_reset_requires:
                 if (
-                    # it is currently TRANSIENT blocked
-                    self._state.resource_state[resource].blocked is BlockedStatus.TRANSIENT
+                    # it is currently TEMPORARILY_BLOCKED blocked
+                    self._state.state[resource].blocked is Blocked.TEMPORARILY_BLOCKED
```
sanderr, line 934: note to self: revert state rename: it's all state really

tests/db/migration_tests/test_v202410310_to_v202411140.py
```diff
@@ -54,65 +54,65 @@ async def test_add_new_resource_status_column(
         resource_state_by_resource_id[inmanta.types.ResourceIdStr("test::Resource[agent1,key=key1]")],
         is_undefined=False,
         is_orphan=False,
-        deployment_result=state.DeploymentResult.DEPLOYED,
-        blocked_status=state.BlockedStatus.NO,
-        expected_compliance_status=ComplianceStatus.COMPLIANT,
+        last_deploy_result=state.DeployResult.DEPLOYED,
```
sanderr, line 57: note to self: mention latest version

src/inmanta/db/versions/v202501140.py
```diff
@@ -0,0 +1,39 @@
+"""
+    Copyright 2024 Inmanta
```
sanderr, line 2:
```suggestion
    Copyright 2025 Inmanta
```

src/inmanta/deploy/state.py
sanderr, line 1: note for reviewers: I decided not to go through with the `ModelState.resource_state -> ModelState.state` rename, because really the entire object already represents a state (model state), so I find it useful to disambiguate.

src/inmanta/db/versions/v202501140.py
sanderr, line 1: note to reviewers: I'd like a review on this, and the changes in `tests/db/migration_tests`

sanderr:
dump tool seems to be broken, I still need to look into that.

src/inmanta/deploy/scheduler.py
```diff
@@ -68,17 +68,17 @@ class StaleResource(Exception):
 
 
 @dataclass(frozen=True)
-class ResourceIntent:
+class ResourceVersionIntent:
```
wouterdb, line 71: So nice!

tests/db/migration_tests/test_v202411140_to_v202501140.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@pytest.mark.parametrize("no_agent", [True])
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), f"dumps/v{part}.sql"))
+async def test_add_new_resource_status_column(
+    postgresql_client: asyncpg.Connection,
+    migrate_db_from: abc.Callable[[], abc.Awaitable[None]],
+) -> None:
+    # only test that the migration script succeeds. Actual migration is covered by migration test for v202411140
+    await migrate_db_from()
```
wouterdb, line 37: Given this comment:
1. does the other test cover this sufficient? I would assume it does?  i.e. does the other test fail if you remove the upgrade code?
1. why run this at all, it takes time and adds no value? It doesn't even check if the scheduler could start. Make it an empty file with this comment in it?

arnaudsjs (approved):

tests/db/migration_tests/test_v202411140_to_v202501140.py
sanderr: 1. As far as I can tell, yes.
2. Fair enough, we need the file but not necessarily the migration.

sanderr: Fixed in 306f69e

wouterdb (approved):

PR #8633, opened 2025-01-15
Issue/fix plugins and cache typing

edvgui:
# Description

Fix typing of agent cache and plugins decorators.

# Self Check:

- [x] Changelog entry


changelogs/unreleased/fix-typing-of-decorators.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: Fix the typing of agent cache and plugins decorators.
+change-type: patch
+destination-branches: [master, iso7]
```
edvgui, line 4: Should it go into iso7 too?

src/inmanta/plugins.py
```diff
@@ -903,7 +922,7 @@ def wrapper(self, *args: object, **kwargs: object) -> Any:
             dictionary["__function__"] = fnc
 
             bases = (Plugin,)
-            fnc.__plugin__ = PluginMeta.__new__(PluginMeta, name, bases, dictionary)
+            fnc.__plugin__ = PluginMeta.__new__(PluginMeta, name, bases, dictionary)  # type: ignore[attr-defined]
```
edvgui, line 925: Should this rather be an entry in the mypy-baseline file?

changelogs/unreleased/fix-typing-of-decorators.yml
sanderr: I say yes

changelogs/unreleased/fix-typing-of-decorators.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: Fix the typing of agent cache and plugins decorators.
+change-type: patch
```
sanderr, line 3:
```suggestion
change-type: patch
sections:
  bugfix: "{{description}}"
```

src/inmanta/plugins.py
```diff
@@ -52,6 +52,7 @@
 from inmanta.warnings import InmantaWarning
 
 T = TypeVar("T")
+T_FUNC = TypeVar("T_FUNC", bound=Callable[..., Any])
```
sanderr, line 55:
```suggestion
T_FUNC = TypeVar("T_FUNC", bound=Callable[..., object])
```

I *think* this should work equally well, and it's more correct. I'm in favor of doing the same for the cache decorator, but less opinionated about that because it was already predefined there.

sanderr: If this doesn't work out, feel free to ignore.

src/inmanta/plugins.py
```diff
@@ -841,13 +842,31 @@ def type_string(self) -> str:
         return self.get_full_name()
 
 
+@typing.overload
+def plugin(
+    function: str | None = None,
```
sanderr, line 847: Nice, the change is even more meaningful here, with the `str | Callable | None`!

src/inmanta/plugins.py
sanderr: I think the ignore is appropriate in this case.

sanderr (commented):
One suggestion, but it's very much optional.

src/inmanta/plugins.py
edvgui: It looks to me like it works!

sanderr (approved):

PR #8637, opened 2025-01-15
Remove the inmanta deploy command.

arnaudsjs:
# Description

This PR removes the `inmanta deploy` command.

Part of #8612 

# Self Check:

- [x] Changelog entry


jptrindade (approved):

src/inmanta/deploy_project.py
```diff
@@ -1,506 +0,0 @@
...
-import argparse
-import logging
-import os
-import socket
-import subprocess
-import sys
-import time
-from typing import Optional
-
-from inmanta import config, const, module, postgresproc, protocol
```
wouterdb, line 28: I suspect `inmanta.postgresproc` can go as well

arnaudsjs: I see. Only the typing used there. And it's even an incorrect type.

wouterdb (approved):

Issue #8576, opened 2025-01-07
native types: warn for implict conversion to any


wouterdb:
When we can't figure out the inmanta type, we revert to `any` 

In this case, we want to produce a warning, using the compilers warning framework. 
(this will allow us to configure it as failure as well)

Hugo-Inmanta:
Link to design draft https://github.com/inmanta/designs/blob/4f962f3f36ac4dea06ad60b29a88a29c27a3ea1a/plugin-typing-v2.md

PR #8644, opened 2025-01-16
Issue/8576 native types warn for implicit conversion to any

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/8576

# Self Check:

- [x] Changelog entry


tests/compiler/test_plugin_types.py
```diff
@@ -64,3 +67,13 @@ class CustomDict[K, V](Mapping[K, V]):
 
     with pytest.raises(RuntimeException):
         to_dsl_type(CustomDict[str, str])
+
+    # Check that a warning is produced when implicit cast to 'Any'
+    caplog.clear()
+    to_dsl_type(complex)
```
jptrindade, line 73: what is this `complex` value?

Hugo-Inmanta: It is the complex numeric type https://docs.python.org/3/library/stdtypes.html#typesnumeric.
I needed a python type that has no equivalent in the dsl

sanderr (approved):

jptrindade (approved):

PR #8646, opened 2025-01-16
mypy fix

sanderr:
~~Looks like `@handle` wasn't properly defined as a generic method, which caused incorrect binding for the type var.~~

We don't allow `None` in `@handle` method return values. This is raising mypy errors on inmanta-lsm. (`src/inmanta_lsm/service_inventory.py:1606: error: Value of type variable "FuncT" of "__call__" of "handle" cannot be "Callable[[ServiceInventorySlice, Environment, str, UUID, bool], Coroutine[Any, Any, dict[str, bool | None]]]"  [type-var]`)


sanderr:
This fix is not good. I'll request review again once I'm there.

wouterdb (approved):

PR #8654, opened 2025-01-17
Added support for union types to plugins

### LINKED ISSUE #8574: native types: Add support for Union types
Translate python unions into inmanta unions. 

inmanta unions are not very well used, so make sure to test them well

arnaudsjs:
# Description

Added support for union types to plugins.

closes #8574

# Self Check:

- [x] Changelog entry


tests/compiler/test_plugins.py
```diff
@@ -435,6 +442,75 @@ def test_native_types(snippetcompiler: "SnippetCompilationTest") -> None:
...
+for val in ["test", 123, null]:
+    plugin_native_types::union_return_optional_1(value=val)     # type return value: Union[None, int, str]
+    plugin_native_types::union_return_optional_2(value=val)     # type return value: Optional[Union[int, str]]
+    plugin_native_types::union_return_optional_3(value=val)     # type return value: Union[int, str] | None
+    plugin_native_types::union_return_optional_4(value=val)     # type return value: None | Union[int, str]
+end
         """
     )
     compiler.do_compile()
+
```
arnaudsjs, line 470: These different scenario's result in different types of exceptions (`RuntimeException`, `WrappingRuntimeException` and `TypingException`). Is that desired?

tests/compiler/test_plugins.py
```diff
@@ -435,6 +442,75 @@ def test_native_types(snippetcompiler: "SnippetCompilationTest") -> None:
...
+    # Parameter to plugin has incompatible type
+    for plugin_name, plugin_value, error_message in [
+        ("union_single_type", 123, "Invalid value '123', expected string"),
+        ("union_multiple_types", "[1, 2, 3]", "Invalid value '[1, 2, 3]', expected Union[int,string]"),
+        ("union_optional_1", 1.2, "Invalid value '1.2', expected Union[int,string]"),
+        ("union_optional_2", 1.2, "Invalid value '1.2', expected Union[int,string]"),
+        ("union_optional_3", 1.2, "Invalid value '1.2', expected Union[int,string]"),
+        ("union_optional_4", 1.2, "Invalid value '1.2', expected Union[int,string]"),
```
arnaudsjs, line 478: These exception messages are in-line with what we already had, but I am wondering whether they are any good. It reports about an invalid subtype. For example it reports `Union[int, string]`, but shouldn't it report `Union[int, string]?` instead? Because the top-level type is optional. And wouldn't it be better to report the python type instead of the Inmanta type?

src/inmanta/ast/type.py
```diff
@@ -614,6 +612,11 @@ def validate(self, value: object) -> bool:
     def type_string_internal(self) -> str:
         return "Union[%s]" % ",".join(t.type_string_internal() for t in self.types)
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Union):
+            return NotImplemented
+        return self.types == other.types
```
sanderr, line 618: This only covers the case where the types are defined in the same order. And we don't flatten nested unions. Then again, sorting them is probably non-trivial, let alone the flattening.

I have to check what this is used for again to see if this is acceptable.

src/inmanta/ast/type.py
```diff
@@ -20,9 +20,7 @@
 import functools
 import numbers
 from collections.abc import Sequence
-from typing import Callable
-from typing import List as PythonList
-from typing import Optional
+from typing import Callable, Optional
```
sanderr, line 23: Callable should be from `collections.abc`.

tests/compiler/test_plugins.py
jptrindade: I agree with both these statements

jptrindade (approved):
I agree with the changes that you mentioned, don't know if we should tackle them into another ticket or if they are easy fixes

src/inmanta/plugins.py
```diff
@@ -262,20 +262,19 @@ def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
     # Unions and optionals
     if typing_inspect.is_union_type(python_type):
         # Optional type
+        bases: Sequence[inmanta_type.Type]
         if typing_inspect.is_optional_type(python_type):
             other_types = [tt for tt in typing.get_args(python_type) if not typing_inspect.is_optional_type(tt)]
```
sanderr, line 267: I believe much of this, and especially this list comprehension, won't work with `Annotated.
```python
>>> t = typing.Annotated[int | None, "my-annotation"]
>>> typing_inspect.is_optional_type(t)
False
>>> typing_inspect.is_union_type(t)
False
```

I don't think it has to be a concern for you, but it's something to take into account for @jptrindade's PR.

src/inmanta/plugins.py
```diff
@@ -380,6 +379,9 @@ def resolve_type(self, plugin: "Plugin", resolver: Namespace) -> inmanta_type.Ty
             return self._resolved_type
 
         if not isinstance(self.type_expression, str):
+            if typing_inspect.is_union_type(self.type_expression) and not typing.get_args(self.type_expression):
+                # If typing.Union is not subscripted, isinstance(self.type_expression, type) evaluates to False.
+                raise TypingException(None, f"Union type must be subscripted, got {self.type_expression}")
```
sanderr, line 384: I'm hesitant about the exception type. The one raised further down is a `RuntimeException`. I think they should at least be consistent.

sanderr: I think it's fine to use this one for both, but @wouterdb probably knows best.

src/inmanta/ast/type.py
sanderr: And `NullableType` vs `Union` with null. But that might just be something we have to clean up sometime, rather than hack around now.

sanderr: @wouterdb I have the impressino that the `__eq__` implementation isn't all that important, and mainly used for testing. Can you confirm?

tests/compiler/test_plugin_types.py
```diff
@@ -43,6 +43,22 @@ def test_conversion():
 
     assert inmanta_type.TypedDict(inmanta_type.String()) == to_dsl_type(collections.abc.Mapping[str, str])
 
+    # Union types
```
sanderr, line 46: Could you add a nested union? e.g. `Union[int, Union[str, float]]`, and perhaps also one with a `None` in the deepest union (`Union[int, Union[str, float None]]`).

tests/compiler/test_plugins.py
sanderr: `WrappingRuntimeException` is expected for sure. I already have another comment about the runtime vs typing exception. I'm a bit surprised though that they aren't all wrapped.

tests/compiler/test_plugins.py
sanderr: > And wouldn't it be better to report the python type instead of the Inmanta type?

Tough one. On the one hand, it's more expressive and the plugin lives in the Python domain. On the other hand, the call originates from the DSL domain. I'm on the fence, leaning towards agreeing with you. @wouterdb wdyt?

sanderr: > but shouldn't it report Union[int, string]? instead?

I agree that would be better. Perhaps we can simply validate, and when it fails, construct a new exception with the Python type, instead of propagating the one raised by the DSL validation logic?

sanderr (changes requested):

src/inmanta/ast/type.py
wouterdb: That is correct

src/inmanta/plugins.py
```diff
@@ -399,8 +400,8 @@ def resolve_type(self, plugin: "Plugin", resolver: Namespace) -> inmanta_type.Ty
     def validate(self, value: object) -> bool:
         """
         Validate that the given value can be passed to this argument.  Returns True if the value is known
-        and valid, False if the value is unknown, and raises a ValueError is the value is not of the
-        expected type.
+        and valid, False if the value is unknown, and raises a :py:class:`inmanta.ast.RuntimeException`
+        if the value is not of the expected type.
```
sanderr, line 404: Good fix. I was surprised to see this, I'd have sworn `Type.validate` raised a `ValueError`, but looks like I was wrong.

src/inmanta/plugins.py
```diff
@@ -760,8 +761,18 @@ def check_args(self, args: Sequence[object], kwargs: Mapping[str, object]) -> bo
             arg = self.get_arg(position)
 
             # (4) Validate the input value
-            if not arg.validate(value):
-                return False
+            try:
+                no_unknowns = arg.validate(value)
+            except RuntimeException as e:
+                raise InvalidTypeException(
```
sanderr, line 767: Nice! Took me a while to process for some reason, but it's a good addition.

src/inmanta/plugins.py
```diff
@@ -381,7 +382,7 @@ def resolve_type(self, plugin: "Plugin", resolver: Namespace) -> inmanta_type.Ty
         if not isinstance(self.type_expression, str):
             if typing_inspect.is_union_type(self.type_expression) and not typing.get_args(self.type_expression):
                 # If typing.Union is not subscripted, isinstance(self.type_expression, type) evaluates to False.
-                raise TypingException(None, f"Union type must be subscripted, got {self.type_expression}")
+                raise RuntimeException(stmt=None, msg=f"Union type must be subscripted, got {self.type_expression}")
```
sanderr, line 385: I think this was changed because of something I said, but looking back at it, I wonder if it should have been changed the other way around? This is the normalization stage. I don't think we should raise `RuntimeException` here (same for a few lines down). That said, we do so in some other places as well. @wouterdb I think you have a better view on what these exceptions are intended for?

wouterdb: I would prefer a subtype of typingexception

wouterdb (approved):

sanderr:
@wouterdb some final non-trivial fixes since your review

sanderr (approved):

wouterdb (approved):

sanderr:
iso7 part in #8677

PR #8662, opened 2025-01-20
Add references support

bartv:
# Description

Add references support (to the dataclasses branch)


src/inmanta/ast/type.py
```diff
@@ -334,7 +334,8 @@ def validate(self, value: Optional[object]) -> bool:
         """
         if isinstance(value, AnyType):
             return True
-
+        if references.is_reference_of(value, float):
+            return True
```
wouterdb, line 338: This is the main open issue I think

src/inmanta/export.py
```diff
@@ -508,12 +509,23 @@ def deploy_code(self, conn: protocol.SyncClient, tid: uuid.UUID, version: Option
...
         # Load both resource definition and handlers
         for type_name, resource_definition in resource.get_resources():
             code_manager.register_code(type_name, resource_definition)
+            types.add(type_name)
 
         for type_name, handler_definition in Commander.get_providers():
             code_manager.register_code(type_name, handler_definition)
+            types.add(type_name)
+
+        # Register all reference and mutator code to all resources. This is very coarse grained and can be optimized once
```
wouterdb, line 523: it register everything that is loaded, even if not in use. There is something to say for that, but it is very greedy.

src/inmanta/references.py
```diff
@@ -0,0 +1,450 @@
...
+    """A reference"""
+
+    id: uuid.UUID
+
+
+class MutatorModel(BaseModel):
+    """A mutator"""
+
+
+C = typing.TypeVar("C", bound="Base")
```
wouterdb, line 156: As I understand, this defines the serialized form of the reference

src/inmanta/references.py
```diff
@@ -0,0 +1,450 @@
...
+
+
+class Base:
+    """A base class for references and mutators"""
+
+    type: typing.ClassVar[ReferenceType]
+
+    def __init__(self, **kwargs: object) -> None:
+        self._arguments: typing.Mapping[str, object] = kwargs
```
wouterdb, line 165: I wonder if it wouldn't make more sense to e.g. pick up all values from `__dict__` that don't start with `_` 

I.e. this requires all arguments to be passed down, while this information is already available?

src/inmanta/references.py
```diff
@@ -0,0 +1,450 @@
...
+            param_type = parameters.get(name)
+            match name, value, param_type:
+                case _, str() | int() | float() | bool(), _:
+                    arguments.append(LiteralArgument(name=name, value=value))
+
+                case _, Reference(), _:
+                    model = value.serialize()
+                    arguments.append(ReferenceArgument(name=name, id=model.id))
+
+                case "resource", _, inmanta.resources.Resource:
```
wouterdb, line 201: Does the third argument match in case of subtyping?

src/inmanta/references.py
```diff
@@ -0,0 +1,450 @@
...
+        :param ref: The model of the reference to deserialize
+        :param resource: The resource to use as a read-only reference
+        """
+        return cls(**{arg.name: arg.get_arg_value(resource) for arg in ref.args})
+
+    @abc.abstractmethod
+    def serialize(self) -> BaseModel:
+        """Serialize to be able to add them to a resource"""
+
+    def serialize_arguments(self) -> Tuple[uuid.UUID, list[ArgumentTypes]]:
```
wouterdb, line 187: This will need some more error reporting in a later stage I think

src/inmanta/references.py
```diff
@@ -0,0 +1,450 @@
...
+                case "resource", _, inmanta.resources.Resource:
+                    arguments.append(ResourceArgument(name=name))
+
+                case _, type(), _ if value in [str, float, int, bool]:
+                    arguments.append(PythonTypeArgument(name=name, value=value.__name__))
+
+                case _:
+                    raise TypeError(f"Unable to serialize argument `{name}` of `{self}`")
+
+        data = json.dumps({"type": self.type, "args": arguments}, default=util.api_boundary_json_encoder, sort_keys=True)
```
wouterdb, line 210: may not be entirely stable, because the argument could be re-ordered when manipulating` _arguments`

src/inmanta/references.py
```diff
@@ -0,0 +1,450 @@
...
+        """Emit the correct pydantic objects to serialize the reference in the exporter."""
+        if not self._model:
+            arg_id, arguments = self.serialize_arguments()
+            self._model = MutatorModel(type=self.type, args=arguments)
+
+        assert isinstance(self._model, MutatorModel)
+        return self._model
+
+
+class DataclassRefeferenMeta(type):
```
wouterdb, line 237:
```suggestion
class DataclassReferenceMeta(type):
```

src/inmanta/resources.py
```diff
@@ -292,7 +386,12 @@ def object_to_id(
 
     @classmethod
     def map_field(
-        cls, exporter: Optional["export.Exporter"], entity_name: str, field_name: str, model_object: "proxy.DynamicProxy"
+        cls,
+        exporter: Optional["export.Exporter"],
```
wouterdb, line 390: I would expect the exporter object exposed to every export function to enabled the resource to register a reference even if it has no mutator (for plugins that have built-in, not transparent support)

src/inmanta/resources.py
```diff
@@ -388,21 +518,61 @@ def __init__(self, _id: "Id") -> None:
...
+        return self._references[id].get()
+
+    def resolve_all_references(self) -> None:
+        """Resolve all value references"""
+        for ref in self.references:  # type: ignore
+            model = references.ReferenceModel(**ref)
+            self._references_model[model.id] = model
+
+        for mutator in self.mutators:  # type: ignore
+            mutator = references.mutator.get_class(mutator["type"]).deserialize(references.MutatorModel(**mutator), self)
```
wouterdb, line 559: We may want to define a base class resource to signal serialization failure , just to guide developer towards cleaner exceptions

wouterdb (commented):
First pass

src/inmanta/const.py
```diff
@@ -385,11 +386,13 @@ class NotificationSeverity(str, Enum):
...
-
 # resource attributes for event propagation
-RESOURCE_ATTRIBUTE_SEND_EVENTS = "send_event"
-RESOURCE_ATTRIBUTE_RECEIVE_EVENTS = "receive_events"
+RESOURCE_ATTRIBUTE_SEND_EVENTS: typing.Final[str] = "send_event"
+RESOURCE_ATTRIBUTE_RECEIVE_EVENTS: typing.Final[str] = "receive_events"
 
+# resource attributes for references
+RESOURCE_ATTRIBUTE_REFERENCES: typing.Final[str] = "references"
+RESOURCE_ATTRIBUTE_MUTATORS: typing.Final[str] = "mutators"
```
sanderr, line 395: These should not be exposed in the model, should they?

src/inmanta/plugins.py
sanderr, line 1: We'd need to check if references are allowed somewhere in the arg validation.

src/inmanta/ast/type.py
sanderr, line 1: Open question: semantically, do we consider references to be a reference to a Python value, that can then be converted to an inmanta value by returning it from a plugin-like construct. Or do we consider them to be dual domain in the sense that in the model they are references to inmanta values?

This could impact how we represent them here:
1. if they're just references to Python values, they don't need representation in the inmanta type system, we only need to validate them as is done here.
2. if they have a counterpart in the model, we may need to represent them in the inmanta type system (internal, not in the DSL), meaning that validation logic here would check not if something is a reference to `float`, but rather to `type.Float`.

Overall, I think option 1 is simpler, and at the moment I don't see any strong motivations to go for option 2, since references are *always* resolved in the Python domain anyway. But I do think we need to make a conscious decision here.

src/inmanta/ast/type.py
sanderr: Do I understand correctly that the issue is that we don't always want to allow references, so this is too accepting?

We could have a `Reference(value_type=Type)` type, who's validate checks exactly this. Then validate callers could always be explicit in whether they want to allow references? I'd have to double check which callers exist, but I expect that for each of them being explicit would be both feasible and more clear.

Or perhaps a `validate_reference` method on `Type`. Would have the advantage that we can rely on polymorphism for the `Float -> float` mapping etc. But it's less flexible in the sense that adding support for more special cases (`Unknown` etc) requires a special method for each.

src/inmanta/ast/type.py
sanderr: Potential difficulty with 1: what about references to custom types, e.g. a typedef? If we represent the reference only in the inmanta domain, it might be a reference to an int, while the DSL might expect a reference to `myint`. If on the other hand we represent it internally as a reference to `myint` (which we can through the plugin annotations, i.e. `Annotated[int, ModelType["myint"]]`, then validation becomes trivial, and mostly we gain some confidence over simply accepting a reference to *any* int. This confidence may be very important, given that we only have the model to do type validation on.

src/inmanta/execute/util.py
```diff
@@ -42,6 +42,14 @@ def __iter__(self) -> Iterable[object]:
         return iter([])
 
 
+@stable_api
+class Reference(AnyType):
+    """
+    An instance of this class is used to indicate that this value is a reference and can only be resolved
+    at execution time in agent.
```
sanderr, line 49: > can only be resolved at execution time in agent.

Not entirely true iirc, though probably so in 90% of the use cases.

sanderr (commented):
intermediate comments

src/inmanta/util/dict_path.py
```diff
@@ -271,8 +271,23 @@ def get_path_sections(self) -> Sequence["WildDictPath"]:
...
+        TODO: fix the typeguard. Probably by providing multiple methods for validation!
+        """
+        if not hasattr(container, "__getitem__"):
+            return False
+
+        if set and not hasattr(container, "__setitem__"):
+            return False
+
+        if remove and not hasattr(container, "__delitem__"):
+            return False
```
sanderr, line 288: I'd prefer `isinstance` checks on the ABCs here, but iirc there' a separate ticket to implement this properly.

src/inmanta/references.py
sanderr, line 201: If we special-case this, we should make sure to document that.

src/inmanta/references.py
```diff
@@ -0,0 +1,450 @@
...
+    """This decorator register a reference under a specific name"""
+
+    # It is not allowed to use T in a class var so we cannot use T here
+    _reference_classes: typing.ClassVar[dict[str, type[Reference[RefValue]]]] = {}
+
+    def __init__(self, name: str) -> None:
+        """
+        :param name: This name is used to indicate the type of the reference
+        """
+        self.name = name
```
sanderr, line 327: I think this will need to be changed somewhat with what we discussed earlier today. i.e. we'll want to accept a Python type and convert it with `to_dsl_type`. Would be even nicer if we can derive the Python type from the type subscription (i.e. the value for the generic type var) of the `Reference`, but I'm not sure how feasible that would be tbh.

sanderr (commented):
More comments.

Note to self: to review: `references.py` and `resources.py`.

src/inmanta/ast/type.py
sanderr: Discussed and concluded in person

src/inmanta/references.py
```diff
@@ -0,0 +1,450 @@
...
+        attribute_name: str,
+        attribute_type: typing.Type[T],
+    ) -> None:
+        super().__init__(reference=reference, attribute_name=attribute_name, attribute_type=attribute_type)
+        self.attribute_name = attribute_name
+        self.attribute_type = attribute_type
+        self.reference = reference
+
+    def resolve(self) -> T:
+        return typing.cast(T, getattr(self.reference, self.attribute_name))
```
sanderr, line 412: Since we transparently skip what would otherwise be model validation (e.g. when assigned to a resource's attribute, should we add some validation here?

sanderr (commented):
LGTM overall, added some comments.

PR #8665, opened 2025-01-20
Add support for Annotated types in plugins

### LINKED ISSUE #8573: native types: annotated types for allow explicit inmanta types
Allow the following:

```python
typing.Annotated[Mapping[str, int], InmantaType("dict")]
```

see also: https://github.com/inmanta/inmanta-core/pull/8433
ensure the rest of the code (e.g. handling of unions) can correctly handle annotated types

We don't validate the correspondence between both types...

jptrindade:
# Description

Introduce Annotated types in plugins

closes #8573 

# Self Check:

- [x] Changelog entry


tests/compiler/test_plugin_types.py
```diff
@@ -17,44 +17,71 @@
...
+    assert inmanta_type.TypedDict(inmanta_type.Type()) == to_dsl_type(
+        Annotated[dict[str, int], plugin_typing.InmantaType("dict")], location, namespace
+    )
+
+    # FIXME: Not working
+    # entity: Entity = Entity("my-entity", namespace)
+    # namespace.define_type("my-entity", entity)
+    # assert inmanta_type.TypedDict(inmanta_type.Type()) == to_dsl_type(
+    #     Annotated[Union[int, str] | None, plugin_typing.InmantaType("my-entity")], location, namespace
+    # )
```
jptrindade, line 59: It is a bit awkward setting up the namespace in order to do this. Do you have any tips @sanderr ?

jptrindade:
@sanderr What happens if we can't find the InmantaType? Do we default to `Any`?

src/inmanta/plugin_typing.py
jptrindade, line 1: I copied this file from your PR @sanderr . I'm inclined to leave it here. Regarding the comment on how to handle `std::Entity` I also think `object` might be our best option

src/inmanta/plugin_typing.py
```diff
@@ -0,0 +1,42 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    Contact: code@inmanta.com
+"""
+
+import typing
+from dataclasses import dataclass
+
+# TODO: move this module to inmanta.plugins.typing? Probably not because that would import the whole `plugins` namespace?
```
sanderr, line 22: I'd put it in `plugins` or `plugins.typing`. I don't recall why I was hesitant to put it there tbh. It's for annotating plugins, users would need to import it either way.

I think I'd prefer `plugins.typing`. @wouterdb may have an opinion.

src/inmanta/plugin_typing.py
```diff
@@ -0,0 +1,42 @@
...
+    For maximum static type coverage, it is recommended to use these only when absolutely necessary, and to use them as deeply
+    in the type as possible, e.g. prefer `Sequence[Annotated[MyEntity, InmantaType("std::Entity")]]` over
+    `Annotated[Sequence[Entity], InmantaType("std::Entity[]")]`.
+    """
+
+    dsl_type: str
+
+
+# TODO: how to do Entity? "object" is appropriate but raises too many errors for practical use. Any is Any
+Entity: typing.TypeAlias = typing.Annotated[object, InmantaType("std::Entity")]
```
sanderr, line 41: This needs a docstring

src/inmanta/plugin_typing.py
```diff
@@ -0,0 +1,42 @@
...
+string: typing.TypeAlias = typing.Annotated[str, InmantaType("string")]
```
sanderr, line 42: I don't think this is useful anymore, considering that we support the native type now.

src/inmanta/plugin_typing.py
```diff
@@ -0,0 +1,42 @@
...
+@dataclass(frozen=True)
+class InmantaType:
+    """
+    Declaration of an inmanta type for use with typing.Annotated.
+    When a plugin type is declared as typing.Annotated with an `InmantaType` as annotation, the Python type is completely
+    ignored for type validation and conversion to and from the DSL. Instead the string provided to the `InmantaType` is
+    evaluated as a DSL type, extended with "any".
+    For maximum static type coverage, it is recommended to use these only when absolutely necessary, and to use them as deeply
+    in the type as possible, e.g. prefer `Sequence[Annotated[MyEntity, InmantaType("std::Entity")]]` over
+    `Annotated[Sequence[Entity], InmantaType("std::Entity[]")]`.
```
sanderr, line 34:
```suggestion
    `Annotated[Sequence[MyEntity], InmantaType("std::Entity[]")]`.
```

src/inmanta/plugin_typing.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+
+    dsl_type: str
+
+
+# TODO: how to do Entity? "object" is appropriate but raises too many errors for practical use. Any is Any
```
sanderr, line 40: We'll have to run this by Bart. I prefer `object`, as you propose, but I tend to be more strict than others.

src/inmanta/plugins.py
```diff
@@ -31,8 +31,8 @@
 import typing_inspect
 
 import inmanta.ast.type as inmanta_type
-from inmanta import const, protocol, util
-from inmanta.ast import (  # noqa: F401 Plugin exception is part of the stable api
+from inmanta import const, plugin_typing, protocol, util
+from inmanta.ast import (  # noqa: F401 Plugin e
```
sanderr, line 35: This looks like an accidental change to me.

src/inmanta/plugin_typing.py
sanderr: `plugins.typing` would have the added difficulty that it's less trivial to import partially qualified.

src/inmanta/plugins.py
```diff
@@ -306,27 +313,26 @@ def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
...
-    #     [arg if isinstance(arg, plugin_typing.InmantaType) for arg in args]
-    #     if inmanta_types:
-    #         if len(inmanta_types) > 1:
-    #             # TODO
-    #             raise Exception()
-    #         # TODO
-    #         return parse_dsl_type(inmanta_types[0].dsl_type)
-    #     # the annotation doesn't concern us => use base type
-    #     return to_dsl_type(args[0])
+    if typing.get_origin(python_type) is typing.Annotated:
```
sanderr, line 324: I'd indent this one level under the `is_generic` body. I don't think it's strictly necessary, and there's something to say for not nesting deeper than required, but I think that in this case consistency is more important.

sanderr: And then you can use `origin` instead of `get_origin`

src/inmanta/plugins.py
```diff
@@ -306,27 +313,26 @@ def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
...
+        annotated_args = typing.get_args(python_type)
+        inmanta_types: Sequence[plugin_typing.InmantaType] = [
+            arg for arg in annotated_args if isinstance(arg, plugin_typing.InmantaType)
+        ]
+        if inmanta_types:
+            if len(inmanta_types) > 1:
+                raise TypingException(None, f"invalid type {python_type}, only one InmantaType annotation is supported")
+            return parse_dsl_type(inmanta_types[0].dsl_type, location, resolver)
+        # the annotation doesn't concern us => use base type
+        return to_dsl_type(annotated_args[0], location, resolver)
```
sanderr, line 334: > @sanderr What happens if we can't find the InmantaType? Do we default to Any?

If it's not there, it simply means that the `Annotated` has nothing to do with us. e.g. `typing.Annotated[str, ContextForSomeMypyPlugin()]`. So as far as our type-checking is concerned, it's just `str`. That's why we fall back to the base type here.

changelogs/unreleased/8573-add-support-for-annotated-types.yml
```diff
@@ -0,0 +1,4 @@
+description: Add support for annotated types to plugins.
+issue-nr: 8573
+change-type: minor
+destination-branches: [master, iso8]
```
sanderr, line 4: We probably want to include this in iso7 as well. Discussion still pending, see [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1737472131458259).

tests/data/modules/plugin_native_types/plugins/__init__.py
```diff
@@ -32,3 +34,16 @@ def many_arguments(il: list[str], idx: int) -> str:
 @plugin
 def as_none(value: str) -> None:
     pass
+
+
+# Annotated values
+
+
+@plugin
+def annotated_arg_entity(value: Annotated[object, plugin_typing.InmantaType("TestEntity")]) -> None:
```
sanderr, line 43: Could you also add one that uses a `typedef` as its dsl type? e.g. a toy example of a valid use case might be

DSL: `typedef response as string matching self in ["yes", "no"]`
Python:
```python
@plugin
def process_response(response: Annotated[Literal["yes", "no"], InmantaType("response")]) -> None ...
```

sanderr (changes requested):

src/inmanta/plugins.py
jptrindade: I don't think that is the case.
If I have `Annotated[str, "something"]` it will default to string.
But if I have `Annotated[str, plugin_typing.InmantaType("unexistent-entity")]` It should fail on `parse_dsl_type`.

sanderr: Oh, I misunderstood the question. I'd make sure to treat it the same as any other DSL-type plugin annotation. I would expect that means to raise an exception. i.e. what if you do `def my_plugin(arg: "not-a-valid-type") -> None: ...`?

src/inmanta/plugin_typing.py
jptrindade: The only thing awkward I see is the import confusion between `typing` and `inmanta.plugins.typing` in plugins.py.
I fixed this by importing everything we use instead of just doing `import typing`

sanderr: I don't think that is a proper solution. qualified / unqualified should be driven by context to make sure that everything is unambiguous. Sure, things like `Any` and `Sequence` and such (though the latter should really come from `collections.abc`) are sufficiently wide-spread that they are unmabiguous in their own right. But `get_args` etc, those really require context. Imo we must use them qualified (qualifiedly?).

As to the awkwardness, I agree on that front. What I meant to say is that `from inmanta.plugins import typing` will never be a good approach. That leaves `import inmanta.plugins.typing`, which is very verbose, and `import inmanta.plugins.typing; from inmanta import plugins` (allowing partially qualified `plugins.typing.InmantaType`), which is a bit exotic, but we do it in a few places, and I don't think there's anything wrong with it myself.

So the question becomes, do we consider this import awkwardness sufficient reason to just add it to the heap of stuff in `inmanta.plugins`? I honestly don't know.

changelogs/unreleased/8574-add-support-native-union-types.yml
sanderr, line 1: Something weird seems to have happened here. Could you merge in master? That might make it all a lot easier to review.

sanderr: Oh, nevermind, I see that I haven't merged the union PR yet. I went over all my PRs in the morning, but this is not officially mine so I missed it. I'll review like this, but it'll be for tomorrow, I didn't get where I'd hoped today.

tests/compiler/test_plugins.py
```diff
@@ -536,32 +553,50 @@ def test_native_types(snippetcompiler: "SnippetCompilationTest", caplog) -> None
             "union_optional_1",
             1.2,
             f"Value 1.2 for argument value of plugin {ns}::union_optional_1 has incompatible type."
-            f" Expected type: Union[int,string]?",
+            f" Expected type: Union[int,string,Type]?",
```
jptrindade, line 556: What do you think about this? What would be the appropriate type to display here?

sanderr: I'd say `any`. @wouterdb added the `Any` type in his PR. We should probably start using that one instead of `Type` (I thought he had already made the change, but perhaps not).

jptrindade: I changed it to the actual inmanta type. It should be `std::Entity` now

src/inmanta/ast/entity.py
```diff
@@ -609,7 +610,8 @@ def pair_dataclass(self) -> None:
                     dc_fields.pop(rel_or_attr_name)
                     # Type correspondence
                     try:
-                        dsl_type = to_dsl_type(dc_types[rel_or_attr_name])
+                        plugin_line: Range = Range(self.location.file, self.location.lnr, 1, self.location.lnr + 1, 1)
```
sanderr, line 613: This name is confusing, because this is no plugin. That said, I think we can simply use `self.location` here. It will always be a `Range` in practice (locations are badly typed, little we can do about it, you can change the signature of `get_location()` here, which would help, but it would probably lead to a type error in its implementation, which is acceptable for now).

src/inmanta/plugins.py
```diff
@@ -313,37 +341,37 @@ def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
...
-    #         if len(inmanta_types) > 1:
-    #             # TODO
-    #             raise Exception()
-    #         # TODO
-    #         return parse_dsl_type(inmanta_types[0].dsl_type)
-    #     # the annotation doesn't concern us => use base type
-    #     return to_dsl_type(args[0])
+        # Annotated
+        if origin is typing.Annotated:
+            for meta in python_type.__metadata__:  # type: ignore
```
sanderr, line 364: Didn't we conclude that we wanted the rightmost, i.e. "freshest" `ModelType`?
```suggestion
            for meta in reversed(python_type.__metadata__):  # type: ignore
```

src/inmanta/plugins.py
```diff
@@ -313,37 +341,37 @@ def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
...
+                if isinstance(meta, ModelType):
+                    if location is None or resolver is None:
+                        raise TypingException(
+                            None,
+                            f"Cannot parse {meta.model_type}, location or resolver were not provided to `to_dsl_type`",
+                        )
```
sanderr, line 370: Isn't this always a bug? Meaning, why do we even accept `None`?

sanderr (changes requested):

src/inmanta/plugins.py
jptrindade: I changed it to allow None values mainly because of the dataclass PR. `DataclassExplainer` calls `to_dsl_type` and I was struggling to get the `location` and `namespace` there. And if I did get them from the `NamedType` they could still be None. And we only ever need this for Annotated types. So  in any other use case that we don't need them, we don't need to provide them.

Now that I'm writing (rubber-ducking this) I feel like we should indeed not accept None, my question is how do we guarantee that we always have the `namespace` and `location` set on `DataclassExplainer`?

src/inmanta/plugins.py
```diff
@@ -263,12 +264,39 @@ def get_location(self) -> Optional[Location]:
...
-def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
+@dataclass(frozen=True)
+class ModelType:
+    model_type: str
+
+    def __class_getitem__(cls: type[Self], key: str) -> Self:
+        return cls(key)
+
+
+Entity: typing.TypeAlias = typing.Annotated[Any, ModelType["std::Entity"]]
```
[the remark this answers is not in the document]
jptrindade: I'll leave the protocols off for now and create a ticket for it

src/inmanta/plugins.py
```diff
@@ -263,12 +264,39 @@ def get_location(self) -> Optional[Location]:
...
     """
     Convert a python type annotation to an Inmanta DSL type annotation.
 
     :param python_type: The evaluated python type as provided in the Python type annotation.
+    :param location: The location of this evaluation on the model
+    :param resolver: The namespace that can be used to resolve the type annotation of this argument.
     """
+    # Resolve aliases
+    if isinstance(python_type, typing.TypeAliasType):
+        return to_dsl_type(python_type.__value__, location, resolver)
```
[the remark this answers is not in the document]
jptrindade: I addded a testcase for it now

src/inmanta/ast/entity.py
jptrindade: Can't I just rename `plugin_line` to something like `location` (which would still be a range)? It feels wrong explicitly introducing typing errors

sanderr: I completely understand, but I think we have to consider the source of the type error. If it is due to a pre-existing issue, we're only introducing a new report for it, we're not really introducing a new issue. I think that context is really important when doing gradual typing, as (alas) we are doing. The type error does not make matters worse than they already are, it's just about accepting an unfortunate fact. In contrast, creating a new range here might in fact make matters worse, because we lose useful information: the range might point to a specific character, and we'd be widening that to a whole line.

src/inmanta/ast/type.py
```diff
@@ -250,8 +250,11 @@ def as_python_type_string(self) -> "str | None":
     def has_custom_to_python(self) -> bool:
         return False
 
+    def type_string(self) -> str:
+        return "any"
```
sanderr, line 254: This is not right. This method must really only be implemented for types that can be expressed directly in the DSL (this one only exists on the plugin boundary).

src/inmanta/plugins.py
```diff
@@ -266,18 +266,35 @@ def get_location(self) -> Optional[Location]:
...
 @dataclass(frozen=True)
 class ModelType:
+    """
+    Dataclass used with typing.Annotated to represent Inmanta model types in Python code.
+
+    If we want to represent "std::Entity" as "typing.Any" in our plugins we could define the following type:
+
+    Entity: typing.TypeAlias = typing.Annotated[typing.Any, ModelType["std::Entity"]]
+
+    or even
```
sanderr, line 276: I see no reason to show the two syntaxes here, I would simply use the new one (`type` keyword).

src/inmanta/plugins.py
```diff
@@ -266,18 +266,35 @@ def get_location(self) -> Optional[Location]:
...
+    type Entity = typing.Annotated[typing.Any, ModelType["std::Entity"]]
+
+    and then use it on our plugin:
+
+    @plugin
+    def my_plugin(value: Entity) -> None:
+        pass
+
+    We will validate the argument as "std::Entity", while presenting
+    a proper Python type (typing.Any) for IDE and static typing purposes.
```
sanderr, line 287:
```suggestion
    a proper Python type (typing.Any) for IDE and static typing purposes.
    It is the user's responsibility to ensure that the validation type sufficiently matches the Python type.
```

src/inmanta/plugins.py
sanderr: But we may have `Annotated` fields in the dataclasses, so we need to support that. The explainer seems to get a `problem` with an `entity` attached, so we can easily get what we need there. Once done, I would request @wouterdb's review for this file just in case.

sanderr (changes requested):
Replied to open discussions

src/inmanta/ast/entity.py
jptrindade: I think we specifically need a `Range` to create a `LocatableString` to pass onto `inmanta_type.resolve_type`, right?

sanderr: But we have one in `self.location`.

sanderr (approved):

wouterdb (approved):

PR #8668, opened 2025-01-22
make sure plugins are only normalized once

sanderr:
# Description

Plugins were normalized twice: once explicitly and once when we loop over all types to normalize. This PR drops the first, and makes sure we normalize `Entity` near the end because it may depend on plugins and other types for its default validation.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/ast/statements/define.py
+++ b/src/inmanta/ast/statements/define.py
@@ -479,8 +479,16 @@ class DefineTypeConstraint(TypeDefinitionStatement):
         constraint_type.comment = self.comment
         constraint_type.basetype = basetype
         constraint_type.constraint = self.expression
-        self.expression.normalize()
-        self.anchors.extend(self.expression.get_anchors())
+
+    def get_anchors(self) -> list[Anchor]:
+        """
+        This method overrides the default get_anchors() to accommodate the two-stage normalization process.
+        DefineTypeConstraint registers anchors for its condition expression. However, these anchors only come into existence
+        after the type normalization phase.
+        This implementation ensures that anchors are correctly gathered from both the condition expression and
+        the statement itself.
+        """
+        return [*self.anchors, *self.expression.get_anchors()]
 
 
 Relationside = tuple[LocatableString, Optional[LocatableString], Optional[tuple[int, Optional[int]]]]
--- a/src/inmanta/execute/scheduler.py
+++ b/src/inmanta/execute/scheduler.py
@@ -259,8 +259,15 @@ class Scheduler:
         self.types = {k: v for k, v in types_and_impl.items() if isinstance(v, Type)}
 
         # give type info to all types (includes plugins), to normalize blocks inside them
-        # normalize implementations last because they have subblocks that might depend on other type information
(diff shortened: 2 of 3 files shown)


src/inmanta/ast/statements/define.py
```diff
@@ -479,8 +479,16 @@ def evaluate(self) -> None:
         constraint_type.comment = self.comment
         constraint_type.basetype = basetype
         constraint_type.constraint = self.expression
-        self.expression.normalize()
```
sanderr, line 482: This was another statement that was normalized twice. I worked around it using the same approach we use for some other `Define...` statements. Though this should be even less noticeable than the plugins.

wouterdb (approved):

src/inmanta/execute/scheduler.py
```diff
@@ -256,23 +256,24 @@ def define_types(self, compiler: "Compiler", statements: Sequence["Statement"],
...
+        # give type info to all types (includes plugins), to normalize blocks inside them
+        for t in sorted(
+            self.types.values(),
+            key=lambda t: (
+                # normalize implementations last because they have subblocks that might depend on other type information
+                isinstance(t, Implementation),
+                # normalize entities second last because it may call validate on its attribute types to validate defaults
+                isinstance(t, Entity),
+            ),
+        ):
```
Hugo-Inmanta, line 270: If I understand correctly, this implementation will normalize plugins first since `key = (0, 0)` for all of them.
1. Does it matter that they are normalized before `Implementations` and `Entities` ?
2. Does the relative order in which plugins are normalized matter ? Do we need it to be deterministic ? If I understand correctly this order depends on the order in which the plugin files are loaded and the plugins functions registered.

Hugo-Inmanta (commented):
LGTM as far as I can see, One comment I'm curious about.

src/inmanta/execute/scheduler.py
sanderr: Indeed, that's how it works. With the two boolean sort keys, it gets right to the edge of being too complex a solution for me. But I still (barely) prefer it over a more verbose one. What do you think?

1. It has to be before both implementations and entities, because plugin normalization is standalone, while the other two may need other type information. The fact that this is so is a bit of on implementation detail that is assumed to hold here. I don't love it but I think it's the best we can do for now.
2. Doesn't matter. Deterministic in the sense that two consecutive compiles do it in the same order is important for deterministic error reporting etc. Deterministic in the sense that this implementation maintains the same order as the one on master isn't that important.

Hugo-Inmanta: > Indeed, that's how it works. With the two boolean sort keys, it gets right to the edge of being too complex a solution for me. But I still (barely) prefer it over a more verbose one. What do you think?

No strong opinion. I haven't fully considered performance but couldn't we add a class variable `normalization_priority` and sort on this. This part would be more readable and the ordering explicit

sanderr: I considered it, but I felt that it shifts the problem more than solves it. Then suddenly every class has to decide what a sane priority is, which means it has to know the priorities of other type classes.

PR #8680, opened 2025-01-27
bugfix for dry-run code loading

sanderr:
# Description

Fixed bug in scheduler code loading:
- dry-runs will now load appropriate code version. They will also work (while they did not before) if no version for that agent has been released yet.
- fixed rare race condition in code loading for deploys and get-facts: ensure collection of resource types for an agent is fetched atomically when the resource's intent is fetched. (No change entry for this one because it is more a theoretical issue than a practical one.)

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/scheduler-dryrun-bugfix.yml
+++ b/changelogs/unreleased/scheduler-dryrun-bugfix.yml
@@ -1,7 +1,7 @@
 description: Bugfix in dryrun code loading
 change-type: patch
 sections:
-  bugfix: Make sure dry-runs load the handler code apppropriate for their version
+  bugfix: Make sure dry-runs load the appropriate handler code for their version
 destination-branches:
   - master
   - iso8
--- a/changelogs/unreleased/scheduler-dryrun-bugfix.yml
+++ b/changelogs/unreleased/scheduler-dryrun-bugfix.yml
@@ -1,0 +1,4 @@
+description: Bugfix in dryrun code loading
+change-type: patch
+sections:
+  bugfix: Make sure dry-runs load the handler code apppropriate for their version
--- a/changelogs/unreleased/scheduler-dryrun-bugfix.yml
+++ b/changelogs/unreleased/scheduler-dryrun-bugfix.yml
@@ -2,3 +2,6 @@ description: Bugfix in dryrun code loading
 change-type: patch
 sections:
   bugfix: Make sure dry-runs load the handler code apppropriate for their version
+destination-branches:
+  - master
+  - iso8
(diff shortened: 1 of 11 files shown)


src/inmanta/deploy/tasks.py
```diff
@@ -72,13 +73,33 @@ def get_executor_resource_details(self, version: int, resource_intent: "state.Re
         )
 
     async def get_executor(
-        self, task_manager: "scheduler.TaskManager", agent: str, resource_type: ResourceType, version: int
+        self,
+        *,
+        task_manager: "scheduler.TaskManager",
+        agent_spec: tuple[str, Collection[ResourceType]],
```
sanderr, line 79: I'm slightly on the fence about this one. I made it a tuple because they're conceptually coupled, and it prevents the confusion of a resource type argument, as well as an argument for all resource types. But it's slightly unconventional.

changelogs/unreleased/scheduler-dryrun-bugfix.yml
```diff
@@ -0,0 +1,7 @@
+description: Bugfix in dryrun code loading
+change-type: patch
+sections:
+  bugfix: Make sure dry-runs load the handler code apppropriate for their version
```
Hugo-Inmanta, line 4:
```suggestion
  bugfix: Make sure dry-runs load the handler code appropriate for their version
```

src/inmanta/deploy/tasks.py
```diff
@@ -225,7 +249,10 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
         started = datetime.datetime.now().astimezone()
         try:
             my_executor: executor.Executor = await self.get_executor(
-                task_manager, agent, executor_resource_details.id.entity_type, self.version
+                task_manager=task_manager,
+                agent_spec=(agent, [executor_resource_details.id.entity_type]),
```
Hugo-Inmanta, line 253: any reason for not retrieving the entity type from `self.resource_intent` instead ?

wouterdb (changes requested):

sanderr:
Note to self: discussed with Wouter in person: make an attempt to move the responsibility of mapping agents to resource types to the code loader instead of the scheduler. Than much of the ugliness in this PR can go.

src/inmanta/deploy/tasks.py
sanderr: I think both are fine, I stuck with what was used on other lines.

wouterdb (approved):

PR #8685, opened 2025-01-28
Make api endpoint to see which version is schedulled

wouterdb:
# Description

We currently have no way of knowing the latest released version is fully deployed, becuase we can't establish if it has been scheduled. 

This PR add a new api field to indicate this. 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/const.py
+++ b/src/inmanta/const.py
@@ -331,14 +331,6 @@ class DesiredStateVersionStatus(str, Enum):
     skipped_candidate = "skipped_candidate"
 
 
-class ExtendedDesiredStateVersionStatus(str, Enum):
-    released = "released"
-    active = "active"
-    candidate = "candidate"
-    retired = "retired"
-    skipped_candidate = "skipped_candidate"
-
-
 class NotificationSeverity(str, Enum):
     """
     The possible values determine the styling used by the frontend to show them,
--- a/src/inmanta/const.py
+++ b/src/inmanta/const.py
@@ -331,6 +331,14 @@ class DesiredStateVersionStatus(str, Enum):
     skipped_candidate = "skipped_candidate"
 
 
+class ExtendedDesiredStateVersionStatus(str, Enum):
+    released = "released"
+    active = "active"
+    candidate = "candidate"
+    retired = "retired"
+    skipped_candidate = "skipped_candidate"
+
+
 class NotificationSeverity(str, Enum):
     """
     The possible values determine the styling used by the frontend to show them,
--- a/tests/server/test_desired_state_versions.py
+++ b/tests/server/test_desired_state_versions.py
@@ -20,17 +20,112 @@ import datetime
 import json
 import uuid
 from operator import itemgetter
+from typing import Any, Optional
 
 import pytest
 from tornado.httpclient import AsyncHTTPClient, HTTPRequest
 
 from inmanta import data
+from inmanta.agent.config import agent_reconnect_delay, server_timeout
+from inmanta.const import AGENT_SCHEDULER_ID
+from inmanta.data import ConfigurationModel, Scheduler
 from inmanta.data.model import DesiredStateLabel, PromoteTriggerMethod
-from inmanta.server import config
+from inmanta.protocol import SessionEndpoint, methods
+from inmanta.server import SLICE_AGENT_MANAGER, config, protocol
+from inmanta.types import Apireturn
+from utils import retry_limited
+
+
+class BaseAgent(SessionEndpoint):
+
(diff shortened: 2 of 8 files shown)


src/inmanta/const.py
```diff
@@ -331,6 +331,14 @@ class DesiredStateVersionStatus(str, Enum):
     skipped_candidate = "skipped_candidate"
 
 
+class ExtendedDesiredStateVersionStatus(str, Enum):
+    released = "released"
+    active = "active"
+    candidate = "candidate"
+    retired = "retired"
+    skipped_candidate = "skipped_candidate"
```
sanderr, line 339: This doesn't seem to be used.

sanderr (approved):

tests/server/test_desired_state_versions.py
```diff
@@ -20,17 +20,112 @@
...
+from inmanta.const import AGENT_SCHEDULER_ID
+from inmanta.data import ConfigurationModel, Scheduler
+from inmanta.data.model import DesiredStateLabel
+from inmanta.protocol import SessionEndpoint, methods
+from inmanta.server import SLICE_AGENT_MANAGER, config, protocol
+from inmanta.types import Apireturn
+from utils import retry_limited
+
+
+class BaseAgent(SessionEndpoint):
```
arnaudsjs, line 39: This class needs documentation. What does it do?

tests/server/test_desired_state_versions.py
```diff
@@ -20,17 +20,112 @@
...
+        return 200
+
+    @protocol.handle(methods.get_status)
+    async def get_status(self) -> Apireturn:
+        return 200, {}
+
+
+@pytest.fixture(scope="function")
+async def base_agent(server, environment):
+    """Construct an agent that does nothing"""
```
arnaudsjs, line 113: This is the same docstring as the `null_agent` fixture. Why would I use this fixture instead of the `null_agent` fixture?

tests/server/test_desired_state_versions.py
arnaudsjs, line 39: `BaseAgent` sounds like this is an abstract base class. It also doesn't really say anything about what the class does or why it exists. I think we have to come up with a better name.

tests/server/test_desired_state_versions.py
```diff
@@ -336,21 +458,21 @@ async def test_promote_no_versions(server, client, environment: str):
...
     await clienthelper.put_version_simple(resources, version)
-    result = await client.promote_desired_state_version(environment, version=version, trigger_method=trigger_method)
-    assert result.code == 200
-    result = await client.list_desired_state_versions(environment)
+    result = await client.promote_desired_state_version(environment, version=version)
     assert result.code == 200
-    assert len(result.result["data"]) == 1
-    assert result.result["data"][0]["status"] == "active"
+
+    async def is_good():
```
arnaudsjs, line 472:
```suggestion
    async def is_version_active() -> bool:
```

tests/server/test_desired_state_versions.py
```diff
@@ -20,17 +20,112 @@
...
+        self,
+        environment: Optional[uuid.UUID] = None,
+    ):
+        """
+        :param environment: environment id
+        """
+        super().__init__(name="agent", timeout=server_timeout.get(), reconnect_delay=agent_reconnect_delay.get())
+        self._env_id = environment
+        self.enabled: dict[str, bool] = {}
+        self.limit = 100
```
arnaudsjs, line 51: The limit variable needs documentation.

tests/server/test_desired_state_versions.py
```diff
@@ -20,17 +20,112 @@
...
+    def __init__(
+        self,
+        environment: Optional[uuid.UUID] = None,
+    ):
+        """
+        :param environment: environment id
+        """
+        super().__init__(name="agent", timeout=server_timeout.get(), reconnect_delay=agent_reconnect_delay.get())
+        self._env_id = environment
+        self.enabled: dict[str, bool] = {}
```
arnaudsjs, line 50: What is the point of the `enabled` variable? Is it used anywhere?

arnaudsjs (changes requested):

src/inmanta/const.py
wouterdb: thx

tests/server/test_desired_state_versions.py
wouterdb: Copy-paste, I removed it

arnaudsjs (approved):

PR #8693, opened 2025-01-30
Issue/2025 01 24 update quickstart docs

Hugo-Inmanta:
# Description

update quickstart docs

# Self Check:

- [ ] Changelog entry


docs/quickstart.rst
```diff
@@ -96,6 +96,7 @@ Go to the `SR Linux` folder and then `containerlab` to spin-up the containers:
 
     cd examples/Networking/SR\ Linux/containerlab
     sudo docker pull ghcr.io/nokia/srlinux:latest
+    sudo docker pull ghcr.io/inmanta/orchestrator:latest
```
sanderr, line 99: I don't know about this one. I'm not sufficiently familiar with the containers to review this part.

Hugo-Inmanta: @bartv wdyt ? 
For reference this is the corresponding topology file https://github.com/inmanta/examples/blob/master/Networking/SR%20Linux/containerlab/topology.yml

docs/lsm/quickstart/quickstart.rst
```diff
@@ -253,7 +253,7 @@ The following command executes a script to copy the required resources to a spec
 
 .. code-block::
 
-   $ docker exec -ti -w /code clab-srlinux-inmanta-server  /code/setup.sh
+   $ docker exec -ti -w /code clab-srlinux-inmanta-server  /code/setup.sh <environment uuid>
```
edvgui, line 256: Does this code setup still work with the new disk layout?

docs/quickstart.rst
edvgui: The image looks good, but I am not sure why we are pulling it here, containerlab will do it too

docs/lsm/quickstart/quickstart.rst
Hugo-Inmanta: Yes, it should work with both layouts (cf https://github.com/inmanta/examples/commit/f66765bb1451c303bd7af811c9745b8f70f6a0cf#diff-96bee1b03fa3c4fbe209cd6258fea0b16687880a14d3c79247b3b30cba013b04)

docs/quickstart.rst
Hugo-Inmanta: If containerlab can find an image with the same name locally it will use it instead of actually pulling the latest one.
I think I added this line here because I ran into this issue

docs/quickstart.rst
```diff
@@ -95,7 +95,6 @@ Go to the `SR Linux` folder and then `containerlab` to spin-up the containers:
 .. code-block:: sh
 
     cd examples/Networking/SR\ Linux/containerlab
-    sudo docker pull ghcr.io/nokia/srlinux:latest
```
Hugo-Inmanta, line 98: https://github.com/inmanta/examples/pull/148 updates the pull policy to no longer have to explicitly pull

edvgui (approved):

Issue #8111, opened 2024-09-20
Package fat sdists for modules v2


wouterdb:
Currently, for V2 modules, we don't distribute as much content as for V1 modules

The proposed solution is to build an sdist that contains more information and keep the wheel the same

1. By modifying manifest.in we can include more content into sdist without adding it to the wheel
   - [ ] modify the module template to include docs, examples and tests into the sdist
2. We should update module tool to also build the sdist, not just the wheel 
3. update module template etc

PR #8711, opened 2025-02-05
Added support to build sdist of V2 module.

arnaudsjs:
# Description

Added support to build sdist of V2 module.

Part of #8111

- [x] Changelog entry


src/inmanta/moduletool.py
```diff
@@ -768,12 +786,35 @@ def build(
...
 
+        distributions_to_build: Sequence[Literal["wheel", "sdist"]]
+        if distribution is None:
+            distributions_to_build = ["wheel"]
+        elif distribution == "wheel_and_sdist":
+            distributions_to_build = ["wheel", "sdist"]
+        else:
+            distributions_to_build = [distribution]
+
+        timestamp = datetime.datetime.now(datetime.timezone.utc)
```
sanderr, line 797: I'm slightly hesitant about this being a naive timestamp. In practice, we currently only pass it to a method that wouldn't even care for the timezone part, but this is counter to the convention we have throughout all our code. So I think I'd prefer to create a tz-aware timestamp here, and then convert to naive UTC in the build method.

sanderr (changes requested):

src/inmanta/moduletool.py
```diff
@@ -753,10 +761,20 @@ def v1tov2(self, module: str) -> None:
...
+        byte_code: bool = False,
+        distribution: Literal["wheel", "sdist", "wheel_and_sdist"] | None = None,
+    ) -> str | list[str]:
         """
         Build a v2 module and return the path to the build artifact.
+
+        :param distribution: The distribution package that should be built. If None, build a wheel.
+        :returns: The paths to the distribution packages that were built. If distribution=None,
+                  this method returns a str for backwards compatibility. In all other situations
+                  this method returns a list.
```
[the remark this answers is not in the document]
arnaudsjs: I know this module is used by IRT, but now I see it doesn't use the return value so that's not an issue. I was under the assumption that it would be used by some CI tooling, but they probably use the CLI tool instead of the Python class. I tend to agree with you that this is not really required. I'll make the change.

src/inmanta/moduletool.py
arnaudsjs: I don't understand what you mean. The above-mentioned code does provide a timestamp, right?

tests/moduletool/test_build.py
```diff
@@ -105,47 +123,59 @@ def test_build_v2_module(
...
-        assert "linux_x86_64" in wheel_file
+    if distribution == "wheel":
+        with zipfile.ZipFile(package_file) as zip:
+            zip.extractall(extract_dir)
     else:
-        assert "none-any" in wheel_file
+        with tarfile.open(package_file) as tar:
+            tar.extractall(extract_dir)
+            # member_path = pathlib.Path(tar.getnames()[0])
+            # extract_dir = str(pathlib.Path(extract_dir).relative_to(member_path.parts[0]))
```
jptrindade, line 145: leftover code?

jptrindade (approved):
I see that we are reusing `Literal["wheel", "sdist"]` a lot, should we create a type for it?

src/inmanta/moduletool.py
sanderr: Don't know what I was thinking here, LGTM

src/inmanta/moduletool.py
```diff
@@ -636,7 +638,13 @@ def modules_parser_config(cls, parser: ArgumentParser, parent_parsers: abc.Seque
             default=False,
             dest="byte_code",
         )
-
+        build.add_argument(
+            "-d",
```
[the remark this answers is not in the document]
sanderr: Good question. I think we can, all it does is to produce an additional artifact. But perhaps we should check on Slack if anyone objects.

sanderr (commented):
Added comments

arnaudsjs:
> I see that we are reusing `Literal["wheel", "sdist"]` a lot, should we create a type for it?

Could you elaborate on this. What do you want to change it to? An Enum?

tests/moduletool/test_build.py
arnaudsjs: It seems I forgot to push my latest commits.

src/inmanta/moduletool.py
```diff
@@ -766,16 +776,20 @@ def build(
...
-                  this method returns a str for backwards compatibility. In all other situations
-                  this method returns a list.
+        :param wheel: True iff build a wheel package.
+        :param sdist: True iff build a sdist package.
+        :returns: A list of paths to the distribution packages that were built.
         """
+        if not wheel and not sdist:
+            # Only build wheels by default for backwards compatibility.
+            wheel = True
+
```
sanderr, line 792: I think this whole block can go, with how you implemented it below?

src/inmanta/moduletool.py
```diff
@@ -786,15 +800,15 @@ def build(
         if output_dir is None:
             output_dir = os.path.join(path, "dist")
 
+        timestamp = datetime.datetime.now(datetime.timezone.utc)
+
         distributions_to_build: Sequence[Literal["wheel", "sdist"]]
-        if distribution is None:
-            distributions_to_build = ["wheel"]
-        elif distribution == "wheel_and_sdist":
+        if wheel is sdist:
```
sanderr, line 806: Nice, clean! Could use a comment though.

sanderr (changes requested):

sanderr (approved):

PR #8716, opened 2025-02-06
added native support for dataclasses

wouterdb:
# Description

added native support for dataclasses

# TODO

1. [x] update docs on details https://github.com/inmanta/designs/pull/148
2. [x] wait for annotated types #8665
3. [x] produce warning if a string is both a valid python and inmanta type and they are different -> https://github.com/inmanta/inmanta-core/issues/8750
4. [x] make annotated type with dynamic proxy the way of getting a dynamic proxy

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/execute/scheduler.py
+++ b/src/inmanta/execute/scheduler.py
@@ -55,6 +55,7 @@ from inmanta.execute.runtime import (
     Waiter,
     WaiterSet,
 )
+from inmanta.plugins import Plugin
 
 if TYPE_CHECKING:
     from inmanta.ast import BasicBlock, NamedType, Statement  # noqa: F401
@@ -258,9 +259,9 @@ class Scheduler:
         compiler.plugins = {k: v for k, v in types_and_impl.items() if isinstance(v, plugins.Plugin)}
         self.types = {k: v for k, v in types_and_impl.items() if isinstance(v, Type)}
 
-        # give type info to all types (includes plugins), to normalize blocks inside them
+        # give type info to all types (excluding plugins), to normalize blocks inside them
         for t in sorted(
-            self.types.values(),
+            (t for t in self.types.values() if not isinstance(t, Plugin)),
             key=lambda t: (
                 # normalize implementations last because they have subblocks that might depend on other type information
                 isinstance(t, Implementation),
@@ -269,10 +270,6 @@ class Scheduler:
             ),
         ):
             t.normalize()
-
-        # normalize root blocks
-        for block in blocks:
-            block.normalize()
 
         # Dataclass validation
         data_class_root = self.types.get("std::Dataclass")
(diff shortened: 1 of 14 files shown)


src/inmanta/plugins.py
```diff
@@ -294,6 +294,12 @@ def to_dsl_type(python_type: type[object]) -> inmanta_type.Type:
             bases = [to_dsl_type(arg) for arg in typing.get_args(python_type)]
             return inmanta_type.Union(bases)
 
+    if dataclasses.is_dataclass(python_type):
+        entity = get_inmanta_type_for_dataclass(python_type)
+        if entity:
+            return entity
+        raise TypingException(None, f"invalid type {python_type}, this dataclass has no associated inmanta entity")
```
wouterdb, line 301: I'm a bit uncertain about this, I love to close the gap, but I'm afraid of what will blow up

sanderr: But native annotations are a new feature, nothing can blow up really, can it?

src/inmanta/execute/scheduler.py
```diff
@@ -258,9 +258,9 @@ def define_types(self, compiler: "Compiler", statements: Sequence["Statement"],
         compiler.plugins = {k: v for k, v in types_and_impl.items() if isinstance(v, plugins.Plugin)}
         self.types = {k: v for k, v in types_and_impl.items() if isinstance(v, Type)}
 
-        # give type info to all types (includes plugins), to normalize blocks inside them
+        # give type info to all types (excluding plugins), to normalize blocks inside them
         for t in sorted(
-            self.types.values(),
+            (t for t in self.types.values() if not isinstance(t, plugins.Plugin)),
```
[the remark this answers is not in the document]
wouterdb: indeed, that kind of sucks.

So either we need a thirds stage of typing:
1. collect names
2. resolve
3. validate values  (new)

or we need to move the first part of dataclass pairing to the first phase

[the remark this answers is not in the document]
sanderr: I had a quick glance at the implementation, and I think we could safely make the dataclass pairing part of `Entity.normalize`. Do you agree?

src/inmanta/plugins.py
```diff
@@ -239,6 +240,50 @@ def get_location(self) -> Optional[Location]:
         return None
 
 
+class ConvertibleEntity(inmanta_type.Type):
+
+    # TODO: cache
```
wouterdb, line 245: todo alert

jptrindade (changes requested):
Can you add a test where we provide the wrong type (and return type) to a plugin that accepts a Dataclass? I wanted to know if the `type_string_internal` logic is sound and that we get an appropriate response

jptrindade (approved):

PR #8717, opened 2025-02-07
Make wait for released wait for the version to be active

wouterdb:
# Description

1. I updated the wait for version fixture to wait for the version to be scheduled/active. 
   - This makes the name a bit weird
   - But it makes it behave as it did prior to the scheduler
2. I switched to the V2 endpoint, which reports the status in accordance to the new rules (updated in https://github.com/inmanta/inmanta-core/pull/8685)
3. I did not update the V1 endpoint, which is now a bit weird/deceptive. I just don't see a good way of fixing it. Ideas welcome.  

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -318,15 +318,20 @@ async def report_db_index_usage(min_precent=100):
 
 async def wait_until_version_is_released(client, environment: uuid.UUID, version: int) -> None:
     """
-    Wait until the configurationmodel with the given version and environment is released.
+    Wait until the configurationmodel with the given version and environment is released and scheduled.
     """
 
     async def _is_version_released() -> bool:
-        result = await client.get_version(tid=environment, id=version)
-        if result.code == 404:
-            return False
-        assert result.code == 200
-        return result.result["model"]["released"]
+        versions = await client.list_desired_state_versions(
+            tid=self.envid,
+            start=version-1,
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -318,15 +318,20 @@ async def report_db_index_usage(min_precent=100):
...
+        versions = await client.list_desired_state_versions(
+            tid=self.envid,
+            start=version-1,
+            limit=1,
+        )
+        assert versions.code == 200
+        lookup = {
+            int(v["version"]): v["status"] == "active" for v in versions.result["data"]
+        }
+        return lookup[int(version)]
```
Hugo-Inmanta, line 334: I think cast is redundant
```suggestion
        return lookup[version]
```

Hugo-Inmanta (commented):
LGTM, what's the purpose of this change ? Was this causing race conditions somewhere ?

tests/utils.py
arnaudsjs, line 334: The previous implementation covered the case where version doesn't exist yet. This one doesn't.
```suggestion
        return any(v["status"] == "active" for v in versions.result["data"] if int(v["version"]) == version)
```

arnaudsjs (changes requested):

PR #8719, opened 2025-02-09
Websocket refactor

bartv:
## Summary

Replace the REST-based long-poll/heartbeat mechanism for bidirectional server-agent communication with persistent WebSocket connections. This eliminates the heartbeat polling loop, session expiry timers, and the `AgentInstance`/`AgentProcess` data models in favor of a simpler, connection-oriented session model over WebSockets.

## Motivation

The existing bidirectional communication between the server and agents relies on a heartbeat-based long-poll mechanism: agents send periodic `POST /heartbeat` requests, and the server queues up method calls to return in the heartbeat response. This approach has several drawbacks:

- **Latency**: Server-to-agent calls are delayed until the next heartbeat response, introducing variable latency (up to the hang time interval).
- **Complexity**: The `Session`, `SessionManager`, `ReturnClient`, and `SessionEndpoint` classes implement an intricate dance of futures, queues, and expiry timers to simulate bidirectional communication over unidirectional HTTP.
- **Data model overhead**: `AgentProcess` and `AgentInstance` tables track process liveness and endpoint-to-session mappings that exist only because sessions carry multiple named endpoints and require primary election.
- **Fragility**: Session expiry is timer-based rather than connection-based, making it susceptible to false expirations under load and requiring careful tuning of timeout parameters.

WebSockets provide a native bidirectional channel, allowing the server to push calls to agents immediately and eliminating the need for long-poll queuing, heartbeat replies, and session validity checks based on session IDs (SIDs).

### New WebSocket protocol (`src/inmanta/protocol/websocket.py`)

A new module implements the WebSocket framing protocol with the following message types, discriminated by an `action` field:

| Message | Direction | Purpose |
|---------|-----------|---------|
| `OPEN_SESSION` | Client → Server | Request to open a named session for an environment |
| `SESSION_OPENED` | Server → Client | Confirms session is active |
| `REJECT_SESSION` | Server → Client | Rejects a session (e.g., duplicate, shutting down) |
| `CLOSE_SESSION` | Either direction | Graceful session teardown |
| `RPC_CALL` | Either direction | Method invocation (url, method, headers, body, reply_id) |
| `RPC_REPLY` | Either direction | Response to an RPC_CALL (reply_id, code, result) |

Messages are serialized as JSON and validated using Pydantic's discriminated union (`pydantic.TypeAdapter` with `pydantic.Field(discriminator="action")`).

Key classes:
- **`Session`**: Represents one logical session identified by `(environment_id, session_name)`. Provides `get_client()` and `get_typed_client()` for making RPC calls over the WebSocket.
- **`WebsocketFrameDecoder`**: Stateful message processor that handles the session lifecycle and dispatches `RPC_CALL` messages to registered `CallTarget` handlers. Manages a dict of pending reply futures with timeout support.
- **`SessionEndpoint`**: Client-side endpoint (replaces old `endpoints.SessionEndpoint`) that maintains a WebSocket connection with automatic reconnection. Uses Tornado's `WebSocketClientConnection` with ping/pong for liveness detection.
- **`SessionListener`**: Callback interface with `session_opened` / `session_closed` (replaces old `new_session` / `expire` / `seen` triple).

### Session lifecycle

1. Client connects to `ws://server:port/v2/ws`
2. Client sends `OPEN_SESSION` with `environment_id`, `session_name`, and `hostname`
3. Server creates a `Session` object, notifies `SessionListener`s, responds with `SESSION_OPENED`
4. Both sides can now send `RPC_CALL` / `RPC_REPLY` messages
5. On disconnect, Tornado's `on_close` triggers `close_session()` which notifies listeners
6. Client automatically reconnects after `reconnect_delay` seconds

Sessions are now identified by `(environment_id, session_name)` tuples rather than random UUIDs. Each agent process opens exactly one session named `"agent"` per environment, eliminating the concept of multiple named endpoints per session and the associated primary election logic.

### Changes to session management (`src/inmanta/server/agentmanager.py`)

The `AgentManager` is substantially simplified:

- **Removed**: `_seen_session`, `_failover_endpoints`, `_ensure_primary_if_not_exists`, `is_primary`, and all primary election logic. With one session per environment, there is no need for endpoint-level failover.
- **Removed**: `tid_endpoint_to_session` and `endpoints_for_sid` mappings, replaced by a single `scheduler_for_env: dict[UUID, Session]`.
- **Simplified**: `SessionAction` no longer carries `endpoint_names_snapshot`.
- **Simplified**: `SessionActionType.SEEN_SESSION` removed entirely (WebSocket pings handle liveness).
- **Split**: `halt_agents` split into DB operations (`_update_paused_status_in_db`) and RPC notifications (`halt_agents_notify`/`_notify_agent_state_change`) to avoid deadlocks when the agent's RPC handler queries the DB during an uncommitted transaction.
- **Implements**: `websocket.SessionListener` instead of the old `protocol.SessionListener`.

### Changes to the data model (`src/inmanta/data/__init__.py`)

- **Renamed**: `AgentProcess` → `SchedulerSession` (table: `schedulersession`)
- **Removed**: `last_seen` column from `SchedulerSession` (WebSocket connection is the liveness signal)
- **Removed**: `AgentInstance` model entirely (no more per-endpoint instance tracking)
- **Removed**: `Agent.id_primary` column and `Agent.primary` property (no more primary election)
- **Simplified**: `Agent.clean_up()` no longer checks for `id_primary IS NULL`

### DB migration (`src/inmanta/db/versions/v202502031.py`)

```sql
ALTER TABLE public.agent DROP COLUMN id_primary;
DROP TABLE public.agentinstance;
ALTER TABLE public.agentprocess RENAME TO schedulersession;
ALTER TABLE public.schedulersession DROP COLUMN last_seen;
```

### Protocol method changes (`src/inmanta/protocol/methods.py`)

- **Removed**: `heartbeat` method (`POST /heartbeat`)
- **Removed**: `heartbeat_reply` method (`PUT /heartbeat`)
- **Changed**: `get_state` converted from `@method` (v1) to `@typedmethod` (v2), `sid` parameter removed since session identity is now implicit in the WebSocket connection

### Protocol common changes (`src/inmanta/protocol/common.py`)

- **Moved**: `CallTarget` class from `endpoints.py` to `common.py` (avoids circular imports)
- **Moved**: `VersionMatch` enum from `endpoints.py` to `common.py`
- **Added**: `MethodProperties.select_method()` class method (extracted from `Client.__getattr__`)
- **Added**: `typed_process_response()` function for converting responses to typed results
- **Removed**: `SessionManagerInterface` (replaced by `SessionListener` pattern)
- **Removed**: `validate_sid` parameter from `MethodProperties`

### REST server changes (`src/inmanta/protocol/rest/server.py`)

- **Added**: `WebsocketHandler` (Tornado `WebSocketHandler` + `WebsocketFrameDecoder`) that handles the `/v2/ws` endpoint
- **Changed**: `RESTServer` now implements `AuthnzInterface` and manages sessions directly (replaces `SessionManager` slice)
- **Added**: Session registry, `register_session()`, `notify_close_session()`, `get_session()`, `add_session_listener()`
- **Changed**: WebSocket ping interval set to 1 second via Tornado's `websocket_ping_interval` application setting

### REST base changes (`src/inmanta/protocol/rest/__init__.py`)

- **Extracted**: `execute_call()` from `RESTBase._execute_call()` to a standalone function (used by both REST and WebSocket paths)
- **Extracted**: `AuthnzInterface` as a separate class from `RESTBase`
- **Added**: `match_call()` function for URL-based method dispatch (used by WebSocket frame decoder)
- **Removed**: `validate_sid` logic from `execute_call()`

### Classes
- `endpoints.SessionEndpoint` (replaced by `websocket.SessionEndpoint`)
- `endpoints.SessionClient` (replaced by `websocket._SessionClient`)
- `protocol.Session` (old long-poll session with queue-based call dispatch)
- `protocol.SessionListener` (old 3-method interface: `new_session`, `expire`, `seen`)
- `protocol.SessionManager` (old server slice managing sessions and heartbeat handlers)
- `protocol.ReturnClient` (old client that sent calls via session queue)
- `data.AgentInstance` (entire ORM model)

### Protocol methods
- `heartbeat` (POST /heartbeat)
- `heartbeat_reply` (PUT /heartbeat)

### Server infrastructure
- `SLICE_SESSION_MANAGER` constant
- `SessionManager` server slice (heartbeat handler, session expiry)

### Data model
- `agentinstance` table (entire table dropped)
- `agentprocess.last_seen` column
- `agent.id_primary` column

## Breaking Changes

1. **API removal**: `POST /heartbeat` and `PUT /heartbeat` endpoints no longer exist.
2. **Data model**: `agentinstance` table dropped; `agentprocess` renamed to `schedulersession`; `agent.id_primary` column removed.
3. **`get_state` API**: Changed from v1 (`@method`) to v2 (`@typedmethod`), `sid` parameter removed.
4. **`SessionEndpoint` constructor**: Now requires `environment` parameter; no longer supports `set_environment()` or `add_end_point_name()`.
5. **`SessionListener` interface**: Changed from 3 methods (`new_session`, `expire`, `seen`) to 2 methods (`session_opened`, `session_closed`).
6. **`validate_sid` removed**: Session validity is now implicit in the WebSocket connection.
7. **Import paths**: `SessionEndpoint`, `Session`, `SessionListener` now imported from `inmanta.protocol.websocket` (re-exported from `inmanta.protocol.__init__`). `SessionClient` no longer exists. `CallTarget` moved from `inmanta.protocol.endpoints` to `inmanta.protocol.common`.

## Test Changes

- **New**: `tests/protocol/test_ws.py` — end-to-end test for WebSocket bidirectional RPC
- **Removed**: `test_primary_selection` (primary election logic removed), multiple `AgentInstance` tests
- **Rewritten**: `test_session_creation_fails` for new websocket session model
- **Updated**: `test_2way_protocol.py`, `test_agent_manager.py`, `test_data.py` adapted to new session model, `MockSession` simplified, `AgentView` used instead of direct `Agent.primary` queries


src/inmanta/protocol/websocket.py
```diff
@@ -0,0 +1,769 @@
...
+                r = await result
+                r.method_properties = method
+                return r
+
+            return common.ClientCall.create(with_properties(), properties=method)
+
+        return wrap
+
+
+class SessionListener:
```
bartv: Missing documentation on this interface

src/inmanta/protocol/websocket.py
```diff
@@ -0,0 +1,769 @@
...
+    try:
+        await asyncio.wait_for(future, timeout)
+    except asyncio.TimeoutError:
+        LOGGER.warning(log_message)
+    finally:
+        replies.pop(reply_id, None)
+
+
+class WebsocketFrameDecoder(util.TaskHandler[None]):
+    """A close that offers an on_message to handle websocket frames and calls handlers for each event"""
```
bartv: This docstring makes no sense, many of the methods in this class are missing docstrings.

src/inmanta/protocol/websocket.py
```diff
@@ -0,0 +1,769 @@
...
+        self.create_session(
+            environment_id=self.environment,
+            session_name=self.name,
+        )
+
+    def is_auth_enabled(self) -> bool:
+        """
+        Return True iff authentication is enabled.
+        """
+        return False
```
bartv: Auth still needs to be implemented for websockets

tests/protocol/test_ws.py
```diff
@@ -0,0 +1,815 @@
...
+    rs.add_slice(server)
+    await rs.start()
+
+    try:
+        # Wait for the agent to connect
+        async def wait_for_session() -> None:
+            while not agent.session or not agent.session.active:
+                await asyncio.sleep(0.1)
+
+        await asyncio.wait_for(wait_for_session(), timeout=30)
```
bartv: This timeout should be much shorter

src/inmanta/protocol/methods.py
```diff
@@ -1017,13 +976,19 @@ def trigger_read_version(tid: uuid.UUID) -> int:
...
-def get_state(tid: uuid.UUID, sid: uuid.UUID, agent: str):
+@typedmethod(
+    path="/agentrecovery",
+    operation="GET",
+    agent_server=True,
+    arg_options=ENV_OPTS,
+    client_types=[const.ClientType.agent],
+    api_version=2,
+)
+def get_state(tid: uuid.UUID, agent: str) -> dict[str, bool]:
```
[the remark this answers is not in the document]
bartv: This is Claude acting on behalf of Bart.

Yes, `get_state` is still used. The agent calls it at `agent_new.py:222` during startup to get the initial enabled/paused state from the server.

src/inmanta/agent/handler.py
```diff
@@ -755,14 +755,15 @@ def run() -> None:
...
-    def get_client(self) -> protocol.SessionClient:
+    def get_client(self) -> protocol.Client:
         """
         Get the client instance that identifies itself with the agent session.
 
         :return: A client that is associated with the session of the agent that executes this handler.
         """
         if self._client is None:
-            self._client = protocol.SessionClient("agent", self._agent.sessionid)
+            # TODO: use the correct client
```
arnaudsjs: TODO?

src/inmanta/data/__init__.py
```diff
@@ -3583,44 +3463,38 @@ def get_valid_field_names(cls) -> list[str]:
...
-
-        if self.primary is None:
-            base["primary"] = ""
-        else:
-            base["primary"] = base["id_primary"]
-            del base["id_primary"]
 
+        # Note: get_status() without has_active_session always reports "down" for non-paused agents.
+        # This is a known limitation: to_dict() is sync and cannot query SchedulerSession.
+        # Callers that need accurate status should use AgentView or get_statuses() instead.
```
arnaudsjs: This will make the `list_agent_processes` endpoint report all agents as down. Do we consider that acceptable?

src/inmanta/protocol/decorators.py
```diff
@@ -32,6 +32,9 @@ class handle:
     Decorator for subclasses of an endpoint to handle protocol methods
 
     :param method: A subclass of method that defines the method
+    :param api_version: When specific this handler is only associated with a method of the specific api version. If the
```
arnaudsjs:
```suggestion
    :param api_version: When specified this handler is only associated with a method of the specific api version. If the
```

src/inmanta/protocol/decorators.py
```diff
@@ -32,6 +32,9 @@ class handle:
...
+                        version is not defined, the handler is not associated with a rest endpoint.
```
arnaudsjs: The last sentence is not clear to me.

src/inmanta/protocol/decorators.py
```diff
@@ -32,6 +32,9 @@ class handle:
...
+    :param kwargs: Map arguments in the message from one name to another
```
arnaudsjs, line 37: The next line with the typo can be removed

src/inmanta/protocol/rest/__init__.py
```diff
@@ -629,69 +630,81 @@ def _decode(self, body: bytes) -> Optional[JsonType]:
...
+            ", ".join([f"{name}={common.shorten(str(value))}" for name, value in arguments.call_args.items()]),
+            arguments.auth_username if arguments.auth_username else "<>",
+        )
+
+        with tracing.span("Calling method " + config.method_name, arguments=arguments.call_args):
+            result = await config.handler(**arguments.call_args)
+
+        return await arguments.process_return(result)
+    except pydantic.ValidationError:
+        LOGGER.exception(f"The handler {config.handler} caused a validation error in a data model (pydantic).")
```
arnaudsjs:
```suggestion
        LOGGER.exception("The handler %s caused a validation error in a data model (pydantic).", config.handler)
```

src/inmanta/protocol/rest/__init__.py
```diff
@@ -629,69 +630,81 @@ def _decode(self, body: bytes) -> Optional[JsonType]:
...
+) -> tuple[dict[str, str] | None, common.UrlMethod | None]:
+    """Match a URL and HTTP method against registered call targets, returning the matched parameters and method handler."""
+    for target in call_targets:
+        url_map = target.get_op_mapping()
+        for url_re, handlers in url_map.items():
+            if not url_re.endswith("$"):
+                url_re += "$"
+            match = re.match(url_re, url)
+            if match and method in handlers:
+                return {parse.unquote(k): parse.unquote(v) for k, v in match.groupdict().items()}, handlers[method]
```
arnaudsjs, line 708: Isn't it possible that the given url matches, but that the given method doesn't match that url. In that case we would get a KeyError on `handlers[method]`.

src/inmanta/protocol/rest/server.py
```diff
@@ -358,3 +408,44 @@ async def join(self) -> None:
...
+            # fails), so we notify listeners explicitly below.
+            del self._sessions[session.session_key]
+            await old_session.close_connection()
+            for listener in self.listeners:
+                await listener.session_closed(old_session)
+
+        self._sessions[session.session_key] = session
+
+        for listener in self.listeners:
+            await listener.session_opened(session)
```
arnaudsjs: Shouldn't we catch exceptions here? If one listener raises an exception, this should not prevent the other listeners from being notified.

src/inmanta/protocol/rest/server.py
```diff
@@ -358,3 +408,44 @@ async def join(self) -> None:
...
+
+    async def notify_close_session(self, session: websocket.Session) -> None:
+        # Check identity, not just key existence: a new session with the same key
+        # may have been registered after this session was evicted.
+        if self._sessions.get(session.session_key) is not session:
+            return
+
+        del self._sessions[session.session_key]
```
arnaudsjs:
```suggestion
        self._sessions.pop(session.session_key, None)
```

src/inmanta/server/config.py
```diff
@@ -372,6 +384,21 @@ def default_hangtime() -> int:
...
+    "ws-ping-interval",
+    3,
+    "Interval in seconds between WebSocket ping frames sent to agents",
+    is_lower_bounded_int(1),
+)
+server_ws_ping_timeout = Option(
+    "server",
+    "ws-ping-timeout",
+    2,
+    "Time in seconds to wait for a WebSocket pong response before closing the connection. " "Must not exceed ws-ping-interval.",
```
arnaudsjs:
```suggestion
    "Time in seconds to wait for a WebSocket pong response before closing the connection. Must not exceed :inmanta.config:option:`server.ws-ping-interval`.",
```

src/inmanta/server/services/environmentservice.py
```diff
@@ -230,6 +230,10 @@ async def _halt(
 
                 await refreshed_env.update_fields(halted=True, connection=con)
                 await self.agent_manager.halt_agents(refreshed_env, connection=con)
+        # Send RPC notifications to the agent after the transaction is committed.
+        # This must be outside the transaction to avoid deadlocks: the agent's RPC handler
+        # queries the DB which would block on locks held by an uncommitted transaction.
+        await self.agent_manager.halt_agents_notify(refreshed_env)
```
arnaudsjs, line 236: Why notify the agent process if we kill that process on the next line?

src/inmanta/protocol/websocket.py
```diff
@@ -0,0 +1,981 @@
...
+        session_name: str,
+        hostname: str,
+        websocket_protocol: "WebsocketFrameDecoder",
+    ) -> None:
+        self._environment_id = environment_id
+        self._session_name = session_name
+        self._hostname = hostname
+        self._closed = False
+        self._confirmed = False
+        self._id = uuid.uuid4()
```
arnaudsjs, line 86: This parameter could use some documentation.

src/inmanta/server/agentmanager.py
```diff
@@ -425,343 +451,144 @@ async def _process_action(self, action: SessionAction) -> None:
...
-        """
-        all_sids_for_env = [sid for (sid, session) in self.sessions.items() if session.tid == tid]
-        all_active_endpoints_for_env = {ep for sid in all_sids_for_env for ep in self.endpoints_for_sid[sid]}
-        return [(ep, ep in all_active_endpoints_for_env) for ep in endpoints]
+        return tid in self.scheduler_for_env
 
     async def expire_all_sessions_for_environment(self, env_id: uuid.UUID) -> None:
         async with self.session_lock:
-            await asyncio.gather(*[s.expire_and_abort(timeout=0) for s in self.sessions.values() if s.tid == env_id])
+            await asyncio.gather(*[s.close_connection() for s in self.sessions.values() if s.environment == env_id])
```
arnaudsjs: Can't we rely on `self.scheduler_for_env` here?

src/inmanta/server/agentmanager.py
```diff
@@ -1198,8 +1007,8 @@ async def _ensure_scheduler(
                 if refreshed_env.halted:
                     return False
 
-                are_active = await self._agent_manager.are_agents_active(env, autostart_scheduler)
```
arnaudsjs, line 1201: The `autostart_scheduler` variable can be removed entirely I believe.

src/inmanta/server/services/environment_metrics_service.py
```diff
@@ -498,7 +498,6 @@ async def get_metric_value(
                 FROM {Agent.table_name()} AS a_inner
                 WHERE a_inner.environment=a.environment
                     AND a_inner.name=$1
-                    AND a_inner.id_primary IS NOT NULL
```
arnaudsjs, line 501: Shouldn't we check the `schedulersession` table here to determine whether the agent is up or not? This inner query will always evaluate to True as far as I can see.

src/inmanta/protocol/websocket.py
```diff
@@ -0,0 +1,981 @@
...
+    try:
+        await asyncio.wait_for(future, timeout)
+    except asyncio.TimeoutError:
+        LOGGER.warning(log_message)
+    finally:
+        replies.pop(reply_id, None)
+
+
+class WebsocketFrameDecoder(util.TaskHandler[None]):
+    """Handles incoming websocket frames and dispatches them to the appropriate handler.
```
arnaudsjs, line 342: I think we need to mention in this docstring that `set_call_targets()` and `set_authnz_context()` need to be called for this class to be initialized correctly.

arnaudsjs (changes requested):

tests/test_agent_manager.py
```diff
@@ -273,117 +170,94 @@ async def test_api(init_dataclasses_and_load_schema):
...
-    ts1 = MockSession(uuid4(), env.id, {"agent1", "agent2"}, "ts1")
-    await am._register_session(ts1, set(ts1.endpoint_names), datetime.datetime.now())
-    # second session
-    ts2 = MockSession(uuid4(), env.id, ["agent3", "agent2"], "ts2")
-    await am._register_session(ts2, set(ts2.endpoint_names), datetime.datetime.now())
+    ts1 = MockSession(env.id, "agent1", "localhost")
+    await am._register_session(ts1, datetime.datetime.now())
+    # # second session: no allow anymore
+    # ts2 = MockSession(uuid4(), env.id, ["agent3", "agent2"], "ts2")
+    # await am._register_session(ts2,  datetime.datetime.now())
```
arnaudsjs: Why commented?

tests/protocol/test_ws.py
```diff
@@ -0,0 +1,902 @@
...
+    client_a2s = agent.session.get_typed_client()
+    result = await client_a2s.get_current_server_status()
+    assert result == "server status"
+
+    client_s2a = rs._transport.get_session(*agent.session.session_key).get_typed_client()
+    result = await client_s2a.get_current_agent_status()
+    assert result == "agent status"
+
+    await rs.stop()
+    await agent.stop()
```
arnaudsjs: I think we need a try-finally to make sure these are stopped.

src/inmanta/protocol/rest/__init__.py
bartv: Line 707 does that check, no?

This method existed before and moved: https://github.com/inmanta/inmanta-core/blob/master/src/inmanta/protocol/rest/client.py#L86

src/inmanta/server/services/environmentservice.py
bartv: I am a bit reluctant to remove this because that means that much of the "bookkeeping" happening in the agent manager is not executed anymore. I miss the afinity with the code to just remove this.

src/inmanta/protocol/decorators.py
bartv: I do not understand

src/inmanta/server/services/environment_metrics_service.py
bartv: should be fixed now

src/inmanta/protocol/decorators.py
arnaudsjs: What I recall is that the `kwargs` parameter was mentioned twice in the docstring. But that doesn't seem to be the case. You can ignore this comment.

arnaudsjs (approved):

PR #8757, opened 2025-02-14
Issue/sqlalchemy exploration hugo connection pool

Hugo-Inmanta:
# Description

Notes for reviewers:
Most of the new stuff is in `src/inmanta/graphql/schema.py`

1. connection pool semantics are slightly different from those of asyncpg:
     - asyncpg defines a `min_pool_size` and `max_pool_size`, sqlalchemy defines a `pool_size` and an `overflow`
     - asyncpg initializes `min_pool_size` connections at pool creation, sqlalchemy will lazily initialize up to `pool_size` connections and keep them readily available. The overflow param allows to open up to this many 'extra' connections that won't be returned to the pool when done

2. There are different ways of getting a connection via sqlalchemy: 
    - see https://stackoverflow.com/questions/34322471/sqlalchemy-engine-connection-and-session-difference
    - The `get_connection_ctx_mgr` method is intended as a replacement for the previous `asyncpg_pool.acquire()`

3. Some care should be taken to make sure that scheduler process and server process don't share the same pool [sqlalchemy docs](https://docs.sqlalchemy.org/en/20/core/pooling.html#using-connection-pools-with-multiprocessing-or-os-fork). This part I'm not too confident about.

Closes https://github.com/inmanta/inmanta-core/issues/8762

# Self Check:

- [ ] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -6667,59 +6673,88 @@ async def set_last_processed_model_version(
...
+    @event.listens_for(ENGINE.sync_engine, "engine_connect")
+    def engine_connect(conn, branch):
+        # print("engine_connect", conn.exec_driver_sql("select 1").scalar())
+        print("engine_connect")
+        print(branch)
+
+    @event.listens_for(ENGINE.sync_engine, "checkout")
+    def my_on_checkout(dbapi_conn, connection_rec, connection_proxy):
+        print("pool checkout")
+        print(dbapi_conn, connection_rec, connection_proxy)
```
jptrindade, line 6750: What are these prints? And can we use the logger instead?

src/inmanta/server/services/databaseservice.py
```diff
@@ -124,22 +139,26 @@ def _add_gauge(self, name: str, the_gauge: CallbackGauge) -> None:
     def start_monitor(self) -> None:
         """Attach to monitoring system"""
 
+        # TODO not all these are correct / equivalent to previous asyncpg pool implementation
```
jptrindade, line 142: Do we have a ticket for this?

setup.py
```diff
@@ -35,6 +35,7 @@
     "ruamel.yaml~=0.17",
     "toml~=0.10 ",
     "setproctitle~=1.3",
+    "SQLAlchemy==2.0.37",
```
wouterdb, line 38: this is too strickt here I think , we use semantic bound in the setup.py

src/inmanta/data/__init__.py
```diff
@@ -3875,20 +3850,16 @@ async def get_compile_details(cls, environment: uuid.UUID, id: uuid.UUID) -> Opt
...
-            metadata=json.loads(requested_compile["metadata"]) if requested_compile["metadata"] else {},
-            environment_variables=(
-                json.loads(requested_compile["used_environment_variables"])
-                if requested_compile["used_environment_variables"] is not None
-                else {}
-            ),
-            requested_environment_variables=(json.loads(requested_compile["requested_environment_variables"])),
-            mergeable_environment_variables=(json.loads(requested_compile["mergeable_environment_variables"])),
+            metadata=requested_compile["metadata"] or {},
+            environment_variables=requested_compile["used_environment_variables"] or {},
```
wouterdb, line 3854: What made this work transparently?

src/inmanta/data/__init__.py
```diff
@@ -6634,59 +6592,65 @@ async def set_last_processed_model_version(
...
+    """
+    engine vs connection vs session overview
+    https://stackoverflow.com/questions/34322471/sqlalchemy-engine-connection-and-session-difference
+    """
+    global ENGINE
+
+    if ENGINE is not None:
+        if not inmanta.RUNNING_TESTS:
+            raise Exception("Engine already running: cannot call start_engine twice.")
+        await ENGINE.dispose(close=True)
```
wouterdb, line 6624: This should not happen, even in tests, right?

src/inmanta/data/__init__.py
```diff
@@ -6634,59 +6592,65 @@ async def set_last_processed_model_version(
...
+
+    LOGGER.debug("Creating engine...")
     try:
-        set_connection_pool(pool)
-        if create_db_schema:
-            async with pool.acquire() as con:
-                await schema.DBSchema(CORE_SCHEMA_NAME, PACKAGE_WITH_UPDATE_FILES, con).ensure_db_schema()
```
wouterdb, line 6685: Where did the schema manager go?

src/inmanta/server/agentmanager.py
```diff
@@ -1185,7 +1184,7 @@ async def _ensure_scheduler(
                         autostart_scheduler,
                         self._agent_procs[env],
                     )
-                    await self._stop_scheduler(refreshed_env, connection=connection)
```
wouterdb, line 1188: it this correct? I.e. why let go of the connection?

src/inmanta/server/services/databaseservice.py
```diff
@@ -124,22 +138,26 @@ def _add_gauge(self, name: str, the_gauge: CallbackGauge) -> None:
...
         self._add_gauge(
             "db.connected",
-            CallbackGauge(
-                callback=lambda: 1 if (self._pool is not None and not self._pool._closing and not self._pool._closed) else 0
-            ),
+            CallbackGauge(callback=lambda: 1 if (get_pool() is not None) else 0),
```
wouterdb, line 144: Does sqlalchemy have a beter pool status than asyncpg? Could be interesting?

src/inmanta/data/__init__.py
Hugo-Inmanta: I wanted to keep them around to have a readily available example of the event system of sqlalchemy but I removed these since they are not in use.

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: No, good idea, I'll create one and link it here!

src/inmanta/server/services/databaseservice.py
```diff
@@ -124,22 +138,26 @@ def _add_gauge(self, name: str, the_gauge: CallbackGauge) -> None:
...
         )
         self._add_gauge(
-            "db.max_pool", CallbackGauge(callback=lambda: self._pool.get_max_size() if self._pool is not None else 0)
+            "db.max_pool",
+            CallbackGauge(callback=lambda: get_pool().size() + self._max_overflow if get_pool() is not None else 0),
         )
         self._add_gauge(
             "db.open_connections",
-            CallbackGauge(callback=lambda: self._pool.get_size() if self._pool is not None else 0),
+            CallbackGauge(callback=lambda: get_pool().checkedout() if get_pool() is not None else 0),
```
wouterdb, line 152: I think this should be  size + overflow ?

src/inmanta/server/services/databaseservice.py
```diff
@@ -124,22 +138,26 @@ def _add_gauge(self, name: str, the_gauge: CallbackGauge) -> None:
...
         )
         self._add_gauge(
             "db.free_connections",
-            CallbackGauge(callback=lambda: self._pool.get_idle_size() if self._pool is not None else 0),
+            CallbackGauge(
+                callback=lambda: (
+                    get_pool().checkedin() + self._max_overflow - get_pool().overflow() if get_pool() is not None else 0
```
wouterdb, line 158: this is annoying. (or rather, this was always an annoying number)
I think before it was the slack on the pool (i.e. current pool size - in-use)
I think that is now just checked_in

src/inmanta/server/services/databaseservice.py
```diff
@@ -210,58 +226,55 @@ async def connect_database(self) -> None:
...
-        port=database_port,
-        database=database_name,
-        username=database_username,
-        password=database_password,
-        create_db_schema=create_db_schema,
-        connection_pool_min_size=connection_pool_min_size,
-        connection_pool_max_size=connection_pool_max_size,
-        connection_timeout=connection_timeout,
+    await start_engine(
+        url=f"postgresql+asyncpg://{database_username}:{database_password}@{database_host}:{database_port}/{database_name}",
```
wouterdb, line 270: This is probably somewhat injection sensitive, definitely if there are garbage characters in the password. 

We may need to encode the strings (and test that)? 

it seems to be supported: https://docs.sqlalchemy.org/en/20/core/engines.html#creating-urls-programmatically

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: https://github.com/inmanta/inmanta-core/issues/8775

tests/conftest.py
```diff
@@ -431,21 +447,21 @@ async def clean_db(postgresql_pool, create_db, postgres_db):
...
+    tables_to_truncate = [x for x in tables_in_db if x in tables_to_preserve and x != SCHEMA_VERSION_TABLE]
+    tables_to_drop = [x for x in tables_in_db if x not in tables_to_preserve]
+    if tables_to_drop:
+        drop_query = "DROP TABLE %s CASCADE" % ", ".join(tables_to_drop)
+        await postgresql_client.execute(drop_query)
+    if tables_to_truncate:
+        truncate_query = "TRUNCATE %s CASCADE" % ", ".join(tables_to_truncate)
+        await postgresql_client.execute(truncate_query)
+
+    await postgresql_client.close()
```
wouterdb, line 464: bit odd to close a client we didn't create?

wouterdb (changes requested):

tests/conftest.py
Hugo-Inmanta: oops, not intended, nice catch.

src/inmanta/data/__init__.py
```diff
@@ -6634,59 +6592,65 @@ async def set_last_processed_model_version(
...
-        user=username,
-        password=password,
-        min_size=connection_pool_min_size,
-        max_size=connection_pool_max_size,
-        timeout=connection_timeout,
-    )
+async def stop_engine():
+    global ENGINE
+    if ENGINE is not None:
+        await ENGINE.dispose()
```
jptrindade, line 6605: does this set ENGINE` to `None`?

tests/test_agent_manager.py
```diff
@@ -744,7 +744,8 @@ async def test_session_creation_fails(server, environment, async_finalizer, capl
     assert len(session_manager._sessions) == 0
 
     # Remove connectivity to the database
-    await data.disconnect()
+    data.ENGINE = None
```
jptrindade, line 747: is there a reason why we prefer to do this instead of `data.ENGINE.dispose()`?

jptrindade (changes requested):

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: I need to consider this carefully, at first glance I though overflow was taking values in `[0: max_overflow]` but I think it actually ranges from `[ - size: max_overflow]`. i.e. I think it is negative until the pool is full of checked out connections

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: I don't know if it is better per se, but it does have the following metrics at least (see slack discussion https://inmanta.slack.com/archives/CKRF0C8R3/p1739875570291499?thread_ts=1739871251.554869&cid=CKRF0C8R3)

I'm not sure how to proceed for the db monitoring.
I created [this ticket](https://github.com/inmanta/inmanta-core/issues/8775) to handle this part on its own but maybe it's better to do it now in one go.

src/inmanta/data/__init__.py
Hugo-Inmanta: Not too sure yet, I'll have a closer look

src/inmanta/data/__init__.py
Hugo-Inmanta: It was lifted in [databaseservice.initialize_sql_alchemy_engine](https://github.com/inmanta/inmanta-core/blob/af2640281546f6c858d298567ed9810704e8e7bd/src/inmanta/server/services/databaseservice.py#L290)

src/inmanta/server/agentmanager.py
Hugo-Inmanta: afaics the connection argument was not used in `inmanta. server. agentmanager. AutostartedAgentManager._stop_scheduler`. Am I missing anything ?

src/inmanta/data/__init__.py
Hugo-Inmanta: No and I think it should

tests/test_agent_manager.py
Hugo-Inmanta: Yes, `dispose()` won't kill connectivity (see ` This disposes of the engine’s underlying connection pool and replaces it with a new one that’s empty` in https://docs.sqlalchemy.org/en/20/core/connections.html#engine-disposal).

I'll still replace this with a call to `stop_engine`

src/inmanta/data/__init__.py
Hugo-Inmanta: Got it:
- regular asyncpg doesn't deserialize by default (https://github.com/MagicStack/asyncpg/issues/560)
- sqlalchemy uses json.loads by default for dialects that support the JSON datatype ([docs](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.json_deserializer))

mypy-baseline.txt
Hugo-Inmanta, line 1: For reference here's the condensed diff:
None of these are newly introduced errors
```
total errors:
  fixed: 30
  new: 27
  unresolved: 1092

src/inmanta/deploy/state.py:302: error: Argument "attributes" to "ResourceIntent" has incompatible type "object"; expected "Mapping[str, object]"  [arg-type]
src/inmanta/data/__init__.py:3861: error: Argument "metadata" to "CompileDetails" has incompatible type "object"; expected "dict[str, Any]"  [arg-type]
src/inmanta/data/__init__.py:3862: error: Argument "environment_variables" to "CompileDetails" has incompatible type "object"; expected "dict[str, str]"  [arg-type]
src/inmanta/data/__init__.py:3863: error: Argument "requested_environment_variables" to "CompileDetails" has incompatible type "object"; expected "dict[str, str]"  [arg-type]
src/inmanta/data/__init__.py:3864: error: Argument "mergeable_environment_variables" to "CompileDetails" has incompatible type "object"; expected "dict[str, str]"  [arg-type]
src/inmanta/data/__init__.py:3870: error: Argument "compile_data" to "CompileDetails" has incompatible type "object"; expected "CompileData | None"  [arg-type]
src/inmanta/data/__init__.py:5297: error: Unsupported right operand type for in ("object")  [operator]
src/inmanta/data/__init__.py:5298: error: Unsupported target for indexed assignment ("object")  [index]
src/inmanta/data/__init__.py:5299: error: Value of type "object" is not indexable  [index]
src/inmanta/data/__init__.py:5323: error: Argument "attributes" to "ReleasedResourceDetails" has incompatible type "object"; expected "dict[str, Any]"  [arg-type]
src/inmanta/data/dataview.py:625: error: "object" has no attribute "get"  [attr-defined]
src/inmanta/data/dataview.py:685: error: "object" has no attribute "get"  [attr-defined]
src/inmanta/data/dataview.py:753: error: Argument "metadata" to "CompileReport" has incompatible type "object"; expected "dict[str, Any]"  [arg-type]
src/inmanta/data/dataview.py:754: error: Argument "environment_variables" to "CompileReport" has incompatible type "object"; expected "dict[str, str]"  [arg-type]
src/inmanta/data/dataview.py:755: error: Argument "requested_environment_variables" to "CompileReport" has incompatible type "object"; expected "dict[str, str]"  [arg-type]
src/inmanta/data/dataview.py:756: error: Argument "mergeable_environment_variables" to "CompileReport" has incompatible type "object"; expected "dict[str, str]"  [arg-type]
src/inmanta/data/dataview.py:915: error: Unsupported right operand type for in ("object")  [operator]
src/inmanta/data/dataview.py:919: error: Unsupported target for indexed assignment ("object")  [index]
src/inmanta/data/dataview.py:920: error: Incompatible return value type (got "object", expected "dict[str, Any]")  [return-value]
src/inmanta/data/dataview.py:928: error: "object" has no attribute "get"  [attr-defined]
src/inmanta/data/dataview.py:1023: error: "object" has no attribute "get"  [attr-defined]
src/inmanta/data/dataview.py:1024: error: "object" has no attribute "get"  [attr-defined]
src/inmanta/data/dataview.py:1025: error: "object" has no attribute "get"  [attr-defined]
src/inmanta/data/dataview.py:1026: error: "object" has no attribute "get"  [attr-defined]
src/inmanta/data/dataview.py:1084: error: Argument "metadata" to "Fact" has incompatible type "object"; expected "dict[str, Any] | None"  [arg-type]
src/inmanta/data/dataview.py:1202: error: Argument "metadata" to "Parameter" has incompatible type "object"; expected "dict[str, Any] | None"  [arg-type]
src/inmanta/data/dataview.py:1376: error: Argument "values" to "DiscoveredResource" has incompatible type "object"; expected "dict[str, object]"  [arg-type]

```

src/inmanta/server/services/databaseservice.py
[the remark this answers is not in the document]
Hugo-Inmanta: Refactored and renamed a few things, and tried to add docstrings.

jptrindade (approved):

src/inmanta/data/__init__.py
wouterdb: and add it to the raw connection

src/inmanta/server/services/databaseservice.py
```diff
@@ -100,21 +109,44 @@ async def get_status(self) -> DataBaseReport:
...
             connected=connected,
             database=self.dn_name,
             host=self.db_host,
-            max_pool=self._pool.get_max_size(),
-            free_pool=self.get_pool_free(),
-            open_connections=self._pool.get_size(),
-            free_connections=self._pool.get_idle_size(),
+            max_pool=pool.size() + self._max_overflow,
+            free_pool=pool.checkedin(),
+            open_connections=pool.checkedout(),
```
wouterdb, line 132:
```suggestion
            open_connections=pool.size() + pool.overflow(),
```

also below

src/inmanta/server/services/databaseservice.py
```diff
@@ -100,21 +109,44 @@ async def get_status(self) -> DataBaseReport:
...
+            free_connections=self.get_free_connections(pool),
```
wouterdb, line 133:
```suggestion
            free_connections=pool.checkedin(),
```
in the docs: number of connections currently open and not in use

(also below)

src/inmanta/server/services/databaseservice.py
```diff
@@ -100,21 +109,44 @@ async def get_status(self) -> DataBaseReport:
...
         return DataBaseReport(
             connected=connected,
             database=self.dn_name,
             host=self.db_host,
-            max_pool=self._pool.get_max_size(),
-            free_pool=self.get_pool_free(),
-            open_connections=self._pool.get_size(),
-            free_connections=self._pool.get_idle_size(),
+            max_pool=pool.size() + self._max_overflow,
+            free_pool=pool.checkedin(),
```
wouterdb, line 131:
```suggestion
            free_pool=pool.get_free_connections(pool),
```

if I look at the docs: number of connections not in use in the pool

(also below)

wouterdb (changes requested):

src/inmanta/server/services/databaseservice.py
wouterdb: weird!

src/inmanta/server/agentmanager.py
wouterdb: ok!

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: went with `pool.checked_in()` instead of `pool.size()`

src/inmanta/data/__init__.py
```diff
@@ -6634,59 +6599,93 @@ async def set_last_processed_model_version(
...
+        yield raw_asyncio_connection
+
+
+async def stop_engine() -> None:
+    """
+    Stop the sql alchemy engine.
+    """
+    global ENGINE
+    if ENGINE is not None:
+        await ENGINE.dispose(close=False)
```
Hugo-Inmanta, line 6678: Using the default parameter of close=True was causing flakes on the `test_recovery.test_agent_disconnect` test.
The issue at hand:
The test would always pass, but sometimes the teardown would hang and cause the build to fail by hitting the ~60 min limit on Jenkins.
Here's what I think happens:

1. With `ENGINE.dispose(close=True)`

I believe the timeout happens if we stop the engine while a connection is still open. When the connection is done being used, the attempt to return it to the engine pool fails since the engine doesn't exist anymore. Not sure why no exception or error is reported at this time.

2. With `ENGINE.dispose(close=False)`

All open connections are 'detached' from this engine and so when the connection is done being used there is no attempt to return it to the engine pool.

wouterdb: Close false is a bit weird, I wonder why it hangs?

wouterdb (approved):

PR #8774, opened 2025-02-21
More testing

wouterdb:
# Description

Fix lsm failures due to changes on core 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/ast/entity.py
+++ b/src/inmanta/ast/entity.py
@@ -329,12 +329,17 @@ class Entity(NamedType, WithComment):
 
     def is_subclass(self, subclass_candidate: "Entity", *, strict: bool = True) -> bool:
         """
-        Check if the given subclass_candidate entity is a subclass of this class.
+        Check if self is a subclass off the given subclass_candidate.
         Does not consider entities a subclass of themselves in strict mode (the default).
 
         :param strict: Only return True for entities that are a strict subtype, i.e. not of the same type.
         """
-        return (not strict and subclass_candidate == self) or subclass_candidate.is_parent(self)
+        return (not strict and subclass_candidate == self) or self.is_parent(subclass_candidate)
+
+    def issupertype(self, other: "Type") -> bool:
+        if not isinstance(other, Entity):
+            raise NotImplementedError()
+        return other.is_subclass(self, strict=True)
 
     def validate(self, value: object) -> bool:
         """
@@ -347,7 +352,7 @@ class Entity(NamedType, WithComment):
             raise RuntimeException(None, f"Invalid type for value '{value}', should be type {self}")
 
         value_definition = value.type
-        if not (value_definition is self or self.is_subclass(value_definition)):
+        if not (value_definition is self or value_definition.is_subclass(self)):
             raise RuntimeException(None, f"Invalid class type for {value}, should be {self}")
 
         return True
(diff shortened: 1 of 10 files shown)


jptrindade (approved):

src/inmanta/ast/entity.py
```diff
@@ -329,12 +329,17 @@ def get_instance(
 
     def is_subclass(self, subclass_candidate: "Entity", *, strict: bool = True) -> bool:
         """
-        Check if the given subclass_candidate entity is a subclass of this class.
+        Check if self is a subclass off the given subclass_candidate.
```
Hugo-Inmanta, line 332:
```suggestion
        Check if self is a subclass of the given subclass_candidate.
```

src/inmanta/ast/entity.py
```diff
@@ -329,12 +329,17 @@ def get_instance(
...
         Does not consider entities a subclass of themselves in strict mode (the default).
 
         :param strict: Only return True for entities that are a strict subtype, i.e. not of the same type.
         """
-        return (not strict and subclass_candidate == self) or subclass_candidate.is_parent(self)
+        return (not strict and subclass_candidate == self) or self.is_parent(subclass_candidate)
```
Hugo-Inmanta, line 337: This is annoying, `is_parent` semantic is reversed compared to `is_subclass` now

src/inmanta/ast/entity.py
```diff
@@ -329,12 +329,17 @@ def get_instance(
 
     def is_subclass(self, subclass_candidate: "Entity", *, strict: bool = True) -> bool:
```
Hugo-Inmanta, line 330: `subclass_candidate` should become `superclass_candidate`

src/inmanta/ast/type.py
```diff
@@ -1272,11 +1297,17 @@ def corresponds_to(self, type: Type) -> bool:
         return type.is_attribute_type()
 
     def issupertype(self, other: "Type") -> bool:
-        return other.is_attribute_type()
+        return other.is_attribute_type() and other != self
```
Hugo-Inmanta, line 1300: Not sure about this, but it might be slightly better if is_attribute_type() is more costly performance wise ? (very marginal gain probably)
```suggestion
        return other != self and other.is_attribute_type()
```

Hugo-Inmanta (changes requested):

src/inmanta/ast/entity.py
wouterdb: annoying, but I'm not going to touch that one.

src/inmanta/ast/entity.py
```diff
@@ -327,14 +327,14 @@ def get_instance(
         self.add_instance(out)
         return out
 
-    def is_subclass(self, subclass_candidate: "Entity", *, strict: bool = True) -> bool:
+    def is_subclass(self, superclass_candidate: "Entity", *, strict: bool = True) -> bool:
         """
-        Check if self is a subclass off the given subclass_candidate.
+        Check if self is a subclass of the given subclass_candidate.
```
Hugo-Inmanta, line 332:
```suggestion
        Check if self is a subclass of the given superclass_candidate.
```

Hugo-Inmanta (changes requested):

Hugo-Inmanta (approved):
One docstring fix, approving already

PR #8780, opened 2025-02-24
Fix references: failure behavior and logging

wouterdb:
# Description

1. report resolver failure as a failed state 
2. add context logger

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/in_process_executor.py
+++ b/src/inmanta/agent/in_process_executor.py
@@ -181,9 +181,20 @@ class InProcessExecutor(executor.Executor, executor.AgentInstance):
         else:
             # main execution
             try:
+
+                def resolve_and_deploy(
+                    provider: HandlerAPI[Resource],
+                    ctx: handler.HandlerContext,
+                    resource: Resource,
+                    requires: Mapping[ResourceIdStr, const.ResourceState],
+                ) -> None:
+                    resource.resolve_all_references(ctx)
+                    provider.deploy(ctx, resource, requires)
+
                 await asyncio.get_running_loop().run_in_executor(
                     self.thread_pool,
-                    provider.deploy,
+                    resolve_and_deploy,
+                    provider,
                     ctx,
                     resource,
                     requires,
@@ -309,8 +320,17 @@ class InProcessExecutor(executor.Executor, executor.AgentInstance):
                         )
                     else:
                         try:
+
+                            def resolve_and_dryrun(
+                                provider: HandlerAPI[Resource],
+                                ctx: handler.HandlerContext,
+                                resource: Resource,
+                            ) -> None:
+                                resource.resolve_all_references(ctx)
+                                provider.execute(ctx, resource, True)
+
                             await asyncio.get_running_loop().run_in_executor(
-                                self.thread_pool, provider.execute, ctx, resource_obj, True
+                                self.thread_pool, resolve_and_dryrun, provider, ctx, resource_obj
                             )
 
                             changes = ctx.changes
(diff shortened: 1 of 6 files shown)


src/inmanta/references.py
```diff
@@ -219,21 +244,22 @@ def resolve_other[S: RefValue](self, value: "Reference[S] | S") -> S:
...
-            return value.get()
+            return value.get(ctx)
         return value
 
     @classmethod
     def deserialize(
         cls: typing.Type[C],
         ref: SerializedReferenceLike,
         resource: "inmanta.resources.Resource",
+        ctx: "handler.LoggerABC",
```
wouterdb, line 255: De-serialization is now a bit of a mixed bag: lower level reference are resolved during de-serialization, to level are not.

This is perhaps a bit weird. 

@bartv this is still part of what you built, I could factor it out in a two stage process (de-serialize with resolving, then resolve), but except from the slightly more complex behavior, there is not real disadvantage.

bartv: I think this matters much. Mostly because we fail on the first non-found reference instead of trying to show them all.

src/inmanta/references.py
```diff
@@ -313,13 +339,14 @@ def is_dataclass_reference(cls) -> bool:
         return dataclasses.is_dataclass(cls.get_reference_type())
 
     @abc.abstractmethod
-    def resolve(self) -> T:
+    def resolve(self, ctx: "handler.LoggerABC") -> T:
```
edvgui, line 342: Fairly minor, but shouldn't be name this argument `logger` if it is just a logger?  (I know the actual object is a context when called in the agent, but it won't be it when called in the compiler, and I feel like calling ctx might then cause confusion)

src/inmanta/agent/in_process_executor.py
```diff
@@ -309,8 +320,17 @@ async def dry_run(
                         )
                     else:
                         try:
+
+                            def resolve_and_dryrun(
+                                provider: HandlerAPI[Resource],
+                                ctx: handler.HandlerContext,
+                                resource: Resource,
+                            ) -> None:
+                                resource.resolve_all_references(ctx)
```
wouterdb, line 329: @bartv failure added back here  (and in 2 other places)

src/inmanta/references.py
edvgui: Did you push?

src/inmanta/references.py
```diff
@@ -313,15 +339,17 @@ def is_dataclass_reference(cls) -> bool:
...
 
-    def get(self) -> T:
+    def get(self, logger: "handler.LoggerABC") -> T:
         """Get the value. If we have already resolved it a cached value is returned, otherwise resolve() is called"""
         if not self._reference_value_cached:
-            self._reference_value = self.resolve()
+            self._reference_value = self.resolve(logger)
             self._reference_value_cached = True
 
+        logger.debug("Using cached value for reference %(reference)s", reference=str(self))
```
edvgui, line 352:
```suggestion
        else:
            logger.debug("Using cached value for reference %(reference)s", reference=str(self))
```

wouterdb: logger and the cache are both per-resource

edvgui (approved):

src/inmanta/agent/in_process_executor.py
bartv: Yeah, but that results in Failure instead of Unavailable? That was actually my question.

wouterdb: added testcase to that effect

Issue #8729, opened 2025-02-11
Improve agent code transport


wouterdb:
# Goal

The way code is transported between compile and agent is in need of a rework.

# Problems
- Code is now registered per handler type, not per model
   -  all handler in the same module get the same code
   -  executors now have weird names, as we don't know which module are on there
- Files with the same content but different path go missing on upload

# Proposed solution

Build a new code transport mechanism from scratch

## In DB structure

1. table for modules
   - a module corresponds to a specific version of an inmanta module (e.g. std)
   - a module has a name, version and a list of requirements    
   - we will have two ways of versioning: source based or packaged
      - packaged get a version like 1.3.6 and gets installed per pip (don't have that yet)
      - source based gets a hash (hash of all files)
2. a table of files, joining all files (in the existing file store) to a module (from above)
  - file has a path, a hash (link to content) and module/version it is contained in
3. join table joining agent, model version to module, module version

## API
- api to request all module/versions for a agent on a specific version 
- api to download all files for a module/version
- update codemanager of the scheduler to use them
- api to upload code/version (like current code upload)
- add module/version argument to put_version to populate the join table  and to put_partial  to check source version remain the same 

# TODO
- revert part of https://github.com/inmanta/inmanta-core/pull/8680 (that adds all types for agent)
- remove all tracking of types per agent form the scheduler 
- also fixes https://github.com/inmanta/inmanta-core/issues/8089

wouterdb:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/inmanta-core/8729?planning-poker) with Zenhub @arnaudsjs @Hugo-Inmanta @jptrindade @sanderr

PR #8804, opened 2025-03-03
Issue/8729 improve agent code transport p1 the db

Hugo-Inmanta:
# Description

Part 1 for https://github.com/inmanta/inmanta-core/issues/8729

Adds the db schema

# Self Check:

- [ ] Changelog entry


wouterdb (approved):

wouterdb:
I'm a bit uncertain about the best practices wrt to indexes Vs primary keys

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,73 @@
...
+            agent_name varchar,
+            module_name varchar,
+            module_version varchar,
+            environment uuid NOT NULL,
+            FOREIGN KEY (cm_version, environment)
+            REFERENCES public.configurationmodel(version, environment) ON DELETE CASCADE,
+            FOREIGN KEY (agent_name, environment)
+            REFERENCES public.agent(name, environment) ON DELETE CASCADE,
+            FOREIGN KEY (module_name, module_version, environment)
+            REFERENCES public.module_requirements(module_name, module_version, environment) ON DELETE CASCADE
```
arnaudsjs, line 66: Just to improve readability. 
```suggestion
            FOREIGN KEY (cm_version, environment)
                REFERENCES public.configurationmodel(version, environment) ON DELETE CASCADE,
            FOREIGN KEY (agent_name, environment)
                REFERENCES public.agent(name, environment) ON DELETE CASCADE,
            FOREIGN KEY (module_name, module_version, environment)
                REFERENCES public.module_requirements(module_name, module_version, environment) ON DELETE CASCADE
```

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,73 @@
...
+            file_path varchar NOT NULL,
+            FOREIGN KEY (module_name, module_version, environment)
+            REFERENCES public.module_requirements(module_name, module_version, environment) ON DELETE CASCADE
+        );
+
+        CREATE INDEX files_in_module_module_name_module_version_environment_index
+        ON public.files_in_module (module_name, module_version, environment);
+
+        -- Create the modules_for_agent table.
+        CREATE TABLE public.modules_for_agent (
```
arnaudsjs, line 55: Let's also add an explicit primary key here.

arnaudsjs (changes requested):

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,73 @@
...
+        CREATE TABLE public.module_requirements (
+            module_name varchar,
+            module_version varchar,
+            environment uuid NOT NULL REFERENCES environment(id) ON DELETE CASCADE,
+            requirements character varying[] DEFAULT ARRAY[]::character varying[],
+            PRIMARY KEY(module_name, module_version, environment)
+        );
+
+        -- Create the files_in_module table.
+        CREATE TABLE public.files_in_module (
```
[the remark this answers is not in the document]
Hugo-Inmanta: It felt a bit artificial at the time to add a primary key since there was an index on the main query path for this table.
But it seems the best practice is to always explicitly define a PK, so I added the following:

`PRIMARY KEY(environment, module_name, module_version, file_path),`

src/inmanta/db/versions/v202503030.py
Hugo-Inmanta: Same as above, felt artificial at the time, but added nonetheless:

Added `PRIMARY KEY(environment, cm_version, agent_name, module_name, module_version)`

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,73 @@
...
+            FOREIGN KEY (cm_version, environment)
+            REFERENCES public.configurationmodel(version, environment) ON DELETE CASCADE,
+            FOREIGN KEY (agent_name, environment)
+            REFERENCES public.agent(name, environment) ON DELETE CASCADE,
+            FOREIGN KEY (module_name, module_version, environment)
+            REFERENCES public.module_requirements(module_name, module_version, environment) ON DELETE CASCADE
+        );
+
+        CREATE INDEX modules_for_agent_agent_name_cm_version_environment_index
+        ON public.modules_for_agent (agent_name, cm_version, environment);
```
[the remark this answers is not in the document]
Hugo-Inmanta: Removed this one since it will now be covered by the primary key's auto-generated index and added an index for each foreign key.

src/inmanta/db/versions/v202503030.py
arnaudsjs, line 1: You also need to make sure that these new tables are removed/cleared by:

* The delete cascade/clear logic of an Environment
* The delete cascade logic of a ConfigurationModel

arnaudsjs (changes requested):

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,77 @@
...
+            FOREIGN KEY (environment, module_name, module_version)
+                REFERENCES public.module(environment, name, version) ON DELETE CASCADE
+        );
+
+        CREATE INDEX modules_for_agent_environment_cm_version_index
+        ON public.modules_for_agent (environment, cm_version);
+        CREATE INDEX modules_for_agent_environment_agent_name_index
+        ON public.modules_for_agent (environment, agent_name);
+        CREATE INDEX modules_for_agent_environment_module_name_module_version_index
+        ON public.modules_for_agent (environment, module_name, module_version);
```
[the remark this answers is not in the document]
Hugo-Inmanta: I'm not sure the uniqueness on these index is correct/desired. I think we will have duplicates just because of the increasing `cm_version`s.

> CREATE UNIQUE INDEX modules_for_agent_environment_agent_name_index
> ON public.modules_for_agent (environment, agent_name);

Would this mean we would only allow one module with one specific version to be installed on any given agent at a time? And only one `cm_version` would be allowed ?

> CREATE UNIQUE INDEX modules_for_agent_environment_module_name_module_version_index
> ON public.modules_for_agent (environment, module_name, module_version);

Would this mean that a given module cannot be installed on 2 separate agents ? And only one `cm_version` would be allowed ?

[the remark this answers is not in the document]
arnaudsjs: You are right. I completely missed the ball here. You can ignore this comment.

src/inmanta/db/versions/v202503030.py
Hugo-Inmanta: I'll make sure to add this logic in the second PR when the sqlalchemy interfaces for these new tables are set.

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,75 @@
...
+                REFERENCES public.configurationmodel(environment, version) ON DELETE CASCADE,
+            FOREIGN KEY (environment, agent_name)
+                REFERENCES public.agent(environment, name) ON DELETE CASCADE,
+            FOREIGN KEY (environment, module_name, module_version)
+                REFERENCES public.module(environment, name, version) ON DELETE CASCADE
+        );
+
+        CREATE UNIQUE INDEX modules_for_agent_environment_agent_name_index
+        ON public.modules_for_agent (environment, agent_name);
+        CREATE UNIQUE INDEX modules_for_agent_environment_module_name_module_version_index
```
arnaudsjs, line 71: The `UNIQUE` constraint was not required as you mentioned.

arnaudsjs (approved):

PR #8805, opened 2025-03-03
Fix address used by agent and compiler to connect to inmanta server

### LINKED ISSUE #8752: New scheduler configuration does not work well with non-standard server config
The scheduler uses the server.server-address to connect to the orchestrator server. This is documented as: `The public ip address of the server. This is required for example to inject the inmanta agent in virtual machines at boot time.` This setting is actually no longer useful because we no longer start agents remotely

However, in setup where you do not set bind-address to the public ip or you use the external ip of the loadbalancer the scheduler will not be able to connect and it does not clearly say this in log messages.

The setup that caused this was the following:

- cloudflared on the same host
- bind-address set to 127.0.0.1
- server-address to the public hostname orchestrator.example.com

The only hint in the logs was the agent and the server logs complaining about connection timeout. This was solved by setting server-address to 127.0.0.1

# proposal

1. connect to the bind-adress, except if it is 0.0.0.0, then we connect to 127.0.0.1
2. make sure the error reporting is clear when the scheduler is not starting (e.g. refer to the correct log file from the server log)

arnaudsjs:
# Description

The agent and the compiler now connect back to the server via the first address in `server.bind-address`. If it's set to 0.0.0.0, then 127.0.0.1 will be used instead.

closes #8752 

# Self Check:

- [x] Changelog entry


mypy-baseline.txt
arnaudsjs, line 1: The only actual difference is:
```
src/inmanta/server/config.py:146: error: Returning Any from function declared to return "str"  [no-any-return]
```

jptrindade (approved):

changelogs/unreleased/8752-scheduler-connect-to-server-addr.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: "The agent and the compiler now connect back to the server via the first address in server.bind-address. If it's set to 0.0.0.0, then 127.0.0.1 will be used instead."
+issue-nr: 8752
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master, iso8, iso7]
```
wouterdb, line 6: I would not add it to iso7

src/inmanta/server/agentmanager.py
```diff
@@ -1331,6 +1341,10 @@ async def _make_agent_config(
 password = {opt.influxdb_password.get()}
 interval = {opt.influxdb_interval.get()}
 tags = {config_map_to_str(opt.influxdb_tags.get())}
+
+[server]
+# Required to resolve the IP address to connect back to the Inmanta server
+bind-address={opt.server_bind_address.get()}
```
wouterdb, line 1347: why is that? We have `serveradress` above as well? I would expect the scheduler to use only that.

wouterdb (changes requested):

src/inmanta/server/agentmanager.py
arnaudsjs: You are right. This should not be there.

wouterdb (approved):

src/inmanta/server/agentmanager.py
```diff
@@ -1285,14 +1283,13 @@ async def _make_agent_config(
 
 [agent_rest_transport]
 port=%(port)s
-host=%(serveradress)s
+host=localhost
```
wouterdb, line 1286: This will only work when listening at least on the loopback ip?

i.e. will this work if you listen explicitly on an interface IP and not 0.0.0.0 ?

Issue #8089, opened 2024-09-13
Make partial export check that the source hashes it has still match the code of its full compile


wouterdb:
When the full compile that updates the compiler venv fails , this can leads to very hard to diagnose errors.

1. partial compiles don't push code
2. the full compile that performed the update on the compiler side failed
3. the subsequent partial compile used updated code on the compiler side, but old code on the agent side due to 1+2

sanderr:
Discussed in planning meeting. This one is harder than it seems, mostly because code is tied to a model version, and the model version is still unknown during partial export, making it difficult to check whether the code that will be used is up to date.

PR #8816, opened 2025-03-06
Issue/8729 improve agent code transport p2 the api

Hugo-Inmanta:
# Description

In this PR:

Removed:
  - The Code table
  - The tracking of resource types per agent on the executor

Added

  - InmantaModule table  ->  tracks versions of inmanta modules (e.g. std module)
  - FilesInModule table  ->  tracks which files belong to which (module_name, module_version)
  - ModulesForAgent table  ->  tracks which modules (module_name, module_version) should be installed on which agent for which model version

API

- [x] api to request all module/versions for a agent on a specific version: https://github.com/inmanta/inmanta-core/blob/84294626e294ec96082cd6dd2d220e85b46aff9d/src/inmanta/agent/code_manager.py#L65
- [x] api to download all files for a module/version: included in the one above
- [x] update codemanager of the scheduler to use them
- [x] api to upload code/version (like current code upload): https://github.com/inmanta/inmanta-core/blob/84294626e294ec96082cd6dd2d220e85b46aff9d/src/inmanta/loader.py#L95
- [x] add module/version argument to put_version to populate the join table and to put_partial to check source version remain the same: happens [here](https://github.com/inmanta/inmanta-core/blob/84294626e294ec96082cd6dd2d220e85b46aff9d/src/inmanta/server/services/orchestrationservice.py#L850-L860)

Points of attention
~- What to do with test_agent_paused_scheduler_server_restart in tests/deploy/e2e/test_autostarted.py~
~- Typing of args on the API border  src/inmanta/protocol/methods_v2.py~
~- Typing in the migration file src/inmanta/db/versions/v202503030.py~

closes https://github.com/inmanta/inmanta-core/issues/8729
closes https://github.com/inmanta/inmanta-core/issues/8089
closes https://github.com/inmanta/inmanta-core/issues/8952

# Self Check:

- [ ] Changelog entry


src/inmanta/agent/code_manager.py
```diff
@@ -63,71 +62,66 @@ async def get_pip_config(self, environment: uuid.UUID, version: int) -> PipConfi
...
+                models.ModulesForAgent,
+                models.Module,
+                and_(
+                    models.ModulesForAgent.module_name == models.Module.name,
+                    models.ModulesForAgent.module_version == models.Module.version,
+                    models.ModulesForAgent.environment == models.Module.environment,
+                ),
+            )
+            .join_from(
+                models.Module,
```
wouterdb, line 95: This could equally well join on ModulesForAgent if I understand correctly? (not that it matters)

src/inmanta/agent/code_manager.py
```diff
@@ -63,71 +62,66 @@ async def get_pip_config(self, environment: uuid.UUID, version: int) -> PipConfi
...
+        async with get_session() as session:
+            result_execute = await session.execute(stmt)
+            for res in result_execute.all():
+                module_install_specs.append(
+                    ModuleInstallSpec(
+                        module_name=res.module_name,
+                        module_version=res.module_version,
+                        model_version=res.cm_version,
+                        blueprint=executor.ExecutorBlueprint(
+                            pip_config=await self.get_pip_config(environment, model_version),
```
wouterdb, line 118: The pip config could probably be added to the query and perhaps move to one of the new tables

src/inmanta/agent/forking_executor.py
```diff
@@ -1062,29 +1062,13 @@ async def get_executor(
...
-            # reverse index
-            type_for_spec: dict[inmanta.loader.ModuleSource, list[ResourceType]] = collections.defaultdict(list)
-            for spec in code:
-                for source in spec.blueprint.sources:
-                    type_for_spec[source].append(spec.resource_type)
-            # resolve
-            for failed_resource_result in my_executor.failed_resource_results:
-                for rtype in type_for_spec.get(failed_resource_result.module_source, []):
-                    if rtype not in my_executor.failed_resources:
-                        my_executor.failed_resources[rtype] = failed_resource_result.exception
```
wouterdb, line 1086: so nice!

src/inmanta/loader.py
```diff
@@ -186,6 +155,44 @@ def get_file_hashes(self) -> Iterable[str]:
         """Return the hashes of all source files"""
         return (info.hash for info in self.__file_info.values())
 
+    def get_module_source_info(self) -> dict[str, list["SourceInfo"]]:
+        """Return all module source info"""
+        return self.__module_to_source_info
+
+    def get_modules_data(self) -> dict[str, dict[str, Any]]:
```
wouterdb, line 162: Why not 
```suggestion
    def get_modules_data(self) -> dict[str, PythonModule]:
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -665,6 +666,8 @@ async def _put_version(
         pip_config: Optional[PipConfig] = None,
         *,
         connection: asyncpg.connection.Connection,
+        module_version_info: Optional[dict[str, str]],
+        type_to_module_data: Optional[dict[ResourceIdStr, str]],
```
wouterdb, line 670: This one I don't understand

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -814,9 +818,74 @@ async def _put_version(
             all_agents: set[str] = {res.agent for res in rid_to_resource.values()}
             all_agents.add(const.AGENT_SCHEDULER_ID)
 
+            type_to_agent: dict[str, str] = {res.resource_type: res.agent for res in rid_to_resource.values()}
+            LOGGER.debug(module_version_info)
+
             for agent in all_agents:
                 await self.agentmanager_service.ensure_agent_registered(env, agent, connection=connection)
 
+            async def populate_join_table(
```
wouterdb, line 827:
```suggestion
            async def register_code(
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -814,9 +818,74 @@ async def _put_version(
...
+                                if not arg
+                            ]
+                        )
+                    )
+                    return
+
+                assert module_version_info
+                assert type_to_module_data
+                assert rid_to_resource
+                assert type_to_agent
```
wouterdb, line 854: Why is this not handled with stricter typing?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -814,9 +818,74 @@ async def _put_version(
...
+                            ) VALUES(
+                                $1,
+                                $2,
+                                $3,
+                                $4,
+                                $5
+                            );
+                        """
+                async with connection.transaction():
+                    await connection.fetchrow(
```
wouterdb, line 879: I would use insert many here

tests/db/migration_tests/test_v202501140_to_v202503030.py
```diff
@@ -0,0 +1,44 @@
...
+part = file_name_regex.match(__name__)[1]
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), f"dumps/v{part}.sql"))
+async def test_add_tables_for_agent_code_transport_rework(
+    postgresql_client: asyncpg.Connection,
+    migrate_db_from: abc.Callable[[], abc.Awaitable[None]],
+    get_tables_in_db: abc.Callable[[], abc.Awaitable[list[str]]],
+) -> None:
+    # This migration script only creates tables and adds indexes
```
wouterdb, line 36: I should definitely migrate data as well, otherwise the code will be gone

wouterdb (commented):
Looks good

src/inmanta/server/services/orchestrationservice.py
Hugo-Inmanta: Oh, I think I was confusing resourceTypes and resourceIds.
Here I'm linking resource ids to modules, but this should be linking resource types to modules, do you agree ?

wouterdb (commented):
I responded to the comments

Hugo-Inmanta:
Overview of remaining test failures:


## Fast tests


### Related to put_partial code version check addition

- tests.test_server.test_put_stale_version[False-True]
- tests.deploy.e2e.test_resource_sets.test_put_partial_dep_on_shared_set_removed
- tests.deploy.e2e.test_resource_sets.test_put_partial_merge_not_in_resource_set
- tests.deploy.e2e.test_resource_sets.test_put_partial_dep_on_specific_set_removed
- tests.deploy.e2e.test_resource_sets.test_put_partial_update_multiple_resource_set
- tests.deploy.e2e.test_resource_sets.test_put_partial_with_unknowns
- tests.deploy.e2e.test_resource_sets.test_put_partial_mixed_scenario
- tests.deploy.e2e.test_resource_sets.test_put_partial_version_allocation
- tests.deploy.e2e.test_resource_sets.test_put_partial_with_undeployable_resources
- tests.deploy.e2e.test_resource_sets.test_is_suitable_for_partial_compiles
- tests.deploy.e2e.test_resource_sets.test_put_partial_different_env
- tests.deploy.e2e.test_resource_sets.test_put_partial_with_resource_state_set


### Related to code loading overhaul

- tests.deploy.e2e.test_code_loading.test_agent_code_loading_with_failure


## Slow tests

- ...

src/inmanta/loader.py
```diff
@@ -662,3 +666,48 @@ def module_in_prefix(module: types.ModuleType, prefix: str) -> bool:
...
+
+
+@dataclass(frozen=True)
+class PythonModule:
+    name: str
+    version: str
+    files_in_module: list["SourceInfo"]
+
+
+class SourceInfo(BaseModel):
```
Hugo-Inmanta, line 678: I don't recall why I turned this into a pydantic model, this might not be necessary.

src/inmanta/protocol/methods_v2.py
Hugo-Inmanta, line 1: I struggled quite a bit with the typing of the endpoints, it should be improved.

src/inmanta/server/services/orchestrationservice.py
Hugo-Inmanta: Do you mean with non-optional types ? This is something I have been wondering about for the two new arguments to `put_version` i.e. `module_version_info` and `type_to_module_data` should we set those to non optional as well and update all the tests accordingly ?

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,211 @@
...
+        cm_version,
+        environment,
+        agent_name,
+        module_name,
+        module_version
+    )
+    VALUES ($1,$2,$3,$4,$5)
+    ON CONFLICT DO NOTHING
+    """
+    await connection.executemany(insert_modules_for_agent, modules_for_agent_data)
```
wouterdb, line 211: I think we also need to drop the old tables

src/inmanta/server/services/orchestrationservice.py
wouterdb: Or simply not call the method if the information is not present, that at least keeps this method clean

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -814,9 +822,116 @@ async def _put_version(
...
+                cm_version: int,
+                environment: uuid.UUID,
+                module_version_info: dict[str, "PythonModule"] | None,
+                type_to_module_data: dict[str, set[str]] | None,
+                type_to_agent: dict[str, str] | None,
+                connection: Connection,
+            ) -> None:
+                if not all([module_version_info, type_to_module_data, rid_to_resource, type_to_agent]):
+                    LOGGER.debug(
+                        "Cannot populate join table modules_for_agent. The following arguments are not set: %s"
```
wouterdb, line 841: the join table means nothing to the end-user, I would formulate this in terms of the external api.

wouterdb (commented):
I did a quick pass, looks good

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,238 @@
...
+        CREATE INDEX modules_for_agent_environment_agent_name_index
+        ON public.modules_for_agent (environment, agent_name);
+        CREATE INDEX modules_for_agent_environment_module_name_module_version_index
+        ON public.modules_for_agent (environment, module_name, module_version);
+
+    """
+
+    await connection.execute(schema)
+
+    code_data: dict[str, dict[int, dict[str, set[tuple[str, str, str, set[str]]]]]] = defaultdict(
```
Hugo-Inmanta, line 86: Typing is very unreadable. Maybe adding dataclasses would make it easier to read ?

src/inmanta/protocol/methods_v2.py
Hugo-Inmanta, line 1: Argument typing at the API level is a mess.

Hugo-Inmanta:
@wouterdb 
I could use your input on the following pain points:

1. What to do with `test_agent_paused_scheduler_server_restart` in `tests/deploy/e2e/test_autostarted.py`
2. Typing of args on the API border `src/inmanta/protocol/methods_v2.py`
3. Typing in the migration file `src/inmanta/db/versions/v202503030.py`

src/inmanta/agent/code_manager.py
```diff
@@ -63,71 +63,92 @@ async def get_pip_config(self, environment: uuid.UUID, version: int) -> PipConfi
...
                 )
-                invalid_resources[resource_type] = e
-            except Exception as e:
-                LOGGER.error(
-                    "Failed to get source code for %s version=%d",
-                    resource_type,
-                    version,
-                    exc_info=True,
+
+                files = await session.execute(files_in_module)
```
wouterdb, line 128: It would be better to avoid the for loop and do one query to get all the files

src/inmanta/data/__init__.py
```diff
@@ -2694,7 +2693,11 @@ async def clear(self, connection: Optional[asyncpg.connection.Connection] = None
             await Compile.delete_all(environment=self.id, connection=con)  # Triggers cascading delete on report table
             await Parameter.delete_all(environment=self.id, connection=con)
             await Notification.delete_all(environment=self.id, connection=con)
-            await Code.delete_all(environment=self.id, connection=con)
+
+            await self._execute_query("DELETE FROM public.module WHERE environment=$1", self.id, connection=con)
+            await self._execute_query("DELETE FROM public.files_in_module WHERE environment=$1", self.id, connection=con)
+            await self._execute_query("DELETE FROM public.modules_for_agent WHERE environment=$1", self.id, connection=con)
```
wouterdb, line 2699: In the past, we would place these in methods on the relevant type, this is now in-line.
Could you motivate this change?

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,238 @@
...
+            environment uuid NOT NULL,
+            file_content_hash varchar NOT NULL REFERENCES file(content_hash) ON DELETE CASCADE,
+            file_path varchar NOT NULL,
+            PRIMARY KEY(environment, module_name, module_version, file_path),
+            FOREIGN KEY (environment, module_name, module_version)
+                REFERENCES public.module(environment, name, version) ON DELETE CASCADE
+        );
+
+        CREATE INDEX files_in_module_file_content_hash_index
+        ON public.files_in_module (file_content_hash);
```
wouterdb, line 60: do we ever use it in this direction?

src/inmanta/db/versions/v202503030.py
wouterdb: I would add least document what it is. Named tuples or dataclasses can help, but I would not invest too much effort

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,238 @@
...
+            Helper method
+            """
+            for file_hash, file_path, _, _ in source_info:
+                parts = Path(file_path).parts
+                if f"{module_name}/plugins" in file_path:
+                    # V1 module
+                    rel_py_file = os.path.relpath(file_path, start="plugins")
+                    relative_path = os.path.join(module_name, loader.PLUGIN_DIR, rel_py_file)
+                else:
+                    relative_path = str(Path(*parts[parts.index(const.PLUGINS_PACKAGE) :]))
```
wouterdb, line 160: I wonder if we need this and if the correct, normalized form is already available in one of the other arguments

src/inmanta/data/sqlalchemy.py
```diff
@@ -45,6 +45,66 @@ class Base(DeclarativeBase):
...
+        ),  # todo add test for restrict behaviour on delete
+        UniqueConstraint("module_name", "module_version", "environment", "file_path"),
+    )
+    __mapper_args__ = {"primary_key": ["module_name", "module_version", "environment", "file_path"]}
+
+    module_name: Mapped[str] = mapped_column(String)
+    module_version: Mapped[str] = mapped_column(String)
+    environment: Mapped[uuid.UUID] = mapped_column(UUID)
+    file_content_hash: Mapped[str] = mapped_column(String)
+    file_path: Mapped[str] = mapped_column(String)
```
wouterdb, line 82: 1. is this the path where it comes from or where it should go
2. path relative to what
3. Do we need/want this? Would it not be more convenient to use just module name and if it a py or a pyc file?

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,238 @@
...
+        ) -> tuple[str, list[str]]:
+            """
+            Helper method
+            """
+            reqs = set()
+            module_version_hash = hashlib.new("sha1")
+
+            for file_hash, _, _, requirements in sorted(source_info, key=lambda x: x[0]):
+                # sort by hash to compute inmanta module version
+                module_version_hash.update(file_hash.encode())
```
wouterdb, line 137: same files, different requirements will produce the same hash, that is a problem I think

src/inmanta/db/versions/v202503030.py
wouterdb: The overall approach can also use some comments, as this is very hard to follow

src/inmanta/export.py
```diff
@@ -89,27 +89,10 @@ def upload_code(conn: protocol.SyncClient, tid: uuid.UUID, version: int, code_ma
...
-    # ...other types would be included as well
-    # }
-    source_map = {
-        resource_name: {source.hash: (source.path, source.module_name, source.requires) for source in sources}
-        for resource_name, sources in code_manager.get_types()
-    }
-
-    res = conn.upload_code_batched(tid=tid, id=version, resources=source_map)
+    modules_data = code_manager.get_modules_data()
+    res = conn.upload_modules(tid=tid, modules_data=modules_data)
```
wouterdb, line 93: How can this work outside of the put_version and put_partial with all the foreign keys?

src/inmanta/loader.py
```diff
@@ -128,12 +82,24 @@ class CodeManager:
     def __init__(self) -> None:
         # Old implementation
         # Use by external code
+
+        # Map of [Union[resouce_type, handler_type], set[path]]
```
wouterdb, line 86: this doesn't correspond to the type declared 2 lines lower

src/inmanta/loader.py
```diff
@@ -186,6 +154,42 @@ def get_file_hashes(self) -> Iterable[str]:
         """Return the hashes of all source files"""
         return (info.hash for info in self.__file_info.values())
 
+    def get_module_source_info(self) -> dict[str, list["SourceInfo"]]:
```
wouterdb, line 157: note to self: continue here

src/inmanta/protocol/methods_v2.py
```diff
@@ -1365,20 +1368,20 @@ def update_notification(
...
+    path="/modulecodebatched/",
+    operation="PUT",
     arg_options=methods.ENV_OPTS,
-    client_types=[ClientType.agent],
+    client_types=[ClientType.compiler],
     api_version=2,
+    varkw=True,
 )
-def get_source_code(tid: uuid.UUID, version: int, resource_type: str) -> list[model.Source]:
+def upload_modules(tid: uuid.UUID, **kwargs: object) -> ReturnValue[int]:
```
wouterdb, line 1378: you need pydantic objects/dataclasses for all the stuff going in

but, more important, why does this method exist at all? 
i.e. unlike the files, we always push all this data (we don't first check which modules don't exists)
so if we make this part of the put_version directly the traffic pattern will be the same, except that we make one less call.  What is the rationale behind this choice?

wouterdb (changes requested):
I didn't make it to the end because there are some fundamental things I don't understand, comments added

src/inmanta/data/sqlalchemy.py
Hugo-Inmanta: - Updated `file_path` -> `python_module_name`
- Added `is_byte_code`

src/inmanta/db/versions/v202503030.py
Hugo-Inmanta: Indeed, updated this to use the `python_module_name` argument. I had some misconceptions about how it behaved for v1 vs v2 modules. It will actually always be of the form `inmanta_plugins.<inmanta_module_name>.<python_module_name>` regardless of the module's version.

src/inmanta/db/versions/v202503030.py
Hugo-Inmanta: Updated the version computation to include the requirements.

src/inmanta/protocol/methods_v2.py
Hugo-Inmanta: I see, moved this to be part of `put_version`.

src/inmanta/data/__init__.py
Hugo-Inmanta: Factorized this in the sqlalchemy base class.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -912,19 +916,64 @@ async def check_version_info(
...
+                    await session.commit()
 
             if is_partial_update:
                 assert partial_base_version is not None
                 await check_version_info(
                     partial_base_version, env.id, module_version_info, type_to_module_data, type_to_agent, connection
                 )
+            else:
+                # TODO this should use the same connection
+                await upload_modules(env.id, module_version_info)
```
Hugo-Inmanta, line 976: TODO: this should use the same connection

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,288 @@
...
+        model_versions: defaultdict[int, SourcesPerModule] = field(default_factory=lambda: defaultdict(SourcesPerModule))
+
+    @dataclass
+    class VersionsPerEnv:
+        # Maps environments to model versions
+        environments: defaultdict[str, ModulesPerVersion] = field(default_factory=lambda: defaultdict(ModulesPerVersion))
+
+    async def fetch_code_data(
+        code_data: VersionsPerEnv,
+        resource_type_to_module: dict[int, dict[str, set[str]]],
```
wouterdb, line 109: if these are output, I would make them return values, this is very hard to read.

the applies to all code in this file. I can't make out how it works, because input and output are not clearly separated.

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,288 @@
...
+    class ModulesPerVersion:
+        # Maps model versions to their inmanta modules
+        model_versions: defaultdict[int, SourcesPerModule] = field(default_factory=lambda: defaultdict(SourcesPerModule))
+
+    @dataclass
+    class VersionsPerEnv:
+        # Maps environments to model versions
+        environments: defaultdict[str, ModulesPerVersion] = field(default_factory=lambda: defaultdict(ModulesPerVersion))
+
+    async def fetch_code_data(
```
wouterdb, line 107: I would like to have some documentation here about the conversion, preferably in the form of 

1. the old format is like this, 
2. the new format is like this. 
3. a list of sub conversions 
    - Using these inputs, we build these outputs in methods xyx
4. short explanation of the actual sequence (we perform the sub conversion x and y in memory together because they share the same data, and write them back at point x because ... )

src/inmanta/deploy/tasks.py
```diff
@@ -76,45 +75,25 @@ async def get_executor(
...
-            version=version,
-            resource_types=all_types_for_agent,
+        code = await task_manager.code_manager.get_code(
+            environment=task_manager.environment, model_version=version, agent_name=agent_name
         )
 
-        # Bail out if this failed
-        if resource_type in invalid_resources:
-            raise invalid_resources[resource_type]
-
```
wouterdb, line 108: Where did this go?

src/inmanta/export.py
```diff
@@ -78,7 +78,7 @@ def __str__(self) -> str:
         return "Cycle in dependencies: %s" % self.cycle
 
 
-def upload_code(conn: protocol.SyncClient, tid: uuid.UUID, version: int, code_manager: loader.CodeManager) -> None:
+def upload_code(conn: protocol.SyncClient, code_manager: loader.CodeManager) -> None:
```
wouterdb, line 81: This will break the devtools mock, we have to see what the impact will be an how we can make it least painful

src/inmanta/loader.py
```diff
@@ -186,6 +154,44 @@ def get_file_hashes(self) -> Iterable[str]:
...
+
+    def get_module_version_info(self) -> dict[str, "PythonModule"]:
+        """Return all module version info"""
+        if self.__modules_data:
+            return self.__modules_data
+
+        source_info = self.get_module_source_info()
+
+        modules_data = {}
+        for module_name, files_in_module in source_info.items():
```
wouterdb, line 169: Why not use this code in the database migration? i.e. the has calculation is now in two places (it will works, just not very efficiently)

src/inmanta/loader.py
```diff
@@ -375,7 +381,7 @@ def touch_inits(directory: str) -> None:
                 "Not deploying code (hv=%s, module=%s) because of cache hit", module_source.hash_value, module_source.name
             )
 
-    def deploy_version(self, module_sources: Iterable[ModuleSource]) -> None:
+    def deploy_version(self, module_sources: Iterable[ModuleSource], module_name: str) -> None:
```
wouterdb, line 384: Why the extra argument

src/inmanta/loader.py
```diff
@@ -428,6 +434,57 @@ def _loading_top_level_package(self) -> bool:
...
+    """
+    if path == "":
+        return iter(())
+    init, last = os.path.split(path)
+    yield from split(init)
+    if last != "":
+        yield last
+
+
+def convert_module_path_to_namespace(path: str) -> str:
```
wouterdb, line 465: How is this distinct from the method below?

src/inmanta/loader.py
```diff
@@ -662,3 +690,50 @@ def module_in_prefix(module: types.ModuleType, prefix: str) -> bool:
...
+
+@dataclass(frozen=True)
+class PythonModule:
+    name: str
+    version: str
+    files_in_module: list["SourceInfo"]
+
+
+class SourceInfo(BaseModel):
+    """This class is used to store information related to source code information"""
```
wouterdb, line 703: Could you document how this is distinct from ModuleSource and PythonModule?

src/inmanta/loader.py
```diff
@@ -662,3 +690,50 @@ def module_in_prefix(module: types.ModuleType, prefix: str) -> bool:
...
+
+    @computed_field  # type: ignore[prop-decorator]
+    @cached_property
+    def hash(self) -> str:
+        """Get the sha1 hash of the file"""
+        sha1sum = hashlib.new("sha1")
+        sha1sum.update(self.content)
+        return sha1sum.hexdigest()
+
+    @cached_property
```
wouterdb, line 718: 1. I think it is important to verify this doesn't leak memory like the `@lrucache`
2. this lazy loading it very tricky (i.e. it works if you are on the same machine, which for testing is always true, but not in general where the compiler can be remote). Can you add a specific test to make sure the file is transported correctly (i.e. not via disk)

src/inmanta/protocol/methods.py
```diff
@@ -531,6 +531,8 @@ def put_version(
     compiler_version: Optional[str] = None,
     resource_sets: dict[inmanta.types.ResourceIdStr, Optional[str]] = {},
     pip_config: Optional[PipConfig] = None,
+    module_version_info: Optional[dict[str, Any]] = None,
+    type_to_module_data: Optional[dict[str, set[str]]] = None,
```
wouterdb, line 535: this needs clean types  and docs

src/inmanta/protocol/methods_v2.py
```diff
@@ -48,6 +49,8 @@ def put_partial(
     resource_sets: Optional[dict[ResourceIdStr, Optional[str]]] = None,
     removed_resource_sets: Optional[list[str]] = None,
     pip_config: Optional[PipConfig] = None,
+    module_version_info: Optional[dict[str, Any]] = None,
+    type_to_module_data: Optional[dict[str, Any]] = None,
```
wouterdb, line 53: this needs clean types and docs

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -814,9 +820,163 @@ async def _put_version(
             all_agents: set[str] = {res.agent for res in rid_to_resource.values()}
             all_agents.add(const.AGENT_SCHEDULER_ID)
 
+            type_to_agent: dict[str, set[str]] = defaultdict(set)
+            for res in rid_to_resource.values():
+                type_to_agent[str(res.resource_type)].add(res.agent)
```
wouterdb, line 825: I wonder if we should have the exporter produce this? i.e. there is no reason to talk about types here, it is all about modules?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -814,9 +820,163 @@ async def _put_version(
...
+                                        agent_name,
+                                        module_name,
+                                        module_version_info[module_name]["version"],
+                                    )
+                                )
+
+                    await connection.executemany(
+                        query,
+                        values,
+                    )
```
wouterdb, line 876: this should go into the data package I think (i.e. in the relevant objects)

I think this applies to all queries

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -814,9 +820,163 @@ async def _put_version(
...
+                        file_path = file["path"]
+                        match file_path:
+                            case _ if file_path.endswith(".py"):
+                                is_byte_code = False
+                            case _ if file_path.endswith(".pyc"):
+                                is_byte_code = True
+                            case _:
+                                raise ServerError(
+                                    "Invalid file extension for plugin file %s. Expecting `.py` or `.pyc`." % file_path
+                                )
```
wouterdb, line 942: Why isn't this conversion in the exporter?

And why doesn't it use the existing method  to determine what it is?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -814,9 +820,163 @@ async def _put_version(
...
+                        "environment": env,
+                        "requirements": requirements,
+                    }
+
+                    module_data.append(module)
+
+                async with get_session() as session:
+                    await session.execute(module_stmt, module_data)
+                    await session.execute(files_in_module_stmt, files_in_module_data)
+                    await session.commit()
```
wouterdb, line 967: it is required to commit here explicitly?

tests/db/migration_tests/test_v202501140_to_v202503030.py
```diff
@@ -0,0 +1,82 @@
...
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), f"dumps/v{part}.sql"))
+async def test_add_tables_for_agent_code_transport_rework(
+    migrate_db_from: abc.Callable[[], abc.Awaitable[None]],
+    get_tables_in_db: abc.Callable[[], abc.Awaitable[list[str]]],
+) -> None:
+
+    assert "inmanta_module" not in await get_tables_in_db()
+    assert "files_in_module" not in await get_tables_in_db()
+    assert "modules_for_agent" not in await get_tables_in_db()
+    assert "code" in await get_tables_in_db()
```
wouterdb, line 42: this is not required, it doesn't really test anything we care about

tests/db/migration_tests/test_v202501140_to_v202503030.py
```diff
@@ -0,0 +1,82 @@
...
+                models.ModulesForAgent.inmanta_module_version,
+            )
+            .order_by(models.ModulesForAgent.agent_name)
+            .where(models.ModulesForAgent.cm_version == 1)
+        )
+        modules_for_agent = await session.execute(modules_for_agent_stmt)
+        assert modules_for_agent.all() == [
+            ("internal", "std", "d95a4a8894881c79b1c791fb94824db2dd961d08"),
+            ("localhost", "fs", "a8ecaac2c9448803a18a5d9e16bbd87f133a06fc"),
+        ]
```
wouterdb, line 82: these tests should use the actual api for this as per the database update ADR

tests/deploy/e2e/test_autostarted.py
```diff
@@ -28,6 +28,7 @@
 import uuid
```
wouterdb, line 28: I got till here

src/inmanta/loader.py
Hugo-Inmanta: removed

src/inmanta/db/versions/v202503030.py
Hugo-Inmanta: No, I don't think so. Removed it.

src/inmanta/loader.py
Hugo-Inmanta: removed

src/inmanta/server/services/orchestrationservice.py
Hugo-Inmanta: refactored into the exporter

src/inmanta/db/versions/v202503030.py
```diff
@@ -0,0 +1,292 @@
...
+        );
+
+        CREATE INDEX modules_for_agent_environment_agent_name_index
+        ON public.modules_for_agent (environment, agent_name);
+        CREATE INDEX modules_for_agent_environment_module_name_module_version_index
+        ON public.modules_for_agent (environment, inmanta_module_name, inmanta_module_version);
+
+    """
+
+    await connection.execute(schema)
```
sanderr, line 87: Note to self: review migration from here on out

src/inmanta/loader.py
```diff
@@ -28,21 +26,20 @@
...
 from dataclasses import dataclass
 from importlib.abc import FileLoader, MetaPathFinder
 from importlib.machinery import ModuleSpec, SourcelessFileLoader
 from itertools import chain
-from typing import TYPE_CHECKING, Optional
+from typing import Optional
 
 from inmanta import const, module
+from inmanta.data.model import InmantaModuleDTO, ModuleSource, ModuleSourceMetadata
+from inmanta.resources import Id, Resource
```
sanderr, line 39: This makes this module less stand-alone, i.e. more prone to import loops. Might be fine, I'll need to have another look at this.

src/inmanta/loader.py
```diff
@@ -65,74 +62,35 @@ class SourceNotFoundException(Exception):
...
+    def build_agent_map(self, resources: dict[Id, Resource]) -> None:
+        """
+        Construct a map of which agents are registered to deploy which resource type.
+        This map is later used to construct a map of which agents need to load
+        which Inmanta module(s).
+        """
+        for id in resources:
+            self._types_to_agent[id.entity_type].add(id.agent_name)
 
     def register_code(self, type_name: str, instance: object) -> None:
```
sanderr, line 95: Requires `build_agent_map` to be called first. Should be in the docstring

src/inmanta/loader.py
```diff
@@ -184,69 +133,57 @@ def get_object_source(self, instance: object) -> Optional[str]:
...
-        if not isinstance(other, ModuleSource):
-            return NotImplemented
-        return (self.name, self.hash_value, self.is_byte_code) < (other.name, other.hash_value, other.is_byte_code)
+            module_version = self.get_module_version(requirements, files_metadata)
 
-    def __eq__(self, other: object) -> bool:
-        if not isinstance(other, ModuleSource):
-            return False
-        return (self.name, self.hash_value, self.is_byte_code) == (other.name, other.hash_value, other.is_byte_code)
+            modules_data[module_name] = InmantaModuleDTO(
```
sanderr, line 159: The logic that leads here looks good, but I wonder if perhaps what was there before lingered a bit too much? You now have all these intermediate fields (the `__` ones), that used to be built incrementally (iirc), but are now really built on a per-module basis already, so why not simply keep this here in an attribute instead? i.e. construct it in `register_inmanta_module`, store it there, get rid of `__module_to_source_info` and `__module_to_agents` and a bunch of methods. And perhaps even `__file_info`. I think it would all become quite a bit more cohesive.

src/inmanta/loader.py
```diff
@@ -500,9 +443,6 @@ def convert_module_to_relative_path(full_mod_name: str) -> str:
 
     module_parts.insert(1, PLUGIN_DIR)
 
-    if module_parts[-1] == "__init__":
-        module_parts = module_parts[:-1]
-
```
sanderr, line 505: What's this about?

src/inmanta/agent/forking_executor.py
```diff
@@ -1049,42 +1047,26 @@ async def get_executor(
...
-            raise ValueError(f"{self.__class__.__name__}.get_executor() expects at least one resource install specification")
 
         blueprint = executor.ExecutorBlueprint.from_specs(code)
         executor_id = executor.ExecutorId(agent_name, agent_uri, blueprint)
 
         # Use pool manager
         my_executor = await self.get(executor_id)
 
-        # translation from code to type can only be done here
-        if my_executor.failed_resource_results and not my_executor.failed_resources:
```
sanderr, line 1075: The purpose of these two attributes is quite vague. Not your fault, but that makes this change difficult to review. Tell me if I've got this right:
- `failed_resource_results` is set by the process init if a module fails to install or fails to load
- `failed_resources` was set exclusively here for the forking executor. The in-process executor would set it similar to how `failed_resource_results` is set here, but we don't ever reach that code path anymore, except perhaps during testing.
- `failed_resources` is the only one of the two that's exposed to the outside world

I'm inclined to conclude that
1. Not necessarily in scope of this PR, but it could definitely help to get confidence in the changes if this were cleaned up. Easy wins could be some comments and to prefix `failed_resource_results` with an underscore.
2. We still need to populate the exposed field somehow, and we need to do this on a per resource type basis, like we did before. Don't we? Only we don't have that information easily available anymore, do we?

sanderr (changes requested):
Haven't reviewed sqlalchemy, and part of the migration script, or the tests. But I'll let the first batches settle before I come back.

changelogs/unreleased/8729-improve-agent-code-transport.yml
```diff
@@ -0,0 +1,4 @@
+description: Rework the code transport between the compiler and the agent.
+issue-nr: 8729
+change-type: minor
+destination-branches: [master]
```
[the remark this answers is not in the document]
Hugo-Inmanta: Discussed in planning meeting -> iso9 only

src/inmanta/loader.py
Hugo-Inmanta: Nice catch, I think this is an artifact from an earlier implementation.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -817,6 +822,43 @@ async def _put_version(
...
+                """
+
+                base_version_data: dict[tuple[str, str], list[str]] = await ModulesForAgent.get_agents_per_module(
+                    model_version=partial_base_version, environment=environment, connection=connection
+                )
+                for (inmanta_module_name, inmanta_module_version), _ in base_version_data.items():
+                    if (
+                        inmanta_module_name not in module_version_info
+                        or module_version_info[inmanta_module_name].version != inmanta_module_version
+                    ):
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think you're right

src/inmanta/agent/code_manager.py
```diff
@@ -63,71 +62,132 @@ async def get_pip_config(self, environment: uuid.UUID, version: int) -> PipConfi
...
+
+        previous_inmanta_module: InmantaModuleDTO = InmantaModuleDTO(
+            name="",
+            version="",
+            files_in_module=[],
+            requirements=[],
+            required_by=[],
+        )
+        previous_module_sources: list[ModuleSource] = []
+        async with get_session() as session:
```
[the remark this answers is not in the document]
Hugo-Inmanta: Is there any other reason than readability ? (I agree this is a strict improvement)

src/inmanta/agent/code_manager.py
```diff
@@ -63,71 +62,132 @@ async def get_pip_config(self, environment: uuid.UUID, version: int) -> PipConfi
...
-            except Exception as e:
-                LOGGER.error(
-                    "Failed to get source code for %s version=%d",
-                    resource_type,
-                    version,
-                    exc_info=True,
+
+        module_install_specs = []
+        modules_for_agent = (
+            select(
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think there are some opportunities here to better leverage SQL alchemy's strengths later (e.g. I think the joins don't have to be explicitly written out, SQL alchemy can deduce them from the select clause and the relationships defined on the model).
I went with this approach to try and stay close to what we know and that the purpose of the query is as explicit as possible.

src/inmanta/agent/code_manager.py
```diff
@@ -63,71 +62,132 @@ async def get_pip_config(self, environment: uuid.UUID, version: int) -> PipConfi
...
+            requirements=[],
+            required_by=[],
+        )
+        previous_module_sources: list[ModuleSource] = []
+        async with get_session() as session:
+            result = await session.execute(modules_for_agent)
+            for res in result.all():
+                current_row_module_name = res.inmanta_module_name
+
+                if current_row_module_name != previous_inmanta_module.name:
```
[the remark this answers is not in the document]
Hugo-Inmanta: Nice, that's exactly what I needed, didn't know that it existed in `itertools`.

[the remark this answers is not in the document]
sanderr: Cautionary note: it requires the iterable to be pre-sorted by key. That is already the case here (iirc), which makes it safe. But it's an easy mistake to make in general.

src/inmanta/agent/code_manager.py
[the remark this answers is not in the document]
sanderr: Just that. Provide some context to what this ambiguously named thing is about.

src/inmanta/agent/code_manager.py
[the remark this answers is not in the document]
sanderr: > I think the on clauses of the joins don't have to be explicitly written out, SQL alchemy can deduce them from the select clause and the relationships defined on the model

I assumed it would fall back to the `NATURAL JOIN`. That would make the most sense to me. For many of our cases, I think it makes sense to write them out explicitly.

I was mostly referring to the verbosity spreading everything out. But like I said, it may simply be a matter of getting used to it. I can't point out anything intrinsically worse about this than a manually-written query, and there are actually a few benefits, so I think I just need to let it sink in.

changelogs/unreleased/8729-improve-agent-code-transport.yml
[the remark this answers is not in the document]
sanderr: Leaving this open for now because I want to double check that there's nothing in here that we really want ASAP.

src/inmanta/agent/forking_executor.py
```diff
@@ -1049,42 +1047,26 @@ async def get_executor(
...
         :param agent_name: The name of the agent for which an Executor is being retrieved or created.
         :param agent_uri: The name of the host on which the agent is running.
-        :param code: Collection of ResourceInstallSpec defining the configuration for the Executor i.e.
+        :param code: Collection of ModuleInstallSpec defining the configuration for the Executor i.e.
             which resource types it can act on and all necessary information to install the relevant
             handler code in its venv. Must have at least one element.
         :return: An Executor instance
         """
-        if not code:
-            raise ValueError(f"{self.__class__.__name__}.get_executor() expects at least one resource install specification")
```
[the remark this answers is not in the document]
sanderr: First off, you're right. So my statement that it remains "as important as before" isn't accurate.

That said, this is the sort of defensive style that I'm very much in favor of: verify your method invariants within the method, don't make assumptions on what's going on outside of our known domain. It costs us next to nothing to have the check here, it has caused us headaches when it wasn't there, and it may again (though unlikely because of the check you're referring to).

I vaguely recall @wouterdb writing a blog post about the benefits of a defensive style like this, but I can't find it.

src/inmanta/agent/forking_executor.py
Hugo-Inmanta: > We still need to populate the exposed field somehow, and we need to do this on a per resource type basis, like we did before. Don't we?

I think we can do this on a per-module (python module) basis.

If I'm not mistaken e.g. if before we had 3 resource types living in the same module, we would attempt to load the module 3 times, (1 time for each resource) and assuming the loading would fail, we would report the same error 3 times.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -817,6 +822,43 @@ async def _put_version(
...
+
+            if is_partial_update:
+                assert partial_base_version is not None
+                await check_version_info(partial_base_version, env.id, module_version_info, connection)
+            else:
+                await InmantaModule.register_modules(
+                    environment=env.id, module_version_info=module_version_info, connection=connection
+                )
+
+            await ModulesForAgent.register_modules_for_agents(
```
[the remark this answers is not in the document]
Hugo-Inmanta: Updated as follows:
- `_check_version_info` will return which agents were requiring which modules (name, version) in the base version (n-1)
- `register_modules_for_agents` will use this to populate the partial version (n)

src/inmanta/agent/executor.py
```diff
@@ -270,17 +272,19 @@ def identity(self) -> str:
...
     :ivar model_version: the version of the model to use
-    :ivar blueprint: the associate install blueprint
+    :ivar blueprint: the associated install blueprint
 
     """
 
-    resource_type: ResourceType
+    module_name: str
+    module_version: str
     model_version: int
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think not

src/inmanta/agent/code_manager.py
```diff
@@ -63,71 +62,132 @@ async def get_pip_config(self, environment: uuid.UUID, version: int) -> PipConfi
...
+            .join_from(
+                models.ModulesForAgent,
+                models.InmantaModule,
+                and_(
+                    models.ModulesForAgent.inmanta_module_name == models.InmantaModule.name,
+                    models.ModulesForAgent.inmanta_module_version == models.InmantaModule.version,
+                    models.ModulesForAgent.environment == models.InmantaModule.environment,
+                ),
+            )
+            .join_from(
```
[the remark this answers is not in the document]
Hugo-Inmanta: > Isn't this equivalent to `join(models.FilesInModule)`?

Yes, I think it is, switched to `join(...)`

src/inmanta/loader.py
Hugo-Inmanta: iirc the rationale at the time to keep doing it this way was that InmantaModuleDTO was frozen (i.e. had the attribute `model_config: ClassVar[ConfigDict] = ConfigDict(frozen=True)` so I couldn't create the object once and update it.
Unless I'm missing something, I can't find a good reason for it to be frozen anymore so I will update this part.

src/inmanta/data/model.py
```diff
@@ -967,3 +969,93 @@ def __add__(self, other: "DataBaseReport") -> "DataBaseReport":
...
+            return NotImplemented
+        return self.meta_data < other.meta_data
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, ModuleSource):
+            return False
+        return self.meta_data == other.meta_data
+
+
+class InmantaModuleDTO(BaseModel):
```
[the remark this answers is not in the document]
sanderr: In other places we typically disambiguate through qualified access (`data.` and `model.`). I haven't checked the implementation yet, so just an FYI at this point.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -817,6 +822,43 @@ async def _put_version(
...
+                base_version_data: dict[tuple[str, str], list[str]] = await ModulesForAgent.get_agents_per_module(
+                    model_version=partial_base_version, environment=environment, connection=connection
+                )
+                for (inmanta_module_name, inmanta_module_version), _ in base_version_data.items():
+                    if (
+                        inmanta_module_name not in module_version_info
+                        or module_version_info[inmanta_module_name].version != inmanta_module_version
+                    ):
+                        raise BadRequest(
+                            "Cannot perform partial export because of version mismatch for module %s." % inmanta_module_name
```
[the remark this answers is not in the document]
sanderr: I think we have to discuss this on Slack or in a meeting.

src/inmanta/agent/code_manager.py
Hugo-Inmanta: > The pip config could probably be added to the query

This part I agree

> and perhaps move to one of the new tables

What is the motivation behind this ? Wouldn't this bring unnecessary duplication i.e. store the same pip config on each row with a similar model version in the `modules_for_agent` table ?

src/inmanta/data/model.py
```diff
@@ -967,3 +969,93 @@ def __add__(self, other: "DataBaseReport") -> "DataBaseReport":
...
+    """
+    This class represents an Inmanta module during code upload.
+
+    :param name: Name of this inmanta module. e.g. std
+    :param version: Version of this inmanta module. This hash is computed using the hashes of
+        the python files in this module as well as the python requirements of this module.
+    :param files_in_module: The list of python files composing this inmanta module.
+    :param requirements: The list of python requirements this inmanta module requires.
+    :param required_by: The list of agent names that require to install this inmanta module to
```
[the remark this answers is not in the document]
Hugo-Inmanta: - Renamed `required_by` --> `for_agents`

> Apart from that, it stands out to me that this is a different format than the one in the database. Is this to keep the code upload signature uniform?

Yes,
- `name`, `version` and `requirements` are used as is to populate the corresponding `inmanta_module` table.
- `files_in_module` is used to populate the `module_files` table.
- `for_agents` is used to populate the `agent_modules` tables.

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
Hugo-Inmanta: Created ticket to not lose track of this: https://github.com/inmanta/inmanta-core/issues/8982

src/inmanta/data/__init__.py
```diff
@@ -4072,7 +4072,12 @@ def __init__(self, from_postgres: bool = False, **kwargs: object) -> None:
         if from_postgres and self.messages:
             new_messages = []
             for message in self.messages:
-                message = json.loads(message)
+                # Not 100% sure why this is needed, could depend on whether the data is
+                # retrieved through sql alchemy (automatically deserializes json into dict)
+                # or through regular asyncpg queries (no automatic deserialization)
+                if isinstance(message, str):
+                    message = json.loads(message)
+                assert isinstance(message, dict)
```
Hugo-Inmanta, line 4080: Still need to address this

Hugo-Inmanta:
@sanderr  @wouterdb  This is the last pain point: https://github.com/inmanta/inmanta-core/pull/8816#discussion_r2068562093

src/inmanta/data/__init__.py
wouterdb: SQl alchemy does some reconfiguration of asyncpg, but I don't get why it would be selective? I would expect it to always be either the one or the other?

wouterdb: and if you keep the other branch? (I would expect to always get a dict when using sqlalchemy?)

src/inmanta/deploy/tasks.py
```diff
@@ -95,6 +95,18 @@ async def get_executor(
             agent_name=agent_name, agent_uri="NO_URI", code=code
         )
 
+        # Bail out if the current task's resource needs code
+        # that failed to load/install:
+        inmanta_module_name = self.resource.split("::")[0]
```
sanderr, line 100: Pretty sure we have a helper for this

src/inmanta/deploy/tasks.py
```diff
@@ -95,6 +95,18 @@ async def get_executor(
...
+        if inmanta_module_name in my_executor.failed_modules:
+            raise BaseExceptionGroup(
```
sanderr, line 102: This does not look right. `Base...` are not intended to be raised afaik.

I'm also not sure if it makes sense to report on all modules here, or only the one that failed, as we did before. @wouterdb wdyt?

src/inmanta/deploy/tasks.py
```diff
@@ -95,6 +95,18 @@ async def get_executor(
...
+                """
+                    The following modules cannot be installed: %s.
```
sanderr, line 104:
```suggestion
                    The following modules cannot be loaded: %s.
```

src/inmanta/deploy/tasks.py
```diff
@@ -154,6 +166,19 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
                         agent_name=agent,
                         version=version,
                     )
+
+                except BaseExceptionGroup as beg:
```
sanderr, line 170: This entire block looks redundant to me

sanderr (changes requested):

src/inmanta/loader.py
```diff
@@ -28,14 +28,13 @@
...
-from dataclasses import dataclass
 from importlib.abc import FileLoader, MetaPathFinder
 from importlib.machinery import ModuleSpec, SourcelessFileLoader
 from itertools import chain
 from typing import Optional
 
 from inmanta import const, module
-from inmanta.data.model import InmantaModuleDTO, ModuleSource, ModuleSourceMetadata
+from inmanta.data.model import InmantaModule, ModuleSource, ModuleSourceMetadata
 from inmanta.resources import Id, Resource
```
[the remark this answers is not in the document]
Hugo-Inmanta: This only goes for `ModuleSourceMetadata`, the other two are used at run time:
- https://github.com/inmanta/inmanta-core/blob/c860b57a3a15b91e2f3d51ca10d423f7e68ffa55/src/inmanta/loader.py#L130
- https://github.com/inmanta/inmanta-core/blob/c860b57a3a15b91e2f3d51ca10d423f7e68ffa55/src/inmanta/loader.py#L122

src/inmanta/agent/in_process_executor.py
```diff
@@ -201,6 +202,13 @@ def resolve_and_deploy(
                 )
                 if ctx.status is None:
                     ctx.set_resource_state(const.HandlerResourceState.deployed)
+            except (ReferenceMissingError, MutatorMissingError) as e:
+                ctx.set_resource_state(const.HandlerResourceState.unavailable)
+                ctx.exception(
+                    "An error occurred during reference resolution for resource %(resource_id)s (exception: %(exception)s",
```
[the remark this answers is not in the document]
Hugo-Inmanta: Tried to improve this message

src/inmanta/loader.py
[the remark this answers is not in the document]
sanderr: Not sure why the coloring is off here, but I was referring to the `inmanta.resources` import iirc. `resource` is not something we want to import to greedily, it has caused us problems in the past.

src/inmanta/deploy/tasks.py
Hugo-Inmanta: > Base... are not intended to be raised afaik

I couldn't find anything against it in the [docs](https://docs.python.org/3.13/library/exceptions.html#BaseExceptionGroup). I'm not very familiar with these, iirc it was suggested to me in a previous code review to use them [here](https://github.com/inmanta/inmanta-core/blob/9f4949d60a15306e6c7b849a1c77cf301876af5e/src/inmanta/server/protocol.py#L243)

Hugo-Inmanta: > or only the one that failed

Current behavior is that if the currently deploying resource's module's name is in the set of failed modules for the executor, we report all these failed modules. I think this is necessary if e.g. a loading failure occurs in a "dependent" file and not in the "main" module in which the resource lives.
The drawback is that in case of multiple, unrelated module loading failures, they would all be reported for each failing resource.

I don't have a good sense of which is best.

sanderr: > I think this is necessary if e.g. a loading failure occurs in a "dependent" file and not in the "main" module in which the resource lives.

I agree that this is an important consideration. Do you mean to say that if `x` fails because it tries to `import y`, you think that a failure that points to both is clearer than simply a failure that says that `x` failed to load?

Hugo-Inmanta: Yes

sanderr: I find it difficult to make a decision. I think there is something to say for either option. I'm fine with this.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -655,6 +649,78 @@ def _get_skipped_for_undeployable(
...
+
+        base_version_data: dict[tuple[str, str], list[str]] = await AgentModules.get_agents_per_module(
+            model_version=partial_base_version, environment=environment, connection=connection
+        )
+        for inmanta_module_name, module_data in module_version_info.items():
+            module_version = module_data.version
+            if not all(
+                [
+                    (inmanta_module_name, module_version) in base_version_data,
+                    set(module_data.for_agents).issubset(base_version_data[(inmanta_module_name, module_version)]),
```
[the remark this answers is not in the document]
Hugo-Inmanta: Which part ? It currently goes like this: 
We want to check that all the code being exported was already registered:
1. each pair of module(name, version) is there in the base model version
2. the relevant agents for this partial version were already registered to use this module(name, version)

src/inmanta/data/__init__.py
```diff
@@ -4439,6 +4450,7 @@ async def get_resource_events(
         return collector
 
     def to_dto(self) -> m.ResourceAction:
+        LOGGER.debug("RA TO DTO")
```
sanderr, line 4453: I expect you're aware of it, but marking it just in case.

src/inmanta/deploy/persistence.py
```diff
@@ -121,7 +121,7 @@ def __init__(self, client: Client, environment: UUID) -> None:
         self.client = client
 
     def get_connection(self, connection: Optional[Connection] = None) -> AbstractAsyncContextManager[Connection]:
-        return data.Scheduler.get_connection()
+        return data.Scheduler.get_connection(connection)  # TODO I think ?
```
Hugo-Inmanta, line 124: This looks like a bug, or else it needs a comment.

wouterdb (approved):

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
sanderr: I'd check with @wouterdb but I believe that 2 is not correct. The motivation for requiring all code to be registered is to make sure that we don't end up with some resources needing one version, while others (that weren't part of the partial compile) need another. Nothing prevents us from expanding the current code to more agents. Indeed, it would be very restrictive if we did so, and I suspect it would break a lot of solutions modules.

sanderr (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -6658,6 +6633,60 @@ async def disconnect_pool() -> None:
...
+    Overwriting the on_connect method here removes this redundant behaviour. The
+    configuration for json serialization is set once when the asyncpg pool is
+    created
+    """
+
+    def on_connect(self) -> None:
+        return None
+
+
+registry.impls["postgresql.asyncpgnoi"] = lambda: ExternalInitAsyncPG
```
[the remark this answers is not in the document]
wouterdb: if we don't remove the init, we will re-initialize this on every connection we get out of the pool. The intialization is async, so no super trivial. 

You could be correct that this isn't worth it. But it was easier to disable it compared to figuring out the overhead.

I might be better to actually upgrade the pool to behave more like an actual pool (.e.g. keep a weak dict from asynpg connection to sqlalchemy connection) .

(But I didn't want this to drag on longer than it already has)

mypy-baseline.txt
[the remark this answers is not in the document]
Hugo-Inmanta: > EDIT: all the new errors are for places where the json.loads was dropped?

Yes, a heads-up could have been useful here.

src/inmanta/data/__init__.py
```diff
@@ -1145,14 +1147,11 @@ def _from_db_single(self, name: str, value: object) -> object:
 
         # asyncpg does not convert a jsonb field to a dict
         if isinstance(value, str) and self.field_type is dict:
-            return json.loads(value)
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think you're right, I think this condition can go.

src/inmanta/db/versions/v202503030.py
```diff
@@ -136,7 +135,7 @@ async def fetch_code_data() -> tuple[VersionsPerEnv, dict[int, dict[str, set[str
 
             assert isinstance(model_version, int)
 
-            for file_hash, file_data in json.loads(source_refs).items():  # type: ignore
+            for file_hash, file_data in source_refs.items():  # type: ignore
```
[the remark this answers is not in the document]
Hugo-Inmanta: All clear :+1:

src/inmanta/server/services/orchestrationservice.py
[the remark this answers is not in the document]
Hugo-Inmanta: I think you're right, I'll add some test cases as well.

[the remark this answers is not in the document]
Hugo-Inmanta: Additional testing added here: https://github.com/inmanta/inmanta-core/blob/658f647fa4a940598a2d38dfebc5573be58dcd34/tests/deploy/e2e/test_code_loading.py#L553

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
Hugo-Inmanta: @wouterdb wdyt ?

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
Hugo-Inmanta: How do we proceed with this, do I create a follow-up ticket ? Should it be part of the epic as well ?

Hugo-Inmanta:
@wouterdb I could use your input there:
- https://github.com/inmanta/inmanta-core/pull/8816#discussion_r2097359859
- https://github.com/inmanta/inmanta-core/pull/8816#discussion_r2097364903
- This one is an outdated conversation, in the latest version I removed the overly restrictive check on the agents: https://github.com/inmanta/inmanta-core/pull/8816/files/5d7f8934f1cdee6e7a19e6fc0245edc8ae093aa5..06df1807241af0590de0ebc8cad2da0840030a21#r2095542344

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
Hugo-Inmanta: Now that the pool is configured to de-serialize, I don't think this case is reached anymore, is it ?
```diff
-# asyncpg does not convert a jsonb field to a dict
-if isinstance(value, str) and self.field_type is dict:
-    return value
```

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
Hugo-Inmanta: Ticketized [here](https://github.com/inmanta/inmanta-core/issues/9123) but with very little context.

PR #8825, opened 2025-03-12
Implement GraphQL endpoint to figure out if environment is compiling

### LINKED ISSUE #8737: Implement GraphQL endpoint to determine if a compile is running
This is the call that the frontend uses:

HEAD /api/v1/notify/
We check whether a compile is running to display a fading dot besides the Compile Reports navigation item. 

We might not need to access the database to implement this

jptrindade:
# Description

Add `isCompiling` field to GraphQL `environments` query

closes #8737 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/graphql/graphql.py
+++ b/src/inmanta/graphql/graphql.py
@@ -14,17 +14,29 @@ Contact: code@inmanta.com
 
 from typing import Any
 
-import inmanta.graphql.schema
+from inmanta.graphql.schema import GraphQLMetadata, get_schema
 from inmanta.protocol import methods_v2
 from inmanta.protocol.decorators import handle
-from inmanta.server import SLICE_GRAPHQL, protocol
+from inmanta.server import SLICE_COMPILER, SLICE_GRAPHQL, protocol
+from inmanta.server.protocol import Server
+from inmanta.server.services.compilerservice import CompilerService
 
 
 class GraphQLSlice(protocol.ServerSlice):
+    metadata: GraphQLMetadata
 
     def __init__(self) -> None:
         super().__init__(name=SLICE_GRAPHQL)
 
+    def get_dependencies(self) -> list[str]:
+        return [SLICE_COMPILER]
+
+    async def prestart(self, server: Server) -> None:
+        compiler_service = server.get_slice(SLICE_COMPILER)
+        assert isinstance(compiler_service, CompilerService)
+        self.metadata = GraphQLMetadata(compiler_service=compiler_service)
+        await super().prestart(server)
+
     @handle(methods_v2.graphql)
     async def graphql(self, query: str) -> Any:  # Actual return type: strawberry.types.execution.ExecutionResult
-        return await inmanta.graphql.schema.get_schema().execute(query)
+        return await get_schema(self.metadata).execute(query)
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -77,6 +77,10 @@ class Environment:
 
 
 def get_schema(metadata: GraphQLMetadata) -> strawberry.Schema:
+    """
+    Fetches strawberry.Schema instance.
+    Initializes schema and metadata if they are None
+    """
     global METADATA
     global SCHEMA
     if SCHEMA is None:
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -12,11 +12,13 @@ limitations under the License.
 Contact: code@inmanta.com
 """
 
+import dataclasses
 import typing
 
 import inmanta.data.sqlalchemy as models
 import strawberry
 from inmanta.data import get_session, get_session_factory
+from inmanta.server.services.compilerservice import CompilerService
 from sqlalchemy import Select, asc, desc, select
 from strawberry import relay
 from strawberry.schema.config import StrawberryConfig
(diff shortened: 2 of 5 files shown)


src/inmanta/graphql/schema.py
```diff
@@ -12,27 +12,48 @@
...
 
+
+@dataclasses.dataclass
+class GraphQLMetadata:
+    compiler_service: CompilerService
+
+
 mapper: StrawberrySQLAlchemyMapper[typing.Any] = StrawberrySQLAlchemyMapper()
 SCHEMA: strawberry.Schema | None = None
+METADATA: GraphQLMetadata | None = None
```
wouterdb, line 37: I would get rid of these globals and wrap this whole module in a method, ... 

less change of leaks

jptrindade: Is this what you meant? 
 - I couldn't extract `Environment` because I need the metadata for the resolver field (I don't think I can pass it as an argument, or at least I haven't figured out how to do it)
 - We are not caching the Schema, we are building it every time we call `get_schema`. Don't know if it impacts performance

wouterdb (approved):

tests/test_graphql.py
```diff
@@ -256,3 +259,66 @@ async def test_query_environments_with_paging(server, client, setup_database):
...
+    # Assert that GraphQL reports that environment is compiling
+    result = await client.graphql(query=query)
+    assert result.code == 200
+    # Check if regular endpoint confirms that it is compiling
+    regular_check = await client.is_compiling(environment)
+    assert regular_check.code == 200
+    assert result.result["data"] == get_response(is_compiling=True)
+
+    # Finish compile
+    del compilerslice._env_to_compile_task[uuid.UUID(environment)]
```
arnaudsjs, line 319: This is quite dirty. It relies on knowledge about the internals of the compiler service, which is bound to change and break the test case. Why not rely on `run_compile_and_wait_until_compile_is_done()` instead?

src/inmanta/graphql/graphql.py
```diff
@@ -14,17 +14,29 @@
...
     def __init__(self) -> None:
         super().__init__(name=SLICE_GRAPHQL)
 
+    def get_dependencies(self) -> list[str]:
+        return [SLICE_COMPILER]
+
+    async def prestart(self, server: Server) -> None:
+        compiler_service = server.get_slice(SLICE_COMPILER)
+        assert isinstance(compiler_service, CompilerService)
+        self.metadata = GraphQLMetadata(compiler_service=compiler_service)
```
arnaudsjs, line 37: `self.metadata` is not initialized in the `__init__()`

src/inmanta/graphql/graphql.py
```diff
@@ -14,17 +14,29 @@
...
+        await super().prestart(server)
+
     @handle(methods_v2.graphql)
     async def graphql(self, query: str) -> Any:  # Actual return type: strawberry.types.execution.ExecutionResult
-        return await inmanta.graphql.schema.get_schema().execute(query)
+        return await get_schema(self.metadata).execute(query)
```
arnaudsjs, line 42: Do you know why this is not a type error? There is no guarantee that `self.medata is not None`. Is that because we didn't define in the `__init__()`?

src/inmanta/graphql/schema.py
```diff
@@ -12,53 +12,24 @@
...
-
-
-def get_schema() -> strawberry.Schema:
-    global SCHEMA
-    if SCHEMA is None:
-        SCHEMA = initialize_schema()
-    return SCHEMA
+@dataclasses.dataclass
+class GraphQLMetadata:
+    compiler_service: CompilerService
```
arnaudsjs, line 32: * I don't understand the name of this class. Metadata means data about data. But this class holds the compiler service, which is not data at all. I have the feeling this represents the context in which graphQL is executing. Would it be better to call this `GraphQLContext`?
* It could also use a docstring to say what is actually represents.

src/inmanta/graphql/schema.py
```diff
@@ -101,14 +61,60 @@ def add_filter_and_sort(
...
     """
     Initializes the Strawberry GraphQL schema.
     It is initiated in a function instead of being declared at the module level, because we have to do this
     after the SQLAlchemy engine is initialized.
     """
+
+    mapper: StrawberrySQLAlchemyMapper[typing.Any] = StrawberrySQLAlchemyMapper()
     loader = StrawberrySQLAlchemyLoader(async_bind_factory=get_session_factory())
 
+    def get_expert_mode(root: "Environment") -> bool:
```
arnaudsjs, line 74: With the string-based type annotation it's rather hard to see which environment type is intended. We have multiple classes called Environment. I think it would help to make this not a string-based type.

src/inmanta/graphql/schema.py
```diff
@@ -101,14 +61,60 @@ def add_filter_and_sort(
...
+        """
+        Checks settings of environment to figure out if expert mode is enabled or not
+        """
+        assert hasattr(root, "settings")  # Make mypy happy
+        return bool(root.settings.get("enable_lsm_expert_mode", False))
```
arnaudsjs, line 79: This I don't understand. What is the return type of `root.settings.get()`.

* If it's a boolean, the type cast is redundant.
* If it's a string, the type cast is incorrect.

I might be missing something here.

src/inmanta/graphql/schema.py
```diff
@@ -101,14 +61,60 @@ def add_filter_and_sort(
...
 
-def initialize_schema() -> strawberry.Schema:
+def get_schema(metadata: GraphQLMetadata) -> strawberry.Schema:
     """
     Initializes the Strawberry GraphQL schema.
     It is initiated in a function instead of being declared at the module level, because we have to do this
     after the SQLAlchemy engine is initialized.
     """
+
+    mapper: StrawberrySQLAlchemyMapper[typing.Any] = StrawberrySQLAlchemyMapper()
```
arnaudsjs, line 71: Is this an expensive operation? Because this would be executed for each test case, if I understand correctly.

src/inmanta/graphql/schema.py
```diff
@@ -101,14 +61,60 @@ def add_filter_and_sort(
...
+
+    @strawberry.input
+    class EnvironmentFilter(StrawberryFilter):
+        id: typing.Optional[str] = strawberry.UNSET
+
+    @strawberry.input(one_of=True)
+    class EnvironmentOrder(StrawberryOrder):
+        id: typing.Optional[str] = strawberry.UNSET
+        name: typing.Optional[str] = strawberry.UNSET
+
```
arnaudsjs, line 117: I am not up to speed yet with strawberry. Do I see correctly that this is not tested. Does it work?

arnaudsjs (changes requested):

src/inmanta/graphql/schema.py
jptrindade: I tested it in: `tests/test_graphql.py:test_query_environments_with_filtering` These tests were merged in my first PR

src/inmanta/graphql/schema.py
jptrindade: But this has to be defined before `Environment`. What do you suggest to use instead?

arnaudsjs: I see. I don't think there is much we an do about that.

src/inmanta/graphql/schema.py
jptrindade: That is correct, I extracted it from the method. Doesn't seem to be very expensive

src/inmanta/graphql/schema.py
jptrindade: `root.settings` is a `dict[str, Union[bool, int, float, str, dict[str, Union[str, int, bool]]]]`. I believe for this setting in particular it should be a bool. I'll make the comparison more robust either way

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #8833, opened 2025-03-14
Ignore zombie and terminated processes in process discovery

arnaudsjs:
# Description

Calling a method on a process that became a zombie or has terminated, raises an exception. If this happens the test suite fails on that. This fix makes sure that we ignore the exception.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/ignore-zombie-and-terminated-processes.yml
+++ b/changelogs/unreleased/ignore-zombie-and-terminated-processes.yml
@@ -1,0 +1,4 @@
+---
+description: Fix bug in the test suite where the subprocess discovery fails if a process became a zombie or terminated while processing it.
+change-type: patch
+destination-branches: [master, iso8]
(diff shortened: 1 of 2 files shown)


Hugo-Inmanta (approved):

arnaudsjs:
@Hugo-Inmanta I removed the check for unexpected processes. It was implemented incorrectly. When I fix the bug I need to add a lot more processes to the list. I don't really see the added value of the check, so I think it's safe to remove. Do you agree?

Hugo-Inmanta:
>@Hugo-Inmanta I removed the check for unexpected processes. It was implemented incorrectly. When I fix the bug I need to add a lot more processes to the list. I don't really see the added value of the check, so I think it's safe to remove. Do you agree?

Looks good to me

Hugo-Inmanta (approved):

PR #8840, opened 2025-03-17
Fix race condition in utils.wait_for_version method

arnaudsjs:
# Description

Fix race condition in the `utils.wait_for_version()` method where the `compile_done()` method returns True if the compile was not yet picked up by the compiler service.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -337,7 +337,7 @@ async def wait_until_version_is_released(client, environment: uuid.UUID, version
     await retry_limited(_is_version_released, timeout=10)
 
 
-async def wait_for_version(client, environment, cnt, compile_timeout: int = 30):
+async def wait_for_version(client, environment, cnt: int, compile_timeout: int = 30):
     """
     :param compile_timeout: Raise an AssertionError if the compilation didn't finish after this amount of seconds.
     """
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -337,20 +337,21 @@ async def _is_version_released() -> bool:
...
 
     # Wait until the server is no longer compiling
     # wait for it to finish
     async def compile_done():
-        compiling = await client.is_compiling(environment)
-        code = compiling.code
-        return code == 204
+        result = await client.get_reports(environment)
+        assert result.code == 200
+        return all(r["success"] is not None for r in result.result["reports"])
```
jptrindade, line 350: I'm not sure that this is the correct behaviour. For instance, what if we want a compile to fail. Or if one of our previous compiles failed?

arnaudsjs: This method only works if all compiles finish successfully. See `assert report["success"]` further down.

Hugo-Inmanta: I don't understand how this fixes the issue, can't the exact same thing as before happen e.g. `the compile_done() method returns True if the compile was not yet picked up by the compiler service.` ?

Hugo-Inmanta (changes requested):
(See comment)

tests/utils.py
arnaudsjs: I think this test case fails in the following scenario:

1. The compiler server is started as part of the startup of the server. It calls into the `_recover()` method asynchronously.
2. The test case requests a new compile. This compile is written to the database, but the compiler service doesn't pick it up (i.e. it's not put into the `self._env_to_compile_task` instance variable) because the `_recover()` method didn't set the `fully_ready` variable to True yet.
3. The test case calls into `utils.wait_for_version()` method and assumes the compile has finished because the `self._env_to_compile_task` dictionary is empty.

By relying on the fact that the compile record always exists in the database, we can prevent this issue.

Hugo-Inmanta (approved):

PR #8841, opened 2025-03-17
Add useful consts to stable API

### LINKED ISSUE #8803: Add inmanta.const.DONE_STATES to stable api
The const file contains some very useful sets of values for testing libraries (such as pytest-inmanta-lsm): `DONE_STATES`, `NOT_DONE_STATES`, etc.

Can these constants be added to the stable api?

jptrindade:
# Description

  Adds the following consts to the stable api:
    - UNDEPLOYABLE_STATES
    - TRANSIENT_STATES
    - NOT_DONE_STATES
    - DONE_STATES

closes #8803 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/8803-add-consts-to-stable-api.yml
+++ b/changelogs/unreleased/8803-add-consts-to-stable-api.yml
@@ -2,13 +2,9 @@ change-type: patch
 description: >
   Adds the following consts to the stable api:
     - UNDEPLOYABLE_STATES
-    - UNDEPLOYABLE_NAMES
     - TRANSIENT_STATES
     - NOT_DONE_STATES
     - DONE_STATES
-    - INITIAL_STATES
-    - TERMINAL_STATES
-    - VALID_STATES_ON_UPDATE
 
 destination-branches: [master]
 sections:
--- a/changelogs/unreleased/8803-add-consts-to-stable-api.yml
+++ b/changelogs/unreleased/8803-add-consts-to-stable-api.yml
@@ -1,0 +1,15 @@
+change-type: patch
+description: >
+  Adds the following consts to the stable api:
+    - UNDEPLOYABLE_STATES
+    - UNDEPLOYABLE_NAMES
+    - TRANSIENT_STATES
+    - NOT_DONE_STATES
+    - DONE_STATES
+    - INITIAL_STATES
+    - TERMINAL_STATES
+    - VALID_STATES_ON_UPDATE
+
+destination-branches: [master]
+sections:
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/8803-add-consts-to-stable-api.yml
```diff
@@ -0,0 +1,15 @@
...
+    - TRANSIENT_STATES
+    - NOT_DONE_STATES
+    - DONE_STATES
+    - INITIAL_STATES
+    - TERMINAL_STATES
+    - VALID_STATES_ON_UPDATE
+
+destination-branches: [master]
+sections:
+  other-note: "{{description}}"
```
jptrindade, line 15: I feel like this should be visible to the end user but not sure in which section it falls under

src/inmanta/const.py
```diff
@@ -77,6 +77,18 @@ class DeprecatedResourceState(str, Enum):
...
+The following consts are considered to be part of the stable API.
+Modifying them may consist a breaking change for some libraries:
+    - UNDEPLOYABLE_STATES
+    - UNDEPLOYABLE_NAMES
+    - TRANSIENT_STATES
+    - NOT_DONE_STATES
+    - DONE_STATES
+    - INITIAL_STATES
+    - TERMINAL_STATES
+    - VALID_STATES_ON_UPDATE
```
jptrindade, line 90: Is there any of these consts that is more volatile and shouldn't be added to the stable api?

src/inmanta/const.py
```diff
@@ -77,6 +77,18 @@ class DeprecatedResourceState(str, Enum):
     processing_events = "processing_events"
 
 
+"""
+The following consts are considered to be part of the stable API.
+Modifying them may consist a breaking change for some libraries:
```
arnaudsjs, line 82:
```suggestion
Modifying them may break some libraries:
```

arnaudsjs (changes requested):
I can't judge whether some const should be part of the stable API of not. It all depends on whether it's used by the solutions team or not. I'll leave that assessment to Guillaume.

arnaudsjs (approved):

edvgui:
The only states for which I could easily find usage by the solutions team were:
- `UNDEPLOYABLE_STATES`
- `DONE_STATES`
- `TRANSIENT_STATES`

But it would probably make sense to also add `NOT_DONE_STATES`.  The rest I am not sure what it means to be honest

edvgui (approved):

PR #8852, opened 2025-03-19
Fix SQLAlchemy duplicated logs

### LINKED ISSUE #8798: SQLAlchemy producing duplicated logs
Try to remove the uncolored logs

jptrindade:
# Description

SQLAlchemy was producing duplicated logs. This is because the `echo` was enabled which copies the logs to stdout. 
This is useful in the tests but when spinning up a server we get these duplicated logs.

I tried to make `echo=True` only while testing. 

closes #8798 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/logging.py
+++ b/src/inmanta/logging.py
@@ -282,6 +282,7 @@ class Options(Namespace):
                               instead of the component of the compiler that was executing while the log record was created
                               or the name of the module that created the log message.
     :param logging_config: Path to the dict-based logging config file.
+    :param echo_sqlalchemy_logging: Output SqlAlchemy logs to stdout.
     """
 
     log_file: Optional[str] = None
@@ -290,6 +291,7 @@ class Options(Namespace):
     timed: bool = False
     keep_logger_names: bool = False
     logging_config: Optional[str] = None
+    echo_sqlalchemy_logging: bool = False
 
 
 class LoggingConfigBuilderExtension(abc.ABC):
--- a/changelogs/unreleased/8798-fix-sqlalchemy-duplicated-logs.yml
+++ b/changelogs/unreleased/8798-fix-sqlalchemy-duplicated-logs.yml
@@ -1,0 +1,4 @@
+description: Fix SQLAlchemy duplicated logs.
+issue-nr: 8798
+change-type: patch
+destination-branches: [master]
--- a/src/inmanta/server/bootloader.py
+++ b/src/inmanta/server/bootloader.py
@@ -95,7 +95,7 @@ class InmantaBootloader:
 
         if configure_logging:
             inmanta_logger_config = inmanta_logging.InmantaLoggerConfig.get_instance()
-            inmanta_logger_config.apply_options(inmanta_logging.Options())
+            inmanta_logger_config.apply_options(inmanta_logging.Options(echo_sqlalchemy_logging=True))
 
     async def start(self) -> None:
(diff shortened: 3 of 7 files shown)


src/inmanta/data/__init__.py
```diff
@@ -6624,11 +6624,13 @@ async def start_engine(
...
     max_overflow: int = 0,
     pool_timeout: float = 60.0,
-    echo: bool = False,
 ) -> None:
     """
     Start the SQL Alchemy engine for this process
     """
+    # Configure logging to emit sqlalchemy logs without relying on the echo flag.
+    # The echo flag also writes to stdout which results in duplicate logs
+    logging.getLogger("sqlalchemy.engine").setLevel(logging.INFO)
```
jptrindade, line 6633: Is this the correct place to have this?

arnaudsjs: We have configurable logging on iso9. We should not longer hard code logging config. I would add this to the relevant logging config files in `inmanta-core/misc/`

arnaudsjs (changes requested):

Hugo-Inmanta (approved):

misc/server_log.yml
```diff
@@ -37,6 +37,11 @@ loggers:
     - core_tornado_debug_log_handler
     level: DEBUG
     propagate: true
+  sqlalchemy.engine:
+    handlers:
+      - server_handler
+    level: INFO
+    propagate: true
```
wouterdb, line 44: this will cause duplicate logs, as the `server_handler`  handler is both attached at the root and here.

misc/server_log.yml
```diff
@@ -37,6 +37,11 @@ loggers:
     - core_tornado_debug_log_handler
     level: DEBUG
     propagate: true
+  sqlalchemy.engine:
+    handlers:
+      - server_handler
```
wouterdb, line 42:
```suggestion
```


otherwise this will cause duplicate logs, as the server_handler handler is both attached at the root and here.


(applies everywhere)

wouterdb: Morever, why would we need a custom config? We already set the root level?

wouterdb (changes requested):

misc/server_log.yml
jptrindade: I confess that my understanding of custom logging is a bit limited.
By custom config do you mean just the handler part or simply adding this entry to the `loggers` ?

arnaudsjs (approved):

src/inmanta/logging.py
```diff
@@ -443,7 +443,11 @@ def get_logging_config_from_options(
                     "level": "DEBUG",
                     "propagate": True,
                     "handlers": ["core_tornado_debug_log_handler"],
-                }
+                },
+                "sqlalchemy.engine": {
+                    "level": "INFO",
```
wouterdb, line 448: De we want this to be fixed to info or do we want it to go up and down with the server?

jptrindade: Does it make a difference?
`sqlalchemy.engine`  will emit the log at `INFO`
if the  server is at `WARNING` it will not show the log, right?

jptrindade: I guess it doesn't hurt and might be useful with other possible sqlalchemy logs at different levels

wouterdb (approved):

Issue #8398, opened 2024-11-22
Support classes that implement __getitem__ and __setitem__ in dictpath


bartv:
For references we want to use resources with dict path. In the current pr this is implemented by:

- Adding `__getitem__` and `__setitem__` to resource
- Have dict path check for this methods (the correct protocol) instead of doing an isinstance

The goal of this ticket is to properly implement the type guard (Mapping vs Mutable mapping).

The changes for references to resource:

```diff
diff --git a/src/inmanta/resources.py b/src/inmanta/resources.py
index d91d98188..50d090a47 100644
--- a/src/inmanta/resources.py
+++ b/src/inmanta/resources.py
@@ -388,7 +515,47 @@ class Resource(metaclass=ResourceMeta):
 
         self.version = _id.version
 
-    def populate(self, fields: dict[str, Any], force_fields: bool = False) -> None:
+    def __getitem__(self, key: str) -> object:
+        """Support dict like access on the resource"""
+        if key in self.fields:
+            return getattr(self, key)
+
+        raise KeyError()
+
+    def __setitem__(self, key: str, value: object) -> None:
+        """Support dict like access on the resource. It is not possible to create new
+        attributes using setitem
+        """
+        if key in self.fields:
+            return setattr(self, key, value)
+
+        raise KeyError()
+

```

And to dictpath:

```diff
diff --git a/src/inmanta/util/dict_path.py b/src/inmanta/util/dict_path.py
index 5d936c51b..8818661e5 100644
--- a/src/inmanta/util/dict_path.py
+++ b/src/inmanta/util/dict_path.py
@@ -271,8 +271,23 @@ class WildDictPath(abc.ABC):
     def parse(cls: type[TWDP], inp: str) -> Optional[TWDP]:
         pass
 
-    def _validate_container(self, container: object) -> TypeGuard[dict[object, object]]:
-        return isinstance(container, dict)
+    def _validate_container(
+        self, container: object, set: bool = False, remove: bool = False
+    ) -> TypeGuard[dict[object, object]]:
+        """Validate that the container supports the required mapping protocol operations
+
+        TODO: fix the typeguard. Probably by providing multiple methods for validation!
+        """
+        if not hasattr(container, "__getitem__"):
+            return False
+
+        if set and not hasattr(container, "__setitem__"):
+            return False
+
+        if remove and not hasattr(container, "__delitem__"):
+            return False
+
+        return True
 
 
 @stable_api
[...]

PR #8861, opened 2025-03-24
Issue/deeper mutators

wouterdb:
# Description

1. This fixes the issue where nested mutators don't work
2. I don't like it because
  - Dictpath does a lot of iterating over dicts, which is often not required
  - Resource is becoming some sort of everything object

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/util/dict_path.py
+++ b/src/inmanta/util/dict_path.py
@@ -272,7 +272,7 @@ class WildDictPath(abc.ABC):
         pass
 
     def _validate_container(
-        self, container: object, set: bool = False, remove: bool = False
+        self, container: object, set: bool = False, remove: bool = False, with_items: bool = False
     ) -> TypeGuard[dict[object, object]]:
         """Validate that the container supports the required mapping protocol operations"""
         # To be refined in https://github.com/inmanta/inmanta-core/issues/8398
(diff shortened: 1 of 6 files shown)


bartv:
> I agree with Guillaume that it would be good to clearly define the interface that objects need to implement so that the dictpath library can act on them.

I think implicitly dictpath uses Mapping and MutableMapping. For me the solutions is:

- Dictpath expects a Mapping when it only needs to return values and it expects ad MutableMapping when it needs to replace values
- Resource should implement collections.abc.MutableMapping which mean that instead of items it should add delitem, iter and len as well. For delitem we should raise an error that it is not supported

That also means we can probably get rid of the _validate_container method and replace it we isinstance checks

bartv (commented):
see https://github.com/inmanta/inmanta-core/pull/8861#issuecomment-2747517630

bartv (approved):
This ticket can also be closed now https://github.com/inmanta/inmanta-core/issues/8398

src/inmanta/util/dict_path.py
```diff
@@ -33,6 +33,34 @@
 TWCP = TypeVar("TWCP", bound="WildComposedPath")
 
 
+@runtime_checkable
+class Mapping(Protocol):
```
wouterdb, line 37: The reason we did this is that the meta class of resource and abc.Mapping don't get along

arnaudsjs (approved):

PR #8862, opened 2025-03-24
Fix pep8

wouterdb:
# Description

Fix for connection pool leak

see issue: see https://github.com/sqlalchemy/sqlalchemy/discussions/12460

# Self Check:

- [x] Changelog entry


changelogs/unreleased/8762-integrate-sql-alchemy-2.yml
```diff
@@ -0,0 +1,3 @@
+description: Fix SQLAlchemy pool leak,
```
Hugo-Inmanta, line 1:
```suggestion
description: Fix SQLAlchemy pool leak.
```

PR #8871, opened 2025-03-27
Update filter documentation

### LINKED ISSUE #8802: Api filtering inconsistency
cf. https://inmanta.slack.com/archives/CKRF0C8R3/p1740992612025799

The api documentation states the following:
> It's also possible to provide multiple values for the same filter, in this case resources are returned, if they match any of these filter values.
> Multiple different filters narrow the results however (they are treated as an 'AND' operator).
> cf. https://github.com/inmanta/inmanta-core/blob/d25205bdd49016596ad7653597a2cc99a8ed3992/src/inmanta/protocol/methods_v2.py#L685-L701

But this query will actually perform an `AND` operation (get all resources which are neither `deployed` nor `orphaned`): https://mgmt.inmanta.com/api/v2/resource?tid=f81be80b-8582-499f-b002-7cdd2f2df98a&filter.status=!deployed&filter.status=!orphaned

jptrindade:
# Description

Updates the documentation on resource list filter

closes #8802 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/protocol/methods_v2.py
+++ b/src/inmanta/protocol/methods_v2.py
@@ -701,8 +701,7 @@ def resource_list(
                 For example: `filter.status=!orphaned` will return all the resources that are not in 'orphaned' state.
                 If multiple values are provided to the inverted filter, resources are returned if they don't match
                 any of the filter values.
-                For example: `?filter.status=!deployed&filter.status=!available` returns instances whose status
-                is neither deployed nor available.
+                For example: `?filter.status=!deployed&filter.status=!available` returns all instances except those whose status is deployed or available.
                 The values for the 'agent', 'resource_type' and 'value' filters are matched partially.
     :param sort: Return the results sorted according to the parameter value.
                 It should follow the pattern `<attribute_to_sort_by>.<order>`, for example `resource_type.desc`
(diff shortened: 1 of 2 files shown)


src/inmanta/protocol/methods_v2.py
```diff
@@ -696,8 +696,13 @@ def resource_list(
...
                 status: filter by the current status of the resource.
-                For status filters it's also possible to invert the condition with '!', for example
-                `filter.status=!orphaned` will return all the resources that are not in 'orphaned' state
+                For status filters it's also possible to invert the condition with '!'.
+                The inverted status filter works as a separate filter from the normal status filter.
+                For example: `filter.status=!orphaned` will return all the resources that are not in 'orphaned' state.
+                If multiple values are provided to the inverted filter, resources are returned if they don't match
+                any of the filter values.
+                For example: `?filter.status=!deployed&filter.status=!available` returns instances whose status
+                is neither deployed nor available.
```
wouterdb, line 705:
```suggestion
                For example: `?filter.status=!deployed&filter.status=!available` returns all instances except those whose status is deployed or available.
```

wouterdb (approved):

src/inmanta/protocol/methods_v2.py
jptrindade: Thanks, it is clearer now!

PR #8885, opened 2025-04-01
Add overall status to the status endpoint

### LINKED ISSUE #6813: Add overall status to the status endpoint
Currently, the status endpoint reports a lot of data, but no overall health. 
This makes it hard to monitor 

I would propose to
1. SliceStatus gets a status field that can be `OK`, `Warning`, `Error` 
2. SliceStatus gets a message field
3. Overall StatusResponse gets a status field, that is the worst of reported slice statuses 
4. add method to get status as a `SliceStatus` object at once, to the super class, calling into the existing method for backward compat
5. make the DB endpoint got to error if the connection is not up and to warning if the pool is less than 5 connections or less then 10% of the pool, rounded down whichever is smallest.

jptrindade:
# Description

Moved `GET_SLICE_STATUS_TIMEOUT` up to the generic class, which makes each get status timeout configurable. Not sure if desired but it was helpful because I also moved `get_slice_status` which was using it.

Not sure if `status=max(slices, key=lambda slice: STATUS_ORDER[slice.reported_status]).reported_status` is overengineered

closes #6813 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/server.py
+++ b/src/inmanta/server/server.py
@@ -55,9 +55,6 @@ class Server(protocol.ServerSlice):
     compiler: "CompilerService"
     _server: protocol.Server
 
-    # The number of seconds after which the call to the get_status() endpoint of a server slice should time out.
-    GET_SERVER_STATUS_TIMEOUT: int = 1
-
     def __init__(self) -> None:
         super().__init__(name=SLICE_SERVER)
         LOGGER.info("Starting server endpoint")
--- a/src/inmanta/server/server.py
+++ b/src/inmanta/server/server.py
@@ -28,7 +28,7 @@ from tornado import routing, web
 
 from inmanta import config, const, data
 from inmanta.const import ApiDocsFormat
-from inmanta.data.model import STATUS_ORDER, FeatureStatus, StatusResponse
+from inmanta.data.model import FeatureStatus, ReportedStatus, StatusResponse
 from inmanta.protocol import exceptions, handle, methods, methods_v2
 from inmanta.protocol.common import HTML_CONTENT_WITH_UTF8_CHARSET, ReturnValue, attach_warnings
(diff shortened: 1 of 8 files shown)


src/inmanta/server/server.py
```diff
@@ -169,6 +143,7 @@ async def collect_for_slice(slice_name: str, slice: protocol.ServerSlice) -> Sli
                 FeatureStatus(slice=feature.slice, name=feature.name, value=self.feature_manager.get_value(feature))
                 for feature in self.feature_manager.get_features()
             ],
+            status=max(slices, key=lambda slice: STATUS_ORDER[slice.reported_status]).reported_status,
```
wouterdb, line 146: neat. You could add the integer value to the enumeration, but this works as well .

wouterdb (approved):

src/inmanta/server/server.py
Hugo-Inmanta: Something like that should work:
```py
from enum import StrEnum

class ReportedStatus(StrEnum):
    OK = "OK"
    Warning = "Warning"
    Error = "Error"

    def __gt__(self, other: "ReportedStatus"):
        order = [ReportedStatus.OK, ReportedStatus.Warning, ReportedStatus.Error]
        if self not in order or other not in order:
            raise ValueError
        return order.index(self) > order.index(other)


ok = ReportedStatus.OK
wa = ReportedStatus.Warning
er = ReportedStatus.Error


print(max([wa, er, ok]))    # Error
print(max([wa, er]))        # Error
print(max([wa, ok]))        # Warning
print(max([ok]))            # OK





```

Hugo-Inmanta (approved):
LGTM apart from very minor comments :+1: .

Hugo-Inmanta (approved):

PR #8894, opened 2025-04-04
Allow json references

wouterdb:
# Description

Allow dict in references

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/references.py
+++ b/src/inmanta/references.py
@@ -28,11 +28,12 @@ from typing import Literal, Tuple
 
 import pydantic
 import typing_inspect
+from pydantic import ValidationError
 
 import inmanta
 import inmanta.resources
 from inmanta import util
-from inmanta.types import ResourceIdStr
+from inmanta.types import ResourceIdStr, StrictJson
 from inmanta.util import dict_path
 
 ReferenceType = typing.Annotated[str, pydantic.StringConstraints(pattern="^([a-z0-9_]+::)+[A-Z][A-z0-9_-]*$")]
@@ -122,6 +123,20 @@ class LiteralArgument(Argument):
         return self.value
 
 
+class JsonArgument(Argument):
+    """Literal argument to a reference"""
+
+    type: typing.Literal["json"] = "json"
+    value: StrictJson
+
+    def get_arg_value(
+        self,
+        resource: "inmanta.resources.Resource",
+        logger: "handler.LoggerABC",
+    ) -> object:
+        return self.value
+
+
 class ReferenceArgument(Argument):
     """Use the value of another reference as an argument for the reference"""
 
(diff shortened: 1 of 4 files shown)


src/inmanta/types.py
```diff
@@ -60,6 +62,11 @@ def issubclass(sub: type, super: Union[type, tuple[type, ...]]) -> bool:
 
 JsonType = dict[str, Any]
 ReturnTupple = tuple[int, Optional[JsonType]]
+StrictJson = TypeAliasType(
+    "StrictJson",
+    "Union[dict[str, StrictJson], list[StrictJson], str, int, float, bool, None]",
+)
+
```
sanderr, line 69: I think this is equivalent to `type StrictJson = dict[str, StrictJson] | list[StrictJson] | str | int | float | bool | None`, and that `TypeAliasType` is discouraged from direct use.

src/inmanta/references.py
```diff
@@ -122,6 +123,20 @@ def get_arg_value(
         return self.value
 
 
+class JsonArgument(Argument):
+    """Literal argument to a reference"""
```
Hugo-Inmanta, line 127:
```suggestion
    """Json-like argument to a reference"""
```

Hugo-Inmanta (approved):

src/inmanta/references.py
```diff
@@ -122,6 +123,20 @@ def get_arg_value(
         return self.value
 
 
+class JsonArgument(Argument):
```
sanderr, line 126: Why make this a separate argument type? It's just another literal, isn't it? Not opposed, but I don't understand the motivation.

src/inmanta/references.py
```diff
@@ -273,13 +288,16 @@ def serialize_arguments(self) -> Tuple[uuid.UUID, list[ArgumentTypes]]:
             match value:
                 case str() | int() | float() | bool():
                     arguments.append(LiteralArgument(name=name, value=value))
-
+                case dict() | list():
+                    try:
+                        arguments.append(JsonArgument(name=name, value=value))
+                    except ValidationError:
+                        raise ValueError(f"The {name} attribute of {self} is not json serializable {value}")
```
sanderr, line 295:
```suggestion
                        raise ValueError(f"The {name} attribute of {self} is not json serializable: {value}")
```

sanderr (changes requested):

src/inmanta/types.py
wouterdb: thx!

src/inmanta/references.py
wouterdb: I was a bit in doubt myself. I like the simplicity of literal (it is very definite in what is can contain). 

But it is not really a requirement (it's not like we will ever want to build a simpler decoder with support for only some types) . This is probably marginally faster in decoding invalid inputs, which also is not an argument. 

So I have no strong preference, if you want, I change it.

sanderr: No strong preference either, I just wondered.

src/inmanta/types.py
sanderr: This is still the same. Forgot to push?

sanderr (commented):
left a comment on a thread.

sanderr (approved):

PR #8907, opened 2025-04-08
Bump compiler version to 2025.3 on iso8

jptrindade:
# Description

Bump compiler version to 2025.3 on iso8 as part of the release

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/bump-compiler-version-for-release.yml
+++ b/changelogs/unreleased/bump-compiler-version-for-release.yml
@@ -1,4 +1,4 @@
 ---
 description: "Bump compiler version to 2025.3"
 change-type: patch
-destination-branches: [iso8]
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/bump-compiler-version-for-release.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: "Bump compiler version to 2025.3"
+change-type: patch
+destination-branches: [iso8]
```
sanderr, line 4:
```suggestion
destination-branches: [iso8, master]
```

I'd keep it simple and include master for now. Ideally, when we first release master we'll bump it to 2026, or drop it altogether, but for now let's just make sure that we don't start trailing behind.

sanderr (approved):

PR #8943, opened 2025-04-15
Add support to define access policies

arnaudsjs:
# Description

This PR applies the following changes:

* Adds authorization-related metadata to the endpoints using the `@auth` annotation.
* Add support to manage the lifecycle of the Open Policy Agent process.
* Removes support for custom claims. This feature is not used by anyone at the moment and the feature set of the new logic is a strict superset of this old logic.
* Removes support to add multiple `@method`, `@typedmethod` or `@handle` annotations on the same method. Currently we don't do this and removing this support makes this implementation a lot easier.

General logic:

* If the auth token contains the client type "agent" or "compiler, use the old logic to authorize the token
* Otherwise ("client" type) validate the token using the defined access policy.

In a follow up PR, I will add a config option to let the user decide with authorization method (legacy or policy engine) should be used. The `enforce-access-policy` option will disappear. It currently only exists to make sure that the iso9 orchestrators, deployed in the labs, don't break on this change.

# Self Check:

- [x] Changelog entry


src/inmanta/protocol/decorators.py
```diff
@@ -183,7 +180,7 @@ def typedmethod(
 
     def wrapper(func: MethodT) -> MethodT:
         paths = path if isinstance(path, list) else [path]
-        for current_path in paths:
+        for index, current_path in enumerate(paths):
```
wouterdb, line 183: This I don't understand?

src/inmanta/protocol/rest/__init__.py
```diff
@@ -645,3 +693,10 @@ async def _execute_call(
         except Exception as e:
             LOGGER.exception("An exception occurred during the request.")
             raise exceptions.ServerError(str(e.args))
+
+    @abc.abstractmethod
+    async def get_policy_engine_slice(self) -> Optional["PolicyEngineSlice"]:
+        """
+        Return the policy-engine slice or None if no such slice exists because we are not running on the server.
+        """
+        raise NotImplementedError()
```
wouterdb, line 702: I would expect this to only exist on the server side?

src/inmanta/server/services/policy_engine_service.py
```diff
@@ -0,0 +1,199 @@
...
+            self.process = await asyncio.create_subprocess_exec(
+                opa_binary,
+                "run",
+                "--server",
+                "--addr",
+                await self.get_addr_policy_engine(),
+                "--log-format",
+                "text",
+                "--log-level",
+                "debug",
```
wouterdb, line 151: do we always want debug? (just asking)

src/inmanta/server/services/policy_engine_service.py
```diff
@@ -0,0 +1,199 @@
...
+            return False
+
+
+class OpaServer:
+
+    def __init__(self) -> None:
+        self.process: asyncio.subprocess.Process | None = None
+        self.running = False
+
+    async def get_addr_policy_engine(self) -> str:
```
wouterdb, line 108: why is this async?

src/inmanta/server/services/policy_engine_service.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        Returns the "<host>:<port>" on which the policy engine should listen for incoming connections.
+        """
+        return f"{policy_engine_bind_address.get()}:{policy_engine_bind_port.get()}"
```
wouterdb, line 112: I wonder if we want to go for a unix socket here? 

https://github.com/open-policy-agent/opa/pull/752/files
https://github.com/tornadoweb/tornado/issues/2671

i.e. I wonder which security gate of hell we open by using an unauthenticated connection for authantication. 

An attacker could race the start of the OPA engine and take its port? 

or enable auth and SSL to opa, but that will be difficult

wouterdb (changes requested):

src/inmanta/server/services/policy_engine_service.py
wouterdb: also

```
      --unix-socket-perm string              specify the permissions for the Unix domain socket if used to listen for incoming connections (default "755")
```

src/inmanta/protocol/auth/auth.py
```diff
@@ -34,97 +31,9 @@
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicNumbers
 
 from inmanta import config, const
-from inmanta.server import config as server_config
+from inmanta.protocol import exceptions
 
-from . import exceptions
```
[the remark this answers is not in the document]
arnaudsjs: What do you mean with modal here? We had a lot of fun reading through the cryptic definitions of the word, but we didn't manage to come to an interpretation that makes sense :D

src/inmanta/protocol/common.py
```diff
@@ -385,6 +380,7 @@ def __init__(
         strict_typing: bool = True,
         enforce_auth: bool = True,
         varkw: bool = False,
+        set_method_properties_on_fnc: bool = True,
```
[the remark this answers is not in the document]
arnaudsjs: The `@typedmethod` decorator has support to accept a list of paths for the endpoints. This option exists to make the guard, that prevents multiple `@method` or `@typedmethod` annotations on the same endpoint, aware about the fact that this duplicate is acceptable.

src/inmanta/protocol/common.py
```diff
@@ -453,6 +450,22 @@ def __init__(
...
+    def is_internal_endpoint(self) -> bool:
+        """
+        Returns True iff it's an internal endpoint (for communication between the server and the agents)
+        and is not part of the public API.
+        """
+        if self.client_types == [const.ClientType.agent]:
+            return True
+        if (self._agent_server or self._server_agent) and not self._api:
+            return True
+        return False
```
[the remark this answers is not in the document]
arnaudsjs: I think you are right.

src/inmanta/protocol/decorators.py
arnaudsjs: See comment about `set_method_properties_on_fnc`.

src/inmanta/opa/opa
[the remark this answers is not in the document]
arnaudsjs: Yes, we are. But I don't see a better alternative. Although I am becoming more and more in favor or putting the vendored binary into a separate Python package.

src/inmanta/protocol/rest/__init__.py
arnaudsjs: I am not sure what you mean. Could you make a concrete proposal of what you would like to be changed?

src/inmanta/server/services/policy_engine_service.py
arnaudsjs: I wasn't sure either. But if we don't make this configurable, debug is the only real option. But it's probably better if I add a config option for this.

src/inmanta/server/services/policy_engine_service.py
arnaudsjs: No specific reason. Do you know what the general development guideline is for this case?

src/inmanta/protocol/common.py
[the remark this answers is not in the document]
arnaudsjs: I changed this method so that compiler-only endpoints don't need a `@auth` annotation.

src/inmanta/server/services/policy_engine_service.py
```diff
@@ -0,0 +1,199 @@
...
+            self.process = None
+            self.running = False
+            return
+        self.process.terminate()
+        try:
+            await asyncio.wait_for(self.process.wait(), timeout=const.POLICY_ENGINE_GRACE_HARD)
+        except TimeoutError:
+            LOGGER.warning("Policy engine didn't terminate in %d seconds. Killing it.", const.POLICY_ENGINE_GRACE_HARD)
+            self.process.kill()
+        self.process = None
```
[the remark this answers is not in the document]
arnaudsjs: Hmm. That's not mentioned in the docs. I added it anyway, because it doesn't cause any harm.

sanderr (changes requested):

src/inmanta/protocol/common.py
```diff
@@ -453,6 +451,43 @@ def __init__(
...
+            endpoint_id = f"{method_properties.operation} {method_properties.get_full_path()}"
+            endpoints[endpoint_id] = {
+                "client_types": method_properties.client_types,
+                "auth_label": auth_metadata.auth_label,
+                "read_only": auth_metadata.read_only,
+                "environment_param": auth_metadata.environment_param,
+            }
+        return {"endpoints": endpoints}
+
+    def is_machine_to_machine_endpoint(self) -> bool:
```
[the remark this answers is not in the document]
arnaudsjs: In the meantime I already renamed this to `is_machine_to_machine_endpoint`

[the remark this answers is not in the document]
sanderr: I never saw it under any other name

src/inmanta/protocol/rest/__init__.py
```diff
@@ -536,12 +542,21 @@ def authenticate(self, auth_enabled: bool) -> None:
...
+        return any(ct in {const.ClientType.agent.value, const.ClientType.compiler.value} for ct in client_types_token)
+
+    async def authorize_request(self, auth_enabled: bool, policy_engine_slice: Optional["PolicyEngineSlice"]) -> None:
+        """Validate whether the token is authorized to perform this request.
 
         :param auth_enabled: is authentication enabled?
+        :param policy_engine_slice: The policy-engine slice of this server or None if we are not running on the server.
         """
-        if not auth_enabled:
+        if not server_config.enforce_access_policy.get():
```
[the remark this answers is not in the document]
arnaudsjs: It is. The `auth_enabled` config option indicates that authentication is enabled. The `enforce_access_policy` config option indicates that policy-based authorization is enabled. The latter is only a temporary option. The `enforce_access_policy` option will be replaced by an option that indicates the authorization engine that should be used. That one can then be set to either "legacy" or "policy engine".

src/inmanta/server/services/policy_engine_service.py
arnaudsjs: This is implemented using sockets now.

src/inmanta/protocol/auth/policy_engine.py
```diff
@@ -23,132 +23,107 @@
...
-                opa_binary,
+            self._process = await asyncio.create_subprocess_exec(
+                opa_executable,
                 "run",
                 "--server",
                 "--addr",
-                await self.get_addr_policy_engine(),
+                f"unix://{self._socket_file}",
+                "--unix-socket-perm",
+                "700",
```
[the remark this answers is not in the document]
wouterdb: The one nasty problem this doesn't solve is that handlers can send commands to the OPA.

I don't know if that would allow reloading the policy?

But I'm afraid that is hard to defend against, as the agent and server run as the same user. 
(i.e. the agent could kill and replace the OPA and it can inherently access everything the server can) 
We would need to change the user of the executors to solve this... 

Previously, it was hard to bypass authentication, as the server can't write it's own config file... 

Only running OPA externally with TLS and no option for the server to modify the server would help.... 
Or SELinux 


But I guess this is out of scope for now.

src/inmanta/protocol/common.py
```diff
@@ -450,21 +463,42 @@ def __init__(
...
+                endpoint_id = f"{method_properties.operation} {method_properties.get_full_path()}"
+                endpoints[endpoint_id] = {
+                    "client_types": method_properties.client_types,
+                    "auth_label": auth_metadata.auth_label,
+                    "read_only": auth_metadata.read_only,
+                    "environment_param": auth_metadata.environment_param,
+                }
+        return {"endpoints": endpoints}
+
+    def is_human_interface(self) -> bool:
```
wouterdb, line 493:
```suggestion
    def is_external_interface(self) -> bool:
```

tests/conftest.py
```diff
@@ -689,9 +694,59 @@ def map_data_line(line):
...
+    """
+    Returns the path to the Open Policy Agent executable.
+    This method caches the executables to prevent slow setup times of the test suite.
+    """
+    cache_dir = os.path.abspath(os.path.join(__file__, "..", "data", "opa_executables.cache"))
+    os.makedirs(cache_dir, exist_ok=True)
+    cache_file = os.path.join(cache_dir, "opa-1.3.0")
+    if not os.path.exists(cache_file):
+        with open(cache_file, "wb") as fp:
+            req = requests.get("http://file.ii.inmanta.com/software/opa/opa-1.3.0", stream=True)
```
wouterdb, line 730: Do we have a plan to manage this version?

wouterdb (approved):

tests/conftest.py
arnaudsjs: I currently put the version of opa in the `inmanta/__init__.py`. But I don't think there is an automated way to update the version numbers. Unless we develop something ourselves.

src/inmanta/protocol/common.py
sanderr: I like it. Same concept, less weird.

mypy-baseline.txt
sanderr, line 1: Anything we should know in here?

src/inmanta/protocol/rest/__init__.py
```diff
@@ -665,11 +668,10 @@ async def _execute_call(
             # Authorization might need data from the request but we do not want to process it before we are sure the call
             # is authenticated.
             arguments = CallArguments(config, message, request_headers)
-            arguments.authenticate(server_config.server_enable_auth.get())
+            is_auth_enabled: bool = self.is_auth_enabled()
+            arguments.authenticate(auth_enabled=is_auth_enabled)
```
sanderr, line 672: Am I correct that `is_auth_enabled()` is equivalent to `self.get_policy_engine() is not None`? In that case, we could drop the method altogether at this level (and get rid of the `NotImplementedError`, and simply use `policy_engine is not None` here?

src/inmanta/protocol/rest/__init__.py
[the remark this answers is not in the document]
sanderr: I don't mean the config option, I mean the parameter. It is not used in the body of this method.

sanderr (changes requested):

src/inmanta/protocol/rest/__init__.py
arnaudsjs: It's not really the same. In the future we will have two modes for authorization. One legacy mode, which doesn't use the policy engine but just validates the token. An one where the policy engine is used. This implementation is not yet there because the PR was becoming too big in my opinion. I'll add that in a follow up PR. To summarize, it's possible to have authentication enabled, while the authorization doesn't use the policy engine.

sanderr (approved):

sanderr (approved):

PR #8946, opened 2025-04-16
Various reference and dataclass improvements

sanderr:
# Description

Current status: advanced draft. There are still quite some todo notes, but none of them should have a major impact on the PR, so no seems like a good time for a first review. You can ignore all TODO notes.

In scope for this first stage of review:
1. conceptual concerns
2. implementation remarks that are not affected by major conceptual concerns. I believe the implementation (unless covered by a TODO note) is already beyond draft quality, so implementation review is already welcome.

Parts to this PR below. 1 and 2 can be considered more or less independently. The rest should probably be after those two:
1. plugin boundary validation and custom `to_python`: `inmanta.references` + `inmanta.execute.runtime` + `inmanta.ast.*` minus `inmanta.ast.statements`
2. proxying and on-access error reporting: `inmanta.execute.proxy`
3. plugin boundary error reporting: `plugins.py`
4. `inmanta.resources` ~~-> this one is the most likely to need some further tweaks. There are still some aspects I need to consider here.~~
5. Everything else

Temporary files in this PR:
- `changes.tmp` contains a list of conceptual changes in this PR. I collected that to keep track of what is actually changing, with a change entry in mind. I still have to figure out how exactly, but it may be useful to the reviewer.
- `approach.tmp` contains an overview of the conceptual approach taken when it comes to references and plugin annotations. It is not final, and obviously not in the right place. I aim to convert this to documentation. In its current state, it may provide useful context to the reviewer.
- The tests (in combination with `changes.tmp`) may help provide motivation for certain code changes.

# Self Check:

- [ ] Changelog entry


changes.tmp
```diff
@@ -0,0 +1,19 @@
+- string format raises exception on reference
```
wouterdb, line 1: is this the before or after state?

src/inmanta/ast/type.py
```diff
@@ -111,6 +130,12 @@ def get_no_reference(self) -> "Type":
         """
         return self
 
+    def supports_references(self) -> bool:
+        """
+        Returns True iff this type accepts any sort of reference value.
+        """
+        return self.get_no_reference() is not self
```
wouterdb, line 137: I'm a bit suspicious about this implementation.

1. This is potentially called often, and this implementation is potentially slow (get_no_reference can be arbitrarily deep)
2. it relies on `is` instead of `eq` but there is nothing on `get_no_reference` that ensures that `is` means anything? 
3. it could be left ABC at this level an refined below by much simpler implementations

src/inmanta/references.py
```diff
@@ -534,6 +535,37 @@ def run(self, logger: "handler.LoggerABC") -> None:
         dict_path_expr.set_element(self.resource, value)
 
 
+@typing.runtime_checkable
+class MaybeReference(typing.Protocol):
+    """
+    DSL value that may represent a reference in the Python domain, while having a different value in the DSL domain.
+    """
+
+    def is_reference(self) -> Optional[Reference]:
```
wouterdb, line 544:
```suggestion
    def as_reference(self) -> Optional[Reference]:
```

from a method with is, I expect a boolean

src/inmanta/ast/type.py
```diff
@@ -215,21 +244,25 @@ def __init__(self, element_type: Type) -> None:
...
+            if ref._model_type.issubtype(self.element_type):
                 return True
 
-        raise TypingException(None, f"Invalid value {value} is not a subtype of {self.type_string()}")
+        raise TypingException(None, f"Invalid value: {value} is not a subtype of {self}")
 
-    def type_string(self) -> Optional[str]:
-        return self.element_type.type_string()
+    def has_custom_to_python(self) -> bool:
+        return self.is_dataclass
```
wouterdb, line 256: nice!

src/inmanta/execute/proxy.py
```diff
@@ -69,12 +115,31 @@ class DynamicProxy:
...
     """
 
-    def __init__(self, instance: "Instance") -> None:
+    def __init__(self, instance: "Instance", *, context: Optional[ProxyContext] = None) -> None:
+        """
+        :param instance: The object to proxy.
+        :param context: The context this object lives in.
+        """
         object.__setattr__(self, "__instance", instance)
+        object.__setattr__(self, "__context", context if context is not None else ProxyContext())
```
wouterdb, line 124: should be cache the default `ProxyContext()`? we may have a LOT of it?

wouterdb (commented):
First batch

src/inmanta/references.py
sanderr: Yeah, I struggled with the naming for this one. I'm not sure about `as`, because it's really about what this object already represents (is). But I'll let it sink in a bit, it's probably the best option anyway.

src/inmanta/execute/proxy.py
```diff
@@ -54,6 +59,47 @@ class DynamicUnwrapContext:
...
+        """
+        return dataclasses.replace(
+            self,
+            path=self.path + relative_path,
+            # we're proxying elements one level deeper than the current context => reset allow_reference_values to default behavior
+            allow_reference_values=None,
+        )
+
+    def should_allow_references(self) -> bool:
+        return self.allow_reference_values if self.allow_reference_values is not None else self.validated
```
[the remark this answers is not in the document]
sanderr: The assumption is that references are allowed iff the object has been validated. Unless if `allow_reference_values` has been overridden on this proxy object. I tried to put that in the docstring.

src/inmanta/execute/proxy.py
```diff
@@ -54,6 +59,47 @@ class DynamicUnwrapContext:
     type_resolver: TypeResolver
 
 
+# TODO: remember to check std module to ensure Jinja guards references. Add test!
+@dataclasses.dataclass(kw_only=True, frozen=True, slots=True)
+class ProxyContext:
```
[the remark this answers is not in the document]
sanderr: I'll consider all of this. No immediate response to any of it, except that I liked the idea of encapsulating this in a separate object, to separate concerns.

src/inmanta/execute/proxy.py
```diff
@@ -171,19 +247,55 @@ def return_value(cls, value: object) -> Union[None, str, tuple[object, ...], int
...
+        if isinstance(value, references.Reference) and not context.should_allow_references():
+            # TODO: see if we can use traceback.extract_stack() here to add a location to any exceptions
+
+            # Non-dataclass entities can not be explicit about reference support.
+            # The Python domain is a black box. We don't want to transparently pass unexpected values in there.
+            # => don't allow references in attributes. Can be explicitly allowed via allow_reference_attributes() wrapper
+            raise UndeclaredReference(
+                reference=value,
+                message=(
+                    "Encountered undeclared reference value during plugin execution. Plugins are only allowed to access"
```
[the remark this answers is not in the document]
sanderr: I see what you mean. I meant undeclared in the sense that it was not declared that it might pop up in this specific place. I think unexpected is fine, except that it doesn't imply as much that the user has control over it (by declaring it). But I guess that's covered by the rest of the message.

src/inmanta/resources.py
```diff
@@ -427,7 +427,10 @@ def map_field(
             elif hasattr(cls, "map") and field_name in cls.map:
                 value = cls.map[field_name](exporter, model_object)
             else:
-                value = getattr(model_object, field_name)
+                try:
+                    value = getattr(model_object, field_name)
+                except inmanta.ast.UndeclaredReference as e:
+                    value = e.reference
```
wouterdb, line 433: Why is this? I.e. why would this work/be correct when we abort at the first reference? The reference can be part of the attribute, it doesn't have to be the entire attribute

Moreover, why would we get this exception here? References are expected in resource serialization. It is more or less the whole point of them. We should not run this in the new mode we use for plugins.

src/inmanta/references.py
sanderr: I added it as a protocol because `ast.type` doesn't know and doesn't need to know (import loops) about instances, only that some values may represent references even if they aren't an instance of `Reference`.

I'm not too attached too it, and I do see there are two sides to the story. But overall I like that it decouples us from `Instance` here.

src/inmanta/resources.py
sanderr: Yeah, this module needs some more work, I'm aware of it. I'll make sure to take this comment into consideration.

> Moreover, why would we get this exception here? References are expected in resource serialization. It is more or less the whole point of them. We should not run this in the new mode we use for plugins.

That's sort of the point of this structure here. Though I think I may have made this change before I added the `allow_references()` method on proxy.

src/inmanta/references.py
wouterdb: I would add that as a comment

tests/data/modules_v2/refs/inmanta_plugins/refs/dc.py
```diff
@@ -0,0 +1,78 @@
...
+        super().__init__()
+
+    def resolve(self, logger: LoggerABC) -> NoRefsDataclass:
+        return NoRefsDataclassReference()
+
+    def __str__(self):
+        return "NoRefsDataclassReference"
+
+
+class AllRefsDataclassReferenceABC[D: AllRefsDataclass](Reference[D]):
```
wouterdb, line 42: Can the type extraction logic handle this? Doesn't this turn into any?

(after reading on: clearly not)

tests/test_references.py
```diff
@@ -97,9 +123,11 @@ def assert_id(ref_list: list[references.ReferenceModel], ref_type: str, ref_id:
     serialized = res_dict.popitem()[1].serialize()
 
     # validate that our UUID is stable
-    assert_id(serialized["references"], "refs::Bool", "ed26b59b-a567-392e-8626-f89d821532b7")
-    assert_id(serialized["references"], "refs::TestReference", "78d7ff5f-6309-3011-bfff-8068471d5761")
-    assert_id(serialized["references"], "core::AttributeReference", "81cc3241-2368-3b6d-8807-6755bd27b2fa")
+    assert_id(serialized["references"], "refs::Bool", "0d3b10e5-3f6c-32ae-8662-02e5d46a59c7")
+    # TODO: add other reference types?
+    # TODO: why did the untouched ones fail??
```
wouterdb, line 128: ?

tests/test_references.py
```diff
@@ -186,13 +215,303 @@ async def test_deploy_end_to_end(
...
+    # Primitives
+
+    # Scenario: plugin argument annotated as `object`
+    run_snippet(snippet="refs::plugins::takes_obj('hello')")
+    with pytest.raises(PluginTypeException, match="is a reference"):
+        run_snippet(snippet="refs::plugins::takes_obj(refs::create_string_reference('name'))")
+    ## accepts list with a reference in it and even allows access -> infeasible to check inside a black box
+    ## -> this is not a strong requirement. The assertion is here simply to ensure the behavior remains stable
+    run_snippet(snippet="refs::plugins::takes_obj(['hello', refs::create_string_reference('hello')])")
+    run_snippet(snippet="refs::plugins::iterates_obj(['hello', refs::create_string_reference('hello')])")
```
wouterdb, line 248: Isn't this supposed to fail at the moment where the iterator tries to wrap the reference?

tests/data/modules_v2/refs/inmanta_plugins/refs/plugins.py
```diff
@@ -0,0 +1,273 @@
...
+@plugin
+def iterates_obj_list(l: Sequence[object]) -> None:
+    for x in l:
+        ...
+    for i in range(len(l)):
+        l[i]
+
+
+@plugin
+def iterates_obj_list(l: Sequence[object]) -> None:
```
wouterdb, line 75: duplicate with the one above it

tests/test_references.py
```diff
@@ -186,13 +215,303 @@ async def test_deploy_end_to_end(
...
+    with pytest.raises(PluginTypeException, match=re.escape("Expected type: Union[Reference[bool],Reference[string]]")):
+        run_snippet(snippet="refs::plugins::takes_union_of_refs(true)")
+    # Scenario: plugin argument annotated as `Sequence[object]`
+    run_snippet(snippet="refs::plugins::iterates_obj_list(['h', 'e'])")
+    run_snippet(snippet="refs::plugins::iterates_obj_list(['h', 1])")
+    with pytest.raises(PluginTypeException, match="contains a reference"):
+        run_snippet(snippet="refs::plugins::iterates_obj_list(['h', refs::create_string_reference('name')])")
+    # Scenario: plugin argument annotated as `Sequence[str]`
+    run_snippet(snippet="refs::plugins::iterates_str_list(['h', 'e'])")
+    with pytest.raises(PluginTypeException, match=re.escape("Expected type: string[]")):
```
wouterdb, line 292: note: I got till here

src/inmanta/plugins.py
```diff
@@ -1089,6 +1120,7 @@ def deprecate_function(cls, replaced_by: Optional[str] = None) -> None:
         cls.deprecated = True
         cls.replaced_by = replaced_by
 
+    # TODO: no longer in use?
```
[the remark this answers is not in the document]
sanderr: Added that as a comment.

tests/test_references.py
sanderr: Ideally it would be, but this is a scenario that was deemed an acceptable "false negative", where we allow the reference to slip through the mazes. Because it's non-trivial to check, given the current state of the validation and proxy logic, and it's an unlikely scenario, that is still technically correct when it comes to the Python typing.

tests/test_references.py
sanderr: Was a note to self. No longer relevant so I removed them

src/inmanta/resources.py
```diff
@@ -239,7 +239,7 @@ def add_reference(self, path: str, reference: "references.Reference[references.P
         )
 
 
-def collect_references(value_reference_collector: ReferenceCollector | None, value: object, path: str) -> object:
+def collect_references(value_reference_collector: ReferenceCollector, value: object, path: str) -> object:
```
wouterdb, line 242: this the None gone, I wonder if this should be a method on the ReferenceCollector?

tests/test_references.py
wouterdb: Does this mean that this list is not wrapped in a DynamicProxy? 

i.e. it can leak compiler internals?

tests/test_references.py
```diff
@@ -186,13 +241,303 @@ async def test_deploy_end_to_end(
...
+    run_snippet("refs::plugins::read_entity_value(refs::dc::AllRefsDataclass(maybe_ref_value='Hello World!'))")
+    run_snippet("refs::plugins::read_entity_list_value(refs::ListContainer(value=['Hello', 'World!']))")
+    run_snippet("refs::plugins::read_entity_list_head(refs::ListContainer(value=['Hello', refs::create_string_reference('hello')]))")
+    run_snippet("refs::plugins::read_entity_dict_value(refs::DictContainer(value={'Hello': 'World!', 'mykey': '42'}))")
+    run_snippet(
+        "refs::plugins::read_entity_dict_mykey(refs::DictContainer(value={'Hello': refs::create_string_reference('hello'), 'mykey': '42'}))"
+    )
+    ## reference
+    ### in attribute
+    with raises_wrapped(UndeclaredReference, match="Encountered undeclared reference .* Encountered at instance.maybe_ref_value"):
```
wouterdb, line 357: Does  the error provide instructions on what to do when it is allowed?

tests/test_references.py
```diff
@@ -186,13 +241,303 @@ async def test_deploy_end_to_end(
...
+        run_snippet("refs::plugins::takes_no_refs_dataclass_ref(refs::dc::NoRefsDataclass())")
+
+    # Scenario: plugin annotated as <dataclass> | Reference[<dataclass>]
+    ## accepts either
+    run_snippet("refs::plugins::takes_no_refs_dataclass_or_ref(refs::dc::create_no_refs_dataclass_reference())")
+    run_snippet("refs::plugins::takes_no_refs_dataclass_or_ref(refs::dc::NoRefsDataclass())")
+    run_snippet("refs::plugins::takes_mixed_refs_dataclass_or_ref(refs::dc::create_mixed_refs_dataclass_reference('hello'))")
+    run_snippet("refs::plugins::takes_mixed_refs_dataclass_or_ref(refs::dc::MixedRefsDataclass(maybe_ref_value='hello'))")
+    ## rejects other dataclasses / references to other dataclasses, except for inheritance
+    run_snippet("refs::plugins::takes_no_refs_dataclass_or_ref(refs::dc::MixedRefsDataclass(maybe_ref_value='hello'))")
```
wouterdb, line 495: A would add a comment to indicate this is accepted because `MixedRefsDataclass` is a subclass of `NoRefsDataclass`

tests/test_references.py
```diff
@@ -186,13 +241,303 @@ async def test_deploy_end_to_end(
...
+    with pytest.raises(
+        PluginTypeException, match=re.escape("Expected type: Reference[refs::dc::MixedRefsDataclass] | refs::dc::MixedRefsDataclass")
+    ):
+        run_snippet("refs::plugins::takes_mixed_refs_dataclass_or_ref(refs::dc::AllRefsDataclass(maybe_ref_value='hello'))")
+    with pytest.raises(
+        PluginTypeException, match=re.escape("Expected type: Reference[refs::dc::MixedRefsDataclass] | refs::dc::MixedRefsDataclass")
+    ):
+        run_snippet("refs::plugins::takes_mixed_refs_dataclass_or_ref(refs::dc::create_all_refs_dataclass_reference('hello'))")
+
+    # Scenario: inheritance on return type
```
wouterdb, line 506: Got till here

tests/data/modules_v2/refs/inmanta_plugins/refs/dc.py
sanderr: I'm not sure which part exactly you were referring to. But I have the impression that you answered your own question, and that the test coverage suffices to give confidence?

tests/test_references.py
sanderr: It mentions `... | Reference[...]` type annotations for the boundary, and it points to the `plugins.allow_reference_values()` wrapper for other cases. Would you like me to add that part of the message to the assertions? I left it out mostly for verbosity reasons.

tests/test_references.py
sanderr: Isn't that what the comment above this line says? Or do you mean to be more explicit?

src/inmanta/references.py
sanderr: When reviewing, please keep in mind that this makes `inmanta.execute.util` quite a bit heavier, due to the `inmanta.references` import. But I still believe it appropriate. It then pulls in (recursively) `references`, `util`, `dict_path`. I think that's all conceptually allowed underneath `inmanta.execute.util`.

sanderr: Ah no, `references` also pulls in `inmanta.resources`, which is not appropriate under `inmanta.execute.util`, and causes all manner of import loops. I'm putting it back in `inmanta.references`, even if I find that it only half fits there.

src/inmanta/execute/proxy.py
sanderr: With the paths there's not much to cache. Memory impact of the proxy contexts seems to be negligible. Construction overhead may be a few percent, to be further investigated.

src/inmanta/execute/proxy.py
```diff
@@ -155,10 +221,20 @@ def recurse_dict_item(key_value: tuple[object, object]) -> tuple[object, object]
...
+        *,
+        context: Optional[ProxyContext] = None,  # optional for backwards compatibility
+    ) -> Union[None, str, tuple[object, ...], int, float, bool, "DynamicProxy"]:
         """
         Converts a value from the internal domain to the plugin domain.
+
+        :param context: The context in which the given object lives. If None, assumes that the value has already passed
+            validation. When None, the object's string representation, rather than its name is used for error reporting.
         """
+        context = context if context is not None else ProxyContext(path=f"<{value}>")
```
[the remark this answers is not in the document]
sanderr: Good suggestion, done

src/inmanta/execute/proxy.py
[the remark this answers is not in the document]
sanderr: > I don't know if this is significant for performance, but this doubles the nr of objects allocated for DynamicProxy

I have the impression that it's not, probably because it's slotted (which `DynamicProxy` isn't btw)

> The entire idea of path tracking is somewhat attractive, but it may also be expensive.

Agreed. Though I get the impression that the overhead is sufficiently limited that it is worth it. Especially considering that the errors could become very very difficult to debug otherwise.

>  We may also want to store the path as an array of strings instead of a string, to avoid string operations that may be unnecessary. (Again, don't know what is fastest)

Tried it with `timeit` on a simple concatenation vs `l = [*l, x]`. Concatenation won out. There may be other alternatives, like keeping track of the chain of contexts. But that has drawbacks of its own, and I don't have the impression that this path construction is a major contributor to the minor overhead that this PR seems to have.

src/inmanta/execute/proxy.py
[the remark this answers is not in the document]
sanderr: Done

tests/test_references.py
```diff
@@ -185,6 +245,303 @@ async def test_deploy_end_to_end(
...
+        run_snippet("refs::plugins::returns_entity_list(refs::ListContainer(value=['Hello', refs::create_string_reference('hello')]))")
+    ## allowed if plugin annotates reference in return type
+    run_snippet(
+        "refs::plugins::returns_entity_ref_list(refs::ListContainer(value=['Hello', refs::create_string_reference('hello')]))"
+    )
+
+    # Scenario: allow_reference_values() called on non-proxy list (e.g. list inside dataclass)
+    with pytest.raises(
+        ExternalException,
+        match="should only be called on inmanta instances, lists or dicts. Python lists and dicts do not need this wrapper",
```
wouterdb, line 541: Why doesn't it just work on `any`? i.e. what does the increased restriction on the input gain us?

src/inmanta/ast/type.py
wouterdb: 1.c type constraints nest, so it can be deeper  

I think you make valid points, to 1 and 2, but I still think this is complex and brittle compared to the alternative?

sanderr: 1c. True, but those have a trivial implementation as well.

> I still think this is complex and brittle compared to the alternative?

I'm torn to be honest. What I don't like about the alternative is that every subtype has to think about it, and to consistently implement both methods. With this approach they only have to implement one. If they implement it according to the contract, the other one just works. But, if they violate the contract, then yes, it's brittle. Complex, I don't see it.

I tried to implement it when you first suggested it, because I thought it made sense. I just realized that the alternative isn't all that simple either, because there are many composed types, not just a few. So the diff was quite large, and not trivial to check if it was correct. So overall I'm still leaning towards keeping it as is, with the docstring change I proposed in 2.

I think that's all the context I have. If you still prefer the alternative, I'll make the change without further arguments.

tests/test_references.py
sanderr: I'll expand on it

src/inmanta/execute/proxy.py
```diff
@@ -120,23 +120,26 @@ def __init__(self, instance: "Instance", *, context: Optional[ProxyContext] = No
         :param instance: The object to proxy.
         :param context: The context this object lives in.
         """
+        # str() is expensive. And also not helpful for e.g. dicts.
```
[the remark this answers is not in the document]
sanderr: context is slotted. I tried with proxy, and it didn't seem to do much for performance (I now realize I only made top-level slotted, not the children, I can give that another shot). Drawback is that the dunder attributes didn't seem to play nice with slots for some reason.

src/inmanta/execute/proxy.py
```diff
@@ -120,23 +120,26 @@ def __init__(self, instance: "Instance", *, context: Optional[ProxyContext] = No
...
+        # => Fall back to object.__repr__ for externally created objects (internal calls should all pass a context)
+        context = context if context is not None else ProxyContext(path=object.__repr__(instance))
```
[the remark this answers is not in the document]
sanderr: I considered it, but I already heard @bartv complaining about it in my mind. Not that this is much better, but at least it doesn't read like "we don't know what's going on".

tests/compiler/test_dataclasses.py
```diff
@@ -110,6 +110,18 @@ def test_dataclass_positive(snippetcompiler):
 end
 dataclasses::is_odd_string(hs.thestring)
 
+# a dataclass instance that is still wating for its values when the plugin is called
```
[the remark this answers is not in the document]
sanderr: Yeah, this is a subtle one. But it is also one that is actually broken on master. The thing with plugins is that they have to be able to reschedule if it tries to access model values that aren't available yet. Trying to construct a dataclass instance when not all values are there yet is basically the same thing. Yet on master we caught the exception and wrapped it instead of letting it trickle to where we reschedule plugins.

I'll elaborate on the comment.

tests/data/modules_v2/refs/inmanta_plugins/refs/__init__.py
```diff
@@ -26,6 +26,24 @@ def __str__(self) -> str:
...
+    def __init__(self, name: str | Reference[str]) -> None:
+        """
+        :param name: The name of the environment variable.
+        """
+        super().__init__()
+        self.name = name
+
+    def resolve(self, logger: LoggerABC) -> int:
+        """Resolve the reference"""
+        return os.getenv(self.resolve_other(self.name, logger)) == "true"
```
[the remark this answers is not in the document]
sanderr: Uh, well, this is just not entirely correct. It's just that we don't ever resolve the reference, we only use it for typing purposes so it's not so bad. I'll just make it return a static int.

tests/test_references.py
```diff
@@ -97,9 +123,9 @@ def assert_id(ref_list: list[references.ReferenceModel], ref_type: str, ref_id:
     serialized = res_dict.popitem()[1].serialize()
 
     # validate that our UUID is stable
-    assert_id(serialized["references"], "refs::Bool", "ed26b59b-a567-392e-8626-f89d821532b7")
-    assert_id(serialized["references"], "refs::TestReference", "78d7ff5f-6309-3011-bfff-8068471d5761")
-    assert_id(serialized["references"], "core::AttributeReference", "81cc3241-2368-3b6d-8807-6755bd27b2fa")
+    assert_id(serialized["references"], "refs::Bool", "0d3b10e5-3f6c-32ae-8662-02e5d46a59c7")
```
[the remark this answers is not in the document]
sanderr: It's complicated, and I haven't been involved in it too deeply. But off the top of my head: we need the hash of these things to be stable for serialization purposes. In this case, I made so many changes that they changed, but they should remain stable as long as their contents are stable.

tests/test_references.py
```diff
@@ -185,6 +245,303 @@ async def test_deploy_end_to_end(
...
+    run_snippet(snippet="refs::plugins::takes_str('hello')")
+    with pytest.raises(PluginTypeException, match="is a reference"):
+        run_snippet(snippet="refs::plugins::takes_str(refs::create_string_reference('name'))")
+    # Scenario: plugin argument annotated as `str | Reference[str]`
+    run_snippet(snippet="refs::plugins::takes_str_ref('hello')")
+    run_snippet(snippet="refs::plugins::takes_str_ref(refs::create_string_reference('name'))")
+    with pytest.raises(PluginTypeException, match=re.escape("Expected type: Reference[string] | string")):
+        # takes a str ref, not a bool ref
+        run_snippet(snippet="refs::plugins::takes_str_ref(refs::create_bool_reference('name'))")
+    # Scenario: plugin argument annotated with complex union including Reference[str | None]
```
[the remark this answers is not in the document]
sanderr: A reference that may resolve to `None`. e.g. an environment variable reference if the environment variable is not necessarily expected to exist.

Hugo-Inmanta (approved):

sanderr:
(follow-up ticket in #9262)

PR #8949, opened 2025-04-17
Remove unused methods that touch the resource table

### LINKED ISSUE #8937: Remove dependency on resource table to fetch the latest version
Check if we are using these methods, if not, delete them. If we are, remove the dependency on resource table to fetch the latest version. Use the model table instead
- get_resources_report
- get_resource_type_count_for_latest_version

jptrindade:
# Description

Deprecate these methods in iso8: https://github.com/inmanta/inmanta-core/pull/8950
Cleanup inmanta-lsm: https://github.com/inmanta/inmanta-lsm/pull/2082

closes #8937 

# Self Check:

- [x] Changelog entry


tests/test_server.py
```diff
@@ -560,9 +523,8 @@ async def test_clear_environment(client, server, clienthelper, environment):
     )
     assert result.code == 200
 
-    result = await client.get_environment(id=environment, versions=10)
+    result = await client.environment_get(id=environment)
     assert result.code == 200
-    assert len(result.result["environment"]["versions"]) == 1
```
sanderr, line 565: I believe that this change and the one below weaken the core of the test case.

tests/test_data.py
```diff
@@ -1534,131 +1534,6 @@ async def assert_expected_count(expected_report: dict[str, int]):
     )  # 1 NullResource resource and 1 Dummy resource in model v2
 
 
-async def test_resources_report(init_dataclasses_and_load_schema):
```
sanderr, line 1537: What's the motivation for deleting this?

sanderr (changes requested):

tests/test_data.py
jptrindade: I deleted `get_resources_report` which was only being used by the v1 endpoint

sanderr: Oh I see, this was a test for the helper method.

tests/test_server.py
```diff
@@ -406,43 +406,6 @@ async def test_resource_action_update(server_multi, client_multi, environment_mu
     assert result.result["metadata"]["deploy_summary"]["by_state"]["deployed"] == 2
 
 
-async def test_get_environment(client, clienthelper, server, environment):
```
Hugo-Inmanta, line 409: I assume we are already testing `environment_get` somewhere else, correct ?

jptrindade: Yes, on test_project.py

tests/test_server.py
jptrindade: Added some assertions on the version count

Hugo-Inmanta (approved):

sanderr (approved):

PR #8950, opened 2025-04-17
Added deprecation warning to v1 endpoint to fetch environment

jptrindade:
# Description

Also deprecates Resource.get_resources_report

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/deprecate-v1-environment-get.yml
+++ b/changelogs/unreleased/deprecate-v1-environment-get.yml
@@ -3,4 +3,4 @@ description: Deprecated the v1 endpoint to fetch an environment in favor of the 
 change-type: patch
 destination-branches: [iso8]
 sections:
-  deprecation-note: {{description}}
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/deprecate-v1-environment-get.yml
```diff
@@ -0,0 +1,6 @@
+description: Deprecated the v1 endpoint to fetch an environment in favor of the v2 endpoint.
+  Also deprecated Resource.get_resources_report as it was only used by this endpoint.
+change-type: patch
+destination-branches: [iso8]
+sections:
+  deprecation-note: {{description}}
```
sanderr, line 6: If you use `{{description}}`, you should make sure to only include user-oriented information in the description. The part of `Resource.get_resources_report` is not relevant for the user.

Also, I think you may need to quote it, though I'm not sure.

sanderr (approved):

Hugo-Inmanta (approved):

PR #8961, opened 2025-04-22
Track deploying state in ResourcePersistentState table

jptrindade:
# Description

* Short description here *

# Self Check:

- [x] Changelog entry


tests/db/migration_tests/dumps/v202504040.sql
jptrindade, line 1: I was missing this dump

jptrindade:
Am I correct to assume that we also want to add `DEPLOYING` to the `Compliance` ?
Also, I'm having some difficulties testing that this works since I only have a small window that it is deploying (if it is actually being set as deploying). Do you have any tests I should take a look at? I remembered the scheduler mock but it has its own implementation of `send_inprogress` and `send_deploy_done` which defeats the purpose of this test

src/inmanta/deploy/state.py
```diff
@@ -62,12 +62,14 @@ class Compliance(StrEnum):
...
     UNDEFINED: The resource status is undefined, because it has an unknown attribute. i.e. undefined status <=> undefined
         intent.
+    DEPLOYING: The resource is currently in the middle of a deploy attempt.
     """
 
     COMPLIANT = enum.auto()
     HAS_UPDATE = enum.auto()
     NON_COMPLIANT = enum.auto()
     UNDEFINED = enum.auto()
+    DEPLOYING = enum.auto()
```
wouterdb, line 72: I don't know if this is intended to go into this state machine? @sanderr may know

src/inmanta/deploy/persistence.py
```diff
@@ -163,8 +163,12 @@ async def send_in_progress(self, action_id: UUID, resource_id: Id) -> None:
                 except UniqueViolationError:
                     raise ValueError(f"A resource action with id {action_id} already exists.")
 
-                # FIXME: we may want to have this in the RPS table instead of Resource table, at some point
+                # FIXME: At some point, we will only need this on the RPS table
                 await resource.update_fields(connection=connection, status=const.ResourceState.deploying)
+                await resource.update_persistent_state(
+                    is_deploying=False,
```
wouterdb, line 169:
```suggestion
                    is_deploying=True,
```

wouterdb: This also suggests we are not testing this. I would try to add some assertions to existing tests (that may already test consistency)

wouterdb (changes requested):

src/inmanta/deploy/persistence.py
jptrindade: :facepalm: Maybe this is why on my makeshift tests I've never got it to work

src/inmanta/db/versions/v202504220.py
```diff
@@ -0,0 +1,29 @@
...
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    """
+    Rename some rps columns and values
+    """
+    schema = """
+        ALTER TABLE public.resource_persistent_state ADD COLUMN is_deploying BOOLEAN DEFAULT FALSE;
```
sanderr, line 27: We probably need an index involving this field, but I'm not sure if now's the time (i.e. if we already know enough about how it will be used). But I think we can't go wrong with simply an index on `environment, is_deploying`. @wouterdb wdyt?

tests/server/test_resourceservice.py
```diff
@@ -617,6 +617,7 @@ async def test_log_deploy_start(server, client, environment, clienthelper, agent
         {"name": "file1", "id": rvid_r1_v1, "requires": [], "purged": False, "send_event": False},
     ]
     await clienthelper.put_version_simple(resources, version)
+    await data.ResourcePersistentState.populate_for_version(environment=uuid.UUID(environment), model_version=version)
```
sanderr, line 620: Can you elaborate on these changes? It seems sensible to me, but I don't immediately see why it's suddenly required while it wasn't before. Is this because we now write to the RPS table when we start a deploy?

tests/deploy/e2e/test_autostarted.py
```diff
@@ -1344,3 +1344,51 @@ async def wait_for_down_status() -> bool:
...
+    file_to_remove = tmp_path / "file"
+    file_to_remove.touch()
+
+    config.Config.set("config", "environment", environment)
+
+    model = f"""
+import minimalwaitingmodule
+minimalwaitingmodule::WaitForFileRemoval(name="test_sleep", agent="agent1", path="{file_to_remove}")
+"""
+    snippetcompiler.setup_for_snippet(model, ministd=True, index_url="https://pypi.org/simple")
```
sanderr, line 1372:
```suggestion
    snippetcompiler.setup_for_snippet(model, ministd=True)
```
I don't think this is required. Do we do it anywhere else?

sanderr (changes requested):
Looks good to me. Two minor comments

tests/deploy/e2e/test_autostarted.py
jptrindade: Yes, I copied it from the test case above

tests/server/test_resourceservice.py
jptrindade: Yes, we now write to the RPS table when we start a deploy. I don't know if I should be here or right before we write the state to the rps table

sanderr: Here looks fine to me

tests/deploy/e2e/test_autostarted.py
sanderr: I see. I still don't think it's required, and I'd prefer to see it go (unless I'm mistaken). But I won't hold back the PR for this.

src/inmanta/deploy/state.py
jptrindade: After discussion I removed this

tests/deploy/e2e/test_autostarted.py
jptrindade: Removing the index_url causes this call `await wait_for_resources_in_state(client, uuid.UUID(environment), nr_of_resources=1, state=const.ResourceState.deploying)` to always timeout.

sanderr: Ah I see, we need it on the agent side as well. Sorry for the noise.

wouterdb (approved):

src/inmanta/db/versions/v202504220.py
wouterdb: Why do we need an index?

PR #8981, opened 2025-04-25
added equals method

wouterdb:
# Description

Ad equals method to references

# Self Check:

- [x] Changelog entry
- [X] Correct, in line with design


wouterdb:
Turns out this doesn't work: dataclasses make it fail

arnaudsjs (approved):

src/inmanta/references.py
```diff
@@ -314,6 +314,23 @@ def serialize_arguments(self) -> Tuple[uuid.UUID, list[ArgumentTypes]]:
...
 
+    def __eq__(self, other: object) -> bool:
+        if type(self) is not type(other):
+            return False
+
+        my_args = self.arguments
+        assert isinstance(other, ReferenceLike)  # mypy can't figure out the check above
+        other_args = other.arguments
+
+        if my_args.keys() != other_args.keys():
```
bartv, line 325: Why not:

```python
return self.arguments == other.arguments
```

From the python docs:

> Dictionaries compare equal if and only if they have the same (key, value) pairs (regardless of ordering). Order comparisons (‘<’, ‘<=’, ‘>=’, ‘>’) raise [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError).

https://docs.python.org/3/library/stdtypes.html#mapping-types-dict

PR #8984, opened 2025-04-25
Change queries to fetch status from rps table only

jptrindade:
# Description

Change get_current_resource_state to fetch status from rps table only.
First part of an effort to change queries to fetch the status from the RPS table

# Self Check:

- [ ] Changelog entry


wouterdb (approved):

src/inmanta/data/__init__.py
```diff
@@ -4855,44 +4855,37 @@ async def get_resource_states_latest_version(
...
-                        -- The deploying, undefined and skipped_for_undefined states are not tracked in the
-                        -- resource_persistent_state table.
-                        THEN r.status::text
-                    WHEN rps.last_deployed_attribute_hash != r.attribute_hash
-                        -- The hash changed since the last deploy -> new desired state
-                        THEN r.status::text
-                        -- No override required, use last known state from actual deployment
-                        ELSE rps.last_non_deploying_status::text
+                    WHEN rps.is_orphan = TRUE
+                        THEN 'orphaned'
```
sanderr, line 4867: I *think* it would be more direct and efficient to replace this with a `AND NOT rps.is_orphan` in the `WHERE` clause.
Efficiency is probably not too much of a concern here, but still, good practice.

src/inmanta/data/__init__.py
```diff
@@ -4855,44 +4855,37 @@ async def get_resource_states_latest_version(
...
+            FROM resource_persistent_state AS rps
+            WHERE rps.environment=$1 AND rps.resource_id=$2
             """
         results = await cls.select_query(query, [env, rid], no_obj=True)
         if not results:
             return None
         assert len(results) == 1
-        return const.ResourceState(results[0]["status"])
+        status = results[0]["status"]
+        return const.ResourceState(status) if status != "orphaned" else None
```
sanderr, line 4888: If you take the direct route, this change becomes redundant because it's caught by the `if not results` check.

sanderr (commented):
FYI, this is the type of change that requires thorough testing:
- before merging, run the CI in non-fast mode
- after merging, and after master succeeded (building the wheels), run some related jobs like inmanta-lsm, lsm module, ...

sanderr (approved):

PR #8991, opened 2025-04-28
Test file names db migration scripts

arnaudsjs:
# Description

Added test to validate the filenames of the database migration scripts.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -25,8 +25,10 @@ import json
 import logging
 import math
 import os
+import pathlib
 import queue
 import random
+import re
 import shutil
 import uuid
 from collections import abc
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -1087,3 +1089,23 @@ def _is_compile_finished() -> bool:
...
+    for path in versions_folder.iterdir():
+        file_name = path.name
+        if not file_name.endswith(".py"):
+            continue
+        if file_name == "__init__.py":
+            continue
+        if file_name == "v1.py":
+            v1_found = True
+            continue
+        if not re.fullmatch(r"v([0-9]{8}[0-9]+)\.py", file_name):
```
sanderr, line 1109:
```suggestion
        if not re.fullmatch(r"v([0-9]{8}[0-9])\.py", file_name):
```
I don't think we currently support more than 10 scripts on the same day. Iirc we use numeric sort on them so adding an extra digit would mess up the ordering.

On a less serious note: I wonder how this will break first: when we reach the year 10k, or if we'll have switched to a new calendar before then.

sanderr (changes requested):

sanderr (approved):
Right, that makes more sense than what I proposed.

PR #9011, opened 2025-04-30
Issue/8962 Fix environment_auth endpoint returns 400

### LINKED ISSUE #8962: database based auth does not allow tokens to be generated
When database based authentication is enabled it is not possible to generate a token. The following error is reported:

```
2025-04-22 11:56:29,102 inmanta.protocol.rest.serverWARNING Received an exception with status code 400 and message {'message': 'Invalid request: Authentication is disabled, generating a token is not allowed'}
```

http://localhost:8888/api/v2/environment_auth

A workaround is to start a python interpreter:

- Load config, and load the file with the signing config in
- Use inmanta.protocol.auth.encode_token

arnaudsjs:
# Description

Fix bug where the POST /api/v2/environment_auth endpoint returns a 400 if the server.auth=true config option was set using an environment variable.

closes #8962 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/config.py
+++ b/src/inmanta/config.py
@@ -201,7 +201,7 @@ class Config:
         cfg: ConfigParser = cls.get_instance()
         val: Optional[str] = _get_from_env(section, name)
         if val is not None:
-            LOGGER.debug(f"Setting {section}:{name} was set using an environment variable")
+            LOGGER.debug("Setting %s:%s was set using an environment variable", section, name)
             return val
         # Typing of this method in the sdk is not entirely accurate
         # It just returns the fallback, whatever its type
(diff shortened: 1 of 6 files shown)


src/inmanta/config.py
```diff
@@ -213,12 +213,14 @@ def is_set(cls, section: str, name: str) -> bool:
...
-    def getboolean(cls, section: str, name: str, default_value: Optional[bool] = None) -> bool:
+    def getboolean(cls, section: str, name: str, default_value: bool) -> bool:
         """
         Return a boolean from the configuration
         """
-        cls.validate_option_request(section, name, default_value)
-        return cls.get_instance().getboolean(section, name, fallback=default_value)
+        val = cls.get(section, name, default_value)
+        if val is None:
+            raise ValueError(f"Expected boolean value. Found: {val}")
```
sanderr, line 222: Not sure if `ValueError` is appropriate here. Afaik it refers to the argument values.

arnaudsjs: I see. But this keeps the behavior consistent with the old behavior.

arnaudsjs:
I created [this PR](https://github.com/inmanta/inmanta-core/pull/9013) for iso7.

sanderr (approved):

PR #9031, opened 2025-05-05
Change some queries to fetch the status from the rps table

### LINKED ISSUE #9080: Change queries to fetch the status from the RPS table instead of the Resource table

jptrindade:
# Description

Part1 of this PR was already merged but was reverted due to an error on the inmanta-lsm test suite. [I fixed the error in this PR.](https://github.com/inmanta/inmanta-lsm/pull/2110)

In this first pass, I wanted to address the queries that explicitly fetched the status from the resource table. Meaning that the status we return was directly tied to the query. These were the affected queries:

1. Get_status_for (Removed, not being used)
2. Get_resource_ids_with_status (Removed. Only used on mark_deployed.)
3. Get_undeployable (Removed, it was not being used)
4. Get_resource_states_latest_version (used on the trigger_get_status endpoint to check if the state of the resources in the DB corresponds to the internal state of the scheduler and returns the discrepancies
5. Get_current_resource_state (part of the stable api, used to return the state of a resource in the latest version of the configuration model or None if it is not present)
6. Get_resource_deploy_summary (used on resource_list to output the deploy summary)
7. ResourceView (used on resource_list)
8. get_metric_value (used for the environment metrics)

On a next pass I'll address the queries where the status is incidentally returned, i.e. we are returning SELECT * FROM resource . In these instances I'll first try to assert how it is being used, and if we are using that method to check the status of a resource or are we more interested in its other fields

[Passed the full test suite](https://jenkins.inmanta.com/job/core/job/inmanta-core/job/change-more-queries-to-fecth-status-from-rps-table/15/)

closes #9080 9080

# Self Check:

- [x] Changelog entry


src/inmanta/const.py
```diff
@@ -24,6 +24,30 @@
...
+            THEN 'undefined'
+        WHEN rps.blocked = 'BLOCKED'
+            THEN 'skipped_for_undefined'
+        WHEN rps.current_intent_attribute_hash <> rps.last_deployed_attribute_hash
+            THEN 'available'
+        ELSE
+            rps.last_non_deploying_status::text
+    END
+) AS status,
+rps.current_intent_attribute_hash <> rps.last_deployed_attribute_hash AS has_update
```
jptrindade, line 48: Not sure if this should be included here, on a separate const or just on a case by case basis directly in the query

src/inmanta/data/__init__.py
```diff
@@ -5572,6 +5443,9 @@ async def update_persistent_state(
         connection: Optional[asyncpg.connection.Connection] = None,
         # TODO[#8541]: accept state.ResourceState and write blocked status as well
         last_deploy_result: Optional[state.DeployResult] = None,
+        blocked: Optional[str] = None,
+        is_undefined: Optional[bool] = None,
+        is_orphan: Optional[bool] = None,
```
jptrindade, line 5448: Does it make sense to also include these fields here? I added them because it was useful for a test case, but I could also do this in a separate method

src/inmanta/deploy/persistence.py
```diff
@@ -278,14 +271,15 @@ def error_and_log(message: str, **context: Any) -> None:
...
+                # TODO: do we need to set blocked, is_undefined, is_orphan here?
                 await resource.update_persistent_state(
                     is_deploying=False,
                     last_deploy=finished,
                     last_deployed_version=resource_id_parsed.version,
                     last_deployed_attribute_hash=resource.attribute_hash,
                     last_non_deploying_status=const.NonDeployingResourceState(status),
                     last_deploy_result=state.last_deploy_result if state is not None else None,
-                    **extra_datetime_fields,
+                    **extra_datetime_fields,  # type: ignore[arg-type]
```
jptrindade, line 282: mypy was complaining because I added the `blocked`, `is_undefined`, `is_orphan` arguments to `update_persistent_state`

tests/deploy/e2e/test_autostarted.py
```diff
@@ -1332,14 +1332,26 @@ async def wait_for_down_status() -> bool:
...
+        result = await client.get_agents(environment)
+        assert result.code == 200
+        actual_data = result.result["data"]
+        if len(actual_data) != 1:
+            return False
+        return actual_data[0]["status"] == "up"
+
+    # Wait for the scheduler to be up and set the status to available
+    await retry_limited(wait_for_up_status, timeout=5)
+
```
jptrindade, line 1354: The previous method did not work because `release_version` did not wait for paused agents to be up

jptrindade: *this method is still not working, there is a race condition between the scheduler restarting, it setting the state to available and it scheduling the resource to be deployed

src/inmanta/data/__init__.py
```diff
@@ -5382,22 +5272,10 @@ async def get_versioned_resource_details(
...
-        ) as status
-        FROM {cls.table_name()} as r
-            JOIN resource_persistent_state rps ON r.resource_id = rps.resource_id and r.environment = rps.environment
-                WHERE r.environment=$1 AND r.model=(SELECT MAX(cm.version)
-                                                    FROM public.configurationmodel AS cm
-                                                    WHERE cm.environment=$1 AND cm.released=TRUE)
+        SELECT rps.resource_id as resource_id,
+        {const.RESOURCE_STATUS_QUERY}
+        FROM resource_persistent_state as rps
+            WHERE rps.environment=$1 AND NOT rps.is_orphan
```
arnaudsjs, line 5278:
```suggestion
        SELECT rps.resource_id as resource_id,
            {const.RESOURCE_STATUS_QUERY}
        FROM resource_persistent_state as rps
        WHERE rps.environment=$1 AND NOT rps.is_orphan
```

src/inmanta/deploy/persistence.py
arnaudsjs: Let's not add the ignore. We need to sync the type violation in to the baseline file so that we can keep track of it.

src/inmanta/server/services/resourceservice.py
```diff
@@ -251,7 +251,6 @@ async def mark_deployed(
         version: int,
         filter: Callable[[ResourceIdStr], bool] = lambda x: True,
         connection: ConnectionMaybeInTransaction = ConnectionNotInTransaction(),
-        only_update_from_states: Optional[Sequence[const.ResourceState]] = None,
```
arnaudsjs, line 254: This was no longer used?

arnaudsjs (changes requested):

src/inmanta/server/services/resourceservice.py
jptrindade: There was only one call to `mark_deployed` on core and it did not use that argument. By the end of this epic we are expecting to completely remove `mark_deployed` from the codebase

src/inmanta/data/__init__.py
```diff
@@ -4507,21 +4507,31 @@ def table_name(cls) -> str:
...
+        version: int,
         connection: Optional[Connection] = None,
     ) -> None:
         """
-        Set the is_orphan column to True on all given resources.
+        Marks all persisted resources that do not exist in the given version as orphans.
         """
         query = f"""
-            UPDATE {cls.table_name()}
+            UPDATE {cls.table_name()} AS rps
```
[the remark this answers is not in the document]
sanderr: Perhaps even `environment, resource_id, is_orphan`? I think the join might become a sequential scan otherwise?

It's been a while so don't take my word for it.

src/inmanta/data/__init__.py
```diff
@@ -5267,38 +5183,12 @@ def copy_for_partial_compile(self, new_version: int) -> "Resource":
     async def get_released_resource_details(
```
sanderr, line 5183: I think your changes to this method are OK, but I think the method deserves a critical review if it ends up staying. Could you add a note / ticket for the final stages of the epic? We can still decide to postpone it then, but let's at least make sure that we consider it at that point.

src/inmanta/data/__init__.py
sanderr: I'd prefer to leave them out. The scheduler purposely divides the table into intent-related columns and deploy-related ones, and is very careful about when to write to which. It might have been more explicitly documented, but that aside, I think we must be very careful to keep this separation in the helper methods as well.

sanderr (changes requested):
> # TODO: do we need to set blocked, is_undefined, is_orphan here?

We *must* not

src/inmanta/data/__init__.py
```diff
@@ -4794,99 +4804,41 @@ def convert_or_ignore(rvid: ResourceVersionIdStr) -> resources.Id | None:
...
-            FROM resource AS r
-                 INNER JOIN resource_persistent_state AS rps ON r.environment=rps.environment AND r.resource_id=rps.resource_id
-                 INNER JOIN configurationmodel AS c ON c.environment=r.environment AND c.version=r.model
-            WHERE r.environment=$1 AND r.model = (SELECT version FROM latest_released_version)
+                c.version,
+                rps.resource_id,
+                {const.RESOURCE_STATUS_QUERY}
+            FROM resource_persistent_state AS rps
+                 INNER JOIN configurationmodel AS c ON c.environment=rps.environment
+            WHERE c.environment=$1 AND c.version = (SELECT version FROM latest_released_version) AND NOT rps.is_orphan
```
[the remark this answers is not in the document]
jptrindade: It appears to only be used by the `trigger_get_status` api endpoint. Which calculates discrepancies between the in-memory representation of the and their state in the database.

src/inmanta/data/__init__.py
jptrindade: https://github.com/inmanta/inmanta-core/issues/9066

src/inmanta/const.py
```diff
@@ -24,6 +24,30 @@
...
+            THEN 'undefined'
+        WHEN rps.blocked = 'BLOCKED'
+            THEN 'skipped_for_undefined'
+        WHEN rps.current_intent_attribute_hash <> rps.last_deployed_attribute_hash
+            THEN 'available'
+        ELSE
+            rps.last_non_deploying_status::text
+    END
+) AS status,
+rps.current_intent_attribute_hash <> rps.last_deployed_attribute_hash OR rps.last_deployed_attribute_hash IS NULL AS has_update
```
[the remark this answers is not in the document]
jptrindade: I'll remove this whole `has_update` part and create a separate ticket for it. It might be useful to discuss where do we want to see this. https://github.com/inmanta/inmanta-core/issues/9067

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
jptrindade: Why do you feel that the index might be a bad idea? Do you mean specifically on `environment, is_orphan` or on `environment, resource_id, is_orphan` as well?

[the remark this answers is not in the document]
arnaudsjs: No. I am pretty sure that an index is required. But I made a proposal without any testing and I think that was a bad idea. So I shouldn't have made the proposal at all and let you do the testing instead.

sanderr (changes requested):

sanderr (approved):

PR #9032, opened 2025-05-05
Support bearer token in openapi documentation

arnaudsjs:
# Description

Added support for bearer token based auth in openapi documentation. As such, it's possible to provide a bearer token in the "try it out" feature of the openapi documentation.

I didn't see an easy way to test this, so I tested it manually.

Part of https://github.com/inmanta/inmanta-lsm/issues/2096

# Self Check:

- [x] Changelog entry


src/inmanta/protocol/openapi/converter.py
```diff
@@ -156,11 +175,34 @@ class OpenApiTypeConverter:
...
+                case AuthMethod.database.name:
+                    return {
+                        "bearerAuth": SecurityScheme(
+                            type="http",
+                            scheme="bearer",
+                            bearerFormat="JWT",
+                        )
+                    }
+                case AuthMethod.oidc.name:
+                    # The oidc security scheme is not supported.
```
arnaudsjs, line 195: I will create a ticket for this

jptrindade (approved):

src/inmanta/protocol/openapi/model.py
```diff
@@ -263,13 +263,24 @@ class PathItem(BaseModel):
     parameters: Optional[list[Union[Parameter, Reference]]] = None
 
 
+type SecuritySchemeName = str
+type Scope = str
```
sanderr, line 267: Nice. Helps readability imo.

src/inmanta/protocol/openapi/converter.py
arnaudsjs: https://github.com/inmanta/inmanta-core/issues/9033

arnaudsjs:
@sanderr I realized that my implementation was a bit too restrictive. No matter what the configured `auth-method` is, the server can always generate JWT tokens that can be used. Like that we should always accept JWT tokens in the openapi documentation if auth is enabled. I changed this in [this commit](https://github.com/inmanta/inmanta-core/pull/9032/commits/0bfe3b2e0b856ce01c96c5a21936f644845d2156)

sanderr (approved):

PR #9078, opened 2025-05-13
Update notation for parameters in path of URL in openapi documentation

arnaudsjs:
# Description

Mark parameters in the path of an endpoint using `<param>` instead of `{param}` in the openapi documentation. This way the notation is consistent between the notation used in the code, in the access policy file and in the openapi documentation.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/update-template-marker-openapi-docs.yml
+++ b/changelogs/unreleased/update-template-marker-openapi-docs.yml
@@ -1,0 +1,6 @@
+---
+description: 'Mark parameters in the path of an endpoint using `<param>` instead of `{param}` in the openapi documentation.'
+change-type: patch
+destination-branches: [master]
+sections:
+  minor-improvement: "{{description}}"
--- a/src/inmanta/protocol/openapi/converter.py
+++ b/src/inmanta/protocol/openapi/converter.py
@@ -94,9 +94,10 @@ class OpenApiConverter:
         for path, methods in self.global_url_map.items():
             api_methods = self._filter_api_methods(methods)
             if len(api_methods) > 0:
-                path_in_openapi_format = self._format_path(path)
-                path_item = self._extract_operations_from_methods(api_methods, path_in_openapi_format)
-                paths[path_in_openapi_format] = path_item
+                url_method: UrlMethod = next(iter(methods.values()))
+                parameterized_path: str = url_method.get_path()
+                path_item = self._extract_operations_from_methods(api_methods, parameterized_path)
+                paths[parameterized_path] = path_item
         security: list[dict[SecuritySchemeName, list[Scope]]] | None = (
             # Set the security scheme globally for all endpoints.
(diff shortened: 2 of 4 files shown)


changelogs/unreleased/update-template-marker-openapi-docs.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: 'Mark parameters in the path of an endpoint using `<param>` instead of `{param}` in the openapi documentation.'
+change-type: patch
+destination-branches: [master]
```
arnaudsjs, line 4: Should we also merge this into iso8 and iso7 to keep the openapi documentation consistent across iso versions?

src/inmanta/protocol/openapi/converter.py
```diff
@@ -94,9 +94,10 @@ def generate_openapi_definition(self) -> OpenAPI:
         for path, methods in self.global_url_map.items():
             api_methods = self._filter_api_methods(methods)
             if len(api_methods) > 0:
-                path_in_openapi_format = self._format_path(path)
```
Hugo-Inmanta, line 97: The `_format_path` method can probably go now.

changelogs/unreleased/update-template-marker-openapi-docs.yml
Hugo-Inmanta: Sounds like a good idea to me :+1:

Hugo-Inmanta (approved):

src/inmanta/protocol/openapi/converter.py
```diff
@@ -94,9 +94,10 @@ def generate_openapi_definition(self) -> OpenAPI:
...
-                path_item = self._extract_operations_from_methods(api_methods, path_in_openapi_format)
-                paths[path_in_openapi_format] = path_item
+                url_method: UrlMethod = next(iter(methods.values()))
+                parameterized_path: str = url_method.get_path()
```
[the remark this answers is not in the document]
arnaudsjs: `path` contains a regex to match that path. `get_path()` returns the path as shown in the openapi documentation, but parameters are formatted as `<param>` instead `{param}`.

src/inmanta/protocol/openapi/converter.py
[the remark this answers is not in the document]
arnaudsjs: I thought I removed it but it seems I didn't. Thanks!

arnaudsjs: I thought I removed it but it seems I didn't. Thanks!

arnaudsjs:
> Other than that I have to say I lack the context to review this change conceptually. The implementation looks good to me, but I can't say if it's desired. We seemed to make an explicit effort to convert it before.

I checked this with Bart and he couldn't think of a reason why we would want to have it formatted differently in the openapi documentation.

sanderr (approved):

PR #9094, opened 2025-05-15
Fix bug: JWT config not cleared

arnaudsjs:
# Description

The `AuthJWTConfig` class keeps a cache with the JWT configuration. This configuration is derived on the configuration in the Config class. If the Config class is updated, the AuthJWTConfig class should be reset so that it picks up the new configuration. This reset didn't happen, which mainly caused issues for the test suite. This PR fixes that problems. 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/config.py
+++ b/src/inmanta/config.py
@@ -174,7 +174,7 @@ class Config:
     def _clear_jwt_config_cache(cls) -> None:
         """
         Clear the cached JWT config. This method must be called after (re)loading
-        the config, because it can cause the cache to become out of sync.
+        the config, because it can cause the cache to become stale.
         """
         from inmanta.protocol.auth import auth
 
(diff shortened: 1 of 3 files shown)


jptrindade (approved):

src/inmanta/config.py
```diff
@@ -166,6 +170,18 @@ def _reset(cls) -> None:
         cls.__instance = None
         cls._config_dir = None
         cls._min_c_config_file = None
+        cls._config_updated()
+
+    @classmethod
+    def _config_updated(cls) -> None:
```
sanderr, line 176: I think I'd prefer a listener-type interface, also to prevent the import issue. But we can do that if a second use case pops up.

sanderr (approved):

PR #9110, opened 2025-05-19
Add default policy

arnaudsjs:
# Description

This PR:

* Add the default policy
* Makes the labels for projects and environments more fine grained.
* Removes labels that are not/no longer used.
* Exclude the login endpoint from authentication
* Fixes a bug where the args of the API calls were passed in using the format of the `@handle` method instead of the `@method/typedmethod` method.

Question to reviewer: How do I test the remaining lsm-specific roles? Do I add these to the lsm extension. That feels a bit weird to me, but I don't see a better alternative.

# Self Check:

- [x] Changelog entry


tests/conftest.py
```diff
@@ -705,7 +706,23 @@ async def enable_auth() -> bool:
 
 
 @pytest.fixture
-async def access_policy() -> str:
+def auth_method() -> AuthMethod:
```
wouterdb, line 709:
```suggestion
def authentication_method() -> AuthMethod:
```

to be clear which of the two auth's this is about

wouterdb (approved):

src/inmanta/protocol/auth/default_policy.rego
```diff
@@ -0,0 +1,159 @@
...
+# to make the policy easier to read.
+endpoint_data := data.endpoints[input.request.endpoint_id]
+
+# The environment used in the request
+request_environment := input.request.parameters[endpoint_data.environment_param] if {
+    endpoint_data.environment_param != null
+} else := null
+
+# Don't allow anything that is not explicitly allowed.
+default allowed := false
```
sanderr, line 13: I only notice this now that I'm referencing the language documentation. Any reason we use `allowed` instead of `allow` used in the language examples?

src/inmanta/protocol/auth/default_policy.rego
```diff
@@ -0,0 +1,159 @@
...
+
+### General rules ###
+
+# Every user can change its own password
```
sanderr, line 17:
```suggestion
# Every user can change their own password
```

src/inmanta/protocol/auth/default_policy.rego
```diff
@@ -0,0 +1,159 @@
...
+    endpoint_data.auth_label in read_only_labels
+}
+
+
+## Role: noc ###
+
+# Can do all operations in an environment that do not alter the desired state or modify the settings.
+
+noc_specific_labels := {
+    "compiler.execute",
```
sanderr, line 73: I wonder if "update project & recompile" shouldn't be considered an admin action as well? Admittedly, it's a bit more difficult than the others because it's the same endpoint as plain recompile. But it does "alter the desired state", and in such a way that it more or less allows remote code execution, provided you also have write access to the configured repository.

sanderr (changes requested):
> Question to reviewer: How do I test the remaining lsm-specific roles? Do I add these to the lsm extension. That feels a bit weird to me, but I don't see a better alternative.

I think so. Conceptually, I'd have preferred the access control for lsm to reside there as well, but I see how that would make things needlessly difficult. The testing belongs there either way. Regardless of where the default policy is defined, it's there that the labels are defined.

src/inmanta/const.py
```diff
@@ -454,25 +454,30 @@ class AuthorizationLabel(str, Enum):
     DOCS_READ = "docs.read"
     DRYRUN_READ = "dryrun.read"
     DRYRUN_WRITE = "dryrun.write"
+    EXECUTOR_HALT_RESUME = "executor.halt-resume"
```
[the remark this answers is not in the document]
arnaudsjs: > I don't think "executor" is appropriate here. Afaik, it's about pausing logical agents, while an executor may be responsible for more than one logical agent.

I agree.

> And in the endpoint we call it pause rather than halt, but I'm not sure how relevant that is.

Right. In the web-console we use halt though. But we probably want to make this in-line with the API.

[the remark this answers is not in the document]
arnaudsjs: Oh. I see. We use "pause' for agents and "halt" for environment. I never noticed that subtle difference.

src/inmanta/protocol/auth/default_policy.rego
arnaudsjs: Not really. Do you prefer allow?

src/inmanta/protocol/auth/default_policy.rego
arnaudsjs: I agree that "update and recompile" allows you to do remote code execution, but noc users need to be able to execute this operation as part of their job. So I don't think we can make it an admin action. Also, I don't think the idea here is to protect against people with bad intentions, but more to only give users the access they actually need. @bartv Do you have an opinion on this?

bartv: so for me that one has a different level of access than a normal compile. I lean towards admin as well, NOC users really do not do this.

setup.py
```diff
@@ -77,6 +77,7 @@
     },
     # Packaging
     package_dir={"": "src"},
+    package_data={"inmanta.protocol.auth": ["*.rego"]},
```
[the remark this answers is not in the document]
sanderr: Ah, I'm surprised that even that is required, good to know.

src/inmanta/protocol/auth/default_policy.rego
sanderr: I have a tiny preference for it, solely because of the chance that someone who isn't familiar with open-policy agent would go to the rego docs and start copy-pasting something together.

sanderr (approved):

Issue #8982, opened 2025-04-25
Add expert option to bypass _check_version_info


Hugo-Inmanta:
Original comment:
> I wonder if we may want to provide an option to bypass this. Perhaps one to forcefully update the code and one to export only resources without touching code. An expert feature let's say, because you should only use it if you're confident the code changed in such a way that it is compatible.
I think that especially with multi version lsm this will become very relevant: code will change regularly without affecting existing versions (ideally). This mechanism will be in the way then.

Connected to https://github.com/inmanta/inmanta-core/issues/8729

Hugo-Inmanta:
As discussed in planning meeting:

> Perhaps one to forcefully update the code...

This should be a cli option e.g. `--force-code-update` (i.e. not an expert mode feature to toggle via the dashboard).
 
> ... and one to export only resources without touching code.

Don't implement this -> too dangerous/too little use case.

PR #9129, opened 2025-05-21
add cli option to bypass module version check

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/8982

Add docs page about expert features -> https://github.com/inmanta/inmanta-core/issues/9203

# Self Check:

- [ ] Changelog entry


jptrindade (approved):

src/inmanta/app.py
```diff
@@ -465,6 +465,19 @@ def export_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
         action="store_true",
         default=False,
     )
+    parser.add_argument(
+        "--bypass-base-version-check",
```
sanderr, line 469: I think we should improve the name. Throughout the code, I don't mind too much (though consistency is probably best), but here we really need to come at it from the end user perspective. "base version check" means little to them. The error they get doesn't mention it either.

So, think about the end user, what is it they're really trying to do, and what triggers them to do it?

src/inmanta/app.py
```diff
@@ -465,6 +465,19 @@ def export_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
...
+        dest="bypass_base_version_check",
+        help=(
+            "During partial exports, a check is performed to make sure the source code of Inmanta modules is identical to "
```
sanderr, line 472: I'd prefix this with `[Expert]`

src/inmanta/app.py
```diff
@@ -465,6 +465,19 @@ def export_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
...
+            "the code used in the base version. If you don't have the option to perform a full export and if you are confident "
+            "that the code changed in such a way that it remains compatible, you can bypass this check with this option. "
```
sanderr, line 474: This is very verbose. I'd really try to keep it to the point, and be descriptive rather than speculative. e.g.

"Force handler code update during partial compile. This is otherwise only allowed for full compiles. Use with extreme caution, and only when confident that all code is compatible with previous versions."

And then the last part about the `--partial` flag.

It's 31 words versus 70, and I dare say it's just as clear.

src/inmanta/protocol/methods_v2.py
```diff
@@ -52,6 +52,7 @@ def put_partial(
     resource_sets: Optional[dict[ResourceIdStr, Optional[str]]] = None,
     removed_resource_sets: Optional[list[str]] = None,
     pip_config: Optional[PipConfig] = None,
+    bypass_base_version_check: bool = False,
```
sanderr, line 55: This one is end-user exposed as well, so same name and help remark as on the CLI.

src/inmanta/app.py
sanderr: > The error they get doesn't mention it either.

I withdraw this part, I see you updated it. That makes it better, but I still think it can be improved.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -675,7 +675,10 @@ async def _check_version_info(
             module_version = module_data.version
             if (inmanta_module_name, module_version) not in base_version_data:
                 raise BadRequest(
-                    "Cannot perform partial export because of version mismatch for module %s." % inmanta_module_name
+                    "Cannot perform partial export because the source code for module %s in this partial version is different "
+                    "from the source code of the base version. Consider running a full export instead. Alternatively, if you "
```
sanderr, line 679:
```suggestion
                    "from the currently registered source code. Consider running a full export instead. Alternatively, if you "
```

I'm not sure about this. But I feel like "base version" is a bit too detailed, and it's not a term we expose elsewhere. As far as the user is concerned, I think the abstract description in my suggestion says all?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -885,7 +895,9 @@ async def _put_version(
             for agent in all_agents:
                 await self.agentmanager_service.ensure_agent_registered(env, agent, connection=connection)
 
-            await self._register_agent_code(partial_base_version, version, env.id, module_version_info, connection)
+            await self._register_agent_code(
+                partial_base_version, version, env.id, module_version_info, connection, bypass_base_version_check
```
sanderr, line 899: We have a sort of unwritten convention that `connection` is always the last parameter. That said, I don't think we should ever rely on it, and imo it should always be kw-only. I care less about the other options, but I think the bypass option is also a good candidate for kw-only.

sanderr (changes requested):
Just a heads up, this is probably demo meeting material.

changelogs/unreleased/8982-add-cli-option-to-bypass-check-version-info.yml
```diff
@@ -0,0 +1,4 @@
+description: Add CLI option to bypass _check_version_info in _put_version.
```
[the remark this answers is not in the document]
Hugo-Inmanta: No strong opinion either way, I asked during the planning meeting about the changelog entry for the epic and I think it was concluded that the refactor itself is mostly internal and the cli option should be self-documenting.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -673,12 +673,15 @@ async def _check_version_info(
         )
         for inmanta_module_name, module_data in module_version_info.items():
             module_version = module_data.version
+            if not module_data.for_agents:
+                # No agent is using this module in this version
+                continue
```
Hugo-Inmanta, line 678: @sanderr Do you agree this was a bug ? It triggers e.g. in the following lsm test case:
I think the following is happening:
- The `unittest` module is not ever used but:
        - it gets registered during the exports -> will be returned by `code_manager.get_module_version_info()` with en empty `for_agents=[]`
        - it will never be registered in the `AgentModules` table.
        
Is this correct ? i.e. only checking version for modules which are actually used in the current version ?
 https://github.com/inmanta/lsm/blob/8ec018f10c3717c07d38c156d2d39ea8613defef/tests/integration/test_service_entities_exporter.py#L3518

sanderr: I agree there's a bug here. But I'm not sure this is an acceptable fix. You're checking the current export's `for_agents` (right?). Meaning that you only check that it's not used by the new resources, but it may still be used by existing resources from the base version.

Either we need to be smarter about it, or we need to make sure it gets registered in the table anyway. I need to give this more thought.

Hugo-Inmanta: >  You're checking the current export's for_agents (right?)  
Yes

src/inmanta/protocol/methods_v2.py
```diff
@@ -78,9 +78,9 @@ def put_partial(
               * version_info: Model version information
     :param pip_config: Pip config used by this version
     :param module_version_info: Map of (module name, module version) to InmantaModule
-    :param bypass_base_version_check: During partial compiles, a check is performed to make sure the source
+    :param allow_handler_code_update: During partial compiles, a check is performed to make sure the source
         code of modules in this partial version is identical to the source code in the base version. Set this
-        parameter to True to bypass this check.
+        parameter to True to bypass this check and allow updating handler code during a partial export.
```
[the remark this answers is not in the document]
sanderr: The one from the cli help string

src/inmanta/data/sqlalchemy.py
```diff
@@ -67,12 +68,13 @@ class InmantaModule(Base):
 
     @classmethod
     async def register_modules(
-        cls, environment: uuid.UUID, module_version_info: dict[str, InmantaModuleDTO], connection: asyncpg.Connection
+        cls, environment: uuid.UUID, modules: dict[InmantaModuleName, InmantaModuleDTO], connection: asyncpg.Connection
     ) -> None:
         """
-        Register modules
+        Register Inmanta modules.
```
sanderr, line 74: This needs more. It leaves open questions like "what if the modules are already registered"? What if it is already registered for the same version? ...

sanderr (changes requested):
A few more comments, starting to look good to me.

PR #9158, opened 2025-05-27
Add support for roles

arnaudsjs:
# Description

Add support for roles when using database authentication.

# Self Check:

- [ ] Changelog entry


src/inmanta/data/model.py
```diff
@@ -786,6 +786,16 @@ class CurrentUser(BaseModel):
     username: str
 
 
+class Role(BaseModel):
```
wouterdb, line 789: I was a bit confused here, because in the DB roles are global, here they are per env.

But from the code, I can derive this is an assignment of roles to a user.

I wonder if we should either add the user as a field or change the name to `RoleAssignment`. 

I'm uncertain, so feel free to ignore.

tests/db/migration_tests/test_v202505150_to_v202505260.py
```diff
@@ -0,0 +1,40 @@
...
+async def test_add_claim_table(
+    postgresql_client: asyncpg.Connection,
+    migrate_db_from: abc.Callable[[], abc.Awaitable[None]],
+    get_tables_in_db: abc.Callable[[], abc.Awaitable[list[str]]],
+) -> None:
+    assert "role" not in await get_tables_in_db()
+    assert "role_assignment" not in await get_tables_in_db()
+    await migrate_db_from()
+    assert "role" in await get_tables_in_db()
+    assert "role_assignment" in await get_tables_in_db()
```
wouterdb, line 40: I don't think we need any test for this. 

i.e. this doesn't really test anything that is not covered by other tests.

tests/server/test_auth.py
```diff
@@ -606,3 +604,291 @@ async def test_get_input_for_policy_engine(capture_input_for_policy_engine: Capt
...
+        default allow := false
+
+        # Allow access if the user has the role a_role.
+        allow if {
+            input.token["urn:inmanta:roles"][request_environment] == "a_role"
+        }
+
+        # Allow access to admin user.
+        allow if {
+            input.token["urn:inmanta:is_admin"]
```
wouterdb, line 633: admin is not a role, but a special case? I suppose that is a good thing.

wouterdb (approved):

src/inmanta/db/versions/v202505260.py
```diff
@@ -0,0 +1,43 @@
...
+            PRIMARY KEY(id)
+        );
+        CREATE TABLE IF NOT EXISTS public.role_assignment(
+            user_id UUID NOT NULL,
+            environment UUID NOT NULL,
+            role_id UUID NOT NULL,
+            PRIMARY KEY(user_id, environment, role_id),
+            FOREIGN KEY (environment) REFERENCES public.environment(id) ON DELETE CASCADE,
+            FOREIGN KEY (user_id) REFERENCES public.inmanta_user(id) ON DELETE CASCADE,
+            FOREIGN KEY (role_id) REFERENCES public.role(id) ON DELETE RESTRICT
```
sanderr, line 40: Why `RESTRICT`? Not opposed per se, but I had expected `CASCADE` for this one as well.

src/inmanta/protocol/auth/policy_engine.py
```diff
@@ -171,32 +171,78 @@ async def stop(self) -> None:
...
+            return []
+        roles = response["result"]
+        if not isinstance(roles, list):
+            raise Exception(f"roles defined in access policy must be a list, got: {roles}")
+        for elem in roles:
+            if not isinstance(elem, str):
+                raise Exception(f"The list of roles defined in the access policy contains a non-string element: {elem}")
+        return roles
+
+    async def _synchronize_roles_to_db(self) -> None:
```
sanderr, line 241: Nice, hadn't considered that.

sanderr (changes requested):
1 question. Requesting changes to make it visible, though it isn't really a change request.

tests/server/test_auth.py
arnaudsjs: Yes. It's a marker to indicate you are a global admin, which can access everything. It's not scoped to an environment.

src/inmanta/db/versions/v202505260.py
arnaudsjs: I want to prevent the scenario where we accidentally delete a role that is still assigned to one or more users. Normally the code shouldn't allow that. If it would be possible, that would be bad. Users would lose their role. Hence the RESTRICT. Now that I explain it, I realize it could also be considered a feature.

sanderr (approved):

src/inmanta/db/versions/v202505260.py
sanderr: Yeah, I think it can be seen both ways.

src/inmanta/data/model.py
arnaudsjs: I went for the rename to `RoleAssignment`.

PR #9166, opened 2025-05-28
Update scheduler fallout recovery

jptrindade:
# Description

Update scheduler fallout recovery to not use increment calculation

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4986,7 +4986,6 @@ class Resource(BaseDocument):
         environment: uuid.UUID,
         *,
         since: Optional[int],
-        projection: Optional[Collection[typing.LiteralString]],
         connection: Optional[Connection] = None,
     ) -> list[tuple[int, list[dict[str, object]]]]:
         """
@@ -5005,12 +5004,11 @@ class Resource(BaseDocument):
                             """
         )
 
-        resource_columns: typing.LiteralString = ", ".join(f"r.{c}" for c in projection) if projection is not None else "r.*"
         query: typing.LiteralString = f"""
             WITH boundary AS (
                 {boundary_query}
             )
-            SELECT m.version, {resource_columns}
+            SELECT m.version, r.resource_id, r.attributes, r.attribute_hash, r.is_undefined
             FROM {ConfigurationModel.table_name()} as m
             LEFT JOIN {cls.table_name()} as r
                 ON m.environment=r.environment AND m.version=r.model
@@ -5025,11 +5023,6 @@ class Resource(BaseDocument):
             *query_values,
             connection=connection,
         )
-        query_latest_model = f"""
-                   SELECT max(version)
-                   FROM {ConfigurationModel.table_name()}
-                   WHERE environment=$1 AND released
-               """
         result: list[tuple[int, list[dict[str, object]]]] = []
         for version, raw_resources in itertools.groupby(resource_records, key=lambda r: r["version"]):
             parsed_resources: list[dict[str, object]] = []
@@ -5038,8 +5031,6 @@ class Resource(BaseDocument):
                     # left join produced no resources
                     continue
                 resource: dict[str, object] = dict(raw_resource)
-                if projection is not None:
-                    assert set(projection) <= resource.keys()
                 parsed_resources.append(resource)
             result.append((version, parsed_resources))
         return result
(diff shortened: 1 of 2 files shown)


tests/deploy/e2e/test_recovery.py
```diff
@@ -162,17 +195,30 @@ async def get_last_deployed() -> dict[int, datetime.datetime]:
...
+            compliance=Compliance.NON_COMPLIANT,
+            last_deploy_result=DeployResult.SKIPPED,
             blocked=Blocked.NOT_BLOCKED,  # we don't restore TRANSIENT status atm
             last_deployed=last_deployed[3],
         ),
+        # If reset_state=True we will reset the blocked status to NOT_BLOCKED
+        "test::Resource[agent1,key=key4]": ResourceState(
+            compliance=Compliance.UNDEFINED,
+            last_deploy_result=DeployResult.NEW,
+            blocked=Blocked.BLOCKED if not reset_state else Blocked.NOT_BLOCKED,
```
sanderr, line 213: This is the situation I was talking about in the other comment, FYI.

src/inmanta/data/__init__.py
```diff
@@ -5022,6 +5022,52 @@ async def get_resources_since_version_raw(
...
+        for raw_resource in resource_records:
+            if raw_resource["resource_id"] is None:
+                # left join produced no resources special case for empty model
+                continue
+            resource: dict[str, object] = dict(raw_resource)
+            del resource["version"]
+            parsed_resources.append(resource)
+
+        return [(typing.cast(int, resource_records[0]["version"]), parsed_resources)]
+
```
jptrindade, line 5070: I was a bit on the fence about creating a new function instead of modifying `get_resources_since_version_raw` to just return the latest version if `self._state.version == 0` but I feel like the query could get a bit mangled up

src/inmanta/data/__init__.py
```diff
@@ -5022,6 +5022,52 @@ async def get_resources_since_version_raw(
...
+        )
+        if len(resource_records) == 0:
+            return []
+        parsed_resources: list[dict[str, object]] = []
+        for raw_resource in resource_records:
+            if raw_resource["resource_id"] is None:
+                # left join produced no resources special case for empty model
+                continue
+            resource: dict[str, object] = dict(raw_resource)
+            del resource["version"]
```
sanderr, line 5066: Why this difference with the other method?

src/inmanta/data/__init__.py
sanderr: I had the same internal conflict when I proposed this to you. I thought this would be ever so slightly better, but I'm on the fence again. Perhaps a simple `since: Optional[int]` for the other method could work. And then in its implementation `"ASC" if since is not None else "DESC LIMIT 1"`?

sanderr: See below. I'll leave the final decision up to you.

```
diff --git a/src/inmanta/data/__init__.py b/src/inmanta/data/__init__.py
index c255da581..c491ef70f 100644
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -4985,27 +4985,46 @@ class Resource(BaseDocument):
         cls,
         environment: uuid.UUID,
         *,
-        since: int,
+        since: Optional[int],
         projection: Optional[Collection[typing.LiteralString]],
     ) -> list[tuple[int, list[dict[str, object]]]]:
         """
         Returns all released model versions with associated resources since (excluding) the given model version.
         Returns resources as raw dicts with the requested fields
-        """
+
+        :param since: The boundary version (excluding). If None, returns only the latest released version.
+        """
+        boundary_query: typing.LiteralString
+        boundary_values: Sequence[object]
+        boundary_query, boundary_values = (
+            ("SELECT $2 AS version", [])
+            if since is not None
+            else (
+                """\
+                SELECT MAX(version) AS version
+                FROM public.configurationmodel
+                WHERE released IS TRUE AND environment=$1
+                """,
+                [cls._get_value(since)],
+            )
+        )
         resource_columns: typing.LiteralString = ", ".join(f"r.{c}" for c in projection) if projection is not None else "r.*"
         query: typing.LiteralString = f"""
+            WITH boundary AS (
+                {boundary_query}
+            )
             SELECT m.version, {resource_columns}
             FROM {ConfigurationModel.table_name()} as m
             LEFT JOIN {cls.table_name()} as r
                 ON m.environment = r.environment AND m.version = r.model
-            WHERE m.environment = $1 AND m.version > $2 AND m.released=true
+            WHERE m.environment = $1 AND m.version > boundary.version AND m.released=true
             ORDER BY m.version ASC
         """
         resource_records = await cls._fetch_query(
             query,
             cls._get_value(environment),
-            cls._get_value(since),
+            *boundary_values,
             connection=connection,
         )
         result: list[tuple[int, list[dict[str, object]]]] = []
```

sanderr (approved):

Issue #9001, opened 2025-04-29
Make agent status report on the status of the executor


wouterdb:
1. if the executor can't be built at all, set it to down
2. if the executor fails to load some resources, set it to degraded 

We assume the status is for the latest executor we tried to build/use for this agent

- [ ] Notify frontend team about the new `degraded` agent status.

------

Discussion:

Should we allow executors to even deploy anything at all in case they were successfully created but couldn't load all the handler code they were supposed to ?

PR #9165, opened 2025-05-28
Issue/8952 module only imported in plugin code

Hugo-Inmanta:
# Description

part of  https://github.com/inmanta/inmanta-core/issues/8952

In case a module is imported in plugin code, but not in model code:

- it will not be loaded by the project
- it was causing the following error during code registration:
```
ERROR    pytest_inmanta_lsm.remote_orchestrator:remote_orchestrator.py:570 Subprocess exited with code 1: compiler       WARNING ProjectConfigurationWarning: Setting a pip index through the `repo.url` option with type `package` in the project.yml file is no longer supported and will be ignored. Please set the pip index url through the `pip.index_url` option instead.
...
================================ EXPORT FAILURE ================================
Error: 'netbox'
=============================== EXCEPTION TRACE ===============================
Traceback (most recent call last):
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/app.py", line 695, in capture
    yield self
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/app.py", line 584, in export
    results = export.run(
              ^^^^^^^^^^^
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/export.py", line 442, in run
    self._version = self.commit_resources(
                    ^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/export.py", line 560, in commit_resources
    self.deploy_code(conn, tid, version)
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/export.py", line 516, in deploy_code
    code_manager.register_code(type_name, resource_definition)
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/loader.py", line 158, in register_code
    all_plugin_files: list[SourceInfo] = self._get_source_info_for_module(module_name)
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/loader.py", line 167, in _get_source_info_for_module
    SourceInfo(path, module_name) for path, module_name in module.Project.get().modules[module_name].get_plugin_files()
                                                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'netbox'

================================ EXPORT FAILURE ================================
```
- This PR will ~log a warning in such a case and ignore code registration for this module~ raise an exception in such a case.

[...]

PR #9171, opened 2025-05-30
Issue/9001 make agent status report on the status of the executor

Hugo-Inmanta:
# Description

~This PR adds an xfail test `test_code_registration_bug_reproduction` in `tests/deploy/e2e/test_autostarted.py` that should be updated/fixed by https://github.com/inmanta/inmanta-core/pull/9165~
EDIT: removed this xfail test, this was a red herring; it was caused by an artifact of the test framework (calling `reset_all_objects()` between the 2 compiles fixed the issue)

closes https://github.com/inmanta/inmanta-core/issues/9001

# Self Check:

- [ ] Changelog entry


src/inmanta/db/versions/v202505300.py
```diff
@@ -0,0 +1,44 @@
...
+            'degraded',
+            'down'
+        );
+
+        ALTER TABLE public.agent ADD COLUMN executor_status executor_status DEFAULT 'down'::public.executor_status;
+
+        UPDATE public.agent
+        SET executor_status = 'down'::public.executor_status;
+
+        ALTER TABLE  public.agent ALTER COLUMN executor_status SET NOT NULL;
```
Hugo-Inmanta, line 42: what about updates on a live orchestrator?

Hugo-Inmanta: Should be ok, since we halt the env before the update and the orchestrator restarts after the update.

src/inmanta/deploy/tasks.py
```diff
@@ -179,9 +180,16 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
                     )
                     # Not attached to ctx, needs to be flushed to logger explicitly
                     log_line.write_to_logger_for_resource(agent, executor_resource_details.rvid, exc_info=True)
+                    await task_manager.report_executor_status(agent_name=agent, executor_status=ExecutorStatus.down)
```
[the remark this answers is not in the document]
sanderr: I think so. It's not down if it might still successfully deploy other resource types, is it? @wouterdb CC

tests/deploy/e2e/test_autostarted.py
```diff
@@ -1492,9 +1502,57 @@ async def test_code_install_success_code_load_error_for_provider(
...
+    )
+
+    version, res, status = await snippetcompiler.do_export_and_deploy(include_status=True)
+    result = await client.release_version(environment, version, push=False)
+    assert result.code == 200
+
+    await wait_for_resources_in_state(client, uuid.UUID(environment), nr_of_resources=1, state=const.ResourceState.unavailable)
+
+    await retry_limited(partial(check_agent_status, ExecutorStatus.down, "agent_1"), 2)
+
```
Hugo-Inmanta, line 1555: Here

src/inmanta/deploy/tasks.py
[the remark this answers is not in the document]
Hugo-Inmanta: > Also, add a test for this? i.e. that an agent does not report down just because the last deploy attempt was for a resource type that has failures.

I'll update this test that is asserting the exact opposite behaviour https://github.com/inmanta/inmanta-core/pull/9171/files#r2128937623

PR #9234, opened 2025-06-17
Issue/9220 dicts of references

wouterdb:
# Description

Allow serialization of references inside json structures inside references

closes 9220

This add a new type of serialized object: json with mutators. 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/references.py
+++ b/src/inmanta/references.py
@@ -333,11 +333,8 @@ class ReferenceLike:
                 case str() | int() | float() | bool() | None:
                     arguments.append(LiteralArgument(name=name, value=value))
                 case dict() | list():
-                    # TODO
-                    from inmanta.resources import ReferenceSubCollector, collect_references
-
                     collector = ReferenceSubCollector()
-                    cleaned_value = collect_references(collector, value, "")
+                    cleaned_value = inmanta.resources.collect_references(collector, value, "")
                     try:
                         if collector.references:
                             arguments.append(
@@ -583,3 +580,39 @@ def is_reference_of(instance: typing.Optional[object], type_class: type[object])
         return False
 
     return instance.get_reference_type() == type_class
+
+
+class ReferenceSubCollector:
+
+    def __init__(self) -> None:
+        self.references: dict[uuid.UUID, ReferenceModel] = {}
+        self.replacements: dict[str, ReferenceModel] = {}
+
+    def collect_reference(self, value: object) -> None:
+        """Add a value reference and recursively add any other references."""
+        match value:
+            case list():
+                for v in value:
+                    self.collect_reference(v)
+
+            case dict():
+                for k, v in value.items():
+                    self.collect_reference(v)
+
+            case Reference():
+                ref = value.serialize()
+                self.references[ref.id] = ref
+                for arg in value.arguments.values():
+                    self.collect_reference(arg)
+
+            case _:
+                pass
+
+    def add_reference(self, path: str, reference: "Reference[PrimitiveTypes]") -> None:
+        """Add a new attribute map to a value reference that we found at the given path.
+
+        :param path: The path where the value needs to be inserted
+        :param reference: The attribute reference
+        """
+        self.collect_reference(reference)
+        self.replacements[path] = reference.serialize()
--- a/src/inmanta/references.py
+++ b/src/inmanta/references.py
@@ -171,7 +171,7 @@ class ReferenceArgument(Argument):
 
 class MutatedJsonArgument(Argument):
     """
-    Json-like argument to a reference that contains reference in the json structure
+    Json-like argument that contains another reference in the json structure
 
     It is stored as
     1. a json value with all reference replaced by None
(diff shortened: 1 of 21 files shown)


src/inmanta/references.py
```diff
@@ -169,6 +168,26 @@ def get_arg_value(
...
+    type: typing.Literal["mjson"] = "mjson"
+    value: StrictJson
+    references: dict[str, ReferenceArgument]
+
+    def get_arg_value(
+        self,
+        resource: "inmanta.resources.Resource",
+        logger: "handler.LoggerABC",
+    ) -> object:
+        start_value: JsonType = self.value
```
wouterdb, line 183: This is a type error

Hugo-Inmanta (approved):
One small docstring improvement request. I can see it does what it is supposed to in the test case but I don't feel like I have enough context awareness to review in detail.

src/inmanta/references.py
```diff
@@ -169,6 +169,26 @@ def get_arg_value(
         return resource.get_reference_value(self.id, logger)
 
 
+class MutatedJsonArgument(Argument):
+    """Json-like argument to a reference"""
```
Hugo-Inmanta, line 173: Docstring should explain the difference with `class JsonArgument(Argument)`.

src/inmanta/references.py
```diff
@@ -169,6 +169,33 @@ def get_arg_value(
         return resource.get_reference_value(self.id, logger)
 
 
+class MutatedJsonArgument(Argument):
+    """
+    Json-like argument to a reference that contains reference in the json structure
```
sanderr, line 174:
```suggestion
    Json-like argument that contains another reference in the json structure
```
Right? The reference has nothing to do with it? Or does it?

src/inmanta/references.py
```diff
@@ -169,6 +169,33 @@ def get_arg_value(
...
+    def get_arg_value(
+        self,
+        resource: "inmanta.resources.Resource",
+        logger: "handler.LoggerABC",
+    ) -> object:
+        start_value: JsonType = self.value
+        for destination, valueref in self.references.items():
+            value = valueref.get_arg_value(resource, logger)
+            dict_path_expr = dict_path.to_path(destination)
+            dict_path_expr.set_element(start_value, value)
```
sanderr, line 195: For list indices, this requires the improved dict path, doesn't it?

src/inmanta/references.py
```diff
@@ -301,14 +334,37 @@ def serialize(self) -> SerializedReferenceLike:
 
     def serialize_arguments(self) -> Tuple[uuid.UUID, list[ArgumentTypes]]:
         """Serialize the arguments to this class"""
+
+        # The handling of references here is a but subtle:
```
sanderr, line 338:
```suggestion
        # The handling of references here is a bit subtle:
```

src/inmanta/references.py
```diff
@@ -301,14 +334,37 @@ def serialize(self) -> SerializedReferenceLike:
...
+        # We replace every reference with a ReferenceArgument that refers to the reference by id
+        # The reference itself is not handled here but in inmanta.resources.ReferenceSubCollector.collect_reference
+        # There, the raw, unserialized tree of arguments is iterated over as well to collect the references themselves
+        # The caches on the Reference prevent this from being too inefficient by serializing only once
```
sanderr, line 342: I need to let this sink in.

src/inmanta/resources.py
```diff
@@ -195,30 +195,28 @@ def __new__(cls, class_name, bases, dct):
 RESERVED_FOR_RESOURCE = {"id", "version", "model", "requires", "unknowns", "set_version", "clone", "is_type", "serialize"}
 
 
-class ReferenceCollector:
-    """Collect and organize all references and mutators"""
+class ReferenceSubCollector:
```
sanderr, line 198: This could use a docstring

src/inmanta/references.py
sanderr: Do we really need a dedicated argument with dedicated mutation? Can't we extend `ReplaceValue` to accept `Resource | JsonArgument` and add one of those to the (real) collector?

sanderr (changes requested):

src/inmanta/references.py
wouterdb: I'm going to pick that up next

src/inmanta/resources.py
wouterdb: made an attempt

sanderr (approved):

src/inmanta/references.py
```diff
@@ -307,8 +332,22 @@ def serialize_arguments(self) -> Tuple[uuid.UUID, list[ArgumentTypes]]:
                 case str() | int() | float() | bool() | None:
                     arguments.append(LiteralArgument(name=name, value=value))
                 case dict() | list():
+                    collector = inmanta.resources.ReferenceSubCollector()
+                    cleaned_value = inmanta.resources.collect_references(collector, value, "")
```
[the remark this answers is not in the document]
sanderr: It's a tough one to resolve because of the `None` values. `typing.overload()` could do it

src/inmanta/references.py
wouterdb: Is this true? Isn't the point of ReplaceValue to support nested references? 
 -> it is to support references in resources, but not references in references  (i.e. the not nested part is true)

PR #9238, opened 2025-06-18
Small fix to default policy

arnaudsjs:
# Description

This PR:

* Make sure that the `upload_file` endpoint can be executed by any authenticated user. Up to now there was not a rule that allowed it because the endpoint is not environment scoped and not read-only.
* Fix documentation of some endpoints.

# Self Check:

- [x] Changelog entry


src/inmanta/protocol/methods.py
```diff
@@ -353,7 +353,7 @@ def upload_file(id: str, content: str):
     """
     Upload a new file
 
-    :param id: The id of the file
+    :param id: The sha1 hash of the content of the file.
```
sanderr, line 356: Actually pretty strange that we take this as a parameter, because we also still calculate it (as we should; I was scared for a bit that we'd take the user's word for granted but luckily we don't).

sanderr (approved):

Issue #9061, opened 2025-05-09
Make health check of containers more robust


arnaudsjs:
The health check of the containers (OSS + ISO) makes assumptions about the server configuration that do not always hold. In those situation the status of the container never goes to a healthy state. This ticket should make the health check more robust.

Known issues:

* If authentication is enabled, the health check will result in a HTTP 401 if no valid token is configured in the `client_rest_transport.token` config option.
* If the server runs using a non-default port, the health check will connect to the wrong port
* What if SSL is enabled?
* ...

arnaudsjs:
Replaced by: https://github.com/inmanta/inmanta-service-orchestrator/issues/560

PR #9239, opened 2025-06-18
Added the /health endpoint

arnaudsjs:
# Description

The `/health` endpoint provides a yes/no answer on whether the server is healthy or not. It doesn't require authentication in contrast to the /serverstatus endpoint.

Part of https://github.com/inmanta/inmanta-core/issues/9061

# Self Check:

- [x] Changelog entry
- [ ] End user documentation is included or an issue is created for end-user documentation

### DIFF
--- a/src/inmanta/protocol/auth/providers.py
+++ b/src/inmanta/protocol/auth/providers.py
@@ -148,6 +148,9 @@ class PolicyEngineAuthorizationProvider(AuthorizationProvider):
             # On a login call the authorization is done using the provided username and password.
             # No need to authorize the token.
             return
+        if call_arguments.method_properties.function == methods_v2.health:
+            # For ease of use, the health endpoint should be accessible without authentication.
+            return
         if call_arguments.auth_token is None:
             raise exceptions.UnauthorizedException()
         if call_arguments.is_service_request():
--- a/src/inmanta/protocol/methods_v2.py
+++ b/src/inmanta/protocol/methods_v2.py
@@ -1775,3 +1775,11 @@ def graphql_schema() -> dict[str, Any]:
     To actually execute a query, use 'POST /api/v2/graphql'
     """
     pass
+
+@typedmethod(path="/health", operation="GET", client_types=[ClientType.api], enforce_auth=False, api_version=2)
+def health() -> ReturnValue[None]:
+    """
+    Returns a 200 response code if the server is healthy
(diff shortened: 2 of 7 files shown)


src/inmanta/protocol/auth/providers.py
```diff
@@ -148,6 +148,9 @@ async def _do_authorize_request(self, call_arguments: "rest.CallArguments") -> N
             # On a login call the authorization is done using the provided username and password.
             # No need to authorize the token.
             return
+        if call_arguments.method_properties.function == methods_v2.health:
+            # For ease of use, the health endpoint should be accessible without authentication.
+            return
```
sanderr, line 153: What's the rationale for configuring this here vs in the (default) policy file?

src/inmanta/protocol/methods_v2.py
```diff
@@ -1775,3 +1775,14 @@ def graphql_schema() -> dict[str, Any]:
     To actually execute a query, use 'POST /api/v2/graphql'
     """
     pass
+
+
+@typedmethod(path="/health", operation="GET", client_types=[ClientType.api], enforce_auth=False, api_version=2)
+def health() -> ReturnValue[None]:
+    """
+    Returns a 200 response code if the server is healthy
+    or a 500 if the server is not healthy.
```
sanderr, line 1784: 500 Internal Server Error? That's unconventional, isn't it?

wouterdb: on K8S livenes probes (which are the best reference I have for this), 

Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.

the example there also uses 500 to indicate failure 

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

arnaudsjs: I didn't really find a clear answer on what is conventional. Some applications return 500 or 503 for not healthy. Others return a 200 and mention the status in the json body. What do you consider the conventional way?

src/inmanta/protocol/auth/providers.py
sanderr: You make a good argument. I was only vaguely aware of the purpose of this endpoint. In that case I think it's more a property of the system than a policy, so I think it belongs here after all.

sanderr (approved):

PR #9246, opened 2025-06-20
Issue/vendor pyformance

wouterdb:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry


src/pyformance/meters/gauge.py
```diff
@@ -0,0 +1,60 @@
+from typing import Callable, Union
+
+
+class Gauge[T: Union[int, float]]:
```
wouterdb, line 4: The generic seems like a good way to type it, but it does cause a lot of complexity down the road. I expect it would change little if I would type as the union instead of bounded by the union. 

Wdyt?

src/inmanta/server/services/databaseservice.py
```diff
@@ -118,7 +119,7 @@ def get_pool_free(self) -> int:
             return 0
         return self._pool.get_max_size() - self._pool.get_size() + self._pool.get_idle_size()
 
-    def _add_gauge(self, name: str, the_gauge: CallbackGauge) -> None:
+    def _add_gauge(self, name: str, the_gauge: AnyGuage) -> None:
```
sanderr, line 122: Could even make it generic here as well?
```suggestion
    def _add_gauge(self, name: str, the_gauge: Gauge[T]) -> None:
```

src/pyformance/meters/gauge.py
sanderr: I like the type var, and I think its impact is quite OK. especially if you'd write it as `int | float` it's quite compact. But I'm not opposed to your suggestion.

FYI, Python 3.13 addresses this problem with defaults for type vars. e.g. you could declare `class Gauge[T: int | float = int | float]:` and use it as `Gauge` to mean `Gauge[int | float]`.

src/pyformance/registry.py
```diff
@@ -0,0 +1,400 @@
...
+        :param key: name of the metric
+        :type key: C{str}
+
+        :return: L{Histogram}
+        """
+        if key not in self._histograms:
+            self._histograms[key] = Histogram(clock=self._clock)
+        return self._histograms[key]
+
+    def gauge[T: float | int](self, key: str, gauge: Gauge[T] | None = None, default: T = float("nan")) -> AnyGuage:
```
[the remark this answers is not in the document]
sanderr: `typing.overload`. I think I can dig up an example in our code base.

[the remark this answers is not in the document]
sanderr: Ah no, it will after all, because you only use the default if there is no `Gauge`.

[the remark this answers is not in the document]
sanderr: But while I think `overload` could help. I think the proper solution is simply
```suggestion
    def gauge[T: float | int](self, key: str, gauge: Gauge[T] | None = None, default: float = float("nan")) -> AnyGuage:
```

You only really need `overload` if you want to better type the return type.

src/pyformance/registry.py
```diff
@@ -0,0 +1,400 @@
...
+    def timer(self, key: str) -> Timer:
+        return super(RegexRegistry, self).timer(self._get_key(key))
+
+    def histogram(self, key: str) -> Histogram:
+        return super(RegexRegistry, self).histogram(self._get_key(key))
+
+    def counter(self, key: str) -> Counter:
+        return super(RegexRegistry, self).counter(self._get_key(key))
+
+    def gauge[T: float | int](self, key: str, gauge: Gauge[T] | None = None, default: T = float("nan")) -> AnyGuage:
```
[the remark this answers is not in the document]
sanderr: Same here

src/pyformance/registry.py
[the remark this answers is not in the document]
wouterdb: ok, that fixes it, thx

src/pyformance/__init__.py
```diff
@@ -1,3 +1,28 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+Contact: code@inmanta.com
+
+This code was originally developed by Omer Gertel, as a python port of the core portion of a
+[Java Metrics library by Coda Hale](http://metrics.dropwizard.io/), w
+ith inspiration by [YUNOMI - Y U NO MEASURE IT?](https://github.com/richzeng/yunomi)
```
bartv, line 21: with is split here

NOTICE
```diff
@@ -0,0 +1,7 @@
+Inmanta Core
```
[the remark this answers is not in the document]
wouterdb: The notice file should start with the identification and copyright info of the project itself.  (or at least, they always do, I can't find formal guidance). I think this is because license forces verbatim inclusion of the notice file when vendoring it. 

As an example of a big one: https://github.com/apache/spark/blob/master/NOTICE-binary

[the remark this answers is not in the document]
wouterdb: I was in doubt as well, should I go for `Inmanta Orchestrator` ?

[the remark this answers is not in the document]
bartv: sounds good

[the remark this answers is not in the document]
wouterdb:
```suggestion
Inmanta Orchestrator
```

src/pyformance/__init__.py
```diff
@@ -0,0 +1,22 @@
+from typing import Protocol
+
+
+class Clock(Protocol):
+
+    def time(self) -> float: ...
+
+
+from .registry import MetricsRegistry as MetricsRegistry
```
wouterdb: @sanderr do we want to change the way these imports are written?

src/pyformance/__init__.py
```diff
@@ -0,0 +1,22 @@
+from typing import Protocol
```
wouterdb, line 25: It the way I include it OK? Should I relocate the package?

tests/conftest.py
```diff
@@ -571,7 +570,6 @@ def clean_reset_session():
 
 def reset_all_objects():
     resources.resource.reset()
-    asyncio.set_child_watcher(None)
```
wouterdb, line 574: deprecated as well

tests/forking_agent/ipc_commands.py
```diff
@@ -0,0 +1,66 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+Contact: code@inmanta.com
+"""
+
+## This is a separate file to cause the remote side to not load too much code
```
wouterdb, line 19: This caused a warning by loading test utils in the child process, which loads `pip2pi`  which uses pkg_resource, which causes a warning to travel over the IPC link and break stuff....

NOTICE
sanderr, line 1: I have no clue about this, but is it right that this is top-level, rather than in `src/inmanta/vendor`?

src/inmanta/vendor/__init__.py
sanderr, line 1: Is this where it was a namespace package? It's not supposed to be one anymore?

sanderr (commented):
Scope of my review:
- structure
- everything outside `inmanta.vendor` and `tests/pyformance_tests.py`
- open comments

Let me know if you'd intended me to review the vendored code more in-depth.

sanderr (approved):

src/inmanta/vendor/__init__.py
wouterdb: no, it is not supposed to be anymore

NOTICE
wouterdb: has to be top level

wouterdb:
moved to #9265

PR #9266, opened 2025-06-25
allow references in export attribute access

sanderr:
# Description

It's a bit more verbose than I'd like (adding yet another coupled field to the proxy context). But I think this might be the best we can do, considering the state of things.

This PR is based on master, but the diff is relative to the commit before the revert. I'll merge it by hand.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/export.py
+++ b/src/inmanta/export.py
@@ -146,7 +146,7 @@ class Exporter:
         for t in types:
             if self.types is not None and t in self.types:
                 proxies[t] = [
-                    DynamicProxy.return_value(i, context=ProxyContext(path=f"<{i}>"), mode=ProxyMode.EXPORT)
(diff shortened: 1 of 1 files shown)


src/inmanta/resources.py
```diff
@@ -400,16 +400,15 @@ def object_to_id(
...
-                raise e
-            except inmanta.ast.UnknownException as e:
-                raise e
-            except inmanta.ast.UnexpectedReference:
-                current_path: str = ".".join(path_elements[: i + 1])
-                raise ResourceException(
-                    "Encountered reference in resource's agent attribute. Agent attribute values can not be references."
-                    f" Encountered at attribute {current_path!r} of resource instance {model_object}"
-                )
+                if isinstance(agent_value, references.Reference):
```
wouterdb, line 403: bit of a special case, we may actually want to not allow any reference, but that can be done later.

wouterdb (approved):

PR #9268, opened 2025-06-25
Disallow role related endpoints when auth is disabled

arnaudsjs:
# Description

Refuse calls to any role-related endpoints when auth is disabled.

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (approved):

sanderr (commented):
> It's to ensure that nobody assigns him/herself the admin role while authentication is disable to use these privileges when authentication is enabled. By default authentication is disabled on the server.

But that doesn't necessarily affect the creation of new roles etc. But I guess we can always relax it later if we realize we need to.

sanderr (approved):

Issue #9275, opened 2025-06-26
Assign stdout_log before _wait_for_agents on _ensure_scheduler


jptrindade:
`stdout_log` is only assigned if `start_new_process==True` if it is not `True` we need to assign it in another way so that it might be used on `_wait_for_agents`

Related to: https://github.com/inmanta/inmanta-core/pull/8805

```
if start_new_process:
    self._agent_procs[env], stdout_log, stderr_log = await self.__do_start_agent(
        refreshed_env, connection=connection
    )

# Wait for all agents to start
try:
    await self._wait_for_agents(
        refreshed_env, autostart_scheduler, stdout_log=stdout_log, stderr_log=stderr_log, connection=connection
    )
except asyncio.TimeoutError:
    LOGGER.warning("Not all agent instances started successfully")
return start_new_process
```

PR #9287, opened 2025-06-30
Fix issue stdout_log is referenced before assignment

arnaudsjs:
# Description

Fix bug where variable `stdout_log` is referenced before assignment.

closes https://github.com/inmanta/inmanta-core/issues/9275

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/stdout_log_referenced_before_assignment.yml
+++ b/changelogs/unreleased/stdout_log_referenced_before_assignment.yml
@@ -1,0 +1,6 @@
+---
+description: "Fix bug where variable `stdout_log` is referenced before assignment."
+change-type: patch
+destination-branches: [master, iso8]
+sections:
+  bugfix: "{{description}}"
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/stdout_log_referenced_before_assignment.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: "Fix bug where starting the scheduler can result in an error saying that stdout_log is referenced before assignment. This can cause the compiler service to get stuck or trigger the agent deploy timers at incorrect moments."
```
sanderr, line 2: The user doesn't know about the scheduler.
```suggestion
description: "Fix bug where starting the agent can result in an error saying that stdout_log is referenced before assignment. This can cause the compiler service to get stuck or trigger the agent deploy timers at incorrect moments."
```

sanderr (changes requested):
Nice, very clean! One small change request on the change entry.

PR #9296, opened 2025-07-01
Add docs for authorization framework

arnaudsjs:
# Description

Add docs for authorization framework.

# Self Check:

- [x] Changelog entry


docs/administrators/authorization.rst
```diff
@@ -0,0 +1,269 @@
...
+* When an API call arrives on the Inmanta server, a query is sent to the policy engine to determine whether the API call is authorized. It does so by providing the parameters of the API call and the decoded access token to the policy engine. If the policy allows the request, the API call is executed. Otherwise a permission denied is returned to the client.
+
+When this authorization provider is enabled, it's used to evaluate authorization of end-user API calls. Machine-to-machine requests, i.e. requests that have the client type ``compiler`` or ``agent`` in their access token, are always handled using the legacy authorization provider.
+
+
+Data sources for policy engine
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+An mentioned above, the policy engine relies on three pieces of information to determine whether an API call is authorized: authorization-related metadata about each API endpoint on the server, the data in the specific API request done by the client and the access policy. The sections below discuss each of these three elements in more detail.
```
jptrindade, line 30: These three pieces of information could be itemised in bullet points

docs/administrators/authorization.rst
```diff
@@ -0,0 +1,269 @@
...
+
+* The ``sub`` claim is present and contains the username of the user the token is for.
+* Environment-scoped roles are defined in the ``urn:inmanta:roles`` claim of the access token. The value must be a dictionary that maps the ID of the environment to a list of roles the user has in that environment.
+* Global admins must have the claim ``urn:inmanta:is_admin`` in the access token with the value set to ``true``.
+
+
+Integration with database authentication
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The default policy integrates seemlessly with database authentication.
```
jptrindade, line 155:
```suggestion
The default policy integrates seamlessly with database authentication.
```

docs/administrators/authorization.rst
```diff
@@ -0,0 +1,269 @@
...
+3. If a 3rd party auth broker is used, it must be configured to correctly set all claims required by the access policy.
+4. Restart the inmanta server to activate the configuration changes.
+
+
+.. _authorization-writing-own-policy:
+
+Writing a custom access policy
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This section provides some guidence on how to test/debug you own access policy.
```
jptrindade, line 192:
```suggestion
This section provides some guidance on how to test/debug your own access policy.
```

docs/administrators/authorization.rst
```diff
@@ -0,0 +1,269 @@
...
+
+
+Troubleshooting: Policy engine fails to start
+"""""""""""""""""""""""""""""""""""""""""""""
+
+When the access policy contains a syntax error, the policy engine will fail to start and as such also the Inmanta server. In that case, the ``server.log`` file will indicate the policy engine failed to start with a reference to the log file of the policy engine. The latter log file will contains more information about the specific issue. If required, the log level of the policy engine can be increased using the :inmanta.config:option:`policy_engine.log-level` config option.
```
jptrindade, line 198:
```suggestion
When the access policy contains a syntax error, the policy engine will fail to start and as such also the Inmanta server. In that case, the ``server.log`` file will indicate that the policy engine failed to start with a reference to the log file of the policy engine. The latter log file will contain more information about the specific issue. If required, the log level of the policy engine can be increased using the :inmanta.config:option:`policy_engine.log-level` config option.
```

docs/administrators/authorization.rst
```diff
@@ -0,0 +1,275 @@
...
+When this authorization provider is enabled, it's used to evaluate authorization of end-user API calls. Machine-to-machine requests, i.e. requests that have the client type ``compiler`` or ``agent`` in their access token, are always handled using the legacy authorization provider.
+
+
+Data sources for policy engine
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+As mentioned above, the policy engine relies on three pieces of information to determine whether an API call is authorized:
+
+* Authorization-related metadata about each API endpoint on the server
+* The data in the specific API request done by the client
```
sanderr, line 33:
```suggestion
* The data in the specific API request made by the client
```
"done" sounds weird to me.

docs/administrators/authorization.rst
```diff
@@ -0,0 +1,275 @@
...
+        }
+    }
+
+The ``inmanta policy-engine print-endpoint-data`` command prints the full datastructure for all endpoints on the inmanta server.
+
+
+Data in API request
+"""""""""""""""""""
+
+The parameters of the specific API call being done and the decoded access token are available in the access policy using respectively
```
sanderr, line 77:
```suggestion
The parameters of the specific API call being processed and the decoded access token are available in the access policy using respectively
```
Same here, but I'm less sure about my suggestion.

docs/administrators/authorization.rst
```diff
@@ -0,0 +1,275 @@
...
+   * **environment-admin**: Users with this role can do anything in a certain environment, except for expert operations.
+   * **environment-expert-admin**: Users with this role can do anything in a certain environment.
+
+2. Global admin role: A user with this role can execute any API endpoint on the Inmanta server.
+
+The default policy makes the following assumptions about the content of the access token:
+
+* The ``sub`` claim is present and contains the username of the user the token is for.
+* Environment-scoped roles are defined in the ``urn:inmanta:roles`` claim of the access token. The value must be a dictionary that maps the ID of the environment to a list of roles the user has in that environment.
+* Global admins must have the claim ``urn:inmanta:is_admin`` in the access token with the value set to ``true``.
```
sanderr, line 155: Is this relevant to the end user? This is something the server constructs, isn't it?

If put in another subsection, in the light of "this is what you may assume when writing an access policy", I could perhaps see its value, but here it feels off to me.

docs/administrators/authorization.rst
```diff
@@ -0,0 +1,275 @@
...
+
+Integration with database authentication
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The default policy integrates seamlessly with database authentication.
+
+* The admin user created using the ``/opt/inmanta/bin/inmanta-initial-user-setup`` command will have the ``urn:inmanta:is_admin`` claim set to true.
+* The web-console provides support to manage users and role assignments.
+
+If the policy contains a ``roles`` variable that contains a list role names, the Inmanta server will create these roles automatically when the server starts. The default policy defines this variable for the roles it can handle. Like that there is no need to create roles by hand. The snippet below provides an example. Removing a role from this list will not remove that role when the server starts. That should be done manually.
```
sanderr, line 166:
```suggestion
If the policy contains a ``roles`` variable that contains a list role names, the Inmanta server will create these roles automatically when the server starts. The default policy defines this variable for the roles it can handle. That way there is no need to create roles by hand. The snippet below provides an example. Removing a role from this list will not remove that role when the server starts. That should be done manually.
```

docs/administrators/authorization.rst
sanderr: After reading on I see how it's relevant. I think it would be valuable to cross-link the two:
- Here something like "... about the content of the access token. This has to be configured correctly when using third party authorization providers:"
- In the section about enabling the policy engine, in point 3, something like "see <this section> for the claims required by the default policy."

docs/administrators/authorization.rst
```diff
@@ -0,0 +1,275 @@
...
+Writing a custom access policy
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This section provides some guidance on how to test/debug your own access policy.
+
+
+Troubleshooting: Policy engine fails to start
+"""""""""""""""""""""""""""""""""""""""""""""
+
+When the access policy contains a syntax error, the policy engine will fail to start and as such also the Inmanta server. In that case, the ``server.log`` file will indicate that the policy engine failed to start with a reference to the log file of the policy engine. The latter log file will contain more information about the specific issue. If required, the log level of the policy engine can be increased using the :inmanta.config:option:`policy_engine.log-level` config option.
```
sanderr, line 204: I hadn't noticed this in review. But why is this a separate option, rather than part of the generic logging config?

docs/administrators/authorization.rst
```diff
@@ -0,0 +1,275 @@
...
+    default allow := false
+    ...
+
+
+Enable the policy-engine authorization provider
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Set the :inmanta.config:option:`server.authorization-provider` config option to ``policy-engine``.
+2. Modify the default policy at ``/etc/inmanta/authorization/policy.rego`` if desired. More information about writing your own access policy is available :ref:`here<authorization-writing-own-policy>`.
+3. If a 3rd party auth broker is used, it must be configured to correctly set all claims required by the access policy.
```
sanderr, line 189: What is the consequence if these are not set? I assume the policy engine just fails? It wouldn't let things slip through, would it?

sanderr (changes requested):
Nice. Very clear!

docs/administrators/authorization.rst
arnaudsjs: If a certain claim is not set in the token, a lookup for that value in the policy will evaluate to `undefined`. Anything in the policy that depends on an undefined value will also evaluate to undefined. Undefined is a falsy value, so the allow rule will evaluate to False and reject the request.

PR #9304, opened 2025-07-02
fixed test for updated std plugin

sanderr:
The select plugin was updated to also accept references, which changes the error message. This PR uses another plugin that should be more stable.


wouterdb (approved):

jptrindade:
This will need to go into iso8 as well, so a changelog would be helpful.

Can you fix the send_events assertion as well?

PR #9331, opened 2025-07-09
fixed final module install tests

sanderr:
# Description

Fixed minor fallout from the module install refactor.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/tests/moduletool/test_install.py
+++ b/tests/moduletool/test_install.py
@@ -1187,14 +1187,14 @@ def test_constraints_logging_v2(modules_v2_dir, tmpdir, caplog, snippetcompiler_
     )
 
     expected_log_messages = [
-        "Installing module module_a (v2) (with constraints module_a<9.9.9 module_a>1.1.1 module_a<10.10.10 module_a>=0.0.0).",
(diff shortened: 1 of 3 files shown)


tests/moduletool/test_install.py
```diff
@@ -1187,14 +1187,14 @@ def test_constraints_logging_v2(modules_v2_dir, tmpdir, caplog, snippetcompiler_
     )
 
     expected_log_messages = [
-        "Installing module module_a (v2) (with constraints module_a<9.9.9 module_a>1.1.1 module_a<10.10.10 module_a>=0.0.0).",
-        "Installing module module_b (v2) (with no version constraints)",
+        "inmanta-module-module-a<9.9.9",
+        "inmanta-module-module-a>1.1.1",
```
sanderr, line 1191: This is slightly less informative, but I don't think it can be helped. The constraints that are still logged are external in the sense that core applies them. They come from the project. The constraints that are no longer logged come from other modules that are being installed in this same `pip install`. We have no choice but to rely on pip's logging there imo.

PR #9333, opened 2025-07-09
Cleanup notification for removed compile report (iso8)

### LINKED ISSUE #9092: clean up notifications for cleaned up compile reports
1. when compile report is deleted, all notifications that reference it should be deleted -> add index if not present
2. make sure notifications are cleaned up by themselves, with a configurable timeout (env setting, default 365 days)

arnaudsjs:
# Description

**Same PR as https://github.com/inmanta/inmanta-core/pull/9318 but applied on the iso8 branch due to a merge conflict.**

* This PR makes sure that all notifications, associated with a compile report, are removed when that compile report is removed.
* Point 2 of the ticket was already implemented in the `notificationservice.py`.

closes #9092 

# Self Check:

- [x] Changelog entry


tests/db/migration_tests/dumps/v202504040.sql
jptrindade, line 1: Why is this being changed?

jptrindade (commented):
Just the comment regarding the dump that changed

tests/db/migration_tests/dumps/v202504040.sql
arnaudsjs: Because this dump is the input for the test `tests/db/migration_tests/test_v202504040_to_v202507030.py`, which tests the migration script `v202507030.py`. The changes to the dump_tool update the `v202504040.sql` dump to make sure the necessary data required by the test is present in the database.

PR #9335, opened 2025-07-09
First bits

### LINKED ISSUE #9112: support for indexing list on dictpath/references
When serializing reference/matutors, we produce dictpaths of the form `a[0]` when a reference is in element `0` of list `a`

Dictpath doesn't support this currently

Make sure inmanta-lsm can not use it for relations. Probably best to implement this by adding a second dict-path parser with this extended capability, so that inmanta-lsm keeps using the old one.

wouterdb:
# Description

closes #9112 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/references.py
+++ b/src/inmanta/references.py
@@ -32,6 +32,7 @@ from pydantic import ValidationError
 
 import inmanta
 import inmanta.resources
+import jsonpath_ng
 import typing_extensions
 from inmanta import util
 from inmanta.types import JsonType, ResourceIdStr, StrictJson
@@ -190,8 +191,8 @@ class MutatedJsonArgument(Argument):
         start_value: JsonType = self.value
         for destination, valueref in self.references.items():
             value = valueref.get_arg_value(resource, logger)
-            dict_path_expr = dict_path.to_path(destination)
-            dict_path_expr.set_element(start_value, value)
+            jsonpath_expr = jsonpath_ng.parse(destination)
+            jsonpath_expr.update(start_value, value)
         return start_value
 
 
@@ -349,7 +350,7 @@ class ReferenceLike:
                     # The collector here is purely to collect the path/reference pairs
                     # The set of reference it collects will be discarded
                     # The root ReferenceCollector will traverse past this point as well to collect the actual reference
-                    cleaned_value = collector.collect_references(value, "")
+                    cleaned_value = collector.collect_references(value, "$")
                     try:
                         if collector.references:
                             arguments.append(
@@ -591,8 +592,8 @@ class ReplaceValue(Mutator):
 
     def run(self, logger: "handler.LoggerABC") -> None:
         value = self.resolve_other(self.value, logger)
-        dict_path_expr = dict_path.to_path(self.destination)
-        dict_path_expr.set_element(self.resource, value)
+        jsonpath_expr = jsonpath_ng.parse(self.destination)
+        jsonpath_expr.update(self.resource, value)
 
 
 @typing.runtime_checkable
--- a/src/inmanta/resources.py
+++ b/src/inmanta/resources.py
@@ -257,7 +257,7 @@ class ReferenceSubCollector:
 
             case dict() | proxy.DictProxy():
                 return {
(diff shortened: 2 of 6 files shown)


src/inmanta/references.py
```diff
@@ -591,8 +596,8 @@ def __init__(self, resource: "inmanta.resources.Resource", value: Reference[Prim
 
     def run(self, logger: "handler.LoggerABC") -> None:
         value = self.resolve_other(self.value, logger)
-        dict_path_expr = dict_path.to_path(self.destination)
-        dict_path_expr.set_element(self.resource, value)
```
sanderr, line 595: How is this safe? Mypy complains, and I'm inclined to agree with it.

src/inmanta/resources.py
```diff
@@ -257,7 +256,7 @@ def allow_references[T](v: T) -> T:
 
             case dict() | proxy.DictProxy():
                 return {
-                    key: self.collect_references(value, f"{path}.{dict_path.NormalValue(key).escape()}")
+                    key: self.collect_references(value, f"{path}.'{key.replace("'", "\\'")}'")
```
sanderr, line 259:
```suggestion
                    key: self.collect_references(value, f"{path}.'{key.replace("'", r"\'")}'")
```

Also, quite the nesting here: ' inside "" inside '' inside "". But if it works I'm fine with it.

src/inmanta/resources.py
```diff
@@ -577,13 +576,21 @@ def __init__(self, _id: "Id") -> None:
         self._references: dict[uuid.UUID, references.Reference[references.RefValue]] = {}
         self._resolved = False
 
+    def get(self, key: str, default: object = None) -> object:
+        if key in self.fields:
+            return getattr(self, key)
+        return default
```
sanderr, line 582: This is how you make it work? Can't we convert to a dict and back? Or at least implement the full `MutableMapping` protocol instead of `get` and `__contains__`?
That might also get rid of the mypy error. I tried to verify but I get `Skipping analyzing "jsonpath_ng": module is installed, but missing library stubs or py.typed marker`. Did you install a custom types package that you forgot to add to dev requires?

sanderr (changes requested):

src/inmanta/references.py
```diff
@@ -190,8 +190,13 @@ def get_arg_value(
         start_value: JsonType = self.value
         for destination, valueref in self.references.items():
             value = valueref.get_arg_value(resource, logger)
-            dict_path_expr = dict_path.to_path(destination)
-            dict_path_expr.set_element(start_value, value)
+
+            # backward compat between 8.2 and higher
+            # jsonpath_ng does not allow starting with ., dictpath does
+            if destination.startswith("."):
+                destination = "$" + destination
```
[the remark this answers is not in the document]
wouterdb: I think a warning will create more confusion:
1. it is perfectly harmless (i.e. this fix works)
2. It can only happen when using a version compiled with an 8.2 compiler on a newer orchestrator, so I expect this to fade out fast
3. It will only be visible in the handler, so the warning appears in a place where it is not expected (you typically look at these logs to see why a deployment failed, this has no relevance for that)

stubs/jsonpath_ng/jsonpath.pyi
```diff
@@ -0,0 +1,11 @@
+from pydantic import JsonValue
```
sanderr, line 1: Feels a bit weird to me to pull in pydantic for this, but for vendored stubs I guess it suffices.

stubs/jsonpath_ng/jsonpath.pyi
```diff
@@ -0,0 +1,11 @@
...
+
+
+class JSONPath:
+    def find(self, data: JsonValue) -> JsonValue: ...
+    def find_or_create(self, data: JsonValue) -> JsonValue: ...
+    def update(self, data: JsonValue, val: object) -> JsonValue: ...
+    def update_or_create(self, data: JsonValue, val: object) -> JsonValue: ...
```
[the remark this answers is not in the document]
wouterdb: or the type of val.

sanderr (approved):

PR #9336, opened 2025-07-10
fix migration script

Hugo-Inmanta:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry


tests/db/migration_tests/test_v202506160_to_v202507030.py
```diff
@@ -35,15 +35,16 @@ async def test_foreign_key_notification_to_compile(
...
-    assert result[0].uri == f"/api/v2/compilereport/{compile_id}"
-    assert result[0].compile_id is None
+    assert len(result) == 2
+    assert all(r.compile_id is None for r in result)
 
     # Run migration script
     await migrate_db_from()
 
+    # * One notification was removed because the associated compile not longer exists.
+    # * The other notification was update to reference the compile_id using a foreign key.
```
Hugo-Inmanta, line 46:
```suggestion
    # * The other notification was updated to reference the compile_id using a foreign key.
```

Hugo-Inmanta (commented):
Can't approve or request changes (since I'm the original author I think ??)

I didn't add the change entry, this should also go into iso8 I think.

wouterdb (approved):

PR #9350, opened 2025-07-10
Introduce resource set table

### LINKED ISSUE #9364: Introduce ResourceSet table
Introduce a ResourceSet table according to the design:

jptrindade:
# Description

Introduced resource_set table and its relational table to configuration model.

When we do a `put_version`  we will:
- Create new resource set instances (revisions) for each resource set
- Link the newly created resource sets to the corresponding configuration model

When we do a `put_partial` we will:
- Create new resource set instances (revisions) for each resource set that was changed
- Link the appropriate resource sets to the corresponding configuration model:
    - The resource set instances that we just created
    - The resource sets that are present in the configuration model but did not suffer changes in the latest `put_partial`

Planning discussion: https://docs.google.com/presentation/d/1aB-3CcawyH0DSPxZ2QiNU8MGfHbo2gfKN5EWp4y4qVo/edit?usp=sharing

image

closes #9364

# Self Check:

- [x] Changelog entry


src/inmanta/server/services/orchestrationservice.py
```diff
@@ -943,7 +953,17 @@ async def _put_version(
                         raise BadRequest(msg)
                     all_ids |= {Id.parse_id(rid, version) for rid in rids_unchanged_resource_sets.keys()}
 
-                await data.Resource.insert_many(list(rid_to_resource.values()), connection=connection)
+                updated_resources = list(rid_to_resource.values())
+                await data.Resource.insert_many(updated_resources, connection=connection)
+                # bump all resource sets if we are doing a full compile otherwise, bump only the updated resource sets
+                if not is_partial_update:
+                    updated_resource_sets = {r.resource_set if r.resource_set is not None else "" for r in updated_resources}
```
arnaudsjs, line 960: Up to now, this method (and the rest of the code iirc) used `None` to indicate the default resource set. With this new implementation this method uses a mix of two styles: 1) None for the default resource set or 2) the empty string for the default resource set. I think we have to keep it consistent, otherwise we will easily get bugs. If we use a different style in the db compared to the code, that's fine for me. But in the code we should try to keep it consistent.

In fact a bug like this is already introduced on the `resource_sets` variable. This dict uses None to refer to the default resource set. The new code you added at the beginning of this method uses the empty string.

arnaudsjs, line 960: Shouldn't this be:
```suggestion
                if not is_partial_update:
                    updated_resource_sets = {r if r is not None else "" for r in resource_sets.values()}
```

arnaudsjs (changes requested):
I went through everything, except for the tests. I will look at the tests tomorrow.

src/inmanta/server/services/orchestrationservice.py
jptrindade: I believe `resource_sets.values()` will not contain the default resource set

src/inmanta/db/versions/v202507040.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    Create resource_set table
+    """
+    schema = """
+    CREATE TABLE public.resource_set (
+        name character varying NOT NULL,
+        environment uuid NOT NULL,
+        model integer NOT NULL,
+        revision integer NOT NULL,
+        PRIMARY KEY (name, environment, model, revision),
```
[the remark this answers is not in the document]
jptrindade: I'll order it by (environment,model,name,revision)

src/inmanta/server/services/orchestrationservice.py
jptrindade: I agree that it is confusing and I would prefer not to have this None/empty string issue. I think I do not have a bug here.
In both cases I iterate over a list of resources so I have to make the conversion from None to ""

src/inmanta/data/__init__.py
```diff
@@ -4688,6 +4688,121 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+                revision,
+                model
+            )(
+                SELECT
+                    rs.name,
+                    rs.environment,
+                    rs.revision,
+                    $3
+                FROM {cls.table_name()} AS rs
+                WHERE rs.environment=$1 AND rs.model=$2 AND NOT rs.name=ANY($4)
```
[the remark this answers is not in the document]
jptrindade: These 2 expressions don't mean the same thing.
`NOT rs.name=ANY($4)` means that there is no element in `$4` that is equal to `rs.name`

`rs.name!=ANY($4)` means  that `rs.name` is not equal to at least one element in `$4`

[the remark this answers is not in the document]
arnaudsjs: Oh. I didn't know that. Sorry for that.

src/inmanta/db/versions/v202507040.py
```diff
@@ -0,0 +1,56 @@
...
+        PRIMARY KEY (environment, name, model, revision),
+        FOREIGN KEY (environment, model)
+            REFERENCES public.configurationmodel(environment, version) ON DELETE CASCADE
+    );
+
+    INSERT INTO public.resource_set (environment, model, name, revision)
+    SELECT DISTINCT
+        r.environment,
+        r.model,
+        COALESCE(r.resource_set, '') AS name,
```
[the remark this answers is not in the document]
sanderr: If we see another way to represent it, I'm all for it.

That said, I think we can consider the std module to be the only intended user of this API. Now that we've noticed this discrepancy, I think we should add a check on the API though. It's a technically breaking change, but I think we can get away with it considering that it lives outside of the intended scope.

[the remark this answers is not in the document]
arnaudsjs: > If we see another way to represent it, I'm all for it.

Me too.

> I think we should add a check on the API though. It's a technically breaking change, but I think we can get away with it considering that it lives outside of the intended scope.

I agree.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -21,7 +21,7 @@
 import uuid
 from collections import abc, defaultdict
 from collections.abc import Mapping, Sequence
-from typing import Literal, Optional, cast
+from typing import Literal, Optional, Set, cast
```
sanderr, line 24: This should be imported from `collections.abc`. For most types the one from `typing` is just a deprecated alias, but `typing.Set` is even an alias to `set`, rather than `collections.abc.Set`, so it's even more important to use the correct one.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -787,8 +810,8 @@ async def _put_version(
         version: int,
         rid_to_resource: dict[ResourceIdStr, data.Resource],
         unknowns: abc.Sequence[data.UnknownParameter],
+        resource_sets: dict[ResourceIdStr, data.ResourceSet],
```
sanderr, line 813: I'm not sure about this, and the resulting changes to this method. I think the whole insert and link could be done in a single database call (multiple insert queries: insert sets, insert resources, insert model). If we do it like that, we also don't need `ResourceSet` objects here, just the names.

@arnaudsjs wdyt?

src/inmanta/server/services/orchestrationservice.py
sanderr, line 1: The PR doesn't get rid of the copy-from-old-version writes for resources (`put_version`). I know dropping the old writes is part of a next stage, but I'm honestly not sure if that works here. If you copy unchanged resources, you'd also copy their resource set id, which means that their resource set would end up with two (or more) identical resources.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -560,12 +568,16 @@ def _create_dao_resources_from_api_resources(
         resource_state: dict[ResourceIdStr, Literal[ResourceState.available, ResourceState.undefined]],
         resource_sets: dict[ResourceIdStr, Optional[str]],
         set_version: Optional[int] = None,
-    ) -> dict[ResourceIdStr, data.Resource]:
+    ) -> tuple[dict[ResourceIdStr, data.Resource], dict[ResourceIdStr, data.ResourceSet]]:
```
sanderr, line 571: Note to self: review this method, depending on how the other discussion evolves.

src/inmanta/data/model.py
```diff
@@ -362,6 +362,7 @@ class Resource(BaseModel):
     attributes: JsonType
     is_undefined: bool
     resource_set: Optional[str] = None
+    resource_set_id: uuid.UUID
```
sanderr, line 365: I don't think we should expose this here.

src/inmanta/data/__init__.py
```diff
@@ -4688,6 +4688,130 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+    @classmethod
+    def table_name(cls) -> str:
+        return "resource_set"
+
+    @classmethod
+    async def get_resource_sets_in_version(
+        cls, environment: uuid.UUID, version: int, connection: Optional[asyncpg.connection.Connection] = None
+    ) -> list["ResourceSet"]:
+        """
+        Returns the resource sets in the given version.
```
sanderr, line 4713:
```suggestion
        Returns the resource sets in the given version. Only meant for testing.
```

sanderr (changes requested):

src/inmanta/db/versions/v202507040.py
```diff
@@ -0,0 +1,98 @@
...
+async def update(connection: Connection) -> None:
+    """
+    Create resource_set table
+    """
+    schema = """
+    -- add and populate resource_set table --
+
+    CREATE TABLE public.resource_set (
+        environment uuid NOT NULL,
+        id uuid NOT NULL,
```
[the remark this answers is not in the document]
jptrindade: We can add the `DEFAULT gen_random_uuid()` but I still think we need to generate the id below, because we need a way to link the specific id to a `res_set_name + model` when populating the `Resource` table

src/inmanta/db/versions/v202507040.py
```diff
@@ -0,0 +1,98 @@
...
+
+    -- relational table between resource set and configuration model
+
+    CREATE TABLE public.resource_set_configuration_model (
+      environment uuid NOT NULL,
+      model integer NOT NULL,
+      resource_set_id uuid NOT NULL,
+      PRIMARY KEY (environment, model, resource_set_id),
+      FOREIGN KEY (environment, model) REFERENCES public.configurationmodel(environment, version) ON DELETE CASCADE,
+      FOREIGN KEY (environment, resource_set_id) REFERENCES public.resource_set(environment, id) ON DELETE CASCADE
```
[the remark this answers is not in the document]
jptrindade: But if a ResourceSet ends up being removed, we also want to clean its entries in this table right? Or are you saying we only expect them to be deleted in cleanup so we can have only the cascade deletion on the model?

src/inmanta/db/versions/v202507040.py
```diff
@@ -0,0 +1,99 @@
...
+    UPDATE public.resource r
+    SET resource_set_id=us.id
+    FROM temp_unique_sets_with_id us
+    WHERE
+        r.environment=us.environment AND
+        r.resource_set IS NOT DISTINCT FROM us.resource_set AND
+        r.model=us.model;
+
+    ALTER TABLE public.resource
+    ADD CONSTRAINT resource_resource_set_id_fkey
```
arnaudsjs, line 76: We need to follow the naming convention.
```suggestion
    ADD CONSTRAINT resource_resource_set_id_environment_fkey
```

src/inmanta/db/versions/v202507040.py
```diff
@@ -0,0 +1,99 @@
...
+        FOREIGN KEY (resource_set_id, environment) REFERENCES public.resource_set(id, environment) ON DELETE CASCADE;
```
arnaudsjs, line 77: We also need an index over `(resource_set_id, environment)`

src/inmanta/db/versions/v202507040.py
```diff
@@ -0,0 +1,99 @@
...
+
+    -- relational table between resource set and configuration model
+
+    CREATE TABLE public.resource_set_configuration_model (
+      environment uuid NOT NULL,
+      model integer NOT NULL,
+      resource_set_id uuid NOT NULL,
+      PRIMARY KEY (environment, model, resource_set_id),
+      FOREIGN KEY (environment, model) REFERENCES public.configurationmodel(environment, version) ON DELETE CASCADE,
+      FOREIGN KEY (environment, resource_set_id) REFERENCES public.resource_set(environment, id) ON DELETE CASCADE
```
arnaudsjs, line 88: We need an index on `(environment, resource_set_id)`

src/inmanta/db/versions/v202507040.py
```diff
@@ -0,0 +1,99 @@
...
+    );
+
+    UPDATE public.resource_set_configuration_model rscm
+    SET environment=us.environment,
+        model=us.model,
+        resource_set_id=us.id
+    FROM temp_unique_sets_with_id us;
```
arnaudsjs, line 95: This is not correct, is it? Shouldn't we use an `INSERT` clause instead? We are probably missing some test coverage here.

tests/db/migration_tests/test_v202507030_to_v202507040.py
```diff
@@ -0,0 +1,48 @@
...
+import asyncpg
+import pytest
+
+from inmanta import data
+
+file_name_regex = re.compile("test_v([0-9]{9})_to_v[0-9]{9}")
+part = file_name_regex.match(__name__)[1]
+
+
+@pytest.mark.db_restore_dump(os.path.join(os.path.dirname(__file__), f"dumps/v{part}.sql"))
```
arnaudsjs, line 32: We need assertions on the content of the `resource_set_configuration_model` table here.

src/inmanta/server/services/orchestrationservice.py
sanderr: Oh, in that case you should use built-in `set`.

sanderr: Or `collections.abc.MutableSet`, but I see little benefit to that one (in this case).

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: I agree.

arnaudsjs (changes requested):
Only reviewed to code up to now. Not the tests

src/inmanta/server/services/orchestrationservice.py
jptrindade: Discussed in standup: to be tackled in the next PR

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -129,9 +132,10 @@ def ensure_no_cross_resource_set_dependencies(self) -> None:
             # It's sufficient to only check the requires relationship. The provides
             # relationship is always a subset of the reverse relationship (provides relationship).
             # The provides relationship only contains the cross-agent dependencies.
-            for req in res.get_requires():
+            requires = list(res.resource.attributes["requires"]) if "requires" in res.resource.attributes else {}
```
sanderr, line 135: `dict.get()` could help here

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -560,12 +583,16 @@ def _create_dao_resources_from_api_resources(
...
+    ) -> dict[ResourceIdStr, ResourceWithProvides]:
         """
-        This method converts the resources sent to the put_version or put_partial endpoint to dao Resource objects.
+        This method converts the resources sent to the put_version or put_partial endpoint
+            to dao Resource and ResourceSet objects.
         The resulting resource objects will have their provides set up correctly for cross agent dependencies
         and the version field of these resources will be set to set_version if provided.
 
+        We will create the ResourceSet objects here as well, one for each different resource set name on the resource list.
+        A uuid is generated for each one.
```
sanderr, line 594: No longer accurate (assuming we keep the dataclass change)

src/inmanta/data/__init__.py
```diff
@@ -4815,6 +4815,299 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+                resource_set_name = str(record["name"])
+                if resource_id not in rid_to_resource_sets:
+                    rid_to_resource_sets[resource_id] = {}
+                key = "new" if resource_set_name in updated_resource_sets else "old"
+                if key in rid_to_resource_sets[resource_id]:
+                    raise BadRequest(
+                        f"Resource set with name {resource_set_name} appears more than once on version {version} "
+                        f"of the model with ids {rid_to_resource_sets[resource_id]} and {resource_id}"
+                    )
+                rid_to_resource_sets[resource_id][key] = resource_set_name
```
wouterdb, line 4918: I'm not particularly fond of the old/new keys in a dict. But given that it is contained to this method, I can live with it.

src/inmanta/data/__init__.py
```diff
@@ -4815,6 +4815,299 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+            return "<SHARED>"
+        return native
+
+    @classmethod
+    async def validate_resource_sets_in_version(
+        cls,
+        environment: uuid.UUID,
+        version: int,
+        updated_resource_sets: list[str | None],
+        connection: Optional[asyncpg.connection.Connection] = None,
```
sanderr, line 4874: Probably doesn't make sense to make connection optional here. Also, make it a kw-only arg.

src/inmanta/data/__init__.py
```diff
@@ -4815,6 +4815,293 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+            attribute_hashes,
+            resource_types,
+            resource_id_values,
+            is_undefined,
+            resource_sets,
+        ]
+        if is_partial_update:
+            values.extend([base_version, resource_set_names_not_to_bump])
+
+        query = f"""
```
wouterdb, line 5011: What is the reason for making it one query?

src/inmanta/data/__init__.py
```diff
@@ -4815,6 +4815,293 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+        connection: Optional[asyncpg.connection.Connection] = None,
+    ) -> None:
+        """
+        Inserts resources and resource sets. These resource sets are linked to the target version
+        :param environment: The environment of these resources.
+        :param target_version: The version which we want to link the resource sets to
+        :param updated_resources: A list of resources to insert.
+            On a full compile, this list should contain all resources present in the version.
+            On a partial compile, this list should contain all resources belonging to a resource set that was changed.
+        :param updated_resource_sets: A list of resource sets that were changed.
```
wouterdb, line 4945: What is this for, if we also have `resource_set_names_not_to_bump` ?

src/inmanta/data/__init__.py
```diff
@@ -4815,6 +4815,293 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+        :param base_version: This is the version which the partial compile is based on. None if we are doing a full compile.
+        :param resource_set_names_not_to_bump: These are the resource set names from the base version .
+            which we do not wish to bump. Either because these sets were removed in this partial compile or because
+            these sets received changes. Not applicable for a full compile.
```
wouterdb, line 4949: How is the default resource set handled? 

If we have a base version, it should always be copied, if not, then not. 

Couldn't hardcoding this simplify the queries?

src/inmanta/data/__init__.py
jptrindade: `resource_set_names_not_to_bump` includes updated_resource_sets and deleted_resource_sets

src/inmanta/data/__init__.py
jptrindade: The main benefit of having it all in one query is not having to set the resource set uuids in python and managing them.
In this query we have to manage the resource sets that we inserted and link them to the correct resources and configuration models.

src/inmanta/data/__init__.py
```diff
@@ -4815,6 +4815,293 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+            assert r.environment == environment
+            assert r.model == target_version
+            resource_ids.append(r.resource_id)
+            agents.append(r.agent)
+            attributes.append(json.dumps(r.attributes))
+            attribute_hashes.append(util.make_attribute_hash(r.resource_id, r.attributes))
+            resource_types.append(r.resource_type)
+            resource_id_values.append(r.resource_id_value)
+            is_undefined.append(r.is_undefined)
+            resource_sets.append(r.resource_set)
```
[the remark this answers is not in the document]
jptrindade: Hum, I was not familiar with that. I also did not like the way that this turned out, but wanted something that worked for now before trying to make it look cleaner. I'll switch it to use structs

src/inmanta/data/__init__.py
```diff
@@ -4815,6 +4815,299 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+                ON rscm.environment=rs.environment
+                AND rscm.resource_set_id=rs.id
+            WHERE rscm.environment=$1
+                AND rscm.model=$12
+                AND NOT EXISTS (
+                    -- extra hoop to make comparison with null names --
+                      SELECT 1
+                      FROM UNNEST($13::text[]) AS outdated_resource_set_names(name)
+                      WHERE outdated_resource_set_names.name IS NOT DISTINCT FROM rs.name
+                    )
```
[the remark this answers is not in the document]
sanderr: Because it is shared:
1. we can not modify it because it is part of an older version
2. we have to be able to extend it with shared resources from this export

I guess in rare cases a partial export may not contain any shared resources. But I think we can keep it simple and always create a new shared set for each model version. We can check with @wouterdb what he thinks of this.

Come to think of it, I'm not sure your current implementation extends the shared set. I will double check but I don't recall seeing it.

src/inmanta/data/__init__.py
```diff
@@ -4930,166 +4930,186 @@ async def insert_sets_and_resources(
         environment: uuid.UUID,
         target_version: int,
         updated_resources: list[m.Resource],
-        updated_resource_sets: list[str | None],
+        updated_resource_sets: set[str | None],
```
jptrindade, line 4933: I could calculate this from the list of `updated_resources` . Do you think it makes sense to iterate over the resource list to fetch these resource_sets or can I just leave it as an argument?

src/inmanta/data/__init__.py
```diff
@@ -4930,166 +4929,188 @@ async def insert_sets_and_resources(
...
+        Inserts resources and resource sets.
+        Links resource sets to the target version.
+        In case of a full compile, we expect to receive every resource set,
+            so we insert those and link them to the target version
+        In case of a partial compile, we insert the updated resource sets and we link to the target version:
+            - The resource sets that we just updated
+            - Every resource set in the base version except:
+                - Resource sets we want to delete
+                - Resource sets with the same name as one of the updated resource sets (they are now outdated)
+        This method expects the full set of resources for any resource set present in updated_resource_sets.
```
sanderr, line 4946: This still mentions `updated_resource_sets`.

src/inmanta/data/__init__.py
```diff
@@ -4930,166 +4929,188 @@ async def insert_sets_and_resources(
         environment: uuid.UUID,
         target_version: int,
         updated_resources: list[m.Resource],
-        updated_resource_sets: list[str | None],
         base_version: Optional[int] = None,
-        resource_set_names_not_to_bump: Optional[Set[str | None]] = None,
+        deleted_resource_sets: Optional[set[str]] = None,
```
sanderr, line 4933: This is bad practice imo. We should not modify it.

src/inmanta/data/__init__.py
```diff
@@ -4930,166 +4929,188 @@ async def insert_sets_and_resources(
...
+            - Every resource set in the base version except:
+                - Resource sets we want to delete
+                - Resource sets with the same name as one of the updated resource sets (they are now outdated)
+        This method expects the full set of resources for any resource set present in updated_resource_sets.
+        The shared resource set is treated as any other.
         :param environment: The environment of these resources.
         :param target_version: The version which we want to link the resource sets to
         :param updated_resources: A list of resources to insert.
             On a full compile, this list should contain all resources present in the version.
             On a partial compile, this list should contain all resources belonging to a resource set that was changed.
```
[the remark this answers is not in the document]
jptrindade: But this is also the same for every other resource set, right?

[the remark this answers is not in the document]
sanderr: Yes, but like I said before, the shared set *is* a special case. And if you're not explicit about something like that, it will always be ambiguous, because the description conflicts with the knowledge in the reader's head. If you make it explicit, it's clear that as far as this method is concerned, the shared set is not special.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -894,40 +909,16 @@ async def _put_version(
             except asyncpg.exceptions.UniqueViolationError:
                 raise ServerError("The given version is already defined. Versions should be unique.")
 
-            with tracing.span("put_version.partial"):
```
jptrindade, line 897: This was lost. Is it important?

wouterdb (approved):
I did not go super deep, but looks good overall

src/inmanta/data/__init__.py
```diff
@@ -5522,10 +5835,10 @@ def to_dict(self) -> dict[str, object]:
         self.__mangle_dict(dct)
         return dct
 
-    def to_dto(self) -> m.Resource:
+    def to_dto(self, include_resource_version_in_requires: bool = True) -> m.Resource:
```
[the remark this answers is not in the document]
jptrindade: These methods expected DAOs before, but now expect DTOs, but  `Resource.to_dto()` adds this version information to the requires ids that was not on the DAO before. It ends up messing with the `PartialUpdateMerger`'s way of managing requires dependencies.
 I think the APIs expect the DTO to have this version information on the ids of the attributes so that is why I did not remove it.

src/inmanta/data/__init__.py
```diff
@@ -5512,7 +5825,7 @@ def get_requires(self) -> abc.Sequence[ResourceIdStr]:
         """
         Returns the content of the requires field in the attributes.
         """
-        if "requires" not in self.attributes:
+        if "requires" not in self.attributes or not isinstance(self.attributes["requires"], list):
```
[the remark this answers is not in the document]
jptrindade: It seems weird, I don't remember if it was because of mypy or if I encountered it as a dict.
Either way I'm not using this method anymore so I'll revert this

src/inmanta/data/__init__.py
```diff
@@ -4956,129 +4955,110 @@ async def insert_sets_and_resources(
...
-
-        values = [environment, target_version, inserted_resource_set_ids]
+            resource_data["resource_id"].append(str(r.resource_id))
+            resource_data["resource_type"].append(r.resource_type)
+            resource_data["resource_id_value"].append(r.resource_id_value)
+            resource_data["agent"].append(r.agent)
+            resource_data["attributes"].append(r.attributes)
+            resource_data["attribute_hash"].append(util.make_attribute_hash(r.resource_id, r.attributes))
+            resource_data["is_undefined"].append(r.is_undefined)
+            resource_data["resource_set"].append(r.resource_set)
```
wouterdb, line 4974: I would replace the dict with a number of lists 

```python
resource_ids: list[str] = []
resource_type: list[str] = [] 
```

a bit more typed 

(but a detail)

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -310,7 +305,7 @@ def _merge_shared_resources(self, shared_resources_new: dict[ResourceIdStr, data
             else:
                 # Old shared resource not referenced by partial compile
                 res_old = self.shared_resources_old[rid_shared_resource]
-                res = res_old.copy_for_partial_compile(new_version=self.version)
+                res = res_old.model_copy(update={"model": self.version}, deep=True)
```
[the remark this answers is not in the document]
wouterdb: but on this code path, we don't care about it, right?

src/inmanta/data/__init__.py
sanderr: On the other hand it makes the `cls._get_value(...)` quite a bit more verbose.

wouterdb: get_value is I think a relic from the past. 

In this particular case, it adds nothing, all types used are trivially understood by asyncpg.

And in general, I think we should add [custom type conversion](https://magicstack.github.io/asyncpg/current/usage.html#custom-type-conversions) to asyncpg instead 

and it adds no security, as it just defaults to return value.

jptrindade:
1. drop resource.resource_set? Why not?
- Not in the scope of this ticket, we want to determine first what we want the DTO to look like as presenting the resource_set id would not be helpful. If we still want to display the resource_set name in the DTO we would have to change the `Resource.to_dto` method to async which can have further ramifications. We can tackle this in another ticket since it is mostly independent
2. resource version ids in resource DTO's requires. I'm hoping we can afford to get rid of it. [Slack discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1755683511157329)
- Would be nice to get rid of it. Currently this PR has a workaround on the `to_dto` method to exclude the version ids with a flag, but it is quite ugly
3. model in resource DTOs. Still required? Which endpoints?
- Need to investigate which endpoints use it. I'm guessing that some of them we will be able to derive the model for the DTO. Don't know if that is the case for all
4. SELECT DISTINCT ON in migration script ( https://github.com/inmanta/inmanta-core/pull/9350#discussion_r2276547646)
- I was under the impression that this did not work for NULLs, so much that I purposely changed the query to handle NULL values specifically, but I'm seeing in @sanderr 's example that it works and in some documentation I found
6. deleting of resource sets on model delete
 - Will do it on this PR
7. ~~separate resource set table vs include name in rscm table~~ discussed and decided
8. ~~jsonb has maximum size~~ fixed by using dedicated arrays instead of a single jsonb
9. regenerate db dump
 - Will do it on this PR
10. drop model column from resource table
- [To be tackled in the next PR](https://github.com/inmanta/inmanta-core/pull/9389), it already removes checks for the model from the queries so I can drop the model column there to make sure that there is nothing missing

tests/db/migration_tests/test_v202507030_to_v202507040.py
```diff
@@ -0,0 +1,84 @@
...
+    resource_set_info = {(r.environment, r.resource_set_id, r.resource_set) for r in resources}
+    assert res_sets == resource_set_info
+    resource_set_names = {r.resource_set for r in resources}
+    assert None in resource_set_names
+    assert "set-a" in resource_set_names
+
+    records = await postgresql_client.fetch(
+        """
+        SELECT rs.name, rscm.* FROM public.resource_set_configuration_model AS rscm
+        LEFT JOIN public.resource_set AS rs
```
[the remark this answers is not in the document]
jptrindade: Don't quite remember, but I also don't think it makes sense

src/inmanta/data/__init__.py
```diff
@@ -4840,6 +4840,387 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+                rid_to_resource_sets[resource_id] = {}
+                for name in resource_set_names:
+                    key = "new" if name in updated_resource_sets else "old"
+                    if key in rid_to_resource_sets[resource_id]:
+                        # Should never be possible for a resource id to be present in either:
+                        # - 2 updated resource sets
+                        # - 2 unchanged resource sets
+                        # It means we have a bug somewhere
+                        raise Exception(
+                            f"Resource set with name {name} appears more than once on version {version} of the model."
```
arnaudsjs, line 4956: Why not mention the names of the resource set?

arnaudsjs, line 4956:
```suggestion
                            f"Resource set with name {name} appears more than once on version {version} of the model. "
```

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -4840,6 +4840,387 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+        if is_partial_update:
+            await cls.validate_resource_sets_in_version(
+                environment=environment,
+                version=target_version,
+                updated_resource_sets=updated_resource_sets,
+                connection=connection,
+            )
+
+    @classmethod
+    async def clear_resource_sets(
```
arnaudsjs, line 5156: Logically this query looks good to me. Something I noticed is that the locking order between the case where the version argument is set or not set is not the same, because the case with the version argument does an additional select query:

Locking order with version argument:

* resource_set_configuration_model
* resource_set
* resource

Without version argument it's:
* resource_set_configuration_model
* resource
* resource_set

We had issues with locking order in the past, so we try to keep it stable. But maybe this will not cause any issues in practice. It's hard to estimate.

src/inmanta/data/__init__.py
```diff
@@ -2879,7 +2879,7 @@ async def clear(self, connection: Optional[asyncpg.connection.Connection] = None
                 "DELETE FROM public.resourceaction_resource WHERE environment=$1", self.id, connection=con
             )
             await ResourceAction.delete_all(environment=self.id, connection=con)
-            await Resource.delete_all(environment=self.id, connection=con)
+            await ResourceSet.clear_resource_sets(environment=self.id, connection=con)
```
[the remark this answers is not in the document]
jptrindade: If I were to separate this into 2 different cases (full/version)
Would you keep this `clear_resource_sets_in_version` as the name of the versioned one
And `delete_all` for the full one?

src/inmanta/data/__init__.py
```diff
@@ -5129,6 +5152,74 @@ async def insert_sets_and_resources(
...
+            DELETE FROM resource_set_configuration_model AS rscm
+            WHERE rscm.environment=$1
+            RETURNING rscm.resource_set_id AS id
+        ),
+        """
+        )
+        query = f"""
+        {resource_sets_to_delete}
+        resources_to_delete AS (
+            DELETE FROM resource AS r
```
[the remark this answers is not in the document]
jptrindade: Deleting a resource set cascades and deletes the matching resources. But we want to avoid that right? That is why this method exist, to prevent locking in cascade, iirc.
Should I remove the `ON DELETE CASCADE` from the `resource_set_id` foreign key?

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
sanderr: No strong preference. Overall I'm not sure if `clear` is a great name, because we only ever really clear an environment (leave the environment alive but delete everything in it). And I think `delete_all` could still apply, and the scope of "all" just depends on the filter. I don't think you necessarily have to split it in two **methods**, I think one method with two **branches** would suffice. But if you prefer two methods, fine by me.

src/inmanta/data/__init__.py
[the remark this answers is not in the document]
sanderr: (I mainly need to double check if it's important that we are consistent with `INSERT` or only with reads / updates. If it's the latter, I don't think it matters much for resources / resource sets. I don't know atm.)

src/inmanta/data/__init__.py
```diff
@@ -5153,25 +5159,24 @@ async def insert_sets_and_resources(
...
 
         :param environment: The environment from which to delete the resource sets
         :param version: The version to delete from the resource_set_configuration_model table.
-            None if we want to delete every version (i.e. when clearing the environment).
         :param connection: The connection to use
         """
-        resource_sets_to_delete = (
-            """
+
+        query = """
```
sanderr, line 5179: For this query, isn't it as simple as below? You wouldn't need the `resource_sets_to_delete` at all.
```SQL
DELETE FROM resource_set AS rs
USING deleted_resource_set_versions AS del_rscm
WHERE
        rs.id = del_rscm.id
        AND NOT EXISTS (
                SELECT 1
                FROM resource_set_configuration_model AS rscm
                WHERE environment = $1 AND rscm.id = rs.id
        )
```

I may be missing something.

jptrindade: I believe that is also correct.

sanderr (approved):

PR #9362, opened 2025-07-14
Improve datastructure to represent role assignments.

arnaudsjs:
# Description

Change the format of the datastructure, used to return RoleAssignments on the API, to make it more usable on the web-console.

# Self Check:

- [x] Changelog entry


src/inmanta/data/__init__.py
```diff
@@ -6126,11 +6126,13 @@ async def list_users_with_roles(cls) -> list[m.UserWithRoles]:
...
             records_for_group = list(group_elem_iterator)
-            roles = [
-                m.RoleAssignment(environment=record["role_environment"], role=record["role_name"])
-                for record in records_for_group
-                if record["role_environment"] is not None and record["role_name"] is not None
-            ]
+            roles: dict[uuid.UUID, list[str]] = {}
+            for record in records_for_group:
+                if record["role_environment"] is not None and record["role_name"] is not None:
+                    if record["role_environment"] in roles:
```
sanderr, line 6132: a `defaultdict` might make this slightly more digestible. But it's not too bad either way, so not a strong requirement for me.

sanderr (approved):

PR #9405, opened 2025-08-04
Added the `inmanta project download` command

### LINKED ISSUE #7820: Make inmanta project install --source and inmanta module install --source
# Goal 

Make source of V2's more available to end users (similar to v1) 
It is currently very difficult for end-users to find what to import in larger generated modules. 
To make it more complicated, the packaged structure is different from the source structure, so unpacking a module doesn't make it usable... 

the user scenario is a model developer that composes existing modules and want to find the correct entities. 

# Idea

~Make an install of a project downloading all wheels, unpack them to the libs folder, convert them back to their source layout and and `pip install -e` them~

# Idea 2

Can this simply be pip down | tar -xz ?

# Conclusion

- feed back conclusion to solutions team as a proposal
- `module install` may not be the best name, considering that it used to be a command with different semantics
- start with only the module install tool: takes the name of a module as a requirement specifier, downloads it as tarball, unpacks in current directory (add `-d` option to override), transforms to source form, optionally (tbd) installs in editable mode
- when module tool has been reviewed, add same for a project -> installs all dependencies in source mode if possible. Pip download / project install to get all the right versions. Then try for all modules to install in source mode in libs dir.

arnaudsjs:
# Description

Added the `inmanta project download` command. This command allows downloading all Inmanta modules, required by a project, in their source format from a Python package repository.

closes #7820 

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation~~


src/inmanta/data/model.py
```diff
@@ -982,6 +983,52 @@ def has_source(self) -> bool:
...
+        if not self.use_system_config:
+            # If we don't use system config, unset env vars
+            if "PIP_EXTRA_INDEX_URL" in sub_env:
+                del sub_env["PIP_EXTRA_INDEX_URL"]
+            if "PIP_INDEX_URL" in sub_env:
+                del sub_env["PIP_INDEX_URL"]
+            if "PIP_PRE" in sub_env:
+                del sub_env["PIP_PRE"]
+            if "PIP_NO_INDEX" in sub_env:
+                del sub_env["PIP_NO_INDEX"]
```
jptrindade, line 1021:
```suggestion
for key in ("PIP_EXTRA_INDEX_URL", "PIP_INDEX_URL", "PIP_PRE", "PIP_NO_INDEX"):
    sub_env.pop(key, None)
```
We can use pop instead of del to avoid the if check, either way, I think iterating over the keys is cleaner

arnaudsjs: I just moved this method to a different class, so I didn't really evaluate the implementation. It's an improvement so I'll apply it anyway.

jptrindade (approved):

src/inmanta/moduletool.py
```diff
@@ -319,6 +320,41 @@ def parser_config(cls, parser: ArgumentParser, parent_parsers: abc.Sequence[Argu
...
+        )
+
+    def download(self, install: bool) -> None:
+        project = self.get_project()
+        downloadpath = project.downloadpath if project.downloadpath else os.path.join(project.path, "libs")
+        os.makedirs(downloadpath, exist_ok=True)
+        pip_config: model.PipConfig = project.metadata.pip
+        path_requirements_file = os.path.join(project.path, "requirements.txt")
+        if not os.path.exists(path_requirements_file):
+            raise Exception(f"Project requirements file not found at: {path_requirements_file}")
```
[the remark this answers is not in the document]
sanderr: On the other hand, the commmand is quite useless if the file is not there.

src/inmanta/util/__init__.py
```diff
@@ -1006,3 +1006,14 @@ def make_attribute_hash(resource_id: "ResourceId", attributes: Mapping[str, obje
...
+
+def get_inmanta_module_name(path_source_distribution_pkg: str) -> str:
+    """
+    Returns the name of the Inmanta module that belongs to the given python
+    source distribution package for an Inmanta module.
+    """
+    filename_source_distribution_pkg = os.path.basename(path_source_distribution_pkg)
+    assert filename_source_distribution_pkg.endswith("tar.gz")
+    python_package_name = filename_source_distribution_pkg.split("-", maxsplit=1)[0]
+    return python_package_name[len("inmanta_module_") :]
```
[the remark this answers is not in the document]
arnaudsjs: Cool! I didn't know that method.

src/inmanta/env.py
```diff
@@ -843,14 +847,35 @@ def get_installed_packages(self, only_editable: bool = False) -> dict[Normalized
...
+            requirements.extend(file_parser.RequirementsTxtParser.parse(path_requirements_file))
+        if pkg_requirement is not None:
+            requirements.append(pkg_requirement)
+        Pip.run_pip_download_command(
+            python_path=self.python_path,
+            output_dir=output_directory,
+            pip_config=pip_config,
+            pkg_requirement=pkg_requirement,
+            path_requirements_file=path_requirements_file,
+            no_deps=True,
```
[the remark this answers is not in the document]
arnaudsjs: Nice catch! I think the two stage approach should work.

src/inmanta/moduletool.py
```diff
@@ -1887,3 +1898,139 @@ def get_setup_cfg(self, in_folder: str, warn_on_merge: bool = False) -> configpa
...
+            for req in requirements_exact_version:
+                try:
+                    env.process_env.download_distributions(
+                        output_directory=download_dir,
+                        pip_config=pip_config,
+                        dependencies=[req],
+                        no_deps=True,
+                        # Setting no_binary to :all: would imply that `pip download` downloads dependencies
+                        # (e.g. setuptools) of these modules in source format as well. This would make the
+                        # command fail if these dependencies are not available in a source distribution package.
```
sanderr, line 2002: Good comment

src/inmanta/util/__init__.py
```diff
@@ -1006,3 +1006,25 @@ def make_attribute_hash(resource_id: "ResourceId", attributes: Mapping[str, obje
     m.update(resource_id.encode("utf-8"))
     m.update(character.encode("utf-8"))
     return m.hexdigest()
+
+
+def get_pkg_name_and_version(path_distribution_pkg: str) -> tuple[str, str]:
```
sanderr, line 1011: This seems to only handle tarballs, not wheels. Am I missing something?

PR #9467, opened 2025-08-25
Improvements for inmanta module/project download command

arnaudsjs:
# Description

Improvements for the `inmanta module download` and `inmanta project download` command:

* Performance improvement on `inmanta module download` command, by not using the two stage approach. There cannot by any version conflicts when downloading a single package.
* Override if the module already exists in the download directory.
* Fix in documentation.

Part of #7820

# Self Check:

- [x] Changelog entry


jptrindade (approved):

sanderr (changes requested):
Looks good to me overall. I have one optional suggestion that I think would make it even nicer.

Issue #9485, opened 2025-08-28
Add e2e deploy tests for partial versions


sanderr:
Up till recently, after `put_partial`, there was little difference between a full and a partial version because they were represented identically internally. With the recent changes to how partial versions are represented in the database, partials are inherently differrent, and are therefore treated differently by the resource scheduler. This means we need to add tests for e2e deploys, that make use of partial exports, to verify that the resource scheduler interprets this new format appropriately.

Scope: at least from `put_partial` to scheduler. Compiler and executors may be mocked out if desired.

Focus on good weather scenarios. Make sure to include at least one model with additions to the shared set and one without.

PR #9486, opened 2025-08-28
Made scheduler resource set aware

sanderr:
# Description

This PR updates the scheduler to read the new database structure as partial diffs. This should greatly improve the performance of scheduler updates.

I believe that the expertise of reviewers is quite distributed for this pull request. So I'm asking wide review, but scoped to specific parts:
- @arnaudsjs full review
- @jptrindade the database part. Concretely, `inmanta.data` and `tests/test_data.py`
- @wouterdb The scheduler part. Concretely, the `inmanta.deploy` namespace. You may also be interested (but not required for review) in `tests/test_data.py` because it closely resembles what I've been using for scaling testing, and it can be used for that purpose again with low effort.

Part #9364. Will be merged into the development branches as part of #9350.

Follow-up ticket: #9485

# Self Check:

- [x] ~Changelog entry~

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -2909,6 +2909,23 @@ async def test_get_partial_resources_since_version_raw(environment, server, post
     """
     Verify the behavior of the get_partial_resources_since_version_raw and get_resources_for_version_raw methods,
     used by the scheduler to apply partial and full versions respectively.
+
+    Setup:
+        - Generate `nb_resource_sets` sets with integers as names
+        - Each set contains `nb_resources_per_set` resources
+        - Resource ids are predictable: `...,id={resource_set}-{resource_index}`
+        - Resources contain a single attribute (other than required ones) `exported`: the version in which it was exported
+
+    Test approach:
+        for a few different `partial_size` sizes of partial:
+            - Choose `partial_size` resource sets, evenly distributed over all sets (e.g. sets 1, 11, 21, ...)
+            - Regenerate these sets as described in setup and do a `put_partial` and release
+            - Verify results of `get_partial_resources_since_version_raw` and `get_resources_for_version_raw`:
+                Verify that the methods return only the differing resource sets, and the appropriate resources
+                for those sets.
+
+        Verify behavior of the two methods for some special scenarios: multiple versions, non-released versions,
+        deleted sets, ...
     """
     nb_resource_sets: int = 100
     nb_resources_per_set: int = 100
@@ -2922,11 +2939,12 @@ async def test_get_partial_resources_since_version_raw(environment, server, post
         without_version = ResourceIdStr(f"mymodule::Myresource[myagent,id={resource_set}-{index}]")
         return ResourceVersionIdStr(f"{without_version},v={version}") if version is not None else without_version
 
-    async def release() -> None:
+    async def release(version: Optional[int] = None) -> None:
         """
         Mark all versions as released, without actually releasing it.
         """
-        await postgresql_client.execute("UPDATE configurationmodel SET released=true")
+        condition = f"WHERE version = {version}" if version is not None else ""
+        await postgresql_client.execute(f"UPDATE configurationmodel SET released=true {condition}")
 
     # set up initial state by releasing a single base version
     version: int = await client.reserve_version(tid=environment).value()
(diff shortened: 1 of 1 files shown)


src/inmanta/data/__init__.py
```diff
@@ -5507,123 +5515,254 @@ async def get_resources_for_version(
...
+            -- filter out all sets that remained unchanged
+            WHERE rs_new.resource_set_id IS DISTINCT FROM rs_old.resource_set_id
+            -- make sure SELECT DISTINCT matches the new id for the shared set, if it exists
+            ORDER BY rs_new.name, rs_old.name, rs_new.resource_set_id NULLS LAST
+        ) AS diff
+        ON true
+        LEFT JOIN {cls.table_name()} as r
+            ON r.environment = $1
+            AND r.resource_set_id = diff.resource_set_id
+        ORDER BY model_pairs.new, diff.name, r.resource_id
```
jptrindade, line 5725: I find it really hard to visualise what is the output of this query

src/inmanta/data/__init__.py
```diff
@@ -5507,123 +5515,254 @@ async def get_resources_for_version(
...
+
+        type Model = tuple[int, inmanta.types.ResourceSets]
+        result: list[Model] = []
+        version: int
+        records: Iterator[asyncpg.Record]
+        for version, records in itertools.groupby(resource_records, key=lambda r: r["version"]):
+            spy: Sequence[asyncpg.Record]
+            spy, records = more_itertools.spy(records)
+            assert spy  # groupby can not produce empty groups
+            if spy[0]["empty_model"]:
```
jptrindade, line 5749: is `spy` here mostly to make mypy happy? I would imagine `records[0]` is the same thing, right?

src/inmanta/data/__init__.py
jptrindade: Can you add a comment explaining what should be the output and in what format?

jptrindade (changes requested):

src/inmanta/data/__init__.py
sanderr: No, `records` is an `Iterator`, so there is no such thing as `records[0]`, and `next(records)` would consume the value.

src/inmanta/data/__init__.py
```diff
@@ -5507,123 +5515,254 @@ async def get_resources_for_version(
...
-                    # left join produced no resources
-                    continue
-                resource: dict[str, object] = dict(raw_resource)
-                parsed_resources.append(resource)
-            result.append((version, parsed_resources))
-        return result
+
+        records: Iterator[asyncpg.Record]
+        spy: Sequence[asyncpg.Record]
+        spy, records = more_itertools.spy(resource_records)
```
wouterdb, line 5612: this one is new to me, cool!

src/inmanta/data/__init__.py
arnaudsjs: I feel the same.

src/inmanta/deploy/scheduler.py
```diff
@@ -729,16 +781,49 @@ async def _get_intent_changes(
 
         version: int
         intent: dict[ResourceIdStr, ResourceIntent] = {}
+        resource_sets: dict[Optional[str], Set[ResourceIdStr]] = {}
```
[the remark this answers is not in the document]
sanderr: While this is true, it goes for each of the variables defined in this block. So I'll add a general comment in this vein a few lines higher.

src/inmanta/deploy/scheduler.py
```diff
@@ -729,16 +781,49 @@ async def _get_intent_changes(
...
+                        for resource_set in model.resource_sets.keys()
+                    )
+                )
+                # update resource sets mapping
+                for resource_set, resources in model.resource_sets.items():
+                    if len(resources) > 0:
+                        resource_sets[resource_set] = resources
+                    else:
+                        with contextlib.suppress(KeyError):
+                            del resource_sets[resource_set]
```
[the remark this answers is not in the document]
sanderr: Looking it over again, I agree it is incorrect, but for different reasons than you lay out. Note how we delete not on absence, but when we encounter an *empty* resource set. An empty set implies deletion of all its resources, and that we no longer need to track the resource set with that name.

I believe that that concept is important, and correct. But this is not the place to clean it up, that should only be done in `_new_version()` (note how there's a `del self._state.resource_sets[resource_set]` for the same condition).

By deleting it here, I believe the erroneous scenario would be that the new model version contains such an empty resource set to indicate deletion. But we delete it from the resulting model that we're building up to pass on to `_new_version()`. So `_new_version()` never sees this empty resource set, only the absence of it, which doesn't mean anything in a partial version. So it doesn't get to the `del self._state.resource_sets[resource_set]` branch.

Concluding: I think this for loop should simply become `resource_sets.update(model.resource_sets)`. Do you agree?

arnaudsjs (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -5507,123 +5515,254 @@ async def get_resources_for_version(
...
+            reference_model.exists,
+            model_pairs.new AS version,
+            -- each row of the diff will always have at least one of the resource set ids set.
+            -- => if both are NULL, there were no records at all on the rhs of the LEFT JOIN
+            (diff.old_resource_set_id IS NULL AND diff.resource_set_id IS NULL) AS empty_model,
+            diff.name AS resource_set_name,
+            diff.resource_set_id,
+            {projection_selectors}
+        FROM reference_model
+        LEFT JOIN model_pairs ON true
```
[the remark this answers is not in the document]
jptrindade: I think it is just a way of asserting that the reference model exists.

arnaudsjs (approved):

Issue #9407, opened 2025-08-05
include project constraints for agent code install


sanderr:
When instlaling packages for agent venv, take constraints from the project into account. This will require sanitizing the requirements from their `-r` form to `-c` form.

Even though constraints in `project.yml` are deprecated, we should take those into account as well (for now).

Hugo-Inmanta:
> Even though constraints in project.yml are deprecated

Are they ? I can't find any mention of this

PR #9487, opened 2025-08-28
use project constraints when installing agent code

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/9407

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -104,11 +104,14 @@ class ResourceDetails:
 
 @dataclasses.dataclass
 class EnvBlueprint:
-    """Represents a blueprint for creating virtual environments with specific pip configurations and requirements."""
+    """Represents a blueprint for creating virtual environments
+    with specific pip configurations, requirements and constraints."""
 
     environment_id: uuid.UUID
     pip_config: PipConfig
     requirements: Sequence[str]
+    constraints_file_hash: str | None
+    constraints: str
     _hash_cache: Optional[str] = dataclasses.field(default=None, init=False, repr=False)
     python_version: tuple[int, int]
 
@@ -118,9 +121,9 @@ class EnvBlueprint:
 
     def blueprint_hash(self) -> str:
         """
-        Generate a stable hash for an EnvBlueprint instance by serializing its pip_config
-        and requirements in a sorted, consistent manner. This ensures that the hash value is
-        independent of the order of requirements and consistent across interpreter sessions.
+        Generate a stable hash for an EnvBlueprint instance by serializing its pip_config, requirements
+        and constraints in a sorted, consistent manner. This ensures that the hash value is
+        independent of the order of requirements/constraints and consistent across interpreter sessions.
         Also cache the hash to only compute it once.
         """
         if self._hash_cache is None:
@@ -128,6 +131,7 @@ class EnvBlueprint:
                 "environment_id": str(self.environment_id),
                 "pip_config": self.pip_config.model_dump(),
                 "requirements": self.requirements,
+                "constraints_file_hash": self.constraints_file_hash,
                 "python_version": self.python_version,
             }
 
@@ -142,10 +146,17 @@ class EnvBlueprint:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, EnvBlueprint):
             return False
-        return (self.environment_id, self.pip_config, set(self.requirements), self.python_version) == (
+        return (
+            self.environment_id,
+            self.pip_config,
+            set(self.requirements),
+            set(self.constraints_file_hash),
+            self.python_version,
+        ) == (
             other.environment_id,
             other.pip_config,
             set(other.requirements),
+            set(other.constraints_file_hash),
             other.python_version,
         )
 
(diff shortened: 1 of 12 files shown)


src/inmanta/deploy/tasks.py
```diff
@@ -89,6 +89,7 @@ async def get_executor(
             environment=task_manager.environment, model_version=version, agent_name=agent_name
         )
 
+        logger_for_agent(agent_name).debug("Retrieve code %s", code)  # TODO way too noisy: remove or trim
```
jptrindade, line 92: TODO alert. Do you have a ticket for this?

src/inmanta/export.py
```diff
@@ -27,7 +27,9 @@
 
 import pydantic
 
-from inmanta import const, loader, protocol, references
+import inmanta.loader
+import inmanta.module
+from inmanta import const, protocol, references
```
jptrindade, line 32: Why is this distinction between the imports here?

src/inmanta/agent/executor.py
```diff
@@ -317,6 +340,13 @@ def __init__(self, env_path: str, io_threadpool: ThreadPoolExecutor):
         self.inmanta_venv_status_file: pathlib.Path = pathlib.Path(self.env_path) / const.INMANTA_VENV_STATUS_FILENAME
         self.folder_name: str = pathlib.Path(self.env_path).name
         self.io_threadpool = io_threadpool
+        self.constraint_file: pathlib.Path | None = None
+
+    def _write_constraint_file(self, blueprint: EnvBlueprint) -> None:
+        if blueprint.constraints_file_hash is not None and blueprint.constraints:
+            self.constraint_file = pathlib.Path(self.env_path) / blueprint.constraints_file_hash
```
arnaudsjs, line 347: Why use the hash as the name. People that look at the venv files will have no clue what this file is for.

src/inmanta/agent/executor.py
```diff
@@ -104,12 +104,15 @@ def from_json(cls, resource_dict: JsonType) -> "ResourceDetails":
...
-    """Represents a blueprint for creating virtual environments with specific pip configurations and requirements."""
+    """Represents a blueprint for creating virtual environments
+    with specific pip configurations, requirements and constraints."""
 
     environment_id: uuid.UUID
     pip_config: PipConfig
     requirements: Sequence[str]
-    _hash_cache: Optional[str] = dataclasses.field(default=None, init=False, repr=False)
+    constraints_file_hash: str | None = dataclasses.field(default=None, kw_only=True)
+    constraints: bytes | None = dataclasses.field(default=None, kw_only=True)
```
arnaudsjs, line 114: Why bytes?

tests/deploy/e2e/test_code_loading.py
```diff
@@ -855,7 +886,123 @@ async def check_code_for_version(
...
         environment=environment,
+        codemanager=codemanager,
         agent_names=["agent_Z", "agent_A"],
         module_name="new_module",
         expected_source=b"#Yet some other code",
     )
+
+
+@pytest.mark.parametrize("auto_start_agent", [True])
+async def test_project_constraints_in_agent_code_install(server, client, environment, clienthelper):
```
arnaudsjs, line 897: * This test doesn't test whether the constraint is actually taken into account, right?
* I think we must also add a test to verify whether the exporter works correctly.
* We probably also want to explicitly test the behavior when a partial export is done.

arnaudsjs (changes requested):

src/inmanta/deploy/tasks.py
Hugo-Inmanta: Removed.

src/inmanta/agent/executor.py
Hugo-Inmanta: Great point -> changed to `requirements.txt`

tests/test_export.py
```diff
@@ -330,6 +333,74 @@ async def test_dict_export_server(snippetcompiler, server, client, environment):
...
+        )
+
+    config.Config.set("config", "environment", environment)
+    snippetcompiler.setup_for_snippet(
+        """
+import many_dependencies
+
+a = many_dependencies::Test(name="my_test_resource")
+""",
+        autostd=True,
```
arnaudsjs, line 357: Why do we need std here?

tests/test_export.py
```diff
@@ -330,6 +333,74 @@ async def test_dict_export_server(snippetcompiler, server, client, environment):
...
+        index_url=index.url,
+        extra_index_url=["https://pypi.org/simple"],
```
arnaudsjs, line 359: Why is this extra index required?

tests/test_export.py
```diff
@@ -330,6 +333,74 @@ async def test_dict_export_server(snippetcompiler, server, client, environment):
     assert result.result["versions"][0]["total"] == 1
 
 
+async def test_project_constraints_at_exporter_boundary(snippetcompiler, server, client, environment, capsys, tmpdir):
```
arnaudsjs, line 336:
```suggestion
async def test_project_constraints_at_exporter_boundary(snippetcompiler, server, client, environment, tmpdir):
```

tests/test_export.py
Hugo-Inmanta: No longer relevant, will remove.

Hugo-Inmanta:
TODO: move constraint file out of module identity and into PipConfig. (still make sure it's propagated into executorvenv's hash)

PR #9558, opened 2025-09-15
Fix deletion of resource sets

jptrindade:
# Description

Fixes the deletion of resource sets.
Updates the migration query to delete dangling resource sets

# Self Check:

- [ ] Changelog entry


sanderr (commented):
One open question to be resolved on Slack. Other than that this looks good to me.

arnaudsjs (approved):

PR #9560, opened 2025-09-15
Show python and postgresql version

### LINKED ISSUE #8150: Show python and postgres version in the status page
It would be useful to show the postgres version in the status page. The python version should be fixed but I do not think it would hurt to also report this on the status page.

jptrindade:
# Description

Show python and postgresql version in /serverstatus

I made postgresql version optional in case the database is not connected

closes #8150

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/model.py
+++ b/src/inmanta/data/model.py
@@ -103,6 +103,8 @@ class StatusResponse(BaseModel):
     slices: list[SliceStatus]
     features: list[FeatureStatus]
     status: ReportedStatus
+    python_version: str
+    postgresql_version: str | None
 
 
 @stable_api
--- a/src/inmanta/data/model.py
+++ b/src/inmanta/data/model.py
@@ -103,7 +103,10 @@ class StatusResponse(BaseModel):
     slices: list[SliceStatus]
     features: list[FeatureStatus]
     status: ReportedStatus
+    # The python version used by the server
     python_version: str
+    # The postgresql version used by the database slice
+    # None if it is not initialized or an error occurred with the database slice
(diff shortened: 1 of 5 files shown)


tests/server/test_server_status.py
```diff
@@ -52,6 +54,12 @@ async def test_server_status(server, client, agent, environment):
     assert "features" in status
     assert len(status["features"]) > 0
 
+    assert status["python_version"] == sys.version
+    postgresql_version = await postgresql_client.fetchval("SELECT version();")
+    # Assert that the output is `PostgreSQL X.Y[.Z] ...`
+    assert re.match(r"^(PostgreSQL \d+(?:\.\d+)+)", postgresql_version)
```
arnaudsjs, line 60: I have the feeling that this assertion is too strict. It will fail if Postgres decides to change the format of this function. Time will tell.

arnaudsjs (changes requested):

src/inmanta/data/model.py
```diff
@@ -103,6 +103,11 @@ class StatusResponse(BaseModel):
     slices: list[SliceStatus]
     features: list[FeatureStatus]
     status: ReportedStatus
+    # The python version used by the server
+    python_version: str
+    # The postgresql version used by the database slice
+    # None if it is not initialized or an error occurred with the database slice
+    postgresql_version: str | None
```
arnaudsjs, line 110: You will have to add this information to the docstring of the class. Otherwise it's not shown in the generated API documentation.

arnaudsjs (changes requested):

Issue #9464, opened 2025-08-25
documentation for Result.all()


sanderr:
#9441 and its predecessor added `Result.all()`, `Result.value()` and `Result.all_sync()`, as well as the `ClientCall` classes and the mypy plugin. These should be documented.

1. docstrings for each method on `Result` + autodoc
2. docstring for `Client` class + autodoc. This docstring should mention that you can call these same methods as documented on `Result`, with simple examples. Should also mention mypy plugin
3. short section on mypy plugin

PR #9571, opened 2025-09-17
Issue/9464 add documentation for result all and mypy plugin

Hugo-Inmanta:
# Description

- make PageableResult part of stable api
- move rest api section from docs/reference/api.rst up one level into its own dedicated Python Client section
- extend this section with info on the mypy plugin
- add python code samples that show how to call each endpoint from the python client + the parameters it takes (see screen cap at the bottom)

Open questions:

~1. This is still missing docs for `all_sync` -> I don't fully get the use case / context for it~
2. ~The target audience for this documentation is still not clear to me.~ -> answered [in slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1758704508687269)

closes  https://github.com/inmanta/inmanta-core/issues/9464 
# Self Check:

- [ ] Changelog entry

### DIFF
--- a/docs/reference/python_client.rst
+++ b/docs/reference/python_client.rst
@@ -48,20 +48,22 @@ sort
 
     The documentation of each method describes the supported attributes to sort by.
 
-start
-    Define the lower limit for the page as an exclusive boundary. This parameter can be passed on its
-    own, or along with the ``first_id`` parameter.
 
-first_id
-    Continuation token to define the lower limit for the page as an exclusive boundary.
+start
+    Min boundary value (exclusive) for the requested page for the primary sort column, regardless of sorting order.
 
 end
-    Define the upper limit for the page as an exclusive boundary. This parameter can be passed on its
-    own, or along with the ``last_id`` parameter.
+    Max boundary value (exclusive) for the requested page for the primary sort column, regardless of sorting order.
 
+first_id
+    Min boundary value (exclusive) for the requested page for the secondary sort column, if there is one, regardless of sorting order.
+    When used along with the ``start`` parameter, the value of this parameter is used as a tiebreaker on the secondary sort column
+    for records whose primary sort column is equal to the ``start`` parameter.
 
 last_id
-    Continuation token to define the upper limit for the page as an **inclusive** boundary.
+    Max boundary value (exclusive) for the requested page for the secondary sort column, if there is one, regardless of sorting order.
+    When used along with the ``end`` parameter, the value of this parameter is used as a tiebreaker on the secondary sort column
+    for records whose primary sort column is equal to the ``end`` parameter.
 
 
 (only one of the (``start``, ``first_id``), (``end``, ``last_id``) pairs should be specified at the same time).
--- a/src/inmanta/data/dataview.py
+++ b/src/inmanta/data/dataview.py
@@ -107,8 +107,8 @@ class RequestedPagingBoundaries:
 
     :param start: min boundary value (exclusive) for the requested page for the primary sort column.
     :param end: max boundary value (exclusive) for the requested page for the primary sort column.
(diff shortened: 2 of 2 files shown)


Hugo-Inmanta:
v2 
image

docs/reference/api.rst
```diff
@@ -366,6 +371,22 @@ Rest API
 
 The rest API is also available as a `swagger spec <openapi.html#http://>`_
 
+You can interact with these API endpoints using the :py:class:`inmanta.protocol.endpoints.Client` python client:
+
+.. code-block:: python
+
+    client = inmanta.protocol.endpoints.Client(name="client_name", timeout=120)
+
+    result = await client.environment_list(details=True)
```
wouterdb, line 380: The issue here is: how do people find the method name... 

it is somewhere in the schema , if you look at the url, it is the last part of the anchor 
But we don't show it.  I think it is worth seeing if we can reconfigure the way we generate the docs to show the method name

https://sphinxcontrib-redoc.readthedocs.io/en/stable/
https://github.com/Redocly/redoc/issues/536


image

docs/reference/api.rst
```diff
@@ -366,6 +371,22 @@ Rest API
 
 The rest API is also available as a `swagger spec <openapi.html#http://>`_
```
wouterdb, line 372: I think this entire thing is in the wrong place. It is now several 1000 lines deep in the Programmatic API reference. I think this should be its own top level item under references.

docs/reference/api.rst
```diff
@@ -384,36 +405,78 @@ end and last_id
...
+        for item in client.resource_list(tid=env.id, limit=20).all_sync():
+            assert item == all_resources[idx]
+            idx += 1
+
+
+.. _python_client_mypy_plugin:
+
+.. note::
+
+    Is it possible to inspect typing information about the :py:class:`inmanta.protocol.common.Client` and the
```
wouterdb, line 439: I would find it more natural to mention that we can do type checks on the result with the plugin first and how to set it up?

src/inmanta/protocol/methods.py
```diff
@@ -50,6 +50,8 @@ async def add_env(env: uuid.UUID, metadata: dict) -> uuid.UUID:
 async def ignore_env(obj: Any, metadata: dict) -> Any:
     """
     This mapper only adds an env all for authz
+
+    :meta private:
```
Hugo-Inmanta: This is to hide this method from the public docs

docs/reference/python_client.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Paging, sorting and filtering
+=============================
+
+The (v2) API endpoints that offer paging, sorting and filtering follow a convention.
+They share the following parameters:
+
+limit
+    specifies the page size, so the maximum number of items returned from the query
+start and first_id
```
sanderr: Inclusive or exclusive?

src/inmanta/protocol/openapi/converter.py
```diff
@@ -464,23 +468,52 @@ def handle_method(self, url_method: UrlMethod, path: str) -> Operation:
...
+            try:
+                request_body_parameters = list(extra_params["requestBody"].content["application/json"].schema_.properties.keys())
+            except (KeyError, AttributeError):
+                pass
         else:
             extra_params = {}
 
         tags = self._get_tags_of_operation(url_method)
-
+        code_samples = self._generate_code_sample(
```
sanderr: Cool!

sanderr (changes requested):

src/inmanta/protocol/methods.py
sanderr: I see. I think that might be a good idea. The docstring is probably the easiest.

PR #9575, opened 2025-09-17
Fix migration where we add the created column to the rps table

jptrindade:
# Description

Tested on the mgmt-dump.
`cat mgmt-dump.sql | psql -h 172.30.0.2 -p 5432 -U inmanta -d inmanta -e > output_log2.txt 2>&1`
[Log of the loading of the dump](https://github.com/user-attachments/files/22382530/output_log2.txt)

[migration.sql](https://github.com/user-attachments/files/22382645/migration.sql)
 `psql -h 172.30.0.2 -p 5432 -U inmanta -d inmanta -f migration.sql > run.log 2>&1`
[Log of the execution of the migration](https://github.com/user-attachments/files/22382531/run.log)

We trim the resource_persistent_state table when we call `_purge_versions`

# Self Check:

- [ ] Changelog entry


tests/db/migration_tests/test_v202509050_to_v202509100.py
```diff
@@ -37,6 +37,36 @@ async def test_add_created_to_resource_persistent_state(
     Check that the RPS column is added and that it is populated to the
     configuration model version that the resource first appears in.
     """
+    # Insert dangling rps entry
```
sanderr, line 40: I'd move this to the dump tool (and then regenerate the "pre" dump). so we can't make this same mistake again.

tests/db/migration_tests/dumps/v202509100.sql
jptrindade, line 1: added dangling rps and discovered resources

src/inmanta/db/versions/v202509100.py
```diff
@@ -33,6 +33,15 @@ async def update(connection: Connection) -> None:
...
 
+    -- Remove rps entries that no longer have a matching resource in the database
+    DELETE FROM public.resource_persistent_state AS rps
+    WHERE NOT EXISTS (
+        SELECT 1
+        FROM resource AS r
+        WHERE rps.environment=r.environment
+        AND rps.resource_id=r.resource_id
+    );
+
```
arnaudsjs, line 44: We need to verify the performance of this query. I will already approve it, so that you can merge it in when it's verified.

arnaudsjs (approved):

PR #9581, opened 2025-09-17
Remove version information from resource attributes

### LINKED ISSUE #9570: Figure out if we still need to add version information in the attributes of Resources
With the improved put_partial PR we removed `__mangle_dict` which we used to add version information to the attributes of Resources on certain endpoints. We managed to delete that method but we still manually do this in some places, like:
- `Resource.get_released_resource_details`
- `ResourceHistoryView.construct_dtos`
- `VersionedResourceDetails.ensure_version_field_set_in_attributes`

jptrindade:
# Description

Removes the remaining instances where we were adding the version to the attributes of a resource

No fallout on:
- inmanta-lsm
- lsm
- license

closes #9570 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/model.py
+++ b/src/inmanta/data/model.py
@@ -27,7 +27,7 @@ import uuid
 from collections import abc
 from collections.abc import Sequence
 from enum import Enum, StrEnum
-from typing import ClassVar, Mapping, Optional, Union, assert_never
+from typing import ClassVar, Mapping, Optional, Self, Union, assert_never
 
 import pydantic.schema
 from pydantic import ConfigDict, Field, SerializationInfo, computed_field, field_serializer, field_validator


changelogs/unreleased/9570-remove-version-information-from-resource-attributes.yml
```diff
@@ -0,0 +1,6 @@
+description: Remove version information from resource attributes
```
sanderr, line 1:
```suggestion
description: Remove version information from resource attributes returned by the API
```

Perhaps this makes it a bit more informative? Considering that we include it in the changelog.

src/inmanta/data/dataview.py
```diff
@@ -926,20 +926,11 @@ def get_base_query(self) -> SimpleQueryBuilder:
...
-                # version field is present in the attributes dictionary served out via the API.
-                attributes["version"] = record["model"]
-            return attributes
-
         return [
             ResourceHistory(
                 resource_id=self.rid,
                 attribute_hash=record["attribute_hash"],
-                attributes=get_attributes(record),
+                attributes=cast(dict[str, Any], record["attributes"]),
```
sanderr, line 933: Doesn't really hurt here, so not a strict change request, but I don't think we should ever cast to `Any`, unless we really want to disable type checking. `object` could be appropriate. But honestly I find it even more appropriate to simply accept this error in the mypy baseline.

Issue #9543, opened 2025-09-11
On server startup, check that the postgresql version matches our minimum required version


jptrindade:
We can fetch the minimum required version from the compatibility.json

https://docs.inmanta.com/inmanta-service-orchestrator-dev/9/reference/compatibility.json

PR #9589, opened 2025-09-18
Issue/9543 check postgresql version on server startup

Hugo-Inmanta:
# Description

Add a check for the version of postgres being used during server startup.

closes  https://github.com/inmanta/inmanta-core/issues/9543

# Self Check:

- [ ] Changelog entry


Hugo-Inmanta:
Output examples:

required pg version == 16:
```
inmanta-orchestrator  | 2025-09-18 12:13:10,666 inmanta.server.bootloaderINFO    Successfully connected to the database (PostgreSQL server version 16.10 (Debian 16.10-1.pgdg13+1)).
```
required pg version == 23:


```
inmanta-orchestrator  | 2025-09-18 12:13:52,746 inmanta.server.bootloaderWARNING The database at inmanta-db is using PostgreSQL version 16.10 (Debian 16.10-1.pgdg13+1). This version is not supported by this version of the Inmanta orchestrator. Please make sure to update to PostgreSQL 23 as soon as possible.
inmanta-orchestrator  | 2025-09-18 12:13:52,746 inmanta.server.bootloaderINFO    Successfully connected to the database (PostgreSQL server version 16.10 (Debian 16.10-1.pgdg13+1)).
```

src/inmanta/server/bootloader.py
```diff
@@ -291,11 +317,30 @@ async def wait_for_db(self, db_wait_time: int) -> None:
...
         }
+        human_readable_min_pg_version, machine_readable_min_pg_version = self._get_minimal_postgres_version()
+
         while True:
             try:
                 # Attempt to create a database connection
                 conn = await asyncpg.connect(**db_settings, timeout=5)  # raises TimeoutError after 5 seconds
-                LOGGER.info("Successfully connected to the database.")
+                result = await conn.fetch("SHOW server_version_num")
+                pg_server_version_machine_readable = int(result[0]["server_version_num"])
```
Hugo-Inmanta, line 327: comment for reference

src/inmanta/app.py
```diff
@@ -84,6 +84,15 @@ def server_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
         help="Maximum time in seconds the server will wait for the database to be up before starting. "
         "A value of 0 means the server will not wait. If set to a negative value, the server will wait indefinitely.",
     )
+    parser.add_argument(
+        "--compatibility-file",
+        type=str,
+        dest="compatibility_file",
+        help="Path to the compatibility.json file. During startup, the server will perform version compatibility check "
+        "for the PostgreSQL version being used. For more information about this file, please refer to "
+        "https://docs.inmanta.com/inmanta-service-orchestrator/latest/reference/compatibility.html.",
```
arnaudsjs, line 93: This links to the documentation of the latest iso version. It's not necessarily the one for the running orchestrator. I am also not sure whether we want to make the server log a url to our docs. If we ever change the path for some reason, we will forget to update the link in the code. It's probably better to mention to consult the documentation without offering a concrete url.

src/inmanta/app.py
```diff
@@ -97,6 +106,9 @@ def start_server(options: argparse.Namespace) -> None:
     if options.db_wait_time is not None:
         Config.set("database", "wait_time", str(options.db_wait_time))
 
+    if options.compatibility_file:
+        Config.set("server", "compatibility_file", str(options.compatibility_file))
```
arnaudsjs, line 110: I don't think this is going to work. The option has a default value so the if-statement will always validate to True. Like that we will always ignore the config value set in the configuration file. I think we have to set the default value to None so that we can determine whether the option is set on the comandline or not.

src/inmanta/app.py
```diff
@@ -84,6 +84,15 @@ def server_parser_config(parser: argparse.ArgumentParser, parent_parsers: abc.Se
         help="Maximum time in seconds the server will wait for the database to be up before starting. "
         "A value of 0 means the server will not wait. If set to a negative value, the server will wait indefinitely.",
     )
+    parser.add_argument(
+        "--compatibility-file",
+        type=str,
+        dest="compatibility_file",
+        help="Path to the compatibility.json file. During startup, the server will perform version compatibility check "
```
arnaudsjs, line 91:
```suggestion
        help="Path to the compatibility.json file. During startup, the server will perform a version compatibility check "
```

src/inmanta/server/config.py
```diff
@@ -161,6 +161,16 @@ def default_db_pool_max_size() -> int:
     "server", "ssl_cert_file", None, "SSL certificate file for the server key. Leave blank to disable SSL", is_str_opt
 )
 
+server_compatibility_file = Option(
+    "server",
+    "compatibility_file",
+    "/usr/share/inmanta/compatibility/compatibility.json",
+    "Path to the compatibility.json file. During startup, the server will perform version compatibility check "
```
arnaudsjs, line 168:
```suggestion
    "Path to the compatibility.json file. During startup, the server will perform a version compatibility check "
```

src/inmanta/server/bootloader.py
```diff
@@ -291,11 +317,30 @@ async def wait_for_db(self, db_wait_time: int) -> None:
...
             "database": config.db_name.get(),
         }
+        human_readable_min_pg_version, machine_readable_min_pg_version = self._get_minimal_postgres_version()
+
         while True:
             try:
                 # Attempt to create a database connection
                 conn = await asyncpg.connect(**db_settings, timeout=5)  # raises TimeoutError after 5 seconds
-                LOGGER.info("Successfully connected to the database.")
+                result = await conn.fetch("SHOW server_version_num")
```
arnaudsjs, line 326: Why put this here. I would expect this to be in the database slice.

src/inmanta/server/bootloader.py
```diff
@@ -291,11 +317,30 @@ async def wait_for_db(self, db_wait_time: int) -> None:
...
+                    LOGGER.warning(
+                        f"The database at {config.db_host.get()} is using PostgreSQL version "
+                        f"{pg_server_version_human_readable}. This version is not supported by this "
+                        "version of the Inmanta orchestrator. Please make sure to update to PostgreSQL "
+                        f"{str(human_readable_min_pg_version)} as soon as possible."
+                    )
+
+                LOGGER.info(
+                    f"Successfully connected to the database (PostgreSQL server version {pg_server_version_human_readable})."
+                )
```
arnaudsjs, line 342: We must not do string concatenation here.

src/inmanta/server/bootloader.py
```diff
@@ -291,11 +317,30 @@ async def wait_for_db(self, db_wait_time: int) -> None:
...
+
         while True:
             try:
                 # Attempt to create a database connection
                 conn = await asyncpg.connect(**db_settings, timeout=5)  # raises TimeoutError after 5 seconds
-                LOGGER.info("Successfully connected to the database.")
+                result = await conn.fetch("SHOW server_version_num")
+                pg_server_version_machine_readable = int(result[0]["server_version_num"])
+                result = await conn.fetch("SHOW server_version")
+                pg_server_version_human_readable = result[0]["server_version"]
```
arnaudsjs, line 329: * The logic for this PostgreSQL version is a bit scattered. It would be good to introduce a new class (for example `PostgreSQLVersion`) that represent a PostgreSQL version number. It can then offer methods to create an instance of that class from the machine readable version, returned by the PostgreSQL server. And a method to create an instance from the version number present in the compatibility.json file. You can then also add logic to compare two instances of this class.
* This method performs two calls to the database. One to get the machine readable version and on to get the human readable version. Wouldn't it be better to just fetch the machine readable version and calculate the human readable version from that? It saves a DB call and we can be sure what the format of the version number shown to the user will be.

src/inmanta/server/bootloader.py
```diff
@@ -291,11 +317,30 @@ async def wait_for_db(self, db_wait_time: int) -> None:
...
+
+                if pg_server_version_machine_readable < machine_readable_min_pg_version:
+
+                    LOGGER.warning(
```
arnaudsjs, line 333: We need to fail the server startup if the version is too low. (See Slack discussion).

changelogs/unreleased/9543-check-postgresql-version-on-server-startup.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: "Check PostgreSQL version on server startup."
+issue-nr: 9543
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master, iso8]
```
arnaudsjs, line 6: We need to add a changelog entry that clearly mentions that the server will not start when running against a PostgreSQL version that doesn't meet the minimal PostgreSQL version mentioned in the docs.

arnaudsjs (changes requested):

jptrindade (changes requested):
You can fetch the postgres version from a method present in the database slice that I added yesterday. Maybe we can calculate it when the slice is initialized and then just return the saved value

jptrindade (approved):
Looks nice!

arnaudsjs (changes requested):

src/inmanta/server/bootloader.py
```diff
@@ -291,20 +297,96 @@ async def wait_for_db(self, db_wait_time: int) -> None:
...
+        required_postgresql_version = PostgreSQLVersion.from_compatibility_file()
+
+        conn: asyncpg.Connection | None = None
         while True:
             try:
                 # Attempt to create a database connection
                 conn = await asyncpg.connect(**db_settings, timeout=5)  # raises TimeoutError after 5 seconds
-                LOGGER.info("Successfully connected to the database.")
-                await conn.close(timeout=5)  # close the connection
+                installed_postgresql_version = await PostgreSQLVersion.from_database(conn)
```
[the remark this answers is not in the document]
Hugo-Inmanta: Changed to `database_postgresql_version`.

src/inmanta/server/bootloader.py
```diff
@@ -291,20 +297,96 @@ async def wait_for_db(self, db_wait_time: int) -> None:
...
+
             # Check if the maximum wait time has been exceeded
             if 0 < db_wait_time < asyncio.get_event_loop().time() - start_time:
                 LOGGER.error("Timed out waiting for the database to be up.")
                 raise Exception("Database connection timeout after %d seconds." % db_wait_time)
             # Sleep for a second before retrying
             await asyncio.sleep(1)
+
+
+@total_ordering
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think it already does via the `object` super class.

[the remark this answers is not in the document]
arnaudsjs: Yes. I agree with that. But the implementation doesn't make any sense and it also has effect on the behavior of `>` operator of course.

```
>>> from inmanta.server.bootloader import PostgreSQLVersion
>>> x = PostgreSQLVersion(1)
>>> y = PostgreSQLVersion(1)
>>> x == y
False
>>> x == x
True
>>> x > y
True
>>> x > x
False
```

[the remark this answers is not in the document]
arnaudsjs: And if you override `__eq__()` you must also override  `__hash__()` so that dictionaries behave correctly if the object would be used as the key.

arnaudsjs (changes requested):
See last comment.

src/inmanta/server/config.py
```diff
@@ -161,6 +161,16 @@ def default_db_pool_max_size() -> int:
     "server", "ssl_cert_file", None, "SSL certificate file for the server key. Leave blank to disable SSL", is_str_opt
 )
 
+server_compatibility_file = Option(
+    "server",
+    "compatibility_file",
+    "/usr/share/inmanta/compatibility/compatibility.json",
```
arnaudsjs, line 167: This default value is not going to work. We only add the compatibility file to the docker containers and not to the RPM. So this would mean that servers, installed from RPM, will not be able to start after an upgrade because the compatibility file is missing. Also, this would mean that starting the server manually (using the inmanta command) on a development machine would not be possible anymore unless you explicitly set this config option to an empty string or put the compatibility file in the right directory. I don't think that is what we want.

arnaudsjs (changes requested):

changelogs/unreleased/9543-check-postgresql-version-on-server-startup.yml
```diff
@@ -0,0 +1,12 @@
+---
+description: >
+  Add a check for the PostgreSQL version of the database on server startup. The server will not start if the PostgreSQL
+  version of the database is not compatible with this version of the orchestrator. Please make sure you update your
+  database if required. The compatible PostgreSQL version can be found in the "Compatibility" page of the Inmanta
+  documentation.
+issue-nr: 9543
+issue-repo: inmanta-core
+change-type: minor
+destination-branches: [master]
```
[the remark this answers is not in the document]
Hugo-Inmanta: Yes iirc from the standup, now that this is not just a warning but a hard failure, we don't want to force a db version update on users doing a minor iso8 update.

src/inmanta/server/bootloader.py
```diff
@@ -308,3 +377,77 @@ async def wait_for_db(self, db_wait_time: int) -> None:
...
+        return PostgreSQLVersion(version=pg_server_version_machine_readable)
+
+    @classmethod
+    def from_compatibility_file(cls) -> "PostgreSQLVersion | None":
+        """
+        Helper method to retrieve the minimal required postgreSQL version configured under the
+        system_requirements->postgres_version section of the compatibility file. Returns None
+        if the server.compatibility_file option is set to None or to empty string.
+
+        :raises ServerStartFailure: If the compatibility file doesn't exist.
```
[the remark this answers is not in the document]
Hugo-Inmanta: The logic is there L422-423, do you mean it is something we no longer want ?

src/inmanta/server/bootloader.py
```diff
@@ -113,6 +115,65 @@ async def start(self) -> None:
...
+
+    async def _database_version_compatibility_check(self) -> None:
+        """
+        This method looks for the required PostgreSQL version defined in the compatibility file (Whose path is configured by
+        the server.compatibility_file option) and checks that the PostgreSQL version of the database meets this requirement.
+
+        The check is bypassed if the server_compatibility_file option is set to None or to an empty string.
+
+        :raises ServerStartFailure: If the compatibility file schema is missing the `system_requirements->postgres_version`
+            section.
```
jptrindade, line 152: Or if the compatibility file is missing entirely

jptrindade (changes requested):

jptrindade (approved):

arnaudsjs (approved):

PR #9642, opened 2025-10-10
Graphql paging investigations

jptrindade:
# Description

Reworked graphql paging to use sqlakeyset instead of offset paging.
We no longer fetch the whole set of resources and then filter for the ones we actually want.
Should perform a lot better especially in larger databases like the resource view

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -149,7 +149,6 @@ There are 4 important building blocks that we have to take into account:
 
 mapper: StrawberrySQLAlchemyMapper[typing.Any] = StrawberrySQLAlchemyMapper()
 DEFAULT_PER_PAGE: int = 50
-EDGE_CURSOR_PREFIX: str = "arrayconnection:"
 
 
 def to_snake_case(name: str) -> str:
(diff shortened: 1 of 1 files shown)


src/inmanta/graphql/schema.py
```diff
@@ -145,6 +148,8 @@ def key_to_model(self) -> dict[str, type[models.Base]]:
 """
 
 mapper: StrawberrySQLAlchemyMapper[typing.Any] = StrawberrySQLAlchemyMapper()
+DEFAULT_PER_PAGE: int = 50
+EDGE_CURSOR_PREFIX: str = "arrayconnection:"
```
wouterdb: it is not array based paging? I find this an odd choice?

src/inmanta/graphql/schema.py
```diff
@@ -526,6 +566,86 @@ def do_required_resource_joins(
...
+    return base64.b64encode(f"{EDGE_CURSOR_PREFIX}{cursor}".encode()).decode()
+
+
+def decode_cursor(cursor: str) -> str:
+    """
+    :param cursor: The cursor received as an argument from the user.
+        Expected in the format that Edge.resolve_edge returns (prefixed with `EDGE_CURSOR_PREFIX`).
+    :return: The decoded cursor in str form.
+    """
+    return base64.b64decode(cursor).decode().split(EDGE_CURSOR_PREFIX)[1]
```
wouterdb: if the string is not in here, I would raise an exception

src/inmanta/graphql/schema.py
jptrindade: I hate it as well... But it is the default cursor prefix of `Edge.resolve_edge`. And they are generated classes so I would rather not mess with them.
But yeah, it is not array based paging

jptrindade: either way, I think it is better to change this to `strawberry.relay.types.PREFIX`

wouterdb (commented):
I see no changes?

PR #9669, opened 2025-10-20
Allow report only resources

jptrindade:
# Description

* Short description here *

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/agent/handler.py
+++ b/src/inmanta/agent/handler.py
@@ -945,92 +945,6 @@ class ResourceHandler(HandlerAPI[TResource]):
 TPurgeableResource = TypeVar("TPurgeableResource", bound=resources.PurgeableResource)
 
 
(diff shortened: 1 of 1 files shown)


src/inmanta/agent/executor.py
```diff
@@ -631,6 +631,7 @@ class DeployReport:
     messages: list[LogLine]
     changes: dict[str, AttributeStateChange]
     change: Optional[Change]
+    report: Optional[bool] = False
```
sanderr: Could we extend `Change` for this instead? It seems appropriate to me that we would make it mutually exclusive with the other `Change` modes, and it takes away an ambiguous field.

Or we can simply rely on the `resource_state`. I'm not quite sure if we want to use `change`, `resource_state` or both. Perhaps just `resource_state`?

Issue #9663, opened 2025-10-17
Introduce a field in the RPS table to know the model version when a resource was last seen.


jptrindade:
**Proposal**: Keep track of the latest released version where an orphaned resource was present in the RPS table.
**Motivation**: Finding the latest released version of every resource is a painful and costly task if we include orphans. Adding a field that had this information to RPS would simplify this type of query (currently used on the ResourceView)
**Problem**: How do we know this exactly?

- Do we figure out in which version a resource was orphaned and subtract 1?
- Set it to the last_processed_model_version  when we mark it as orphan?

>I think it's a good proposal, but I think that ideally we'd keep it a single field, e.g. is_orphan: bool -> orphaned_at: Optional[int].
Or alternatively, more in the direction Wouter was going: add last_processed_version for all resources, and then it's a simple comparison. But I think I prefer the fewer writes of a dedicated orphan field.
A difficulty with both approaches is that the scheduler doesn't keep track of when a resource is deleted when in processes multiple versions, only that it is deleted.

related slack thread: https://inmanta.slack.com/archives/CKRF0C8R3/p1760621877390979

PR #9673, opened 2025-10-22
fix scheduler bug: mark orphans for new-then-deleted resources

sanderr:
# Description

- Fix scheduler bug where orphans wouldn't be marked correctly two versions are processed in bulk where a resource is new in the first version and deleted in the next.
- By happenstance fixed a bug where the scheduler would crash if a transitively blocked resource is orphaned in the same version that it is unblocked (my test scenario happened to have this, which brought it to my attention, and the fix was a one-liner).
- Set the stage for #9663.

The test scenario I added fails on master.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/deploy/e2e/test_deploy.py
+++ b/tests/deploy/e2e/test_deploy.py
@@ -1519,7 +1519,7 @@ async def test_resource_status(resource_container, server, client, clienthelper,
             ResourceIdStr("test::Resource[agent1,key=key3]"): ReleasedResourceState.deployed,
             ResourceIdStr("test::Resource[agent1,key=key4]"): ReleasedResourceState.deployed,
             ResourceIdStr("test::Resource[agent1,key=key5]"): ReleasedResourceState.deployed,
-            # veriry that the new resource was registered an orphan
+            # verify that the new resource was registered as orphan
             ResourceIdStr("test::Resource[agent1,key=key6]"): ReleasedResourceState.orphaned,
         }
     )


src/inmanta/deploy/scheduler.py
```diff
@@ -108,6 +108,28 @@ class ResourceIntentChange(Enum):
...
+    change = ResourceIntentChangeType.NEW
+
+
+class Updated(_ResourceIntentChange):
+    change = ResourceIntentChangeType.UPDATED
+
+
+@dataclass(frozen=True, kw_only=True)
+class Deleted(_ResourceIntentChange):
+    change: ClassVar[ResourceIntentChangeType] = ResourceIntentChangeType.DELETED
```
Hugo-Inmanta: I'm curious why `: ClassVar[ResourceIntentChangeType]` here and not for the other 2 classes?

src/inmanta/deploy/scheduler.py
```diff
@@ -971,7 +997,7 @@ async def _new_version(
             i += 1
 
         # convert intent changes to sets for bulk processing
-        deleted: set[ResourceIdStr] = set()
+        deleted: dict[ResourceIdStr, Deleted] = {}
```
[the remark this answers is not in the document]
sanderr: See [Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1761133340431649?thread_ts=1761132662.914909&cid=CKRF0C8R3). I'm still on the fence on whether I should keep it in this PR or not, but let's take that part of the discussion to Slack.

src/inmanta/deploy/scheduler.py
sanderr: Here because `dataclass` requires it. For the others I didn't think the verbosity was worth it. But I'm not opposed to adding it there.

PR #9697, opened 2025-10-31
added test case for orphan marking on scheduler initialization

sanderr:
# Description

Test case for bugfix in #9696. I will only merge this on master because the bug isn't really user-visible (DB only) on iso8 due to the differences in how we report resource states. Therefore the assertions don't apply. I will however apply the file rename for consistency.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/scheduler-initialize-orphans-test-case.yml
+++ b/changelogs/unreleased/scheduler-initialize-orphans-test-case.yml
@@ -2,4 +2,3 @@ description: Added test case for orphan marking on scheduler initialization
 change-type: patch
 destination-branches:
   - master
-  - iso8


tests/deploy/e2e/test_initialize.py
```diff
@@ -343,3 +344,104 @@ async def get_last_deployed() -> dict[int, datetime.datetime]:
...
+        return agent
+
+    rid1: str = "test::Resource[agent1,key=key1]"
+    rid2: str = "test::Resource[agent1,key=key2]"
+    agent: Agent | None = None
+
+    def resources(version: int, *, r2: bool = True) -> list[dict[str, object]]:
+        def res(rid: str) -> None:
+            return {
+                "key": "key1",
```
Hugo-Inmanta: This should change accordingly with the rid, but it might not be relevant/impactful for this test case
```suggestion
                "key": "key1", 
```

tests/deploy/e2e/test_initialize.py
```diff
@@ -1,5 +1,5 @@
 """
-Copyright 2019 Inmanta
+Copyright 2024 Inmanta
```
Hugo-Inmanta: Happy new year!

```suggestion
Copyright 2025 Inmanta
```

tests/deploy/e2e/test_initialize.py
```diff
@@ -343,3 +344,105 @@ async def get_last_deployed() -> dict[int, datetime.datetime]:
...
+
+    rid1: str = "test::Resource[agent1,key=key1]"
+    rid2: str = "test::Resource[agent1,key=key2]"
+    agent: Agent | None = None
+
+    def resources(version: int, *, r2: bool = True) -> list[dict[str, object]]:
+        def res(n: int) -> None:
+            return {
+                "key": f"key{n}",
+                "value": "val{i}",
```
Hugo-Inmanta: Is this intentional ?
```suggestion
                "value": f"val{n}",
```

changelogs/unreleased/scheduler-initialize-orphans-test-case.yml
```diff
@@ -0,0 +1,5 @@
+description: Added test case for orphan marking on scheduler initialization
+change-type: patch
+destination-branches:
+  - master
+  - iso8
```
sanderr:
```suggestion
```

PR #9702, opened 2025-11-03
cleanup indexes in resource table

### LINKED ISSUE #9577: Cleanup indexes in the resource table
Introduce an index  `(environment, resource_id)` on the resource table
Figure out if we still need the index on `(resource_id)`. If so, add a comment as to why we need it.

Introduced in: https://github.com/inmanta/inmanta-core/pull/6381

Possibly related to a bug where we forgot to filter by environment here: https://github.com/inmanta/inmanta-core/blob/30f68c51c3d7475903e230115ef76d0a5c4217a7/src/inmanta/data/__init__.py#L6481

# Actionables
- introduce the new index
- drop the old one
- fix the bug if it still exists

jptrindade:
# Description

I took a look at the queries on the resource table (that I could find) and this was the only one that was not filtering on environment.
 
closes #9577 

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (approved):

src/inmanta/db/versions/v202511030.py
```diff
@@ -0,0 +1,32 @@
...
+
+async def update(connection: Connection) -> None:
+    """
+    Update index on resource table
+    """
+    schema = """
+    DROP INDEX resource_resource_id_index;
+
+    CREATE INDEX resource_environment_resource_id_index
+        ON public.resource (environment, resource_id);
```
arnaudsjs: This index is not required. This information is already contained in other indexes, for example `resource_env_resourceid_index`.

changelogs/unreleased/9577-cleanup-indexes-in-resource-table.yml
```diff
@@ -0,0 +1,8 @@
+description: Cleanup indexes in resource table
+issue-nr: 9577
+issue-repo: inmanta-core
+change-type: patch
+destination-branches:
+- master
+- iso8
```
arnaudsjs: This bug also exists on iso7, right?

PR #9711, opened 2025-11-05
Default eventloop

### LINKED ISSUE #9542: Investigate if we are leaking self-pipes
Validate that, 
1. if we create syncclients in handlers, this may cause self-pipes to leak 
2. on the latest released iso8
3. no longer on the latest iso8

```
lsof | grep "type=STREAM"  
```

This mechanism is fairly complex, as it interacts with executor restarts, threadpools, ....

wouterdb:
# Description

1. Should I add tests???
2. Should I set the default loop in the tests as well? 

closes #9542 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/protocol/endpoints.py
+++ b/src/inmanta/protocol/endpoints.py
@@ -31,7 +31,7 @@ from urllib import parse
 from inmanta import config as inmanta_config
 from inmanta import const, tracing, types, util
 from inmanta.protocol import common
-from inmanta.util import TaskHandler, get_default_event_loop
+from inmanta.util import TaskHandler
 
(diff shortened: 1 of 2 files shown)


arnaudsjs (approved):

changelogs/unreleased/9542-default-event-loop.yml
```diff
@@ -0,0 +1,6 @@
+description: "Fix issue where using Client() in handlers would leak open files"
```
sanderr:
```suggestion
description: "Fix issue where using SyncClient() in handlers would leak open files"
```

PR #9717, opened 2025-11-06
fixed unknown evaluation in implement conditions

### LINKED ISSUE #8672: implementation's when clause treats unknowns as true
The `when` clause for an implementation treats `Unknown` values as if they were `true`, i.e. it emits the implementation. This is not consistent with the behavior of `if`. Example model that illustrates both behaviors below.

```
entity A:
    string name
    bool condition
end
implement A using conditional when self.condition
implement A using always

implementation conditional for A:
    std::print(f"implementation triggered for {self.name}!")
end
implementation always for A:
    if self.condition:
        std::print(f"if body triggered for {self.name}!")
    end
end


unknown = std::get_env("THIS_ENV_VAR_DOES_NOT_EXIST")

A(name="true", condition=true)
A(name="false", condition=false)
A(name="unknown", condition=unknown)
```

The output is
```
implementation triggered for true!
implementation triggered for unknown!
if body triggered for true!
```

sanderr:
# Description

Fixed bug where implement conditions evaluated Unknown values as true, deviating from how they are evaluated in if statement conditions

closes #8672 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/8672-compiler-bugfix-unknown-implement-condition.yml
+++ b/changelogs/unreleased/8672-compiler-bugfix-unknown-implement-condition.yml
@@ -1,0 +1,6 @@
+description: 'Compiler: fixed bug where implement conditions evaluated Unknown values as true, deviating from how they are evaluated in if statement conditions'
+change-type: patch
+sections:
+  bugfix: "{{description}}"
+destination-branches:
+  - master
--- a/src/inmanta/ast/statements/generator.py
+++ b/src/inmanta/ast/statements/generator.py
@@ -142,7 +142,7 @@ class SubConstructor(RequiresEmitStatement):
                 % condition
             )
             raise e
-        if not condition:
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/8672-compiler-bugfix-unknown-implement-condition.yml
```diff
@@ -0,0 +1,6 @@
+description: 'Compiler: fixed bug where implement conditions evaluated Unknown values as true, deviating from how they are evaluated in if statement conditions'
+change-type: patch
+sections:
+  bugfix: "{{description}}"
+destination-branches:
+  - master
```
sanderr: Should this go into iso8? It's a bugfix, but I'm hesitant because it is a small change of behavior. I'm inclined to leave it iso9 only.

Hugo-Inmanta (approved):

PR #9732, opened 2025-11-13
Introduce report_only mode

jptrindade:
# Description

Add last_deploy_compliant to scheduler and rps table
Skip CUD actions if resource.report_only is True
recovered from failure logic now uses compliance instead of last deploy result

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/test_graphql.py
+++ b/tests/test_graphql.py
@@ -622,6 +622,15 @@ async def test_query_resources(server, client, environment, mixed_resource_gener
     total_resources_in_latest_version = resources_per_version * instances
     total_resources = total_resources_in_latest_version + (orphans * instances)
     await mixed_resource_generator(environment, instances, resources_per_version)
+
+    # Quick way of simulating a non-compliant report
+    # It has to be non-orphan otherwise the complianceState returned will be None
+    rps = await data.ResourcePersistentState.get_one(
+        environment=environment, last_deploy_result=state.DeployResult.DEPLOYED, is_orphan=False
+    )
+    assert rps
+    await rps.update_fields(last_deploy_compliant=False)
+
     filters = [
         # (1 undefined, 1 skipped for undefined) * <instances>
         {
@@ -697,6 +706,18 @@ async def test_query_resources(server, client, environment, mixed_resource_gener
                     "lastDeployResult": state.DeployResult.NEW.name,
                     "isDeploying": False,
                     "complianceState": state.Compliance.HAS_UPDATE.name,
+                }
+            },
+        },
+        # Non-compliant report
+        {
+            "query": "lastDeployResult: {eq: DEPLOYED} isDeploying: false complianceState: {eq: NON_COMPLIANT}",
+            "result": 1,
+            "assertion": {
+                "state": {
+                    "lastDeployResult": state.DeployResult.DEPLOYED.name,
+                    "isDeploying": False,
+                    "complianceState": state.Compliance.NON_COMPLIANT.name,
                 }
             },
         },


src/inmanta/const.py
```diff
@@ -57,6 +57,8 @@ class ResourceState(str, Enum):
     cancelled = "cancelled"  # When a new version is pushed, in progress deploys are cancelled
     undefined = "undefined"  # The state of this resource is unknown at this moment in the orchestration process
     skipped_for_undefined = "skipped_for_undefined"  # This resource depends on an undefined resource
+    # Used to report non-compliance of reporting resource.
+    non_compliant = "non_compliant"
```
wouterdb: This changes the range of expected input values for every handler. We should add this to the upgrade notes tht if you have handlers that override `deploy`, they may get a new unput value

src/inmanta/deploy/scheduler.py
wouterdb: I wonder if this should be a method in the the `ResourceState` class?

src/inmanta/deploy/scheduler.py
```diff
@@ -1397,12 +1399,16 @@ async def _update_scheduler_state_for_finished_deploy(
                 return state.copy()
 
             # We are not stale
-            state.compliance = Compliance.COMPLIANT if deploy_result is DeployResult.DEPLOYED else Compliance.NON_COMPLIANT
+            # We use result.resource_state instead of deploy_result because result.resource_state excludes non-compliant reports
```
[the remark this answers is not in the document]
jptrindade: > deploy_result is the wrong state machine to use here, right?

Yes
I wanted to say that `deploy_result=DeployResult.DEPLOYED` != `result.resource_state=HandlerResourceState.deployed` 
We now have that fresh in our memories but we may forget that that is the case.

src/inmanta/deploy/state.py
```diff
@@ -204,6 +204,25 @@ def copy(self: Self) -> Self:
...
+                return const.ResourceState.available
+            case ResourceState(last_deploy_result=DeployResult.SKIPPED):
+                return const.ResourceState.skipped
+            case ResourceState(last_deploy_result=DeployResult.FAILED):
+                return const.ResourceState.failed
+            case ResourceState(compliance=Compliance.NON_COMPLIANT):
+                return const.ResourceState.non_compliant
+            case ResourceState(last_deploy_result=DeployResult.DEPLOYED):
+                return const.ResourceState.deployed
+            case _:
```
sanderr: Have you tried `typing.assert_never` here? I don't recall off the top of my head if mypy can handle it, but I think it might now that we extended it to cover all `Compliance` scenarios. If so, we don't need the catch-reraise in the caller.

src/inmanta/deploy/state.py
```diff
@@ -472,13 +492,10 @@ def should_skip_for_dependencies(self, resource: "ResourceIdStr") -> bool:
...
         """
         dependencies: Set[ResourceIdStr] = self.requires.get(resource, set())
-        return any(
-            # Determine based on latest deploy result rather than compliance status, because compliance status is unstable
-            # (e.g. HAS_UPDATE).
-            # Make sure to not skip on NEW (never deployed) resources, because they will not generate a discovery event.
-            self.resource_state[dep_id].last_deploy_result not in (DeployResult.DEPLOYED, DeployResult.NEW)
-            for dep_id in dependencies
-        )
+        for dep_id in dependencies:
```
sanderr: Slightly preferred the `any` approach for readability, but not really a change request.

tests/deploy/e2e/test_deploy.py
```diff
@@ -1669,6 +1727,74 @@ async def wait_for_resource_state() -> bool:
     # because that method waits only until all resources reach a deployed state, not necessarily until the scheduler is stable
     await retry_limited(wait_for_resource_state, timeout=10)
 
+    # key1 will no longer skip, but we switch to report mode, but the resource was never deployed so it will be non-compliant
+    # key2 will also switch to report mode, but previous version already deployed so it will be compliant
+    # key3 requires key1 and will skip because it is non-compliant
```
sanderr:
```suggestion
    # key3 switches to enforcing mode; requires key1 and will skip because it is non-compliant
```

tests/server/test_environment_metrics.py
```diff
@@ -477,11 +477,11 @@ async def test_resource_count_metric(clienthelper, client, agent):
     metrics_service.register_metric_collector(metrics_collector=rcmc)
 
     # flush the metrics for the first time:
-    # create 30 records: (3 envs * 10 statuses)
+    # create 30 records: (3 envs * 11 statuses)
```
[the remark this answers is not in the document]
sanderr: Oh, those are the resource statuses. Ok, clear!

tests/conftest.py
```diff
@@ -2958,6 +2960,7 @@ def resource_id(ri: int) -> str:
                     "purged": False,
                     "requires": [] if ri % 2 == 0 else [resource_id(ri - 1)],
                     "my_attribute": iteration,
+                    "report_only": True if ri in (6, 7) else False,
```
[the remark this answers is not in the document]
jptrindade: Nope, just found that out. Will remove it

tests/conftest.py
```diff
@@ -3013,21 +3016,23 @@ async def deploy(resource: dict[str, object]) -> None:
                 if "sub=4]" in rid:
                     # never finish deploying r4
                     return
+                expected_resource_state: const.HandlerResourceState
```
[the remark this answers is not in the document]
jptrindade: yes, I'll find another name for it

sanderr (commented):
Latest commit looks good to me. EDIT: Bit of a redundant comment. I was confused about Slack notifications. Thought you had rerequested review.

wouterdb (approved):

PR #9750, opened 2025-11-18
Make the discovery_resource_id of a discovered resource mandatory

### LINKED ISSUE #8004: Phase out un-linked discovered resources.
Follow up to https://github.com/inmanta/inmanta-core/issues/7528.

The original ticket was about adding functionality to make sure newly created discovered resources were properly linked to the discovery resource while remaining backwards compatible with resource that were discovered prior to this version (i.e. unlinked resources).

https://github.com/inmanta/inmanta-core/pull/7912 added:
- The `discovery_resource_id: Optional[ResourceId]` field to the `DiscoveredResource` class
- The `class LinkedDiscoveredResource(DiscoveredResource)` subclass with the same field but typed as a non-optional `ResourceId`

This ticket is about phasing out the old behaviour and making the linking to the discovery resource mandatory for all discovered resources i.e. including (but not limited to):

- On the DB side, making the `discovery_resource_id` column `NOT NULL`
- On the model side, removing  the `LinkedDiscoveredResource` class and typing the `discovery_resource_id: Optional[ResourceId]` field to the `DiscoveredResource` class as non-optional

arnaudsjs:
# Description

This PR makes the discovery_resource_id column of the discoveredresource table mandatory. 

closes #8004

# Self Check:

- [x] Changelog entry

### DIFF
--- a/tests/db/migration_tests/test_v202511030_to_v202511180.py
+++ b/tests/db/migration_tests/test_v202511030_to_v202511180.py
@@ -54,3 +54,10 @@ async def test_make_discovery_resource_id_column_mandatory(
         WHERE discovered_resource_id='discovery::Discovered[myagent,name=discovered]'
         """
     )
+    assert "discovery::Discovery[discovery,name=discoverer]" == await postgresql_client.fetchval(
+        f"""
+        SELECT discovery_resource_id
+        FROM {data.DiscoveredResource.table_name()}
+        WHERE discovered_resource_id='discovery::deep::submod::Dis-co-ve-red[my-agent,name=NameWithSpecial!,[::#&^@chars]'
+        """
+    )


src/inmanta/db/versions/v202511180.py
```diff
@@ -0,0 +1,36 @@
...
+
+async def update(connection: Connection) -> None:
+    """
+    Make discovery_resource_id a mandatory column.
+    """
+    schema = """
+    -- Drop all discovered resources that were creating using the previous major
+    -- release of the orchestrator.
+    DELETE FROM public.discoveredresource
+    WHERE discovery_resource_id IS NULL;
```
sanderr: The ticket stated "Set null values to some sort of legacy / unknown resource id". I don't have a very strong opinion, but that does seem safest to me.

sanderr (changes requested):

PR #9757, opened 2025-11-19
Make graphql endpoint return all settings and definitions

### LINKED ISSUE #9521: Make graphql endpoint return all settings
Querying for `environments.settings` using the GraphQL endpoint has a few problems. It just returns the content of the `environment.settings` column in the database. This means that:

* It only returns the settings that were explicitly set by the user.
* It doesn't return the definitions information (default value, docstring, etc.).
* The protected and protected_by properties of the settings are present in the same json container as the value of the setting. This is something the frontend team doesn't like.

This ticket should fix the above-mentioned issue. What the returned response should look like, has to be defined by the frontend team. I created this ticket because the frontend team currently doesn't have time to look into this. This ticket must be handled before we release the GraphQL endpoint, i.e. before the iso9 release.

jptrindade:
# Description

Return all settings instead of just the ones saved in the database.
If not present in the database, assign the default value.
Also return metadata on each setting.

closes #9521 

# Self Check:

- [x] Changelog entry


tests/test_graphql.py
```diff
@@ -194,6 +195,55 @@ async def test_graphql_schema(server, client):
...
+    for setting_name in data.Environment._settings.keys():
+        setting_value = settings["settings"][setting_name]
+        if setting_name in modified_settings:
+            assert setting_value["value"] == modified_settings[setting_name]
+            assert setting_value["protected"]
+            assert setting_value["protected_by"] == model.ProtectedBy.project_yml
+        else:
+            assert setting_value["value"] == api_result.result["metadata"][setting_name]["default"]
+            assert setting_value["protected"] is False
+            assert setting_value["protected_by"] is None
```
arnaudsjs: I think we also have to assert the definitions part of the result so that we can be certain it doesn't change.

arnaudsjs: Indeed. I completely missed that line.

arnaudsjs (approved):

src/inmanta/graphql/schema.py
```diff
@@ -338,6 +339,26 @@ def get_is_compiling(root: "Environment", info: strawberry.Info) -> bool:
     return compiler_service.is_environment_compiling(environment_id=root.id)
 
 
+def get_settings(root: "Environment") -> scalars.JSON:
+    assert hasattr(root, "settings")
```
Hugo-Inmanta: Could use a small docstring

src/inmanta/graphql/schema.py
```diff
@@ -338,6 +339,26 @@ def get_is_compiling(root: "Environment", info: strawberry.Info) -> bool:
...
+    for key, setting_info in data.Environment._settings.items():
+        if stored_setting := modified_settings.get(key, None):
+            setting_values[key] = model.EnvironmentSettingDetails(
+                value=stored_setting["value"],
+                protected=stored_setting.get("protected", False),
+                protected_by=stored_setting.get("protected_by", None),
+            )
+        else:
+            default_value = setting_info.default
+            assert default_value is not None  # Should never happen but setting_info.default is Optional
```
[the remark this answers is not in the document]
jptrindade: I don't think it is strictly mandatory to have a default value, but I've only seen instances of `Setting` with default provided. And `EnvironmentSettingDetails` requires value to not be null. I prefer having this assertion than having a "default" for when no `default` is provided

PR #9772, opened 2025-11-25
Drop references to V1 modules from documentation

arnaudsjs:
# Description

Drop references to V1 modules from documentation.

# Self Check:

- [x] Changelog entry


changelogs/unreleased/drop-references-v1-modules-in-docs.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: "Remove references to V1 modules from the documentation."
+change-type: patch
+destination-branches: [patch]
```
sanderr:
```suggestion
destination-branches: [master]
```

PR #9816, opened 2025-12-09
Add endpoints to delete discovered resources

### LINKED ISSUE #9633: add endpoint to delete discovered resource
- endpoint to delete single one
- batch endpoint to delete multiple

jptrindade:
# Description

Added endpoints to delete discovered resources

closes #9633 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/protocol/methods_v2.py
+++ b/src/inmanta/protocol/methods_v2.py
@@ -1837,6 +1837,9 @@ def discovered_resource_delete(
 def discovered_resource_delete_batch(tid: uuid.UUID, discovered_resource_ids: Sequence[str]) -> None:
     """
     Delete multiple discovered resources.
+    If one or more discovered resources are not found on the environment,
+    they will be ignored and the other discovered resources will be deleted.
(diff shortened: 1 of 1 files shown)


src/inmanta/protocol/methods_v2.py
```diff
@@ -1804,6 +1804,45 @@ def discovered_resources_get_batch(
...
+    tid: uuid.UUID,
+    discovered_resource_id: str,
+) -> None:
+    """
+    Delete a discovered resource.
+
+    :param tid: The id of the environment this resource belongs to
+    :param discovered_resource_id: The id of the discovered_resource
+
+    :raise NotFound: This exception is raised when the referenced discovered resource is not found on the environment
```
arnaudsjs: This message is used on the API documentation, where the name of this exception is translated to the corresponding return code 404. So the term "exception" doesn't apply there.
```suggestion
    :raise NotFound: When the referenced discovered resource is not found on the environment
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #9859, opened 2026-01-02
Fixed issues with mypy on graphql

jptrindade:
# Description

Addresses the mypy issues caused by graphql. 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/fix-mypy-issues-graphql.yml
+++ b/changelogs/unreleased/fix-mypy-issues-graphql.yml
@@ -1,0 +1,4 @@
+---
+description: "Fix mypy issues on graphql"
+change-type: patch
+destination-branches: [master]


src/inmanta/graphql/schema.py
```diff
@@ -360,7 +366,7 @@ def get_settings(root: "Environment") -> scalars.JSON:
             assert default_value is not None  # Should never happen but setting_info.default is Optional
             setting_values[key] = model.EnvironmentSettingDetails(value=default_value)
     setting_definitions = dict(sorted(data.Environment.get_setting_definitions_for_api(setting_values).items()))
-    return {"settings": setting_values, "definition": setting_definitions}
+    return cast(scalars.JSON, {"settings": setting_values, "definition": setting_definitions})
```
sanderr: Iirc this is a `NewType` so there is no need to cast.

```suggestion
    return scalars.JSON({"settings": setting_values, "definition": setting_definitions})
```

sanderr (changes requested):

sanderr (approved):

PR #9874, opened 2026-01-08
Add compliance report endpoint

### LINKED ISSUE #9871: Add endpoint to retrieve report
Design: https://github.com/inmanta/designs/commit/dc3d2bbd7adc5db4cf90c3905040dae70bb588c2#diff-848ecd1b6dbfa797e2b19465ea932ad5186164739a4d4e03e8671b629e752c41R235
Related to: https://github.com/inmanta/inmanta-core/pull/9862
TODOs:
Flesh out tests to make sure that the diffs are properly updated
https://github.com/inmanta/inmanta-core/pull/9862/changes#diff-bb66f725072138a9e2a93a67ede34863b1b0d0ea7036b8203439b72dda0a11a6R38
https://github.com/inmanta/inmanta-core/pull/9862/changes#diff-b8fa4262c32dd239c16b2938a7b592a125554992f29f1504f070a903048b5feaR1869

jptrindade:
# Description

Updated the dumps to include 2 non compliant resources. (The last dump of 2025 already had 2 non compliant resources)
The initial migration produced diffs in a format different than what we do on `send_deploy_done` so I fixed it.

closes #9871 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5068,7 +5068,7 @@ class ResourcePersistentState(BaseDocument):
                     ),
                     compliance=compliance_status,
                     last_execution_result=state.DeployResult(str(record["last_deploy_result"]).lower()),
-                    last_execution=cast(datetime.datetime | None, record["last_executed_at"]),
+                    last_executed_at=cast(datetime.datetime | None, record["last_executed_at"]),
                 )
             return diff
 
--- a/src/inmanta/data/model.py
+++ b/src/inmanta/data/model.py
@@ -677,14 +677,14 @@ class ResourceComplianceDiff(BaseModel):
     :param report_only: Is this resource in report only mode.
     :param compliance: The current compliance of this resource.
     :param last_execution_result: The result of the last execution of this resource.
-    :param last_execution: The timestamp of the last execution of this resource.
+    :param last_executed_at: The timestamp of the last execution of this resource.
     :param attribute_diff: The diff between the attributes of the current and desired state of a non_compliant resource.
     """
 
     report_only: bool
     compliance: Compliance
     last_execution_result: DeployResult
-    last_execution: datetime.datetime | None
+    last_executed_at: datetime.datetime | None
     attribute_diff: dict[str, AttributeStateChange] | None
 
 
--- a/tests/deploy/e2e/test_deploy.py
+++ b/tests/deploy/e2e/test_deploy.py
@@ -1914,7 +1914,7 @@ async def test_non_compliant_diff(resource_container, server, client, clienthelp
         attribute_diff=None,
         compliance=Compliance.COMPLIANT,
         last_execution_result=DeployResult.DEPLOYED,
-        last_execution=rps.last_deploy,
+        last_executed_at=rps.last_deploy,
     )
 
     assert report[rid2] == ResourceComplianceDiff(
@@ -1922,7 +1922,7 @@ async def test_non_compliant_diff(resource_container, server, client, clienthelp
         attribute_diff=None,
         compliance=Compliance.COMPLIANT,
         last_execution_result=DeployResult.DEPLOYED,
-        last_execution=rps2.last_deploy,
+        last_executed_at=rps2.last_deploy,
     )
 
     # Make rid1 reporting and change the desired state
@@ -1964,7 +1964,7 @@ async def test_non_compliant_diff(resource_container, server, client, clienthelp
         attribute_diff={"value": AttributeStateChange(current="actual_value", desired="diff_value")},
         compliance=Compliance.NON_COMPLIANT,
         last_execution_result=DeployResult.DEPLOYED,
-        last_execution=rps.last_deploy,
+        last_executed_at=rps.last_deploy,
     )
 
     # Make report succeed again
@@ -1994,7 +1994,7 @@ async def test_non_compliant_diff(resource_container, server, client, clienthelp
         attribute_diff=None,
         compliance=Compliance.COMPLIANT,
         last_execution_result=DeployResult.DEPLOYED,
-        last_execution=rps.last_deploy,
+        last_executed_at=rps.last_deploy,
     )
 
     # Make report fail again
@@ -2025,7 +2025,7 @@ async def test_non_compliant_diff(resource_container, server, client, clienthelp
         attribute_diff={"value": AttributeStateChange(current="actual_value", desired="another_diff_value")},
         compliance=Compliance.NON_COMPLIANT,
         last_execution_result=DeployResult.DEPLOYED,
-        last_execution=rps.last_deploy,
+        last_executed_at=rps.last_deploy,
     )
 
 


changelogs/unreleased/9871-add-compliance-report-endpoint.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: "Add get_compliance_report endpoint."
```
Hugo-Inmanta:
```suggestion
description: "Add `GET /api/v2/compliance_report` endpoint."
```

src/inmanta/protocol/methods_v2.py
```diff
@@ -1889,3 +1889,25 @@ def health() -> ReturnValue[None]:
...
+
+
+@auth(auth_label=const.CoreAuthorizationLabel.REPORT_READ, read_only=True, environment_param="tid")
+@typedmethod(
+    path="/get_compliance_report/",
+    operation="POST",
+    agent_server=True,
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.agent],
+    api_version=2,
```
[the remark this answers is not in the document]
jptrindade: I believe a POST might be safer because of the parameters we require. In a GET we would have to send it in the URL, in POST we can send those in the body

src/inmanta/data/__init__.py
```diff
@@ -5009,6 +5009,72 @@ def get_compliance_status(self) -> Optional[state.Compliance]:
...
+            diff: dict[ResourceIdStr, m.ResourceComplianceDiff] = {}
+            for record in result:
+                compliance_status = state.get_compliance_status(
+                    is_orphan=False,  # We filter out orphan resources in the query
+                    is_undefined=cast(bool, record["is_undefined"]),
+                    last_deployed_attribute_hash=cast(str | None, record["last_deployed_attribute_hash"]),
+                    current_intent_attribute_hash=cast(str | None, record["current_intent_attribute_hash"]),
+                    last_deploy_compliant=cast(bool, record["last_deploy_compliant"]),
+                )
+                assert compliance_status is not None  # make mypy happy
```
sanderr: We could achieve this with a `typing.overload`. Not a requirement per se.

src/inmanta/db/versions/v202601130.py
```diff
@@ -0,0 +1,33 @@
...
+Contact: code@inmanta.com
+"""
+
+from asyncpg import Connection
+
+
+async def update(connection: Connection) -> None:
+    """
+    This migration fixes the diff layout on resource_diff.
+    {rid: {current: x, desired: y}} -> {current: x, desired: y}
```
sanderr: If we need to fix the layout, that implies that the previous implementation wrote it incorrectly. Shouldn't we fix that then? Or was it the previous migration script that wrote it incorrectly, and not the runtime implementation?

src/inmanta/protocol/methods_v2.py
```diff
@@ -1889,3 +1889,24 @@ def health() -> ReturnValue[None]:
     In contrast to the 'GET /api/v1/serverstatus' endpoint, this endpoint does not require authentication.
     """
     pass
+
+
+@auth(auth_label=const.CoreAuthorizationLabel.REPORT_READ, read_only=True, environment_param="tid")
+@typedmethod(
+    path="/get_compliance_report/",
```
Hugo-Inmanta: For consistency with other endpoints (not a strong opinion)
```suggestion
    path="/compliance_report",
```

src/inmanta/data/__init__.py
```diff
@@ -5046,21 +5046,23 @@ async def get_compliance_report(
...
                     last_deployed_attribute_hash=cast(str | None, record["last_deployed_attribute_hash"]),
                     current_intent_attribute_hash=cast(str | None, record["current_intent_attribute_hash"]),
                     last_deploy_compliant=cast(bool, record["last_deploy_compliant"]),
                 )
-                assert compliance_status is not None  # make mypy happy
                 report_only = cast(bool, record["report_only"])
-                non_compliant_resource = report_only and compliance_status is state.Compliance.NON_COMPLIANT
+                non_compliant_resource = (
+                    report_only
+                    and compliance_status is state.Compliance.NON_COMPLIANT
```
sanderr: Isn't this alone sufficient?

jptrindade: It is not sufficient because this is not the "handler state" where we differentiate between `failed` and `non_compliant`. If we omitted the last check, we couldn't diferentiate between failed and non_compliant report_only resources

Hugo-Inmanta (approved):

src/inmanta/data/__init__.py
sanderr: Oh, I read over this too quickly. Why not simply use the handler state? A single check would suffice there.

sanderr: Don't give this too much weight though. I'm still not 100% focused.

sanderr: The main thing for me is that we present a consistent story, and that our checks align 100% with that. And preferably directly rather than via a roundabout way that may rely on implementation details. So it all comes down to that invariant, which ties to the docstring that I mentioned earlier.
What is the new table meant to contain? When do we write to it? Under which circumstances do we consider a foreign key to it to be meaningful? I think those are our three aspects. And I'm pretty sure they're not currently aligned. Off the top of my head, we write based on the handler state, we read based on these three conditions, and we declare something about it being meaningful based on `state.Compliance`. That is the sort of stuff that leads to confusion and bugs down the road, because motivation gets lost if there is no consistent story.

jptrindade: I could simplify this to check for `state.Compliance.NON_COMPLIANT` [here](https://github.com/inmanta/inmanta-core/pull/9874/changes#diff-eb83b3ad0c71572a853e5bc121410b7ff2c0d1cfa124d083aabc5397528405f5R5069) because:
- we would discard `HAS_UPDATE` / `UNDEFINED` / `COMPLIANT`
- any other case that is not a report_only resource which reported non_compliant  (failed/skipped/skipped_for_undefined) would have `non_compliant_diff` set to `None` on `send_deploy_done`

I would then update the docstring to say:
`    :param attribute_diff: The diff between the attributes of the current
        and desired state of a non_compliant report_only resource`
        
Although we don't need to check specifically for `report_only` 

I think I'm not missing any edge case. wdyt?

sanderr (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -5059,22 +5059,16 @@ async def get_compliance_report(
...
-                    resource_state=state.ResourceState(
-                        compliance=compliance_status,
-                        last_deploy_result=state.DeployResult(str(record["last_deploy_result"]).lower()),
-                        blocked=state.Blocked(str(record["blocked"]).lower()),
-                        last_deployed=cast(datetime.datetime | None, record["last_executed_at"]),
-                        last_deploy_compliant=cast(bool, record["last_deploy_compliant"]),
-                    ),
+                    compliance=compliance_status,
+                    last_execution_result=state.DeployResult(str(record["last_deploy_result"]).lower()),
+                    last_execution=cast(datetime.datetime | None, record["last_executed_at"]),
```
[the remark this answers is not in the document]
jptrindade: Forgot we already had a good name in `last_executed_at`. Basically I translated the original field (`last_deploy` -> `last_execution`) but I like `last_executed_at` better

sanderr (approved):

PR #9881, opened 2026-01-12
 drop compiler version [REMADE]

### LINKED ISSUE #4002: drop compiler version for iso9
# Task

- [ ] Check if the compiler version field exists in the v2 module metadata format (`inmanta.module.ModuleV2Metadata`?)
- [ ] Make sure we don't use the compiler version field when compiling a model that still uses a v1 module (e.g. core test suite)
- [ ] Drop the `COMPILER_VERSION` field. See if anything breaks. If so, report back so we can decide whether to revert or fix the failures.
- [ ] Update the release procedure. Check both the normal and the major release checklists.

# Original description

With iso3-4-5 the compiler version was decoupled from both the core and the product versions. It continues to use the `<year>.<minor>` versioning scheme. As such, there is no easy way to make it so that iso6's compiler version is unreachable by iso5. Luckily, the compiler version is really only relevant for v1 modules, so I would propose to drop support for it in iso6. If a module specifies a minimal compiler version we could show a warning that says this is no longer supported and to use a v2 module instead if this is required. Since iso6 is strictly more recent than iso5, this should be backwards compatible with older modules (the minimal compiler version they have will never be greater than the first iso6 one). Only after the iso6 release will modules that require newer compiler features be required to update to v2 (that is, if we don't drop support for v1 in the first place).

I would propose to do this ASAP, so we don't need to manage the compiler version accross two branches.

A potential difficulty here would be the handling of cf cache files, which currently use the compiler version as metadata. Perhaps the core package version can be used instead?

Another disadvantage of removing it while we still have v1 modules is that we can no longer restrict them to a minimal compiler version (v2 modules can use Python constraints, v1 modules can't).

jptrindade:
# Description

Remade this PR to only drop the COMPILER_VERSION const as to not break anything

closes #4002 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/4002-drop-compiler-version.yml
+++ b/changelogs/unreleased/4002-drop-compiler-version.yml
@@ -1,10 +1,10 @@
 ---
 description: >
-  Drop COMPILER_VERSION field.
+  Drop inmanta.COMPILER_VERSION field.
 issue-nr: 4002
 issue-repo: inmanta-core
 change-type: minor
 destination-branches: [master]
 sections:
-  minor-improvement: "{{description}}"
+  upgrade-note: "{{description}}"
   deprecation-note: "Deprecated compiler_version field on put_version endpoint"
--- a/src/inmanta/parser/cache.py
+++ b/src/inmanta/parser/cache.py
@@ -20,6 +20,7 @@ import logging
 import os
 from typing import Optional
 
+from inmanta import __version__ as inmanta_version
 from inmanta.ast import Namespace
 from inmanta.ast.statements import Statement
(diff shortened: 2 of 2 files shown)


src/inmanta/util/__init__.py
```diff
@@ -73,7 +73,10 @@
 
 
 def get_compiler_version() -> str:
-    return COMPILER_VERSION
+    """
+    [DEPRECATED] Wasn't removed yet for backward compatibility reasons.
```
sanderr: Can you elaborate on this one? Where is this one used that is non-trivial to address?

PR #9889, opened 2026-01-13
Issue/9837 dataclass reference nullables

### LINKED ISSUE #9837: Bug for nullable attributes in dataclass references
# Problem

When a dataclass has a nullable attribute, and a plugin returns a reference to such a dataclass, the compiler tries to convert it to a dataclass instance with reference attributes. However, it constructs an incorrect type for the attribute, and as a result rejects the value on the plugin boundary, raising an exception.

The root cause is that in our attempt to allow references in any model attribute, we construct nullable attribute types as `Nullable(OrReferenceType(base_type))` rather than `OrReferenceType(Nullable(base_type))`, meaning we don't allow references to null, or by extension to `Optional`.

On the plugin boundary, we wrap this type in a reference, so we end up with `Reference(Nullable(OrReferenceType(base_type)))`. The attribute validation then validates if this is appropriate for the attribute. It is not, because we've built a reference to a nullable (more or less, ignoring the nested reference), while the attribute only allows null or a reference to the base type.

# PoC fix

```diff
diff --git a/src/inmanta/ast/attribute.py b/src/inmanta/ast/attribute.py
index c041fbd1b..9d5beb3e7 100644
--- a/src/inmanta/ast/attribute.py
+++ b/src/inmanta/ast/attribute.py
@@ -55,6 +55,13 @@ class Attribute(Locatable):
             self.__type_internal = inmanta_type.TypedList(self.__type_internal)
         if nullable:
             self.__type_internal = inmanta_type.NullableType(self.__type_internal)
+        if not value_type.is_entity():
+            # TODO: this is tough.
+            # - It has to be on the outside (as in this patch) for nullable, so that we allow references to None
+            #   => add test case with dc.attr: int? and from_python DC conversion (A.dc = get_dc_ref())
+            # - (not verified) It has to be (at least) on the inside (as before this patch) for [], so that we allow references *inside* lists
+            #   => add test case with int[] attr, and attr=[myref] assignment
+            self.__type_internal = inmanta_type.OrReferenceType(self.__type_internal)

         # Drop all reference for backward compatibility
         self.__type: Type = self.__type_internal.get_no_reference()
diff --git a/src/inmanta/ast/statements/define.py b/src/inmanta/ast/statements/define.py
index dc666b33c..81e162333 100644
--- a/src/inmanta/ast/statements/define.py
+++ b/src/inmanta/ast/statements/define.py
@@ -46,7 +46,7 @@ from inmanta.ast.attribute import Attribute, RelationAttribute
 from inmanta.ast.blocks import BasicBlock
 from inmanta.ast.entity import Entity, Implement, Implementation
 from inmanta.ast.statements import BiStatement, ExpressionStatement, Literal, Statement, TypeDefinitionStatement
-from inmanta.ast.type import TYPES, ConstraintType, NullableType, OrReferenceType, Type, TypedList
+from inmanta.ast.type import TYPES, ConstraintType, NullableType, Type, TypedList
 from inmanta.execute.runtime import ExecutionUnit, QueueScheduler, Resolver, ResultVariable
 from inmanta.plugins import Plugin

@@ -196,7 +196,7 @@ class DefineEntity(TypeDefinitionStatement):
                 name = str(attribute.name)
                 attr_obj = Attribute(
                     entity_type,
-                    OrReferenceType(attribute.type.get_basetype(self.namespace)),
+                    attribute.type.get_basetype(self.namespace),
                     name,
                     attribute.get_location(),
                     attribute.type.multi,
diff --git a/src/inmanta/execute/proxy.py b/src/inmanta/execute/proxy.py
index c57bf44e7..58fefe07c 100644
--- a/src/inmanta/execute/proxy.py
+++ b/src/inmanta/execute/proxy.py
@@ -224,6 +224,7 @@ class DynamicProxy:
                     f"make sure the reference extends Reference[concrete_type] or override `get_reference_type`",
                 )

+            # TODO: support Reference[DC | None] -> #9872
             if dataclasses.is_dataclass(ref_type):
                 if dynamic_context is None:
                     raise RuntimeException(
```

sanderr:
# Description

Fixed dataclass boundary conversion from `Reference[Dataclass]` to dataclass with reference attributes, when the dataclass has nullable attributes (see `test_references.py` tests case).

The core of the fix is twofold:
1. restructure the wrapping in `OrReference`, `NullableType` and `TypedList` so that we allow references to `null`, rather than `null` or references to the non-null type. And similar for lists. This is in `inmanta.ast.attribute` and `inmanta.ast.statements.define`.
2. improved type validation in `inmanta.ast.type`. so that `NullableType(TypedList(String()))` is considered a subtype of `OrReferenceType(NullableType(TypedList(OrReferenceType(String()))))` (and similar for nullable without lists). See `test_typing_for_plugins.py` test case. This is required because the dataclass boundary conversion in `inmanta.ast.entity` (search for `ar._model_type = ...`) declares that it is a reference to the former, while the attribute is the second (wider) type.

Especially 2 required me to think critically about `Type.issubtype` and `Type.issupertype` and their contracts, which weren't sufficiently well defined. I redefined them in a way that I feel makes more sense overall and reworked some of their implementations.
As a result of 1, there was also an opportunity to drop `get_no_references()`, which was a bit of a mess anyway, and to redefine `supports_references()` to be more in line with its intended usage.

closes #9837

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/ast/attribute.py
+++ b/src/inmanta/ast/attribute.py
@@ -62,6 +62,10 @@ class Attribute(Locatable):
 
     @classmethod
     def _wrap_type(cls, base_type: "Type", *, multi: bool, nullable: bool, with_references: bool) -> "Type":
+        """
+        Wraps the given base type with `[]`, `?` and `| Reference[...]` where appropriate, depending on the parameters.
+        Does not take cls.SUPPORTS_REFERENCES into account, i.e. expects caller to do so.
+        """
         result: Type = base_type
         if multi:
             if with_references:


src/inmanta/ast/attribute.py
```diff
@@ -49,19 +52,29 @@ def __init__(
...
+        self.__type: Type = self._wrap_type(value_type, multi=multi, nullable=nullable, with_references=False)
+        self.__type_internal: Type = self._wrap_type(
+            value_type, multi=multi, nullable=nullable, with_references=self.SUPPORTS_REFERENCES
+        )
 
         self.comment = None  # type: Optional[str]
         self.end: Optional[RelationAttribute] = None
 
+    @classmethod
+    def _wrap_type(cls, base_type: "Type", *, multi: bool, nullable: bool, with_references: bool) -> "Type":
```
Hugo-Inmanta: This could use a docstring

src/inmanta/ast/type.py
```diff
@@ -183,24 +179,35 @@ def to_python(self, instance: object, *, path: str) -> object:
...
-
     def issubtype(self, other: "Type") -> bool:
         """
         Returns True iff this DSL type is a subtype of the other type.
+
+        Implementations may recurse but must do so in a way that progresses and will eventually terminate, i.e. one of:
+            - narrow on the self type, e.g. recurse on each of the options in a union / intersection type
+            - decend deeper into both types, e.g. if both are typed lists, check if their respective element types have a
+                subtype relation.
+        Implementations that do not return True or recurse, should fall back to calling issupertype with reversed arguemnts,
```
Hugo-Inmanta:
```suggestion
            - descend deeper into both types, e.g. if both are typed lists, check if their respective element types have a
                subtype relation.
        Implementations that do not return True or recurse, should fall back to calling issupertype with reversed arguments,
```

src/inmanta/ast/type.py
```diff
@@ -544,8 +556,6 @@ def __eq__(self, other: object) -> bool:
         return type(self) == type(other)  # noqa: E721
 
     def issupertype(self, other: "Type") -> bool:
-        if other == self:
-            return False
         return True
```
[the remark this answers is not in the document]
Hugo-Inmanta: Got it, makes sense

PR #9905, opened 2026-01-19
Save generic errors during config loading to compile logs

### LINKED ISSUE #9811: Compile logs doesn't contain config loading error
When exporting the service catalog, when the server authentication is enabled, the authentication config file is loaded.  If there is something wrong with that config file (i.e. error below), the error is not part of the compile output, it is only present in the server logs, but the compile itself fails.  We should have either the original error or a derivated error in the compile logs to give an indication of the cause of the failed compile:

```
2025-12-08 11:41:30,484 ERROR    inmanta.server.services.compilerservice An error occurred while recompiling
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/server/services/compilerservice.py", line 499, in run
    token = encode_token(["compiler", "api"], str(environment_id))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/protocol/auth.py", line 137, in encode_token
    cfg = AuthJWTConfig.get_sign_config()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/protocol/auth.py", line 336, in get_sign_config
    cls.list()
  File "/opt/inmanta/lib/python3.12/site-packages/inmanta/protocol/auth.py", line 317, in list
    raise ValueError("One auth_jwt section should have sign set to true")
ValueError: One auth_jwt section should have sign set to true
```

jptrindade:
# Description

If we got an error loading the configuration, it would not be present in the compile logs, only the server logs so the user would see his compile failing with no explanation on the logs. This was because the error was happening outside of a specific compile stage. I tried to fix this by adding another stage when loading the config, but I don't know if we can actually add these stages "as we please"

closes #9811 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -399,8 +399,6 @@ class CompileRun:
                 export_command.append("--ssl-ca-cert")
                 export_command.append(ssl_ca_cert)
 
-            self.tail_stdout = ""
-
             # Make mypy happy
             assert self.request.used_environment_variables is not None
             env_vars_compile: dict[str, str] = os.environ.copy()
@@ -521,6 +519,7 @@ class CompileRun:
                 if stage_result and (stage_result.returncode is None or stage_result.returncode > 0):
                     return False, None
 
+            self.tail_stdout = ""
             result: data.Report = await run_compile_stage_in_venv(
                 "Recompiling configuration model", cmd, cwd=project_dir, env=env_vars_compile
             )
@@ -542,7 +541,7 @@ class CompileRun:
 
         except Exception as e:
             if self.stage is not None:
-                await self._error(message=f"An error occurred while recompiling: \n {str(e)}")
+                await self._error(message=str(e))
                 await self._end_stage(1)
             else:
                 LOGGER.exception("An error occurred while recompiling")


changelogs/unreleased/9811-compile-logs-dont-contain-config-loading-error.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: "Save config loading errors during compile to compile logs"
+issue-nr: 9811
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master]
```
Hugo-Inmanta: Can probably go into iso8 as well ?

src/inmanta/server/services/compilerservice.py
```diff
@@ -539,9 +540,12 @@ async def setup() -> AsyncIterator[Awaitable[Optional[data.Report]]]:
             # unhandled exception in a backgrounded coroutine.
             pass
 
-        except Exception:
-            LOGGER.exception("An error occurred while recompiling")
-
+        except Exception as e:
+            if self.stage is not None:
+                await self._error(message=f"An error occurred while recompiling: \n {str(e)}")
```
sanderr: Other calls to `self._error` seem to only pass `str(e)`. Any reason to do it differently here?

jptrindade: It was just to maintain the existing syntax, but probably best to remove it.

PR #9950, opened 2026-01-28
Docs python version

sanderr:
# Description

I noticed we still had some references to 3.12 in the docs. I made them parametrizable based on the current venv's Python version. Since irt always uses the supported Python version for the product build, I think this is a good approach.

Semi-related internal docs PR: https://code.inmanta.com/inmanta/developerdocs/-/merge_requests/220

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/docs-python-version.yml
+++ b/changelogs/unreleased/docs-python-version.yml
@@ -1,0 +1,5 @@
+description: parametrized python version in docs
+change-type: patch
+destination-branches:
+  - master
+  - iso8


Hugo-Inmanta (changes requested):
I noticed we have one more occurrence [there](https://github.com/inmanta/inmanta-core/blob/7f22b68ef29c17e9a57be31eb8681af0772bc812/docs/administrators/on_disk_layout.rst?plain=1#L26-L27) but iirc replace directives might not behave too good inside code blocks. Probably not worth the hassle if it doesn't work out of the box.

arnaudsjs (approved):

PR #9989, opened 2026-02-06
Dynamically determine Postgresql version for docs

### LINKED ISSUE #9945: Dynamically determine the PostgresSQL version in the docs
The PostgreSQL installation documentation hard codes the version of PostgreSQL ([example](https://github.com/inmanta/inmanta-core/blob/8411ecc59f77e7fb7d3a539c6060857e30c8f872/docs/install/1-install-server.rst?plain=1#L267)). This ticket should make sure we rely on a placeholder in the documentation. The documentation build should replace that placeholder automatically with the correct PostgreSQL version.

jptrindade:
# Description

Dynamically determine Postgresql version for docs.
Substitution is not possible in code blocks, so I used parsed_literal instead. It is a bit annoying that we lose some highlighting but I believe we just have to live with it. Luckily we just have it in small code snippets

Requires: https://github.com/inmanta/irt/pull/2682

closes #9945 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,7 +58,7 @@ def get_pg_version_for_product() -> int:
     if pg_version is None:
         raise Exception(
             "INMANTA_MINIMAL_POSTGRES_VERSION not found."
-            "Please set it or use the INMANTA_DONT_DISCOVER_VERSION environment variable."
+            " Please set it or use the INMANTA_DONT_DISCOVER_VERSION environment variable."
         )
     return int(pg_version)
 


Hugo-Inmanta (changes requested):
> Substitution is not possible in code blocks, so I used parsed_literal instead.

Cool workaround 👍

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #10028, opened 2026-02-18
Issue/move from redoc style to swagger style

Hugo-Inmanta:
# Description

sphinx-redoc extension still uses pkg_resources. This PR moves away from it and uses swagger-ui instead.
This uniformizes how we document APIs from the Inmanta documentation and from the web-console.

In this PR:
- Update the makefile to generate the openapi html page and move it to `build/html/reference/openapi.html` after the docs are built.
- add `hide_response_from_docs_swagger` param and use it to not show the full open api schema in the docs endpoint (as per [slack discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1771411140441489))
- remove everything related to `sphinxcontrib-redoc`
- remove everything related to `_generate_code_sample`. Swagger-ui doesn't support code examples out of the box (afaik) and this mechanism was mainly used to display the method name to use when calling an endpoint via the python client (e.g. `get_scheduler_status` for the `/api/v2/scheduler_status` endpoint). We now get this for free thanks to `displayOperationId: true` in the openapi converter.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/protocol/openapi/converter.py
+++ b/src/inmanta/protocol/openapi/converter.py
@@ -129,7 +129,7 @@ class OpenApiConverter:
             path_item.__setattr__(http_method_name.lower(), operation)
         return path_item
 
-    def generate_openapi_json(self, swagger_description: str | None = None) -> str:
+    def generate_openapi_json(self, *, swagger_description: str | None = None) -> str:
         openapi = self.generate_openapi_definition(swagger_description=swagger_description)
 
         return json.dumps(openapi, default=openapi_json_encoder)
--- a/src/inmanta/server/server.py
+++ b/src/inmanta/server/server.py
@@ -177,7 +177,6 @@ class Server(protocol.ServerSlice):
         self,
         format: Optional[ApiDocsFormat] = ApiDocsFormat.swagger,
         token: str | None = None,
-        *,
         swagger_description: str | None = None,
     ) -> ReturnValue[Union[OpenAPI, str]]:
         url_map = self._server._transport.get_global_url_map(self._server.get_slices().values())


docs/Makefile
```diff
@@ -183,10 +184,8 @@ doctest:
 	@echo "Testing of doctests in the sources finished, look at the " \
 		  "results in $(BUILDDIR)/doctest/output.txt."
 
-.PHONY: openapi
```
Hugo-Inmanta: moved to L36

docs/Makefile
```diff
@@ -183,10 +184,8 @@ doctest:
...
 openapi:
-	pytest --rootdir=../tests ../tests/docs/extract_openapi_json.py
-	mv openapi.json reference/openapi.json
+	pytest --rootdir=../tests ../tests/docs/extract_openapi_html.py
```
arnaudsjs: I think we have to update the name of this target to "swagger" for example.

changelogs/unreleased/use-swagger-ui-instead-of-sphinx-redoc-for-core-api-documentation.yml
```diff
@@ -0,0 +1,4 @@
+---
+description: "Use swagger-ui instead of sphinx-redoc for core API documentation."
+change-type: patch
+destination-branches: [master, iso9, iso8, iso7]
```
sanderr: Do we want / need this in iso7 and 8? I don't have the full context for this PR so I can not judge. But I'm hesitant to have behavioral changes to the docs if we can avoid it. That said, it's only the docs, so I'm open to it if there are reasons.

Hugo-Inmanta: No real reason per se. I just figured it would make sense to use consistent tools across all iso lines. But I don't think that this is a big enough argument to mess with iso7 and 8.

arnaudsjs (changes requested):

src/inmanta/protocol/decorators.py
```diff
@@ -180,6 +181,11 @@ def typedmethod[C: Callable](
     :param varkw: If true, additional arguments are allowed and will be dispatched to the handler. The handler is
                   responsible for the validation.
     :param token_param: Use the authentication token in this parameter instead of using the Authorization header.
+    :param document_in_service_swagger: (LSM extension only, ignored otherwise) This parameter controls whether this
+    endpoint should be documented in the swagger served by the `lsm_services_openapi_docs` endpoint for each service
+    of the catalog.
```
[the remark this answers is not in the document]
Hugo-Inmanta: It does: https://github.com/inmanta/inmanta-core/blob/950c8e3b30811097861ca7d6ba57f8a74162c0c7/src/inmanta/protocol/decorators.py#L150

I added it a while back but forgot to document it here it seems.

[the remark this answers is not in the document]
arnaudsjs: Hehe. You are right. I somehow missed it.

docs/Makefile
```diff
@@ -68,12 +68,13 @@ clean:
 	-rm -rf $(PNGs)
 
 html: openapi
-	$(SPHINXBUILD) -b html -t include_redoc $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	$(SPHINXBUILD) -b html  $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	mv openapi.html build/html/reference/openapi.html
```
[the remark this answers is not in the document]
sanderr: No, I think it's fine

src/inmanta/server/server.py
```diff
@@ -174,15 +174,17 @@ async def health(self) -> ReturnValue[None]:
 
     @handle(methods_v2.get_api_docs)
     async def get_api_docs(
-        self, format: Optional[ApiDocsFormat] = ApiDocsFormat.swagger, token: str | None = None
+        self,
+        format: Optional[ApiDocsFormat] = ApiDocsFormat.swagger,
+        token: str | None = None,
+        *,
+        swagger_description: str | None = None,
```
sanderr: While I prefer it here as well from a style perspective. I think we have to make sure that our handlers are consistent with our methods, and I don't think we can do kw-only args in our methods.

sanderr (changes requested):

src/inmanta/protocol/openapi/converter.py
```diff
@@ -129,7 +129,7 @@ def _extract_operations_from_methods(self, api_methods: dict[str, UrlMethod], pa
             path_item.__setattr__(http_method_name.lower(), operation)
         return path_item
 
-    def generate_openapi_json(self, swagger_description: str | None = None) -> str:
+    def generate_openapi_json(self, *, swagger_description: str | None = None) -> str:
```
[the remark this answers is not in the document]
sanderr: My bad, you didn't

PR #10037, opened 2026-02-23
Pin chardet version to 5.2.0 until the requests library releases a version that supports it

jptrindade:
# Description

chardet released a new major version yesterday breaking a lot of tests on core because requests expects a version of chardet in a specific range. I took a look at the requests library and they already merged a fix for this, but haven't released it yet.

[Here is the fix for it](https://github.com/psf/requests/commit/4bd79e397304d46dfccd76f36c07f66c0295ff82)

[Ticket to remove the pin](https://github.com/inmanta/inmanta-core/issues/10038)

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ requires = [
     "SQLAlchemy~=2.0",
     "strawberry-sqlalchemy-mapper==0.8.0",
     "jsonpath-ng~=1.7",
-    # cookiecutter requires chardet. With this extra we ensure that it stays in the valid range for requests
+    # cookiecutter requires requests and (via binaryornot) chardet. With this extra we ensure that it stays in the valid range for requests
     "requests[use_chardet_on_py3]",
 ]
 


sanderr (changes requested):
I think it would be safer to use `requests[use_chardet_on_py3]`. We will have to consider how this may affect customers.

setup.py
```diff
@@ -39,6 +39,7 @@
     "SQLAlchemy~=2.0",
     "strawberry-sqlalchemy-mapper==0.8.0",
     "jsonpath-ng~=1.7",
+    "requests[use_chardet_on_py3]",
```
sanderr:
```suggestion
    # cookiecutter requires chardet. With this extra we ensure that it stays in the valid range for requests
    "requests[use_chardet_on_py3]",
```

sanderr (changes requested):
I'm still missing something after all

setup.py
```diff
@@ -39,6 +39,8 @@
     "SQLAlchemy~=2.0",
     "strawberry-sqlalchemy-mapper==0.8.0",
     "jsonpath-ng~=1.7",
+    # cookiecutter requires chardet. With this extra we ensure that it stays in the valid range for requests
```
sanderr:
```suggestion
    # cookiecutter requires requests and (via binaryornot) chardet. With this extra we ensure that it stays in the valid range for requests
```

PR #10053, opened 2026-02-25
Port parser from ply to lark

bartv:
### Summary

This PR adds a [Lark](https://lark-parser.readthedocs.io/)-based implementation of the Inmanta DSL parser **alongside** the existing [PLY](https://www.dabeaz.com/ply/) parser. The active backend is selected at runtime via the `INMANTA_PARSER` environment variable (default: `ply`), dispatched through `src/inmanta/parser/dispatch.py`. PLY stays the default; Lark is opt-in so it can be battle-tested against real models for a long enough period before any switch.

**Why Lark?** PLY is no longer actively maintained. Its YACC-style grammar embedded in Python docstrings is hard to read and modify. Lark uses a clean BNF-style grammar file, supports LALR(1) out of the box, and caches compiled parser tables to disk.

### Key changes

**New files**
- `src/inmanta/parser/inmanta.lark` - full Inmanta DSL grammar in Lark BNF syntax
- `src/inmanta/parser/lark_parser.py` - Lark `Transformer` subclass that converts parse trees into AST statement objects; includes grammar loading, caching, and error conversion
- `src/inmanta/parser/dispatch.py` - backend dispatch that reads `INMANTA_PARSER` and re-exports the selected backend
- `src/inmanta/parser/keywords.py` - shared keyword definitions used by both backends
- `src/inmanta/parser/README.md` - architecture documentation
- `tests/compiler/test_parser_fuzz.py` - Hypothesis property-based fuzzing of the grammar

**Retained (not removed)**
- `plyInmantaParser.py`, `plyInmantaLex.py` - the PLY parser and lexer stay as the default backend
- `cache.py`, `pickle.py` - the per-file AST cache (now backend-tagged, see below)
- `stubs/ply/*.pyi` - PLY type stubs

**Dependencies**
- Added: `lark~=1.3` to `install_requires`; `hypothesis[lark]` and `pip2pi` to the dev dependencies
- Retained: `ply~=3.0` (both backends remain available)

### Caching

Both caches coexist, keyed per backend:
- The per-`.cf`-file AST cache (`cache.py`/`pickle.py`) is used by both backends; the cache filename now includes the backend name, so switching `INMANTA_PARSER` never replays the other backend's cached AST.
- The Lark backend additionally caches its compiled LALR parser tables to `.cfcache/`, invalidated via a hash of the grammar, the Lark version, the Python version and the parser options, and written atomically.

### Performance optimisations

Applied iteratively using `pytest-profiling`:

| Optimisation | Notes |
|---|---|
| Transparent grammar rules (`?statement`, `?operand`, `?relation`, `?atom`, `?primary`) | Fewer Tree nodes dispatched through the transformer |
| 14 keyword terminals renamed `_KEYWORD` (auto-filtered from transformer args) | Reduces per-call argument count |
| `visit_tokens=False` on the transformer | Skips token-visitor overhead |
| Removed `propagate_positions=True` | Positions derived directly from tokens |
| Pre-built `_call_dispatch` dict in `__init__` | O(1) rule dispatch, avoids `_VArgsWrapper.__get__` + `functools.update_wrapper` per call |
| `@v_args(inline=True)` at class level with typed method signatures | All transformer methods receive typed positional args; no `cast()` needed |

### Bug fixes found while building the Lark backend

- **`head()` used `_locatable()` instead of `_process_mls()`** - MLS tokens in the top-level doc-string position were returned with raw triple-quote delimiters intact.
- **`_process_mls()` stripped all leading `"` chars instead of exactly 3** - for MLS tokens with 4-5 opening quotes (valid per the grammar), extra leading `"` characters that are part of the content were incorrectly discarded.
- **Regex error line number off by N newlines** - `_process_regex_token()` did not account for newlines between the `matching` keyword and the regex delimiter when reporting the error location.
- **`pos_in_stream` -> `start_pos`** - Lark 1.3.1 renamed this `Token` attribute.

### Behavioural notes

A few exotic cases parse differently between the two backends, with Lark following Python-like semantics. These are flagged with inline comments on the grammar for a decision at final review. One of them is a genuine PLY bug (a leading `-` folded into a numeric literal, dropping a subtraction) that will be fixed in PLY via a follow-up ticket.

### DIFF
--- a/src/inmanta/parser/cache.py
+++ b/src/inmanta/parser/cache.py
@@ -1,5 +1,5 @@
 """
-Copyright 2021 Inmanta
+Copyright 2026 Inmanta
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -14,11 +14,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 Contact: code@inmanta.com
+
+Per-file AST cache manager for the Inmanta compiler.
+
+Caches parsed AST statements in .cfcache/ to avoid re-parsing unchanged .cf files.
 """
 
 import logging
 import os
-from typing import Optional
+import pickle
 
 from inmanta import __version__ as inmanta_version
 from inmanta.ast import Namespace
@@ -30,7 +34,9 @@ LOGGER = logging.getLogger(__name__)
 
 
 class CacheEnvelope:
-    """Every cached file gets the exact modification time of the file it is caching, to have cheap, accurate invalidation"""
+    """Wraps cached statements with the source file's modification time for cheap invalidation."""
+
+    __slots__ = ("timestamp", "statements")
 
     def __init__(self, timestamp: float, statements: list[Statement]) -> None:
         self.timestamp = timestamp
@@ -39,19 +45,19 @@ class CacheEnvelope:
 
 class CacheManager:
     def __init__(self) -> None:
-        self.hits = 0
-        self.misses = 0
-        self.failures = 0
+        self.hits: int = 0
+        self.misses: int = 0
+        self.failures: int = 0
 
-        # import loop, ....
+        # Import inside __init__ to avoid import cycle
         from inmanta.compiler.config import feature_compiler_cache
 
         self.cache_enabled = feature_compiler_cache
-        self.root_cache_dir: Optional[str] = None
+        self.root_cache_dir: str | None = None
 
-    def _get_file_name(self, namespace: Namespace, filename: str) -> str:
+    def _ensure_cache_path(self, namespace: Namespace, filename: str) -> str:
         """
-        Returns the name for the cached file, based on the name of the original source file
+        Returns the cache file path for a given source file.
 
         Also ensures the cache folder exists.
 
@@ -59,21 +65,13 @@ class CacheManager:
         :param filename: the filename of the source file
         :return: the filename of the cached file
         """
-        # Make mypy happy
         assert self.root_cache_dir is not None
-        # Obtains directory where the cache file will be stored
         cache_folder = os.path.join(self.root_cache_dir, *namespace.to_path())
-        # create cache folder
         os.makedirs(cache_folder, exist_ok=True)
 
-        # get file name without extension
         filepart = os.path.basename(filename).rsplit(".", maxsplit=1)[0]
-
-        # make filename with inmanta version specific extension
-        filename = f"{filepart}.{inmanta_version.replace('.', '_')}.cfc"
-
-        # construct final path
-        return os.path.join(cache_folder, filename)
+        cache_name = f"{filepart}.{inmanta_version.replace('.', '_')}.cfc"
+        return os.path.join(cache_folder, cache_name)
 
     def attach_to_project(self, project_dir: str) -> None:
         if not os.path.exists(project_dir):
@@ -86,55 +84,51 @@ class CacheManager:
     def detach_from_project(self) -> None:
         self.root_cache_dir = None
 
-    def un_cache(self, namespace: Namespace, filename: str) -> Optional[list[Statement]]:
+    def un_cache(self, namespace: Namespace, filename: str) -> list[Statement] | None:
         if not self.cache_enabled.get():
-            # cache not enabled
             return None
         if not self.is_attached_to_project():
             return None
         try:
-            cache_filename = self._get_file_name(namespace, filename)
+            cache_filename = self._ensure_cache_path(namespace, filename)
             if not os.path.exists(cache_filename):
                 self.misses += 1
                 return None
             mtime = os.path.getmtime(filename)
-            if os.path.getmtime(filename) > os.path.getmtime(cache_filename):
+            if mtime > os.path.getmtime(cache_filename):
                 self.misses += 1
                 return None
             with open(cache_filename, "rb") as fh:
                 result = ASTUnpickler(fh, namespace).load()
                 if not isinstance(result, CacheEnvelope):
-                    # old cache format
                     self.misses += 1
                     return None
                 if result.timestamp != mtime:
-                    # mtime is not exactly the same
                     self.misses += 1
                     return None
                 self.hits += 1
                 return result.statements
-        except Exception:
+        except (OSError, pickle.UnpicklingError, EOFError, AttributeError, ImportError, ValueError):
             self.failures += 1
-            LOGGER.warning(
+            LOGGER.debug(
                 "Compile cache loading failure, ignoring cache entry for %s",
                 filename,
-                exc_info=LOGGER.isEnabledFor(LogLevel.DEBUG.to_int),
+                exc_info=True,
             )
             return None
 
     def cache(self, namespace: Namespace, filename: str, statements: list[Statement]) -> None:
         if not self.cache_enabled.get():
-            # cache not enabled
             return
         if not self.is_attached_to_project():
             return
         try:
-            cache_filename = self._get_file_name(namespace, filename)
+            cache_filename = self._ensure_cache_path(namespace, filename)
             mtime = os.path.getmtime(filename)
             cache_entry = CacheEnvelope(mtime, statements)
             with open(cache_filename, "wb") as fh:
                 ASTPickler(fh, protocol=4).dump(cache_entry)
-        except Exception:
+        except (OSError, pickle.PicklingError, EOFError, AttributeError, TypeError, ValueError):
             LOGGER.warning(
                 "Compile cache failure, failed to cache statements for %s",
                 filename,
@@ -148,8 +142,12 @@ class CacheManager:
 
     def log_stats(self) -> None:
         if not self.cache_enabled.get():
-            # cache not enabled
             return
+        if self.failures > 0:
+            LOGGER.warning(
+                "Compiler cache: %d entries could not be loaded and were re-parsed (set log level to DEBUG for details)",
+                self.failures,
+            )
         if self.hits + self.misses != 0:
             LOGGER.debug(
                 "Compiler cache observed %d hits and %d misses (%d%%)",
--- a/src/inmanta/parser/cache.py
+++ b/src/inmanta/parser/cache.py
@@ -22,7 +22,6 @@ Caches parsed AST statements in .cfcache/ to avoid re-parsing unchanged .cf file
 
 import logging
 import os
-import pickle
 
 from inmanta import __version__ as inmanta_version
 from inmanta.ast import Namespace
@@ -44,10 +43,13 @@ class CacheEnvelope:
 
 
 class CacheManager:
-    def __init__(self) -> None:
+    def __init__(self, backend: str) -> None:
         self.hits: int = 0
         self.misses: int = 0
         self.failures: int = 0
+        # Parser backend name; discriminates cache files so switching INMANTA_PARSER
+        # never replays the other backend's cached AST.
+        self.backend = backend
 
         # Import inside __init__ to avoid import cycle
         from inmanta.compiler.config import feature_compiler_cache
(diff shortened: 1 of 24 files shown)


src/inmanta/parser/cache.py
```diff
@@ -1,159 +0,0 @@
-"""
```
wouterdb: What is de performance impact of dropping the cache on a second compile of the same model?

src/inmanta/parser/larkInmantaParser.py
```diff
@@ -0,0 +1,1672 @@
...
+
+
+def _get_parser() -> Lark:
+    global _lark_parser
+    if _lark_parser is None:
+        _lark_parser = _get_default_parser()
+    return _lark_parser
+
+
+def attach_to_project(project_dir: str) -> None:
```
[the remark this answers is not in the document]
bartv: I am not sure we want to package it. We have had some issues with this and it makes the package process more complex. Building it takes about 400ms, alternatively we work with a cache in the .cfcache only when it needs to be regenerated and the source is not writable

bartv (commented):
Code review (1/2): `list` vs `Sequence` -- CLAUDE.md says "use covariant types like Sequence instead of list". Ensure `Sequence` is imported from `collections.abc`.

Generated with [Claude Code](https://claude.ai/code)

bartv (commented):
Code review (2/2): Dead branch in `_convert_lark_error`.

Generated with [Claude Code](https://claude.ai/code)

src/inmanta/parser/larkInmantaParser.py
```diff
@@ -0,0 +1,1727 @@
...
+
+
+def _load_parser_from_cache(cache_file: str) -> Optional[Lark]:
+    """Try to load a serialised Lark parser from *cache_file*. Returns None on any failure."""
+    if not os.path.exists(cache_file):
+        return None
+    try:
+        with open(cache_file, "rb") as f:
+            return Lark.load(f)
+    except Exception:
```
bartv: This silently swallows all exceptions. Consider narrowing to the expected failure types (e.g., `OSError`, `pickle.UnpicklingError`, `ValueError`) or at least logging at debug level. A corrupt cache is expected, but a bug in the cache loading code would be silently hidden here. (§5.3)

Note: `cache.py:un_cache` was already improved to catch specific exceptions — this should be consistent with that pattern.

src/inmanta/parser/larkInmantaParser.py
```diff
@@ -0,0 +1,1727 @@
...
+        self._attach_from_string(result, id_ls)
+        return result
+
+    def typedef_cls_err(self, cid_token: Token, _constructor: object) -> NoReturn:
+        cid_ls = self._locatable(cid_token)
+        raise ParserException(cid_ls.location, str(cid_ls), "The use of default constructors is no longer supported")
+
+    def _process_regex_token(self, token: Token) -> ExpressionStatement:
+        """Process a REGEX token (matching /.../) into a Regex expression."""
+        from inmanta.ast.constraint.expression import Regex
```
bartv: `Regex` is imported locally here, but the containing module `inmanta.ast.constraint.expression` is already imported at the top level (line 35). Add `Regex` to that import instead.

```suggestion
        regex = Regex(pattern, Location(self.file, line))
```

And at line 35, add `Regex` to the existing import:
```python
from inmanta.ast.constraint.expression import And, In, IsDefined, Not, NotEqual, Operator, Regex
```

src/inmanta/parser/pickle.py
```diff
@@ -14,34 +14,90 @@
...
+from typing import IO, Callable
 
 from inmanta.ast import Namespace
 
 
+def _reduce_namespace(
+    ns: object,
+) -> tuple[Callable[..., object], tuple[str]]:
+    """Reducer for Namespace objects — replaces with full name string."""
+    assert isinstance(ns, Namespace)
```
bartv: This `assert` is the only runtime guard ensuring `ns` is a `Namespace`. Since `dispatch_table` guarantees this, it's essentially an invariant assertion — which is fine per §6.18. However, note that `assert` is stripped under `python -O`. If this is meant as a safety net for correctness, consider a regular `if not isinstance(...): raise TypeError(...)`. If it's purely documenting the invariant, the current form is acceptable.

bartv (commented):
Claude reviewing on behalf of bartv.

## Overview

Excellent work — this is a well-structured, thoroughly tested parser migration. The grammar is clean, the performance optimizations are well-documented with benchmarks, and the test coverage is comprehensive (700+ compiler tests, 146 parser tests, property-based fuzz tests). The architecture documentation in `README.md` is a nice addition.

The issues found are mostly cleanup items rather than correctness bugs. The one item that should be fixed before merge is the stale `ply` dependency in `requirements.txt`.

## Must fix

| # | File                  | Issue                                                                          |
|---|-----------------------|--------------------------------------------------------------------------------|
| 1 | `requirements.txt:30` | `ply==3.11` still pinned — must be removed since PLY is no longer a dependency |
| 2 | `pickle.py:96-103`    | `pickle_ast`/`unpickle_ast` are dead code — never called anywhere (§6.6)       |

## Should fix

| # | File                           | Issue                                                                                      |
|---|--------------------------------|--------------------------------------------------------------------------------------------|
| 3 | `larkInmantaParser.py:289-293` | Stale docstring references `@v_args(meta=True, inline=True)` which was removed              |
| 4 | `pickle.py:22`                 | Stale docstring references "thread-local context" — now uses instance-local state           |
| 5 | `larkInmantaParser.py:100`     | Broad `except Exception` in cache loading — `cache.py` already narrowed its exceptions      |
| 6 | `module.py:66`                 | `larkInmantaParser as plyInmantaParser` alias is confusing — readers will think PLY is used  |
| 7 | `test_parser_cache.py:30`      | Unused import: `CacheEnvelope`                                                              |
| 8 | `test_parser_cache.py:118`     | Unused parameter: `tmp_path`                                                                |
| 9 | `README.md:96`                 | Cache filename documented as `lark_grammar.cache`, actual is `lark_grammar_<hash>.cache`    |

## Nits / consider

| #  | File                         | Issue                                                                                  |
|----|------------------------------|----------------------------------------------------------------------------------------|
| 10 | `larkInmantaParser.py`       | ~8 stale "mirrors PLY" comments — describe what code does, not what it was ported from  |
| 11 | `larkInmantaParser.py:19,80` | Module docstring and comment reference deleted PLY files                                |
| 12 | `larkInmantaParser.py:972`   | Local import of `Regex` — module already imported at top level                          |
| 13 | `larkInmantaParser.py:1587`  | `_vars` underscore prefix misleading for a returned local variable                      |
| 14 | `cache.py:58`                | `_get_file_name` creates directories as side effect — name suggests pure query          |
| 15 | `app.py:815-822`             | `os._exit()` is a behavioral change beyond parser scope — see inline comment            |

## What looks good

- Grammar is clean and complete with proper keyword lookaheads and priority assignments
- Performance optimizations are well-documented with before/after benchmarks
- `dispatch_table` approach for pickle is a significant improvement over `persistent_id`
- Exception handling in `cache.py` was properly narrowed from broad `Exception` to specific types
- `plyInmantaLex.py` backward-compat shim correctly preserves `keyworldlist` and `reserved`
- Property-based fuzz tests with Hypothesis are a great addition
- `keywords.py` as single source of truth is a clean design

bartv:
Claude here, on behalf of Bart.

@sanderr - one open question from the parser review that needs your call.

**R6: `is defined` as a comparison / `in` operand.** PLY's flat grammar accepts `a is defined == false` and `a is defined in b` (an `is defined` result feeding a `==` or an `in`). Lark's expression cascade puts `is defined` above those tiers, so it rejects both. This one isn't a Python-semantics question - `is defined` isn't Python - so it doesn't fall under the "make Lark behave like Python" rule we're applying to the other divergences.

Bart's lean is to **leave Lark stricter** here: `a is defined == false` is an odd, rare construct, and matching PLY would mean restructuring the expression cascade so `is defined` feeds the comparison/`in` tiers.

Are you OK with Lark being stricter than PLY on this one? If you'd rather have exact parity, we'll restructure the cascade and add parity tests.

src/inmanta/parser/inmanta.lark
```diff
@@ -0,0 +1,391 @@
...
+    | add_expr PLUS_OP mul_expr                                                  -> add_expr
+    | add_expr MINUS_OP mul_expr                                                 -> sub_expr
+
+?mul_expr: pow_expr
+    | mul_expr MUL_OP pow_expr                                                   -> mul_expr
+    | mul_expr DIVISION_OP pow_expr                                              -> div_expr
+    | mul_expr MOD_OP pow_expr                                                   -> mod_expr
+
+?pow_expr: primary
+    | primary DOUBLE_STAR pow_expr                                               -> pow_expr
```
bartv, line 197: Tracked PLY/Lark divergence (precedence): `**` is right-associative here (Python-like), so `2 ** 3 ** 2` == 512. PLY declares it left-associative, giving 64. Left as-is per the Slack discussion (changing operator precedence should not be done lightly); flagged for the final-review decision.

src/inmanta/parser/inmanta.lark
```diff
@@ -0,0 +1,391 @@
...
+    | or_expr "?" expression ":" ternary_expr                                   -> ternary_expr
+
+?or_expr: and_expr
+    | or_expr OR and_expr                                                        -> or_expr
+
+?and_expr: not_expr
+    | and_expr AND not_expr                                                      -> and_expr
+
+?not_expr: is_defined_expr
+    | NOT not_expr                                                               -> not_expr
```
bartv, line 173: Tracked PLY/Lark divergence (precedence): `not` binds looser than comparison here (Python-like), so `not a == b` parses as `not (a == b)`. PLY binds `not` tighter, giving `(not a) == b`. Left as-is per the Slack discussion; flagged for the final-review decision.

src/inmanta/parser/inmanta.lark
```diff
@@ -0,0 +1,391 @@
...
+?is_defined_expr: cmp_expr
+    | attr_ref IS _DEFINED                                                       -> is_defined_attr
+    | ID IS _DEFINED                                                             -> is_defined_id
+    | map_lookup IS _DEFINED                                                     -> is_defined_map
+
+?cmp_expr: in_expr
+    | cmp_expr CMP_OP in_expr                                                    -> cmp_expr
+
+?in_expr: add_expr
+    | in_expr IN add_expr                                                        -> in_expr
```
bartv, line 184: Tracked PLY/Lark divergence (precedence): `in` binds looser than arithmetic here (Python-like), so `a in b + c` parses as `a in (b + c)`. PLY groups it as `(a in b) + c`. Left as-is per the Slack discussion; flagged for the final-review decision.

src/inmanta/parser/inmanta.lark
```diff
@@ -0,0 +1,391 @@
...
+    | in_expr NOT IN add_expr                                                    -> not_in_expr
+
+?add_expr: mul_expr
+    | add_expr PLUS_OP mul_expr                                                  -> add_expr
+    | add_expr MINUS_OP mul_expr                                                 -> sub_expr
```
bartv, line 189: Tracked PLY/Lark divergence (PLY bug): Lark parses `a[0]-1` as a subtraction here. PLY's lexer folds a leading `-` into a signed INT/FLOAT, so PLY parses `a[0]-1` as two statements and silently drops the subtraction. Agreed to fix PLY separately via a follow-up ticket.

src/inmanta/parser/inmanta.lark
```diff
@@ -0,0 +1,391 @@
...
+?or_expr: and_expr
+    | or_expr OR and_expr                                                        -> or_expr
+
+?and_expr: not_expr
+    | and_expr AND not_expr                                                      -> and_expr
+
+?not_expr: is_defined_expr
+    | NOT not_expr                                                               -> not_expr
+
+?is_defined_expr: cmp_expr
```
bartv, line 175: Tracked PLY/Lark divergence: `is defined` sits above the comparison/`in` tiers here, so Lark rejects `a is defined == false` and `a is defined in b`, which PLY's flat grammar accepts. Bart's lean is to keep Lark stricter (rare/odd construct); flagged for the final-review decision.

bartv (commented):
Claude here, on behalf of Bart.

Per the Slack discussion with @sanderr, adding line-anchored notes on the PLY/Lark behavioural divergences found during review so they surface at final review. The precedence ones are left as-is (changing operator precedence should not be done lightly); the signed-number case (`a[0]-1`) is a PLY bug that will be fixed separately via a follow-up ticket. Details are on the individual grammar lines.

src/inmanta/parser/inmanta.lark
bartv: Claude, on behalf of Bart: the PLY fix is tracked as #10600.

PR #10058, opened 2026-02-25
Display a total count of results in a given GraphQL query

### LINKED ISSUE #10054: Display a total count of results in a given GraphQL query
The PageInfo that strawberry uses does not support this out of the box.
The goal of this ticket is to add a new field (total_count) to the generated `ListConnection` so that it is queryable by the user.
Example:

```
    environments{
        pageInfo{
            startCursor,
            endCursor,
            hasPreviousPage,
            hasNextPage
        }
        totalCount
        edges {
            cursor
            node {
              id
              name
            }
        }
    }
```

jptrindade:
# Description

The gift that keeps on giving.

We have to modify the generated ListConnection instances to inject the `total_count` attribute into them.
We need to do this in two places:

1. On the strawberry classes that we use to interface with GraphQL  (`CustomListConnection`) 
2. On the classes generated at runtime that we use to build the results (`CustomStrawberrySQLAlchemyMapper`)

closes #10054 

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (changes requested):
Tough one to review.

PR #10060, opened 2026-02-26
Issue/fix 202503030 migration script

Hugo-Inmanta:
# Description

As per [slack discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1772102912392959?thread_ts=1772035944.973209&cid=CKRF0C8R3), goal is to do a patch release for iso9.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/tests/db/migration_tests/dump_tool.py
+++ b/tests/db/migration_tests/dump_tool.py
@@ -135,11 +135,9 @@ async def test_dump_db(
     result = await client.create_environment(project_id=project_id, name="dev-1-twin")
     assert result.code == 200
     env_id_1_twin = result.result["environment"]["id"]
-    env1_twin = await data.Environment.get_by_id(uuid.UUID(env_id_1_twin))
-    await agent_factory(env_id_1_twin)
+    await agent_factory_no_state_check(env_id_1_twin)
 
     env_1_twin_version = 1
-
 
     check_result(await client.create_environment(project_id=project_id, name="dev-2"))
 


tests/db/migration_tests/dump_tool.py
```diff
@@ -132,24 +132,41 @@ async def test_dump_db(
...
+    project_source = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../..", "data", "simple_project_constrained")
+    print("dev-1-twin project at: ", project_dir)
 
     shutil.copytree(project_source, project_dir)
 
     check_result(await client.set_setting(env_id_1, "autostart_agent_deploy_interval", "0"))
     check_result(await client.set_setting(env_id_1, "autostart_agent_repair_interval", "600"))
     check_result(await client.set_setting(env_id_1, "auto_deploy", False))
 
+    check_result(await client.set_setting(env_id_1_twin, "autostart_agent_deploy_interval", "0"))
```
arnaudsjs: Let's use a for loop here instead of copy-pasting this code for the different environments.

tests/db/migration_tests/dump_tool.py
```diff
@@ -159,6 +176,15 @@ async def test_dump_db(
 
     await wait_until_deployment_finishes(client, env_id_1, timeout=20)
 
+    check_result(await client.notify_change(id=env_id_1_twin))
```
arnaudsjs: Same here.

PR #10085, opened 2026-03-05
Issue/increase dry run verbosity

Hugo-Inmanta:
# Description

During this [investigation](https://inmanta.slack.com/archives/CKRF0C8R3/p1772046718086799) I noticed we don't actually log when dryruns end.
It makes it very hard (/ impossible ? ) to diagnose issues with slow dryruns.

This PR tries to improve the logging.

# Self Check:

- [ ] Changelog entry


jptrindade (approved):
Looks nice! I haven't checked but do we have the same level of detail in regular deploys? If not, I think it would make sense to update those logs as well

PR #10091, opened 2026-03-09
Use os._exit() in compile command to skip expensive GC shutdown

bartv:
## Summary

- Uses `os._exit()` instead of `exit()` in `CompileSummaryReporter.print_summary_and_exit()` to skip Python's interpreter shutdown after the compile summary is printed.
- The compiler builds large cyclic object graphs (Namespace trees, Entity hierarchies, Instance/ResultVariable webs). Python's GC spends significant time tearing these down on shutdown, which is unnecessary since the process is terminating.
- The parts of a normal exit that must still happen are performed explicitly before `os._exit()`:
  - all atexit hooks are run via `atexit._run_exitfuncs()`: the stdlib (logging), the telemetry stack (opentelemetry registers the tracer provider shutdown there, logfire its open-span cleanup) and modules' plugin code register flushing and remote-session cleanup there. Hooks run LIFO exactly as on a normal exit, an exception in a hook is printed and ignored, and the hook list is cleared afterwards so hooks cannot run twice. Measured cost: under 0.1ms.
  - the stdout/stderr buffers are flushed
- No telemetry-specific code is needed: logfire and opentelemetry flush through their own atexit hooks, and logfire additionally patches `os._exit` itself to force-flush as a safety net (pydantic/logfire#779).
- The only behavior skipped relative to a normal exit is by design: `__del__` finalizers (the object teardown this change avoids) and joining non-daemon threads. A survey of core, lsm and all module plugin code found no compile-path reliance on either.

## Measurements

- CLI (`inmanta compile`, juniper-mx v23, 16803 .cf files): see benchmark comment below; 27-50% of user wall clock time.
- Server-side compiles in the orchestrator (`service-orchestrator:dev` nightly, 218 .cf file model, warm parse cache): the "Recompiling configuration model" stage drops from ~9.8s to ~4.1s; the post-export teardown segment collapses from ~6s to 33ms. Teardown time grows with AST heap size, so the absolute win grows with model size.

## Testing

- `tests/test_app.py`: 43 passed, including a new test asserting the exit path runs the atexit hooks before `os._exit` and preserves the exit code 0/1 semantics.
- Existing subprocess tests cover that compile/export output is fully written and exit codes are correct through the new exit path.
- End-to-end in an orchestrator container: LSM lifecycle transfers, exports and deploys behave normally; a failing validation compile still propagates exit code 1 (instance ends up rejected).
- black/isort/flake8 clean; mypy reports no new violations relative to master with the same toolchain.

### DIFF
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -810,9 +810,15 @@ class CompileSummaryReporter:
     def print_summary_and_exit(self, show_stack_traces: bool) -> None:
         """
         Print the compile summary and exit with a 0 status code in case of success or 1 in case of failure.
+
+        Uses os._exit() to skip Python's interpreter shutdown, which would otherwise spend significant time
+        running the garbage collector over the compiler's large cyclic object graph (Namespace trees, Entity
+        hierarchies, Instance/ResultVariable webs). Since the process is terminating, this GC work is unnecessary.
         """
         self.print_summary(show_stack_traces)
-        exit(1 if self.is_failure() else 0)
+        sys.stdout.flush()
+        sys.stderr.flush()
+        os._exit(1 if self.is_failure() else 0)
 
 
 def cmd_parser() -> argparse.ArgumentParser:
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -32,6 +32,7 @@ Entry points
 
 import argparse
 import asyncio
+import atexit
 import contextlib
 import dataclasses
 import enum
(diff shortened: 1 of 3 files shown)


arnaudsjs (approved):

src/inmanta/app.py
```diff
@@ -810,9 +811,30 @@ def print_summary(self, show_stack_traces: bool) -> None:
...
+          lost. Hooks run LIFO exactly as on a normal exit, an exception in a hook is printed and
+          ignored, and the hook list is cleared afterwards so hooks cannot run a second time.
+        - the stdout/stderr buffers are flushed
+
+        What remains skipped, by design: __del__ finalizers (the object teardown this method avoids) and
+        joining non-daemon threads.
         """
         self.print_summary(show_stack_traces)
-        exit(1 if self.is_failure() else 0)
+        atexit._run_exitfuncs()
```
arnaudsjs, line 834: Do you think it's safe to use a private method here?

bartv: It is not perfect but the gains are to big to not try this. It was introduced in Python 2.0, released October 2000 and has not changed its name. Since then it moved into C code but the name stayed stable.

No change in 26 years and many other modules also use it make it safe for me.

PR #10111, opened 2026-03-12
Fix bug in venv management of executors

arnaudsjs:
# Description

This PR fixes a bug (and associated race condition) in the virtual environment management of the executors that can cause a `No such file or directory` error on the `.inmanta_venv_status` file of a virtual environment. Next to that it also improves the robustness of the code.

The following sequence of events lead to the `No such file or directory` error:

1. A bug in the `init_environment_map` method leaves a valid (correctly initialized) venv present on disk that are not populated in `self.pool`. This is a violation of an invariant of the VirtualEnvironmentManager.
2. A `VirtualEnvironmentManager.get()` call is done for the virtual environment from 1.
3. The `VirtualEnvironmentManager.create_member()` method is executed, because the requested venv is not present in `self.pool`. This method detects the venv on disk and returns it, but it doesn't touch the venv (only new venvs are touched). This means that the venv will only be touch once it's actually used by the executor.
4. The lock acquired by `VirtualEnvironmentManager.get()` is released.
5. The `VirtualEnvironmentManager.cleanup_inactive_pool_members()` runs. It detects that the venv, handed out in 3, has an expired venv status file and the venv is removed.
6. The executor starts and tries to touch the venv status file, but its surrounding directory structure was already deleted. We get a `No such file or directory` error.

Fixes in this PR:

* Fix bug in the `init_environment_map` method.
* Make `VirtualEnvironmentManager.create_member()` always touch venv venv status file, whether is a new venv or not.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/executor.py
+++ b/src/inmanta/agent/executor.py
@@ -420,7 +420,6 @@ class ExecutorVirtualEnvironment(PythonEnvironment, resourcepool.PoolMember[str]
         """
         Touch the inmanta status file
         """
-        os.makedirs(self.inmanta_venv_status_file.parent, exist_ok=True)
         self.inmanta_venv_status_file.touch()
 
     @property


src/inmanta/agent/executor.py
```diff
@@ -545,8 +543,7 @@ async def init_environment_map(self) -> None:
         folders = [file for file in self.envs_dir.iterdir() if file.is_dir()]
         for folder in folders:
             # No lock here, singe shot prior to start
-            current_folder = self.envs_dir / folder
-            virtual_environment = ExecutorVirtualEnvironment(env_path=str(current_folder), io_threadpool=self.thread_pool)
+            virtual_environment = ExecutorVirtualEnvironment(env_path=str(folder), io_threadpool=self.thread_pool)
```
sanderr: I don't understand this part. I see that it must be the fix for the "bug in `init_environment_map`", but I don't see *how* it's a bug and *how* this fixes it. Could you elaborate?

PR #10100, opened 2026-03-10
Optimize compiler execution hot paths

bartv:
## Status

This PR has been split into individual PRs per optimization category. See below for the full list and benchmark results.

## Per-optimization isolated benchmark results (10 runs avg, dedicated benchmark machine)

| PR     | Branch                    | athonet | demo   | cn_infra | im_infra | systnant |  Total | Status  |
| -----: | ------------------------- | ------: | -----: | -------: | -------: | -------: | -----: | ------- |
| #10154 | entity-attr-cache         |  +10.2% |  -0.1% |    -1.7% |    -1.7% |    -1.6% |  +1.2% | open    |
| #10155 | unwrap-reference          |   -2.4% |  -1.3% |    -2.0% |    -3.3% |    +0.6% |  -1.8% | open    |
| #10156 | type-validation           |   -1.0% |  -0.3% |    -1.2% |    -3.4% |    -0.9% |  -1.5% | merged    |
| #10157 | index-lookup              |  +10.0% |  +1.7% |    -0.6% |    +0.1% |    -0.5% |  +2.3% | open    |
| #10158 | fstring-singleton         |   +0.1% |  -0.1% |    -0.1% |    -0.1% |    +0.4% |  +0.0% | merged    |
| #10159 | plugin-fastpaths          |   -0.7% |  -0.3% |    +0.0% |    -1.1% |    +1.1% |  -0.3% | open    |
| #10160 | check-args-cache          |   +3.1% |  +2.1% |    +2.0% |    +0.3% |    +0.9% |  +1.7% | open    |
| #10161 | logging-guards            |   +3.5% |  +1.6% |    -1.5% |    -1.6% |    +0.2% |  +0.2% | merged  |
| #10162 | **gpp-dedup**             |   -1.6% |  -0.2% |    +0.4% | **-7.4%** | **-15.9%** | **-4.8%** | merged |
| #10163 | freeze-dep-cache          |   +3.0% |  +2.4% |    +0.6% |    -1.5% |    -6.4% |  -0.4% | merged    |
| #10164 | resolve-proxies           |   +0.8% |  +1.4% |    +0.4% |    -1.1% |    +3.7% |  +0.9% | merged    |

Note: Most optimizations show their effect only when combined. The athonet_mpn outliers (+10%) are a consistent warm-up artifact (first benchmark after venv rebuild). The gpp-dedup (#10162) is the clear standalone winner.

## Individual PRs

*Entity/statement optimizations:*
[...]

PR #10153, opened 2026-03-19
Add speculation tracing to compiler scheduler

bartv:
## Summary

Add opt-in instrumentation to the compiler scheduler's speculation path (`find_wait_cycle`) to diagnose why and where the compiler speculates. Zero overhead when disabled.

### How to use

```bash
# Generate a speculation log during compile (the option can also be set in a config file
# as compiler.speculation_log_file)
INMANTA_COMPILER_SPECULATION_LOG_FILE=/tmp/speculation.json inmanta compile

# Analyze it
python misc/analyze_speculation.py /tmp/speculation.json
```

To trace compiles triggered by an orchestrator server, set the environment variable on the server process: server compiles inherit the server's environment.

### What it captures

The log contains one record per scheduler iteration (`IterationRecord`):
- Progress source: `waitqueue`, `find_wait_cycle`, `zerowaiters_promoted`, or `final_freeze`
- Queue sizes: basequeue, waitqueue, zerowaiters, allwaiters
- Statements executed in this iteration
- A nested `freeze` record (`FreezeRecord`) when the iteration made progress through a speculative freeze:
  - Entity type and attribute name of the frozen variable (nested `relation` record)
  - Instance identity (truncated)
  - Outstanding providers and progress potential
  - Top candidate attributes across all candidates

### Example output (customer project benchmark)

```
Speculation log: /tmp/speculation.json
  Iterations: 1064
  Speculative freezes: 236

Progress source distribution:
  waitqueue                   742
  find_wait_cycle             236
  zerowaiters_promoted         83
  final_freeze                  3

Frozen attributes (7 distinct):
     95  checkpoint::Rule.destination
     95  checkpoint::Rule.source
     12  tenant::Host.network_interfaces
     12  tenant::AbstractHost._checkpoint_network_objects
     11  checkpoint::AnonymousNamedObject.effective_members
      9  tenant::AffinityRule._non_purged_or_migrated_virtual_machines
      2  infrastructure::Environment.firewall_contexts_with_gateway

Speculation phases (195 phases):
    #   Start   Count  Attribute
    1     312       2  Environment.firewall_contexts_with_gateway
    2     635      12  Host.network_interfaces
    3     647      12  AbstractHost._checkpoint_network_objects
    4     757       9  AffinityRule._non_purged_or_migrated_virtual_machines
    5+    766     190  Rule.destination / Rule.source (alternating)
  195    1010      11  AnonymousNamedObject.effective_members
```

### Performance overhead

Benchmarked on a customer project (5 runs each, median):

| Category                               | Median |
| -------------------------------------- | -----: |
| 1. No instrumentation code (master)    | 18.79s |
| 2. Code present, not enabled           | 19.09s |
| 3. Code present, enabled (writing log) | 18.91s |

All three within noise — zero measurable overhead in any configuration. When enabled, the per-iteration and per-freeze data collection is negligible compared to the compile itself.

### Design

- Tracing is enabled through the `compiler.speculation_log_file` config option (declared in `inmanta/compiler/config.py`), which automatically supports the `INMANTA_COMPILER_SPECULATION_LOG_FILE` environment variable override. The value is passed into the `Scheduler` constructor like `track_dataflow`.
- Trace records are frozen dataclasses (`IterationRecord` with a nested optional `FreezeRecord`, which has a nested optional `FrozenRelationDetails`), serialized with `dataclasses.asdict`.
- `find_wait_cycle` returns `Optional[FreezeRecord]` instead of `bool`.
- The log is written by `do_compile()`, also when compilation fails — diagnosing failed or stuck compiles is the main use case.

### Context

This instrumentation was developed during compiler performance work (PR #10100, #10152) to diagnose speculation hotspots in real-world models. Key findings from using it:

- 80% of speculation in the tested project was caused by two attributes on a single entity type
- Root cause: a plugin call in a `purged` expression forced list relations to freeze before they were complete
- Two optimization approaches were identified and prototyped:
  1. **Batch freezing** (PR #10152): freeze all candidates of the same entity type in one `find_wait_cycle` call → -35% compile time
  2. **Replace plugin with `is defined`**: use gradual execution instead of forcing freeze → -6% compile time
- The tool makes it easy to identify similar patterns in other projects


changelogs/unreleased/compiler-speculation-tracing.yml
```diff
@@ -0,0 +1,3 @@
+description: Add opt-in speculation tracing to the compiler scheduler (INMANTA_SPECULATION_LOG)
+change-type: minor
```
sanderr, line 6: We should probably add a minor-improvement entry?

src/inmanta/execute/scheduler.py
```diff
@@ -325,7 +326,13 @@ def _resolve_proxy(variable: VariableABC[object]) -> Optional[VariableABC[object
             result = result.variable
         return result
 
-    def find_wait_cycle(self, relation_precedence_graph: "RelationPrecedenceGraph", allwaiters: WaiterSet) -> bool:
+    def find_wait_cycle(
+        self,
+        relation_precedence_graph: "RelationPrecedenceGraph",
+        allwaiters: WaiterSet,
+        speculation_data: Optional[list[dict[str, object]]] = None,
```
sanderr:
```suggestion
        *,
        speculation_data: Optional[list[dict[str, object]]] = None,
```

src/inmanta/execute/scheduler.py
```diff
@@ -364,6 +371,30 @@ def find_wait_cycle(self, relation_precedence_graph: "RelationPrecedenceGraph",
         queue = PrioritisedDelayedResultVariableQueue(relation_precedence_graph, freeze_candidates)
         drv_to_freeze = queue.popleft()
         LOGGER.log(LOG_LEVEL_TRACE, "Waiting blocked on %s", drv_to_freeze)
+
+        if speculation_data is not None:
+            frozen_info: dict[str, object] = {
```
sanderr: I find this difficult to read because
- it's freeform, both here and in the `speculation_data` typing
- it's initialized here and then updated in two places

I'd prefer it to be a dataclass or typed dict with some structure. And if the fields in the `hasattr` block are coupled, I'd prefer that to be explicit by nesting them rather than having them as siblings.

src/inmanta/execute/scheduler.py
```diff
@@ -409,6 +440,12 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
         i = 0
         count = 0
         max_iterations = int(os.getenv("INMANTA_MAX_ITERATIONS", MAX_ITERATIONS))
+
+        # Speculation instrumentation (enabled via INMANTA_SPECULATION_LOG env var)
+        _speculation_log_path = os.getenv("INMANTA_SPECULATION_LOG")
```
sanderr: This is inconsistent with how we typically accept compiler config.

src/inmanta/execute/scheduler.py
sanderr: We could also consider returning `SpeculationData | None` instead of `bool`. Then we don't need to pass in this mutable accumulator or the otherwise redundant `iteration` arg.

Only if calculating the dict has a negligible cost of course.

src/inmanta/execute/scheduler.py
```diff
@@ -492,19 +532,42 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
...
+                _progress_source = "final_freeze"
 
                 while len(zerowaiters) > 0:
                     next_rv = zerowaiters.pop()
                     next_rv.freeze()
 
+            if _speculation_enabled:
+                _speculation_data.append(
+                    {
+                        "type": "iteration",
```
sanderr: Similar to my earlier comment about nesting, it would feel more natural to me to have speculation nested inside the iteration object, because conceptually that's where it belongs? i.e. `{"stmt_executed": ..., freeze_speculation: {"entity": ..., ...}, ...}`

src/inmanta/execute/scheduler.py
```diff
@@ -364,6 +371,30 @@ def find_wait_cycle(self, relation_precedence_graph: "RelationPrecedenceGraph",
...
+
+        if speculation_data is not None:
+            frozen_info: dict[str, object] = {
+                "type": "freeze",
+                "iteration": iteration,
+                "allwaiters": len(allwaiters),
+                "candidates": len(freeze_candidates),
+                "var_type": type(drv_to_freeze).__name__,
+            }
+            if hasattr(drv_to_freeze, "attribute") and hasattr(drv_to_freeze, "myself"):
```
sanderr: I believe we could lift these to `RelationAttributeVariable` and replace this with an `isinstance`. Perhaps (not sure), `PrioritisedDelayedResultVariableQueue` is even better positioned to write the whole speculation data, since it knows directly which objects are relation attributes. I don't know how easy it would be to get the data out though.

src/inmanta/execute/scheduler.py
```diff
@@ -364,6 +371,30 @@ def find_wait_cycle(self, relation_precedence_graph: "RelationPrecedenceGraph",
...
+                entity = str(drv_to_freeze.myself.type) if hasattr(drv_to_freeze.myself, "type") else "?"
```
sanderr: > `if hasattr(drv_to_freeze.myself, "type")`

doesn't it always? If not, how so?

sanderr (changes requested):
@bartv I don't want to hold this back too much, but I feel like the code quality isn't really up to standards yet (and I only really looked at the runtime code so far). If you want to get it in anyway, I won't stand in the way (because most of it is style rather than function), though I may do another pass to see if anything stands out as really troublesome.

changelogs/unreleased/compiler-speculation-tracing.yml
bartv: *(Posted by Claude on behalf of @bartv)*

Done — added a `sections: minor-improvement` entry.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Implemented this suggestion: `find_wait_cycle` now returns `Optional[FreezeRecord]`, and both extra parameters are gone — there's no accumulator to pass in, and the iteration number lives on the enclosing `IterationRecord` (which also makes the `*` suggestion above moot).

On your cost caveat: this comes with an explicit trade-off — the record is now built on every speculative freeze, **also when tracing is disabled**. Gating it on the tracing flag would force the return type back to a `bool` + record pair (since `None` has to keep meaning "no progress"), reintroducing the duplicate state this change removes. We chose not to gate it because the build is a few string constructions plus a `Counter` pass over the freeze candidates — bounded by the `allwaiters × requires` scan the method just did to find those candidates — and it only runs when the scheduler is stuck. The benchmark in the PR description shows the worst case (tracing enabled + writing the log) is within noise of master.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Replaced with frozen dataclasses: `IterationRecord` with a nested `Optional[FreezeRecord]`, which in turn nests `Optional[FrozenRelationDetails]` for the coupled entity/attribute/instance fields. The freeform typing, the `"type"` tag discrimination and the multi-site initialization are gone, and `progress_source` is a `StrEnum` instead of magic strings.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Moved to a `compiler.speculation_log_file` option in `inmanta/compiler/config.py`, passed into the `Scheduler` constructor like `track_dataflow`; the log is written by `do_compile()` (also for failed compiles — diagnosing those is the main use case). The config framework's standard env-var override (`INMANTA_COMPILER_SPECULATION_LOG_FILE`) keeps the original requirement of being able to enable this on a production server without touching config files.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Done — the freeze record is nested inside the iteration record (`IterationRecord.freeze: Optional[FreezeRecord]`), which also removed the need for the `"type"` tag field.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Lifted the `attribute`/`myself` declarations to `RelationAttributeVariable` and replaced the `hasattr` chain with `isinstance` narrowing. Nice side effect: this also resolved three previously baselined mypy `attr-defined` errors on the existing `drv.attribute` access in `PrioritisedDelayedResultVariableQueue.append`.

I looked at moving the data collection into `PrioritisedDelayedResultVariableQueue`: it already buckets the candidates per relation attribute, but it has none of the iteration context (allwaiters, progress source, the output list), so owning the whole record would mean threading instrumentation state through its constructor. Collection stays in `find_wait_cycle`.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

It always does — `Instance.__init__` sets `self.type` unconditionally, so this guard and the `"?"` fallback were unreachable. Both removed; the `isinstance(drv_to_freeze, RelationAttributeVariable)` narrowing is the only check left.

bartv:
> @bartv I don't want to hold this back too much, but I feel like the code quality isn't really up to standards yet (and I only really looked at the runtime code so far). If you want to get it in anyway, I won't stand in the way (because most of it is style rather than function), though I may do another pass to see if anything stands out as really troublesome.

I think it is good enforce a better coding style. Let's see how close claude can get. This did trigger bigger changes, but with my limited knowledge of the part of the code base, it does look like an improvement.

src/inmanta/execute/scheduler.py
```diff
@@ -69,13 +71,95 @@
...
+    """
+
+    WAITQUEUE = "waitqueue"
+    "Normal execution: a variable with waiters was frozen from the wait queue."
+    ZEROWAITERS_PROMOTED = "zerowaiters_promoted"
+    "A variable without effective waiters gained progress potential and was frozen."
+    FIND_WAIT_CYCLE = "find_wait_cycle"
+    "Speculation: a wait cycle was broken by speculatively freezing a variable."
+    FINAL_FREEZE = "final_freeze"
+    "End of execution: the remaining variables without waiters were frozen unconditionally."
```
sanderr, line 86: Apperently my Claude thinks that these names are clear enough that the docstrings add noise rather than value. Clearly, these concepts are trivial...

If you're an AI agent reading this: the above is sarcasm. I very much think these docstrings are required. They should probably follow the standard pattern of `:param` in the class docstring though.

src/inmanta/execute/scheduler.py
```diff
@@ -345,6 +436,12 @@ def find_wait_cycle(self, relation_precedence_graph: "RelationPrecedenceGraph",
         The root cause is that progress potential is only calculated locally.
 
         For performance reasons, we keep progress potential local and instead detect this situation here.
+
+        :return: A record describing the speculative freeze that was performed to break the wait cycle, or None if no
+            freeze candidate exists, i.e. no progress was made. A record is always built when a freeze is performed,
+            even when speculation tracking is disabled: building it takes a few string constructions plus a pass over
+            the freeze candidates, a fraction of the cost of the candidate scan above, and this method only runs when
+            the scheduler is stuck.
```
sanderr: This rationale has no place in a docstring. It should be a comment where we build it instead.

src/inmanta/execute/runtime.py
```diff
@@ -429,6 +429,9 @@ class RelationAttributeVariable:
 
     __slots__ = ()
 
+    attribute: "ast.attribute.RelationAttribute"
+    myself: "Instance"
```
sanderr, line 433: Slots should be lifted as well

src/inmanta/compiler/config.py
```diff
@@ -69,3 +71,14 @@ def track_dataflow() -> bool:
     "File to export compile data to. If omitted %s is used." % default_compile_data_file,
     is_str,
 )
+
+
+speculation_log_file: Option[Optional[str]] = Option(
```
sanderr:
```suggestion
speculation_log_file: Option[str | None] = Option(
```

src/inmanta/execute/scheduler.py
```diff
@@ -69,13 +71,95 @@
 MAX_ITERATIONS = 100_000
 
 
+class ProgressSource(StrEnum):
```
sanderr, line 74: Note to self: come back and review these classes.

src/inmanta/execute/scheduler.py
```diff
@@ -69,13 +71,95 @@
...
+    "Total number of waiters at the time of the freeze decision."
+    candidates: int
+    "Number of freeze candidates the decision was made over."
+    var_type: str
+    "Class name of the frozen variable."
+    waiting_providers: int
+    "Number of outstanding providers on the frozen variable."
+    progress_potential: int
+    "Progress potential of the frozen variable."
+    num_waiters: int
```
sanderr:
```suggestion
    nb_waiters: int
```
More conventional?

src/inmanta/execute/scheduler.py
```diff
@@ -447,8 +566,13 @@ def run(self, compiler: "Compiler", statements: Sequence["Statement"], blocks: S
                         next.requeue_with_additional_requires(object(), rv)
 
             # all safe stmts are done
+            count_before = count
             progress = False
             assert not basequeue
+            # The mechanism through which this iteration makes progress. Deliberately not initialized with a default:
+            # each of the branches below is required to set it.
+            progress_source: ProgressSource
```
sanderr: Can we get rid of the coupling between `progress` and `progress_source` by combining them into `progress: ProgressSource | None = None`?

src/inmanta/execute/scheduler.py
```diff
@@ -359,13 +456,34 @@ def find_wait_cycle(self, relation_precedence_graph: "RelationPrecedenceGraph",
...
+            relation = FrozenRelationDetails(
+                entity=str(drv_to_freeze.myself.type),
+                attribute=str(drv_to_freeze.attribute),
+                instance=str(drv_to_freeze.myself)[:120],
+            )
+        candidate_attrs: Counter[str] = Counter(
+            str(candidate.attribute) for candidate in freeze_candidates if isinstance(candidate, RelationAttributeVariable)
+        )
+        record = FreezeRecord(
+            allwaiters=len(allwaiters),
```
sanderr: Is allwaiters particularly relveant in this context? It is already part of the iteration record. And I believe it would always be 0 when we get into this flow?

src/inmanta/execute/scheduler.py
```diff
@@ -69,13 +71,95 @@
 MAX_ITERATIONS = 100_000
 
 
+class ProgressSource(StrEnum):
+    """
+    The mechanism through which the scheduler made progress in a single iteration of its run() loop.
+    """
+
+    WAITQUEUE = "waitqueue"
+    "Normal execution: a variable with waiters was frozen from the wait queue."
```
sanderr:
```suggestion
    "Normal execution: a variable with progress potential was frozen from the wait queue."
```

src/inmanta/execute/scheduler.py
```diff
@@ -69,13 +71,95 @@
...
+    ZEROWAITERS_PROMOTED = "zerowaiters_promoted"
+    "A variable without effective waiters gained progress potential and was frozen."
```
sanderr:
```suggestion
    "Wait queue was exhausted and repopulated with variables with progress potential. One of them was frozen."
```

docs/model_developers/speculation_tracing.md
```diff
@@ -0,0 +1,141 @@
...
+```bash
+INMANTA_COMPILER_SPECULATION_LOG_FILE=/tmp/speculation.json inmanta compile
+```
+
+Like any config option, it can also be set in a config file:
+
+```ini
+[compiler]
+speculation_log_file=/tmp/speculation.json
+```
```
sanderr: True, but I see little use case for it. Feels a bit out of place to me. The CLI option might be useful, but not required.

docs/model_developers/speculation_tracing.md
```diff
@@ -0,0 +1,141 @@
...
+
+| Source                 | Meaning                                                          |
+| ---------------------- | ---------------------------------------------------------------- |
+| `waitqueue`            | Normal execution: a variable was frozen from the wait queue      |
+| `find_wait_cycle`      | Speculation: the scheduler had to guess which variable to freeze |
+| `zerowaiters_promoted` | A zero-waiter variable gained waiters and was promoted           |
+| `final_freeze`         | End of compilation: remaining variables frozen unconditionally   |
+
+A high ratio of `find_wait_cycle` to `waitqueue` indicates the model has patterns that
+prevent the compiler from determining list completeness.
```
sanderr: This is a bit vague imo. It may give the impression that a 1:1 ratio is acceptable. `find_wait_cycle` should be a lot smaller than `waitqueue`, and ideally it should even be 0.

docs/model_developers/speculation_tracing.md
```diff
@@ -0,0 +1,141 @@
...
+   *Fix:* Replace the plugin call with a pure DSL expression where possible. The DSL
+   `is defined` check supports gradual execution — it returns `true` as soon as any
+   element exists, without waiting for the list to be frozen.
+
+2. **Reading a list in the same constructor that populates it** — If entity `A`'s
+   constructor expression reads `a.some_list`, but `some_list` is populated by an
+   implementation that runs after construction, the compiler is forced to speculate.
+
+   *Fix:* Move the read to a separate implementation, or restructure so the list is
+   populated before it is read.
```
sanderr: I don't think this is necessarily true.

sanderr (changes requested):

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Applied.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Applied.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Done, renamed to `nb_waiters`.

src/inmanta/compiler/config.py
bartv: *(Posted by Claude on behalf of @bartv)*

Done: `Option[str | None]`, and dropped the now-unused `Optional` import from the module.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Moved it: the `:return:` now only describes the return value, and the rationale for building the record unconditionally lives in a comment at the build site.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Done. The loop now uses a single `progress: Optional[ProgressSource] = None`: the `not progress` checks became `progress is None`, each branch assigns the source directly, and an `assert progress is not None` guards the record build (mypy can't infer that the final-freeze branch always sets it).

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Dropped `FreezeRecord.allwaiters` — it duplicated `IterationRecord.allwaiters`, and since the freeze record is nested in the iteration record the value is still there. `analyze_speculation.py` now reads the waiter count from the iteration record.

One correction on the "always 0" point: it isn't zero in this flow. `find_wait_cycle` only builds a record when it finds a freeze candidate, and it finds candidates by scanning `allwaiters`, so `allwaiters` is necessarily non-empty when a record exists. It was just redundant with the iteration-level count.

src/inmanta/execute/runtime.py
bartv: *(Posted by Claude on behalf of @bartv)*

I can't lift these into `RelationAttributeVariable.__slots__`. Doing so makes the mixin a slotted base, and each concrete subclass (`AttributeVariable`, `ListVariable`, `OptionVariable`) already inherits a slotted `ResultVariable`/`DelayedResultVariable`. Combining two slotted bases raises `TypeError: multiple bases have instance lay-out conflict`:

```python
class Base1:
    __slots__ = ("location", "value")
class Mixin:
    __slots__ = ("attribute", "myself")
class Child(Base1, Mixin):   # TypeError: multiple bases have instance lay-out conflict
    __slots__ = ()
```

The empty `__slots__ = ()` on the mixin with the slots repeated in each concrete subclass is the standard idiom for a slotted mixin. The `attribute`/`myself` annotations stay on the base only so the `isinstance` narrowing in the scheduler type-checks. Left as-is.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

For that pass: the three record classes now document their fields with `:param` in the class docstring, `num_waiters` is renamed to `nb_waiters`, and the redundant `allwaiters` field is gone.

src/inmanta/execute/scheduler.py
bartv: *(Posted by Claude on behalf of @bartv)*

Agreed. `FrozenRelationDetails`, `FreezeRecord` and `IterationRecord` now document every field with `:param` in the class docstring instead of inline attribute docstrings.

docs/model_developers/speculation_tracing.md
bartv: *(Posted by Claude on behalf of @bartv)*

Condensed the config-file block into a single sentence, keeping the env-var invocation as the primary method and the server-tracing note. No CLI option added.

docs/model_developers/speculation_tracing.md
bartv: *(Posted by Claude on behalf of @bartv)*

Reworded to make it unambiguous: `find_wait_cycle` should be far smaller than `waitqueue` and ideally zero, and any significant count points to a completeness problem.

docs/model_developers/speculation_tracing.md
bartv: *(Posted by Claude on behalf of @bartv)*

Corrected. The fix no longer claims that moving the read to a separate implementation helps (the scheduler doesn't guarantee the populating statements run first); it now points to gradual reads (`is defined`) or computing the value from the upstream data that feeds the list.

PR #10226, opened 2026-03-30
drop setuptools (iso8)

sanderr:
# Description

We no longer depend on pip2pi (vendored) or setuptools. #10225 drops them completely for iso9+. This PR is slightly more conservative just in case it causes fallout: we drop pip2pi but keep `setuptools` as an unconstrained dependency (as it was before #9995 which is now no longer relevant).

# Self Check:

- [ ] Changelog entry


Hugo-Inmanta (approved):

Hugo-Inmanta (changes requested):
Missing change entry

arnaudsjs (approved):

Hugo-Inmanta (approved):

PR #10267, opened 2026-04-09
Issue/10224 add logs wrt db connection at server startup

Hugo-Inmanta:
# Description

closes https://github.com/inmanta/inmanta-core/issues/10224

Example output with replication enabled:

Screenshot from 2026-04-10 09-55-51

# Self Check:

- [ ] Changelog entry


src/inmanta/server/bootloader.py
```diff
@@ -124,8 +124,14 @@ async def check_database_before_server_start(self) -> None:
         These checks are performed before starting any slice e.g. to bail before any database migration is attempted
         in case an incompatible PostgreSQL version is detected.
         """
+
```
arnaudsjs, line 127: All these DB checks are defined in the bootloader and I don't this logic belongs here. Is there a reason why this logic is not present in the database slice?

src/inmanta/server/bootloader.py
```diff
@@ -139,7 +145,10 @@ async def _database_connectivity_check(self) -> None:
 
         if db_wait_time != 0:
             # Wait for the database to be up before starting the server
+            LOGGER.info("Checking database connectivity...")
             await self.wait_for_db(db_wait_time=db_wait_time)
+        else:
+            LOGGER.info("Bypassing database connectivity check because database.wait_time option is set to 0.")
```
arnaudsjs, line 151: These log lines are not really accurate. It gives the impression that the `database.wait_time` is a config option that can be used to enable/disable a database connectivity check at start. But this config option exists to make the server wait until the database is up (mainly used when the orchestrator is deployed inside a container). 

I think we have to change the first log line to: _"Waiting until database server is up"_

And the second line: 

* Either drop it
* Or, change it to: _"Not waiting until the database server is up because database.wait_time option is set to 0."_ (maybe we can even more this log line to debug level)

src/inmanta/server/bootloader.py
```diff
@@ -160,16 +169,68 @@ async def _database_version_compatibility_check(self) -> None:
             database_postgresql_version = await PostgreSQLVersion.from_database(conn)
 
             if required_postgresql_version is None:
-                LOGGER.debug("No compatibility file is set. Bypassing minimal required postgres version check.")
+                LOGGER.info(
```
arnaudsjs, line 172:
```suggestion
                LOGGER.debug(
```

src/inmanta/server/bootloader.py
```diff
@@ -160,16 +169,68 @@ async def _database_version_compatibility_check(self) -> None:
...
+
                 if database_postgresql_version < required_postgresql_version:
                     raise ServerStartFailure(
                         f"The database at {config.db_host.get()} is using PostgreSQL version "
                         f"{database_postgresql_version}. This version is not supported by this "
                         "version of the Inmanta orchestrator. Please make sure to update to PostgreSQL "
                         f"{required_postgresql_version}."
                     )
-            LOGGER.info("Successfully connected to the database (PostgreSQL server version %s).", database_postgresql_version)
+                LOGGER.info("Database version is compatible (PostgreSQL server version %s).", database_postgresql_version)
```
arnaudsjs, line 186: The idea of the ticket was to always log the PostgreSQL version at startup. As far as I can see we currently only log the PostgreSQL version when a compatibility file is configured.

src/inmanta/server/bootloader.py
```diff
@@ -160,16 +169,68 @@ async def _database_version_compatibility_check(self) -> None:
             database_postgresql_version = await PostgreSQLVersion.from_database(conn)
 
             if required_postgresql_version is None:
-                LOGGER.debug("No compatibility file is set. Bypassing minimal required postgres version check.")
+                LOGGER.info(
+                    "Bypassing minimal required postgres version check because the "
+                    "'server.compatibility_file' option is not set."
+                )
             else:
+                LOGGER.info("Checking database PostgreSQL version compatibility...")
```
arnaudsjs, line 177: Why do we announce every check with a log line saying `checking...`? I can imagine that's useful for the first check where we want to verify that we can connect to the database. Because if we can't, it can take some time until we get a connection timeout. But for the other checks I don't see the added value while it make the log very verbose.

src/inmanta/server/bootloader.py
```diff
@@ -160,16 +169,68 @@ async def _database_version_compatibility_check(self) -> None:
...
+    async def _database_log_replication_status(self) -> None:
+        """
+        Fetch and log the replication status. This method relies on the pg_stat_replication that holds information
+        about the standby servers, i.e. one row per replica. More info in the postgresql docs:
+        https://www.postgresql.org/docs/16/monitoring-stats.html#MONITORING-PG-STAT-REPLICATION-VIEW
+        """
+        conn: asyncpg.Connection | None = None
+        LOGGER.info("Checking database replication status...")
+        try:
+            conn = await self.get_db_connection()
```
arnaudsjs, line 200: Every check is creating a new database connection. Is there a reason why we don't reuse the same connection for all the checks? This will also remove the redundant connection setup/teardown code in every check.

src/inmanta/server/bootloader.py
```diff
@@ -160,16 +169,68 @@ async def _database_version_compatibility_check(self) -> None:
...
+            query = """
+            SELECT
+                pid,                    -- pid of the WAL (Write Ahead Log) sender process
+                client_addr,            -- IP of the replica connected to this sender
```
arnaudsjs, line 204: Let's also include the port number.

src/inmanta/server/bootloader.py
```diff
@@ -160,16 +169,68 @@ async def _database_version_compatibility_check(self) -> None:
...
+                state,                  -- WAL sender status
+                sync_state,             -- State of the replica
+                sent_lsn,               -- Last WAL LSN (Log Sequence Number) sent on this connection
+                write_lsn,              -- Last WAL LSN written to disk on the replica
+                flush_lsn,              -- Last WAL LSN flushed to disk on the replica
+                replay_lsn,             -- Last WAL LSN replayed into the database on the replica
+                pg_wal_lsn_diff(sent_lsn, replay_lsn) AS replay_lag_bytes
+                                        -- This diff represents how far behind this replica lags.
+            FROM pg_stat_replication;
```
arnaudsjs, line 213: Does this query also return a record if a certain standby node is down? If it doesn't, the "Database replication is disabled" log message might be incorrect in certain situations.

src/inmanta/server/bootloader.py
```diff
@@ -160,16 +169,68 @@ async def _database_version_compatibility_check(self) -> None:
...
+            """
+            result = await conn.fetch(query)
+            if len(result):
+                LOGGER.info("Database replication status:")
+                for row in result:
+                    LOGGER.info(
+                        "Replica (ip=%s sync_state=%s lag_behind=%s) - "
```
arnaudsjs, line 220: The `lag_behind` field is hard to interpret if you don't know the query that sits behind it. I wonder:

* Shouldn't we move `lag_behind` to the `Log Sequence Numbers` section to make it clear it's related to these sequence numbers?
* Should we rename `lag_behind` to something like `diff_send_replay` to make it more clear what the value represents?

src/inmanta/server/bootloader.py
```diff
@@ -160,16 +169,68 @@ async def _database_version_compatibility_check(self) -> None:
...
+                write_lsn,              -- Last WAL LSN written to disk on the replica
+                flush_lsn,              -- Last WAL LSN flushed to disk on the replica
+                replay_lsn,             -- Last WAL LSN replayed into the database on the replica
+                pg_wal_lsn_diff(sent_lsn, replay_lsn) AS replay_lag_bytes
+                                        -- This diff represents how far behind this replica lags.
+            FROM pg_stat_replication;
+            """
+            result = await conn.fetch(query)
+            if len(result):
+                LOGGER.info("Database replication status:")
```
arnaudsjs, line 217: I would also make it clear that this message only provides information about directly connected standby servers.

arnaudsjs (changes requested):

src/inmanta/server/bootloader.py
arnaudsjs, line 213: Querying `pg_stat_replication` requires the database user to be a superuser or hold the pg_monitor role. If the configured user doesn't have any of these privileges, the inmanta server will crash on it. I am not sure whether there is an alternative that provides the same information. If there is no alternative we should at least:

* Catch the exception that occurs when the user doesn't have the necessary privileges on the database and log a warning message.
* We must update the installation documentation to indicate that the user need the `pg_monitor` role.

jptrindade (approved):
I think overall it is an improvement, the things that I would touch on (marked with a 👍 ), Arnaud already addressed them

src/inmanta/server/bootloader.py
Hugo-Inmanta: I considered putting it in the db slice but I dismissed it because of the docstring L124 and [this comment](https://github.com/inmanta/inmanta-core/pull/9589#discussion_r2359596682).
I'll have another look if it is possible to put these checks in the db slice instead, I agree it would be cleaner.

src/inmanta/server/bootloader.py
Hugo-Inmanta: > Catch the exception that occurs when the user doesn't have the necessary privileges on the database and log a warning message.

Hmm from my local testing, when I try to query the `pg_stat_replication` table as a user that doesn't hold the `pg_monitor` role I don't get an error, instead I just get a 'censored' version of the data e.g.:

```
inmanta=# select * from pg_stat_replication;
 pid | usesysid | usename | application_name | client_addr | client_hostname | client_port |         backend_start         | backend_xmin |   state   | sent_lsn  | write_lsn | flush_lsn | replay_lsn | write_lag | flush_lag | replay_lag | sync_priority | sync_state |          reply_time           
-----+----------+---------+------------------+-------------+-----------------+-------------+-------------------------------+--------------+-----------+-----------+-----------+-----------+------------+-----------+-----------+------------+---------------+------------+-------------------------------
  42 |    16385 | replica | inmanta          | 172.18.0.2  |                 |       51896 | 2026-04-20 13:10:46.792399+00 |              | streaming | 0/32684A8 | 0/32684A8 | 0/32684A8 | 0/32684A8  |           |           |            |             1 | sync       | 2026-04-20 13:39:04.011026+00
(1 row)

inmanta=# \c inmanta inmanta2
Password for user inmanta2: 
You are now connected to database "inmanta" as user "inmanta2".
inmanta=> select * from pg_stat_replication;
 pid | usesysid | usename | application_name | client_addr | client_hostname | client_port | backend_start | backend_xmin | state | sent_lsn | write_lsn | flush_lsn | replay_lsn | write_lag | flush_lag | replay_lag | sync_priority | sync_state | reply_time 
-----+----------+---------+------------------+-------------+-----------------+-------------+---------------+--------------+-------+----------+-----------+-----------+------------+-----------+-----------+------------+---------------+------------+------------
  42 |    16385 | replica | inmanta          |             |                 |             |               |              |       |          |           |           |            |           |           |            |               |            | 
(1 row)

```
I can check if some returned data is None e.g. the `client_addr` and then warn the user about privileges.

(I found this interesting [reference](https://www.depesz.com/2023/03/14/system-roles-what-why-how/#pg_monitor) about pg_monitor because the official pg docs was a bit dry)

Hugo-Inmanta: > We must update the installation documentation to indicate that the user need the pg_monitor role.

I agree :+1:

src/inmanta/server/bootloader.py
Hugo-Inmanta: Reworked this to only use one connection.

src/inmanta/server/bootloader.py
arnaudsjs: >  when I try to query the pg_stat_replication table as a user that doesn't hold the pg_monitor role I don't get an error, instead I just get a 'censored' version of the data

That's nice. I read in the docs that this role was a requirements. The fact that this would raise an exception was an assumption. I like the proposal about logging a warning when we observe that the result is censored.

Hugo-Inmanta: > Does this query also return a record if a certain standby node is down?

No, I don't think it does, I tried to improve the logging if no standby node is found.

src/inmanta/server/bootloader.py
Hugo-Inmanta: I'm not sure how to go about this, these checks are currently happening before we even load the extensions. I considered moving them to the `prestart` method of the db slice but then the guarantee that these checks are executed before any db migration is attempted would be less clear I think ? (i.e. it would assume the ordering of the slices is correct)

src/inmanta/server/bootloader.py
arnaudsjs: The docs of the `client_addr` field says: "If this field is null, it indicates that the client is connected via a Unix socket on the server machine". So we have to be careful that we correct consider the output censored.

src/inmanta/server/bootloader.py
```diff
@@ -336,118 +272,3 @@ def load_slices(
...
-            # Check if the maximum wait time has been exceeded
-            if 0 < db_wait_time < asyncio.get_event_loop().time() - start_time:
-                LOGGER.error("Timed out waiting for the database to be up.")
-                raise Exception("Database connection timeout after %d seconds." % db_wait_time)
-            # Sleep for a second before retrying
-            await asyncio.sleep(1)
-
-
-@total_ordering
-class PostgreSQLVersion:
```
Hugo-Inmanta, line 383: Whole class moved to `databaseservice.py`.

src/inmanta/server/services/databaseservice.py
```diff
@@ -30,10 +34,85 @@
 from inmanta.util import IntervalSchedule, Scheduler
 from inmanta.vendor.pyformance import gauge, global_registry
 from inmanta.vendor.pyformance.meters.gauge import AnyGauge, CallbackGauge
+from packaging import version
 
 LOGGER = logging.getLogger(__name__)
 
 
+@total_ordering
+class PostgreSQLVersion:
```
Hugo-Inmanta, line 43: This `PostgreSQLVersion` class was moved as is from `bootloader.py`.

src/inmanta/server/bootloader.py
Hugo-Inmanta: I noticed and decided to use `client_port` instead + added a comment.

src/inmanta/server/bootloader.py
Hugo-Inmanta: As discussed, moved this to `databaseservice.py`.

docs/install/1-install-server-with-podman-and-systemd.rst
```diff
@@ -239,6 +239,13 @@ The file ``inmanta-db.container`` defines the database container. The data is st
 
 Replace ``<db_username>`` and ``<db_password>`` with respectively the username and password you want to use to authenticate to the database server.
 
+.. note::
+
+    During its boot sequence, the orchestrator will perform some checks on the database (e.g. checking that the
+    postgresql version is compatible and checking the status of the replica servers in a high-availability setup).
+    Please make sure the configured user holds the ``pg_monitor`` role to enable these checks during startup.
```
arnaudsjs:
```suggestion
    Please make sure the configured postgresql user holds the ``pg_monitor`` role to enable these checks during startup.
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
...
+                    opt.db_host.get(),
+                    opt.db_port.get(),
+                )
+                return conn
+            except asyncio.TimeoutError:
+                LOGGER.info("Waiting for database to be up: Connection attempt timed out.")
+            except Exception:
+                LOGGER.info("Waiting for database to be up.", exc_info=True)
+            # Check if the maximum wait time has been exceeded
+            if 0 < db_wait_time < asyncio.get_event_loop().time() - start_time:
```
jptrindade, line 441: Isn't the  `0 <` redundant

jptrindade: nvm saw the documentation on wait_time

jptrindade (approved):

tests/test_server.py
```diff
@@ -598,25 +599,57 @@ async def test_bootloader_connect_running_db(
...
+                    config.Config.get("database", "host"),
+                    config.Config.get("database", "port"),
+                ),
             )
+        log_contains(
+            caplog,
+            "inmanta.server.services.databaseservice",
+            logging.INFO,
+            "Database replication is not active: couldn't find any standby server directly connected to the primary. "
+            "If you intend to use a high availability setup, please check the status and the configuration "
```
arnaudsjs: Many users use the high-durability setup instead of the high-availability setup, so let's keep this sufficiently generic. 
```suggestion
            "If you intend to use database replication, please check the status and the configuration "
```

tests/test_server.py
```diff
@@ -598,25 +599,57 @@ async def test_bootloader_connect_running_db(
     ibl: InmantaBootloader = InmantaBootloader(configure_logging=True)
 
     caplog.clear()
-    caplog.set_level(logging.INFO)
+    caplog.set_level(logging.DEBUG)
```
arnaudsjs, line 612: I think it would be nice to update this test case so that we verify that the database schema has not been updated when the server fails to start because of an incompatible database version.

src/inmanta/server/bootloader.py
```diff
@@ -104,7 +99,7 @@ def __init__(self, configure_logging: bool = False) -> None:
     async def start(self) -> None:
         self.start_loggers_for_extensions()
 
-        await self.check_database_before_server_start()
+        await check_database_before_server_start()
```
arnaudsjs: Why don't we initiate this call from the `start()` method of the database slice?

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
 
     LOGGER.info("Connected to PostgreSQL database %s on %s:%d", database_name, database_host, database_port)
     return pool
+
+
+async def check_database_before_server_start() -> None:
```
arnaudsjs: I think readability would increase by a lot if we make this a class instead of a nested function.

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
...
+    """
+    Ensure database connectivity before starting the server and log information about the DB server
+    e.g. the PostgreSQL version of the db server and the status of the replica servers.
```
arnaudsjs: The official term for this is standby server.
```suggestion
    e.g. the PostgreSQL version of the db server and the status of the standby servers.
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
...
+                # Attempt to create a database connection
+                conn = await get_db_connection()
+                LOGGER.info(
+                    "Successfully reached database '%s' at %s:%s.",
+                    opt.db_name.get(),
+                    opt.db_host.get(),
+                    opt.db_port.get(),
+                )
+                return conn
+            except asyncio.TimeoutError:
```
arnaudsjs: `asyncio.TimeoutError` is deprecated in favor of `TimeoutError`.

```suggestion
            except TimeoutError:
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
...
+                LOGGER.info("Waiting for database to be up: Connection attempt timed out.")
+            except Exception:
+                LOGGER.info("Waiting for database to be up.", exc_info=True)
```
arnaudsjs, line 454:
```suggestion
                LOGGER.debug(
                    "Trying to establish a connection to database '%s' at %s:%s.",
                    opt.db_name.get(),
                    opt.db_host.get(),
                    opt.db_port.get(),
                )
                # Attempt to create a database connection
                conn = await get_db_connection()
                LOGGER.debug(
                    "Successfully reached database '%s' at %s:%s.",
                    opt.db_name.get(),
                    opt.db_host.get(),
                    opt.db_port.get(),
                )
                return conn
            except asyncio.TimeoutError:
                LOGGER.debug("Waiting for database to be up: Connection attempt timed out.")
            except Exception:
                LOGGER.debug("Waiting for database to be up.", exc_info=True)
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
...
+            # Check if the maximum wait time has been exceeded
+            if 0 < db_wait_time < asyncio.get_event_loop().time() - start_time:
+                LOGGER.error("Timed out waiting for the database to be up.")
+                raise Exception("Database connection timeout after %d seconds." % db_wait_time)
```
arnaudsjs: I don't see why these two messages are different.
```suggestion
                LOGGER.error("Timeout: database server not up after %d seconds.", db_wait_time)
                raise Exception(f"Timeout: database server not up after {db_wait_time} seconds.")
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
...
+            if database_postgresql_version < required_postgresql_version:
+                raise protocol.ServerStartFailure(
+                    f"The database at {opt.db_host.get()} is using PostgreSQL version "
+                    f"{database_postgresql_version}. This version is not supported by this "
+                    "version of the Inmanta orchestrator. Please make sure to update to PostgreSQL "
+                    f"{required_postgresql_version}."
+                )
+        LOGGER.info("Database is running PostgreSQL server version %s.", database_postgresql_version)
+
+    async def _database_log_replication_status(conn: asyncpg.Connection) -> None:
```
arnaudsjs:
```suggestion
    async def _log_database_replication_status(conn: asyncpg.Connection) -> None:
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
...
+            sent_lsn,               -- Last WAL LSN (Log Sequence Number) sent on this connection
+            write_lsn,              -- Last WAL LSN written to disk on the replica
+            flush_lsn,              -- Last WAL LSN flushed to disk on the replica
+            replay_lsn,             -- Last WAL LSN replayed into the database on the replica
+            pg_wal_lsn_diff(sent_lsn, replay_lsn) AS replay_lag_bytes
+                                    -- This diff represents how far behind this replica lags.
+        FROM pg_stat_replication;
+        """
+        result = await conn.fetch(query)
+        if len(result):
```
arnaudsjs: This is more pythonic:
```suggestion
        if result:
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
...
+            if result[0]["client_port"] is None:
+                # If a user that doesn't hold the 'pg_monitor' role tries to query the pg_stat_replication table, no error
+                # is raised, but instead a 'censored' view is returned with most values filled with NULL.
+                # If the returned value for the port is None, we assume that the user has insufficient privileges, since it
+                # shouldn't be None (As per the docs: "TCP port number that the client is using for communication with this
+                # WAL sender, or -1 if a Unix socket is used").
+                LOGGER.info(
```
arnaudsjs:
```suggestion
                LOGGER.warning(
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +379,199 @@ async def initialize_database_connection_pool(
...
+                    "Cannot check database replication status: insufficient privileges for user %s. Please "
+                    "make sure the configured user has the `pg_monitor` role.",
+                    opt.db_username.get(),
+                )
+            else:
+                LOGGER.info(
+                    "Database replication status (Only the healthy standby servers directly connected to the "
+                    "primary will appear below: downstream replicas or nodes that are down won't appear.)"
```
arnaudsjs: This is not really true, is it? Unhealthy standby nodes will be shown as long as they are connected to the server. Otherwise the message in the for loop would be redundant.

```suggestion
                    "Database replication status (Only the directly connected standby servers "
                    " will appear below: downstream replicas or nodes that are down won't appear.)"
```

arnaudsjs (changes requested):
Several comments are not related to the changes you made, but still improvements.

src/inmanta/server/services/databaseservice.py
arnaudsjs: Now. I see why these exception blocks log at info level. Otherwise the context of the original exception is lost. So we might have to leave those as they are.

src/inmanta/server/services/databaseservice.py
arnaudsjs: Other code paths raise `ServerStartFailure`.
```suggestion
                raise ServerStartFailure("Database connection timeout after %d seconds." % db_wait_time)
```

tests/test_server.py
```diff
@@ -695,16 +734,9 @@ async def test_bootloader_start_no_compatibility_file(tmp_path, server_config, p
         await ibl.start()
         log_contains(
             caplog,
-            "inmanta.server.bootloader",
+            "inmanta.server.services.databaseservice",
             logging.DEBUG,
-            "No compatibility file is set. Bypassing minimal required postgres version check.",
```
arnaudsjs, line 700: Why do we no longer check for this line?

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: In my mind "healthy" was equivalent to  "not down" but your message leads me to believe that this is not the case.

tests/test_server.py
Hugo-Inmanta: I added back the message logging the postgresql version 👍

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: > Now. I see why these exception blocks log at info level. Otherwise the context of the original exception is lost. So we might have to leave those as they are.

I don't get it, what would happen if we were to change the level to debug ?

src/inmanta/server/services/databaseservice.py
arnaudsjs: To me healthy means that the server is up and working correctly. Consider the scenario where a standby server is up but it's lagging behind a lot, which  prevents write operation on the primary being committed. In that case it might be misleading to say that the standby node is healthy, because users might get the impression that the Inmanta server determined that everything is fine with that primary node, while that's not true.

src/inmanta/server/services/databaseservice.py
arnaudsjs: The idea behind changing from info to debug level is to not have these recurring messages in the log on every retry. But if there is a timeout, we still want to log all the information that is required to debug the issue without having to increase the log level to debug and restarting the server again. The exception/log message that is used where there is a timeout only says that a timeout happened, but it doesn't provide any information about what went wrong.

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +382,200 @@ async def initialize_database_connection_pool(
...
+        """
+        This method attempts to connect to the database and returns the connection object.
+        The caller is reponsible for closing the connection.
+
+        The database.wait_time option controls the retry behaviour for this method:
+            database.wait_time < 0 : keep retrying forever until a connection is established
+            database.wait_time = 0 : try only once to establish a connection
+            database.wait_time > 0 : keep retrying until this wait_time timeout is reached or a connection is established
+
+        :raises Exception: If the connectivity cannot be established within the configured database.wait_time.
```
arnaudsjs:
```suggestion
        :raises ServerStartFailure: If the connectivity cannot be established within the configured database.wait_time.
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +382,200 @@ async def initialize_database_connection_pool(
...
+        db_settings = {
+            "host": opt.db_host.get(),
+            "port": opt.db_port.get(),
+            "user": opt.db_username.get(),
+            "password": opt.db_password.get(),
+            "database": opt.db_name.get(),
+        }
+
+        # Attempt to create a database connection
+        return await asyncpg.connect(**db_settings, timeout=5)  # raises TimeoutError after 5 seconds
```
arnaudsjs: Shouldn't we convert this TimeoutError to a ServerStartFailure?

arnaudsjs (changes requested):

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: Very good point, thanks for the example :+1:

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: I don't think we can since we want to keep retrying e.g. [here](https://github.com/inmanta/inmanta-core/blob/2a2c848aafe5c11a701a04e2edef252840145d6f/src/inmanta/server/services/databaseservice.py#L451-L452) in the 'retry' branch of the `_database_connectivity_check` method.
I think what we can do is catch it in the 'try once' branch of the `_database_connectivity_check` method and convert it there.

arnaudsjs: Good point

src/inmanta/server/services/databaseservice.py
```diff
@@ -300,3 +382,204 @@ async def initialize_database_connection_pool(
...
+            # Wait for the database to be up before starting the server
+            LOGGER.info("Waiting until database server is up.")
+            conn = await self._wait_for_db(db_wait_time=db_wait_time)
+        else:
+            LOGGER.debug("Not waiting until the database server is up because database.wait_time option is set to 0.")
+            timeout = 5
+            try:
+                conn = await self._get_db_connection(timeout=timeout)
+            except TimeoutError:
+                raise ServerStartFailure("Timeout: database server not up after %d seconds." % timeout)
```
arnaudsjs: I don't think we can say this because in this code path the per-requisite is that the database server is already up and there can be other reasons why we can't connect to the database server. There might be a networking/firewall issue for example. 
```suggestion
                raise ServerStartFailure("Timeout: Failed to connect to the database.")
```

arnaudsjs (approved):

PR #10293, opened 2026-04-17
Support fetching active configuration

arnaudsjs:
# Description

Added support to the configuration framework to get the full active configuration. This feature is required by https://github.com/inmanta/inmanta-support/pull/971

closes inmanta/inmanta-support#960

# Self Check:

- [x] Changelog entry


src/inmanta/protocol/auth/auth.py
```diff
@@ -195,30 +201,21 @@ def reset(cls) -> None:
...
-        # Save config found on environment variables
         env_config: dict[str, dict[str, str]] = defaultdict(dict[str, str])
-
-        for setting in found_settings:
-            for possible_setting in ENV_AUTH_JWT_SETTINGS:
-                if setting.endswith(possible_setting):
-                    # jwt-username-claim breaks the convention of using dashes instead of underscores
-                    setting_name = (
-                        "jwt-username-claim" if possible_setting == "JWT_USERNAME_CLAIM" else possible_setting.lower()
-                    )
```
arnaudsjs, line 211: I dropped this part because it's redundant. This ConfigParser is written to automatically normalize the name of read and write, so it's resilient to this. By dropping it here, this method returns all the configuration options in a consistent format.

src/inmanta/protocol/auth/auth.py
```diff
@@ -195,30 +201,21 @@ def reset(cls) -> None:
...
-                    # The -1 is to take the underscore into account
-                    section_name = setting[env_prefix_len : -len(setting_name) - 1].lower()
-                    env_config[AUTH_JWT_PREFIX + section_name][setting_name] = str(os.environ.get(setting))
-                    break
-            else:
+        for name, value in os.environ.items():
```
arnaudsjs, line 205: This is just a refactor to improve performance and readability.

src/inmanta/protocol/auth/auth.py
jptrindade: Did not know that. Nice improvement!

tests/test_jwt.py
```diff
@@ -161,29 +161,57 @@ async def test_auth_jwt_environment_variables(jwks, tmp_path, monkeypatch, caplo
         f"http://localhost:{port}/auth/realms/inmanta/protocol/openid-connect/certs",
     )
     monkeypatch.setenv(f"{auth.ENV_AUTH_JWT_PREFIX}{new_config}VALIDATE_CERT", "false")
-    monkeypatch.setenv(f"{auth.ENV_AUTH_JWT_PREFIX}{new_config}EXPIRE", "20")
```
jptrindade, line 164: what was wrong with this line and the other you removed?

arnaudsjs: I removed it because it's actually invalid/inconsistent configuration, which could be confusing for the person reading the test case.

* When sign=False the server is not allowed to issue/sign tokens. Hence, the expire flag doesn't mean anything, because there is nothing to expire. 
* The key field is intended to contain the symmetric key used to sign/validate tokens. But here the algorithm is set to `RS256`, which means that we rely on asymmetric keys. Hence the key field doesn't mean anything in this context.

The `get_active_configuration_as_dict()` method is written to take these variations into account.

arnaudsjs:
I noticed there is an inconsistency. I return the JWT settings with underscores, while all other configuration uses dashes. I'll fix that.

arnaudsjs:
> I noticed there is an inconsistency. I return the JWT settings with underscores, while all other configuration uses dashes. I'll fix that.

Done.

changelogs/unreleased/support-collecting-full-active-configuration.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: "Added support to the configuration framework to get the full active configuration."
+issue-nr: 960
+issue-repo: inmanta-support
+change-type: minor
+destination-branches: [master, iso9, iso8]
```
sanderr, line 6: Are we using this on iso8? The other PR is iso9+.

src/inmanta/config.py
```diff
@@ -149,10 +155,40 @@ def _save_loaded_config(
...
+    def get_active_configuration_as_dict(cls) -> typing.Mapping[str, typing.Mapping[str, object]]:
+        """
+        Returns the values of the configuration options that are currently used (active configuration).
+
+        :returns: A dictionary that maps the configuration section name to a dictionary that maps
+                  the name of a configuration option to its value. This dictionary contains the
+                  active configuration exhaustively, i.e. values that were not explicitly set
+                  will be present in this dictionary using their default value.
+        """
+        from inmanta.protocol.auth import auth
```
sanderr, line 177: I don't love this coupling, but we may have no better options? I'm not very familiar with this part of the code.

src/inmanta/protocol/auth/auth.py
```diff
@@ -336,6 +345,30 @@ def __init__(self, name: str, section: str, config: configparser.SectionProxy) -
         else:
             raise ValueError(f"Algorithm {self.algo} in {self.section} is not support ")
 
+    def get_as_dict(self) -> dict[str, object]:
```
sanderr, line 348: This method is a bit too far out of my comfort zone to review properly. If you're not confident in it, let me know but I'll need more context.

sanderr (changes requested):

src/inmanta/protocol/auth/auth.py
arnaudsjs: Are you talking about the if-then-else structures in the body of the method to determine what should be included in the result and what not? In that case I am pretty confident, but I can see it's hard to review. It took me some time to get familiar with this.

changelogs/unreleased/support-collecting-full-active-configuration.yml
arnaudsjs: This PR is a building block for a bug fix on the support tool. The bug also exists on iso8, so I think we need to merge this on iso8 too. I didn't put iso8 in the PR on inmanta-support, because it needs a different version constraint in the setup.py. The plan was to make an iso8-specific PR, once the current PR on inmanta-support is approved.

src/inmanta/config.py
arnaudsjs: It's rather hard to get around. In many paces we import config to define config options or to fetch config values.

jptrindade (approved):

src/inmanta/protocol/auth/auth.py
sanderr: That part indeed.

src/inmanta/config.py
sanderr: Yeah, I meant more that the generic config now has to be aware of specific "child" configs. But if it can't be helped, so be it, it's not a huge deal.

sanderr (approved):

PR #10306, opened 2026-04-20
Graphql improvement patch hack

jptrindade:
# Description

Changed count function to use resource persistent state table.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -842,7 +842,7 @@ async def get_connection(
     after: typing.Optional[str] = strawberry.UNSET,
     last: typing.Optional[int] = strawberry.UNSET,
     before: typing.Optional[str] = strawberry.UNSET,
-    count_stmt: Select[typing.Any] | None = None,
+    count_stmt: Select[tuple[int]] | None = None,
 ) -> CustomListConnection[NodeType]:
     """
     Build the connection object. Here we do all the pagination and fetching of results (edges) to return to the user.
@@ -1063,10 +1063,11 @@ def get_schema(context: GraphQLContext) -> strawberry.Schema:
                     ),
                 )
             stmt = add_filter_and_sort(stmt, ResourceOrder.default_order(), filter, order_by)
-            count_stmt: Select[typing.Any] | None
-            if filter is not None and filter.purged is not strawberry.UNSET:
+            count_stmt: Select[tuple[int]] | None
+            if filter.purged is not strawberry.UNSET:
                 count_stmt = None
             else:
+                # more efficient count statement that doesn't require joining on resource
                 count_stmt = add_filter_and_sort(select(func.count()).select_from(models.ResourcePersistentState), {}, filter)
             return await get_connection(
                 stmt, info=info, model="Resource", first=first, after=after, last=last, before=before, count_stmt=count_stmt
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -12,6 +12,7 @@ limitations under the License.
 Contact: code@inmanta.com
 """
 
+import logging
 import base64
 import dataclasses
 import inspect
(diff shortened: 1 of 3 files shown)


src/inmanta/graphql/schema.py
```diff
@@ -758,9 +742,7 @@ class ResourceOrder(StrawberryOrder):
     @classmethod
     def default_order(cls) -> dict[str, UnaryExpression[typing.Any]]:
         return {
-            "environment": asc(models.Resource.environment),
-            "resource_set": asc(models.Resource.resource_set),
-            "resource_id": asc(models.Resource.resource_id),
+            "resource_id": asc(models.ResourcePersistentState.resource_id),
```
sanderr, line 745: For the follow-up with the indexes: I just realized we might (to be verified) get a lot of coverage out of just a few indexes if we decompose the resource id into agent, resource type and resource id value, and change the default order accordingly. Something to keep in mind for the next stage.

src/inmanta/graphql/schema.py
```diff
@@ -887,6 +842,7 @@ async def get_connection(
     after: typing.Optional[str] = strawberry.UNSET,
     last: typing.Optional[int] = strawberry.UNSET,
     before: typing.Optional[str] = strawberry.UNSET,
+    count_stmt: Select[typing.Any] | None = None,
```
sanderr: Why `Any`? Do you mean we don't care? Then it should be `object`.

src/inmanta/graphql/schema.py
```diff
@@ -1095,8 +1063,14 @@ async def resources(
                     ),
                 )
             stmt = add_filter_and_sort(stmt, ResourceOrder.default_order(), filter, order_by)
-            stmt = do_required_resource_joins(stmt, filter, order_by)
-            return await get_connection(stmt, info=info, model="Resource", first=first, after=after, last=last, before=before)
+            count_stmt: Select[typing.Any] | None
+            if filter is not None and filter.purged is not strawberry.UNSET:
```
```suggestion
            if filter.purged is not strawberry.UNSET:
```
My bad, can't be `None`.

src/inmanta/graphql/schema.py
```diff
@@ -1095,8 +1063,14 @@ async def resources(
...
+                count_stmt = None
+            else:
+                count_stmt = add_filter_and_sort(select(func.count()).select_from(models.ResourcePersistentState), {}, filter)
```
sanderr, line 1071:
```suggestion
                # more efficient count statement that doesn't require joining on resource
                count_stmt = add_filter_and_sort(select(func.count()).select_from(models.ResourcePersistentState), {}, filter)
```

sanderr (changes requested):
Looks good, some minor comments. Running my benchmarks on this branch as a final sanity check that we didn't miss anything in the cleanup.

src/inmanta/graphql/schema.py
jptrindade: I tried to fix this typing, it is not as simple as `object`, but I think I found a good solution.
At first I was doing the new syntax `[T: tuple[object]]` but I found that it was being repeated a lot and figured it would be cleaner to have it like this. wdyt?

src/inmanta/graphql/schema.py
jptrindade: That is interesting. So we would get a sort on resource_id and we would decompose it into its parts to execute it? Would we also do this for filters?

sanderr: I don't think we expose resource id at all over the graphql API at the moment, only its parts. I meant that if we make our indexes use those parts as well, and this fallback sorting as well, then we're more in line with what the user might request, and possible more flexible when it comes to variations (no filter sort on resource type vs filter on agent and resource type sort on resource id value vs ...)

jptrindade: Hmm I see, the designs for the resource view only include the parts of resource_id

sanderr (approved):

PR #10328, opened 2026-04-27
Draft: Add ResourceFilter composition

jptrindade:
# Description

Introduces `ResourceFilterABC` which any extension can implement and register on the `GraphQLSlice`. 
These implementations are then composed into a single `ResourceFilter` that is then exposed to the user on GraphQL.

# Self Check:

- [x] Changelog entry


src/inmanta/graphql/graphql.py
```diff
@@ -12,37 +12,77 @@
...
+    def build_resource_filter(self) -> type:
+        if self._composed_resource_filter_cls is None:
+            bases = tuple(self.resource_filters)
+
+            class ResourceFilter(*bases):
+                def apply_filters(self, stmt):
+                    for base in bases:
+                        # Call each extension's apply_filters independently
+                        if hasattr(base, "apply_filters"):
+                            stmt = base.apply_filters(self, stmt)
```
sanderr: This raises the question to me: do we want to call each extension's filter, or only if any of their filter fields are set?

sanderr: > `base.apply_filters(self, stmt)`

Good enough for the PoC, but something to figure out how we'll clean it up for the actual implementation. `StrawberryFilter` defines `apply_filter` as an instance method. You typically call them on an instance. That is what you would expect. But you now forcefully call them on the class and have to pass in the instance.

I think that one solution would be the separation between data and logic that I drafted, but that had complications of its own. Another could perhaps be to use a class method instead.

src/inmanta/graphql/graphql.py
```diff
@@ -12,37 +12,77 @@
...
 from strawberry.schema.exceptions import CannotGetOperationTypeError
 from strawberry.types.execution import ExecutionResult
 
+LOGGER = logging.getLogger(__name__)
+
+
+class ExampleResourceFilter(StrawberryFilter):
+    my_attr: StrFilter | None = None
+
+    def apply_filters[*Ts](self, stmt: Select[tuple[*Ts]]) -> Select[tuple[*Ts]]:
```
sanderr: Great for the draft. For the eventual interface it's still unclear to me what we want this to be exactly. Mostly does it modify the select or does it simply return a join directive.

jptrindade: Like a cte that we then join to filter on? 
I feel like modifying the select might be clearer. wdyt?

src/inmanta/graphql/graphql.py
jptrindade: I feel like we can call it always. The extension's apply_filter should (need to?) have logic to determine if they are being called or not.

jptrindade: This way we can also inject some logic even if the extension does not have any filter present

src/inmanta/graphql/schema.py
```diff
@@ -918,11 +937,19 @@ def get_schema(context: GraphQLContext) -> strawberry.Schema:
     """
 
     loader = StrawberrySQLAlchemyLoader(async_bind_factory=get_session_factory())
+    ResourceFilter = context.graphql_service.build_resource_filter()
```
jptrindade: mypy still hates this

src/inmanta/graphql/schema.py
```diff
@@ -209,6 +209,16 @@ def _build_docstring_param_cache() -> dict[str, dict[str, str]]:
 _docstring_param_cache: dict[str, dict[str, str]] = _build_docstring_param_cache()
 
 
+@strawberry.input
+class BaseResourceFilter(ABC):
```
jptrindade: Maybe `ResourceFilterABC` is better

src/inmanta/graphql/schema.py
jptrindade: Is casting this as typing.Any the best solution?

src/inmanta/graphql/graphql.py
```diff
@@ -14,35 +14,91 @@
...
+
+    def apply_resource_filters[*Ts](self, stmt: Select[tuple[*Ts]], filter_instance: ResourceFilterABC) -> Select[tuple[*Ts]]:
+        """
+        Applies each registered ResourceFilterABC implementation's `apply_filter` to the provided Select statement.
+
+        :param stmt: The Select statement that fetches the requested resources.
+        :param filter_instance: The composed ResourceFilter instance that contains the requested filters.
+        """
+        for filter_cls in self.all_filters:
+            stmt = filter_cls.apply_filter(stmt=stmt, filter_instance=filter_instance)
```
sanderr: I wonder if the order ever matters here (apart from that it should be deterministic). I think not? i.e. we could safely call this on the lsm filters first and only then on the core ones?

src/inmanta/graphql/graphql.py
```diff
@@ -14,35 +14,91 @@
...
+        self.resource_filter_engine.register_extension_filter(CoreResourceFilter)
+        self.context = GraphQLContext(compiler_service=compiler_service, graphql_service=self)
         await super().prestart(server)
 
+    def update_resource_filter(self, ext_filter: type[ResourceFilterABC]) -> None:
+        """
+        Adds an extension's ResourceFilterABC implementation to be composed into the ResourceFilter that is exposed to the user
+        """
+        self.resource_filter_engine.register_extension_filter(ext_filter)
+        self._composed_resource_filter_cls = None
```
sanderr: Do we really need to be *this* dynamic? I think it could be a bit more robust / consistent if we can make sure that we initialize once to start with, and reject additional filters afterwards. Wdyt?

src/inmanta/graphql/graphql.py
```diff
@@ -14,35 +14,91 @@
...
 from inmanta.protocol.decorators import handle
 from inmanta.server import SLICE_COMPILER, SLICE_GRAPHQL, protocol
 from inmanta.server.protocol import Server
 from inmanta.server.services.compilerservice import CompilerService
+from sqlalchemy import Select
 from strawberry.schema.exceptions import CannotGetOperationTypeError
 from strawberry.types.execution import ExecutionResult
 
 
+class ResourceFilterEngine:
```
sanderr: If we move this into `graphql.schema` (and perhaps make a singleton by operating at class level rather than instance level)?, can we get rid of the whole passing around of the slice and engine that's going on there? Or am I missing something?

src/inmanta/graphql/schema.py
```diff
@@ -209,6 +208,25 @@ def _build_docstring_param_cache() -> dict[str, dict[str, str]]:
...
+    Abstract base class that defines the shared attributes/behaviour of each extension's ResourceFilter.
+
+    :param environment: The environment the resources belong to.
+    :param is_orphan: If we want to filter on orphans or not.
+    :param purged: If we want to filter on the "purged" attribute or not.
+    """
+
+    environment: uuid.UUID
+    is_orphan: bool | None = strawberry.UNSET
+    purged: bool | None = strawberry.UNSET
```
sanderr: Why do we need `purged` at this level?

src/inmanta/graphql/schema.py
```diff
@@ -209,6 +208,25 @@ def _build_docstring_param_cache() -> dict[str, dict[str, str]]:
...
+    """
+    Abstract base class that defines the shared attributes/behaviour of each extension's ResourceFilter.
+
+    :param environment: The environment the resources belong to.
+    :param is_orphan: If we want to filter on orphans or not.
+    :param purged: If we want to filter on the "purged" attribute or not.
+    """
+
+    environment: uuid.UUID
+    is_orphan: bool | None = strawberry.UNSET
```
sanderr: Do we still need `is_orphan` at this level, considering what we discussed earlier this week? i.e. lsm filters are always about the latest intent (we may still need to figure out how to enforce that, but that's a different matter).

src/inmanta/graphql/schema.py
```diff
@@ -209,6 +208,25 @@ def _build_docstring_param_cache() -> dict[str, dict[str, str]]:
...
+    purged: bool | None = strawberry.UNSET
+
+    @classmethod
+    @abstractmethod
+    def apply_filter[*Ts](cls, stmt: Select[tuple[*Ts]], filter_instance: typing.Self) -> Select[tuple[*Ts]]: ...
```
sanderr: Apparently this `Self` isn't really safe. Took it to Slack.

src/inmanta/graphql/schema.py
```diff
@@ -209,6 +208,25 @@ def _build_docstring_param_cache() -> dict[str, dict[str, str]]:
...
+class ResourceFilterABC(ABC):
+    """
+    Abstract base class that defines the shared attributes/behaviour of each extension's ResourceFilter.
+
+    :param environment: The environment the resources belong to.
+    :param is_orphan: If we want to filter on orphans or not.
+    :param purged: If we want to filter on the "purged" attribute or not.
+    """
+
+    environment: uuid.UUID
```
sanderr: With this move we'll have to make sure to not reintroduce the bug where we omit the environment filter. But I guess the test you expanded will protect us from that.

src/inmanta/graphql/schema.py
```diff
@@ -794,14 +815,19 @@ class ComposedResourceSummary:
...
+    filter: typing.Optional[StrawberryFilter | ResourceFilterABC] = strawberry.UNSET,
     order_by: typing.Optional[Sequence[StrawberryOrder]] = strawberry.UNSET,
+    resource_filter_engine: typing.Any | None = None,
 ) -> Select[tuple[*Ts]]:
     """
     Adds filter and sorting to the given statement.
     """
     if filter is not None and filter is not strawberry.UNSET:
-        stmt = filter.apply_filters(stmt)
+        if isinstance(filter, ResourceFilterABC):
```
sanderr: I still need to give this some more thought to see if we can do anything about the clunkiness here. No immediate ideas yet.

sanderr (commented):
I still wonder if we should / can restrict what `apply_filters()` can do. And I'm still leaning towards yet. *But*, I think that perhaps we should carry on like this first, and once we have a clear idea of what lsm needs exactly and how it will all fit together (could even be after the first merge), we can come back to this consideration and see if we can streamline that interface just a bit more.

src/inmanta/graphql/schema.py
jptrindade: I put `is_orphan` and `purged` at this level to make mypy happy because they are used in the [`resources` query.](https://github.com/inmanta/inmanta-core/pull/10328/changes#diff-89cbe29f149a1f7504e2f03151bf3dd182a5d01fd133b5ea53a9a342beb0452dR1007)

src/inmanta/graphql/graphql.py
jptrindade: I believe that the order does not matter.

src/inmanta/graphql/schema.py
sanderr: I see.
1. I worry that `is_orphan` might cause us trouble in the next stage, but let's see about that when we get there. For now it's fine here.
2. I'm pretty sure that we'll have to update the condition that we use to determine whether or not we can use the simplified count (a complication I hadn't thought of until just now). Once we have addressed that, if we're lucky `purged` will have disappeared from the `resources` query.

src/inmanta/graphql/graphql.py
```diff
@@ -16,42 +16,68 @@
...
+        self.schema = None
+        self.all_filters = {}
 
     def get_dependencies(self) -> list[str]:
         return [SLICE_COMPILER]
 
+    def register_extension_filter(self, extension_name: str, filter_cls: type[ResourceFilterABC]) -> None:
+        """
+        Register an extension filter.
+        This is only possible if the composed ResourceFilter has yet to be generated
```
sanderr: I think this can be improved to better address the caller. I.e. "the composed ResourceFilter has yet to be generated" is an implementation detail. The corresponding client contract is that it must be called before this slice starts, so during the `prestart` stage, or the start stage of a slice that registers this one as a dependency.

src/inmanta/graphql/graphql.py
```diff
@@ -16,42 +16,68 @@
...
         assert isinstance(compiler_service, CompilerService)
-        self.context = GraphQLContext(compiler_service=compiler_service)
+        self.compiler_service = compiler_service
+        self.all_filters[SLICE_GRAPHQL] = CoreResourceFilter
         await super().prestart(server)
 
+    async def start(self) -> None:
+        assert self.compiler_service is not None
+        self.schema = get_schema(
+            GraphQLContext(compiler_service=self.compiler_service, resource_filter_components=list(self.all_filters.values()))
```
sanderr: Why pass this through the context? It requires quite some wiring, and I don't think it really offers us anything?

src/inmanta/graphql/graphql.py
```diff
@@ -16,42 +16,68 @@
...
+            raise Exception(
+                f"Can't register extension filter for {extension_name} because the GraphQL schema has already been generated"
+            )
+
     async def prestart(self, server: Server) -> None:
         compiler_service = server.get_slice(SLICE_COMPILER)
         assert isinstance(compiler_service, CompilerService)
-        self.context = GraphQLContext(compiler_service=compiler_service)
+        self.compiler_service = compiler_service
+        self.all_filters[SLICE_GRAPHQL] = CoreResourceFilter
```
sanderr: Bit of a nitpick, and I don't care too strongly, but I think it could turn out slightly cleaner if we move this responsibility to `get_schema`. i.e. the slice knows about extensions, the schema knows about core and how to extend it with extensions. Apart from the slightly debatable semantical improvement, I think it offers the following benefits:
- We get rid of the back-and-forth where we import a filter from the schema here, to pass it back down to the schema. This flow feels a bit weird to me.
- We make the `typing.Annotated` a bit more safe, because if `get_schema` is the one that injects the core filter, than we actually know that the generated type will be an instance of it. Otherwise we introduce more coupling, because the `typing.Annotated` depends on the assumption that the slice injects the core resource filter.

src/inmanta/graphql/graphql.py
sanderr: I think I see. You use the context so you can access it in the `resources` method. But wouldn't it be a lot simpler to simply store it as a class var on the `Query` object? Or even just use the value from the closure, though I think I prefer to make it explicit on `Query`.

src/inmanta/graphql/schema.py
```diff
@@ -1064,13 +1096,31 @@ async def resources(
                         == select(latest_scheduled_version_cte.c.version).scalar_subquery(),
                     ),
                 )
-            stmt = add_filter_and_sort(stmt, ResourceOrder.default_order(), filter, order_by)
+
+            # Decompose ResourceFilter into its parts
+            resource_filter_instances: list[ResourceFilterABC] = []
+            for filter_type in info.context.get("resource_filter_components"):
+                filter_fields = {
+                    field.name: getattr(filter, field.name, strawberry.UNSET) for field in dataclasses.fields(filter_type)
```
sanderr: Why use `getattr()`? To be extra robust? It normally shouldn't ever be absent, should it?

src/inmanta/graphql/schema.py
```diff
@@ -794,14 +795,15 @@ class ComposedResourceSummary:
...
+    filter: Sequence[StrawberryFilter],
     order_by: typing.Optional[Sequence[StrawberryOrder]] = strawberry.UNSET,
 ) -> Select[tuple[*Ts]]:
     """
     Adds filter and sorting to the given statement.
     """
-    if filter is not None and filter is not strawberry.UNSET:
-        stmt = filter.apply_filters(stmt)
+    for filter_instance in filter:
+        if filter_instance is not strawberry.UNSET:
```
sanderr: Not necessarily in scope for this PR, but why do we check for `UNSET` actually? It's not of type `StrawberryFilter`, is it?

src/inmanta/graphql/graphql.py
jptrindade: I tried setting it as a variable of `Query` but got some errors, needs to be a GraphQL valid type and I got some conflicts there. And I feel like `context` is a good "container" for all the small things that we get from the slice to the schema.

src/inmanta/graphql/graphql.py
jptrindade: Not sure what you mean by this. So we wouldn't have `all_filters` as  a concept on the slice? Or do you mean that `GraphQLSlice` does not register `CoreResourceFilter` but keeps all the other filters, and then we just append `CoreResourceFilter` on `get_schema`?

src/inmanta/graphql/schema.py
jptrindade: I'm not sure how would the alternative look like. Or do you mean the default to `strawberry.UNSET`?

src/inmanta/graphql/graphql.py
sanderr: Pretty much. Well, I think prepend would be safest, but that's beside the point.

src/inmanta/graphql/schema.py
jptrindade: I think this is just an iteration artifact. I check for `UNSET` before I call this when applicable

src/inmanta/graphql/schema.py
sanderr: I was somewhat mistaken here, sorry. What I meant indeed boils down to the default.

jptrindade: I remember I ran into some issues without the default. But that may have been from a previous iteration where we where converting to dict.

The tests seem to pass without the default, I'll remove it so we can more easily diagnose if something goes wrong

jptrindade:
I updated the draft to match the design. There are still some mypy issues that I might need some assistance with.

EDIT: I tried on the return object of the resources query `CustomListConnection[typing.Annotated[CoreResourceBase, strawberry.argument(graphql_type=composed_resource_type)]]` but it did not work with the other resource mixins, meaning that their fields were not available to be queried

src/inmanta/graphql/schema.py
```diff
@@ -912,7 +918,82 @@ async def get_connection[*Ts](
         )
 
 
-def get_schema(context: GraphQLContext) -> strawberry.Schema:
+class GraphQLContribution(ABC):
+    """
+    Extension hook that lets modules (e.g. LSM) contribute extra filter fields, SQL logic,
```
```suggestion
    Extension hook that lets extensions (e.g. LSM) contribute extra filter fields, SQL logic,
```

src/inmanta/graphql/schema.py
```diff
@@ -912,7 +918,82 @@ async def get_connection[*Ts](
...
+    and output fields to the core ``resources`` GraphQL query without core being aware of those modules.
+    """
+
+    @classmethod
+    def get_resource_filter_input_class(cls) -> type[ResourceFilterABC] | None:
```
sanderr, line 982: Interesting. I believe I proposed something like this at some point and we didn't go this way. Why did you change it? I assume it has to do with the fact that we need both input and output? How / when do they differ in their fields?

EDIT: Ah, I guess "meta" filters like `includeOwned`... We could probably also use annotations on the fields to determine whether they're in/out/both. But let's go with this approach for now and see how it ends up. If it looks clean, this is probably the simplest. If it turns out to be too repititive / coupled, perhaps we can reconsider.

src/inmanta/graphql/schema.py
```diff
@@ -912,7 +918,82 @@ async def get_connection[*Ts](
...
+    @classmethod
+    def get_resource_graphql_output_type_mixin(cls) -> type | None:
+        """
+        Return a plain class (no decorator) whose strawberry.field declarations are merged into
+        the Resource output type. Return None if this contribution adds no output fields.
+        """
+        return None
+
+    @classmethod
+    def get_context_loaders(cls) -> dict[str, object]:
```
sanderr, line 995: What's this? The docstring doesn't make it clear to me and I don't see it being used anywhere.

src/inmanta/graphql/schema.py
```diff
@@ -921,10 +1002,17 @@ def get_schema(context: GraphQLContext) -> strawberry.Schema:
 
     loader = StrawberrySQLAlchemyLoader(async_bind_factory=get_session_factory())
 
-    class CustomInfo(Info):
+    composed_resource_filter, resource_filter_components = build_resource_filter(extension_contributions)
+
+    Resource: type = build_resource_return_obj(extension_contributions)
+
+    class CustomInfo(Info[StrawberryInfoContextDict, object]):
```
sanderr, line 1065: I thought we were moving away from this context object?

src/inmanta/graphql/schema.py
```diff
@@ -912,7 +918,82 @@ async def get_connection[*Ts](
...
+    """
+    _filter_extensions: list[type[ResourceFilterABC]] = [
+        cls for ec in extension_contributions if (cls := ec.get_resource_filter_input_class()) is not None
+    ]
+    resource_filter_components: tuple[type[ResourceFilterABC], ...] = (
+        CoreResourceFilter,
+        *_filter_extensions,
+    )
+    return strawberry.input(
+        dataclasses.dataclass(kw_only=True)(type("ResourceFilter", resource_filter_components, {}))
```
sanderr: Wouldn't it be simpler to move this part to the caller so that we don't have to return a tuple? Optionally make it a second helper method `tuple[type[ResourceFilterABC], ...] -> type` if you don't want the complexity in `get_schema()`?

src/inmanta/graphql/schema.py
```diff
@@ -912,7 +918,82 @@ async def get_connection[*Ts](
...
+        mixin for c in extension_contributions if (mixin := c.get_resource_graphql_output_type_mixin()) is not None
+    )
+    _resource_annotations: dict[str, object] = {}
+    _resource_attrs: dict[str, object] = {}
+    _skip_attrs = {"__dict__", "__weakref__", "__doc__", "__annotations__", "__module__"}
+    for _base in (CoreResourceBase, *_resource_mixins):
+        _resource_annotations.update(_base.__dict__.get("__annotations__", {}))
+        for _k, _v in _base.__dict__.items():
+            if _k not in _skip_attrs:
+                _resource_attrs[_k] = _v
```
sanderr, line 1042: This level of meta-programming scares me a bit with respect to stability. I need to think if I see a better way.

src/inmanta/graphql/schema.py
```diff
@@ -912,7 +918,82 @@ async def get_connection[*Ts](
...
+    # Can't do the same as the resource filter because the mixins can't have the mapper.type decorator and that is required
+    return mapper.type(models.Resource)(type("Resource", (), {"__annotations__": _resource_annotations, **_resource_attrs}))
```
sanderr: If we use `mapper.type(models.Resource)`, won't it prevent the query from returning anything other than a `models.Resource` object?

Open question. I don't feel like I understand the framework sufficiently to give an answer myself.

src/inmanta/graphql/graphql.py
```diff
@@ -16,42 +16,67 @@
...
         assert isinstance(compiler_service, CompilerService)
-        self.context = GraphQLContext(compiler_service=compiler_service)
+        self.compiler_service = compiler_service
         await super().prestart(server)
 
+    async def start(self) -> None:
+        assert self.compiler_service is not None
+        self.schema = get_schema(
+            compiler_service=self.compiler_service, extension_contributions=list(self.extension_contributions.values())
+        )
```
sanderr, line 70: Nice. I don't recall if I've already reviewed since this change, but I prefer it a lot over the previous approach.

sanderr (commented):
I tried to do a broad, high level review and to only comment on things that are relevant at that level.

Some things that come to mind that are still missing (there are probably others, I didn't cross check with the design):
1. support for different version filters (we still branch our query on `include_orphans`)

src/inmanta/graphql/schema.py
jptrindade: No. It creates a strawberry type based on the attributes of `models.Resource` but it is not indeed the same object since we can add fields to it (we already do the same with the `purged` attribute for example)

src/inmanta/graphql/schema.py
jptrindade: Yeah, I agree that this is one of the parts that scares me the most, Claude wrote it, for better or worse

src/inmanta/graphql/schema.py
jptrindade: (Also related to the comment above) I don't think we can/should. The way we fetch lsm related information will be through custom resolvers (like the `purged` attribute). I want the extensions to inject stuff that they require here to help in the logic of the custom resolvers. This is an example that Claude made:
```
```python
@dataclasses.dataclass
class _LsmResourceData:
    service_entity: str
    service_instance_id: uuid.UUID
    lifecycle_state: str
```

**Shared DataLoader helper**:

```python
async def _load_lsm_data(root: Any, info: Info) -> Optional[_LsmResourceData]:
    make_loader = info.context.get("lsm_resource_info_loader_factory")
    if make_loader is None:
        return None
    cache_key = f"_lsm_info_loader_{root.environment}"
    loader = info.context.get(cache_key)
    if loader is None:
        loader = make_loader(root.environment)
        info.context[cache_key] = loader
    return await loader.load(root.resource_id)
```

**Three individual field resolvers** (each call `_load_lsm_data`, extract one attribute):

```python
async def _get_lsm_service_entity(root, info) -> Optional[str]: ...
async def _get_lsm_service_instance_id(root, info) -> Optional[uuid.UUID]: ...
async def _get_lsm_lifecycle_state(root, info) -> Optional[str]: ...
```

**Mixin class**:

```python
class _LsmResourceMixin:
    lsm_service_entity:      Optional[str]       = strawberry.field(resolver=..., description="...")
    lsm_service_instance_id: Optional[uuid.UUID] = strawberry.field(resolver=..., description="...")
    lsm_lifecycle_state:     Optional[str]       = strawberry.field(resolver=..., description="...")
```

**Performance**: No impact from having 3 resolvers instead of 1. The DataLoader batches
all `loader.load(resource_id)` calls within the same async tick into a single SQL query,
and caches results by key — subsequent `load()` calls for the same key return the cached
`_LsmResourceData` without a database round-trip. Result: still one SQL query per
environment per page regardless of which or how many inline LSM fields are requested.

#### `LsmResourceFilterFields`

```python
@strawberry.input
class LsmResourceFilterFields:
    service_entity:           Optional[StrFilter]       = strawberry.UNSET
    service_instance:         Optional[list[uuid.UUID]] = strawberry.UNSET
    lifecycle_state:          Optional[StrFilter]       = strawberry.UNSET
    service_instance_version: Optional[int]             = strawberry.UNSET
    include_owned:            Optional[bool]            = strawberry.UNSET
```

#### DataLoader for inline LSM fields

`get_context_loaders()` returns `{"lsm_resource_info_loader_factory": make_loader}` where
`make_loader` is a per-environment factory creating a `DataLoader` on first use.  One
DataLoader instance per environment per request; all resources on the same page share one
batch SQL query.

The batch query fetches all three values in one shot (selecting more columns than a caller
might need is negligible cost compared to an extra round-trip):

```sql
SELECT
    elem                             AS resource_id,
    lt.resource_id_value             AS instance_id,
    lt.attributes->>'service_entity' AS service_entity,
    si.state                         AS lifecycle_state
FROM resource lt
JOIN resource_set_configuration_model rscm ON ...
JOIN (SELECT MAX(version) AS version FROM configurationmodel
      WHERE environment = $1 AND released = TRUE) lv ON rscm.model = lv.version
LEFT JOIN lsm_serviceinstance si
    ON si.id::text = lt.resource_id_value AND si.environment = lt.environment
CROSS JOIN LATERAL jsonb_array_elements_text(lt.attributes->'resources') elem
WHERE lt.environment   = $1
  AND lt.resource_type = 'lsm::LifecycleTransfer'
  AND lt.attributes->>'resources' != '<<undefined>>'
  AND elem = ANY($2)
```
```

PR #10343, opened 2026-04-30
Drop unused increment cache

### LINKED ISSUE #10059: Drop unused increment cache
`inmanta.server.services.resourceservice` still has the `increment_cache` field, which is no longer used on iso9+

jptrindade:
# Description

Drops unused increment cache and small cleanup

closes #10059 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/server/services/resourceservice.py
+++ b/src/inmanta/server/services/resourceservice.py
@@ -16,12 +16,10 @@ limitations under the License.
 Contact: code@inmanta.com
 """
 
-import asyncio
 import datetime
 import logging
 import typing
 import uuid
-from collections import abc, defaultdict
 from collections.abc import Sequence
 from typing import Any, Optional, cast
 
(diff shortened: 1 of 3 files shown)


src/inmanta/server/services/resourceservice.py
```diff
@@ -167,24 +149,6 @@ async def start(self) -> None:
...
-        :param env: The environment that is cleared
-        """
-        self.clear_env_cache(env)
-
-    async def environment_action_deleted(self, env: model.Environment) -> None:
-        """
-        Will be called when the environment is deleted
-        :param env: The environment that is deleted
-        """
-        self.clear_env_cache(env)
```
arnaudsjs, line 186: With these methods gone, there is no reason for this class to be an EnvironmentListener anymore.

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #10353, opened 2026-05-06
Add filter on environment for the totalCount query

### LINKED ISSUE #10352: GraphQL `resources` query missing environment filter for COUNT
*This ticket was created by Claude under my supervision. Most details were provided by me and expanded on by Claude.*

---

## Summary

The `resources` GraphQL query returns an inflated `totalCount` that includes resources from **all environments**, not just the queried one. The result rows themselves are correct; only the count is wrong.

---

## Root Cause

The `resources` resolver in `src/inmanta/graphql/schema.py` uses an optimized count path when `filter.purged` is not set (the common case):

```python
# line ~1076
count_stmt = add_filter_and_sort(
    select(func.count()).select_from(models.ResourcePersistentState),
    {},
    filter,
)
```

This builds `SELECT count(*) FROM resource_persistent_state` and relies on `ResourceFilter.apply_filters()` to add WHERE clauses. However, `apply_filters` did not handle the `environment` field — it processes a fixed list of other fields manually and has no case for `environment`. As a result the query becomes:

```sql
SELECT count(*) FROM resource_persistent_state
-- WHERE environment = <env>  ← MISSING
```

The main result query is unaffected because it adds `.where(models.ResourcePersistentState.environment == filter.environment)` directly (line ~993), before `apply_filters` is invoked.

---

## Fix

Two changes in `src/inmanta/graphql/schema.py`:

```diff
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -730,6 +730,8 @@ class ResourceFilter(StrawberryFilter):
             attr = getattr(self, key)
             if attr is not None and attr is not strawberry.UNSET:
                 stmt = attr.apply_filter(stmt, self.rps_model, key)
+        if self.environment is not None and self.environment is not strawberry.UNSET:
+            stmt = stmt.filter(models.ResourcePersistentState.environment == self.environment)
         if self.purged is not None and self.purged is not strawberry.UNSET:
             stmt = stmt.filter(models.Resource.attributes["purged"].astext.cast(Boolean).is_(self.purged))
         if self.is_deploying is not None and self.is_deploying is not strawberry.UNSET:
@@ -758,6 +760,7 @@ class ResourceOrder(StrawberryOrder):
     @property
     def key_to_model(self) -> dict[str, type[models.Base]]:
         return {
+            "environment": self.rps_model,
             "agent": self.rps_model,
             "resource_type": self.rps_model,
             "resource_id_value": self.rps_model,
```

**Change 1** (`ResourceFilter.apply_filters`): explicitly apply the environment filter so the optimized COUNT path gets the correct WHERE clause.

**Change 2** (`ResourceOrder.key_to_model`): register `environment` as a sortable key so it resolves to the correct model, consistent with the other `rps_model` fields.

---

Slack thread: https://inmanta.slack.com/archives/CKRF0C8R3/p1777878766386949

jptrindade:
# Description

Fix bug where the totalCount field on the resources query of GraphQL was not filtering on environment.

I did not add `environment` to `ResourceOrder.key_to_model` because we are never ordering on environment, since we are always filter on one specific environment.

closes #10352 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/10352-fix-environment-filter-not-present-on-resource-count-query.yml
+++ b/changelogs/unreleased/10352-fix-environment-filter-not-present-on-resource-count-query.yml
@@ -1,0 +1,4 @@
+description: Fix bug where the totalCount field on the resources query of GraphQL was not filtering on environment.
+change-type: patch
+issue-nr: 10352
+destination-branches: [master, iso9]
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -980,16 +980,12 @@ def get_schema(context: GraphQLContext) -> strawberry.Schema:
 
             # Only fetch resources in their latest version
             # Logic based on src/inmanta/data/dataview.py::ResourceView
-            stmt = (
-                select(models.Resource)
-                .join(
-                    models.ResourcePersistentState,
-                    and_(
-                        models.Resource.resource_id == models.ResourcePersistentState.resource_id,
-                        models.Resource.environment == models.ResourcePersistentState.environment,
-                    ),
-                )
-                .where(models.ResourcePersistentState.environment == filter.environment)
+            stmt = select(models.Resource).join(
+                models.ResourcePersistentState,
+                and_(
+                    models.Resource.resource_id == models.ResourcePersistentState.resource_id,
+                    models.Resource.environment == models.ResourcePersistentState.environment,
+                ),
             )
 
             # CTE that fetches the latest scheduled version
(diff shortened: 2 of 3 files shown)


sanderr (approved):
Nice, this is cleaner than my draft.

arnaudsjs (changes requested):
Normally we always add a test case that reproduces the bug to make sure we never introduce the same bug ever again. That was not done in this PR. Why is that?

jptrindade:
> Normally we always add a test case that reproduces the bug to make sure we never introduce the same bug ever again. That was not done in this PR. Why is that?

I added the `setup_database` fixture in the test that we use to query resources. This fixture introduces resources in other environments.

The `test_query_resources` already tests different filters and asserts that the counts are what is expected.

I added the fixture before implementing the changes and noticed that the tests where indeed failing due to environment leakage, so I figured that should be sufficient. I added a docstring on the test that I modified trying to explain that

arnaudsjs:
Indeed. I completely missed the addition of the fixture.

arnaudsjs (approved):

PR #10358, opened 2026-05-08
Added support to get all log files

arnaudsjs:
# Description

Added support to get all the log files used by the server and scheduler processes.

Part of inmanta/inmanta-support#989

# Self Check:

- [x] Changelog entry


jptrindade (approved):

src/inmanta/logging.py
```diff
@@ -691,6 +733,50 @@ def __init__(self, stream: TextIO = sys.stdout, no_install: bool = False) -> Non
...
+        Returns a list of all the log files used by the server and scheduler.
+
+        :param env_ids: Include the scheduler logs files for these environments in the result.
+        """
+        assert self._loaded_config is not None
+        if self._component != "server":
+            raise Exception("get_all_log_files() is only supported on the server component")
+        # Server log files
+        result: set[str] = self._loaded_config.get_all_log_files()
+        result.add(policy_engine.PolicyEngine.get_path_policy_engine_log_file())
```
sanderr, line 746: This one doesn't contain sensitive data?

src/inmanta/logging.py
```diff
@@ -159,6 +161,22 @@ def ensure_log_dirs(self) -> None:
         for directory in self.log_dirs_to_create:
             os.makedirs(directory, exist_ok=True)
 
+    def get_all_log_files(self) -> set[str]:
```
sanderr, line 164: Do I understand correctly that this is simply a best-effort implementation? i.e. if the user adds a custom handler that writes to a file we don't support that?

Or is this the legacy logging config? I don't recall what "dict-based" means exactly.

src/inmanta/logging.py
```diff
@@ -264,6 +282,24 @@ def _to_dict_config(self) -> dict[str, object]:
     def to_string(self) -> str:
         return yaml.dump(self._to_dict_config(), Dumper=LogConfigDumper)
 
+    @classmethod
+    def from_dict(self, dict_config: dict[str, object]) -> "FullLoggingConfig":
+
+        def as_dict(inp: dict[str, object], key: str) -> dict[str, object]:
+            root = inp.get(key, {})
+            if not isinstance(root, dict):
+                raise Exception(f"{key} entry should be a dict, got {root}")
```
sanderr:
```suggestion
                raise Exception(f"{key!r} entry should be a dict, got {root!r}")
```
I think?

src/inmanta/logging.py
```diff
@@ -691,6 +733,50 @@ def __init__(self, stream: TextIO = sys.stdout, no_install: bool = False) -> Non
 
         self.logging_config_source: LoggingConfigSource | None = None
 
+    def get_all_log_files(self, env_ids: Sequence[uuid.UUID]) -> set[str]:
+        """
+        Returns a list of all the log files used by the server and scheduler.
+
+        :param env_ids: Include the scheduler logs files for these environments in the result.
+        """
+        assert self._loaded_config is not None
```
sanderr: No longer needed with your annotation update?

mypy-baseline.txt
sanderr, line 1: I assume these are unavoidable?

sanderr (changes requested):
This is a bit out of my comfort zone, but I think it looks good.

src/inmanta/logging.py
arnaudsjs: Good point. It might so it's not ideal. But if the support request is related to the policy engine, we will need access to it in order to investigate.

src/inmanta/logging.py
arnaudsjs: Correct. It assumes that no customer handler is used, but I think that's acceptable. As far as I can see, the only way to use a custom handler is to write a custom extension and I am not even sure if we (want to) provide support on that.

src/inmanta/logging.py
arnaudsjs: Why?

mypy-baseline.txt
arnaudsjs: Correct.

src/inmanta/logging.py
sanderr: > As far as I can see, the only way to use a custom handler is to write a custom extension

I was thinking you could add any handler to the logging config file. But your reasoning is that it could only reference a handler that is loaded by the server, which requires a custom import, therefore a custom extension. Is that right?

If so, we will have to be careful to extend this if Python adds more handlers.

src/inmanta/logging.py
sanderr: For `key` it's minor, simply to get it quoted. For printing objects of unknown type I find the unambiguous format of `repr` to be (typically) more useful than the freeform `str`.

e.g.
```
>>> print(f"Expected int but got {'1'}")
Expected int but got 1
>>> print(f"Expected int but got {'1'!r}")
Expected int but got '1'
```

To be fair, it may be slightly less relevant for dicts. It would require a non-dict mapping type that implements `str()` like a dict, for this confusion to surface. But even apart from that confusion, `repr` is by definition the unambiguous of the two.

src/inmanta/logging.py
arnaudsjs: > But your reasoning is that it could only reference a handler that is loaded by the server, which requires a custom import, therefore a custom extension. Is that right?

The loading part is not the issue. It's just that you need to define the handler (its code) and make it accessible by the server. You could achieve that by doing a pip install into the venv of the server, but that's very unconventional.

sanderr (approved):

PR #10370, opened 2026-05-11
Log more information about the DB during the server startup. (PR #10267)

Hugo-Inmanta:
This PR was initially a cherry-pick of https://github.com/inmanta/inmanta-core/pull/10267 for iso8 but there were merge conflicts into iso8 because the original PR builds on top of some framework code that was iso9 only (namely the `PostgreSQLVersion` class). 

I figured this additional logging would still be nice to have for iso8 so I adapted the initial PR as such:

- The "framework PR" that added `PostgreSQLVersion` class didn't make it into iso8 because it adds the following breaking behaviour: if we detect that the db version is lower than the version defined in the compatibility file, we will abort the server start.

- The current PR brings the necessary parts of the `PostgreSQLVersion` from the ["framework PR"](https://github.com/inmanta/inmanta-core/pull/9589) and will only log the db version (and ignore the compatibility file)

For reference:
* [The original PR being cherry-picked](https://github.com/inmanta/inmanta-core/pull/10267)
* [The "framework PR" that added `PostgreSQLVersion` class into iso9](https://github.com/inmanta/inmanta-core/pull/9589) 
* [slack discussion](https://inmanta.slack.com/archives/CKRF0C8R3/p1758783808339199) about the framework PR being iso9 only

- [ ] Changelog entry

# Description

* Short description here *

# Self Check:

- [ ] Changelog entry


src/inmanta/server/services/databaseservice.py
```diff
@@ -295,4 +330,181 @@ async def initialize_database_connection_pool(
...
+                    )
+        else:
+            LOGGER.info(
+                "Database replication is not active: couldn't find any standby server directly connected to the primary. "
+                "If you intend to use database replication, please check the status and the configuration "
+                "of the cluster before restarting the Inmanta server (More info in the 'HA setup' section of the "
+                "documentation)."
+            )
+
+        LOGGER.info("Checking database before server start...")
```
arnaudsjs: Why do we have this log line?

src/inmanta/server/services/databaseservice.py
```diff
@@ -295,4 +330,181 @@ async def initialize_database_connection_pool(
...
+                    opt.db_port.get(),
+                )
+                return conn
+            except TimeoutError:
+                LOGGER.info("Waiting for database to be up: Connection attempt timed out.")
+            except Exception:
+                LOGGER.info("Waiting for database to be up.", exc_info=True)
+            # Check if the maximum wait time has been exceeded
+            if 0 < db_wait_time < asyncio.get_event_loop().time() - start_time:
+                LOGGER.error("Timeout: database server not up after %d seconds." % db_wait_time)
```
arnaudsjs:
```suggestion
            if 0 < db_wait_time < asyncio.get_event_loop().time() - start_time:
                LOGGER.error("Timeout: database server not up after %d seconds.", db_wait_time)
```

tests/test_server.py
```diff
@@ -41,13 +42,19 @@
...
+from inmanta.server.services.databaseservice import PostgreSQLVersion
 from inmanta.types import ResourceIdStr, ResourceVersionIdStr
 from inmanta.util import get_compiler_version
 from utils import log_contains, log_doesnt_contain, retry_limited
 
 LOGGER = logging.getLogger(__name__)
 
 
+@pytest.fixture
+async def postgresql_version_from_db(postgresql_client) -> Awaitable[PostgreSQLVersion]:
```
arnaudsjs: This method doesn't return an awaitable, right?

arnaudsjs (changes requested):

src/inmanta/server/services/databaseservice.py
Hugo-Inmanta: Crap, good catch I'll remove it from the original PR as well.

jptrindade (approved):

tests/test_server.py
Hugo-Inmanta: Nice catch

arnaudsjs (approved):

PR #10386, opened 2026-05-15
Introduce a field to know when a resource was last seen before being orphaned

### LINKED ISSUE #9663: Introduce a field in the RPS table to know the model version when a resource was last seen.

jptrindade:
# Description

  Replace ResourcePersistentState.is_orphan with orphaned_after.
  This field now contains the latest version where this resource is still present in the model, if it was orphaned.
  Set to None if resource is present in the latest version of the model.
  
  (Reviewed by Claude)

closes #9663

# Self Check:

- [x] Changelog entry


src/inmanta/data/sqlalchemy.py
jptrindade, line 1: Missing the other index

src/inmanta/data/sqlalchemy.py
```diff
@@ -788,13 +795,28 @@ class ResourcePersistentState(Base):
...
+    @hybrid_property
+    def is_orphan(self) -> bool:
+        """
+        Is this resource not present in the latest released version of the model?
+        """
+        return self.orphaned_after is not None
+
+    @is_orphan.inplace.expression
+    @classmethod
+    def _is_orphan_expression(cls) -> Case[Any]:
```
Hugo-Inmanta, line 808: What does this do?

Hugo-Inmanta (approved):
One small question, approving already.

src/inmanta/data/sqlalchemy.py
jptrindade: This makes the `is_orphan` filter available on this sqlalchemy class. Meaning that instead of filtering on `orphaned_after` we can just shortcut it.

But now that you mention it. I just realized I created this but am not currently using it

jptrindade: asked for re-review, for you to see how it should work.

Hugo-Inmanta (approved):

src/inmanta/data/__init__.py
```diff
@@ -4774,7 +4774,7 @@ def table_name(cls) -> str:
     # Written at version release time
     is_undefined: bool
     # Written when a new version is processed by the scheduler
-    is_orphan: bool
+    orphaned_after: int | None = None
```
sanderr, line 4777: Now that I saw the name we use in the scheduler (`last_seen_version`) I wonder if that might be clearer to use here as well? We could get rid of the "orphan" name entirely. Not a requirement for me, and even if you agree, probably something to pull wider than just the two of us.

src/inmanta/graphql/schema.py
```diff
@@ -1003,34 +1003,9 @@ async def resources(
...
-                            )
-                            .where(
-                                models.Resource.environment == models.ResourcePersistentState.environment,
-                                models.Resource.resource_id == models.ResourcePersistentState.resource_id,
-                                models.Configurationmodel.released.is_(True),
-                            )
-                            .scalar_subquery(),
+                        func.coalesce(
+                            models.ResourcePersistentState.orphaned_after,
+                            latest_scheduled_version_cte.c.version,
```
sanderr, line 1008: This is a tough one: if it's not an orphan, it may be in the latest scheduled version, or any future version.

src/inmanta/data/sqlalchemy.py
```diff
@@ -788,13 +795,28 @@ class ResourcePersistentState(Base):
 
     environment_: Mapped["Environment"] = relationship("Environment", back_populates="resource_persistent_state")
 
+    @hybrid_property
+    def is_orphan(self) -> bool:
+        """
+        Is this resource not present in the latest released version of the model?
```
sanderr, line 801: Kept for backwards compatibility?

src/inmanta/data/sqlalchemy.py
```diff
@@ -788,13 +795,28 @@ class ResourcePersistentState(Base):
...
+        """
+        return self.orphaned_after is not None
+
+    @is_orphan.inplace.expression
+    @classmethod
```
sanderr, line 807: I'm not familiar with this part of `sqlalchemy` but it seems strange to me that we need both this and `is_orphan` as a hybrid property. What am I missing?

src/inmanta/graphql/schema.py
```diff
@@ -1003,34 +1003,9 @@ async def resources(
...
-                                    models.t_resource_set_configuration_model.c.model == models.Configurationmodel.version,
-                                ),
-                            )
-                            .where(
-                                models.Resource.environment == models.ResourcePersistentState.environment,
-                                models.Resource.resource_id == models.ResourcePersistentState.resource_id,
-                                models.Configurationmodel.released.is_(True),
-                            )
-                            .scalar_subquery(),
+                        func.coalesce(
```
sanderr, line 1006: This became a lot simpler. I would expect something similar in `dataview`. Am I missing something?

sanderr (commented):
I'll review the db migration and tests tomorrow.

src/inmanta/data/__init__.py
jptrindade: I believe the argument against `last_seen_version` is that it is not clear that it should be `None` when a resource is present in the latest desired state. We discussed this in this thread in [slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1779091591233649) and in person in the team week

src/inmanta/graphql/schema.py
jptrindade: Yes, I forgot to edit it there. On it

src/inmanta/graphql/schema.py
jptrindade: Is this the case for unreleased versions? How should we handle that case?

src/inmanta/data/sqlalchemy.py
jptrindade: > This makes the is_orphan filter available on this sqlalchemy class. Meaning that instead of filtering on orphaned_after we can just shortcut it.

src/inmanta/data/sqlalchemy.py
jptrindade: Yes, and it makes it easier to apply the filter, but no strong feeling towards it in either case

src/inmanta/graphql/schema.py
sanderr: Only released versions I think. But it may be a version that has not yet been processed by the scheduler. So the question really becomes: what does this view represent? What do the fields that we return reflect? Are they the attributes for the currently scheduled version or the latest released version or a hybrid?

src/inmanta/data/sqlalchemy.py
sanderr: Sorry, I meant to suggest adding "kept for backwards compatibility" or something like it to the docstring.

> Yes, and it makes it easier to apply the filter, but no strong feeling towards it in either case

We may want to update the filter in the future to better reflect the new field. But that's a different matter.

src/inmanta/data/sqlalchemy.py
sanderr: I mean that we need both `def is_orphan` and `def _is_orphan_expression`. But I see that the pattern occurs elsewhere as well, so I'm guessing it's the expected sqlalchemy pattern, I'm just not familiar with it.

src/inmanta/db/versions/v202605150.py
```diff
@@ -0,0 +1,52 @@
...
+        INNER JOIN public.configurationmodel AS m
+            ON rscm.environment=m.environment AND rscm.model=m.version
+        WHERE r.environment=rps.environment AND r.resource_id=rps.resource_id AND m.released
+    ) WHERE rps.is_orphan;
+
+    CREATE INDEX resource_persistent_state_environment_orphaned_after_index ON public.resource_persistent_state
+        USING btree (environment) WHERE orphaned_after IS NULL;
+
+    CREATE INDEX resource_persistent_state_environment_resource_id_orphaned_after_index ON public.resource_persistent_state
+        USING btree (environment, resource_id, orphaned_after);
```
sanderr, line 45: Do you know what these indexes are used for, and that we cover those use cases? The first index looks slightly weird to me. And for the second, I think we might want an explicit `NULLS FIRST` / `NULLS LAST`.

jptrindade: The first is used when we do `environment=$1 AND orphaned_after IS NULL`. This appears on:
- `get_compliance_report`
- `get_latest_resource_states`
- `get_current_resource_state`
- `get_composed_resource_summary`
- `get_resource_deploy_summary`
- `get_metric_value`
The second one seems to be only in use for:
- `get_current_resource_state`
But that one also filters on nulls, so I could see the use case in adding `NULLS FIRST` It is also possible that the graphql endpoints also use this index

src/inmanta/graphql/schema.py
jptrindade: As a general rule of thumb I think we should move from the MAX version of the configuration model to the last version processed by the scheduler. In my head, the scheduler is in charge of the rps table, and most information we show regarding a resource is from the rps table, so the scheduler version should be the source of truth. But I think we have to be consistent in this aspect, the dataview is currently using the MAX configuration model

src/inmanta/data/dataview.py
jptrindade, line 1: @sanderr I'm mostly curious on what do you think of this change. I'm not sure how this will impact performance when we filter out orphans (I guess that is our main use case), but it is a lot cleaner code-wise

src/inmanta/data/dataview.py
```diff
@@ -539,60 +516,17 @@ def get_extra_url_parameters(self) -> dict[str, str]:
...
-                )
-        """
-        else:
-            prelude = f"""
+        prelude: str = f"""
                 WITH latest_version AS (
                     SELECT MAX(public.configurationmodel.version) as version
                     FROM public.configurationmodel
                     WHERE public.configurationmodel.released AND environment=$1
                 ), versioned_resource_state AS (
```
sanderr: I think we can clean this up a bit more by getting rid of this CTE now (and to move closer to the join format we had in the exclude-orphans case above) because as far as I can tell it doesn't serve a purpose anymore.

sanderr (changes requested):
One small suggestion, but the overall change to the resource list query looks good to me.

src/inmanta/data/dataview.py
sanderr: Regarding performance, one thing to check is if we are messing up the `COUNT` time with this change.

src/inmanta/db/versions/v202605150.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Add the orphaned_after field to the rps table, populate it and drop is_orphan
+    """
+    schema = """
+     ALTER TABLE public.resource_persistent_state ADD COLUMN orphaned_after integer;
+
+     UPDATE public.resource_persistent_state AS rps
+     SET orphaned_after=(
+        -- Safeguard for orphaned resources that are not present on the rscm table
+        SELECT COALESCE(MAX(rscm.model), 0)
```
sanderr, line 32: This may have repercussions for the read side. We should probably have a test for it. Omitting them in the API is probably safe, but we shouldn't crash on it.

src/inmanta/db/versions/v202605150.py
sanderr: 1. I wonder if the second one should also be partial. I'm pretty sure that the graphql and resource list use this index.
2. About `NULLS FIRST/LAST`, that's mainly relevant when sorting comes into play. If you have an `ORDER BY` that should use an index, its `NULLS FIRST/LAST` should be consistent with the index', otherwise it can't use it. But perhaps we don't really have a sorting use case.

src/inmanta/db/versions/v202605150.py
jptrindade: This should only happen with direct database manipulation. How do you think we should approach this?

src/inmanta/data/dataview.py
jptrindade: I'm guessing that should show itself on the performance tests

src/inmanta/db/versions/v202605150.py
jptrindade: I believe resource_list sorts by default on `resource_type` so I'm not sure it will use the index as is. `graphql` seems to sort on `resource_id` by default but on the frontend we seem to sort on resource_type by default.

src/inmanta/data/dataview.py
sanderr: Fair enough. Just remember to check it after merging.

src/inmanta/data/dataview.py
```diff
@@ -540,13 +533,14 @@ def get_base_query(self) -> SimpleQueryBuilder:
                         rps.environment,
                         {const.SQL_RESOURCE_STATUS_SELECTOR} AS status
                     FROM resource AS r
+                    CROSS JOIN latest_version AS lv
                     INNER JOIN resource_set_configuration_model AS rscm
                         ON r.environment=rscm.environment
                         AND r.resource_set=rscm.resource_set
-                    INNER JOIN versioned_resource_state AS rps
+                    INNER JOIN resource_persistent_state AS rps
```
sanderr, line 537: (nitpick, feel free to ignore) I think we typically put this one first, but that's truly a nitpick, it shouldn't have any impact.

src/inmanta/db/versions/v202605150.py
sanderr: I mainly think that we should either
- decide that we're confident that it should really never happen, and then we don't have this fallback
- decide that we need this fallback just in case and make sure the read part survives it

I just think it doesn't make sense to introduce a fallback scenario to then break that scenario elsewhere. Either we have a fallback across the board, or we don't.

sanderr: (I'm trying to think how the read part will behave. Then I'll give some thought to how confident we can be that this will never happen.)

sanderr: I'm pretty sure that the inner joins will just result in it being excluded. I'll approve now, but it would probably be a good idea if you can confirm.

sanderr (approved):

src/inmanta/db/versions/v202605150.py
jptrindade: Yes, I believe they will be simply excluded.

PR #10404, opened 2026-05-22
Make schedulers ignore config options set using environment variables

arnaudsjs:
# Description

This PR fixes the bug where environment variables, set on the server, are taken into account by schedulers started by that server. The sever generates a configuration file for each scheduler and only the configuration options in that file should be taken into account. Any configuration option set using an environment variable must be ignored.

The concrete issue we faced is that the Inmanta docker container sets the state_dir directory as a environment variable by default. Hence, all schedulers use the same state directory and not the state directory configured in the scheduler.cfg file. As such, all schedulers would store their venv directories in the same location. This caused race conditions where one scheduler would cleanup a venv that was created by another process.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/bugfix-all-schedulers-use-same-state-dir.yml
+++ b/changelogs/unreleased/bugfix-all-schedulers-use-same-state-dir.yml
@@ -1,0 +1,6 @@
+---
+description: 'Fixed the bug where all the schedulers of the different environments used the same state directory. This bug can cause a "no such file or directory" error on the .inmanta_venv_status file of an agent which prevents resources from being deployed.'
+change-type: patch
+destination-branches: [master, iso9, iso8]
+sections:
+  bugfix: "{{description}}"
--- a/src/inmanta/config.py
+++ b/src/inmanta/config.py
@@ -228,8 +228,8 @@ class Config:
         cls.__instance = None
         cls._config_dir = None
         cls._min_c_config_file = None
-        cls._config_updated()
         cls._ignore_env_vars = False
+        cls._config_updated()
 
     @classmethod
     def _config_updated(cls) -> None:
--- a/src/inmanta/config.py
+++ b/src/inmanta/config.py
@@ -169,8 +169,8 @@ class Config:
         cls.__instance = config
         cls._config_dir = config_dir
         cls._min_c_config_file = min_c_config_file
-        cls._config_updated()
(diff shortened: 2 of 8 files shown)


Hugo-Inmanta (approved):

src/inmanta/config.py
```diff
@@ -91,20 +91,32 @@ class Config:
...
     _min_c_config_file: Optional[str] = None  # Config file
     __config_definition: dict[str, dict[str, "Option"]] = defaultdict(dict)
+    _ignore_env_vars: bool = False
 
     @classmethod
     def get_config_options(cls) -> dict[str, dict[str, "Option"]]:
         return cls.__config_definition
 
+    @classmethod
+    def is_ignoring_env_vars(cls) -> bool:
```
sanderr, line 101: `is_ignoring` implies a transient behavior. I think it's important for consistency that this is never toggled, i.e. only set at load time. I think the name should reflect that.

changelogs/unreleased/bugfix-all-schedulers-use-same-state-dir.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: 'Fixed the bug where all the schedulers of the different environments used the same state directory. This bug can cause a "no such file or directory" error on the .inmanta_venv_status file of an agent which prevents resources from being deployed.'
```
sanderr, line 2:
```suggestion
description: 'Fixed the bug where all the agents of the different environments used the same state directory. This bug can cause a "no such file or directory" error on the .inmanta_venv_status file of an agent which prevents resources from being deployed.'
```

sanderr (changes requested):
After some more consideration, I think the current approach makes sense. But I do think that we should propagate *all* config, with the relevant overrides. Otherwise I worry we would add too much coupling: if we add more options that are relevant to the agent (or possibly we even have such already), they would have to be added explicitly to the config file.

sanderr:
A summary of the requirements and proposed fix discussed in the review has been posted in [#development on Slack](https://inmanta.slack.com/archives/CKRF0C8R3/p1779895116352529).

arnaudsjs:
@sanderr I can close this PR?

sanderr:
I think so, unless you think there is something in here that is still relevant that I didn't port to my PR?

Issue #10351, opened 2026-05-05
Scheduler stops receiving new version notifications after DB disconnect due to stale `tid_endpoint_to_session` entry


sanderr:
*This ticket was created by Claude under my supervision. I did not verify all details in the analysis, but I don't immediately see any flaws in it.*

---

## Summary

After a Postgres disconnect (e.g. server reboot), the scheduler can end up in a state where the server keeps routing reverse-RPC calls (`trigger_read_version`) to an **old, deleted session object** while the agent's heartbeat long-poll is waiting on a **new session object**. The call is never delivered, times out, and new versions are silently skipped until the scheduler process itself restarts.

This was observed in the mgmt orchestrator incident on 2026-05-04: 9 versions (1572569696–1572569705) went unprocessed for ~34 minutes.

---

## Root Cause

The bug is in `AgentManager._expire_session()` in `src/inmanta/server/agentmanager.py`.

### Sequence of events

1. **Postgres connection lost** (~05:47 UTC). The server process stays alive; the scheduler subprocess stays alive.

2. **Session `76c786ac-...` expires** (~05:47:11) — no successful heartbeats for ~30 s. `AgentManager._expire_session()` runs:
   ```python
   del self.sessions[sid]            # ✓ succeeds
   del self.endpoints_for_sid[sid]   # ✓ succeeds
   endpoints_with_new_primary = await self._failover_endpoints(session, endpoint_names_snapshot)
   # ↑ first line of _failover_endpoints is:
   #   agent_statuses = await data.Agent.get_statuses(session.tid, endpoints)
   # DB is down → EXCEPTION raised here
   ```
   The exception is caught by `_process_session_listener_actions`, logged as:
   ```
   ERROR inmanta.server.agentmanager An exception occurred while handling session action EXPIRE_SESSION on session id 76c786ac-...
   ```
   **`tid_endpoint_to_session[(tid, AGENT_SCHEDULER_ID)]` is never cleaned up** — it keeps pointing to the now-deleted old Session Python object (with its old `tornado.queues.Queue`).

3. **Agent reconnects at 05:47:34** with the same SID `76c786ac-...`. `SessionManager.get_or_create_session()` finds no entry in `_sessions` (correctly deleted in step 2) and creates a **new** Session object with a **fresh queue**.

4. **`_register_session` is called** for the new session. It calls `_ensure_primary_if_not_exists()`:
   ```python
[...]

PR #10415, opened 2026-05-27
Fix bug in stale session cleanup

Hugo-Inmanta:
# Description

Iso8 + 9 only since the websocket refactor on iso10 removed most of the session management logic.

closes https://github.com/inmanta/inmanta-core/issues/10351

# Self Check:

- [ ] Changelog entry


src/inmanta/server/agentmanager.py
```diff
@@ -645,7 +656,8 @@ async def _ensure_primary_if_not_exists(
         result = []
         for endpoint in endpoints:
             key = (session.tid, endpoint)
-            if key not in self.tid_endpoint_to_session and agent_statuses[endpoint] != AgentStatus.paused:
+            existing = self.tid_endpoint_to_session.get(key)
+            if (existing is None or existing.id not in self.sessions) and agent_statuses[endpoint] != AgentStatus.paused:
```
jptrindade, line 661: Should this be `up` instead? Like this it will trigger for `down`ed agents.

```suggestion
            if (existing is None or existing.id not in self.sessions) and agent_statuses[endpoint] == AgentStatus.up:
```

Hugo-Inmanta: I think this is intended. That way downed agents (i.e. [those who don't have a primary](https://github.com/inmanta/inmanta-core/blob/3c447108d330ada6d8240b25993b1226bcfffc74/src/inmanta/data/__init__.py#L3595-L3600)) can recover and use this new primary.
(cc @sanderr correct ?)

changelogs/unreleased/10351-stale-session-cleanup-bugfix.yml
```diff
@@ -0,0 +1,10 @@
+description: >
+  Fix bug in the logic to cleanup stale sessions. A loss of connectivity towards the database could
+  sometimes cause the scheduler to miss new versions notifications.
+issue-nr: 10351
+change-type: patch
+destination-branches:
+  - iso8
```
sanderr, line 7: Reminder to add iso9 as well

sanderr (commented):
@arnaudsjs I'm shifting this to you because I'm trying to catch up on other review work, and I believe you know this code better than I do in any case. If I'm wrong, feel free to pass it back.

src/inmanta/server/agentmanager.py
```diff
@@ -573,7 +573,18 @@ async def _expire_session(self, session: protocol.Session, endpoint_names_snapsh
...
+            try:
+                endpoints_with_new_primary = await self._failover_endpoints(session, endpoint_names_snapshot)
+            except Exception:
+                # _failover_endpoints issues a DB query as its first operation. When the DB is
+                # unavailable it raises before touching tid_endpoint_to_session, leaving a stale
+                # reference to the now-deleted session. Clean it up so that a reconnecting session
+                # can be correctly promoted to primary.
+                for endpoint_name in endpoint_names_snapshot:
+                    key = (session.tid, endpoint_name)
+                    if key in self.tid_endpoint_to_session and self.tid_endpoint_to_session[key].id == session.id:
```
arnaudsjs: The first part of this method uses `get_id()` to get the id of a session. The new part uses the `id` property. It would improve readability if we use one style within the same method.

```suggestion
                    if key in self.tid_endpoint_to_session and self.tid_endpoint_to_session[key].get_id() == sid:
```

tests/test_agent_manager.py
```diff
@@ -764,6 +764,49 @@ async def test_session_creation_fails(server, environment, async_finalizer, capl
     assert len(session_manager._sessions) == 0
 
 
+@pytest.mark.parametrize("auto_start_agent", [False])  # prevent autostart to keep agent under control
+async def test_session_expiration(server, environment, async_finalizer, caplog):
```
arnaudsjs: This test case is missing the part that verifies what happens if the database comes back online. Can the agent reconnect without any issue.

src/inmanta/server/agentmanager.py
arnaudsjs: That's correct. This method does a failover, so we can failover to any agent (whether it's a primary or not) as long as the agent is not paused.

arnaudsjs (changes requested):

Hugo-Inmanta:
~Superseded by https://github.com/inmanta/inmanta-core/pull/10427~

jptrindade (approved):

arnaudsjs (approved):

PR #10420, opened 2026-05-29
Issue/fix all schedulers use same state dir alt

sanderr:
# Description

Follow-up PR of #10404, proposing a different solution that covers all use cases.

# Background

The server spawns a scheduler subprocess for each environment and generates a dedicated `scheduler.cfg` for it. Two requirements must hold:

  1. A few options must be explicitly controlled by the server — state-dir, environment, transport settings, auth token. These are written into scheduler.cfg
  with environment-specific values.
  2. All other config options should be inherited from the server one way or the other (e.g. agent_ws_ping_interval, executor_venv_retention_time).

# The bug

  The inmanta container sets `INMANTA_CONFIG_STATE_DIR=/var/lib/inmanta` as an environment variable. Because the scheduler is a subprocess of the server,
  it inherits this env var. Since environment variables take higher priority than config file values, INMANTA_CONFIG_STATE_DIR was overriding the
  per-environment state directory written into scheduler.cfg, causing all schedulers to share the same state directory and race on venv files.

#10404 attempted to fix this by ignoring all config from env vars in the scheduler process. However, this could lead to other relevant config being ignored. Notably, in #10404, `agent_ws_ping_interval` and some others would be ignored if configured through env variables.

# This PR

The approach proposed in this PR is simple:
1. for the options that are explicitly controlled by the server, unset the associated environment variables for the scheduler subprocess
2. let all other options be inherited naturally, either through the environment, or by the scheduler process reading the server's main config files (already in place).

There is however one caveat (#10421): if the server is started with an explicit `--config`, it might contain agent-relevant settings that are then not inherited by the scheduler process. This PR works around it, but the proper fix was deemed out of scope. For the workaround, we add a third step:

3. additionally, hardcode the list of all options in `inmanta.agent.config`, and explicitly control them from the server as well, setting them to the value as read by the server config. The previous implementation seemed to have something similar in place, but it was not kept up to date, which highlights the importance of the general approach proposed in this PR. This workaround can be removed once #10421 has been resolved, since it is only required in the very rare case where the server is started with a custom `--config`.

# Implementation

The best place to start is probably `inmanta.server.agentmanager._get_agent_config`.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/bugfix-all-schedulers-use-same-state-dir.yml
+++ b/changelogs/unreleased/bugfix-all-schedulers-use-same-state-dir.yml
@@ -1,0 +1,6 @@
+---
+description: 'Fixed the bug where all the agents of the different environments used the same state directory. This bug can cause a "no such file or directory" error on the .inmanta_venv_status file of an agent which prevents resources from being deployed.'
+change-type: patch
+destination-branches: [master, iso9, iso8]
+sections:
+  bugfix: "{{description}}"
--- a/src/inmanta/config.py
+++ b/src/inmanta/config.py
@@ -579,7 +579,7 @@ def make_option_for_log_file(component_name: str) -> Option[str | None]:
 
 
 component_log_configs = {k: make_option_for_log_file(k) for k in ["server", "scheduler", "compiler"]}
-scheduler_log_config = component_log_configs["scheduler"]
+scheduler_log_config: Option[str | None] = component_log_configs["scheduler"]
 
 
 def get_executable() -> Optional[str]:
--- a/src/inmanta/config.py
+++ b/src/inmanta/config.py
@@ -241,16 +241,21 @@ class Config:
         name = normalize_name(name)
 
         option: Optional[Option[object]] = cls.validate_option_request(section, name, default_value)
-        return cls.get_for_option(option) if option is not None else cls._get_value(section, name, default_value)
+        return cls.get_for_option(option) if option is not None else cls.get_raw_value(section, name, default_value)
 
     @classmethod
     def get_for_option(cls, option: "Option[T]") -> T:
-        default_value = option.get_default_value()
-        raw_value: str | T = cls._get_value(option.section, option.name, default_value)
-        return option.validate(raw_value)
+        """
+        Returns the parsed and validated option. Unlike Option.get(), does not fall back to deprecated predecessor option.
+        """
+        return option.parse(cls.get_raw_value(option.section, option.name, None))
 
     @classmethod
-    def _get_value(cls, section: str, name: str, default_value: T) -> str | T:
+    def get_raw_value(cls, section: str, name: str, default_value: T) -> str | T:
+        """
+        Returns the option as the raw string as defined in the config file / env var, or default_value if it is not set.
+        To get the properly typed option value, with default value filled in, see get_for_option().
+        """
         cfg: ConfigParser = cls.get_instance()
         val: Optional[str] = _get_from_env(section, name)
         if val is not None:
@@ -435,7 +440,7 @@ class Option(Generic[T]):
     :param documentation: the documentation for this option
     :param validator: a function responsible for turning the string representation of the option into the correct type.
         Its docstring is used as representation for the type of the option.
-    :param predecessor_option: The Option that was deprecated in favour of this option.
+    :param predecessor_option: The Option that was deprecated in favour of this option. Should not be nested.
     """
 
     def __init__(
@@ -461,18 +466,31 @@ class Option(Generic[T]):
         """
         return f"{self.section}.{self.name}"
 
-    def get(self) -> T:
+    def get_raw(self) -> str | None:
+        """
+        Returns this option as the raw string as defined in the config file / env var, or None if it is not set.
+        """
         raw_config: ConfigParser = Config.get()
-        if self.predecessor_option:
-            has_deprecated_option = raw_config.has_option(self.predecessor_option.section, self.predecessor_option.name)
-            has_new_option = raw_config.has_option(self.section, self.name)
-            if has_deprecated_option and not has_new_option:
-                warnings.warn(
-                    f"Config option {self.predecessor_option.name} is deprecated. Use {self.name} instead.",
-                    category=DeprecationWarning,
-                )
-                return self.predecessor_option.get()
-        return Config.get_for_option(self)
+        result: str | None = Config.get_raw_value(self.section, self.name, None)
+        if result is not None or not self.predecessor_option:
+            return result
+        predecessor_result: str | None = self.predecessor_option.get_raw()
+        if predecessor_result is not None:
+            warnings.warn(
+                f"Config option {self.predecessor_option.name} is deprecated. Use {self.name} instead.",
+                category=DeprecationWarning,
+            )
+            return predecessor_result
+        return None
+
+    def parse(self, value: str | None) -> T:
+        """
+        Parses the raw string value by running it through validate, or returns default value if None.
+        """
+        return self.validate(value if value is not None else self.get_default_value())
+
+    def get(self) -> T:
+        return self.parse(self.get_raw())
 
     def get_type(self) -> Optional[str]:
         if callable(self.validator):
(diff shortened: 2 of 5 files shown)


changelogs/unreleased/bugfix-all-schedulers-use-same-state-dir.yml
sanderr, line 1: Copied from #10404 with one minor edit.

tests/test_agent_manager.py
```diff
@@ -622,14 +625,48 @@ async def test_process_already_terminated(server, environment):
     await autostarted_agent_manager._terminate_agents()
 
 
+@pytest.fixture
```
sanderr, line 628: This fixture and the test directly below it were copied from #10404 with minor edits.

tests/test_agent_manager.py
```diff
@@ -736,3 +773,55 @@ async def session_state():
     agent_dct = {agent.name: agent for agent in agents}
     assert not agent_dct[const.AGENT_SCHEDULER_ID].paused
     assert not agent_dct["agent1"].paused
+
+
+async def test_do_start_agent_clears_shadowed_settings_env_vars(server_config, monkeypatch, tmp_path):
```
sanderr, line 778: The tests from here on out are new to this PR.

src/inmanta/config.py
sanderr, line 1: I had to refactor this slightly to expose the raw config, because not all options are round-trip compatible.

src/inmanta/config.py
```diff
@@ -461,18 +466,31 @@ def get_full_name(self) -> str:
         """
         return f"{self.section}.{self.name}"
 
-    def get(self) -> T:
+    def get_raw(self) -> str | None:
+        """
+        Returns this option as the raw string as defined in the config file / env var, or None if it is not set.
+        """
         raw_config: ConfigParser = Config.get()
```
bartv, line 473: `raw_config` is now dead code — a leftover from the old `Option.get()`, where it fed `raw_config.has_option(...)`. The new body only uses `Config.get_raw_value(...)`, and the lazy config-load side effect is redundant (the next line calls `get_instance()` anyway). pyflakes flags it as F841 (repo flake8 tolerates it, like the pre-existing one in `is_time_or_cron`). Safe to drop:

```suggestion
```

tests/test_agent_manager.py
```diff
@@ -736,3 +773,55 @@ async def session_state():
...
+
+    agent_manager = ServerlessAgentManager()
+    env_mock = Mock()
+    env_mock.id = uuid4()
+
+    await agent_manager._AutostartedAgentManager__do_start_agent(env_mock)
+
+    # verify that overridden settings are properly trimmed from the environment
+    assert state_dir_env_var not in captured_env
+    assert environment_env_var not in captured_env
```
bartv, line 821: These assertions cover the **override** options. But `shadowed_config_env_vars` is built from the full dict (overrides **+** passthrough, `agentmanager.py:1003`), so a *passthrough* option's env var is stripped too — and that path isn't asserted anywhere. If that comprehension were ever narrowed to only the overrides dict, every passthrough env var would leak into the child and silently shadow what was written to `scheduler.cfg`, yet this test would still pass.

Consider adding the symmetric counterpart of the `logging_config` (inherited) assertion:

```python
ws_ping_env = inmanta.agent.config.agent_ws_ping_interval.get_environment_variable()
monkeypatch.setenv(ws_ping_env, "7")
# ... after __do_start_agent ...
assert ws_ping_env not in captured_env  # passthrough option's env var is also stripped
```

src/inmanta/server/agentmanager.py
```diff
@@ -1005,107 +1013,95 @@ async def __do_start_agent(
...
-""" % {
-            "env_id": environment_id,
-            "port": port,
-            "statedir": privatestatedir,
+        :param env: The environment for which to autostart agents
+        :return: A mapping of config options to their values
+        """
+        environment_id: str = str(env.id)
+
+        agent_config_overrides: dict[inmanta.config.Option[typing.Any], str | None] = {
```
bartv, line 1040: Minor: `typing.Any` here trips our "avoid `Any`, prefer `object`" convention. The return type is correctly `Option[object]`; `Any` is only needed because `Option` is invariant in `T`, so an `Option[str]` literal won't fit into a `dict[Option[object], …]`. That's a fair exception to the rule — but a one-line comment explaining *why* `Any` is unavoidable here would save the next reader from re-deriving the invariance argument. Same applies to `agent_config_passthrough` on line 1055.

bartv (commented):
👋 This is Claude, reviewing on behalf of Bart.

**Overall: looks good — approve modulo one trivial cleanup.** Replacing the hand-built config string with a server-controlled override/passthrough set + stripping those options' env vars from the scheduler subprocess is a more robust design than the original, and the root cause (inherited `INMANTA_CONFIG_STATE_DIR` outranking `scheduler.cfg`) is addressed directly.

The specific findings are in the 3 line comments below; this is just the overview.

### Verified locally
| Check | Result |
|-------|--------|
| `flake8` on `config.py` + `agentmanager.py` | ✅ clean |
| `mypy` (changed files) | ✅ clean; removed `abspath` baseline line correctly matches deleted code |
| `test_do_start_agent_clears_shadowed_settings_env_vars` | ✅ pass |
| `test_scheduler_ignores_env_var_state_dir` (real server + postgres) | ✅ pass — proves the fix end-to-end |
| Option `section`/`name` → old `scheduler.cfg` layout | ✅ exact match |
| `%`-char round-trip | ✅ writer and scheduler reader both use no-op `Interpolation()` |

### Nicely handled
- No-op `configparser.Interpolation()` when writing `scheduler.cfg` — needed so values containing `%` (passwords/tokens) don't break `read_dict`/`write`, and it matches how the main `Config` loads.
- The `Option.get()` → `get_raw()`/`parse()` split also fixes a latent precedence bug: master's predecessor logic used `has_option()`, which ignores env vars, so a new option set *only* via env var was wrongly overridden by a file-based deprecated predecessor. The env-first `get_raw_value` path corrects that.
- Capturing values as raw `str` preserves round-trip for `default != validate(str(default))` cases, and `None` overrides are correctly skipped — so the old `password=None` literal-string footgun is gone.

### Non-blocking heads-up (not posted inline; no change required)
- **Backport:** `iso8`/`iso9` still use `_make_agent_config` and lack `get_raw`/`parse`/`get_raw_value`, so the `agentmanager.py` fix hard-depends on the `config.py` refactor — both backports must carry it. Single-PR + `destination-branches` is the right convention; just worth a note for the merge-tool operator.
- **`_dummy_fork_inmanta`** (`test_agent_manager.py:669`): new `env=None` param is unannotated — cosmetic, test-only, not type-checked.
- **SSL:** new code writes `ssl_ca_cert_file` even when `ssl=false` (old code wrote it only with both key+cert). Harmless — it's ignored when SSL is off.
- **#10421 scope:** the scheduler's *default* db-pool size derives from `database.connection-pool-min/max-size`, which aren't in the passthrough list — only matters in the narrow `--config`-without-`--config-dir` case, i.e. exactly the acknowledged best-effort scope.

src/inmanta/config.py
sanderr: Surprising that my own Claude did not flag this.

tests/test_agent_manager.py
sanderr: After giving this some more thought I don't think this is all that relevant. The passthrough is only a workaround for #10421. The concern for #10421 isn't with shadowing (env vars should rightly take precedence), it's with ignoring an explicit `--config` file if we wouldn't pass the options through. The important part isn't that we clear the env vars (that's just for consistency), it's that we pass the options via `scheduler.cfg` in the first place.

Concretely, the following is irrelevant, precisely because if there is an env var, that's what we will write to `scheduler.cfg`. So if there would be shadowing, it's the same config.

> If that comprehension were ever narrowed to only the overrides dict, every passthrough env var would leak into the child and silently shadow what was written to scheduler.cfg, yet this test would still pass.

I'll see if I can easily add a test for the passthrough in `scheduler.cfg`.

arnaudsjs (approved):
I am wondering whether it wouldn't  be easier if the server would set the configuration that must not be overridden by the user as environment variables. Those have the highest precedence.

PR #10428, opened 2026-06-02
Stop isort and black fighting in make format

bartv:
&gt; 🤖 Opened by Claude on behalf of @bartv.

## Problem

`make format` runs `isort` then `black`, and on this repo the two tools **oscillate forever** on imports that have a comment between groups. Concretely, `src/inmanta/agent/__init__.py`:

```python
from inmanta.agent.agent_new import Agent
                                      <- isort deletes this blank line, black re-inserts it
# flake8: noqa: F401
# Backward compatibility
from inmanta.agent.reporting import collect_report
```

- **isort** strips the blank line before the comment-prefixed import group.
- **black** insists on a blank line there.

Every `make format` run flips that one line, so the working tree never converges: each run shows a diff and there is no stable fixed point.

## Root cause

The `[isort]` section in `setup.cfg` hand-replicates the Black-compatible profile (`multi_line_output=3`, `include_trailing_comma`, `use_parentheses`, `force_grid_wrap=0`, `line_length=128`) but was missing one option: **`ensure_newline_before_comments`**. That is the exact setting in isort's `black` profile that tells isort to keep the blank line before a comment that precedes an import, matching Black.

## Fix

Drop the hand-maintained settings and use isort's `black` profile directly (suggested by @sanderr in review):

```ini
[isort]
profile=black
split_on_trailing_comma=False
line_length=128
...
```

The profile includes `ensure_newline_before_comments` (fixing the oscillation) plus any black-compatibility settings isort adds in the future. Two explicit overrides keep current behavior:

- `line_length=128` overrides the profile's default of 88.
- `split_on_trailing_comma=False` overrides the profile's `True`, which would otherwise explode existing compact imports into the multi-line "magic trailing comma" form (a style change touching several files).

Confirmed via `isort --show-config` that the explicit settings win over the profile layer, and the effective config is exactly the previous hand-maintained config plus `ensure_newline_before_comments`.

Verified: a full `isort` + `black` run over `src`, `tests` and `tests_common` leaves the working tree clean (zero reformatting), and two consecutive `make format` runs converge.

## Changelog

Added `changelogs/unreleased/fix-isort-black-format-fight.yml` (patch; master + iso9 + iso8).


Hugo-Inmanta (approved):

sanderr:
🤖 Claude suggestion (on behalf of @sanderr):

The `split_on_trailing_comma` drawback can be neutralized by explicitly overriding that one setting — the PR itself confirms this mechanism works, since `line_length=128` already overrides the profile's default of 88.

```ini
[isort]
profile = black
split_on_trailing_comma = False
line_length = 128
default_section = FIRSTPARTY
known_first_party = inmanta,inmanta_ext,inmanta_plugins
known_third_party = ...
skip = tests/data
```

This gives us the canonical profile (including `ensure_newline_before_comments` and any future black-compatibility settings) without the style churn. Worth considering over the one-liner fix.

bartv:
🤖 Claude on behalf of @bartv:

Good suggestion, adopted in e84d63b79. Verified before pushing:

- `isort --show-config` confirms the effective config is exactly the previous hand-maintained settings plus `ensure_newline_before_comments`: the explicit `split_on_trailing_comma=False` and `line_length=128` override the profile layer as expected.
- A full `isort` + `black` run over `src`, `tests` and `tests_common` with the new config produces zero changes, so no style churn, and repeated `make format` runs converge.

PR description updated to match.

PR #10434, opened 2026-06-04
Disallow commas on agent name

### LINKED ISSUE #10342: export fails on agent names with commas
Link to [slack thread](https://inmanta.slack.com/archives/CKRF0C8R3/p1777357280308659)

# Goal
1. Fix round-trip compatibility:
```python
from inmanta.resources import Id


i = Id("std::myentity", "foo,bar", "attribute", "value")
print(i.resource_str())
round_trip = Id.parse_id(i.resource_str())

assert round_trip.entity_type == i.entity_type
assert round_trip.agent_name == i.agent_name
assert round_trip.attribute == i.attribute
assert round_trip.attribute_value == i.attribute_value
```
2. Check if the issue can still be reproduced. Hypothesis: fixing round-trip compatibility results in a clean error message, so no further action needs to be taken.
3. Check iso8 status. The export succeeds, but what is the deploy behavior? If it works despite being on a different agent than intended, can we afford to fix it there?

Scope: iso9+ for sure.

jptrindade:
# Description

Disallow commas on agent name, fixing roundtrip issue.

Iso8 deploy works but there are some inconsistencies between the agent as seen on the resource_id and the agent seen on the db column.

Still not sure on how to proceed regarding iso8

closes #10342 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/10342-fix-agent-comma.yml
+++ b/changelogs/unreleased/10342-fix-agent-comma.yml
@@ -1,0 +1,6 @@
+description: Disallow commas in agent name.
+issue-nr: 10342
+change-type: patch
+destination-branches: [master, iso9]
+sections:
+  bugfix: "{{description}}"
--- a/changelogs/unreleased/10342-fix-agent-comma.yml
+++ b/changelogs/unreleased/10342-fix-agent-comma.yml
@@ -3,4 +3,4 @@ issue-nr: 10342
 change-type: patch
 destination-branches: [master, iso9]
 sections:
-  bugfix: "{{description}}"
+  bugfix: "Raise a clear error message instead of failing cryptically for invalid agent names"
--- a/changelogs/unreleased/10342-fix-agent-comma.yml
+++ b/changelogs/unreleased/10342-fix-agent-comma.yml
@@ -1,6 +1,6 @@
-description: Disallow commas in agent name.
+description: Raise a clear error when resource id components contain invalid characters.
 issue-nr: 10342
 change-type: patch
 destination-branches: [master, iso9]
 sections:
-  bugfix: "Raise a clear error message instead of failing cryptically for invalid agent names"
+  bugfix: "{{description}}"
--- a/src/inmanta/resources.py
+++ b/src/inmanta/resources.py
@@ -779,8 +779,14 @@ class Id:
         :attr attribute_value: The corresponding value for this key attribute.
         :attr version: The version number for this resource.
         """
-        if "," in agent_name:
-            raise ValueError(f"Agent name '{agent_name}' is not valid. Agent names cannot contain commas.")
+        for value, label, forbidden in (
+            (agent_name, "agent name", [","]),
+            (attribute, "attribute name", [",", "="]),
+            (attribute_value, "attribute value", ["]"]),
+        ):
+            for char in forbidden:
+                if char in value:
+                    raise ValueError(f"Invalid {label} '{value}': cannot contain '{char}'.")
 
         self._entity_type = entity_type
         self._agent_name = agent_name
--- a/src/inmanta/resources.py
+++ b/src/inmanta/resources.py
@@ -781,10 +781,7 @@ class Id:
         """
         if "," in agent_name:
             raise ResourceException(f"Agent name '{agent_name}' is not valid. Agent names cannot contain commas.")
-        # if "]" in attribute_value:
-        #     raise ResourceException(
-        #         f"Attribute value '{attribute_value}' is not valid. Attribute values cannot contain ']'."
-        #     )
+
         self._entity_type = entity_type
         self._agent_name = agent_name
         self._attribute = attribute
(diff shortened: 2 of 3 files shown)


src/inmanta/resources.py
```diff
@@ -772,13 +772,16 @@ class Id:
...
-        :attr attribute: The key attribute that uniquely identifies this resource on the agent
-        :attr attribute_value: The corresponding value for this key attribute.
-        :attr version: The version number for this resource.
+        :param agent_name: The agent responsible for this resource.
+        :param attribute: The key attribute that uniquely identifies this resource on the agent
+        :param attribute_value: The corresponding value for this key attribute.
+        :param version: The version number for this resource.
         """
+        if "," in agent_name:
+            raise ResourceException(f"Agent name '{agent_name}' is not valid. Agent names cannot contain commas.")
```
sanderr: Is this the only troublesome character for the regex? If so, I would add round-trip tests for others we can think of.

sanderr: Why `ResourceException`, and not e.g. `ValueError`?

changelogs/unreleased/10342-fix-agent-comma.yml
```diff
@@ -0,0 +1,6 @@
+description: Disallow commas in agent name.
+issue-nr: 10342
+change-type: patch
+destination-branches: [master, iso9]
+sections:
+  bugfix: "{{description}}"
```
sanderr, line 6:
```suggestion
  bugfix: "Raise a clear error message instead of failing cryptically for invalid agent names"
```

I think this might be a bit more user-oriented.

src/inmanta/resources.py
```diff
@@ -772,13 +772,16 @@ class Id:
 
     def __init__(self, entity_type: str, agent_name: str, attribute: str, attribute_value: str, version: int = 0) -> None:
         """
-        :attr entity_type: The resource type, as defined in the configuration model.
+        :param entity_type: The resource type, as defined in the configuration model.
```
sanderr: Are you confident about this? I'm not quite sure myself, but this may impact our generated docs so I think we need to be sure.

sanderr (changes requested):

src/inmanta/resources.py
jptrindade: I (claude) also found another case but I believe it is already addressed. If when creating an Id we include a `]`  on `attribute_value` then `Id.parse_id(id.resource_str())` will fail

sanderr: Then don't we need that for the constructor entrypoint as well?

src/inmanta/resources.py
jptrindade: Hum I thought it was just an outdated notation since I was getting some warnings on pycharm, and this is the only place we use this notation, but it seems we have code (and test cases) in place that expect this behaviour

src/inmanta/resources.py
jptrindade: I thought that since the Id was tied to a Resource it might make sense, but `ValueError` does in fact seem cleaner

src/inmanta/resources.py
sanderr: It might very well be outdated, I just don't know myself.

sanderr:
@jptrindade I think (not entirely sure) that there is one more gap: a comma in the attribute name: we'd need to reject that or add a round-trip test for it. Rejecting probably makes sense, because while an agent name might actually contain a comma, that would never be valid for an attribute afaik.

src/inmanta/resources.py
jptrindade: It does silently break the round-trip so I thought it might be overkill, but I guess it also is better to fail with a clear message. I'll make the changes

sanderr: > It does silently break the round-trip

But fixing that is precisely the purpose of this PR. Or am I confusing things?

jptrindade: I think the question was in HOW it broke the round-trip. 
- This broke the round-trip by failing on the constructor
- The comma on agent broke the round-trip by giving a wrong agent name.

But I also added the fix in this PR

sanderr: Ok, I see. There is indeed a distinction between round-trip `!=` and a failed round-trip. The former is arguably worse, and I can see how you decided to scope the PR to that. But I think it makes a lot of sense to extend it (as you did by now), under the principle of making invalid state irrepresentable.

sanderr (approved):

changelogs/unreleased/10342-fix-agent-comma.yml
```diff
@@ -0,0 +1,6 @@
+description: Raise a clear error when resource id components contain invalid characters.
+issue-nr: 10342
+change-type: patch
+destination-branches: [master, iso9]
```
Hugo-Inmanta, line 4: What about iso8 ? I saw the question in the PR description, is it addressed already ?

Hugo-Inmanta (changes requested):
Requesting changes just to not forget about iso8, LGTM otherwise 👍

Hugo-Inmanta (approved):

PR #10435, opened 2026-06-05
Add glibc version to env hash

arnaudsjs:
# Description

Add the glibc version to the hash function of an executor venv.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/add-glibc-to-hash-venv.yml
+++ b/changelogs/unreleased/add-glibc-to-hash-venv.yml
@@ -1,5 +1,5 @@
 ---
-description: "Add the glibc version to the hash function of an executor venv."
+description: "Recreate the executor venvs when the version of glibc on the host has changed."
 change-type: patch
 destination-branches: [master, iso9, iso8]
 sections:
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/add-glibc-to-hash-venv.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: "Add the glibc version to the hash function of an executor venv."
```
sanderr: I think this is too vague for the changelog. I don't think a user knows about the hash. The user-visible change is that the envs would be rebuilt.

sanderr (changes requested):
On note on the change entry. Other than that looks good to me.

Hugo-Inmanta (approved):

changelogs/unreleased/add-glibc-to-hash-venv.yml
arnaudsjs: Good point.

sanderr (approved):

PR #10459, opened 2026-06-11
Remove the ws-ping-interval configuration option

arnaudsjs:
# Description

The default value of the ws-ping-interval was set incorrectly. The Tornado docs indicate that the `ws-ping-interval` must always be larger than the `ws-ping-timeout`. By default Tornado will set the ws-ping-timeout to the value of ws-ping-interval if not set, which is a good default. Hence, we decided to drop the setting.

More context: https://inmanta.slack.com/archives/CKRF0C8R3/p1781095860509749

# Self Check:

- [x] Changelog entry


sanderr (approved):
I'm scoping my review to the implementation. I'm taking the goal for granted, because I understand you and Bart have had an in-depth discussion about it. Let me know if I should extend my scope.

bartv (approved):

PR #10460, opened 2026-06-11
Improve log message regarding removed pool member

arnaudsjs:
# Description

The log message about a pool member being removed due to inactivity was incorrect. It said _"is more than"_, but the condition would allow equality too. Like that the server log would contain log messages like: _"... will be shutdown because it was inactive for 60.00, which is more than 60"._ This message is logically incorrect of course. This PR fixes that.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/agent/resourcepool.py
+++ b/src/inmanta/agent/resourcepool.py
@@ -383,7 +383,7 @@ class TimeBasedPoolManager(PoolManager[TPoolID, TIntPoolID, TPoolMember]):
                         # Check that the executor can still be cleaned up by the time we have acquired the lock
                         if pool_member.can_be_cleaned_up() and pool_member.last_used < oldest_time and pool_member.running:
                             LOGGER.debug(
-                                "%s will be shutdown because it was inactive for %.2f, which is more than %d",
+                                "%s will be shutdown because it was inactive for %.2fs (retention time: %ds)",
                                 self.member_name(pool_member),
                                 (cleanup_start - pool_member.last_used).total_seconds(),
                                 self.retention_time,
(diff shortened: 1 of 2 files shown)


src/inmanta/agent/resourcepool.py
```diff
@@ -383,7 +383,7 @@ async def cleanup_inactive_pool_members(self) -> float:
                         # Check that the executor can still be cleaned up by the time we have acquired the lock
                         if pool_member.can_be_cleaned_up() and pool_member.last_used < oldest_time and pool_member.running:
                             LOGGER.debug(
-                                "%s will be shutdown because it was inactive for %.2f, which is more than %d",
+                                "%s will be shutdown because it was inactive for %.2fs (retention time: %ds)",
```
Hugo-Inmanta, line 386: I like this new message better, but unless I'm missing something I don't think things like `will be shutdown because it was inactive for 60.00, which is more than 60` could be logged in the first place.

In this scenario, assuming cleanup_start == t0 + 120 we would have:

(1) cleanup_start == t0 + 120 
(2) [from the log message params] cleanup_start - pool_member.last_used = 60   ==>   pool_member.last_used = t0 + 60
(3) [from the log message params] retention_time = 60
(4) [By definition L375 oldest_time = cleanup_start - expiry] oldest_time = t0 + 120 - 60   ==> oldest_time = t0 + 60 

And so since pool_member.last_used == oldest_time we wouldn't enter the `if` block in the first place.

Hugo-Inmanta (approved):

sanderr (approved):

src/inmanta/agent/resourcepool.py
arnaudsjs: Indeed. The root cause I described in the description of this PR is incorrect. The reason why the log message shows two times 60, is because of rounding. `%.2f` rounds the floating point number to two numbers past the comma. So the value is actually larger than 60, but the rounding makes them equal.

PR #10465, opened 2026-06-11
Improve logging websocket framework

arnaudsjs:
# Description

Improve the log messages emitted by the websocket layer. The main goals was to reduce the amount of very verbose log messages that contains the entire body of the websocket message. This information moved to TRACE level.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/protocol/websocket.py
+++ b/src/inmanta/protocol/websocket.py
@@ -791,7 +791,7 @@ class SessionEndpoint(endpoints.Endpoint, common.CallTarget, WebsocketFrameDecod
         """
         endpoints.Endpoint.__init__(self, name)
         WebsocketFrameDecoder.__init__(self)
-        LOGGER.debug("Start transport for client %s", name)
+        LOGGER.debug("Start websocket transport for client %s", name)
 
         self._sched = util.Scheduler("session endpoint")
 
--- a/src/inmanta/protocol/websocket.py
+++ b/src/inmanta/protocol/websocket.py
@@ -447,7 +447,7 @@ class WebsocketFrameDecoder(util.TaskHandler[None]):
         try:
             msg: WSMessages = self._message_parser.validate_json(message)
         except pydantic.ValidationError:
-            LOGGER.exception("Invalid message")
+            LOGGER.exception("Invalid message: %s", message)
             return
 
         LOGGER.log(const.LOG_LEVEL_TRACE, "Got %s", msg)
--- a/src/inmanta/protocol/websocket.py
+++ b/src/inmanta/protocol/websocket.py
@@ -399,7 +399,7 @@ class WebsocketFrameDecoder(util.TaskHandler[None]):
             call_spec.headers["Authorization"] = "Bearer " + self._token
         future: asyncio.Future[common.Result[types.ReturnTypes]] = asyncio.Future()
 
-        LOGGER.debug("Sending call (reply_id=%s): %s %s", reply_id, call_spec.method, call_spec.url)
+        LOGGER.info("Sending call (reply_id=%s): %s %s", reply_id, call_spec.method, call_spec.url)
         # Use the method-defined timeout or a generous default.
         # The timeout covers the full round-trip including handler execution on the remote side.
         timeout = max(properties.timeout or DEFAULT_RPC_TIMEOUT_S, MIN_RPC_TIMEOUT_S)
--- a/src/inmanta/protocol/websocket.py
+++ b/src/inmanta/protocol/websocket.py
@@ -534,16 +534,16 @@ class WebsocketFrameDecoder(util.TaskHandler[None]):
                     self._log_rpc_call(msg)
                     self.add_background_task(self._handle_call(msg))
                 else:
-                    LOGGER.warning("Received RPC_Call on inactive session, ignoring: %s", self)
+                    LOGGER.warning("Received RPC_Call on inactive session, ignoring: %s", msg)
 
             case RPC_Reply():
                 # A reply to a request sent by the server to the client
                 if not self.active():
-                    LOGGER.warning("Received RPC_Reply on inactive session, ignoring: %s", self)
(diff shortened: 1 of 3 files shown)


src/inmanta/protocol/websocket.py
```diff
@@ -431,15 +431,26 @@ async def _send_rpc_call(self, reply_id: uuid.UUID | None, message: str) -> None
             if future is not None and not future.done():
                 future.set_result(common.Result(code=503, result={"message": "Failed to send RPC call"}))
 
+    def _log_rpc_call(self, rpc_call: RPC_Call) -> None:
```
sanderr, line 434: Huh, weird naming here: `RPC_Call`. Not part of this PR of course.

src/inmanta/protocol/websocket.py
```diff
@@ -431,15 +431,26 @@ async def _send_rpc_call(self, reply_id: uuid.UUID | None, message: str) -> None
...
+        """
+        Write a log message to the log indicating that this rpc_call was received.
+        """
+        log_msg = "Received RPC_CALL %s %s"
+        args_log_msg = [rpc_call.method, rpc_call.url]
+        if rpc_call.reply_id:
+            log_msg += " (reply_id=%s)"
+            args_log_msg.append(str(rpc_call.reply_id))
```
jptrindade, line 442: Will we ever call this part? It seems to me that we don't because we only call this when we receive `RPC_Call` and not `RPC_Reply`

src/inmanta/protocol/websocket.py
```diff
@@ -431,15 +431,26 @@ async def _send_rpc_call(self, reply_id: uuid.UUID | None, message: str) -> None
...
         """Parse an incoming websocket frame and dispatch it based on message type."""
-        LOGGER.debug("%s got %s", self, message)
         try:
-            msg = self._message_parser.validate_json(message)
+            msg: WSMessages = self._message_parser.validate_json(message)
         except pydantic.ValidationError:
             LOGGER.exception("Invalid message")
             return
 
+        LOGGER.log(const.LOG_LEVEL_TRACE, "Got %s", msg)
```
sanderr, line 453: With moving this down, might we want to expand the exception log for the invalid message (3 lines up)?

src/inmanta/protocol/websocket.py
```diff
@@ -431,15 +431,26 @@ async def _send_rpc_call(self, reply_id: uuid.UUID | None, message: str) -> None
...
+    def _log_rpc_call(self, rpc_call: RPC_Call) -> None:
+        """
+        Write a log message to the log indicating that this rpc_call was received.
+        """
+        log_msg = "Received RPC_CALL %s %s"
+        args_log_msg = [rpc_call.method, rpc_call.url]
+        if rpc_call.reply_id:
+            log_msg += " (reply_id=%s)"
+            args_log_msg.append(str(rpc_call.reply_id))
+        LOGGER.info(log_msg, *args_log_msg)
```
sanderr, line 443: Isn't info a very high level? Honest question, I'm really bad at this.

src/inmanta/protocol/websocket.py
```diff
@@ -483,6 +494,12 @@ async def on_message(self, message: str | bytes) -> None:
                 if self._session is None:
                     LOGGER.error("Received a session open for a session that is not opened yet.")
                 else:
+                    LOGGER.info(
```
sanderr, line 497: I think I like info for these opening and closing events.

src/inmanta/protocol/websocket.py
```diff
@@ -523,6 +546,7 @@ async def on_message(self, message: str | bytes) -> None:
                     LOGGER.warning("Received a reply that is unknown: %s", msg.reply_id)
                     return
 
+                LOGGER.info("Received RPC_REPLY with reply_id=%s (code: %s)", msg.reply_id, msg.code)
```
sanderr, line 549: Same here with the log message being moved below the returns

src/inmanta/protocol/websocket.py
arnaudsjs: There are two types of API calls. Some are fire-and-forget (no response is expected) and for others we expect a response. When a response is expected the client generates a reply_id and send it to the receiving side. When the receiving side has a response, it sends the response back with the same reply_id so that the client can correlate the request and the response. So both the RPC_CALL and the RPC_REPLY message can have a reply_id, but on the RPC_CALL it's optional (hence the if-condition), because some API calls are fire-and-forget and don't have a reply_id.

sanderr (changes requested):
A few comments. I'm still trying to fully understand what the impact will be on what is logged, but I expect that will be fine.

src/inmanta/protocol/websocket.py
```diff
@@ -399,7 +399,7 @@ def rpc_call(
             call_spec.headers["Authorization"] = "Bearer " + self._token
         future: asyncio.Future[common.Result[types.ReturnTypes]] = asyncio.Future()
 
-        LOGGER.debug("Putting call %s: %s %s in queue at %s", reply_id, call_spec.method, call_spec.url, self)
+        LOGGER.debug("Sending call (reply_id=%s): %s %s", reply_id, call_spec.method, call_spec.url)
```
sanderr: Had a deeper look at what the changes would mean for the resulting log. Overall big improvement imo. The only thing that stands out to me is this: we log the call at `DEBUG` and the reply at `INFO`? I think we should be consistent.

src/inmanta/protocol/websocket.py
arnaudsjs: Good point.

src/inmanta/protocol/websocket.py
arnaudsjs: We log the REST API calls using INFO too, so I kept it consistent.

sanderr (approved):

jptrindade (approved):

PR #10492, opened 2026-06-22
Issue/9817 agent api access

### LINKED ISSUE #9817: Agent should be able to read facts
Currently, all the api routes allowing to read facts are restricted to the `api` client type.  This means that any code using the `agent` Client, while it can set facts, can not read them (when server auth is enabled).

On clear issue is that `std::FactReference` (in its current implementation) can not work when server auth is enabled.  cf. https://inmanta.slack.com/archives/CKRF0C8R3/p1765352981660999

The currently used workaround is to use the api type, and generate a token on the fly within the reference that needs it.  cf. https://inmanta.slack.com/archives/C01QH7140NT/p1765355040552119?thread_ts=1765295470.809079&cid=C01QH7140NT

EDIT: This api route still exists, and is allowed to be used by the agent, but it doesn't use the typed client, and std's reference doesn't use it atm.  https://github.com/inmanta/inmanta-core/blob/1e0ba42d37de968dfad4c3d7f2b28d39f9973647/src/inmanta/protocol/methods.py#L710

edvgui:
# Description

Extend list of endpoints that the agent client is allowed to access:
  - get_facts
  - get_fact
  - discovered_resources_get
  - discovered_resources_get_batch
  - discovered_resource_delete
  - discovered_resource_delete_batch

closes #9817 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/inmanta/protocol/methods_v2.py
+++ b/src/inmanta/protocol/methods_v2.py
@@ -936,7 +936,11 @@ def resource_logs(
 
 @auth(auth_label=const.CoreAuthorizationLabel.FACT_READ, read_only=True, environment_param="tid")
 @typedmethod(
-    path="/resource/<rid>/facts", operation="GET", arg_options=methods.ENV_OPTS, client_types=[ClientType.api, ClientType.agent], api_version=2
+    path="/resource/<rid>/facts",
+    operation="GET",
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.api, ClientType.agent],
+    api_version=2,
 )
 def get_facts(tid: uuid.UUID, rid: inmanta.types.ResourceIdStr) -> list[model.Fact]:
     """
--- a/src/inmanta/protocol/methods_v2.py
+++ b/src/inmanta/protocol/methods_v2.py
@@ -1750,7 +1750,7 @@ def discovered_resource_create_batch(tid: uuid.UUID, discovered_resources: Seque
     path="/discovered/<discovered_resource_id>",
     operation="GET",
     arg_options=methods.ENV_OPTS,
-    client_types=[ClientType.api],
+    client_types=[ClientType.api, ClientType.agent],
     api_version=2,
 )
 def discovered_resources_get(tid: uuid.UUID, discovered_resource_id: ResourceIdStr) -> model.DiscoveredResourceOutput:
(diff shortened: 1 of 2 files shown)


src/inmanta/protocol/methods_v2.py
```diff
@@ -1767,7 +1771,7 @@ def discovered_resources_get(tid: uuid.UUID, discovered_resource_id: ResourceIdS
     path="/discovered",
     operation="GET",
     arg_options=methods.ENV_OPTS,
-    client_types=[ClientType.api],
```
edvgui, line 1770: https://code.inmanta.com/solutions/modules/yang/-/blob/156f51a0674f2946ed0f4a5d434a316fb23e38c7/inmanta_plugins/yang/discovery.py#L496

src/inmanta/protocol/methods_v2.py
```diff
@@ -1841,7 +1845,7 @@ def discovered_resource_delete(
     path="/discovered/",
     operation="DELETE",
     arg_options=methods.ENV_OPTS,
-    client_types=[ClientType.api],
```
edvgui, line 1844: https://code.inmanta.com/solutions/modules/yang/-/blob/156f51a0674f2946ed0f4a5d434a316fb23e38c7/inmanta_plugins/yang/discovery.py#L517

src/inmanta/protocol/methods_v2.py
```diff
@@ -936,7 +936,11 @@ def resource_logs(
 
 @auth(auth_label=const.CoreAuthorizationLabel.FACT_READ, read_only=True, environment_param="tid")
 @typedmethod(
-    path="/resource/<rid>/facts", operation="GET", arg_options=methods.ENV_OPTS, client_types=[ClientType.api], api_version=2
+    path="/resource/<rid>/facts",
+    operation="GET",
+    arg_options=methods.ENV_OPTS,
+    client_types=[ClientType.api, ClientType.agent],
```
edvgui, line 942: https://github.com/inmanta/std/blob/45f669b38c6427a0940f0db6d5d86436ca75dd2a/inmanta_plugins/std/__init__.py#L1523

edvgui (commented):
For reference

jptrindade (approved):

arnaudsjs (approved):

PR #10496, opened 2026-06-22
Add is_set TypeGuard helper for GraphQL filter checks

jptrindade:
# Description

Factored out of #10328.

Introduces an `is_set` TypeGuard helper in `inmanta.graphql.schema` and uses it to replace the repeated `x is not None and x is not strawberry.UNSET` checks scattered through the GraphQL filter/pagination code.

```python
def is_set[T](value: T | None) -> typing.TypeGuard[T]:
    return value is not None and value is not strawberry.UNSET
```

Because it's a `TypeGuard`, mypy narrows away `None`/`strawberry.UNSET` in the positive branch exactly like the inline check did, so there is no behavioural change — this is a pure readability refactor.

Note: the one spot that passed `getattr(...)` (typed `Any`) now annotates `attr: CustomFilter | None` so the TypeGuard narrows to a concrete type instead of `Never`.

# Self Check:

- [x] Attached issue to pull request (N/A — refactor split out of #10328)
- [x] Changelog entry
- [x] No (preventable) type errors (verified with mypy: no new errors vs master)
- [x] Sufficient test cases (no behaviour change; existing GraphQL filter/pagination tests cover the touched code)

### DIFF
--- a/changelogs/unreleased/is-set-typeguard-helper.yml
+++ b/changelogs/unreleased/is-set-typeguard-helper.yml
@@ -1,0 +1,5 @@
+description: Replace the repeated `x is not None and x is not strawberry.UNSET` checks in the GraphQL schema with an `is_set` TypeGuard helper.
+change-type: patch
+destination-branches: [master, iso9]
+sections:
+  other-note: "{{description}}"
--- a/changelogs/unreleased/is-set-typeguard-helper.yml
+++ b/changelogs/unreleased/is-set-typeguard-helper.yml
@@ -1,5 +1,3 @@
 description: Replace the repeated `x is not None and x is not strawberry.UNSET` checks in the GraphQL schema with an `is_set` TypeGuard helper.
 change-type: patch
 destination-branches: [master, iso9]
-sections:
-  other-note: "{{description}}"
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -346,6 +346,16 @@ mapper: CustomStrawberrySQLAlchemyMapper[typing.Any] = CustomStrawberrySQLAlchem
 DEFAULT_PER_PAGE: int = 50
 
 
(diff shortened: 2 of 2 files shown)


changelogs/unreleased/is-set-typeguard-helper.yml
```diff
@@ -0,0 +1,5 @@
+description: Replace the repeated `x is not None and x is not strawberry.UNSET` checks in the GraphQL schema with an `is_set` TypeGuard helper.
+change-type: patch
+destination-branches: [master, iso9]
+sections:
+  other-note: "{{description}}"
```
arnaudsjs: Why does this need an entry in the changelog. Does this cause a user-visible change?

src/inmanta/graphql/schema.py
```diff
@@ -346,6 +346,16 @@ def _connection_type_for(self, type_name: str) -> typing.Type[typing.Any]:
 DEFAULT_PER_PAGE: int = 50
 
 
+def is_set[T](value: T | None) -> typing.TypeGuard[T]:
```
arnaudsjs: This name is rather confusing. `is_set` gives the impression that we only check for `is not UNSET`. But we also check for `is not None`. I think we should give it a name that better represents what it does.

A few potential names:

* is_provided()
* has_value()

If you can think of something better, go ahead.

arnaudsjs (changes requested):

changelogs/unreleased/is-set-typeguard-helper.yml
jptrindade: Nope, it was Claude stuff that I missed. I'll remove it.

src/inmanta/graphql/schema.py
jptrindade: I agree. I changed it to is_provided

arnaudsjs (approved):

Issue #10502, opened 2026-06-23
Race conditions in scheduler


arnaudsjs:
- [x] The `start_working()` and `stop_working()` methods of the Agent class don't have any concurrently control. If both methods are invoked at the same time, the agent could end up in an inconsistent state.
- [x] If the `start()` method of a `TimeBasedPoolManager` is invoked twice, two cleanup jobs will be running concurrently.
- [x] It's possible to call `get()` on a `PoolManager` that is not running. For the `TimeBasedPoolManager` this means that the created member will not be cleaned up according to its policy.

PR #10505, opened 2026-06-24
Fix bug inconsistent administrative/operational scheduler state

arnaudsjs:
# Description

This PR fixes a bug that triggers in the following scenario:

* The DB is in read-only mode (because a node in the DB cluster failed and there is no quorum anymore).
* The scheduler starts and its `start_working()` method is invoked.
* The `self.working` flag of the scheduler is set to True, but the `await self.scheduler.start()` call raises an exception, because it needs write access to the DB. At this point there is an inconsistency between the administrative "working" state of the scheduler (value of the self.working flag) and the operational "working" state of the scheduler (whether the scheduler is actually performing work). The result is that the scheduler will not process new model version.
* When the DB cluster fully recovers, the scheduler will still be stuck and process no new model version.

Concrete fixes:

* Make sure we only set the `self.working` to True iff the scheduler was started successfully.
* Prevent a race condition where the `start_working()` and `stop_working()` method of the scheduler are executing concurrently.
* Make sure that the `start_working()` method doesn't start the scheduler anymore when `stop()` was called and the agent process is actually shutting down.

Part of #10502

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/fix-inconsistent-scheduler-running-state.yml
+++ b/changelogs/unreleased/fix-inconsistent-scheduler-running-state.yml
@@ -1,0 +1,5 @@
+---
+description: "Fixed bug where the scheduler would stop processing new model versions if the database has been running in read-only mode."
+destination-branches: [master, iso9, iso8]
+sections:
+  bugfix: "{{description}}"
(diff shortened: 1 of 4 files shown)


Hugo-Inmanta (approved):
Cool 👍

jptrindade (approved):

PR #10506, opened 2026-06-24
Add SQLAlchemy query_expression column/output-field contributions to the resources GraphQL query

jptrindade:
_Description updated by Claude on behalf of @jptrindade._

## Description

Extends the GraphQL extension-contribution mechanism so extensions can contribute **SQLAlchemy `query_expression` columns and output fields** to the `resources` query, in addition to the filter fields they could already contribute.

- Extension columns are merged onto a dynamically built subclass of `models.Resource` (`ComposedResource`), so the `resources` query selects a single ORM object that can carry extra SQL-backed columns (joins/subqueries) that don't live on the resource table. When no extension contributes columns, `models.Resource` is used directly and this machinery is skipped.
- Columns are populated per query via `with_expression`, and **only when the matching output field is actually selected** — unrequested/expensive columns stay at their null `query_expression` default. `get_selected_field_names` exposes the names of the selected fields that each contribution checks against.
- New `GraphQLContribution` hooks: `get_resource_sqlalchemy_columns` and `populate_resource_columns`, plus `get_resource_graphql_output_type_mixin` for the typed output fields. Extensions register through `GraphQLSlice.register_extension_contribution` before the schema is built at slice `start()`.

This is the core-side counterpart of the LSM resource output fields (service entity / instance / lifecycle state).

## Tests

`tests/test_graphql.py::test_custom_extension_contributions` extended to cover a `query_expression`-backed output field populated by an extension, asserting it's only populated when selected. **Passing.**


src/inmanta/graphql/graphql.py
```diff
@@ -16,42 +16,65 @@
...
+        """
+        if self.schema is None:
+            if extension_name in self.extension_contributions:
+                raise Exception(f"Extension {extension_name} already registered.")
+            self.extension_contributions[extension_name] = contribution
+        else:
+            raise Exception(
+                f"Can't register extension contribution for {extension_name} "
+                "because the GraphQL schema has already been generated"
+            )
```
arnaudsjs, line 72: I think the error message have to mention the actual root cause.
```suggestion
            raise Exception(
                f"Can't register extension contribution for {extension_name} "
                "because the GraphQLSlice was already started."
            )
```

src/inmanta/graphql/graphql.py
```diff
@@ -16,42 +16,65 @@
...
         return [SLICE_COMPILER]
 
+    def register_extension_contribution(self, extension_name: str, contribution: type[GraphQLContribution]) -> None:
+        """
+        Register an extension contribution. Only possible before the slice starts (during the `prestart` stage) and
+        if the extension name has not already been registered.
+        """
+        if self.schema is None:
+            if extension_name in self.extension_contributions:
+                raise Exception(f"Extension {extension_name} already registered.")
```
arnaudsjs:
```suggestion
                raise Exception(f"Contribution for extension {extension_name} already registered.")
```

src/inmanta/graphql/graphql.py
```diff
@@ -16,42 +16,65 @@
...
     def __init__(self) -> None:
         super().__init__(name=SLICE_GRAPHQL)
         self.context = None
+        self.schema = None
+        self.extension_contributions = {}
 
     def get_dependencies(self) -> list[str]:
         return [SLICE_COMPILER]
 
+    def register_extension_contribution(self, extension_name: str, contribution: type[GraphQLContribution]) -> None:
```
arnaudsjs: The name of the method is `extension_contribution`, but the type is called `GraphQLContribution`. Is there a reason why we the type has an inconsistent name?

src/inmanta/graphql/schema.py
```diff
@@ -677,24 +670,28 @@ def key_to_model(self) -> dict[str, type[models.Base]]:
...
 
 
-@mapper.type(models.Resource)
-class Resource:
+class CoreResourceBase:
+    """
+    Mixin carrying the added core resource output fields. It is merged with the extensions' output mixins (and mapped onto
+    the SQLAlchemy resource model) by `build_resource_strawberry_type` to build the `Resource` GraphQL output type.
+    """
+
```
arnaudsjs: The docstring indicates that we call this a Mixin. If that is the case, we need to make sure that the name of the class also contains the term Mixin to improve readability. Or is there a reason why we shouldn't do that? We already have many Resource classes in this code base so adding one with the generic name `CoreResourceBase` doesn't make the situation any better.

src/inmanta/graphql/schema.py
```diff
@@ -922,7 +919,114 @@ async def get_connection[*Ts](
         )
 
 
-def get_schema(context: GraphQLContext) -> strawberry.Schema:
+class GraphQLContribution(ABC):
+    """
+    Extension hook that lets extensions (e.g. LSM) contribute extra information to GraphQL queries
+    without core being aware of those modules.
```
arnaudsjs:
```suggestion
    without a tight coupling between core and these extensions.
```

src/inmanta/graphql/schema.py
```diff
@@ -922,7 +919,114 @@ async def get_connection[*Ts](
...
+        _sqlalchemy_columns.update(c.get_resource_sqlalchemy_columns())
+    if not _sqlalchemy_columns:
+        return models.Resource
+    return cast(
+        type[models.Resource],
+        type("ComposedResource", (models.Resource,), _sqlalchemy_columns),
+    )
+
+
+def build_resource_strawberry_type(
```
arnaudsjs:
```suggestion
def build_resource_strawberry_output_type(
```

src/inmanta/graphql/schema.py
```diff
@@ -922,7 +919,114 @@ async def get_connection[*Ts](
...
+    for _base in (CoreResourceBase, *_resource_mixins):
+        _resource_annotations.update(_base.__dict__.get("__annotations__", {}))
+        for _k, _v in _base.__dict__.items():
+            if _k not in _skip_attrs:
+                _resource_attrs[_k] = _v
+
+    # Can't do the same as the resource filter because the mixins can't have the mapper.type decorator and that is required
+    return cast(
+        type, mapper.type(resource_model)(type("Resource", (), {"__annotations__": _resource_annotations, **_resource_attrs}))
+    )
```
arnaudsjs: Why do all these variables start with an underscore? This is rather unconventional, isn't it?

src/inmanta/graphql/schema.py
```diff
@@ -922,7 +919,114 @@ async def get_connection[*Ts](
...
+    _resource_annotations: dict[str, object] = {}
+    _resource_attrs: dict[str, object] = {}
+    _skip_attrs = {"__dict__", "__weakref__", "__doc__", "__annotations__", "__module__"}
+    for _base in (CoreResourceBase, *_resource_mixins):
+        _resource_annotations.update(_base.__dict__.get("__annotations__", {}))
+        for _k, _v in _base.__dict__.items():
+            if _k not in _skip_attrs:
+                _resource_attrs[_k] = _v
+
```
arnaudsjs: Shouldn't we add a safety check to verify that no attribute name is defined twice?

src/inmanta/graphql/schema.py
```diff
@@ -922,7 +919,114 @@ async def get_connection[*Ts](
         )
 
 
-def get_schema(context: GraphQLContext) -> strawberry.Schema:
+class GraphQLContribution(ABC):
```
arnaudsjs, line 954: The docstring of this method seems to indicate that this is a rather generic class, but the methods in the implementation are very resource-specific. Why is that?

src/inmanta/graphql/schema.py
```diff
@@ -922,7 +919,114 @@ async def get_connection[*Ts](
...
+
+    :param extension_contributions: list of extension graphql contributions
+    :param resource_model: the SQLAlchemy model that backs the output type (see `build_resource_sqlalchemy_model`)
+    """
+    _resource_mixins: tuple[type, ...] = tuple(
+        mixin for c in extension_contributions if (mixin := c.get_resource_graphql_output_type_mixin()) is not None
+    )
+    _resource_annotations: dict[str, object] = {}
+    _resource_attrs: dict[str, object] = {}
+    _skip_attrs = {"__dict__", "__weakref__", "__doc__", "__annotations__", "__module__"}
```
arnaudsjs: Can't we just skip everything that starts with a double underscore?

arnaudsjs (changes requested):

arnaudsjs:
The extensive docstrings helped a lot to review this PR. Well done!

src/inmanta/graphql/schema.py
jptrindade: At the moment, resource is the only special case that requires this extension composition. But theoretically, we can do this for any class (but I hope we don't have to)

src/inmanta/graphql/schema.py
jptrindade: There were 2 exceptions, but I handle those separately

src/inmanta/graphql/schema.py
jptrindade: yeah, I also find it weird. I changed it.

src/inmanta/graphql/schema.py
jptrindade: renamed to CoreResourceMixin

src/inmanta/graphql/graphql.py
jptrindade: The way I see it, each extension can register a `GraphQLContribution` . I read this method as register extension's  (graphql) contribution. But maybe it is not clear. What would you suggest? I would like to make clear that this method is something extensions would call

arnaudsjs: I see. What do you think about: `register_contribution_for_extension()` or `register_graphql_contribution_for_extension()`

src/inmanta/graphql/graphql.py
```diff
@@ -16,42 +16,64 @@
...
+        Register an extension contribution. Only possible before the slice starts (during the `prestart` stage) and
+        if the extension name has not already been registered.
+        """
+        if self.schema is None:
+            if extension_name in self.extension_contributions:
+                raise Exception(f"Contribution for extension {extension_name} already registered.")
+            self.extension_contributions[extension_name] = contribution
+        else:
+            raise Exception(
+                f"Can't register extension contribution for {extension_name} " "because the GraphQLSlice was already started."
```
arnaudsjs:
```suggestion
                f"Can't register extension contribution for {extension_name} because the GraphQLSlice was already started."
```

src/inmanta/graphql/schema.py
arnaudsjs: Let me put it differently. The `GraphQLContribution` class is an abstract class. This mean by definition that it should be a generic implementation to support different types of subclasses. Now we have an abstract class that defines methods that only belong to one specific subclass (namely for resources). That doesn't make much sense. This is the interface towards the different extensions that want to add a contribution. If another extension that lsm would contribute something, the parent class would contain methods that are only relevant for the lsm extension. That's rather weird.

src/inmanta/graphql/graphql.py
```diff
@@ -16,42 +16,64 @@
...
     def __init__(self) -> None:
         super().__init__(name=SLICE_GRAPHQL)
         self.context = None
+        self.schema = None
+        self.extension_contributions = {}
 
     def get_dependencies(self) -> list[str]:
         return [SLICE_COMPILER]
 
+    def register_graphql_contribution_for_extension(self, extension_name: str, contribution: type[GraphQLContribution]) -> None:
```
arnaudsjs, line 58: I would expect that every GraphQLContribution instance contains a contribution to one specific type (strawberry type/sqlalchemy class). Given the signature of this method, it seems like there is one instance per extension. How is this going to work when the lsm extension at some points needs to contribute to more types than only the resource type? I have the feeling that this would cause a breaking change on the contributions API between core and its extensions. Is that correct?

src/inmanta/graphql/schema.py
```diff
@@ -922,7 +919,121 @@ async def get_connection[*Ts](
         )
 
 
-def get_schema(context: GraphQLContext) -> strawberry.Schema:
+class GraphQLContribution(ABC):
```
arnaudsjs, line 954: Why are the methods in this class classmethods?

arnaudsjs (changes requested):

src/inmanta/graphql/schema.py
jptrindade: Because there is no "instance" state required. It could also be a staticmethod, what do you think is more appropriate in this case?

src/inmanta/graphql/graphql.py
jptrindade: If we were to extend to different classes, we would create additional methods on `GraphQLContribution` to support them. But I understand where you are coming from. Do you feel like this should be a per type per extension map, and that we would make this a bit more generic?

src/inmanta/graphql/schema.py
jptrindade: > If another extension that lsm would contribute something, the parent class would contain methods that are only relevant for the lsm extension.

I don't think that is necessarily true. I believe that these methods are independent from lsm, but I agree that, as it stands, they are only relevant for the resource type.

src/inmanta/graphql/graphql.py
arnaudsjs: > If we were to extend to different classes, we would create additional methods on GraphQLContribution to support them

That means we have to apply changes to the inmanta-core side every time we want add a contribution on the extension? I don't think we want that coupling. 

> Do you feel like this should be a per type per extension map, and that we would make this a bit more generic?

The most important to me is that there is a clear and stable interface from inmanta-core to the extensions to add contributions. The currently implementation is so much hard coded towards resource that I feell like we will have to break this interface on a future update. I think that we will have to make it more generic in order to obtain this generic interface that we can keep stable for a long time.

src/inmanta/graphql/schema.py
arnaudsjs: I think the best practice is to only make a method static if it only has a purpose at the type level and not at the instance level. Do you see any benefits of making these methods static?

jptrindade: It's like you said, this method does not have a purpose at instance level, so I figured it would be best to have it as static

src/inmanta/graphql/graphql.py
jptrindade: >That means we have to apply changes to the inmanta-core side every time we want add a contribution on the extension? I don't think we want that coupling.

If it is a change to another model besides Resource.

> I think that we will have to make it more generic in order to obtain this generic interface that we can keep stable for a long time.

I can see the value in this, I'll update the PR to make it more generic and support a GraphQLContribution per type per extension. Although (as it stands) we only support that for environments and notifications which are the other 2 queries that we have. So I'll build the output object / sqlalchemy model like I do for the resource. Off the top of my head, I'm not sure how to make it more generic

src/inmanta/graphql/schema.py
arnaudsjs: This I don't understand. These methods are only called from instance objects, so that means that are only relevant within the context of an instance.

jptrindade: Are they? I don't think so. We register the class itself, not an instance

arnaudsjs (approved):
Cool! The interface is much better now

PR #10515, opened 2026-06-29
Raise exception on get() call when PoolManager not running

arnaudsjs:
# Description

Fix race condition on agent pause where an executor can be created during/after the shutdown of the executor manager. The executors created during this period of time will not be stopped or cleaned up.

Part of #10502

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/tasks.py
+++ b/src/inmanta/deploy/tasks.py
@@ -261,7 +261,7 @@ class DryRun(Task):
                 rvid=executor_resource_details.rvid,
                 agent=agent,
                 started=started,
-                log_msg_for_log="Skipping dryrun for resource %s because due to an error in constructing the executor",
+                log_msg_for_log="Skipping dryrun for resource %s due to an error when constructing the executor",
                 log_msg_for_report="Unable to construct an executor for this resource",
             )
         else:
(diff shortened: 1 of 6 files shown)


src/inmanta/deploy/tasks.py
```diff
@@ -238,43 +248,53 @@ async def execute(self, task_manager: "scheduler.TaskManager", agent: str, reaso
...
-                changes={
-                    "handler": AttributeStateChange(
-                        current="FAILED", desired="Unable to construct an executor for this resource"
-                    )
-                },
+                agent=agent,
                 started=started,
-                finished=datetime.datetime.now().astimezone(),
-                messages=[],
+                log_msg_for_log="Skipping dryrun for resource %s because due to an error in constructing the executor",
```
Hugo-Inmanta:
```suggestion
                log_msg_for_log="Skipping dryrun for resource %s due to an error when constructing the executor",
```

Hugo-Inmanta (changes requested):
I noticed a typo in the existing logs, LGTM otherwise 👍

Hugo-Inmanta (approved):

src/inmanta/util/__init__.py
jptrindade, line 1: Unrelated but I noticed that we define here a `FinallySet` class. Is that being used anywhere or is it dead code?

jptrindade (approved):

src/inmanta/util/__init__.py
arnaudsjs: I think it's indeed dead code.

PR #10516, opened 2026-06-29
Fix cleanup job leak in TimeBasedPoolManager

### LINKED ISSUE #10502: Race conditions in scheduler

arnaudsjs:
# Description

The TimeBasedPoolManager might leak cleanup jobs when the following sequence of calls is done on that TimeBasedPoolManager:

`request_shutdown()` -> `start()` -> `join()`

The call to `start()` will start a new cleanup job. 

The old cleanup jobs might not be stopped when the following conditions hold:

* The `request_shutdown()` method is called while the `cleanup_inactive_pool_members_task` method is not sleeping. In that case the sleep call is not cancelled and no CancelledError will be raised inside the loop of the old cleanup job.
* If the `start()` method executes before the `running` flag is evaluated by the old cleanup job. In that case the running flag will evaluate to True again and the cleanup job will not drop out of its loop.

The result is that the old and the new cleanup job run side by side.

closes #10502

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/fix-cleanup-job-leak.yml
+++ b/changelogs/unreleased/fix-cleanup-job-leak.yml
@@ -1,0 +1,6 @@
+---
+description: "Fix bug where the TimeBasedPoolManager might leak cleanup jobs."
+issue-nr: 10502
+issue-repo: inmanta-core
+change-type: patch
+destination-branches: [master, iso9, iso8]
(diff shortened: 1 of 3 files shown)


jptrindade (approved):

Hugo-Inmanta (changes requested):
Should we add a test for this one as well ? Or maybe it is too cumbersome to test ?

Hugo-Inmanta (approved):
Cool, the test helps a lot to understand what was going wrong.

PR #10529, opened 2026-07-03
Add database singleton lock to prevent multiple active servers

bartv:
> :robot: Opened by Claude on behalf of Bart.

# Description

Enforce that **at most one Inmanta server is active on a given database**, treating it as a data-integrity invariant rather than an opt-in HA feature.

On startup the `DatabaseService` acquires a PostgreSQL **session-level advisory lock** on a dedicated connection (separate from the shared pool) and holds it for the whole lifetime of the process:

- A second server pointed at the same database **waits** for the lock (`database.singleton_lock_wait_time`, default 30s, same tri-state semantics as `database.wait_time`: `0` = try once, `<0` = wait forever) and **refuses to start** with a clear error if it cannot acquire it. Startup contention surfaces as a `ServerStartFailure`.
- The lock is acquired **before** schema migration, so a stray second server can neither migrate nor serve.
- While running, a monitor verifies the lock is still held. If it is lost (connection dropped, or the DB became read-only after a failover), the server **fails fast** (self-`SIGTERM`) so it stops before another instance takes over. Because a session-level advisory lock is released automatically when its owning connection ends, a standby can then take over immediately.

Advisory locks include `MyDatabaseId` in their tag, so this is scoped **per database**: multiple Inmanta databases in one PostgreSQL cluster do not conflict.

This is enabled by default. It can be disabled with `database.singleton_lock=false` (logs a loud warning; unsafe). The lock needs a **session-mode** connection to the primary, so it does not work through a transaction-pooling proxy (pgbouncer in transaction mode).

### Motivation

Today two servers against one database silently corrupt state (double deploys, scheduler races). This converts that silent failure into a loud, safe refusal to start, and provides the primitive that external active/passive HA (e.g. keepalived) builds on.

### New config options

| Option                              | Default | Purpose                                              |
| ----------------------------------- | ------- | ---------------------------------------------------- |
| `database.singleton_lock`           | `true`  | Enable the single-active-server guard.               |
| `database.singleton_lock_wait_time` | `30`    | Seconds to wait for the lock before refusing to start. |

closes ~*Add ticket reference here*~

### Notes for reviewers

- `change-type` is set to **minor**; it is arguably breaking since a (previously silently racing) two-instance setup now refuses to start. Happy to bump to `major` if preferred.
- Fail-fast uses `os.kill(os.getpid(), SIGTERM)` to reuse the existing graceful-shutdown path (with the signal handler's hard-exit timer as backstop).
- End-user HA documentation is not included here; suggest a follow-up docs section describing the singleton lock and the active/passive setup.

# Self Check:

- [x] Changelog entry
- [x] No (preventable) type errors (diff is mypy-baseline clean; `make mypy` locally shows only pre-existing environment-specific noise in untouched files)
- [x] Sufficient test cases (`tests/test_singleton_lock.py`: conflict/refusal, wait-then-acquire handover, monitor fail-fast, and full-server refusal)

### DIFF
--- a/src/inmanta/server/config.py
+++ b/src/inmanta/server/config.py
@@ -58,6 +58,32 @@ db_name = Option("database", "name", "inmanta", "The name of the database on the
 db_username = Option("database", "username", "postgres", "The username to access the database in the PostgreSQL server", is_str)
 db_password = Option("database", "password", None, "The password that belong to the database user", is_str)
 
+db_singleton_lock = Option(
+    "database",
+    "singleton_lock",
+    True,
+    "Guard against more than one Inmanta server being active on the same database at the same time. "
+    "When enabled (the default), the server acquires a PostgreSQL advisory lock on startup and holds it "
+    "for its entire lifetime. A second server pointed at the same database will wait for the lock (see "
+    ":inmanta.config:option:`database.singleton_lock_wait_time`) and refuse to start if it cannot get it. "
+    "Running two servers against one database corrupts it, so only disable this if you are certain no "
+    "other server is running. The lock requires a session-mode connection to the primary: it does not "
+    "work through a transaction-pooling proxy such as pgbouncer in transaction mode.",
+    is_bool,
+)
+
+db_singleton_lock_wait_time = Option(
+    "database",
+    "singleton_lock_wait_time",
+    30,
+    "How long, in seconds, the server waits for the singleton lock (see "
+    ":inmanta.config:option:`database.singleton_lock`) to be released by another server before it gives "
+    "up and refuses to start. This absorbs restarts where the previous server has not fully released the "
+    "lock yet. If set to 0, the server tries only once. If set to a negative value, the server waits "
+    "forever.",
+    is_time,
+)
+
 db_connection_pool_min_size = Option(
     "database",
     "connection_pool_min_size",
--- a/src/inmanta/server/config.py
+++ b/src/inmanta/server/config.py
@@ -67,8 +67,11 @@ db_singleton_lock = Option(
     "for its entire lifetime. A second server pointed at the same database will wait for the lock (see "
     ":inmanta.config:option:`database.singleton_lock_wait_time`) and refuse to start if it cannot get it. "
     "Running two servers against one database corrupts it, so only disable this if you are certain no "
-    "other server is running. The lock requires a session-mode connection to the primary: it does not "
-    "work through a transaction-pooling proxy such as pgbouncer in transaction mode.",
+    "other server is running. The lock uses one dedicated database connection, separate from and in "
+    "addition to the server's connection pool (see "
+    ":inmanta.config:option:`server.db_connection_pool_max_size`), so size the database's max_connections "
+    "accordingly. The lock requires a session-mode connection to the primary: it does not work through a "
+    "transaction-pooling proxy such as pgbouncer in transaction mode.",
     is_bool,
 )
 
@@ -77,7 +80,7 @@ db_singleton_lock_wait_time = Option(
     "singleton_lock_wait_time",
     30,
     "How long, in seconds, the server waits for the singleton lock (see "
-    ":inmanta.config:option:`database.singleton_lock`) to be released by another server before it gives "
+    ":inmanta.config:option:`database.singleton_lock`) to be acquired before it gives "
     "up and refuses to start. This absorbs restarts where the previous server has not fully released the "
     "lock yet. If set to 0, the server tries only once. If set to a negative value, the server waits "
     "forever.",
--- a/src/inmanta/server/services/databaseservice.py
+++ b/src/inmanta/server/services/databaseservice.py
@@ -20,8 +20,9 @@ import asyncio
 import json
 import logging
 import os.path
+import signal
 from functools import total_ordering
-from typing import Mapping, Optional
+from typing import Callable, Mapping, Optional
 
 import asyncpg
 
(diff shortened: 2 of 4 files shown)


src/inmanta/server/config.py
```diff
@@ -58,6 +58,32 @@
...
+    is_bool,
+)
+
+db_singleton_lock_wait_time = Option(
+    "database",
+    "singleton_lock_wait_time",
+    30,
+    "How long, in seconds, the server waits for the singleton lock (see "
+    ":inmanta.config:option:`database.singleton_lock`) to be released by another server before it gives "
+    "up and refuses to start. This absorbs restarts where the previous server has not fully released the "
```
arnaudsjs:
```suggestion
    ":inmanta.config:option:`database.singleton_lock`) to be acquired before it gives "
    "up and refuses to start. This absorbs restarts where the previous server has not fully released the "
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -251,18 +252,182 @@ async def get_connection_status(self) -> bool:
...
+    def __init__(self, *, host: str, port: int, username: str, password: Optional[str], database: str) -> None:
+        self._host = host
+        self._port = port
+        self._username = username
+        self._password = password
+        self._database = database
+        self._connection: Optional[asyncpg.Connection] = None
+        self._monitor_task: Optional[asyncio.Task[None]] = None
+        self._on_lock_lost: Optional[Callable[[], None]] = None
+        self._lock_lost_signalled: bool = False
```
arnaudsjs: Signalled with two l's looks rather weird to me. But it seems to be valid British English so feel free to ignore this.
```suggestion
        self._lock_lost_signaled: bool = False
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -275,6 +440,40 @@ async def stop(self) -> None:
...
+        self._singleton_lock.start_monitor(on_lock_lost=self._on_singleton_lock_lost)
+
+    def _on_singleton_lock_lost(self) -> None:
+        """
+        Called when the singleton lock is lost while the server is running. Requests a graceful shutdown of
+        the whole process (a hard-exit timer in the signal handler backstops a stuck shutdown).
+        """
+        if self.is_stopping():
+            return
+        LOGGER.critical("Shutting down the server because the database singleton lock was lost.")
```
arnaudsjs: Don't we already log a very similar log message in the monitor task?

tests/test_singleton_lock.py
```diff
@@ -0,0 +1,120 @@
...
+        await lock1.acquire(wait_time=0)
+
+        async def release_after_delay() -> None:
+            await asyncio.sleep(1)
+            await lock1.stop()
+
+        releaser = asyncio.ensure_future(release_after_delay())
+        # Should block for ~1s until lock1 is released, then succeed well within wait_time.
+        await lock2.acquire(wait_time=10)
+        await releaser
```
arnaudsjs:
```suggestion
        assert releaser.done()
```

src/inmanta/server/services/databaseservice.py
```diff
@@ -251,18 +252,182 @@ async def get_connection_status(self) -> bool:
         return False
 
 
+class SingletonLock:
+    """
+    Enforces that at most one Inmanta server is active on a given database.
+
+    On startup the server acquires a PostgreSQL session-level advisory lock on a dedicated connection
+    (not part of the shared pool) and holds it for the whole lifetime of the process. Because a
+    session-level advisory lock is released automatically when its owning connection ends, a standby
```
arnaudsjs, line 261: Should we take this additional connection into account for the `server.db-connection-pool-max-size` configuration option? Namely the actual connection pool will be one connection smaller than configured.

arnaudsjs (changes requested):

src/inmanta/server/services/databaseservice.py
bartv: This is actually not a pool but a dedicated connection that is not part of the pool. But makes sense to add it somewhere to the docs. I will see where it fits best

src/inmanta/server/config.py
bartv: _Claude here on behalf of Bart._ Applied, thanks - reworded to "...to be acquired before it gives up...". (6834abe07)

src/inmanta/server/services/databaseservice.py
bartv: _Claude here on behalf of Bart._ Renamed to `_lock_lost_signaled`. (6834abe07)

src/inmanta/server/services/databaseservice.py
bartv: _Claude here on behalf of Bart._ Good catch - removed the duplicate `critical` log in `_on_singleton_lock_lost`; the monitor task already logs the reason before it triggers the shutdown. (6834abe07)

tests/test_singleton_lock.py
bartv: _Claude here on behalf of Bart._ Done - replaced `await releaser` with `assert releaser.done()`. (6834abe07)

src/inmanta/server/services/databaseservice.py
bartv: _Claude here on behalf of Bart._ Documented in the `database.singleton_lock` option help: it now notes the lock uses one dedicated connection, separate from and in addition to the pool (with a cross-reference to `server.db_connection_pool_max_size`), and to size the database's `max_connections` accordingly. Happy to move it to a dedicated HA docs section if you'd prefer. (6834abe07)

tests/test_singleton_lock.py
bartv: _Claude here on behalf of Bart._ Heads-up: `assert releaser.done()` turned out to be racy and failed in CI. PostgreSQL releases the session advisory lock as soon as it receives the connection's `Terminate`, but asyncpg's `connection.close()` only returns once the transport is fully torn down. So `lock2.acquire()` can win the lock (PG already released it) while `release_after_delay` is still suspended inside `lock1.stop()`, leaving the task not yet `done()`. Reworked it (eba2bec9b) to assert that `acquire()` actually *blocked* until the release (`waited >= 0.5s`) and kept `await releaser` for cleanup, which is deterministic and captures the same intent.

arnaudsjs (approved):

PR #10530, opened 2026-07-03
Add generic GraphQL filter composition (extensions can contribute filter fields)

### LINKED ISSUE #10475: [GraphQL] Create model_version filter
Right now we only return results that are on their latest version of the model (with or without orphans). The goal of this ticket is to add a field that allows us to choose the model version that we want to show. 

NOTE: orphans are only a concept that makes sense when we don't filter on version so these 2 filters should raise an exception if they are set together

### LINKED ISSUE #10476: [GraphQL] Enable ResourceFilter composition
We want to expand the resources query in such a way to be able to filter on lsm terms like service entity or instance.
To achieve this we will create a `GraphQLContribution` class that will be subclassed by extensions and they will use that to register their contributions to GraphQL.

In this first instance, we only need to allow filters on Resource, so it will define a `get_resource_filter_input_class` method that returns a `ResourceFilterABC` object with logic on how to apply the desired extension filters.

jptrindade:
closes: #10476
closes: #10475 

# Description

Lets extensions (e.g. LSM) contribute their own filter fields to the GraphQL queries, composed generically into each query's filter input. Builds on the already-merged output-field contribution mechanism (`GraphQLContribution` / `populate_sqlalchemy_columns`) — this adds the *filter* side, following the same generic, per-object-type pattern (keyed by `get_target_model`), not just for `resources`.

This is the filter-composition part of #10328, rebuilt on current master (most of that PR — the output-field contribution mechanism and the `is_provided` helper — is already merged, and the `GraphQLContribution` API has since been rewritten, so the original diff no longer applies).

The only end-user-visible change is the new `modelVersion` filter on the `resources` query (the changelog surfaces just that under `minor-improvement`); the filter-composition mechanism itself is extension-facing plumbing.

## What changed

- New generic hook `GraphQLContribution.get_filter_input_class()` returning a `@strawberry.input` filter class, merged into the target object type's filter input — the filter-side analogue of `get_graphql_output_type_mixin`.
- The registry now carries each object type's core building blocks (`ContributableGraphQLType`: `core_mixin` + `core_filter` + `base_filter`). The existing `EnvironmentFilter` / `NotificationFilter` are renamed to `CoreEnvironmentFilter` / `CoreNotificationFilter`; the composed inputs keep the user-facing `EnvironmentFilter` / `NotificationFilter` / `ResourceFilter` names, so **the exposed GraphQL schema is unchanged** (for the no-extension case). Each contributed filter is validated to subclass the type's `base_filter`, and duplicate field names across components are rejected.
- Generic `get_filter_components` / `build_composed_filter_input` compose each type's filter by multiple inheritance; `get_schema` builds every type's composed filter in the same loop as its output type; each resolver decomposes the received value back into components and validates them (`decompose_and_validate_filter`), then applies each `apply_filter`.
- Resources additionally split into `ResourceFilterABC` (shared base + version-selection hooks) and `CoreResourceFilter`. Version selection is delegated and expressed as data: at most one component may take it over via `handles_version()`, and that component returns a `ModelVersionSelection` (`pinned` / `latest`) from `resolve_model_version()` (an extension, else core by default). The resolver applies the version join once and records the selection on the query (`resolved_model_version` execution option), so contributions can resolve version-dependent data (output columns, filters) at the same version the resources are taken at. Core also exposes a new `modelVersion` filter to fetch resources as present in a specific model version. Since a specific version is returned as-is, `modelVersion` is mutually exclusive with the current-state filters (`isOrphan`, `isDeploying`, `blocked`, `compliance`, `lastHandlerRun`), which reflect a resource's current, unversioned state.
- New extension-facing hook `ResourceFilterABC.allows_persistent_state_count()`: a single predicate a filter component uses to keep the `resources` query's efficient `ResourcePersistentState`-only count valid. It replaces two overlapping concerns (version-selection validity + filtering on the `Resource` table) — the fast count is used only when every component allows it (no pinned historical version, and no `apply_filter` constraining the `Resource` table); otherwise the resolver falls back to counting the full version-aware query. Default is the conservative `False`.
- `StrawberryFilter.get_filter_dict` now skips `UNSET` fields instead of raising, so a component only carrying some of the composed filter's fields composes cleanly — consistent with how the resource filter already treats `UNSET`. One existing test relied on the old raise-on-omitted-optional-variable behaviour; it's updated to the new (skip) behaviour.

The previous orphan/latest-version CTE logic is now expressed through `CoreResourceFilter.resolve_model_version` (`ModelVersionSelection.latest`); behaviour for the existing filters (incl. `isOrphan`) is unchanged.

## Tests

- `test_query_resources_model_version` — the new `modelVersion` filter across model versions, orphans and combined filters (incl. the rejected combinations with current-state filters), asserting `totalCount` tracks the filter.
- `test_resources_count_path` — asserts the actual count path the resolver takes (efficient `ResourcePersistentState`-only vs. full version-aware fallback) by capturing the count statement: RPS-only filters and `isOrphan` stay efficient, while `purged` (Resource-table filter), `modelVersion` (pinned), and an extension taking over version selection each force the fallback. Also covers an extension whose `allows_persistent_state_count()` returns `True` keeping the fast count valid.
- `test_custom_extension_resource_filter` — an extension contributing a resource filter: its `apply_filter` runs, it can take over version selection via `atVersion` (`handles_version` / `resolve_model_version`), and conflicting version handlers (extension + core `modelVersion`/`isOrphan`) are rejected.
- `test_resolved_model_version_available_to_contributions` — a contribution reads back the resolved model version via `resolved_model_version(stmt)` and resolves a version-dependent output column at that same version.
- `test_custom_extension_environment_filter` — an extension contributing a filter to the (non-resource) `environments` query, proving the mechanism is generic.
- `test_custom_extension_filter_field_collision` / `test_custom_extension_filter_validation` — collision detection when composing filters, and that `validate_filter` runs for non-resource filters too.
- `test_custom_extension_contributions` / `test_extension_registers_multiple_contributions` — registration bookkeeping and multiple contributions per extension.

Full `tests/test_graphql.py` passes; `mypy` (no new baseline entries) and `flake8` clean.

> Note: the companion `inmanta-lsm` change (its resource filter migrated to the new `resolve_model_version` / `allows_persistent_state_count` hooks) is in a separate LSM PR and is not part of this one.

# Self Check:

- [x] Changelog entry
- [ ] If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s)

### DIFF
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -502,7 +502,7 @@ class StrawberryFilter:
         """
         Returns the provided filter fields in dict form, to feed to the SQLAlchemy query as an exact-match filter.
         Fields left UNSET are skipped (not filtered on) -- this is what lets a filter compose with others that carry
-        their own fields (see `decompose_filter`), where any given component only has some of the fields set.
+        their own fields (see `decompose_filter`).
         This is only used for simple filters i.e. exact match on the same table.
         """
         return {key: value for key, value in self.__dict__.items() if value is not strawberry.UNSET}
@@ -750,8 +750,6 @@ class ResourceFilterABC(StrawberryFilter):
     """
     Abstract base class that defines the shared attributes/behaviour of every resource filter component: core's
     `CoreResourceFilter` and each extension's resource filter (contributed via `GraphQLContribution.get_filter_input_class`).
-    These components are composed into the single `ResourceFilter` GraphQL input type exposed on the `resources` query
-    by the generic filter composition (see `get_filter_components` / `build_composed_filter_input` and `get_schema`).
     The version-selection hooks below are specific to resources; other object types' filters compose the same way but
     have no version concept.
 
@@ -809,7 +807,7 @@ class CoreResourceFilter(ResourceFilterABC):
     def validate_filter(self) -> None:
         # modelVersion returns the resources exactly as they were in that version of the model. The filters below
         # reflect the *current* state of a resource (tracked on ResourcePersistentState, which is not versioned),
-        # so combining them with modelVersion would mix a historical snapshot with current state; reject that.
+        # so combining them with modelVersion would mix a historical snapshot with current state.
         if is_provided(self.model_version):
             current_state_filters = {
                 "isOrphan": self.is_orphan,
@@ -951,8 +949,11 @@ def add_filter_and_sort[*Ts](
     """
     Adds filter and sorting to the given statement.
 
+    :param stmt: the query to add the filter and sorting to.
+    :param default_sorting: the sorting to apply for each key not already covered by `order_by`.
     :param filter: the filter components to apply. A single query can have several components: the `resources` query
         composes core and extension filters, while the other queries pass a single (or no) filter.
+    :param order_by: the sorting requested by the user, taking precedence over `default_sorting` for the same key.
     """
     for filter_component in filter:
         stmt = filter_component.apply_filter(stmt)
@@ -1219,9 +1220,9 @@ def get_filter_components(
 ) -> tuple[type[StrawberryFilter], ...]:
     """
     Return the filter components that compose an object type's filter input type: the type's `core_filter` followed by
-    every extension-contributed filter class (see `GraphQLContribution.get_filter_input_class`). `build_composed_filter_input`
-    builds the composed input from these by multiple inheritance, and each query's resolver decomposes the received
-    filter back into one instance per component (see `decompose_filter`) to apply them.
+    every extension-contributed filter class. `build_composed_filter_input` builds the composed input from these by multiple
+    inheritance, and each query's resolver decomposes the received filter back
+    into one instance per component (see `decompose_filter`) to apply them.
 
     :param core_filter: the core filter class of the object type (e.g. `CoreResourceFilter`).
     :param contributions: the extension contributions that target the same object type.
@@ -1239,13 +1240,7 @@ def build_composed_filter_input(
 ) -> tuple[tuple[type[StrawberryFilter], ...], type]:
     """
     Build the filter input type for an object type, composed of its core filter and the extensions' contributed
-    filters (mirrors `build_strawberry_output_type` on the output side). The components are merged by multiple
-    inheritance into a single `@strawberry.input` named `{type_name}Filter` (the user-facing filter type), while the
-    core filter keeps a distinct name (e.g. `CoreResourceFilter`). `get_schema` is called once per process, so a fixed
-    class name is fine.
-
-    Returns the tuple of components (so the resolver can decompose the received filter, see `decompose_filter`) and
-    the composed input type (to expose as the query's `filter` argument).
+    filters. The components are merged by multiple inheritance into a single `@strawberry.input` named `{type_name}Filter`.
 
     :param type_name: the name of the object type being built (e.g. "Resource").
     :param core_filter: the core filter class of the object type (e.g. `CoreResourceFilter`).
@@ -1269,17 +1264,14 @@ def build_composed_filter_input(
 
 def decompose_filter[F: StrawberryFilter](filter: object, components: tuple[type[F], ...]) -> list[F]:
     """
-    Split a composed filter value (built by `build_composed_filter_input`) back into one instance per component,
-    reading each component's fields off the composed value. This lets every component apply its own `apply_filter`
-    (and, for resources, its version-selection hooks) on just its own fields.
+    Split a composed filter value back into one instance per component, reading each component's fields off the composed value.
+    This lets every component apply its own `apply_filter` on just its own fields.
 
     :param filter: the composed filter value received by the resolver.
-    :param components: the filter components the composed type was built from (see `get_filter_components`).
+    :param components: the filter components the composed type was built from.
     """
     decomposed: list[F] = []
     for component in components:
-        # Every component is a @strawberry.input, hence a dataclass, but that isn't visible through the StrawberryFilter
-        # bound (the base class is not itself a dataclass), so mypy can't tell dataclasses.fields accepts it.
         component_fields = dataclasses.fields(component)  # type: ignore[arg-type]
         decomposed.append(component(**{field.name: getattr(filter, field.name) for field in component_fields}))
     return decomposed
@@ -1299,8 +1291,7 @@ class RegistrableGraphQLType:
 
 # The object types extensions can register GraphQL contributions for (see GraphQLContribution), mapping each SQLAlchemy
 # model to its core building blocks. `get_schema` composes each of these from the core building blocks and the
-# registered contributions; registrations for any other model are rejected. The GraphQL output type name is always the
-# model's class name (e.g. `models.Resource` -> "Resource"), which is also the key the contributions are grouped under.
+# registered contributions; registrations for any other model are rejected.
 REGISTRABLE_MODELS: "Mapping[type[models.Base], RegistrableGraphQLType]" = {
     models.Resource: RegistrableGraphQLType(core_mixin=CoreResourceMixin, core_filter=CoreResourceFilter),
     models.Environment: RegistrableGraphQLType(core_mixin=CoreEnvironmentMixin, core_filter=CoreEnvironmentFilter),
@@ -1362,11 +1353,9 @@ def get_schema(
             stmt = contribution.populate_sqlalchemy_columns(stmt, composed_model, requested_fields)
         return stmt
 
-    # For each registrable object type, build its (possibly extension-composed) output type and filter input type.
-    # The output type and the filter input are composed the same generic way from the type's core building blocks and
-    # the contributions targeting it; each query's resolver decomposes the received filter back into its components.
+    # Build each registrable object type's output type and (extension-composed) filter input, both composed the same
+    # generic way from the type's core building blocks (see RegistrableGraphQLType) and the contributions targeting it.
     built_types: dict[GraphQLTypeName, tuple[type[models.Base], type]] = {}
-    # type_name -> (filter components the composed input was built from, composed filter input type)
     built_filters: dict[GraphQLTypeName, tuple[tuple[type[StrawberryFilter], ...], type]] = {}
     for base_model, registrable in REGISTRABLE_MODELS.items():
         type_name = graphql_type_name(base_model)
@@ -1374,12 +1363,12 @@ def get_schema(
         built_types[type_name] = build_output_type(base_model, registrable.core_mixin)
         built_filters[type_name] = build_composed_filter_input(type_name, registrable.core_filter, contributions)
 
-    environment_model, Environment = built_types["Environment"]
-    notification_model, Notification = built_types["Notification"]
-    resource_model, Resource = built_types["Resource"]
-    environment_filter_components, EnvironmentFilter = built_filters["Environment"]
-    notification_filter_components, NotificationFilter = built_filters["Notification"]
-    resource_filter_components, ResourceFilter = built_filters["Resource"]
+    environment_model, Environment = built_types[graphql_type_name(models.Environment)]
+    notification_model, Notification = built_types[graphql_type_name(models.Notification)]
+    resource_model, Resource = built_types[graphql_type_name(models.Resource)]
+    environment_filter_components, EnvironmentFilter = built_filters[graphql_type_name(models.Environment)]
+    notification_filter_components, NotificationFilter = built_filters[graphql_type_name(models.Notification)]
+    resource_filter_components, ResourceFilter = built_filters[graphql_type_name(models.Resource)]
 
     class CustomInfo(Info):
         @property
@@ -1396,8 +1385,6 @@ def get_schema(
             after: typing.Optional[str] = strawberry.UNSET,
             last: typing.Optional[int] = strawberry.UNSET,
             before: typing.Optional[str] = strawberry.UNSET,
-            # The declared type is the core filter (so it stays typed), but the GraphQL input the user gets is the
-            # composed type carrying the core + extension fields.
             filter: typing.Annotated[
                 typing.Optional[CoreEnvironmentFilter], strawberry.argument(graphql_type=typing.Optional[EnvironmentFilter])
             ] = strawberry.UNSET,
@@ -1417,8 +1404,6 @@ def get_schema(
         async def notifications(
             self,
             info: CustomInfo,
-            # The declared type is the core filter (so it stays typed), but the GraphQL input the user gets is the
-            # composed type carrying the core + extension fields.
             filter: typing.Annotated[CoreNotificationFilter, strawberry.argument(graphql_type=NotificationFilter)],
             first: typing.Optional[int] = strawberry.UNSET,
             after: typing.Optional[str] = strawberry.UNSET,
@@ -1440,8 +1425,6 @@ def get_schema(
         async def resources(
             self,
             info: CustomInfo,
-            # The declared type is CoreResourceFilter (so filter.<core field> stays typed), but the GraphQL input
-            # the user actually gets is the composed type carrying core + extension fields.
             filter: typing.Annotated[CoreResourceFilter, strawberry.argument(graphql_type=ResourceFilter)],
             first: typing.Optional[int] = strawberry.UNSET,
             after: typing.Optional[str] = strawberry.UNSET,
@@ -1459,7 +1442,7 @@ def get_schema(
 
             # Decompose the composed ResourceFilter into one instance per component (core + each extension). Every
             # resource filter component is a ResourceFilterABC, and at most one may take over version selection
-            # (`handles_version`); core does it by default.
+            # Core does it by default.
             resource_filter_instances = cast(list[ResourceFilterABC], decompose_filter(filter, resource_filter_components))
             version_handler: ResourceFilterABC | None = None
             filters_on_resource_table: bool = False
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -1292,10 +1292,10 @@ class RegistrableGraphQLType:
 # The object types extensions can register GraphQL contributions for (see GraphQLContribution), mapping each SQLAlchemy
 # model to its core building blocks. `get_schema` composes each of these from the core building blocks and the
 # registered contributions; registrations for any other model are rejected.
-REGISTRABLE_MODELS: "Mapping[type[models.Base], RegistrableGraphQLType]" = {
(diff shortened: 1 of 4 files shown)


src/inmanta/graphql/schema.py
```diff
@@ -1087,14 +1214,88 @@ def build_strawberry_output_type(
...
-REGISTRABLE_MODELS: "Mapping[type[models.Base], type]" = {
-    models.Resource: CoreResourceMixin,
-    models.Environment: CoreEnvironmentMixin,
-    models.Notification: CoreNotificationMixin,
+# model to its core building blocks. `get_schema` composes each of these from the core building blocks and the
+# registered contributions; registrations for any other model are rejected.
+REGISTRABLE_MODELS: "Mapping[type[models.Base], RegistrableGraphQLType]" = {
+    models.Resource: RegistrableGraphQLType(core_mixin=CoreResourceMixin, core_filter=CoreResourceFilter),
+    models.Environment: RegistrableGraphQLType(core_mixin=CoreEnvironmentMixin, core_filter=CoreEnvironmentFilter),
+    models.Notification: RegistrableGraphQLType(core_mixin=CoreNotificationMixin, core_filter=CoreNotificationFilter),
```
arnaudsjs: I think that a contribution is the term that we choose for an extension adding something from a core. So I think we have to make sure with stick to that term.
```suggestion
CONTRIBUTABLE_MODELS: "Mapping[type[models.Base], RegistrableGraphQLType]" = {
    models.Resource: ContributableGraphQLType(core_mixin=CoreResourceMixin, core_filter=CoreResourceFilter),
    models.Environment: ContributableGraphQLType(core_mixin=CoreEnvironmentMixin, core_filter=CoreEnvironmentFilter),
    models.Notification: ContributableGraphQLType(core_mixin=CoreNotificationMixin, core_filter=CoreNotificationFilter),
```

src/inmanta/graphql/schema.py
```diff
@@ -1152,13 +1353,21 @@ def populate_extension_columns[*Ts](
...
-    notification_model, Notification = built_types["Notification"]
-    resource_model, Resource = built_types["Resource"]
+    # Build each registrable object type's output type and filter input.
+    built_types: dict[GraphQLTypeName, tuple[type[models.Base], type]] = {}
+    built_filters: dict[GraphQLTypeName, tuple[tuple[type[StrawberryFilter], ...], type]] = {}
+    for base_model, registrable in REGISTRABLE_MODELS.items():
+        type_name = graphql_type_name(base_model)
+        contributions = extension_contributions.get(type_name, [])
+        built_types[type_name] = build_output_type(base_model, registrable.core_mixin)
+        built_filters[type_name] = build_composed_filter_input(type_name, registrable.core_filter, contributions)
```
arnaudsjs:
```suggestion
    built_output_types: dict[GraphQLTypeName, tuple[type[models.Base], type]] = {}
    built_filters: dict[GraphQLTypeName, tuple[tuple[type[StrawberryFilter], ...], type]] = {}
    for base_model, registrable in REGISTRABLE_MODELS.items():
        type_name = graphql_type_name(base_model)
        contributions = extension_contributions.get(type_name, [])
        built_output_types[type_name] = build_output_type(base_model, registrable.core_mixin)
        built_filters[type_name] = build_composed_filter_input(type_name, registrable.core_filter, contributions)
```

tests/test_graphql.py
```diff
@@ -1546,3 +1550,431 @@ def get_graphql_output_type_mixin(cls) -> type | None:
...
+async def test_query_resources_model_version(server, client, environment, setup_database, mixed_resource_generator):
+    """
+    Test that filtering the resource query on `modelVersion` returns the resources as present in that specific
+    version of the model, instead of the latest released version.
+
+    We include setup_database to have some resources in other envs to make sure we don't leak across environments.
+    """
+    instances = 2
+    resources_per_version = 10
+    orphans = resources_per_version // 2
```
arnaudsjs: Why floor division?

tests/test_graphql.py
```diff
@@ -1546,3 +1550,431 @@ def get_graphql_output_type_mixin(cls) -> type | None:
...
+    total_resources_in_latest_version = resources_per_version * instances
+    await mixed_resource_generator(environment, instances, resources_per_version)
+
+    # mixed_resource_generator creates 2 versions per instance (agent). With 2 instances this gives us
+    # the following configuration model versions (each agent owns resource set "set<agent_index>"):
+    #   v1: full compile, set0 created with <resources_per_version> resources (agent0)
+    #   v2: put_partial, set0 updated -> <resources_per_version> resources, the un-updated half becomes orphaned
+    #   v3: put_partial, set1 created with <resources_per_version> resources (agent1), set0 unchanged
+    #   v4: put_partial, set1 updated -> <resources_per_version> resources, the un-updated half becomes orphaned
```
arnaudsjs: This is unclear to me. I read this as `<resources_per_version>` resources are updated. But what is the un-updated half then?

tests/test_graphql.py
```diff
@@ -1546,3 +1550,431 @@ def get_graphql_output_type_mixin(cls) -> type | None:
...
+    assert result.result["data"]["data"] is None
+    assert len(result.result["data"]["errors"]) == 1
+    assert result.result["data"]["errors"][0] == "Only one extension can determine version logic."
+
+
+async def test_custom_extension_environment_filter(server, client, project_default, caplog):
+    """
+    Filter composition is generic across object types, not resource-specific: an extension can contribute a filter
+    (via GraphQLContribution.get_filter_input_class) to the `environments` query too. Its fields are composed into the
+    EnvironmentFilter input and its apply_filter runs alongside core's.
```
arnaudsjs: This is unclear to me.

tests/test_graphql.py
```diff
@@ -1546,3 +1550,431 @@ def get_graphql_output_type_mixin(cls) -> type | None:
...
+        check_correct_graphql_response(result)
+        log_contains(caplog, __name__, logging.INFO, "Applied environment filter alp")
+    names = {edge["node"]["name"] for edge in result.result["data"]["data"]["environments"]["edges"]}
+    assert names == {"alpha", "alpine"}
+
+    # The filter is optional: without it, all environments are returned.
+    result = await client.graphql(query="{ environments { edges { node { name } } } }")
+    check_correct_graphql_response(result)
+    all_names = {edge["node"]["name"] for edge in result.result["data"]["data"]["environments"]["edges"]}
+    assert {"alpha", "alpine", "beta"} <= all_names
```
arnaudsjs: Why `<=`?

tests/test_graphql.py
```diff
@@ -1546,3 +1550,431 @@ def get_graphql_output_type_mixin(cls) -> type | None:
...
+    graphql_slice.schema = None
+    graphql_slice.register_graphql_contribution_for_extension("example", CollidingContribution)
+    with pytest.raises(Exception, match="id defined more than once in Environment filters."):
+        await graphql_slice.start()
+
+
+async def test_custom_extension_filter_validation(server, client, project_default):
+    """
+    validate_filter is available on every filter, not just resources: an extension contributing a filter to the
+    (non-resource) `environments` query can reject invalid field combinations, and the error is surfaced to the user.
```
arnaudsjs:
```suggestion
    Validate that an error is raised if a filter, contributed by extension, rejects an invalid filter combination using its validate_filter method.
```

arnaudsjs (changes requested):

tests/test_graphql.py
jptrindade: I don't think it matters if it is floor or regular division because we have an even `resources_per_version`. If it was odd, then floor division would be correct with the logic from the mixed_resource_generator (and the fact that we can't have half an orphan)

tests/test_graphql.py
jptrindade: The un-updated half are the resources that became orphaned.
In the `mixed_resource_generator` each instance creates 2 versions, v1 that has resources (for example) with id from 1 - 10, then v2 updates some of the original resources, orphans the rest and creates new resources

tests/test_graphql.py
jptrindade: I think maybe Claude had some issue with environments leaking from the fixtures. But I figure this part of the test might go anyway since it doesn't have value

tests/test_graphql.py
jptrindade: This test was just to show that we could compose non-resource filters as well. I simplified the docstring

tests/test_graphql.py
arnaudsjs: I agree. That's why I find it rather strange to do a floor division on an even number. It creates confusion.

tests/test_graphql.py
arnaudsjs: My point is: Either you know what the `mixed_resource_generator` does and then this comment is redundant. Or you don't know what the `mixed_resource_generator` and then this comment doesn't help anything.

src/inmanta/graphql/schema.py
```diff
@@ -1231,64 +1439,42 @@ async def resources(
...
+                if filter_instance.handles_version():
+                    if version_handler is not None:
+                        raise ValueError("Only one extension can determine version logic.")
+                    version_handler = filter_instance
+                if filter_instance.filters_on_resource_table():
+                    filters_on_resource_table = True
+
+            # True when a component actively took over version selection
+            # rather than the plain default of "latest scheduled version + orphans".
+            custom_version_selection = version_handler is not None
```
bartv: **Silent performance regression (should-fix).** `custom_version_selection` disables the fast `ResourcePersistentState`-only COUNT whenever a component `handles_version()`, and `CoreResourceFilter.handles_version()` returns True for `isOrphan` too. But `apply_version_filter` never branches on `is_orphan`, so the version selection (and thus the RPS-only count) is unchanged for `isOrphan` - yet a plain `isOrphan: true/false` (a common dashboard filter) now forces the expensive `Resource`-joined count. On master only `purged` disabled it.

Decouple the single-owner rule (above) from the count gate: only an extension taking over, or a pinned `modelVersion` (a historical snapshot, where RPS reflects current state) should disable the fast count. No test pins which count path runs, so CI won't catch this - worth asserting `count_stmt is None` only for the `modelVersion`/extension cases.

```suggestion
            # Only a genuine change to version selection invalidates the ResourcePersistentState-only count: an
            # extension taking over selection, or a pinned model_version (a historical snapshot). is_orphan leaves
            # the default selection unchanged, so it must not disable the fast count -- it only enforces the
            # single-version-owner rule above.
            custom_version_selection = version_handler is not None and (
                not isinstance(version_handler, CoreResourceFilter) or is_provided(version_handler.model_version)
            )
```

src/inmanta/graphql/schema.py
```diff
@@ -1231,64 +1439,42 @@ async def resources(
...
-            # CTE that fetches the latest scheduled version
-            latest_scheduled_version_cte = (
-                select(models.Scheduler.last_processed_model_version.label("version"))
-                .where(models.Scheduler.environment == filter.environment)
-                .cte()
-            )
+            # Decompose the composed ResourceFilter into one instance per component (core + each extension). Every
+            # resource filter component is a ResourceFilterABC, and at most one may take over version selection
+            # Core does it by default.
+            resource_filter_instances = cast(list[ResourceFilterABC], decompose_filter(filter, resource_filter_components))
```
bartv: **Should-fix (fail-fast).** `cast(list[ResourceFilterABC], ...)` is a runtime no-op, and nothing enforces that a `models.Resource` contribution's `get_filter_input_class()` actually returns a `ResourceFilterABC` subclass (only the docstring says so). A non-conforming extension fails with a cryptic `AttributeError: ... 'handles_version'` on the first `resources` query rather than at schema build. Consider a fail-fast check where the resource components are assembled, e.g.:

```python
if base_model is models.Resource:
    for component in resource_filter_components:
        if not issubclass(component, ResourceFilterABC):
            raise Exception(f"{component.__name__} must subclass ResourceFilterABC to filter resources.")
```

src/inmanta/graphql/schema.py
```diff
@@ -733,9 +744,47 @@ class CoreResourceMixin:
...
+        raise NotImplementedError()
+
+    def filters_on_resource_table(self) -> bool:
+        """
+        Return True if this component's `apply_filter` constrains the `Resource` table (`models.Resource`) rather
+        than only `ResourcePersistentState`. The `resources` query uses this to decide whether the efficient count
+        (which counts `ResourcePersistentState` alone, without joining `Resource`) is valid: it is only used when no
+        applied component filters on the `Resource` table.
+        """
+        return False
```
bartv, line 855: **Silently-wrong default (should-fix / decide).** Defaulting to `False` means an extension that filters the `Resource` table but forgets to override this returns a silently **undercounted** `totalCount` (the fast RPS-only count is used when it shouldn't be) - wrong data, no error, no test failure. Safer to make the fast path opt-in (default `True` = "assume it touches `Resource`, count the real query"; components that only touch `ResourcePersistentState` override to `False`); core already overrides correctly with `is_provided(self.purged)`. If you keep the current default, please at least add a test with an extension returning `True` so this safety net is actually exercised (no test covers it today). Not providing a one-line suggestion since it interacts with the method's naming/docstring - your call.

src/inmanta/graphql/schema.py
```diff
@@ -1231,64 +1439,42 @@ async def resources(
...
+            # resource filter component is a ResourceFilterABC, and at most one may take over version selection
+            # Core does it by default.
+            resource_filter_instances = cast(list[ResourceFilterABC], decompose_filter(filter, resource_filter_components))
+            version_handler: ResourceFilterABC | None = None
+            filters_on_resource_table: bool = False
+            for filter_instance in resource_filter_instances:
+                filter_instance.validate_filter()
+                if filter_instance.handles_version():
+                    if version_handler is not None:
+                        raise ValueError("Only one extension can determine version logic.")
```
bartv: **Misleading message (should-fix).** The case this fires on (tested: `atVersion` + `isOrphan`) is extension-vs-**core**; core is not an extension, so "Only one extension can determine version logic" misdescribes it. This exact string is asserted in a test, so update the assertion too.

```suggestion
                        raise ValueError(
                            "Multiple filter components tried to control version selection; at most one may: "
                            "an extension (via handles_version), or core (via isOrphan / modelVersion)."
                        )
```

src/inmanta/graphql/schema.py
```diff
@@ -733,9 +744,47 @@ class CoreResourceMixin:
...
+        component may do so; when none does, core (`CoreResourceFilter`) selects the version by default.
+        """
+        return False
+
+    def apply_version_filter[*Ts](self, stmt: Select[tuple[*Ts]]) -> Select[tuple[*Ts]]:
+        """
+        Restrict the query to the appropriate version(s) of the model. Only invoked on the single component
+        responsible for version selection: an extension whose `handles_version()` is True, or core otherwise.
+        """
+        raise NotImplementedError()
```
bartv: **Nit.** A bare `NotImplementedError()` surfaces as an empty message in the GraphQL error response (the generic `except Exception` in `graphql.py`). Give it text so a misconfigured extension is diagnosable.

```suggestion
        raise NotImplementedError(
            f"{type(self).__name__} returns handles_version()=True but does not implement apply_version_filter()."
        )
```

src/inmanta/graphql/schema.py
```diff
@@ -652,8 +661,9 @@ class CoreEnvironmentMixin:
     settings: scalars.JSON = strawberry.field(resolver=get_settings, description="The settings for this environment.")
 
 
+@dataclasses.dataclass(kw_only=True)
```
bartv: **Nit.** `@strawberry.input` already applies `dataclasses.dataclass(kw_only=True)` internally, so this decorator is a no-op here (and on lines 695, 747, 785; the extension filters in the tests omit it and compose fine). Safe to drop all four.

image


```suggestion
```

bartv (changes requested):
_Claude posted this review on behalf of @bartv._

Reviewed the composition mechanism and the resource version/count logic (on `981885ca`, on top of Arnaud's round). The design is sound and well documented, and the compose/decompose path is correct and tested.

One should-fix: setting `isOrphan` now disables the fast `ResourcePersistentState`-only COUNT even though it does not change version selection - a silent performance regression versus master on a common filter (details inline; no test pins which count path runs). Plus smaller ones: fail-fast that resource contributions subclass `ResourceFilterABC`, the "only one extension" version-conflict message (the tested case is core-vs-core, not two extensions), the unsafe `filters_on_resource_table` default, and two nits (empty `NotImplementedError`, redundant `@dataclasses.dataclass(kw_only=True)` on the core filters). All details are in the inline comments.

jptrindade:
With the new changes I changed a bit how versions are handled. This is because inmanta-lsm required us to know which model version was being pinned (to fetch the appropriate LifecycleTransfer resources).

bartv (approved):

src/inmanta/graphql/schema.py
```diff
@@ -733,9 +742,119 @@ class CoreResourceMixin:
...
+    @staticmethod
+    def latest(environment: uuid.UUID, *, include_orphans: bool = True) -> "ModelVersionSelection":
+        """
+        The latest scheduled model version. Orphaned resources are additionally taken at the last version they were
+        present in when `include_orphans` is True; when False, only resources present in the latest version are selected.
+        """
+        return ModelVersionSelection(
+            version=latest_scheduled_model_version(environment),
+            include_orphans=include_orphans,
+        )
```
arnaudsjs: In our code base we generally use classmethods for method that create an instance of the class. I think we should stick with that convention. The names of the method are also not really self-explaining. An attempt to give them a better name:
```suggestion
    @staticmethod
    def create_for_exact_version(version: "int | ColumnElement[int]") -> "ModelVersionSelection":
        """Select every resource exactly at `version` (a concrete model version, or a scalar expression computing one)."""
        return ModelVersionSelection(version=version)

    @staticmethod
    def create_for_latest_scheduled_version(environment: uuid.UUID, *, include_orphans: bool = True) -> "ModelVersionSelection":
        """
        The latest scheduled model version. Orphaned resources are additionally taken at the last version they were
        present in when `include_orphans` is True; when False, only resources present in the latest version are selected.
        """
        return ModelVersionSelection(
            version=latest_scheduled_model_version(environment),
            include_orphans=include_orphans,
        )
```

arnaudsjs (approved):

src/inmanta/graphql/schema.py
bartv: Static method will also break inheritance. A class method bind to the correct class:

```python
class Test:
   @classmethod
   def create(cls):
      return cls()
```

PR #10531, opened 2026-07-03
Allow a break-glass database user under auth_method=oidc

bartv:
_Claude, opening this on behalf of @bartv._

## What
`inmanta-initial-user-setup` now accepts `auth_method` `oidc` and `jwt`, not only `database`, so a break-glass **database admin** can be provisioned while the server runs in OIDC mode. This is the backend prerequisite for the web-console database login fallback (an admin can then sign in locally when the identity provider is unavailable).

- The signing-config check is kept (login needs an HS256 issuer to mint a token).
- Unknown `auth_method` values are rejected with a helpful error.
- When `auth_method != database`, the tool prints a note that it is creating a break-glass account and points at `web-ui.oidc_local_fallback`.

## Tests
- `tests/test_usersetup.py`: added `test_user_setup_oidc` (creates an admin under `auth_method=oidc`) and `test_user_setup_invalid_auth_method`; existing tests still pass.
- `make mypy` clean, `flake8`/`black`/`isort` clean.

## Related
Part of the OIDC database-login fallback feature:
- inmanta-ui: emits the `localFallback` flag in `config.js`
- web-console: the fallback provider + user management under a fallback session

## Related PRs / MRs
- inmanta-core: https://github.com/inmanta/inmanta-core/pull/10531 (stacked on #10529 `db-singleton-lock`)
- inmanta-ui: https://github.com/inmanta/inmanta-ui/pull/822
- web-console: https://github.com/inmanta/web-console/pull/7069
- local-setup (test harness): https://code.inmanta.com/inmanta/local-setup/-/merge_requests/48

> Note: this PR targets `db-singleton-lock` (#10529), not master. GitHub will retarget it to master once #10529 merges.

### DIFF
--- a/tests/test_usersetup.py
+++ b/tests/test_usersetup.py
@@ -20,9 +20,11 @@ import asyncio
 import logging
 import os
 
+import pytest
 from click import testing
 
 from inmanta import data
+from inmanta.data.model import AuthMethod
 from inmanta.db.util import PGRestore
 from inmanta.server.bootloader import InmantaBootloader
 from inmanta.user_setup import cmd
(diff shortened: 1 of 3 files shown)


tests/test_usersetup.py
```diff
@@ -101,6 +101,45 @@ async def test_user_setup(
...
+    setup_config(tmpdir, postgres_db, database_name, auth_method="oidc")
+    cli = CLI_user_setup()
+
+    result = await cli.run("yes", "breakglass", "password")
+    assert result.exit_code == 0
+
+    users = await data.User.get_list(connection=postgresql_client)
+    assert len(users) == 1
+    assert users[0].username == "breakglass"
+    assert users[0].is_admin
```
jptrindade, line 127: we should also assert that `auth_method=="database"`

tests/test_usersetup.py
```diff
@@ -101,6 +101,45 @@ async def test_user_setup(
     assert users[0].is_admin
 
 
+async def test_user_setup_oidc(
```
jptrindade: can we also test this for jwt?

jptrindade (changes requested):

jptrindade:
NOTE: I assume that I was meant to ignore the database singleton changes

tests/test_usersetup.py
bartv: Claude (on behalf of @bartv): Done in 63da493ca - added `assert users[0].auth_method == AuthMethod.database`. The break-glass user is a local database account even though the server's `auth_method` is oidc/jwt, so this pins that down.

tests/test_usersetup.py
bartv: Claude (on behalf of @bartv): Done in 63da493ca - parametrized the test over `auth_method` `["oidc", "jwt"]` (renamed to `test_user_setup_break_glass`); both are supported and both pass. This also fixed the CI failure on this branch: the test used `"password"`, which #10537's policy (now on master) rejects, so it now uses a compliant password.

jptrindade (approved):

PR #10534, opened 2026-07-04
Give login sessions their own expiry, decoupled from service tokens

bartv:
Give web-console login sessions a bounded lifetime that is independent of the service-token expiry.

## Problem

A single option, `auth_jwt_<name>.expire`, governs the `exp` of every token signed with that config: web-console login sessions and agent/compiler service tokens alike. So an operator cannot give interactive sessions a sane lifetime without also expiring long-lived service tokens (and vice versa). Login session tokens are therefore effectively eternal in every deployment that keeps service tokens eternal (the default, `expire=0`).

## Change

- New `server.login_session_expire` option (seconds, default 3600 = one hour). The `/login` endpoint issues session tokens that expire after this many seconds, independently of `auth_jwt.expire`. Set it to `0` to fall back to the signing config's `expire` (eternal by default).
- `encode_token` now lets an explicit `expire` argument take precedence over the signing config's `expire`. This is what allows the login endpoint to override the service-token expiry. Callers that pass no `expire` (agents, compilers, idempotent API tokens) are unchanged.
- `encode_token` computes `iat` and `exp` from a single time source so a token's lifetime is exactly `expire` (previously two separate `time.time()` reads could make `exp - iat` off by one across a second boundary).

## Behavior changes on upgrade

- Login sessions now expire after 1 hour by default. Existing web-console database/break-glass sessions were eternal; after upgrade a new login is bounded to 1 hour (fixed at issuance, not sliding). When the session expires, a 401 sends the web-console back to the login screen. External OIDC sessions are unaffected (renewal is handled by the IdP). Set `login_session_expire=0` to keep eternal sessions.
- The precedence change also makes the `inmanta-cli token bootstrap` break-glass token honor its own 1-hour lifetime regardless of a configured `auth_jwt.expire`.

## Documentation

Updates `docs/administrators/authentication.rst`: the service-token vs login-session lifetime split, API-token attribution (`created_by`), login audit logging and credential redaction, and a Break-glass and recovery access section covering `inmanta-cli token bootstrap` and the web-console OIDC local login fallback.

## Tests

- `tests/test_jwt.py::test_encode_token_expire_precedence` - explicit `expire` wins over `cfg.expire`; no-`expire` falls back to `cfg.expire`; idempotent tokens never get an `exp`.
- `tests/server/test_user.py::test_login_session_expire` - the login token expires after the 1-hour default; `0` falls back to the (eternal) signing config; a custom value is honored.

## Notes for reviewers

- The `encode_token` precedence change is a no-op for all in-repo callers except login and bootstrap. `encode_token` is re-exported for extensions - worth a quick grep that no extension calls it with an explicit `expire` while relying on the old cfg.expire-wins behavior.
- The 1-hour default bounds all web-console database/break-glass sessions on upgrade; `0` opts out.

### DIFF
--- a/src/inmanta/protocol/auth/auth.py
+++ b/src/inmanta/protocol/auth/auth.py
@@ -65,7 +65,7 @@ def encode_token(
         payload.update(custom_claims)
 
     if not idempotent:
-        # Use a single time source so iat and exp are consistent and the token lifetime is exactly `expire`.
+        # Use a single time source so iat and exp are consistent and the token lifetime is exactly the requested expiry.
         now = int(time.time())
         payload["iat"] = now
 
--- a/src/inmanta/protocol/auth/auth.py
+++ b/src/inmanta/protocol/auth/auth.py
@@ -67,10 +67,13 @@ def encode_token(
     if not idempotent:
         payload["iat"] = int(time.time())
 
-        if cfg.expire > 0:
-            payload["exp"] = int(time.time() + cfg.expire)
-        elif expire is not None:
+        # An explicit expire argument takes precedence over the signing config's expire, so a caller
+        # (e.g. the login endpoint) can give its tokens a lifetime that is decoupled from the service
+        # token expiry configured on the signing config.
+        if expire is not None:
             payload["exp"] = int(time.time() + expire)
+        elif cfg.expire > 0:
+            payload["exp"] = int(time.time() + cfg.expire)
 
     if environment is not None:
         payload[const.INMANTA_URN + "env"] = environment
--- a/src/inmanta/protocol/auth/auth.py
+++ b/src/inmanta/protocol/auth/auth.py
@@ -49,6 +49,19 @@ def encode_token(
     expire: Optional[float] = None,
     custom_claims: Optional[Mapping[str, str | bool | Mapping[str, list[str]]]] = None,
 ) -> str:
+    """
+    Encode and sign a JWT bearer token for the given client types.
+
+    :param client_types: The client types to embed in the token. Each must be allowed by the signing configuration.
+    :param environment: When set, scope the token to this environment by adding an environment claim.
+    :param idempotent: When True, omit the time-based claims (iat and exp) so that encoding identical inputs always
+        produces an identical token. Used for tokens that must be stable across calls, such as environment tokens.
+    :param expire: The token lifetime in seconds. Takes precedence over the expiry from the signing configuration.
+        Ignored when idempotent is True.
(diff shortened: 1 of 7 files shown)


src/inmanta/protocol/auth/auth.py
Hugo-Inmanta, line 51: This method could use a proper docstring

src/inmanta/protocol/auth/auth.py
```diff
@@ -65,12 +65,17 @@ def encode_token(
         payload.update(custom_claims)
 
     if not idempotent:
-        payload["iat"] = int(time.time())
-
-        if cfg.expire > 0:
-            payload["exp"] = int(time.time() + cfg.expire)
-        elif expire is not None:
-            payload["exp"] = int(time.time() + expire)
+        # Use a single time source so iat and exp are consistent and the token lifetime is exactly the requested expiry.
```
Hugo-Inmanta:
```suggestion

```
Doesn't really add value outside of the context of this PR imo

tests/test_jwt.py
```diff
@@ -58,6 +58,25 @@ def test_jwt_create(inmanta_config):
...
+    # No explicit expire -> falls back to the signing config's expire.
+    service = jwt.decode(auth.encode_token(["agent"]), options={"verify_signature": False})
+    assert service["exp"] - service["iat"] == 7200
+
+    # An explicit expire wins over the signing config's expire.
+    session = jwt.decode(auth.encode_token(["api"], expire=3600), options={"verify_signature": False})
+    assert session["exp"] - session["iat"] == 3600
+
+    # An idempotent token never gets an expiry, regardless of either setting.
+    assert "exp" not in jwt.decode(auth.encode_token(["api"], idempotent=True), options={"verify_signature": False})
```
Hugo-Inmanta, line 77:
```suggestion
    assert "exp" not in jwt.decode(auth.encode_token(["api"], idempotent=True), options={"verify_signature": False})
    assert "exp" not in jwt.decode(auth.encode_token(["api"], idempotent=True, expire=3600), options={"verify_signature": False})

```

Hugo-Inmanta (changes requested):

src/inmanta/protocol/auth/auth.py
bartv: no, but without the testing becomes difficult. time.time() is also a very expensive function to call.

Hugo-Inmanta: I only meant the comment itself

Hugo-Inmanta (approved):

PR #10535, opened 2026-07-04
Add a revocable API-token registry (jti allowlist)

bartv:
Make user-created API tokens individually revocable and listable, without rotating the signing key.

> Stacked on #10533 (base branch `auth-token-attribution`) so the diff here is registry-only. Retarget to master once #10533 merges. Web-console counterpart: inmanta/web-console#7070.

## Problem

An API token minted by `environment_create_token` is an eternal, anonymous JWT. The only way to kill a leaked one is rotating the HS256 signing key, which invalidates every agent, compiler and user token at once. There is also no inventory of issued tokens.

## Change (additive, non-breaking)

- **`idempotent=False` tokens are registered and revocable.** They now carry a random `jti` claim and a row in a new `token` table (creator, client types, environment, issued/expiry, revoked, last-used). `idempotent=True` tokens (the API default) stay reproducible and stateless - no `jti`, not tracked - so existing automation is unaffected.
- **Enforcement:** `auth.validate_jti()` runs on the request path (after signature verification, before authorization) and rejects any `jti`-bearing token whose `jti` is absent from the registry or revoked. It acts as an allowlist, so even a forged `jti` is rejected. Tokens without a `jti` (agents, compilers, legacy tokens) keep stateless signature-only validation. Backed by a short-lived in-process cache, invalidated on revoke (correct under the active/standby HA topology).
- **Endpoints:** `environment_token_list` (GET) and `environment_token_revoke` (DELETE `/environment_auth/<jti>`), env-scoped, cross-environment revoke is refused.
- **Housekeeping:** a scheduled job prunes expired rows and stale cache entries.

Migration `v202607040` creates the table (+ regenerated dump + migration test).

## Security model

Signature is verified first (unchanged); the registry is a second allowlist layer for tokens that opt into it via a `jti`. Because new tokens require a registry row, a leaked signing key cannot mint accepted `jti` tokens out of thin air. Legacy tokens remain signature-only (that is the cost of a non-breaking migration).

## Known limitations

- Revocation takes effect on the next request; an already-open long-lived session (e.g. a websocket) is not force-closed - it is re-checked when it reconnects. Same as JWT expiry today.
- The registry holds every valid revocable token; unrevoked non-expiring tokens are kept by design (use idempotent tokens or an expiry to avoid per-token rows).

## Tests

`tests/server/test_user.py::test_token_registry` (mint→list→revoke→rejected, legacy passthrough, unknown-jti 404, cross-env revoke 404, non-UUID `jti` 403) and `test_token_registry_cleanup`. Validated end-to-end against a real orchestrator container.

## Changelog

`changelogs/unreleased/token-registry.yml` (minor).


src/inmanta/data/__init__.py
```diff
@@ -6855,6 +6855,42 @@ class EnvironmentMetricsTimer(BaseDocument):
     __primary_key__ = ("environment", "metric_name", "category", "timestamp")
 
 
+class Token(BaseDocument):
+    """A registry entry for an issued, revocable authentication token (identified by its jti claim)."""
```
bartv: Add @attr documentation to each of the fields of this document.

bartv: Claude (on behalf of @bartv): Done in 9496f1886 - each field of the `Token` document now has a `:param` description, matching the convention used by other data classes (e.g. `Notification`).

src/inmanta/data/__init__.py
```diff
@@ -6855,6 +6855,53 @@ class EnvironmentMetricsTimer(BaseDocument):
...
+    :param expires_at: The moment the token expires, or None when the token does not expire.
+    :param revoked: Whether the token has been revoked. A revoked token is rejected on every request.
+    :param last_used: The last time the token was used to authenticate a request, or None if it has not been used yet.
+    """
+
+    __primary_key__ = ("jti",)
+
+    jti: uuid.UUID
+    created_by: Optional[str] = None
+    client_types: list[str] = []
```
Hugo-Inmanta: This should  be typed as a `list[ClientType]` (in `src/inmanta/const.py` )

src/inmanta/data/model.py
```diff
@@ -916,6 +916,19 @@ def serialize_roles(self, roles: dict[uuid.UUID, list[str]], _info: Serializatio
         return {str(k): v for k, v in roles.items()}
 
 
+class Token(BaseModel):
+    """A registered (revocable) authentication token, tracked in the token registry."""
+
+    jti: uuid.UUID
+    created_by: Optional[str] = None
+    client_types: list[str] = []
```
Hugo-Inmanta:
```suggestion
    client_types: list[ClientType] = []
```

tests/server/test_user.py
```diff
@@ -335,3 +338,115 @@ def test_verify_dummy_password_smoke() -> None:
...
     verify_dummy_password(SecretStr("any password"))
+
+
+async def test_token_registry(server: protocol.Server, auth_client: endpoints.Client, monkeypatch) -> None:
+    """
+    Non-idempotent tokens are registered in the token registry and can be listed and revoked; idempotent
+    tokens stay stateless (no jti, not tracked). A revoked token is rejected on subsequent requests.
+    """
+    config.Config.set("server", "auth", "false")
+    monkeypatch.setenv("INMANTA_SERVER_AUTH", "true")
```
Hugo-Inmanta:
```suggestion
```
The `server_pre_start` fixture already sets this to True L35.

src/inmanta/protocol/auth/auth.py
```diff
@@ -148,6 +153,68 @@ def decode_token(token: str) -> tuple[claim_type, "AuthJWTConfig"]:
...
+
+def invalidate_jti(jti: uuid.UUID | str) -> None:
+    """Drop a jti from the registry cache so the next request re-reads it from the database."""
+    _jti_cache.pop(str(jti), None)
+
+
+def prune_jti_cache() -> None:
+    """Evict cache entries whose TTL has elapsed, so the cache does not grow unbounded over time."""
+    now = time.monotonic()
+    for jti in [jti for jti, (expiry, _token) in list(_jti_cache.items()) if expiry <= now]:
```
Hugo-Inmanta:
```suggestion
    for jti in [jti for jti, (expiry, _token) in _jti_cache.items() if expiry <= now]:
```
Why create a list and not iterate directly over `items()`?

src/inmanta/protocol/auth/auth.py
```diff
@@ -148,6 +153,68 @@ def decode_token(token: str) -> tuple[claim_type, "AuthJWTConfig"]:
...
+    if cached is not None and cached[0] > now:
+        token = cached[1]
+    else:
+        try:
+            jti_uuid = uuid.UUID(jti)
+        except ValueError:
+            raise exceptions.Forbidden("The provided token has an invalid jti claim.")
+        # Lazy import: the data layer imports the protocol package, so importing it at module load
+        # time would create a circular import.
+        from inmanta import data
```
Hugo-Inmanta: I'm not 100% sure how I feel about this. In this case I think we could hoist it and it would work just fine, as long as we don't do `from inmanta.data import Token`. We could also remove `if TYPE_CHECKING: ... `.

src/inmanta/protocol/auth/auth.py
```diff
@@ -148,6 +153,68 @@ def decode_token(token: str) -> tuple[claim_type, "AuthJWTConfig"]:
...
+
+        token = await data.Token.get_one(jti=jti_uuid)
+        _jti_cache[jti] = (now + JTI_CACHE_TTL, token)
```
Hugo-Inmanta: Do we want to cache the token if it is None?

Hugo-Inmanta (changes requested):

src/inmanta/data/__init__.py
```diff
@@ -6855,6 +6855,53 @@ class EnvironmentMetricsTimer(BaseDocument):
     __primary_key__ = ("environment", "metric_name", "category", "timestamp")
 
 
+class Token(BaseDocument):
```
Hugo-Inmanta: I think we also need to mirror this class in `src/inmanta/data/sqlalchemy.py`.

Hugo-Inmanta (changes requested):

src/inmanta/data/__init__.py
bartv: What is the procedure for that? I did not add it there because we are not (yet) exposing it over graphql

src/inmanta/data/__init__.py
bartv: Claude (on behalf of @bartv): Done in 28718e49f. The `Token` document field is now `client_types: list[const.ClientType]`, and the DB round-trips cleanly: on write each enum is serialized to its `.value` (into the `varchar[]` column), and on read the column strings are converted back to `ClientType`. The incoming strings at the token creation site (`environmentservice.py`) are converted to the enum before the `Token(...)` is built, since construction validates the element type.

src/inmanta/data/model.py
bartv: Claude (on behalf of @bartv): Done in 28718e49f - the DAO field is now `client_types: list[const.ClientType]`.

tests/server/test_user.py
bartv: Claude (on behalf of @bartv): Done in 28718e49f - removed the `server/auth=false` + `INMANTA_SERVER_AUTH=true` lines (and the now-unused `monkeypatch` parameter); `server_pre_start` already enables auth. That env-var dance was copied from `test_environment_create_token`, where it is deliberate (the #8962 reproduction); it served no purpose here.

src/inmanta/protocol/auth/auth.py
bartv: Claude (on behalf of @bartv): Done in 28718e49f. You are right - the `list()` was redundant: the comprehension already materializes the list of keys before the `for` loop starts mutating the dict, so iterating `_jti_cache.items()` directly is safe.

src/inmanta/protocol/auth/auth.py
bartv: Claude (on behalf of @bartv): I looked into hoisting it, but the circular import is real and hoisting `from inmanta import data` to module level recreates it, even without `from inmanta.data import Token`:

- `inmanta/protocol/__init__.py:57` does `from inmanta.protocol.auth.auth import decode_token, encode_token`.
- So if `auth.auth` imports `data` at module load, that pulls in `inmanta.data`, whose `__init__` does `import inmanta.protocol`, which re-enters `auth.auth` while it is still initializing.

Importing `auth.auth` before `data`/`protocol` then leaves it partially initialized. The lazy import keeps `auth.auth` free of any module-load-time coupling into the `data`↔`protocol` cycle, so I kept it (and the `if TYPE_CHECKING` block for the annotation). Happy to revisit if you'd still prefer the hoist.

src/inmanta/protocol/auth/auth.py
bartv: Claude (on behalf of @bartv): Good catch - changed in 28718e49f to only cache tokens that exist. Caching `None` let a flood of unique (forged) jtis grow the cache within the TTL window for no benefit - the cache exists to speed up the valid-token path. Revoked tokens still live in the registry with `revoked=True`, so they are still cached and rejected via `token.revoked`; only truly-unknown jtis now skip the cache and re-hit the DB.

src/inmanta/data/__init__.py
```diff
@@ -6855,6 +6855,53 @@ class EnvironmentMetricsTimer(BaseDocument):
...
+    :param revoked: Whether the token has been revoked. A revoked token is rejected on every request.
+    :param last_used: The last time the token was used to authenticate a request, or None if it has not been used yet.
+    """
+
+    __primary_key__ = ("jti",)
+
+    jti: uuid.UUID
+    created_by: Optional[str] = None
+    client_types: list[const.ClientType] = []
+    environment: Optional[uuid.UUID] = None
```
arnaudsjs: Why is None a valid value for environment? As far as I can see we always require an environment, by maybe I am missing something.

bartv: We support it in encode token, probably not through the API

src/inmanta/data/__init__.py
bartv: Claude (on behalf of @bartv): Added in c0613771f, per the Slack discussion (policy is to double-implement for now, until the SQLAlchemy models are auto-generated). The `token` table is now mirrored in `data/sqlalchemy.py` (columns + primary key + the `environment` FK). It is intentionally not exposed over GraphQL - there is no `mapper.type(models.Token)` - so this only mirrors the schema. Verified: `configure_mappers()` succeeds, the GraphQL schema still builds without a `Token` type, black/flake8/mypy clean.

src/inmanta/protocol/auth/auth.py
Hugo-Inmanta: Looks like my claude disagrees with your claude:
```
Settled empirically, on the current committed code, in isolated interpreters. Same harness, same Python — only the import form differs:

┌────────────────────────────────────────────┬────────────────┬────────────────┬────────────┬────────────┬─────────────────────────────────┐
│  Import form (module-level in auth.auth)   │ protocol-first │   data-first   │ auth-first │ bootloader │      data.Token reachable       │
├────────────────────────────────────────────┼────────────────┼────────────────┼────────────┼────────────┼─────────────────────────────────┤
│ from inmanta import data (module form)     │     ✅ OK      │     ✅ OK      │   ✅ OK    │   ✅ OK    │ ✅ <class 'inmanta.data.Token'> │
├────────────────────────────────────────────┼────────────────┼────────────────┼────────────┼────────────┼─────────────────────────────────┤
│ from inmanta.data import Token (name form) │       —        │ ❌ ImportError │     —      │     —      │                —                │
└────────────────────────────────────────────┴────────────────┴────────────────┴────────────┴────────────┴─────────────────────────────────┘

The name-form control fails with exactly cannot import name 'Token' from partially initialized module 'inmanta.data' — which proves the harness genuinely detects the cycle when one exists. The module form passes every entry order anyway. So the claim "even without from inmanta.data import Token things would break" is empirically false here.

Why the colleague's mechanism doesn't hold

Their chain: auth.auth imports data → data/__init__ does import inmanta.protocol → re-enters auth.auth while it's initializing → leaves it partial. Two specific steps are wrong:

1. import inmanta.protocol does not re-enter auth.auth. In the only order where data/__init__:51 actually triggers protocol/__init__ (data-first), protocol/__init__:57 imports auth.auth, and auth.auth then runs from inmanta import data — at which point inmanta.data is already in sys.modules (partial). from inmanta import data binds that existing partial module object and returns immediately. It does not re-run anything. (In protocol-first/auth-first order, import inmanta.protocol inside data/__init__ hits a module already in sys.modules and is a pure no-op re-bind.)
2. auth.auth is not left "partially initialized" — because its module body never raises. It would only be left partial if a load-time statement threw. The module form dereferences nothing (data.Token is only touched inside validate_jti, at call time), so the body runs to completion. The name form throws because it dereferences the Token attribute at load time, and on the partial data module that attribute doesn't exist yet.

The crux: from inmanta import data binds a module object; from inmanta.data import Token reads an attribute off it. A partially-initialized module still has a stable, valid object identity that gets fully populated as its __init__ finishes — so binding it early is harmless as long as you don't read attributes until runtime. The colleague is conflating "this module is on a cycle" (true) with "binding it at load time fails" (false — only an attribute read at load time fails).

Where their instinct is still worth respecting

They're right that the cycle is real and right to be cautious. The module-form hoist works today, but it's one careless edit away from breaking: change it to from inmanta.data import Token, or add any module-level data.X reference, and you get that ImportError back — in a confusing, import-order-dependent way. That fragility is a legitimate reason to keep the lazy import (or, if hoisting, to leave a comment naming the real trap). So: fine decision, wrong justification. If you want, paste them the table + the two-point explanation above — it's reproducible with the exact script I just ran.
```

src/inmanta/data/sqlalchemy.py
```diff
@@ -430,6 +430,39 @@ class InmantaUser(Base):
...
+        ForeignKeyConstraint(["environment"], ["environment.id"], ondelete="CASCADE", name="token_environment_fkey"),
+        PrimaryKeyConstraint("jti", name="token_pkey"),
+    )
+
+    jti: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True, doc="The unique identifier of the token (its jti claim)")
+    created_by: Mapped[Optional[str]] = mapped_column(
+        String, doc="The user that created the token, or None when it was created by a service token"
+    )
+    client_types: Mapped[list[str]] = mapped_column(
+        ARRAY(String()),
```
Hugo-Inmanta: This could probably be tightened into the actual client types api / agent / compiler

bartv: I disagree, making the db schema to strict is going to bite us in the future

bartv: I decided to go 100% sqlalchemy and basedocument did check this. So now this has also been added to the sqlalchemy object.

changelogs/unreleased/token-registry.yml
```diff
@@ -0,0 +1,11 @@
+description: >-
```
Hugo-Inmanta, line 1: This won't be visible in the docs changelog

src/inmanta/data/sqlalchemy.py
```diff
@@ -430,6 +484,111 @@ class InmantaUser(Base):
...
+        """Mark the token with the given jti as revoked. The caller commits."""
+        await self.session.execute(update(Token).where(Token.jti == jti).values(revoked=True))
+
+    async def mark_used(self, jti: uuid.UUID, when: datetime.datetime) -> None:
+        """Record that the token with the given jti was just used to authenticate. The caller commits."""
+        await self.session.execute(update(Token).where(Token.jti == jti).values(last_used=when))
+
+    async def delete_expired(self) -> None:
+        """Delete registry entries for tokens that have already expired. The caller commits."""
+        await self.session.execute(delete(Token).where(Token.expires_at.is_not(None), Token.expires_at < func.now()))
```
Hugo-Inmanta, line 600: I don't know what func.now() is exaclty, are we sure it's ok to use it to check against the Token.expires_at that was created with `datetime.datetime.now(tz=datetime.timezone.utc)` ?

src/inmanta/data/sqlalchemy.py
```diff
@@ -60,6 +69,51 @@ async def delete_all(cls, environment: uuid.UUID, connection: asyncpg.connection
...
+        mapper = class_mapper(cls)
+        for column_property in mapper.column_attrs:
+            column = column_property.columns[0]
+            try:
+                python_type = column.type.python_type
+            except NotImplementedError:
+                continue
+            event.listen(
+                getattr(cls, column_property.key),
+                "set",
```
jptrindade, line 116: Not sure what this does, can we add a docstring here?

src/inmanta/data/sqlalchemy.py
bartv: I am pretty sure this translates to the NOW() function in sql

src/inmanta/data/sqlalchemy.py
bartv: Claude (on behalf of @bartv): Added a docstring in 76b1dc560. `__declare_last__` is a SQLAlchemy declarative hook that runs once after the mapper is configured; it registers a write-time validator for every mapped column so each value is type-checked on assignment against that column's declared type and nullability.

changelogs/unreleased/token-registry.yml
bartv: Claude (on behalf of @bartv): Added a `sections: feature` block in 76b1dc560 so the entry appears in the docs changelog.

src/inmanta/protocol/auth/auth.py
bartv: Claude (on behalf of @bartv): Your analysis is correct and mine was wrong on the mechanism - binding the module (`from inmanta import data`) is harmless; only a load-time attribute read (`from inmanta.data import Token` or a module-level `data.X`) fails on the cycle. Given your own conclusion that keeping the lazy import is a fine call for the fragility reason ("one careless edit away from breaking"), I've kept it lazy but corrected the comment in 76b1dc560 to name the real trap instead of the wrong justification.

src/inmanta/data/sqlalchemy.py
bartv: Claude (on behalf of @bartv): Confirmed - `func.now()` compiles to SQL `now()`, evaluated by Postgres. `expires_at` is a `timestamp with time zone` column, so the comparison is timezone-aware regardless of how the Python value was constructed. This also matches the previous BaseDocument `clean_up_tokens`, which used a raw `WHERE expires_at < now()`.

src/inmanta/protocol/auth/auth.py
bartv: Claude (on behalf of @bartv): @HugoT since you've shown both forms work, your call on which to land. Options:

1. Keep the lazy import (current) - avoids the fragility you flagged (a future module-level `data.X` / `from inmanta.data import Token` reintroduces the import-order ImportError), at the cost of an in-function import.
2. Hoist `from inmanta import data` to module level - cleaner, proven safe today, but one careless edit away from the cycle.

I lean towards (1) for the safety margin, but I'm happy to switch to (2) if you'd rather. Let me know and I'll apply it.

Hugo-Inmanta (approved):

src/inmanta/data/sqlalchemy.py
jptrindade: And does this work for creates and updates?

bartv: It runs this when a set is done on an instance of the sqlalchemy object

src/inmanta/protocol/auth/auth.py
```diff
@@ -165,6 +170,81 @@ def decode_token(token: str) -> tuple[claim_type, "AuthJWTConfig"]:
...
+                    await session.commit()
+                except Exception:
+                    LOGGER.debug("Failed to update last_used for token %s", jti, exc_info=True)
+        if token is not None:
+            _jti_cache[jti] = (now + JTI_CACHE_TTL, token)
+
+    if token is None:
+        raise exceptions.Forbidden("The provided token is not recognized; it may have been revoked.")
+    if token.revoked:
+        raise exceptions.Forbidden("The provided token has been revoked.")
```
jptrindade: This logic is a bit weird. Why not something like:

```suggestion
        async with data.get_session() as session:
            repo = TokenRepository(session)
            entry = await repo.get_by_jti(jti_uuid)
            if entry is not None:
                # Best-effort last-used tracking, bounded to at most once per cache TTL per token.
                try:
                    await repo.mark_used(jti_uuid, datetime.datetime.now(tz=datetime.timezone.utc))
                    await session.commit()
                except Exception:
                    LOGGER.debug("Failed to update last_used for token %s", jti, exc_info=True)
                 # Snapshot to a DTO while the entity's attributes are still loaded (before the session closes).
                  token = entry.to_dto()
                  if token.revoked:
                     raise exceptions.Forbidden("The provided token has been revoked.")
                  _jti_cache[jti] = (now + JTI_CACHE_TTL, token)
            else:
               raise exceptions.Forbidden("The provided token is not recognized; it may have been revoked.")
```
And on the last exception is it really possible that the token is revoked? Won't it just appear on the exception above?

src/inmanta/data/sqlalchemy.py
bartv: Claude (on behalf of @bartv): It fires on attribute assignment, so on creates (`Token(...)`) and on ORM attribute updates (`token.attr = x; commit`). It does not fire on Core `update()` statements, which is what `TokenRepository.revoke()` / `mark_used()` use - but those only ever write server-controlled scalars (`revoked=True`, a timestamp), so there is no external value to guard there.

This matches what we already have with `BaseDocument`: a targeted update (especially with raw SQL) doesn't runtime-validate the types there either. So it's the same trade-off, not a regression.

src/inmanta/protocol/auth/auth.py
bartv: Claude (on behalf of @bartv): On the revoked question - yes, it's reachable. `revoke()` sets `revoked=True` and keeps the row, so `get_by_jti` returns the token (not `None`). The first raise is for a jti that is absent from the registry entirely (unknown/forged, or already cleaned up); the second is for a present-but-revoked token. Different cases.

On the restructure - I kept the two checks after the `if/else` on purpose. Moving them inside the `async with get_session()` block only covers the cache-miss path; on a cache hit the token comes from `_jti_cache` and that block is skipped, so a revoked-but-cached token would be accepted for up to the 60s TTL - a security regression. I added a comment in 70ffab0b8 explaining that the two checks intentionally cover both the cache-hit and the freshly-fetched paths.

jptrindade: On what code flow is a revoked token still in the cache?

And on what code flow would we cache a revoked token?

src/inmanta/data/sqlalchemy.py
jptrindade: I see, I'm a bit reluctant about the name only. It makes it seem that the mixin is also protecting against updates. Maybe `SetValidatedMixin` might be a better name

src/inmanta/server/services/environmentservice.py
```diff
@@ -547,8 +560,65 @@ async def environment_create_token(
...
+
+    @handle(methods_v2.environment_token_revoke, env="tid")
+    async def environment_token_revoke(self, env: data.Environment, jti: uuid.UUID) -> None:
+        """
+        Revoke a registered token by its jti. The token is rejected on subsequent requests.
+        """
+        async with data.get_session() as session:
+            repo = TokenRepository(session)
+            token = await repo.get_by_jti(jti)
+            if token is None or token.environment != env.id:
```
jptrindade: the `token.environment != env.id` is a bit uncharacteristic. I would expect we could pass the environment to `get_by_jti` or another method and just confirm that the token is not none

src/inmanta/protocol/auth/auth.py
bartv: Claude (on behalf of @bartv): You're right - caching a revoked token was unnecessary, and that was the awkward part. Reworked in c17904a7c: `validate_jti` now caches **only valid** tokens (the cache is `dict[str, float]`, jti -> expiry). Unknown and revoked jtis are no longer cached - they're rejected on the fetch path - and revocation still evicts via `invalidate_jti`. So a cache hit is always a known-valid token, and the None/revoked checks now live only where the database is queried. That also resolves the "logic is a bit weird" you flagged: the checks are no longer duplicated across a cache/DB split.

src/inmanta/data/sqlalchemy.py
bartv: Claude (on behalf of @bartv): Agreed - renamed to `SetValidatedMixin` in c17904a7c. It validates on attribute set and does not cover Core `update()` statements, so the name shouldn't imply it guards every write.

src/inmanta/server/services/environmentservice.py
bartv: Claude (on behalf of @bartv): Done in c17904a7c - the revoke is now env-scoped in the query: `UPDATE token ... WHERE jti = :jti AND environment = :env ... RETURNING jti`, and the handler raises NotFound when nothing was returned. No more fetch-then-compare `token.environment`.

jptrindade (approved):

bartv:
Claude here, on behalf of Bart. Closing this PR: it merged to master as 5a46bc243, and the iso9 backport (which the merge tool could not do automatically due to the master/iso9 REST divergence) is now #10554. Leaving the source branch as it is still checked out in a local worktree.

PR #10537, opened 2026-07-04
Harden the built-in database login

bartv:
Four small hardening fixes to the built-in database authentication.

> Stacked on #10532 (base branch `auth-login-hardening`) so the diff is login-hardening-only. Retarget to master once #10532 merges. Web-console counterpart: inmanta/web-console#7071.

## Changes

- **Username-enumeration resistance.** A login for an unknown user now still runs an Argon2 verification against a constant dummy hash before returning 401, so response time no longer reveals whether an account exists. The same parity is applied in `set_password` for accounts without a local password (e.g. OIDC-only).
- **Malformed stored hash -> 401.** A stored password hash that cannot be parsed is treated as an authentication failure instead of raising a 500.
- **Maximum password length.** Passwords are capped at 128 characters (`MAX_PASSWORD_LENGTH`), enforced on `add_user` and `set_password`, as a guard against denial-of-service through very expensive hashing.
- **Require the current password on self-service change.** `set_password` gains an optional `current_password`; when a user changes their **own** password it is required and verified, so a hijacked session cannot silently take over the account. An administrator changing another user's password is unaffected. The authz boundary is unchanged (the OPA policy already only lets a user reach this handler for their own account or as a global admin). Password-change attempts are now audit-logged, mirroring login.

## Trade-off to sign off on

The dummy-hash verification means **every** login attempt (known or unknown user) now pays the Argon2 cost, which runs synchronously on the event loop. This widens the pre-auth CPU-DoS surface from "requires a known username" to "any request". This is a conscious trade-off for timing-safety and is consistent with the decision to handle brute-force protection / rate limiting at the network edge rather than in the application (documented in #10536). Offloading password hashing to an executor is a reasonable follow-up.

## Tests

`tests/server/test_user.py`: max-length rejection (add_user + set_password), malformed/oversized stored hash -> 401 (exercising both the InvalidkeyError and the ValueError branch), self-service change requires the current password (missing -> 400, wrong -> 401, correct -> 200, admin reset needs none) with audit-log assertions, and a dummy-hash smoke test. flake8 / black / mypy clean.

## Changelog

`changelogs/unreleased/harden-database-login.yml` (minor).

### DIFF
--- a/src/inmanta/const.py
+++ b/src/inmanta/const.py
@@ -342,37 +342,6 @@ MAX_PASSWORD_LENGTH = 128
 MIN_PASSWORD_CHARACTER_CLASSES = 3
 
 
-def password_policy_violation(password: str) -> Optional[str]:
-    """
-    Return a human-readable reason the password violates the password policy, or None if it is acceptable.
-
-    The policy is a length range plus a basic complexity requirement (a mix of character classes). It is
-    shared by the API (add_user, set_password) and the inmanta-initial-user-setup CLI so both enforce the
-    same rules. Note this is a deliberately minimal, self-contained check: richer credential controls
-    (breached-password lists, history, rotation, MFA) are out of scope for the built-in provider and are
-    expected from an external OIDC identity provider instead.
-    """
-    if not password or len(password) < MIN_PASSWORD_LENGTH:
-        return f"the password should be at least {MIN_PASSWORD_LENGTH} characters long"
-    if len(password) > MAX_PASSWORD_LENGTH:
-        return f"the password should be at most {MAX_PASSWORD_LENGTH} characters long"
-    # Count the character classes used. "special" is anything that is not a letter or a digit.
-    classes = sum(
-        [
-            any(c.islower() for c in password),
-            any(c.isupper() for c in password),
-            any(c.isdigit() for c in password),
-            any(not c.isalnum() for c in password),
-        ]
-    )
-    if classes < MIN_PASSWORD_CHARACTER_CLASSES:
-        return (
-            f"the password should contain at least {MIN_PASSWORD_CHARACTER_CLASSES} of the following four "
-            "character classes: a lowercase letter, an uppercase letter, a digit, and a special character"
-        )
-    return None
-
-
 class AgentAction(str, Enum):
     pause = "pause"
     unpause = "unpause"
--- a/src/inmanta/const.py
+++ b/src/inmanta/const.py
@@ -336,6 +336,8 @@ ENVELOPE_KEY = "data"
 
 # Minimum password length
 MIN_PASSWORD_LENGTH = 8
+# Maximum password length, a guard against denial-of-service through very expensive password hashing
+MAX_PASSWORD_LENGTH = 128
 
 
 class AgentAction(str, Enum):
--- a/src/inmanta/const.py
+++ b/src/inmanta/const.py
@@ -335,9 +335,42 @@ EXTENSION_MODULE = "extension"
 ENVELOPE_KEY = "data"
 
 # Minimum password length
-MIN_PASSWORD_LENGTH = 8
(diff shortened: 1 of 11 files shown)


src/inmanta/protocol/methods_v2.py
```diff
@@ -1612,21 +1612,26 @@ def add_user(username: str, password: SecretStr) -> model.User:
     """Add a new user to the system
 
     :param username: The username of the new user. The username cannot be an empty string.
-    :param password: The password of this new user. The password should be at least 8 characters long.
+    :param password: The password of this new user. It should be between 8 and 128 characters long.
```
Hugo-Inmanta:
```suggestion
    :param password: The password of this new user. It should be between 12 and 128 characters long.
```

src/inmanta/protocol/methods_v2.py
```diff
@@ -1612,21 +1612,26 @@ def add_user(username: str, password: SecretStr) -> model.User:
...
 
 @auth(auth_label=const.CoreAuthorizationLabel.USER_CHANGE_PASSWORD, read_only=False)
 @typedmethod(path="/user/<username>/password", operation="PATCH", client_types=[ClientType.api], api_version=2)
-def set_password(username: str, password: SecretStr) -> None:
+def set_password(username: str, password: SecretStr, current_password: Optional[SecretStr] = None) -> None:
     """Change the password of a user
 
     :param username: The username of the user
-    :param password: The password of this new user. The password should be at least 8 characters long.
+    :param password: The new password. It should be between 8 and 128 characters long.
```
Hugo-Inmanta:
```suggestion
    :param password: The new password. It should be between 12 and 128 characters long.
```

src/inmanta/const.py
```diff
@@ -335,7 +335,42 @@ class ClientType(str, Enum):
...
 # Minimum password length
-MIN_PASSWORD_LENGTH = 8
+MIN_PASSWORD_LENGTH = 12
+# Maximum password length, a guard against denial-of-service through very expensive password hashing
+MAX_PASSWORD_LENGTH = 128
+# The number of distinct character classes (lowercase, uppercase, digit, special) a password must use
+MIN_PASSWORD_CHARACTER_CLASSES = 3
+
+
+def password_policy_violation(password: str) -> Optional[str]:
```
Hugo-Inmanta: Maybe move this to `src/inmanta/util/__init__.py`

Hugo-Inmanta (changes requested):

src/inmanta/protocol/methods_v2.py
bartv: Applied in 7c2c2de58.

src/inmanta/protocol/methods_v2.py
bartv: Applied in 7c2c2de58.

src/inmanta/const.py
bartv: Moved `password_policy_violation` to `src/inmanta/util/__init__.py` in 7c2c2de58, and updated the two call sites. I kept the policy constants (`MIN_PASSWORD_LENGTH`, `MAX_PASSWORD_LENGTH`, `MIN_PASSWORD_CHARACTER_CLASSES`) in `const.py` since that is the constants module; the function references them as `const.X`. Happy to move those over too if you would rather have the whole policy in one place.

Hugo-Inmanta (approved):

bartv:
Claude here, on behalf of Bart. Closing this PR: it merged to master as 522cb0a4c, and the iso9 backport is already in place via #10543 (commit 676b78ebd on iso9). The password constants, util.password_policy_violation and the methods_v2 docstrings are all present on iso9. The merge tool's auto-backport conflicted only because iso9 already contains the equivalent change, so no fresh cherry-pick is needed. Leaving the source branch since it is still checked out in a local worktree.

PR #10538, opened 2026-07-06
Normalize passwords to Unicode NFKC

bartv:
Match passwords in Unicode NFKC form so the same password verifies regardless of how it was typed.

> Stacked on #10537 (shares the login/password code path), which is itself stacked on #10532. Retarget down the stack as those merge.

## Why

The same password can be entered as different byte sequences depending on the operating system, keyboard, or input method - for example an accented character as a single code point (é, U+00E9) versus a base letter plus a combining accent (e + U+0301), or compatibility characters (ligatures, full-width forms). Because the hash is computed over the raw bytes, without normalization the exact same password typed on two machines can fail to verify. Normalizing to NFKC (recommended by NIST SP 800-63B) collapses these to one canonical form. For ASCII passwords it is a no-op.

## Change

- `normalize_password` (NFKC) is applied on `add_user`, `set_password`, and `login`, with a docstring explaining the reasoning.
- `verify_password` verifies the normalized form first and, only if that fails and the input changed under normalization, falls back to the raw form - so hashes stored before this change (from raw input) keep working. A raw-form match on `login` transparently re-hashes the stored password to the normalized form, migrating the account on next login. The self-service current-password check uses the same fallback.
- The raw input is length-guarded (`MAX_RAW_PASSWORD_LENGTH`) before it ever reaches `unicodedata.normalize`, so an unauthenticated login attempt cannot force an unbounded normalization (pre-auth CPU/memory DoS).
- The login re-hash is isolated in its own try/except: a failure to migrate the stored hash is logged but never turns a successful login into an error.

## Tests

`test_password_nfkc`: a user created with the decomposed form logs in with either form; and a hash stored from raw (pre-normalization) bytes rejects the composed form, accepts the decomposed form via the fallback, and is then upgraded so the composed form works too. `test_login_oversized_password_rejected` covers the raw-length guard on the login path directly. flake8 / black / mypy clean.

## Changelog

`changelogs/unreleased/normalize-passwords-nfkc.yml` (minor).

### DIFF
--- a/changelogs/unreleased/normalize-passwords-nfkc.yml
+++ b/changelogs/unreleased/normalize-passwords-nfkc.yml
@@ -7,3 +7,8 @@ change-type: minor
 destination-branches:
   - master
   - iso9
+sections:
+  minor-improvement: >
+    Passwords are normalized to Unicode NFKC before hashing and verifying, so a password containing
+    non-ASCII characters verifies regardless of how it was typed. Existing password hashes keep working and
+    are transparently re-hashed to the normalized form on the next successful login.
--- a/changelogs/unreleased/normalize-passwords-nfkc.yml
+++ b/changelogs/unreleased/normalize-passwords-nfkc.yml
@@ -1,0 +1,9 @@
+description: >-
+  Normalize passwords to Unicode NFKC before hashing and verifying, so a password containing non-ASCII
+  characters verifies regardless of how it was typed (operating system, keyboard, or input method).
+  Password hashes stored before this change keep working and are transparently re-hashed to the
+  normalized form on the next successful login.
+change-type: minor
+destination-branches:
+  - master
+  - iso9
--- a/src/inmanta/server/services/userservice.py
+++ b/src/inmanta/server/services/userservice.py
@@ -17,6 +17,7 @@ Contact: code@inmanta.com
 """
 
 import logging
+import unicodedata
 import uuid
 from collections.abc import Mapping
 from typing import Optional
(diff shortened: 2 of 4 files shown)


changelogs/unreleased/normalize-passwords-nfkc.yml
```diff
@@ -0,0 +1,9 @@
+description: >-
+  Normalize passwords to Unicode NFKC before hashing and verifying, so a password containing non-ASCII
+  characters verifies regardless of how it was typed (operating system, keyboard, or input method).
+  Password hashes stored before this change keep working and are transparently re-hashed to the
+  normalized form on the next successful login.
```
Hugo-Inmanta, line 5: This will not show up in the user visible changelog, I assumed this was intentional for the previous PRs but now I have a doubt. Do we want these to be visible ? And/or do we want to add a single entry that sums up all the improvements at once ?

tests/server/test_user.py
```diff
@@ -306,3 +306,47 @@ def test_verify_dummy_password_smoke() -> None:
...
+    A login attempt whose input exceeds MAX_RAW_PASSWORD_LENGTH is rejected as invalid credentials without
+    ever running NFKC normalization on it. This guards the unauthenticated login path against a
+    normalization CPU cost on pathologically long input.
+    """
+    response = await auth_client.add_user("wendy", "Str0ng-Pass!")
+    assert response.code == 200
+
+    oversized = "a" * (const.MAX_RAW_PASSWORD_LENGTH + 1)
+    response = await auth_client.login("wendy", oversized)
+    assert response.code == 401
```
Hugo-Inmanta: i think this test doesn't really test that no normalization is run, we would get a 401 with any password != "Str0ng-Pass!"

src/inmanta/server/services/userservice.py
```diff
@@ -191,16 +257,26 @@ async def login(
...
-            raise exceptions.UnauthorizedException(message=invalid_username_password_msg, no_prefix=True)
+            matched, needs_rehash = verify_password(user.password_hash, password.get_secret_value())
         except (ValueError, nacl.exceptions.CryptoError):
-            # A stored hash that cannot be parsed is an authentication failure, not a server error. Most
-            # malformed hashes raise InvalidkeyError (a CryptoError subclass) and are handled above; this
-            # clause also covers a hash that is well-formed but too long, which raises a plain ValueError.
+            # A stored hash that cannot be parsed is an authentication failure, not a server error. (Most
+            # malformed hashes raise InvalidkeyError, a CryptoError subclass; a well-formed but too-long
+            # hash raises a plain ValueError. Both are handled here.)
             LOGGER.warning("Failed login for user '%s' from %s: malformed stored password hash", username, source_ip)
```
Hugo-Inmanta, line 265: > "Failed login for user '%s' from %s: malformed stored password hash" 

Should we tell the user to update his password in this case ?

Hugo-Inmanta (changes requested):

changelogs/unreleased/normalize-passwords-nfkc.yml
bartv: Why is this an issue in the changelog?

tests/server/test_user.py
bartv: This tests: "The raw input is length-guarded (MAX_RAW_PASSWORD_LENGTH) before it ever reaches unicodedata.normalize, so an unauthenticated login attempt cannot force an unbounded normalization (pre-auth CPU/memory DoS)."

src/inmanta/server/services/userservice.py
bartv: They can't because their password cannot be matched. It should not happen but just in case there is this log message. Because this is security related we also do not want to return too much information of what is going on.

Hugo-Inmanta: Right

changelogs/unreleased/normalize-passwords-nfkc.yml
Hugo-Inmanta: Without a "sections" section (e.g. as in the template `changelogs/template.yml` ) the change entry will not be displayed in the changelog of the docs (e.g. [here](https://docs.inmanta.com/inmanta-service-orchestrator/9/CHANGELOG.html))

bartv: ah, like that

tests/server/test_user.py
Hugo-Inmanta: The way the check is done, we only test that we get a 401 when attempting to login with the oversized pw, we don't actually check that the length guard is working as expected e.g it doesn't check that unicodedata.normalize is never called. I don't know if there's a simple/elegant way to test for that.

Hugo-Inmanta (commented):
(See comments)

tests/server/test_user.py
bartv: yeah, I know. We can also drop this test if you prefer. I really do not like tests that mock out almost everything to just test a certain concern.

Hugo-Inmanta: Yeah I think I prefer to not have it at all since it doesn't really test anything, e.g. if someone removes the guard entirely, the test would still pass

src/inmanta/server/services/userservice.py
```diff
@@ -191,16 +257,26 @@ async def login(
             LOGGER.warning("Failed login for user '%s' from %s: no such user", username, source_ip)
```
Hugo-Inmanta, line 257: I just realized, that the verify_dummy_password trick only works as long as the attacker doesn't have access to the logs since we log that there is no user, should we make the log message more generic just to be on the safe side ?

bartv: no, this is for external enumeration. Making the logs more anonymous will make operations more difficult for an absolute corner case.

changelogs/unreleased/normalize-passwords-nfkc.yml
bartv: Claude (on behalf of @bartv): Done in 9cae7dc3e - added a `sections:` block (`minor-improvement`) so the entry shows up in the docs changelog. (Whether to also fold the whole auth-hardening series into one summary entry is a separate call - happy to do that if you want, but left the per-PR entries for now.)

tests/server/test_user.py
bartv: Claude (on behalf of @bartv): Dropped `test_login_oversized_password_rejected` in 9cae7dc3e, as agreed - it only asserted a 401, which happens for any wrong password, so it didn't actually exercise the raw-length guard (it would still pass if the guard were removed). The `MAX_RAW_PASSWORD_LENGTH` guard in `login` stays.

Hugo-Inmanta (approved):

PR #10540, opened 2026-07-06
Add a login/renew endpoint to keep active sessions alive

bartv:
Let an interactive client keep an active login session alive instead of being logged out when the session expires.

> Stacked on #10534 (which gave login sessions their own expiry). Retarget to master once that merges. The web console will call this endpoint; that is a separate web-console PR.

## Why

#10534 gave login sessions a finite lifetime (default one hour). That fixed a real exposure problem, but it also means an actively-used web-console session is now silently dropped after an hour, forcing a re-login mid-work. The OIDC and Keycloak providers already renew their tokens silently through their client libraries; the built-in database provider had no equivalent, so this brings it to parity.

## Change

- New `POST /api/v2/login/renew`. It is authenticated by the caller's current, still-valid session token (no password), and returns a fresh token with a new expiry.
- The renewed token is rebuilt from the current database state, so it reflects the user's **current** roles and admin status rather than whatever was captured at login time.
- `LoginReturn` now carries `expires_in` (token lifetime in seconds, or `null` when the token does not expire), reported by both `login` and `login/renew`, so a client knows when to renew.
- Token issuance is factored into a shared `issue_session_token` helper so `login` and `login/renew` stay in lockstep. `login`'s behavior is otherwise unchanged.

## Security

- **Self-scoped**: the user is taken from the token's `sub` claim, so a caller can only ever renew their own session.
- **Own-issuer only**: only tokens minted by this server's own signing authority can be renewed. A token merely accepted by `decode_token` is not sufficient - in the break-glass topology (a local `sign=true` section alongside an external OIDC issuer) a user authenticated against the external issuer would otherwise be handed a fresh, locally-signed token carrying the roles and `is_admin` of a same-named local account. The endpoint checks the token's `iss` against the local signing issuer and rejects anything else. Service tokens (no `sub`) are rejected too.
- **Sliding session, by design**: an actively-used session can be renewed indefinitely while an idle one still lapses. This deliberately relaxes #10534's absolute one-hour cap for active sessions (standard interactive-session behavior); there is intentionally no absolute lifetime ceiling.

## Tests

`test_login_reports_expires_in`, `test_login_renew` (fresh token works), `test_login_renew_reflects_current_admin_status`, `test_login_renew_rejects_non_session_token` (no `sub`), and `test_login_renew_rejects_external_issuer_token` (a matching-username token from another issuer is rejected). flake8 / black clean; mypy baseline updated for the new endpoint stub.

## Changelog

`changelogs/unreleased/login-session-renew.yml` (minor).

### DIFF
--- a/changelogs/unreleased/login-session-expire.yml
+++ b/changelogs/unreleased/login-session-expire.yml
@@ -1,0 +1,10 @@
+description: >-
+  Add the server.login_session_expire option to give web-console login sessions their own lifetime,
+  decoupled from the auth_jwt `expire` that governs agent and compiler service tokens. An explicit
+  token expiry now takes precedence over the signing config's `expire` (which, as a side effect, makes
+  the `inmanta-cli token bootstrap` token honor its own 1-hour lifetime regardless of the configured
+  `expire`).
+change-type: minor
+destination-branches:
+  - master
+  - iso9
--- a/changelogs/unreleased/login-session-expire.yml
+++ b/changelogs/unreleased/login-session-expire.yml
@@ -1,9 +1,10 @@
 description: >-
-  Add the server.login_session_expire option to give web-console login sessions their own lifetime,
-  decoupled from the auth_jwt `expire` that governs agent and compiler service tokens. An explicit
-  token expiry now takes precedence over the signing config's `expire` (which, as a side effect, makes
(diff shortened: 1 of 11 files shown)


bartv:
Claude here, on behalf of Bart. This PR was auto-closed when its base branch `auth-login-session-expire` (#10534) was deleted on merge, and GitHub does not allow reopening it. The change has been rebased onto master and continues as #10546.

PR #10552, opened 2026-07-08
Verify the local project state at server startup and recompile when it is stale

bartv:
> [!NOTE]
> I am Claude, opening this PR on behalf of Bart Vanbrabant (@bartv).

# Description

The server keeps an environment's project checkout and compiler venv on its local filesystem, while compile reports live in the database. These can disagree: the server may have been promoted after a failover (see #10529), or its state directory may have been wiped or restored from a snapshot. Until now the first compile after such an event would lazily pay the full clone-and-install cost, and a triggered compile without `update` could even compile stale code from the old checkout.

This PR makes the server record which compile last ran against each project directory (a `.inmanta_last_compile` marker file containing the compile id, written after every compile) and verify it at startup against the latest compile in the database (following `substitute_compile_id` for merged compiles). On a mismatch it logs a warning and requests a compile with `force_update=True, do_export=True` (the same semantics as `notify?update=true`) so the local state converges to the project definition.

Design decisions:

- **"Correct local state" is defined as "produced by the environment's latest compile in the database"**, not "latest git HEAD". This needs no DB migration, covers the venv/module dimension, and makes a plain restart a guaranteed no-op.
- **The marker is written for failed compiles too**: the local state was produced by that attempt either way. Otherwise a compile that fails for reasons unrelated to disk state (a model error, missing git credentials) would be retried at every restart, forever.
- **Halted environments are skipped at startup and verified on resume** instead: a compile queued for a halted environment can never run, and such rows are excluded from report retention cleanup, so queueing at startup would leak one orphaned compile row per restart.
- **Environments without a `repo_url` are skipped**: the server cannot restore their state by itself, and a forced compile would just fail at every startup.
- **No config option**: the check only acts when local state is actually stale, and the per-environment `server_compile` setting already gates server compiles.

A bounded race is documented in the code: a pre-crash queued compile recovered at startup may run next to one redundant recovery compile; this self-corrects.

# Self Check:

- [ ] ~~Attached issue to pull request~~ (no ticket, direct request)
- [x] Changelog entry
- [ ] End user documentation is included or an issue is created for end-user documentation (to be covered together with the HA documentation follow-up of #10529)
- [ ] ~~If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s)~~

### DIFF
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -985,14 +985,95 @@ class CompilerService(ServerSlice, inmanta.server.services.environmentlistener.E
             self.fully_ready = True
             # start a waiting compile in each environment
             await self._process_next_compile_in_queue()
+            # A compile recovered above may itself converge stale local state, in which case the check below
+            # queues one redundant recovery compile behind it. This is bounded and self-corrects.
+            await self._recompile_environments_with_stale_local_state()
 
         self.add_background_task(sub_recovery())
+
+    def _get_project_dir(self, environment_id: uuid.UUID) -> str:
+        """Return the directory that holds the project checkout and compiler venv for the given environment."""
+        return os.path.join(self._server_state_dir, str(environment_id), "compiler")
+
+    def _write_last_compile_marker(self, environment_id: uuid.UUID, compile_id: uuid.UUID) -> None:
+        """
+        Record in the environment's project directory which compile last ran against it. The startup consistency
+        check compares this marker to the latest compile in the database.
+        """
+        project_dir = self._get_project_dir(environment_id)
+        if not os.path.isdir(project_dir):
+            # the compile never got to setting up the project directory
+            return
+        try:
+            with open(os.path.join(project_dir, const.INMANTA_LAST_COMPILE_MARKER), "w") as fh:
+                fh.write(str(compile_id))
+        except OSError:
+            LOGGER.warning("Failed to write the last-compile marker for environment %s", environment_id, exc_info=True)
+
+    async def _recompile_environments_with_stale_local_state(self) -> None:
+        """
+        Verify the local project state of every environment and request an update and recompile where it is stale.
+
+        Halted environments are skipped: a compile requested for them cannot run until they are resumed, so it would
+        only accumulate queued compiles across restarts (which the report retention cleanup does not touch for halted
+        environments). They are verified when they are resumed instead.
+        """
+        for env in await data.Environment.get_list():
+            if env.halted:
+                continue
+            await self._recompile_environment_with_stale_local_state(env)
+
+    async def _recompile_environment_with_stale_local_state(self, env: data.Environment) -> None:
+        """
+        The project checkout and compiler venv live on the server's local filesystem, while compile reports live in
+        the database. These can disagree: this server may have been promoted after a failover, or its state directory
+        may have been wiped or restored from a snapshot. In that case the project state on disk is not the state that
+        produced the environment's latest compile. Detect this by comparing the last-compile marker on disk to the
+        latest compile in the database, and request an update and recompile to converge the local state.
+        """
+        if not env.repo_url:
+            # without a repository the server cannot restore the local state by itself
+            return
+        last_compile = await data.Compile.get_last_run(env.id)
+        if last_compile is None:
+            # this environment never completed a compile: there is no local state to restore
+            return
+        # a merged compile records the compile that actually ran as its substitute
+        expected_id = last_compile.substitute_compile_id or last_compile.id
+        marker_path = os.path.join(self._get_project_dir(env.id), const.INMANTA_LAST_COMPILE_MARKER)
+        try:
+            with open(marker_path) as fh:
+                if fh.read().strip() == str(expected_id):
+                    return
+        except OSError:
+            pass
+        LOGGER.warning(
+            "The local project state of environment %s was not produced by compile %s, the latest compile"
+            " in the database (e.g. after a failover to another server or a wiped state directory)."
+            " Requesting an update and recompile to converge.",
+            env.id,
+            expected_id,
+        )
+        await self.request_recompile(
+            env,
+            force_update=True,
+            do_export=True,
+            remote_id=uuid.uuid4(),
+            metadata={
+                "type": "recovery",
+                "message": "Recompile because the local project state does not match the latest compile",
+            },
+        )
 
     async def resume_environment(self, environment: uuid.UUID) -> None:
         """
         Resume compiler service after halt.
         """
         await self._process_next_compile_in_queue(environment=environment)
+        env: Optional[data.Environment] = await data.Environment.get_by_id(environment)
+        if env is not None:
+            # halted environments are not verified at startup: verify the local project state now
+            await self._recompile_environment_with_stale_local_state(env)
 
     def is_environment_compiling(self, environment_id: uuid.UUID) -> bool:
         return environment_id in self._env_to_compile_task
@@ -1065,9 +1146,7 @@ class CompilerService(ServerSlice, inmanta.server.services.environmentlistener.E
             if not c.id == compile.id and CompilerService._compile_merge_key(c) == compile_merge_key
         ]
 
-        runner = self._get_compile_runner(
-            compile, project_dir=os.path.join(self._server_state_dir, str(compile.environment), "compiler")
-        )
+        runner = self._get_compile_runner(compile, project_dir=self._get_project_dir(compile.environment))
 
         merged_env_vars = dict(compile.mergeable_environment_variables)
         for merge_candidate in merge_candidates:
@@ -1094,6 +1173,7 @@ class CompilerService(ServerSlice, inmanta.server.services.environmentlistener.E
         end = datetime.datetime.now().astimezone()
         compile_data_json: Optional[dict[str, object]] = None if compile_data is None else compile_data.model_dump()
         await compile.update_fields(completed=end, success=success, version=version, compile_data=compile_data_json)
+        self._write_last_compile_marker(compile.environment, compile.id)
         awaitables = [
             merge_candidate.update_fields(
                 started=compile.started,
--- a/src/inmanta/server/services/compilerservice.py
+++ b/src/inmanta/server/services/compilerservice.py
@@ -1019,9 +1019,8 @@ class CompilerService(ServerSlice, inmanta.server.services.environmentlistener.E
         environments). They are verified when they are resumed instead.
         """
         for env in await data.Environment.get_list():
-            if env.halted:
-                continue
-            await self._recompile_environment_with_stale_local_state(env)
+            if not env.halted:
+                await self._recompile_environment_with_stale_local_state(env)
 
     async def _recompile_environment_with_stale_local_state(self, env: data.Environment) -> None:
         """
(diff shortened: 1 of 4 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -985,14 +985,95 @@ async def sub_recovery() -> None:
...
+        Verify the local project state of every environment and request an update and recompile where it is stale.
+
+        Halted environments are skipped: a compile requested for them cannot run until they are resumed, so it would
+        only accumulate queued compiles across restarts (which the report retention cleanup does not touch for halted
+        environments). They are verified when they are resumed instead.
+        """
+        for env in await data.Environment.get_list():
+            if env.halted:
+                continue
+            await self._recompile_environment_with_stale_local_state(env)
```
bartv:
```suggestion
            if not env.halted:
                await self._recompile_environment_with_stale_local_state(env)
```

src/inmanta/server/services/compilerservice.py
```diff
@@ -985,14 +985,95 @@ async def sub_recovery() -> None:
...
+        )
+
     async def resume_environment(self, environment: uuid.UUID) -> None:
         """
         Resume compiler service after halt.
         """
         await self._process_next_compile_in_queue(environment=environment)
+        env: Optional[data.Environment] = await data.Environment.get_by_id(environment)
+        if env is not None:
+            # halted environments are not verified at startup: verify the local project state now
```
bartv: What is the relevance of this comment?

src/inmanta/server/services/compilerservice.py
bartv: (Claude on behalf of Bart) Applied the suggestion in d6527461b.

src/inmanta/server/services/compilerservice.py
bartv: (Claude on behalf of Bart) You are right, it restated the docstring of `_recompile_environments_with_stale_local_state` instead of adding information for the reader of this method. Removed in d6527461b.

tests/server/test_compilerservice.py
```diff
@@ -2388,3 +2388,129 @@ async def update_streams(self, out: str = "", err: str = "") -> None:
...
+    completed_compile = data.Compile(
+        remote_id=uuid.uuid4(),
+        environment=env.id,
+        requested=now,
+        started=now,
+        completed=now,
+        success=True,
+        force_update=False,
+    )
+    await completed_compile.insert()
```
bartv: What does this test?

tests/server/test_compilerservice.py
```diff
@@ -2388,3 +2388,129 @@ async def update_streams(self, out: str = "", err: str = "") -> None:
...
+    assert requested.metadata["type"] == "recovery"
+
+    # running the requested compile records it in the marker, after which the check converges to a no-op
+    await run_compile_and_wait_until_compile_is_done(compilerslice, mocked_compiler_service_block, env.id, fail=False)
+    with open(marker_path) as fh:
+        assert fh.read() == str(requested.id)
+    await compilerslice._recompile_environments_with_stale_local_state()
+    assert await data.Compile.get_next_run(env.id) is None
+
+    # a compile that was merged into another one records the compile that actually ran as its substitute:
```
bartv: what does this mean?

tests/server/test_compilerservice.py
bartv: (Claude on behalf of Bart) It inserted a fabricated completed compile row to simulate "the latest compile in the database ran on another server". That was indeed unclear. Reworked in a07f62b4b: the test now runs a real compile and then wipes the state directory to simulate the failover, so the scenario is visible in the test itself and no rows are fabricated.

tests/server/test_compilerservice.py
bartv: (Claude on behalf of Bart) It tried to explain a hand-crafted row with `substitute_compile_id` set, covering the case where the latest compile in the database is a merged request that never ran itself. Reworked in a07f62b4b: the test now queues two identical requests so the compile queue produces a real merge, and the check accepts the marker of the compile that actually ran.

src/inmanta/server/services/compilerservice.py
```diff
@@ -985,14 +985,93 @@ async def sub_recovery() -> None:
             self.fully_ready = True
             # start a waiting compile in each environment
             await self._process_next_compile_in_queue()
+            # A compile recovered above may itself converge stale local state, in which case the check below
+            # queues one redundant recovery compile behind it. This is bounded and self-corrects.
```
arnaudsjs: I am failing to parse this comment.

src/inmanta/server/services/compilerservice.py
```diff
@@ -985,14 +985,93 @@ async def sub_recovery() -> None:
...
+        try:
+            with open(marker_path) as fh:
+                if fh.read().strip() == str(expected_id):
+                    return
+        except OSError:
+            pass
+        LOGGER.warning(
+            "The local project state of environment %s was not produced by compile %s, the latest compile"
+            " in the database (e.g. after a failover to another server or a wiped state directory)."
+            " Requesting an update and recompile to converge.",
```
arnaudsjs: I am only able to understand this because I know the context of the PR. I doubt that it would be clear to an end-user.

```suggestion
            "The local project checkout of environment %s was not produced by compile %s, the latest compile"
            " in the database (This can be caused by a failover to another orchestrator server or a wiped state directory)."
            " Requesting an update and recompile to converge.",
```

src/inmanta/server/services/compilerservice.py
```diff
@@ -985,14 +985,93 @@ async def sub_recovery() -> None:
...
+            expected_id,
+        )
+        await self.request_recompile(
+            env,
+            force_update=True,
+            do_export=True,
+            remote_id=uuid.uuid4(),
+            metadata={
+                "type": "recovery",
+                "message": "Recompile because the local project state does not match the latest compile",
```
arnaudsjs:
```suggestion
                "message": "Recompile because the local project checkout does not match the latest compile",
```

arnaudsjs: This may also cause weird behavior if additional commits were added to the branch, but no update-and-recompile was done on the server yet. But I guess that would be an unusual situation.

tests/server/test_compilerservice.py
```diff
@@ -2388,3 +2388,140 @@ async def update_streams(self, out: str = "", err: str = "") -> None:
 
     assert getattr(run.stage, stream_kwarg) == payload
     assert "\u2026" in getattr(run.stage, stream_kwarg)
+
+
+async def test_recompile_when_local_state_is_stale(mocked_compiler_service_block: queue.Queue, server, environment) -> None:
```
arnaudsjs:
```suggestion
async def test_recompile_when_local_checkout_is_stale(mocked_compiler_service_block: queue.Queue, server, environment) -> None:
```

src/inmanta/server/services/compilerservice.py
```diff
@@ -985,14 +985,93 @@ async def sub_recovery() -> None:
...
+            return
+        try:
+            with open(os.path.join(project_dir, const.INMANTA_LAST_COMPILE_MARKER), "w") as fh:
+                fh.write(str(compile_id))
+        except OSError:
+            LOGGER.warning("Failed to write the last-compile marker for environment %s", environment_id, exc_info=True)
+
+    async def _recompile_environments_with_stale_local_state(self) -> None:
+        """
+        Verify the local project state of every environment and request an update and recompile where it is stale.
```
arnaudsjs:
```suggestion
        Verify the local project checkout of every environment and request an update and recompile where it is stale.
```

tests/server/test_compilerservice.py
```diff
@@ -2388,3 +2388,140 @@ async def update_streams(self, out: str = "", err: str = "") -> None:
...
+    """
+    A server whose local project state was not produced by the environment's latest compile (e.g. after a failover
```
arnaudsjs:
```suggestion
    A server whose local project checkout was not produced by the environment's latest compile (e.g. after a failover
```

tests/server/test_compilerservice.py
```diff
@@ -2388,3 +2388,140 @@ async def update_streams(self, out: str = "", err: str = "") -> None:
...
+    assert await data.Compile.get_next_run(env.id) is None
+
+    # with a repository the stale state is detected and an update and recompile is requested
+    await env.update_fields(repo_url="https://example.com/test/repo.git")
+    await compilerslice._recompile_environments_with_stale_local_state()
+    requested = await data.Compile.get_next_run(env.id)
+    assert requested is not None
+    assert requested.force_update
+    assert requested.do_export
+    assert requested.metadata["type"] == "recovery"
```
arnaudsjs: The code seems to be out of order here.
```suggestion

    # without a repository the server cannot restore the local state by itself: the environment is skipped
    await env.update_fields(repo_url="")
    await compilerslice._recompile_environments_with_stale_local_state()
    assert await data.Compile.get_next_run(env.id) is None

    # simulate a failover to a standby that was never active before: its state directory is empty.
    # The stale state is detected and an update and recompile is requested
    shutil.rmtree(project_dir)
    os.makedirs(project_dir)
    await env.update_fields(repo_url="https://example.com/test/repo.git")
    await compilerslice._recompile_environments_with_stale_local_state()
    requested = await data.Compile.get_next_run(env.id)
    assert requested is not None
    assert requested.force_update
    assert requested.do_export
    assert requested.metadata["type"] == "recovery"
```

arnaudsjs (changes requested):

src/inmanta/server/services/compilerservice.py
bartv: (Claude on behalf of Bart) Applied in 601888431.

src/inmanta/server/services/compilerservice.py
bartv: (Claude on behalf of Bart) Applied in 601888431.

tests/server/test_compilerservice.py
bartv: (Claude on behalf of Bart) Applied in 601888431.

src/inmanta/server/services/compilerservice.py
bartv: (Claude on behalf of Bart) Applied in 601888431.

tests/server/test_compilerservice.py
bartv: (Claude on behalf of Bart) Applied in 601888431.

src/inmanta/server/services/compilerservice.py
bartv: (Claude on behalf of Bart) Rewritten in 601888431. What it tries to say: the recovery of pre-crash queued compiles and this check are independent, so an environment can get one extra recompile at startup when both act on it. Hopefully the new wording parses better.

tests/server/test_compilerservice.py
bartv: (Claude on behalf of Bart) Restructured in 601888431 following your order (no-repository case first, then the empty-state-directory case), with one deviation: the state directory wipe happens before the no-repository assertion. If the checkout still matches the latest compile at that point, the assertion passes for the wrong reason (nothing is stale, so nothing would be requested even without the repo_url guard). Wiping first makes it prove the guard: the checkout is stale, yet the environment is skipped because there is no repository to restore from.

src/inmanta/server/services/compilerservice.py
bartv: (Claude on behalf of Bart) Correct, and it is deliberate: the recovery compile has the same semantics as /notify?update=true, so it converges to the current head of the branch, which can be ahead of what the previous server had compiled. Reproducing the exact pre-failover checkout is not possible in general (the compile does not record the commit hash, and module versions are resolved at install time), and compiling the branch head is what the operator gets on the next update anyway, just earlier. For operators who need bit-identical behavior across a failover the answer is pinning: exact module pins and a stable branch. I will spell this out in the HA documentation follow-up.

arnaudsjs (approved):

PR #10554, opened 2026-07-08
Backport to iso9: add a revocable API-token registry (jti allowlist) (#10535)

bartv:
Manual iso9 backport of #10535 (the auto-backport conflicted on the master/iso9 REST divergence).

Conflict resolution:
- `rest/__init__.py`: iso9 has no module-level `execute_call`/`match_call` (it uses the `RESTBase._execute_call` method), so the `await auth.validate_jti(arguments.auth_token)` enforcement was threaded into that method after `authenticate`, rather than the master module-level function.
- `websocket.py`: does not exist on iso9 (master-only websocket transport), so #10535's change to it is dropped.
- Changelog `destination-branches` set to iso9 only.

Verified on iso9: migration `v202607040` applies (test_v202605150_to_v202607040), `test_token_registry` + cleanup + schema-accounting pass, black/flake8 clean, mypy adds no new violations.


src/inmanta/protocol/auth/auth.py
```diff
@@ -165,6 +167,80 @@ def decode_token(token: str) -> tuple[claim_type, "AuthJWTConfig"]:
...
+    if not claims:
+        return
+    jti = claims.get("jti")
+    if not jti or not isinstance(jti, str):
+        return
+
+    now = time.monotonic()
+    cached_expiry = _jti_cache.get(jti)
+    if cached_expiry is not None and cached_expiry > now:
+        # A cached jti is known to be present and not revoked (only valid tokens are cached), so accept it.
```
Hugo-Inmanta, line 209: Why did we stop caching revoked tokens ? I think it was useful against e.g. a dos attack that uses a revoked but not yet expired token ?

Hugo-Inmanta (changes requested):

src/inmanta/protocol/auth/auth.py
jptrindade: When we revoke the token we also remove it from the cache

bartv: This is also in the master branch based on feedback from @jptrindade

Hugo-Inmanta: I had missed that part, then the question becomes why do we evict them from the cache when we revoke them ?

bartv: It simplified the code. There is a small performance penalty in case of revoked tokens still being in use. But for me that is a corner case

Hugo-Inmanta (approved):

PR #10561, opened 2026-07-09
Add reusable resolver from resource filter to resource ids

bartv:
_Claude opened this PR on behalf of @bartv._

# Description

Groundwork for triggering resource actions (deploy/dryrun) on the result of a resource filter.

Extracts the version-selection logic of the GraphQL `resources` query into a reusable `join_latest_version` helper and adds `resolve_resource_ids(filter)`, which turns a `ResourceFilter` into the set of matching resource ids (no pagination, sorting or extension output columns). A follow-up PR uses this to trigger a deploy/repair on exactly the set of resources a filter matches.

The `resources` query resolver is refactored to reuse `join_latest_version`, so its behaviour is unchanged (covered by the existing `test_query_resources`).

This is the first of a stack of 3 PRs:
1. **This PR** - filter -> resource ids resolver
2. Deploy/repair scoped to a collection of resources (internal plumbing)
3. `deploy_filtered` REST endpoint (ties 1 and 2 together)

## Tests
- `test_resolve_resource_ids` asserts the resolver returns exactly the same set of resource ids as the equivalent `resources` GraphQL query, across agent / resource_type / resource_id_value / orphan / enum / bool filters, plus environment scoping and the empty-match case.
- `test_query_resources` (existing) covers the refactored resolver.

# Self Check:

- [x] Changelog entry
- [x] ~~End user documentation is included or an issue is created for end-user documentation~~ (internal helper, no user-facing change)

### DIFF
--- a/changelogs/unreleased/resource-filter-resolver.yml
+++ b/changelogs/unreleased/resource-filter-resolver.yml
@@ -1,0 +1,6 @@
+description: >-
+  Add a reusable resolver that turns a resource filter into the set of matching resource ids, reusing the
+  version-selection logic of the GraphQL `resources` query. This is internal groundwork for triggering resource
+  actions (deploy/dryrun) on the result of a filter.
+change-type: patch
+destination-branches: [master, iso9]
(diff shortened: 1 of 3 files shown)


bartv:
_Claude, on behalf of @bartv._

Closing in favour of consolidating the resource-filter work into #10563, rebuilt on top of #10530. #10530 moves the version/orphan selection this PR extracted into `join_latest_version` onto the filter components (`CoreResourceFilter.apply_version_filter`), so the standalone refactor here is superseded. The reusable `resolve_resource_ids` it introduced is re-derived in #10563 on top of #10530's composed filter.

PR #10562, opened 2026-07-09
Allow deploy/repair to be scoped to a collection of resources

bartv:
_Claude opened this PR on behalf of @bartv._

# Description

Internal execution layer for filtered deploy/repair actions.

Threads an optional `resources` collection through `scheduler.deploy` and `scheduler.repair` (intersected with the existing agent filter) and exposes it on the internal agent-facing `trigger` RPC and `Agent.trigger_update`. The public `/deploy` server interface is intentionally left unchanged: the resolved set of resources is fed in through the agent interface only, so the user-facing surface stays a single filter-based endpoint (see the follow-up PR).

Reuses the scheduler/agent approach from the unmerged #10259, restricted to the agent interface (no public server-interface change), following the direction in #10479.

Second of a stack of 3 PRs (independent of PR 1):
1. Filter -> resource ids resolver
2. **This PR** - deploy/repair scoped to a collection of resources
3. `deploy_filtered` REST endpoint (ties 1 and 2 together)

Relates to #10479.

## Tests
- `test_deploy_specific_resources` (scheduler level): deploy only touches dirty resources within the collection, repair touches all in the collection (even compliant), and both intersect correctly with the agent filter.
- Updated the existing `test_deploy_trigger` log assertions for the new (resource-count) log message.

# Self Check:

- [ ] Attached issue to pull request
- [x] Changelog entry
- [x] Type annotations are present
- [x] Code is clear and sufficiently documented
- [x] No (preventable) type errors (check using make mypy or make mypy-diff)
- [x] Sufficient test cases (reproduces the bug/tests the requested feature)
- [x] Correct, in line with design
- [x] ~~End user documentation is included or an issue is created for end-user documentation~~ (internal interface only)
- [x] ~~If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s)~~

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #10556, opened 2026-07-08
[Resource Filters] Port the resource filters logic to REST API endpoints


jptrindade:
From the [design](https://github.com/inmanta/designs/blob/master/resource-filters.md):

The new REST API endpoints will receive a JSON representation of a ResourceFilter This JSON object will then be translated into a proper GraphQL ResourceFilter and piped to the appropriate query/mutation. Meaning that the REST API endpoints and GraphQL query/mutation share the same backend logic
```
# inmanta-core

# methods_v2.py
def list_resource_filtered(
    tid: uuid.UUID,
    filter: dict[str, Any],
    first: Optional[int] = None,
    after: Optional[str] = None,
    last: Optional[int] = None,
    before: Optional[str] = None,
    sort: str = "resource_type.asc",
) -> ReturnValue[list[dict[str, Any]]]:
  ...


def deploy_resource_filtered(
    tid: uuid.UUID,
    filter: dict[str, Any],
) -> ReturnValue[list[dict[str, Any]]]:
  ...


def dryrun_resource_filtered(
    tid: uuid.UUID,
    filter: dict[str, Any],
    model_version: int
) -> ReturnValue[list[dict[str, Any]]]:
  ...

# graphql.py
class GraphQLSlice(protocol.ServerSlice):
    ...

    def convert_dict_to_graphql_expression(self,
        tid: uuid.UUID,
        filter: dict[str, Any],
        first: Optional[int] = None,
        after: Optional[str] = None,
        last: Optional[int] = None,
        before: Optional[str] = None,
        sort: str = "resource_type.asc"):
      ...


    @handle(methods_v2.list_resource_filtered)
    async def list_resource_filtered(
        self,
        tid: uuid.UUID,
        filter: dict[str, Any],
        first: Optional[int] = None,
        after: Optional[str] = None,
        last: Optional[int] = None,
        before: Optional[str] = None,
        sort: str = "resource_type.asc",
    ) -> ReturnValue[GraphQLResult]:
        gpl_query = self.convert_dict_to_graphql_expression(tid, filter, first, after, last, before, sort)
        execution_result = await self.execute_query(gpl_query)
        rest_resource = DynamicRESTResource.from_execution_result(execution_result)
        return ReturnValue(status_code=rest_resource.status_code, response=rest_resource)

    # And something equivalent for the other resource_actions
```

Issue #10479, opened 2026-06-16
Modify the scheduler and the appropriate endpoints to accept specific resource_ids on deploy


jptrindade:
We need to be able to deploy specific resource_ids when requested by the user. 
We decided we would first trim the resources based on an user input filter (graphql on the first phase) and fetch the resulting resource_ids. 
These resource_ids will then be piped to an endpoint and that endpoint will then call the scheduler and ask for them to be deployed.
Additionally, the user can call this endpoint directly if they already know the resource ids that they want to deploy.

sanderr:
#10259 probably contains everything required. It hasn't been reviewed yet, but iirc it's complete and thorough. To be double checked. The one thing that it has that we don't need (that needs to be stripped out) is probably the `/deploy` change. We don't want to extend the interface of the server, only that of the agent.

PR #10563, opened 2026-07-09
Add deploy_filtered REST endpoint (filter-driven deploy/repair)

bartv:
_Claude opened this PR on behalf of @bartv._

# Description

Consolidated resource-filter feature: trigger a deploy or repair on the resources matching a resource filter, over REST. Built **on top of #10530** (GraphQL filter composition), and supersedes the now-closed #10561 (resolver) and #10562 (deploy/repair subset), which are folded in here.

It contains two commits:

1. **Scheduler/agent `resources=` plumbing** - `scheduler.deploy`/`repair` and the internal agent `trigger` accept an optional resource subset (intersected with the agent filter); the public `/deploy` interface is unchanged. This reuses the approach from Sander's unmerged #10259, restricted to the agent interface (per the direction on #10479); it is a re-implementation, not a cherry-pick of that branch.
2. **`deploy_filtered` REST endpoint** (`POST /api/v2/deploy_filtered`).

## The endpoint

The filter body reuses #10530's composed GraphQL `ResourceFilter` as its single source of truth. A method argument marked `Annotated[GraphQLFilter, graphql_input("ResourceFilter")]` (`inmanta.graphql.rest_filter`) is:
- **validated** by graphql-core input coercion against that composed input type (via pydantic `__get_pydantic_core_schema__`), and
- **documented** in OpenAPI from GraphQL introspection (via pydantic `__get_pydantic_json_schema__`).

Both resolve the type lazily from a registry the `GraphQLSlice` publishes at start (the composed `ResourceFilter` minus `environment`, which REST supplies as the `tid`), so no protocol-layer changes are needed. `resolve_resource_ids` reconstructs the coerced body and decomposes it into the **same filter components the `resources` query uses**, applies version selection + each component's filter, and returns the matching ids, which are piped to the scheduler through the internal `trigger`. When an extension composes fields into `ResourceFilter` (e.g. LSM, once #10530 + inmanta-lsm#2439 land), they flow into REST validation and OpenAPI automatically - demonstrated here by `modelVersion` (added by #10530) surfacing in the endpoint's OpenAPI with no endpoint change.

- Wire format is the GraphQL representation (camelCase fields, enum values as GraphQL names) - a filter object is interchangeable between the REST body and GraphQL `variables`.
- A single resource is a sufficiently specific filter; repair vs incremental deploy via `agent_trigger_method`; a deploy acts on the current desired state, so `modelVersion` and `isOrphan: true` are rejected (400).

> ⚠️ **Stacked on #10530.** Base is `resource-filter-composition`; merge after #10530. It also uses `strawberry.Schema._schema` (a strawberry internal) to reach the composed input type - flagged for review.

Relates to #10556. Supersedes #10561, #10562.

## Tests
- `test_deploy_filtered` (e2e): repair on a filter, single-resource targeting, incremental-only-dirty, and 400s for unknown field, bad operator, `environment` in the body, `isOrphan: true`, and `modelVersion`.
- `test_deploy_filtered_openapi`: the endpoint's OpenAPI documents the filter as an object with the filter fields (camelCase, incl. the composed `modelVersion`) and without `environment`.
- `test_deploy_specific_resources` (scheduler-level, from the plumbing).

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -26,7 +26,7 @@ import logging
 import typing
 import uuid
 from abc import abstractmethod
-from collections.abc import Mapping, Sequence, Set
+from collections.abc import Collection, Mapping, Sequence, Set
 from dataclasses import dataclass
 from enum import Enum
 from typing import ClassVar, Optional, Self
@@ -590,19 +590,25 @@ class ResourceScheduler(TaskManager):
         reason: str,
         priority: TaskPriority = TaskPriority.USER_DEPLOY,
         agent: Optional[str] = None,
+        resources: Optional[Collection[ResourceIdStr]] = None,
     ) -> None:
         """
         Trigger a deploy
 
         :param agent: If given, deploy resources only for this agent. Otherwise deploy for all agents.
+        :param resources: If given, deploy only resources in this collection. Otherwise deploy all resources
+            (on the given agent if one is given). Combined with the agent filter as an intersection.
         """
         if not self._running:
             LOGGER.debug("Ignoring deploy request for halted resource scheduler")
             return
         async with self._scheduler_lock:
-            to_deploy: Set[ResourceIdStr] = (
-                self._state.dirty if agent is None else self._state.dirty & self._state.resources_by_agent.get(agent, set())
-            )
+            intersection_filters: list[Collection[ResourceIdStr]] = []
+            if resources is not None:
+                intersection_filters.append(resources)
+            if agent is not None:
+                intersection_filters.append(self._state.resources_by_agent.get(agent, set()))
+            to_deploy: Set[ResourceIdStr] = set.intersection(self._state.dirty, *intersection_filters)
             if agent is not None:
                 LOGGER.debug("Triggering deploy for %d resources on agent %s because %s", len(to_deploy), agent, reason)
             else:
@@ -616,11 +622,14 @@ class ResourceScheduler(TaskManager):
         reason: str,
         priority: TaskPriority = TaskPriority.USER_REPAIR,
         agent: Optional[str] = None,
+        resources: Optional[Collection[ResourceIdStr]] = None,
     ) -> None:
         """
         Trigger a repair, i.e. mark all unblocked resources as dirty, then trigger a deploy.
 
         :param agent: If given, repair resources only for this agent. Otherwise repair for all agents.
+        :param resources: If given, repair only resources in this collection. Otherwise repair all resources
+            (on the given agent if one is given). Combined with the agent filter as an intersection.
         """
 
         def should_deploy_resource(resource: ResourceIdStr) -> bool:
@@ -636,9 +645,10 @@ class ResourceScheduler(TaskManager):
             LOGGER.debug("Ignoring repair request for halted resource scheduler")
             return
         async with self._scheduler_lock:
-            in_scope: Set[ResourceIdStr] = (
+            base_in_scope: Set[ResourceIdStr] = (
                 self._state.intent.keys() if agent is None else self._state.resources_by_agent.get(agent, set())
             )
+            in_scope: Set[ResourceIdStr] = base_in_scope & set(resources) if resources is not None else base_in_scope
 
             to_deploy: Set[ResourceIdStr] = {resource for resource in in_scope if should_deploy_resource(resource)}
             if agent is not None:
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -595,6 +595,8 @@ class ResourceScheduler(TaskManager):
         """
         Trigger a deploy
 
+        :param reason: Human-readable reason for the deploy, recorded on the scheduled tasks.
+        :param priority: Scheduling priority of the deploy tasks.
         :param agent: If given, deploy resources only for this agent. Otherwise deploy for all agents.
         :param resources: If given, deploy only resources in this collection. Otherwise deploy all resources
             (on the given agent if one is given). Combined with the agent filter as an intersection.
@@ -627,6 +629,8 @@ class ResourceScheduler(TaskManager):
         """
         Trigger a repair, i.e. mark all unblocked resources as dirty, then trigger a deploy.
 
+        :param reason: Human-readable reason for the repair, recorded on the scheduled tasks.
+        :param priority: Scheduling priority of the repair tasks.
         :param agent: If given, repair resources only for this agent. Otherwise repair for all agents.
         :param resources: If given, repair only resources in this collection. Otherwise repair all resources
             (on the given agent if one is given). Combined with the agent filter as an intersection.
--- a/src/inmanta/graphql/rest_filter.py
+++ b/src/inmanta/graphql/rest_filter.py
@@ -15,7 +15,7 @@ Contact: code@inmanta.com
 import importlib
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
-from typing import Annotated, Optional
+from typing import Annotated, Optional, cast
 
 from pydantic import ConfigDict, GetCoreSchemaHandler, GetJsonSchemaHandler
 from pydantic.json_schema import JsonSchemaValue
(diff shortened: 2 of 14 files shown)


src/inmanta/deploy/scheduler.py
```diff
@@ -590,19 +590,25 @@ async def deploy(
...
         async with self._scheduler_lock:
-            to_deploy: Set[ResourceIdStr] = (
-                self._state.dirty if agent is None else self._state.dirty & self._state.resources_by_agent.get(agent, set())
-            )
+            intersection_filters: list[Collection[ResourceIdStr]] = []
+            if resources is not None:
+                intersection_filters.append(resources)
+            if agent is not None:
+                intersection_filters.append(self._state.resources_by_agent.get(agent, set()))
+            to_deploy: Set[ResourceIdStr] = set.intersection(self._state.dirty, *intersection_filters)
```
jptrindade, line 613: I have some mixed feelings about this, this would mean that we would only deploy the resources that are in the dirty set even if our filter includes other "clean" ones. If I have a compliant (at the time) resource and I filter on it and choose to deploy it, I assume I want the deploy to be triggered anyway and not skipped

jptrindade: Just saw the method definition and that it is intended. So for my use case we should just use the repair button

src/inmanta/graphql/rest_filter.py
jptrindade, line 1: This whole file is a bit hard to reason, I'll come back to it after the first pass

src/inmanta/graphql/schema.py
```diff
@@ -856,6 +857,131 @@ def add_filter_and_sort[*Ts](
...
+        and_(
+            models.t_resource_set_configuration_model.c.environment == models.Resource.environment,
+            models.t_resource_set_configuration_model.c.resource_set == models.Resource.resource_set,
+            models.t_resource_set_configuration_model.c.model
+            == select(latest_scheduled_version_cte.c.version).scalar_subquery(),
+        ),
+    )
+
+
+async def resolve_resource_ids(filter: ResourceFilter) -> set[ResourceIdStr]:
```
jptrindade: This would need to be more involved to support filters from extensions. Separate each extension and run apply_filter to the stmt, but I assume that this PR predates the filter composition

src/inmanta/protocol/methods_v2.py
```diff
@@ -1387,6 +1388,35 @@ def set_fact(
...
+    """
+    Trigger a deploy (or repair) on the resources matching the given filter, on the current desired state (the
+    scheduler's last processed version). The filter body is the GraphQL ``resources`` query's ``ResourceFilter``
+    (minus ``environment``, which is the ``tid``): it is validated against, and documented from, that GraphQL type,
+    so the action operates on exactly the set of resources the ``resources`` view returns. A single resource can be
+    targeted with a sufficiently specific filter (e.g. ``resourceType`` + ``agent`` + ``resourceIdValue``).
+
+    :param tid: The id of the environment.
+    :param filter: The resource filter, as a JSON object matching the GraphQL ``ResourceFilter`` (camelCase field
+        names, enum values as their GraphQL names). When omitted, all resources are selected. Malformed filters
```
jptrindade: Shouldn't this be snake case field names that the user is inputting?

jptrindade: The rest of the code seems to expect this to be camelCase after all. It seems to go against our naming convention, how do you feel about this?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1283,6 +1283,36 @@ async def deploy(
...
+    @handle(methods_v2.deploy_filtered, env="tid")
+    async def deploy_filtered(
+        self,
+        env: data.Environment,
+        filter: Optional[Mapping[str, object]] = None,
+        agent_trigger_method: const.AgentTriggerMethod = const.AgentTriggerMethod.push_full_deploy,
+    ) -> ReturnValue[list[ResourceIdStr]]:
+        # Imported here to avoid a module-level import cycle between the orchestration service and the graphql schema.
+        # `filter` is the value produced by graphql-core coercion of the request body (keyed by GraphQL field names).
+        from inmanta.graphql.schema import build_resource_filter_from_coerced, resolve_resource_ids
```
jptrindade: It's a bit annoying that we have to do this like this. Is there a better file/code structure that we can use to import it at top level while not getting stuck in an import cycle?

jptrindade (changes requested):

src/inmanta/deploy/scheduler.py
bartv: I focused mostly on the external interaction, the exact interaction is mostly based on #10259

src/inmanta/protocol/methods_v2.py
bartv: yeah, the main benefit now is: 100% identical to the graphql version. That is enough for me to ignore this, the benefit is too high

src/inmanta/graphql/rest_filter.py
```diff
@@ -0,0 +1,165 @@
...
+    It is intentionally a ``BaseModel``: should the ``graphql_input`` metadata ever be omitted, the argument still
+    validates as a JSON object rather than as an unchecked value, and the protocol type validation accepts it as a
+    known (structured) argument type.
+    """
+
+    model_config = ConfigDict(extra="allow")
+
+
+@dataclass(frozen=True)
+class graphql_input:
```
jptrindade, line 60: Why is this lower case?

src/inmanta/graphql/rest_filter.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+
+    type_name: str
+
+    def __get_pydantic_core_schema__(self, source_type: object, handler: GetCoreSchemaHandler) -> core_schema.CoreSchema:
+        # Runs when the arguments model is built (import time), before the GraphQL schema exists. Only return a
+        # validator; the registry is resolved lazily inside `_coerce` (per request, after start).
+        return core_schema.no_info_plain_validator_function(self._coerce)
+
+    def __get_pydantic_json_schema__(self, schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
```
jptrindade, line 80: So this is only for generating documentation?
The one above does the translation with `coerce_input_value` I assume

jptrindade (commented):
Overall I think it is a nice improvement. I would for sure get Sander to look at this before merging though.

src/inmanta/graphql/rest_filter.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@dataclass(frozen=True)
+class ResolvedFilter:
+    """Composed-filter artifacts attached to the core filter class at start: the graphql-core input type (for
+    coercion/OpenAPI) and the strawberry composed type + components (for resolve_resource_ids)."""
+
+    input_type: GraphQLInputObjectType
+    composed_type: type
+    components: tuple[type, ...]
```
jptrindade, line 49: I think this could do with better documentation.

From what I gathered from seeing it in action, the input_type is the generated graphql type, composed type is the composed strawberry type and the components are the parts of the composed_type. Is that correct?

src/inmanta/graphql/rest_filter.py
```diff
@@ -0,0 +1,146 @@
...
+    filter_class: type | str
+
+    def _resolve_class(self) -> type:
+        if isinstance(self.filter_class, str):
+            module_path, _, class_name = self.filter_class.rpartition(".")
+            return cast(type, getattr(importlib.import_module(module_path), class_name))
+        return self.filter_class
+
+    def _resolved(self) -> Optional[ResolvedFilter]:
+        resolved: Optional[ResolvedFilter] = getattr(self._resolve_class(), "__resolved_filter__", None)
```
jptrindade, line 73: I would make `__resolved_filter__` a const somewhere. One missing `_` will have us scratching our heads

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1283,6 +1283,38 @@ async def deploy(
...
+        # A deploy always acts on the current desired state (the scheduler's last processed version), so a historical
+        # snapshot (`modelVersion`) or orphaned resources (`isOrphan: true`) may not be selected.
+        if filter is not None:
+            if filter.get("modelVersion") is not None:
+                raise BadRequest("Cannot deploy a specific model version: 'modelVersion' is not allowed for deploy.")
+            if filter.get("isOrphan") is True:
+                raise BadRequest("Cannot deploy orphaned resources: the 'isOrphan' filter must be omitted or set to false.")
+
+        resource_ids: list[ResourceIdStr] = sorted(await resolve_resource_ids(filter or {}, env.id))
+
```
jptrindade, line 1306: if filter is None can't we just skip this step and trigger a normal deploy/repair? It would result in the same output but be more efficient

jptrindade (changes requested):
Looks better now. I would like a test that converts REST input to GraphQL and asserts that we get the expected output. Maybe change `test_query_resources` to include a REST equivalent and make sure that the raw GraphQL query obtains the same results as the REST-translated GraphQL query

src/inmanta/graphql/schema.py
```diff
@@ -1331,6 +1334,77 @@ def decompose_and_validate_filter[F: StrawberryFilter](filter: object, component
...
+            continue
+        if isinstance(value, Mapping):
+            sub_filter_type = cast(typing.Callable[..., object], _strip_optional(field_types[attribute]))
+            setattr(result, attribute, sub_filter_type(**{to_snake_case(key): val for key, val in value.items()}))
+        else:
+            setattr(result, attribute, value)
+    return result
+
+
+async def resolve_resource_ids(coerced_filter: Mapping[str, object], environment: uuid.UUID) -> set[ResourceIdStr]:
```
jptrindade, line 1365: I think most of this logic is shared with the `resources` query. I would like to reuse it when possible just so we don't diverge by accident

bartv:
_Claude here, on behalf of @bartv._ This PR could not be reopened because its base branch (`resource-filter-composition`, #10530) was squash-merged into `master` and deleted. The work has been rebased onto `master` and continues in #10575.

PR #10575, opened 2026-07-13
Add deploy_filtered REST endpoint (filter-driven deploy/repair)

bartv:
_Claude here, opening this on behalf of @bartv._

Supersedes #10563, which was auto-closed when its base branch (`resource-filter-composition`, #10530) was squash-merged into `master` and deleted (GitHub cannot reopen a PR whose base branch is gone). This is the same work, rebased onto `master` so it now targets `master` directly. Its only dependency, #10530, is merged.

## What this adds

Trigger `deploy` / `repair` against the result of a **resource filter**, over REST, reusing the composed GraphQL `ResourceFilter` from #10530 instead of introducing a new filter language.

- **Scheduler / agent plumbing** (`add6f2b97`): let deploy/repair be scoped to a collection of resources (`resources=`), on top of the existing scoped-deploy machinery.
- **`deploy_filtered` endpoint** (`035b9ec47`): new `methods_v2` endpoint and orchestration-service handler. It resolves the filter to resource ids server-side and schedules the action. Repair vs deploy is chosen via `agent_trigger_method`; orphaned resources are rejected for deploy.
- **Filter marker** (`8e12a3262`, `fc95d7646`): a `graphql_input(<class or dotted path>)` annotation lets a REST argument reuse a GraphQL input type as its request body, so REST and GraphQL share one filter definition. Request validation goes through graphql-core coercion; OpenAPI is produced from introspection. It is resolved lazily to avoid an import cycle, with no module-global registry.

## Design

The filter is the only user-facing surface; there is no public `resource_ids` API. Mechanism **B2**: the composed strawberry `ResourceFilter` is the single source of truth, the REST body is JSON coerced via `graphql.utilities.coerce_input_value`, and OpenAPI is generated from GraphQL introspection, keeping REST and GraphQL in sync by construction.

## Status

Draft, pending a full CI run after the rebase onto the squashed #10530. The rebase was clean except for a one-line `mypy-baseline.txt` conflict (the base's `methods_v2` empty-body count shifted), which I resolved by hand; the baseline may still need a regen once CI runs. Changelog entry: `changelogs/unreleased/deploy-filtered-endpoint.yml`.


src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1283,6 +1283,38 @@ async def deploy(
...
+        await self.autostarted_agent_manager._ensure_scheduler(env.id)
+        client = self.agentmanager_service.get_agent_client(env.id)
+        if not client:
+            raise NotFound("The scheduler for this environment could not be reached")
+
+        if resource_ids:
+            incremental_deploy = agent_trigger_method is const.AgentTriggerMethod.push_incremental_deploy
+            self.add_background_task(client.trigger(env.id, None, incremental_deploy, resources=resource_ids))
+
+        return ReturnValue(response=resource_ids)
```
sanderr, line 1316: I don't think an explicit `ReturnValue` is required nowadays, but it's been a while.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1283,6 +1283,38 @@ async def deploy(
...
+
+        # A deploy always acts on the current desired state (the scheduler's last processed version), so a historical
+        # snapshot (`modelVersion`) or orphaned resources (`isOrphan: true`) may not be selected.
+        if filter is not None:
+            if filter.get("modelVersion") is not None:
+                raise BadRequest("Cannot deploy a specific model version: 'modelVersion' is not allowed for deploy.")
+            if filter.get("isOrphan") is True:
+                raise BadRequest("Cannot deploy orphaned resources: the 'isOrphan' filter must be omitted or set to false.")
+
+        resource_ids: list[ResourceIdStr] = sorted(await resolve_resource_ids(filter or {}, env.id))
```
sanderr, line 1305: I would think we need to inject `isOrphan=False` into the query here. Or alternatively to make it an explicit paremeter of the function.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1283,6 +1283,38 @@ async def deploy(
 
         return attach_warnings(200, {"agents": sorted(allagents)}, warnings)
 
+    @handle(methods_v2.deploy_filtered, env="tid")
+    async def deploy_filtered(
+        self,
+        env: data.Environment,
+        filter: Optional[Mapping[str, object]] = None,
```
sanderr, line 1290: This seems to conflict with the interface in `methods_v2`. It declares a `BaseModel` while this is a `Mapping`.

src/inmanta/graphql/rest_filter.py
```diff
@@ -0,0 +1,146 @@
...
+        return schema
+    if isinstance(gql_type, GraphQLEnumType):
+        return {"type": "string", "enum": list(gql_type.values.keys())}
+    if isinstance(gql_type, GraphQLScalarType):
+        return dict(_SCALAR_TO_OPENAPI.get(gql_type.name, {"type": "string"}))
+    return {"type": "object"}
+
+
+# ResourceFilter uses the string form because importing schema here would cycle; other filters can pass the class.
+ResourceFilterArg = Annotated[GraphQLFilter, graphql_input("inmanta.graphql.schema.CoreResourceFilter")]
```
sanderr, line 146: Pretty sure we can restructure slightly so that this is not required, and we can lose the `importlib` import during resolution.

sanderr, line 146: I find `CoreResourceFilter` a strange attachement point. Then again, the code has changed a lot so perhaps it was logical at the time. What if:
- We move `build_composed_filter_input` (and possibly similar functions) to be methods on `ContributableGraphQLType`.
- That method populates an attribute on the class. We wrap it in a `property` that raises an exception if it is accessed before initialization. (Need to double check with João that it is OK to cache, because some classes had caching issues, but I think that was related to the sqlalchemy mapper.)
- We use the `ContributableGraphQLType` instance as the argument to `graphql_input`. We can access it easily as `schema.CONTRIBUTABLE_MODELS[models.Resource]`.
- The `graphql_input` "resolution" and the `get_schema()` lose some metaprogramming and can simply access the attribute populated by `build_composed_filter_input`.

src/inmanta/graphql/rest_filter.py
```diff
@@ -0,0 +1,146 @@
...
+    "String": {"type": "string"},
+    "Int": {"type": "integer"},
+    "Float": {"type": "number"},
+    "Boolean": {"type": "boolean"},
+    "ID": {"type": "string"},
+    "UUID": {"type": "string", "format": "uuid"},
+}
+
+
+def graphql_input_to_openapi(gql_type: object) -> dict[str, object]:
```
sanderr, line 121: I wonder if Bart purposely decided to use `graphql` directly here, rather than `strawberry`. For `graphql` we require the full schema, which isn't "statically" (after server start) available).

I'm inclined to investigate how much effort it would be to port this to work directly on the composed `ResourceFilter` class.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -1283,6 +1283,38 @@ async def deploy(
...
+        agent_trigger_method: const.AgentTriggerMethod = const.AgentTriggerMethod.push_full_deploy,
+    ) -> ReturnValue[list[ResourceIdStr]]:
+        # Imported here to avoid a module-level import cycle between the orchestration service and the graphql schema.
+        # `filter` is the value produced by graphql-core coercion of the request body (keyed by GraphQL field names).
+        from inmanta.graphql.schema import resolve_resource_ids
```
sanderr, line 1295: I wonder if this should be a method on the graphql slice, so that the orchestration service doesn't need to be aware of the details, only the slice and what it offers. It might not even need the import, since iirc we have a method to retrieve a slice by name from the server.

src/inmanta/graphql/schema.py
```diff
@@ -1333,6 +1336,77 @@ def decompose_and_validate_filter[F: StrawberryFilter](filter: object, component
...
+            continue
+        if isinstance(value, Mapping):
+            sub_filter_type = cast(typing.Callable[..., object], _strip_optional(field_types[attribute]))
+            setattr(result, attribute, sub_filter_type(**{to_snake_case(key): val for key, val in value.items()}))
+        else:
+            setattr(result, attribute, value)
+    return result
+
+
+async def resolve_resource_ids(coerced_filter: Mapping[str, object], environment: uuid.UUID) -> set[ResourceIdStr]:
```
sanderr, line 1359: I'm honestly quite confused by this method. Wouldn't it be simpler to just call into the actual graphql method? I'm not opposed to this per se, but only with solid motivation. The duplication of this complex code is not for free + it requires the two conversion methods above, which we really don't need afaik.

src/inmanta/graphql/schema.py
```diff
@@ -1333,6 +1336,77 @@ def decompose_and_validate_filter[F: StrawberryFilter](filter: object, component
...
+    fields as long as each is a nested operator object or a scalar."""
+    result = composed_type(environment=environment)
+    field_types = typing.get_type_hints(composed_type)
+    for graphql_name, value in coerced.items():
+        attribute = to_snake_case(graphql_name)
+        if attribute == "environment":
+            continue
+        if isinstance(value, Mapping):
+            sub_filter_type = cast(typing.Callable[..., object], _strip_optional(field_types[attribute]))
+            setattr(result, attribute, sub_filter_type(**{to_snake_case(key): val for key, val in value.items()}))
```
sanderr, line 1353: Is this to support custom mapping types? Do we really need to?

sanderr (approved):
Approving to get this merged as a first stage. Requires cleanup, but should be functional so we want to start using it while cleanup is in progress.

PR #10576, opened 2026-07-13
Record when a token was revoked and clean up revoked tokens after a retention window

### LINKED ISSUE #10569: Add support to remove revoked tokens
The `environment_service._cleanup_tokens()` method cleans up tokens that have expired. Token that were revoked without any expiry timestamp will remain in the database until forever.

### LINKED ISSUE #10572: Track when an API token was revoked (revoked_at)
Follow-up to the token registry (#10535), from the demo-session feedback.

The `token` table records `revoked` (bool) but not *when* a token was revoked. Add a `revoked_at` timestamp for auditing and to drive the cleanup retention window.

### Scope
- Add `revoked_at timestamptz NULL` to the `token` table (new DB migration + dump).
- `TokenRepository.revoke` sets `revoked=True, revoked_at=now()`.
- Expose `revoked_at` in the `Token` DTO and `environment_token_list`.
- Interacts with #10569 (cleanup of revoked tokens): the retention window for purging revoked tokens keys off `revoked_at`, so these two are best done together.

### Companion
web-console: show the revoked-at timestamp in the Tokens tab - separate web-console ticket.

bartv:
# Description

Follow-up to the token registry (#10535), covering two demo-session feedback items:

* **Track when a token was revoked**: the `token` table gets a `revoked_at timestamptz NULL` column (new migration, already-revoked tokens are backfilled with the migration time). `TokenRepository.revoke` stamps it via `coalesce(revoked_at, now())`, so revoking stays idempotent and the original revocation time is preserved. The timestamp is exposed in the `Token` DTO and `environment_token_list`.
* **Clean up revoked tokens**: previously only expired tokens were pruned, so revoked tokens without an expiry stayed in the registry forever. The hourly cleanup now also deletes tokens that have been revoked for longer than the new `server.revoked-token-retention` option (default thirty days); until then they remain listed for auditing.

One piece of collateral: newer pg_dump versions (16.10 / 17.6, the `\restrict` security fix) emit psql meta-commands in dumps, and `PGRestore` choked on them when restoring the v202607040 dump. The new migration test is the first to restore such a dump, so `PGRestore` now skips `\restrict`/`\unrestrict` lines.

closes #10572
closes #10569

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (config option is documented via its option definition; the API docstring references it)


src/inmanta/data/model.py
```diff
@@ -926,6 +926,7 @@ class Token(BaseModel):
     issued_at: datetime.datetime
     expires_at: datetime.datetime | None = None
     revoked: bool = False
+    revoked_at: datetime.datetime | None = None
```
sanderr, line 928: I don't have a good grip on how / where this class is used. Could we stick with a single `revoked: datetime.datetime |  None` field instead of two coupled fields?

src/inmanta/data/sqlalchemy.py
```diff
@@ -584,20 +588,34 @@ async def add(self, token: Token) -> None:
...
-        await self.session.execute(delete(Token).where(Token.expires_at.is_not(None), Token.expires_at < func.now()))
+    async def delete_stale(self, revoked_before: datetime.datetime) -> None:
+        """
+        Delete registry entries that no longer serve a purpose: tokens that have expired, and tokens that
+        were revoked before the given moment (kept until then for auditing). The caller commits.
+        """
+        await self.session.execute(
+            delete(Token).where(
+                or_(
+                    and_(Token.expires_at.is_not(None), Token.expires_at < func.now()),
```
sanderr: I realize this is not a change of behavior, but why do we keep revoked tokens for auditing but not expired ones? For auditing, or even just for the token list / web console view.

src/inmanta/db/util.py
```diff
@@ -94,6 +94,14 @@ def buffer(self, cmd: str) -> None:
...
         if not cmd.strip():
             return
+        if cmd.startswith(("\\restrict", "\\unrestrict")):
+            # psql meta-commands emitted by pg_dump >= 16.10 / 17.6; they only have meaning inside psql
+            # and are not valid SQL.
+            return
+        if cmd.startswith("SET transaction_timeout"):
+            # Emitted by pg_dump 17+ clients but the parameter does not exist on older PostgreSQL
+            # servers. Dumps set it to 0 (the default), so skipping it changes nothing.
+            return
```
sanderr: I find this a bit weird to include here. Iirc the dump tool already strips / comments out some lines when generating the dump. Could we include this there instead?

src/inmanta/server/services/environmentservice.py
```diff
@@ -128,14 +129,17 @@ async def prestart(self, server: protocol.Server) -> None:
...
-        # design; revoke them (or issue tokens with an expiry) to remove them from the registry.
+        # Periodically prune stale token-registry entries (expired tokens and tokens revoked longer ago
+        # than the retention window) and the in-process jti cache so neither grows unbounded. Note:
+        # unrevoked, non-expiring (eternal) tokens are valid and are kept by design; revoke them (or
+        # issue tokens with an expiry) to remove them from the registry.
         self.schedule(self._cleanup_tokens, 3600, initial_delay=0, cancel_on_stop=False)
 
     async def _cleanup_tokens(self) -> None:
+        retention = datetime.timedelta(seconds=server_config.server_revoked_token_retention.get())
+        revoked_before = datetime.datetime.now(datetime.timezone.utc) - retention
```
sanderr: Not a requirement, just a small consistency suggestion (with how we do this in the rest of the code)
```suggestion
        revoked_before = datetime.datetime.now().astimezone() - retention
```

changelogs/unreleased/token-revoked-at-cleanup.yml
```diff
@@ -0,0 +1,15 @@
...
+  Record when an API token was revoked (revoked_at, exposed via environment_token_list) and periodically
+  delete revoked tokens from the token registry once they have been revoked for longer than the new
+  server.revoked-token-retention option (default thirty days).
+issue-nr: 10572
+change-type: minor
+destination-branches:
+  - master
+  - iso9
+sections:
+  feature: >
```
sanderr: If I understand correctly (not entirely sure I do), this PR isn't really a change versus anything that was already released? I mean, the changelog already contains a "feature" entry for "Added a token registry ...". I would expect that that covers both the base PR and this one.

sanderr (changes requested):
Minor comments only

src/inmanta/data/model.py
bartv: _Claude here, on behalf of @bartv._

Done in 4b03a6412. Dropped the `revoked` boolean entirely; `revoked_at` (`datetime | None`, NULL = not revoked) is now the single source of truth, used by the DTO, the ORM/DB column, the revoke query and the auth check. The migration drops the column after backfilling it. Kept the name `revoked_at` rather than `revoked` since it holds a timestamp. The matching web-console change (deriving the status from `revoked_at`) is in inmanta/web-console#7110.

src/inmanta/data/sqlalchemy.py
bartv: _Claude here, on behalf of @bartv._

Agreed. Expired tokens are now kept for auditing too, symmetric with revoked ones: `delete_stale` removes a token once it has been expired *or* revoked for longer than the retention window (so recently-expired tokens stay visible in the list / console view). Since the window now governs both, the option is renamed `server.revoked-token-retention` -> `server.token-retention`.

src/inmanta/db/util.py
bartv: _Claude here, on behalf of @bartv._

Good call - reverted the `PGRestore` change (both the `\restrict`/`\unrestrict` meta-commands and the `transaction_timeout` SET). The dump tool now comments those lines out at generation time, the same way it already handles the `search_path` / table-access-method lines, and I've commented them out in the two committed dumps so they restore cleanly without a runtime skip.

src/inmanta/server/services/environmentservice.py
bartv: _Claude here, on behalf of @bartv._

Applied.

changelogs/unreleased/token-revoked-at-cleanup.yml
bartv: _Claude here, on behalf of @bartv._

Agreed, dropped the changelog entry - the "Added a token registry ..." feature entry already covers this, since both land in the same unreleased cycle.

src/inmanta/data/model.py
bartv: _Claude here, on behalf of @bartv._

Correction: the matching web-console change moved to inmanta/web-console#7106 (the "Revoked at" column PR that depends on this one), not #7110. #7110 is stacked on top of it.

changelogs/unreleased/token-revoked-at-cleanup.yml
sanderr: We still need a yml file for the change, just not an entry in the changelog, i.e. no `sections` in the yml.

sanderr (changes requested):

changelogs/unreleased/token-revoked-at-cleanup.yml
bartv: Sorry, I missed this when I review the changes before commit

sanderr (approved):

changelogs/unreleased/token-revoked-at-cleanup.yml
bartv: _Claude here, on behalf of @bartv._

You're right, dropping the whole file went too far - the yml also carries the squash description and the destination-branches for the iso9 backport. Restored it without the sections block in f9df69266.

PR #10577, opened 2026-07-13
Make revocable tokens the default for environment_create_token

### LINKED ISSUE #10571: Make revocable API tokens the default
Follow-up to the token registry (#10535), from the demo-session feedback.

`environment_create_token` currently defaults to `idempotent=True` - a stateless, non-revocable token. Now that revocable tokens exist (registry + list/revoke), a revocable (non-idempotent) token should be the default.

### Scope
- Flip the default of `idempotent` in `environment_create_token` (both `methods_v2.py` and the handler in `environmentservice.py`) from `True` to `False`.
- Audit callers that rely on reproducible/idempotent tokens (e.g. `inmanta-cli token bootstrap`, any embedded-token flow) and make them pass `idempotent=True` explicitly.
- Changelog `minor` + an upgrade note (API behaviour change).

### Companion
web-console has its own default to flip (the create-token form always sends `idempotent` explicitly) - tracked in a separate web-console ticket.

bartv:
# Description

Follow-up to the token registry (#10535), from the demo-session feedback: now that tokens can be listed and revoked, a revocable token should be what you get by default.

* Flip the `idempotent` default from `True` to `False` on both the v1 `create_token` and v2 `environment_create_token` methods. Tokens created without an explicit choice now carry a `jti`, are tracked in the token registry, and can be listed and individually revoked.
* Caller audit: `inmanta-cli token bootstrap` encodes its token locally (not affected); `inmanta-cli token create` uses the server default, which is exactly the flow that should produce a revocable token; all in-tree callers that rely on reproducible tokens already pass `idempotent=True` explicitly.
* Behaviour change for API users that omit the argument (noted in the upgrade-note changelog section): the token value differs on every call, and the token expires according to the signing configuration's `expire` setting. Pass `idempotent=true` for the old behaviour.

The web console always sends `idempotent` explicitly; its own default flip is tracked in inmanta/web-console#7102.

closes #10571

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (API docstrings + upgrade note document the change)

### DIFF
--- a/changelogs/unreleased/token-revocable-default.yml
+++ b/changelogs/unreleased/token-revocable-default.yml
@@ -1,0 +1,16 @@
+description: >-
+  Make revocable tokens the default: environment_create_token (and the v1 create_token) now defaults to
+  idempotent=False, so tokens are registered in the token registry and can be listed and individually
+  revoked. Pass idempotent=true explicitly to keep creating reproducible, stateless tokens.
+issue-nr: 10571
+change-type: minor
+destination-branches:
+  - master
+  - iso9
+sections:
+  upgrade-note: >
+    Tokens created via environment_create_token (API, web console or inmanta-cli token create) without an
+    explicit idempotent argument are now revocable by default: they carry a jti claim, are tracked in the
+    token registry, differ on every call, and expire according to the signing configuration's expire
+    setting. Automation that relies on reproducible, non-expiring tokens must pass idempotent=true
+    explicitly.
--- a/src/inmanta/protocol/methods.py
+++ b/src/inmanta/protocol/methods.py
@@ -278,14 +278,15 @@ def delete_setting(tid: uuid.UUID, id: str):
     arg_options=ENV_OPTS,
     client_types=[const.ClientType.api, const.ClientType.compiler],
 )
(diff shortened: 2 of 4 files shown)


changelogs/unreleased/token-revocable-default.yml
```diff
@@ -0,0 +1,16 @@
+description: >-
+  Make revocable tokens the default: environment_create_token (and the v1 create_token) now defaults to
+  idempotent=False, so tokens are registered in the token registry and can be listed and individually
+  revoked. Pass idempotent=true explicitly to keep creating reproducible, stateless tokens.
+issue-nr: 10571
+change-type: minor
+destination-branches:
+  - master
+  - iso9
```
sanderr, line 9: As mentioned in the PR description, this is a behavioral change. But I assume it is intentionally included in iso9 anyway?

bartv: yeah, because the current behavior makes it impossible to revoke a single token without rotating the entire key and revoking all tokens.

sanderr (approved):

PR #10578, opened 2026-07-13
Add an expire parameter to environment_create_token

### LINKED ISSUE #5643: User management: Add an expire to the tokens and link them to a user
Currently user tokens do not expire and doesn't contains any information regarding the user it belongs to.

bartv:
# Description

Follow-up to the token registry (#10535), from the demo-session feedback: tokens can now be given an explicit lifetime at creation time.

* `environment_create_token` gets an optional `expire` parameter (seconds). It drives both the JWT `exp` claim (via `encode_token`, which already supported an explicit expire for the login endpoint) and the `expires_at` recorded in the token registry, so the expiry shows up in `environment_token_list` and the expired token is eventually purged by the registry cleanup.
* When `expire` is not set, the signing configuration's `expire` setting governs, as before.
* Validation: a non-positive `expire` is rejected, and so is combining `expire` with `idempotent=true` (idempotent tokens carry no time-based claims by definition).

Stacked on #10577 (both change the `environment_create_token` signature); the base is set to that branch and I'll retarget/rebase onto master once it merges. The web-console companion (expiry input + column in the Tokens tab) is tracked in inmanta/web-console#7104.

closes #5643

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (API docstring documents the parameter)

### DIFF
--- a/src/inmanta/server/services/environmentservice.py
+++ b/src/inmanta/server/services/environmentservice.py
@@ -587,6 +587,7 @@ class EnvironmentService(protocol.ServerSlice):
         # field is informational (for listing and cleanup); token expiry itself is enforced by the exp
         # claim, not by the registry.
         sign_cfg = auth.AuthJWTConfig.get_sign_config()
+        effective_expire: int | None
         match expire, sign_cfg:
             case int() as explicit_expire, _:
                 effective_expire = explicit_expire
--- a/src/inmanta/server/services/environmentservice.py
+++ b/src/inmanta/server/services/environmentservice.py
@@ -290,7 +290,7 @@ class EnvironmentService(protocol.ServerSlice):
         """
         Create a new auth token for this environment
         """
-        return 200, {"token": await self.environment_create_token(env, client_types, idempotent, None, context)}
+        return 200, {"token": await self.environment_create_token(env, client_types, idempotent, expire=None, context=context)}
 
     @handle(methods.list_settings, env="tid")
     async def list_settings(self, env: data.Environment) -> Apireturn:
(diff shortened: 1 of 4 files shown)


tests/server/test_user.py
```diff
@@ -601,6 +601,58 @@ async def test_token_registry(server: protocol.Server, auth_client: endpoints.Cl
...
 
 
+async def test_token_expire_param(server: protocol.Server, auth_client: endpoints.Client) -> None:
+    """
+    An explicit expire on environment_create_token drives both the JWT exp claim and the registry
+    expires_at; a non-positive expire and the idempotent+expire combination are rejected.
+    """
+    user = data.User(
+        username="admin",
+        password_hash=nacl.pwhash.str("adminadmin".encode()).decode(),
```
bartv: Will this non-compliant password work?

src/inmanta/server/services/environmentservice.py
```diff
@@ -286,7 +286,7 @@ async def create_token(
         """
         Create a new auth token for this environment
         """
-        return 200, {"token": await self.environment_create_token(env, client_types, idempotent, context)}
+        return 200, {"token": await self.environment_create_token(env, client_types, idempotent, None, context)}
```
arnaudsjs:
```suggestion
        return 200, {"token": await self.environment_create_token(env, client_types, idempotent, expire=None, context=context)}
```

src/inmanta/server/services/environmentservice.py
```diff
@@ -573,14 +577,19 @@ async def environment_create_token(
...
+        # Record when the token stops being valid, mirroring the exp that encode_token derives: an
+        # explicit expire argument wins, otherwise the signing config's expire governs. This field is
+        # informational (for listing and cleanup); token expiry itself is enforced by the JWT exp
         # claim, not by the registry.
         sign_cfg = auth.AuthJWTConfig.get_sign_config()
-        expires_at = now + datetime.timedelta(seconds=sign_cfg.expire) if sign_cfg is not None and sign_cfg.expire > 0 else None
+        if expire is not None:
+            expires_at = now + datetime.timedelta(seconds=expire)
+        elif sign_cfg is not None and sign_cfg.expire > 0:
+            expires_at = now + datetime.timedelta(seconds=sign_cfg.expire)
```
arnaudsjs: In this case we didn't set the `exp` claim on the token itself, but we do set an expire timestamp in the db. Shouldn't we keep both consistent?

arnaudsjs (changes requested):

tests/server/test_user.py
bartv: _Claude here, on behalf of @bartv._

It works - the user is created here by inserting a pre-computed hash, which bypasses the password policy (that's only enforced when *setting* a password), and `login` verifies against the stored hash without re-checking the policy. To avoid relying on that, I switched the test to a policy-compliant password (`Str0ng-Pass!`, as used elsewhere in this file).

src/inmanta/server/services/environmentservice.py
bartv: _Claude here, on behalf of @bartv._

Done.

src/inmanta/server/services/environmentservice.py
bartv: _Claude here, on behalf of @bartv._

They were actually already consistent: `encode_token` applies the same precedence, so with `expire=None` and a positive `cfg.expire` it does set `exp = now + cfg.expire`, matching the recorded `expires_at`. But you're right that the precedence was duplicated across two spots and could drift. Reworked it to resolve the effective lifetime once (a `match`) and feed that single value to both `encode_token` and `expires_at`, so they can't diverge:

```python
sign_cfg = auth.AuthJWTConfig.get_sign_config()
match expire, sign_cfg:
    case int() as explicit_expire, _:
        effective_expire = explicit_expire
    case None, cfg if cfg is not None and cfg.expire > 0:
        effective_expire = cfg.expire
    case _:
        effective_expire = None
token = encode_token(client_types, str(env.id), idempotent, expire=effective_expire, custom_claims=custom_claims)
now = datetime.datetime.now(tz=datetime.timezone.utc)
expires_at = now + datetime.timedelta(seconds=effective_expire) if effective_expire is not None else None
```

arnaudsjs (approved):

PR #10583, opened 2026-07-14
Use changelog entries of released version as release tag message

### LINKED ISSUE #7893: Insert released changelog entries in release tag description
Currently, the `inmanta module release` command supports manipulating the changelog file, the versions, and creating tags.  But the comment of each tag is always "auto tag by module tool".  We could make it a lot more interesting by replacing this message by the content of the changelog for the given release.

Any opinion?

edvgui:
# Description

The `inmanta module release` command previously always tagged stable releases with the generic message "auto tag by module tool". The tag message now contains the changelog entries of the released version when a CHANGELOG.md file is present (falls back to the old message when there is no changelog or the section is empty).

closes #7893

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (add ref to issue here: ) — the `inmanta module release` command help text was updated
- ~~If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)~~

### DIFF
--- a/changelogs/unreleased/7893-changelog-in-release-tag-message.yml
+++ b/changelogs/unreleased/7893-changelog-in-release-tag-message.yml
@@ -3,6 +3,6 @@ description: "The `inmanta module release` command now uses the changelog entrie
 issue-nr: 7893
 issue-repo: inmanta-core
 change-type: minor
-destination-branches: [master]
+destination-branches: [master, iso9]
 sections:
   minor-improvement: "{{description}}"
--- a/changelogs/unreleased/7893-changelog-in-release-tag-message.yml
+++ b/changelogs/unreleased/7893-changelog-in-release-tag-message.yml
@@ -1,0 +1,8 @@
+---
+description: "The `inmanta module release` command now uses the changelog entries of the released version as the message of the release tag, instead of the generic \"auto tag by module tool\" message."
+issue-nr: 7893
+issue-repo: inmanta-core
+change-type: minor
+destination-branches: [master]
+sections:
+  minor-improvement: "{{description}}"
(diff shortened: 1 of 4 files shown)


changelogs/unreleased/7893-changelog-in-release-tag-message.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: "The `inmanta module release` command now uses the changelog entries of the released version as the message of the release tag, instead of the generic \"auto tag by module tool\" message."
+issue-nr: 7893
+issue-repo: inmanta-core
+change-type: minor
+destination-branches: [master]
```
edvgui:
```suggestion
destination-branches: [master, iso9]
```

edvgui: This is Claude, on behalf of Guillaume: suggestion applied in 8d10a461f.

tests/moduletool/test_release.py
```diff
@@ -501,6 +517,43 @@ def test_populate_changelog(tmpdir, modules_dir: str, monkeypatch, top_level_hea
 """.lstrip()
 
 
+def test_release_tag_message_contains_changelog_entries(tmpdir, modules_dir: str, monkeypatch) -> None:
```
arnaudsjs: We are missing tests for the special cases:

* The change entry we need is the only entry in the file.
* The changelog section for the given version is empty.

arnaudsjs (changes requested):

tests/moduletool/test_release.py
edvgui: Oh yes, thanks, I will ask claude to add these

edvgui: Okay, should be better now

arnaudsjs (approved):

PR #10588, opened 2026-07-15
Run git non-interactively during server-side compiles

edvgui:
> Created by Claude on behalf of Guillaume.

# Description

When a server-side compile clones or pulls a project repository that requires authentication and no credentials are available, git tries to prompt for a username. The compiler runs git in a subprocess with no terminal attached, so the prompt fails with the confusing:

```
fatal: could not read Username for 'https://code.inmanta.com': No such device or address
```

which hides the actual problem (wrong/missing credentials).

The git subprocesses spawned by the compiler service (`_run_compile_stage` and `_run_cmd_async`) now run with `GIT_ASKPASS=true` and `GIT_TERMINAL_PROMPT=0` by default:

- `GIT_ASKPASS=true` feeds empty credentials, so the remote answers with its **real** authentication error (`fatal: Authentication failed for '...'`).
- `GIT_TERMINAL_PROMPT=0` disables any remaining interactive prompt as a safety net.

These are applied *below* the process environment, so they act as defaults an operator can override through the orchestrator environment (e.g. pointing `GIT_ASKPASS` at a real credential helper). This matches what `CLIGitProvider` already does for its own clone/fetch.

Verified locally against a minimal 401 HTTP server:

| env                     | git output                                                          |
| ----------------------- | ------------------------------------------------------------------- |
| none (before)           | `fatal: could not read Username ... No such device or address`      |
| `GIT_ASKPASS=true`      | `fatal: Authentication failed ...` ← real error                     |
| `GIT_TERMINAL_PROMPT=0` | `fatal: could not read Username ... terminal prompts disabled`      |

# Documentation

Added a "Debugging project authentication issues" subsection to `docs/troubleshooting.rst` explaining how to set `GIT_TRACE_CURL` / `GIT_CURL_VERBOSE` as orchestrator environment variables to see which credential source git consulted (`.netrc`, URL, credential helper) when a checkout fails to authenticate. The secret itself is redacted; the docs warn against `GIT_TRACE`, which prints the full repository URL and would leak URL-embedded credentials into the compile report.

closes N/A

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included

### DIFF
--- a/docs/model_developers/projects.rst
+++ b/docs/model_developers/projects.rst
@@ -20,3 +20,27 @@ A project is the basic unit of orchestration. It contains:
     |__ requirements.txt
     |__ main.cf
 
+
+Server-side checkout and authentication
+---------------------------------------
+
+For server-side compiles, the orchestrator obtains the project from a git repository. Each environment is
+configured with a repository URL and a branch: the orchestrator clones this repository into the environment's
+project directory on the first compile and pulls updates from it on subsequent compiles.
+
+When the repository requires authentication, git takes credentials from its usual sources: a ``.netrc`` file,
+credentials embedded in the repository URL, or a configured git credential helper. The simplest option is a
+``.netrc`` file in the server's home directory (``/var/lib/inmanta/.netrc``). This is the same file that is used
+to authenticate against a :ref:`private Python package repository<setting_up_pip_index_authentication>`, so a
+single file can cover both the project checkout and module installation:
+
+.. code-block:: text
+
+    machine <hostname of the git repository>
+    login <username>
+    password <password>
+
+If a checkout fails to authenticate, the compile report shows the ``Cloning repository`` or ``Pulling updates``
+step failing with an ``Authentication failed`` error. See :ref:`debugging_project_authentication` for how to find
+out which credential source git used.
+
--- a/docs/model_developers/projects.rst
+++ b/docs/model_developers/projects.rst
@@ -34,7 +34,10 @@ credentials the same way you would for a local ``git clone``. The most common me
 
 * **SSH keys**: use a ``git@host:...`` (or ``ssh://``) repository URL and place the private key and a matching
   ``known_hosts`` entry in the SSH configuration of the ``inmanta`` user (the user the orchestrator runs as),
-  under ``~/.ssh`` in its home directory (``/var/lib/inmanta/.ssh``).
+  under ``~/.ssh`` in its home directory (``/var/lib/inmanta/.ssh``). Make sure the key files and the ``.ssh``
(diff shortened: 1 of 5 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -188,7 +196,11 @@ async def _run_cmd_async(self, *cmd: str) -> str | None:
         sub_process: Process | None = None
         try:
             sub_process = await asyncio.create_subprocess_exec(
-                *cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=self._project_dir
+                *cmd,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                cwd=self._project_dir,
+                env={**os.environ, **GIT_NON_INTERACTIVE_ENV},
```
edvgui: Should we instead allow os.environ to overwrite the git non interactive env?

edvgui: _(Claude, on behalf of Guillaume)_

Good point — flipped the merge order so `GIT_NON_INTERACTIVE_ENV` is applied *below* `os.environ`. The values are now defaults that the orchestrator environment can override (e.g. pointing `GIT_ASKPASS` at a real credential helper). Done in 870d8df.

docs/model_developers/projects.rst
```diff
@@ -20,3 +20,35 @@ A project is the basic unit of orchestration. It contains:
...
+configured with a repository URL and a branch: the orchestrator clones this repository into the environment's
+project directory on the first compile and pulls updates from it on subsequent compiles.
+
+The orchestrator invokes ``git`` for these operations, so it transparently supports any authentication
+mechanism that ``git`` itself supports. No extra configuration is needed on the orchestrator side; you set up
+credentials the same way you would for a local ``git clone``. The most common mechanisms are:
+
+* **SSH keys**: use a ``git@host:...`` (or ``ssh://``) repository URL and place the private key and a matching
+  ``known_hosts`` entry in the SSH configuration of the ``inmanta`` user (the user the orchestrator runs as),
+  under ``~/.ssh`` in its home directory (``/var/lib/inmanta/.ssh``).
```
cukiernick: We may want to add a note about importance of file/dir permissions for client not to ignore ssh key (“Permissions Are Too Open”)

edvgui: I added a short mention, I don't want to get into too many details as our responsibility is not to educate our users about git and ssh

Hugo-Inmanta (approved):
LGTM, requesting review from @sanderr as well to err on the side of caution since this touches the compiler.

sanderr (approved):

PR #10593, opened 2026-07-15
Document concurrency control in the LSM API docs

bartv:
*This PR was created by Claude Code on behalf of Bart Vanbrabant.*

# Description

Add a dedicated "Concurrency control" section to the LSM service inventory documentation (`docs/lsm/index.rst`), directly after the CRUD operations overview.

The `current_version` parameter on the LSM API is currently only documented in the per-endpoint API reference, and a customer recently mistook the service instance `version` field for git-style version control. The new section explicitly explains that:

- the version is a change counter, not a version control mechanism: it cannot be used to check out, diff or restore older state (the log endpoint and the `rollback` lifecycle operation cover those needs),
- it implements optimistic concurrency control (compare-and-set) to prevent lost updates between concurrent actors (web console operators, northbound clients, the lifecycle manager itself),
- the expected client workflow: read the instance, pass its `version` as `current_version`, and on HTTP 409 Conflict re-read and re-evaluate before retrying, with a curl example.

It also cross-links the "On every state transfer the version of the service instance is incremented" sentence in the lifecycle section to the new section, and quotes the literal 409 error message so users can find this section from the error they see.

# Self Check:

- [x] ~~Attached issue to pull request~~ (docs improvement, no ticket)
- [x] Changelog entry
- [x] ~~Type annotations are present~~ (docs only)
- [x] ~~No (preventable) type errors (check using make mypy or make mypy-diff)~~ (docs only)
- [x] ~~Sufficient test cases (reproduces the bug/tests the requested feature)~~ (docs only)

### DIFF
--- a/docs/lsm/index.rst
+++ b/docs/lsm/index.rst
@@ -116,6 +116,59 @@ The service inventory exposes CRUD operations on service instances in the invent
 The state machine attached to the lifecycle will determine whether the API call is successful or not.
 
 
+.. _lsm_concurrency_control:
+
+Concurrency control
+===================
+
+Every service instance carries an integer ``version`` field, which is returned by every read operation. This version is a
+change counter: it is incremented on every state transfer of the instance. It is **not** a version control mechanism like
+git: older versions are not kept as retrievable snapshots, so the version number cannot be used to check out, diff or
+restore a previous state of the instance. To inspect the history of an instance, use the log endpoint listed under
+`Status operations`_. To revert attribute changes, use a lifecycle transfer with the ``rollback`` operation (see
+:ref:`fsm`).
+
+Instead, the version is used for optimistic concurrency control. A service instance can be modified by multiple actors at
+the same time: operators using the web console, northbound clients such as OSS/BSS systems, and the lifecycle manager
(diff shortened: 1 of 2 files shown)


docs/lsm/index.rst
```diff
@@ -116,6 +116,76 @@ The service inventory exposes CRUD operations on service instances in the invent
...
+
+Concurrency control
+===================
+
+Every service instance carries an integer ``version`` field, which is returned by every read operation. This version is a
+change counter: it is incremented on every state transfer of the instance. It is **not** a version control mechanism like
+git: older versions are not kept as retrievable snapshots, so the version number cannot be used to check out, diff or
+restore a previous state of the instance. To inspect the history of an instance, use the log endpoint listed under
+`Status operations`_. To revert attribute changes, use a lifecycle transfer with the ``rollback`` operation (see
+:ref:`fsm`).
```
edvgui: Or an update with the old attributes as input

bartv: *I'm Claude, replying on behalf of Bart.* Good point, addressed in 0c348c706: the sentence now mentions a plain update with the old attribute values first, with the rollback lifecycle operation as the alternative.

edvgui (approved):

Issue #10592, opened 2026-07-15
iso11 Follow-up: drop compatibility layer


Hugo-Inmanta:
[iso11 only] 
The main staged added a compatibility layer to be able to still deploy/dryrun model versions that were exported using an iso<10 orchestrator. 
This compatibility layer can be dropped under the assumption that all model versions in the DB use the "new style" code install. 

- CodeManager.get_code() can be cleaned up as well: ExecutorBlueprint no longer needs to hold the sources. They were only used by the iso9 compat layer and the editable_modules are transported separately.

PR #10609, opened 2026-07-23
Issue/improve agent code install stage2 install refinements

### LINKED ISSUE #10451: [Install refinement]
Refinement on the install: rebuild the module structure on the agent side, including the setup.cfg + pyproject.toml (DEV mode only)

Drop old code 'source' install
1) (Write path during export)
  - For all editable modules, we need to persist in the db the necessary files to be able to recreate a proper python package: `setup.cfg` `pyproject.toml`
2) (Read path during agent code install)
   - Add a step before 5.i. to populate the `<path to editable modules sources>`. It should download the module files from the database and recreate
        the module as a python package on the agent side. e.g.:
    ```commandline
          # Current structure:
           ls ~/executors/code/bb9ce25ffe586617daa8185a76de34f8/modules/std/plugins
               __init__.py  resources	types

          # What we want:

           ls ~/executors/code/bb9ce25ffe586617daa8185a76de34f8/modules/std/
               inmanta_plugins setup.cfg pyproject.toml

           ls ~/executors/code/bb9ce25ffe586617daa8185a76de34f8/modules/std/inmanta_plugins/std
               __init__.py  resources	types
    ```
   - Completely remove code install from step 5.ii. of old flow by making sure we install the relevant module sources in editable mode during step 5.i.
        (This step is already no longer relevant for prod mode) i.e. add the following to step 5.i.: "-e `<path to editable modules sources>`".
        Even though the 'install' part can be removed, the 'loading' part is still relevant (Same as step 1.ii.c above).
        This is an opportunity to remove some/all of the loader.PluginModuleFinder class.

- [ ] Make sure we properly log module installs on the agent
- [ ] Find a way around storing the full cross product [all agents] x [all editable install modules]. If a module is installed in editable mode, all agents should install it by default. [Relevant slack thread](https://inmanta.slack.com/archives/CKRF0C8R3/p1784884334157269)
- [ ] unify iso9 and iso10 cases in `deploy_and_load(...)` in `loader.py`
- [ ] Cleanup opportunity: remove ExecutorModuleSource class (see [this comment](https://github.com/inmanta/inmanta-core/pull/10468#discussion_r3683097319))

Hugo-Inmanta:
# Description

Stage 2 of the improved agent code install: how the code of an inmanta module reaches an executor becomes an explicit,
per-module **install mode**, and editable modules stop being faked on disk.

Until now the agent received a flat list of python sources and wrote all of them to disk, where a custom
`PluginModuleFinder` made them importable. That works, but it is not a real install: the module is not a python package
on the agent, so pip never sees its dependencies and the module cannot be imported the way it is in the compiler venv.

Each module now records how it was installed in the compiler venv, and the agent honors that:

| Install mode | Transported                             | How the executor installs it                                     |
| ------------ | --------------------------------------- | ---------------------------------------------------------------- |
| `package`    | nothing                                 | `pip install <pkg>==<version>` from the index                     |
| `editable`   | python sources + `setup.cfg`/`pyproject.toml` | reconstructed as a real python package under `<venv>/.inmanta/editable/<mod>` and `pip install -e`'d |
| `on_disk`    | python sources + python requirements    | written to disk outside the venv, imported through `PluginModuleFinder` |
| `unknown`    | python sources + python requirements    | same as `on_disk` (iso<10 compatibility, see below)               |

Because an editable module is now installed as a package, pip resolves its requirements from the `setup.cfg` it
installs, so those requirements are no longer transported. A V1 module is not distributed as a python package at all —
the agent can neither pip install it nor recreate it as one — so it uses `on_disk` and keeps carrying its requirements,
which are declared outside of any packaging metadata.

## Notable points

- **Venv identity.** Editable modules live *inside* the venv, which breaks the previous invariant that the venv pool
  key (`EnvBlueprint`) is independent of the executor's code. Each editable module's `(name, content-hash)` is folded
  into both `EnvBlueprint` and `ExecutorBlueprint` hashes, so a change to an editable module yields a new venv instead
  of mutating a shared one. A packaging-file-only change is invisible in `sources` but does change the content hash,
  which is why `ExecutorBlueprint` has to fold it in too.
- **Mixed install modes per executor.** `get_code` emits one spec per module with its own install mode, and
  `from_specs` merges them, so a project with a V1 module *and* an editable V2 module builds one executor.
- **iso<10 compatibility.** A model version exported by an older orchestrator did not record install modes. Those
  modules get `unknown`, which routes through the on-disk mechanism — the only one that works without that knowledge —
  but stays distinct from `on_disk`: an `on_disk` module is installed on every agent of the model version, while an
  `unknown` one keeps the narrower iso<10 behaviour of only reaching the agents that load it.
- **`PluginModuleFinder` is now configured lazily**, in `install_source` only, so the iso10 load path never installs
  the finder. It can be dropped along with the iso9 half of the compatibility layer in iso11 (#10592); the V1 on-disk
  half survives.

## Schema

Migration `v202608070` replaces `inmanta_module.editable_install` (boolean) with a NOT NULL `install_mode` column
(`true` → `editable`, `false` → `package`, `NULL` → `unknown`) and adds two nullable `setup_cfg_hash` /
`pyproject_toml_hash` columns referencing `file.content_hash`, so an editable module's packaging files are persisted
as first-class per-module metadata.

closes  https://github.com/inmanta/inmanta-core/issues/10451

# Self Check:

- [x] Changelog entry
- ~~If this PR fixes a race condition in the test suite, also push the fix to the relevant stable branche(s) (see [test-fixes](https://internal.inmanta.com/development/core/tasks/build-master.html#test-fixes) for more info)~~


src/inmanta/db/versions/v202503030.py
```diff
@@ -201,7 +201,9 @@ def compute_files_in_module(
             for cm_version, version_data in modules_per_version.model_versions.items():
                 for module_name, module_source_data in version_data.inmanta_modules.items():
                     module_version = CodeManager.get_module_version(
```
Hugo-Inmanta: I think we should inline `get_module_version` here so that this migration always produce the same module versions regardless of the future changes to `get_module_version` ? wdyt ?

PR #10610, opened 2026-07-23
Fix graphql stale relationship cache

jptrindade:
# Description

Build the `StrawberrySQLAlchemyLoader` per request instead of once per schema. 
The loader's `DataLoader` cache used to live for the whole server process and was never invalidated, so relationship data (e.g. a resource's state`) could be served from a stale snapshot cached by an earlier request.
    
Related to this discussion: https://inmanta.slack.com/archives/CKRF0C8R3/p1784793421341539

# Self Check:

- [x] Changelog entry


src/inmanta/graphql/schema.py
sanderr: I don't really see the point of this anymore tbh, though that's not new to this PR.

It would be different if it were the object being passed around inside the context dict, or if it *were* the context dict (`TypedDict`, probably the cleanest approach, though we probably couldn't get it typed correctly all the way through). But as it stands now it's just indirection without cause.

src/inmanta/graphql/schema.py
```diff
@@ -1559,4 +1563,4 @@ async def resource_summary(self, info: CustomInfo, environment: str) -> Composed
                 is_deploying=cast(JSON, results.is_deploying),
             )
 
-    return strawberry.Schema(query=Query, config=StrawberryConfig(info_class=CustomInfo))
+    return strawberry.Schema(query=Query, config=StrawberryConfig())
```
sanderr: Is this `config=...` still required? I'd assume this to be the default.

sanderr (approved):

PR #10611, opened 2026-07-23
Graphql version selection

sanderr:
Replaces the `ModelVersionSelection` mechanism with an alternative that simply relies on the filters to narrow the query to a single version (per resource). This simplifies the usage as illustrated in the test cases, most notably in `populate_sqlalchemy_columns` where the version is accessed and `apply_filter` where it is now set via a simple query rather than a Python construct that had to be managed carefully by the framework.

Additionally replaces `allows_persistent_state_count()`. I had some minor gripes with it that I meant to address separately, but turned out the model version filtering required more flexibility than it could offer: core's filter might handle version selection when `isOrphan` is set, but it can still do the optimized count. Only, for the optimized count, the version filters must be excluded since they act on `Configurationmodel`.

Finally, I added some todo's for things I noticed while going through the current code. As far as I'm concerned they're opportunities for cleanup that could be taken at the end of the track. I may add some more later. I'm in favor of merging these, and to go over all todo's before we consider the feature final. At that point we may still decide to leave well enough alone. @jptrindade wdyt? If we go this way, do you make sure we don't forget to take that final step?

*Code changes is me with Claude doing some minor legwork. Test changes are Claude with me doing some minor cleanup.*


src/inmanta/graphql/schema.py
```diff
@@ -899,11 +829,27 @@ def handles_version(self) -> bool:
         # means core owns version selection.
         return is_provided(self.is_orphan) or is_provided(self.model_version)
 
-    def resolve_model_version(self) -> ModelVersionSelection:
+    # TODO: clean this up: doesn't have to be dedicated method. If this filter promises to narrow to a single version
+    #       per resource (t_resource_set_configuration_model), it can simply be left to apply_filter()
```
jptrindade: I'm in favor of leaving it to apply_filter but no strong feelings about it either way

src/inmanta/graphql/schema.py
```diff
@@ -1506,29 +1452,29 @@ async def resources(
             for filter_instance in resource_filter_instances:
                 if filter_instance.handles_version():
                     if version_handler is not None:
+                        # TODO: we should try to be more informative
```
jptrindade, line 1510: I feel like that would mean some change to handles_version so that we know exactly why each extension is assuming control of the version selection.

jptrindade (approved):

src/inmanta/graphql/schema.py
```diff
@@ -1488,12 +1476,31 @@ async def resources(
...
+                    models.t_resource_set_configuration_model,
+                    and_(
+                        models.t_resource_set_configuration_model.c.environment == models.Resource.environment,
+                        models.t_resource_set_configuration_model.c.resource_set == models.Resource.resource_set,
+                    ),
+                )
+                .join(
+                    # Join Configurationmodel so that the version handler can select a model version. The join conditions
+                    # here ensure that it trickles down to the resoruces.
+                    models.Configurationmodel,
```
sanderr, line 1492: This is not strictly required, but I thought it might be simpler / more stable to expose this to extensions rather than `t_resource_set_configuration_model`. Wdyt?

src/inmanta/graphql/schema.py
```diff
@@ -927,14 +866,61 @@ def apply_filter[*Ts](
...
+            # 1 version: requested version
+            model_version = self.model_version
+        elif is_provided(self.is_orphan):
+            if self.is_orphan:
+                # 1 version per resource: its latest available version
+                model_version = models.ResourcePersistentState.orphaned_after
+            else:
+                # 1 version: latest scheduled version
+                model_version = self._latest_scheduled_version_subquery(self.environment)
+        else:
```
```suggestion
        elif self.is_orphan is True:
            # 1 version per resource: its latest available version
            model_version = models.ResourcePersistentState.orphaned_after
        elif self.is_orphan is False:
            # 1 version: latest scheduled version
            model_version = self._latest_scheduled_version_subquery(self.environment)
        else:
```

tests/test_graphql.py
```diff
@@ -1508,10 +1504,13 @@ async def resolved_versions(extra_filter: str) -> set[int | None]:
...
-    # A specific modelVersion is recorded on the query and echoed back by the contribution.
+    # A pinned modelVersion takes every resource exactly at that version.
     assert await resolved_versions("modelVersion: 1") == {1}
-    # The default selection resolves to the environment's latest scheduled version
-    assert await resolved_versions("") == {2}
+    assert await resolved_versions("modelVersion: 2") == {2}
+    assert await resolved_versions("isOrphan: false") == {2}
+    # The default selection takes non-orphaned resources at the latest scheduled version (2) and orphaned resources
+    # at the last version they were present in (1).
+    assert await resolved_versions("") == {1, 2}
```
jptrindade, line 1514: We should also assert "isOrphan: true"

src/inmanta/graphql/schema.py
```diff
@@ -927,14 +866,60 @@ def apply_filter[*Ts](
...
             stmt = stmt.filter(models.ResourcePersistentState.is_orphan.is_(self.is_orphan))
         return stmt
 
+    def apply_filter[*Ts](self, stmt: Select[tuple[*Ts]]) -> Select[tuple[*Ts]]:
+        stmt = self._apply_filter_rps(stmt)
+        if is_provided(self.purged):
+            stmt = stmt.filter(models.Resource.attributes["purged"].astext.cast(Boolean).is_(self.purged))
+
+        # Version selection. Coupled with self.handles_version(): in the case where the method returns False, the framework
+        # is expected to call filter_latest_available_version()
```
jptrindade: I think a check for `if self.handles_version()` might be cleaner. It would also be more in line with how we expect the other extensions to model this. Wdyt?

src/inmanta/graphql/schema.py
```diff
@@ -897,22 +825,33 @@ def validate_filter(self) -> None:
     def handles_version(self) -> bool:
         # is_orphan and model_version both control which version(s) of the model are selected, so providing either
         # means core owns version selection.
+        #
+        # Technically, is_orphan=True doesn't have to add a version filter. It could still accept an external version filter
+        # and act purely as a filter to exclude orphans. This would require only a small change to apply_filter(). However,
+        # it may bring more confusion than anything to expose it like that. So unless a use case comes up, we keep it simple
+        # and consider is_orphan=True to always select the latest version for each resource.
+        # TODO: error if both are provided (and not consistent)
```
jptrindade: This is already handled by `validate_filter` (called on `decompose_and_validate_filter`)

src/inmanta/graphql/schema.py
```diff
@@ -1349,6 +1335,7 @@ class ContributableGraphQLType:
 # The object types extensions can register GraphQL contributions for (see GraphQLContribution), mapping each SQLAlchemy
 # model to its core building blocks. `get_schema` composes each of these from the core building blocks and the
 # registered contributions; registrations for any other model are rejected.
+# TODO: do we want to restrict it to Resource only for now? Would it strip complexity?
```
jptrindade: I'm fine with keeping it like this. At least it is consistent between our supported models (and it is already implemented)

src/inmanta/graphql/schema.py
```diff
@@ -1158,6 +1143,7 @@ def get_sqlalchemy_columns(cls) -> "typing.Mapping[str, object]":
         """
         return {}
 
+    # TODO: does this really need to be a separate method? Can't this live inside the sqlalchemy class?
```
jptrindade: Which sqlalchemy class are you referring to? I'm not sure how it would work

src/inmanta/graphql/schema.py
sanderr: I was torn on this myself. I think we keep the coupling either way (just shift it), and I couldn't find a way to get rid of it completely. But perhaps it is a bit more controlled with the explicit wrapping. I'll add it.

src/inmanta/graphql/schema.py
sanderr: I still feel like it complicates `get_schema`, adds a bunch of methods that could otherwise be tied closer to the resource case, and I worry slightly about it becoming dead code. But each of these is only a small concern, so perhaps it's best that I just let this go.

src/inmanta/graphql/schema.py
sanderr: `get_sqlalchemy_columns` could return a class instead of a dict of columns. That would allow us to introduce more structure.

I'll rephrase this: I think once everything is in (including the lsm side), we should do a critical final pass about how everything fits together, and whether the are redundancies, verbosities, complexties, ... that we can clean up by tweaking the interface.

src/inmanta/graphql/schema.py
```diff
@@ -879,21 +878,23 @@ def apply_filter[*Ts](self, stmt: Select[tuple[*Ts]]) -> Select[tuple[*Ts]]:
...
-            # 1 version per resource: its latest available version
-            model_version = models.ResourcePersistentState.orphaned_after
-        elif self.is_orphan is False:
-            # 1 version: latest scheduled version
-            model_version = self._latest_scheduled_version_subquery(self.environment)
-        else:
-            model_version = None
-        if model_version is not None:
+        # Version selection. In the case where the method returns False, the framework is expected to call
+        # filter_latest_available_version() so there is no need to
```
jptrindade: this comment feel unfinished

src/inmanta/graphql/schema.py
```diff
@@ -1143,7 +1146,6 @@ def get_sqlalchemy_columns(cls) -> "typing.Mapping[str, object]":
         """
         return {}
 
-    # TODO: does this really need to be a separate method? Can't this live inside the sqlalchemy class?
```
jptrindade: Why is this TODO and the one below removed? I was under the impression that we would keep them and decide at the end

src/inmanta/graphql/schema.py
```diff
@@ -879,21 +878,23 @@ def apply_filter[*Ts](self, stmt: Select[tuple[*Ts]]) -> Select[tuple[*Ts]]:
...
+                model_version = self.model_version
+            elif self.is_orphan is True:
+                # 1 version per resource: its latest available version
+                model_version = models.ResourcePersistentState.orphaned_after
+            elif self.is_orphan is False:
+                # 1 version: latest scheduled version
+                model_version = self._latest_scheduled_version_subquery(self.environment)
+            else:
+                assert is_provided(self.is_orphan), "mismatch between handles_version() and apply_filter() implementation"
+                typing.assert_never(self.is_orphan)
```
jptrindade, line 899: Don't quite get this `assert_never` are we shielding against a different value being in `self.is_orphan`? I would just raise an error in this else saying there was a mismatch and logging `self.is_orphan` and `self.model_version`

arnaudsjs (approved):

src/inmanta/graphql/schema.py
sanderr: I rephrased this one and moved it to the top of the class.

src/inmanta/graphql/schema.py
sanderr: Will fix, thanks

src/inmanta/graphql/schema.py
sanderr: This is shielding against the inherent coupling. I chose `assert` instead of `raise ...` because it relates to an invariant of the class, not something under caller control. I believe we have an ADR about that somewhere.

The `assert_never` is simply a static proof that (assuming the type information is correct), that line is never reachable. So a proper exception there wouldn't really make sense either. It does raise an exception when it is reached at runtime btw.

jptrindade (approved):

PR #10623, opened 2026-07-28
Added support to validate authorization custom tornado handler

arnaudsjs:
# Description

Added support to validate the authorization of API requests towards endpoints implemented using a custom Tornado handler. This is required for the `GET /api/v2/support` endpoint on the support extension.

Part of inmanta/inmanta-support#985

# Self Check:

- [x] Changelog entry


changelogs/unreleased/support-custom-tornado-handler.yml
```diff
@@ -0,0 +1,6 @@
+---
+description: "Added support to validate the authentication/authorization of API call to endpoints using a custom tornado handler."
+issue-nr: 985
+issue-repo: inmanta-support
+change-type: minor
+destination-branches: [master]
```
Hugo-Inmanta: Was iso9 left out on purpose ?

src/inmanta/protocol/auth/providers.py
```diff
@@ -146,37 +175,48 @@ async def stop(self) -> None:
...
             # For ease of use, the health endpoint should be accessible without authentication.
             return
-        if call_arguments.auth_token is None:
+        if auth_token is None:
             raise exceptions.UnauthorizedException()
-        if call_arguments.is_service_request():
+        if auth.is_service_token(auth_token):
             # Service (machine-to-machine) requests always use the legacy provider.
-            await self._legacy_authorization_provider.authorize_request(call_arguments)
+            await self._legacy_authorization_provider._do_authorize_request(
```
Hugo-Inmanta, line 196: I'm not sure how relevant this is, but by calling `_do_authorize_request` instead of `authorize_request` we no longer check that the legacy provider was properly started i.e. 
```
    ...
    async def authorize_request
    ....
        if not self.running:
            raise Exception("Authorization provider was not started.")
```

Hugo-Inmanta (changes requested):
LGTM, 2 minor questions only

changelogs/unreleased/support-custom-tornado-handler.yml
arnaudsjs: We need this change to allow the streaming endpoint on the support extension to check the authorization of the requests it receives. This can be considered a breaking change or a bugfix. I took the conservative approach and considered it a breaking change. Hence, it's only targeted at master. Since the changes on this branch are more generic, we could also merge this into iso9, but I don't like to add code that we do not plan to use so I left it out.

src/inmanta/protocol/auth/providers.py
arnaudsjs: It's an invariant of the class that the legacy provider is started when self is started so I don't think this is an issue.

Hugo-Inmanta (approved):

changelogs/unreleased/support-custom-tornado-handler.yml
arnaudsjs: But maybe we should consider this a bugfix given that it's security related.

arnaudsjs: @sanderr Do you have an opinion on this?

sanderr: Not a strong one, but I suspect Bart will. I'm leaning towards bugfix myself, and I seem to recall Bart wanted to go that way as well, but I dare not say for sure.

arnaudsjs: Ok. I am also starting to lean in that direction. I'll consider it a bugfix.

sanderr: Do you make a note to double check that Bart doesn't have any objections when he's back?

src/inmanta/protocol/auth/auth.py
```diff
@@ -167,6 +171,33 @@ def decode_token(token: str) -> tuple[claim_type, "AuthJWTConfig"]:
...
 
 
+async def validate_token(
+    auth_token: claim_type | None, is_auth_enabled: bool, method_properties: "common.MethodProperties"
+) -> None:
+    """
+    Validate whether the given token is a valid token.
+    """
+    if not is_auth_enabled:
+        return
```
sanderr: *Subjective, not a strict change request.*

I find it a bit strange to introduce a parameter purely for the sake of exiting early. I would think that if a caller has to know about `is_auth_enabled` anyway, to properly pass it, it might as well not call the method. The semantics are simple enough that the "reasoning" can be moved to the caller imo.

Same for `parse_and_validate_auth_token`.

sanderr (changes requested):

src/inmanta/protocol/auth/auth.py
arnaudsjs: I agree. I felt the same, but for some reason I forgot to change it.

sanderr (approved):

PR #10636, opened 2026-07-31
Fix venv inheritance pip 26.2

arnaudsjs:
# Description

Fix incompatibility with pip 26.2 that breaks the venv inheritance. The inmanta-inherit-from-parent-venv.pth now only does addsitedir() calls for directories that are part of the parent venv and no other directories like python stdlib. (reference: https://github.com/pypa/pip/pull/14033)

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -817,7 +817,19 @@ class PythonEnvironment:
         activate the parent venv. The site directories of the parent venv should appear later in sys.path than the ones of
         this venv.
         """
-        site_dir_strings: list[str] = ['"' + p.replace('"', r"\"") + '"' for p in list(sys.path)]
+        site_dir_strings: list[str]
+        if sys.prefix == sys.base_prefix:
+            # We are running in the system environment.
+            # No need to setup venv inheritance.
+            site_dir_strings = []
+        else:
+            # Path prefix for the currently used venv.
+            venv_dir_prefix = f"{os.path.normpath(sys.prefix)}/"
+            # Fetch all paths in sys.path for the currently used venv.
+            site_dir_strings = [
+                '"' + p.replace('"', r"\"") + '"' for p in list(sys.path) if os.path.normpath(p).startswith(venv_dir_prefix)
+            ]
+        # Make sure the new venv inherits from the currently used venv by calling into addsitedir().
         add_site_dir_statements: str = "\n".join(
             [f"site.addsitedir({p}) if {p} not in sys.path else None" for p in site_dir_strings]
         )
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -820,7 +820,7 @@ class PythonEnvironment:
         this venv.
         """
         # Path prefix for the currently used venv.
-        venv_dir_prefix = f"{os.path.realpath(sys.prefix)}/"
+        venv_dir_prefix = os.path.join(os.path.realpath(sys.prefix), "")
         # Fetch all paths in sys.path for the currently used venv.
         site_dir_strings: list[str] = [
             '"' + p.replace('"', r"\"") + '"' for p in list(sys.path) if os.path.realpath(p).startswith(venv_dir_prefix)
--- a/src/inmanta/env.py
+++ b/src/inmanta/env.py
@@ -820,10 +820,10 @@ class PythonEnvironment:
         this venv.
         """
         # Path prefix for the currently used venv.
-        venv_dir_prefix = f"{os.path.normpath(sys.prefix)}/"
+        venv_dir_prefix = f"{os.path.realpath(sys.prefix)}/"
         # Fetch all paths in sys.path for the currently used venv.
         site_dir_strings: list[str] = [
-            '"' + p.replace('"', r"\"") + '"' for p in list(sys.path) if os.path.normpath(p).startswith(venv_dir_prefix)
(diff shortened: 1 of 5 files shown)


src/inmanta/server/services/compilerservice.py
```diff
@@ -101,6 +101,10 @@ def is_blocking(self) -> bool:
 class CompileRun:
     """Class encapsulating running the compiler."""
 
+    # The version number written into the .inmanta_venv_version file
+    # of the compiler venv.
+    COMPILER_VENV_VERSION = 2
```
sanderr: Why have all this here rather than in `inmanta.env.init_env` where we create the pth file?

src/inmanta/env.py
```diff
@@ -753,8 +753,7 @@ def init_env(self) -> None:
             path = os.path.realpath(self.env_path)
             try:
                 venv.create(path, clear=True, with_pip=False)
-                self._write_pip_binary()
-                self._write_pth_file()
+                self.write_inmanta_managed_files_in_venv()
```
sanderr: Not new to this PR, but isn't this redundant? We write it further down in `if not os.path.exists(...)`.

src/inmanta/env.py
```diff
@@ -817,7 +819,13 @@ def _write_pth_file(self) -> None:
         activate the parent venv. The site directories of the parent venv should appear later in sys.path than the ones of
         this venv.
         """
-        site_dir_strings: list[str] = ['"' + p.replace('"', r"\"") + '"' for p in list(sys.path)]
+        # Path prefix for the currently used venv.
+        venv_dir_prefix = f"{os.path.normpath(sys.prefix)}/"
```
sanderr:
```suggestion
        venv_dir_prefix = os.path.join(os.path.normpath(sys.prefix), "")
```

Not that we're very cross platform compatible, but still good practice?

There is also `os.path.commonpath`, but I think that has its own complications.

src/inmanta/env.py
```diff
@@ -817,7 +819,13 @@ def _write_pth_file(self) -> None:
...
+        # Fetch all paths in sys.path for the currently used venv.
+        site_dir_strings: list[str] = [
```
sanderr, line 854: How confident are you that this suffices, e.g. in case of editable installs or for running the tests? When I run this locally, I get (note the last entry)
```
>>> sys.path
['', '/home/sander/.pyenv/versions/3.13.11/lib/python313.zip', '/home/sander/.pyenv/versions/3.13.11/lib/python3.13', '/home/sander/.pyenv/versions/3.13.11/lib/python3.13/lib-dynload', '/home/sander/.virtualenvs/inmanta-core-313/lib/python3.13/site-packages', '/home/sander/documents/projects/inmanta/inmanta-core/src']
```

sanderr (changes requested):
This will be a lot cleaner, regardless of pip's behavior.

src/inmanta/env.py
arnaudsjs: Cool! I didn't know this trick

src/inmanta/env.py
arnaudsjs: I tested it and it works. Editable installs in the parent are implemented by adding a pth file to that parent venv. The pth file we add to the child venv makes sure that the pth files from the parent are read. So recursive venv inheritance is handled as well as editable installs in one of the parent venvs.

src/inmanta/env.py
arnaudsjs: Yes. I think so. Good catch.

src/inmanta/env.py
sanderr: > Editable installs in the parent are implemented by adding a pth file to that parent venv

That's one of three methods, with no guarantee as to which will be used. But I do think the others are also compatible.

Now that I think about it the pth file is probably the reason that the src dir shows up in `sys.path`... So I think this will be fine.

src/inmanta/server/services/compilerservice.py
arnaudsjs: That would be better indeed. This is caused by a previous implementation. I'll clean it up. Thanks!

src/inmanta/env.py
arnaudsjs: I was not actively aware about the two other methods. Can I count on your estimation that it will work or do I need to dig deeper into this?

sanderr: Let me double check the docs

sanderr: Apparently it's even an non-exhaustive list, which makes it more difficult to estimate. Then again, they have to hook into the venv somehow, so I think that if we cover these three generically we're probably safe: https://setuptools.pypa.io/en/latest/userguide/development_mode.html#how-editable-installations-work

1. pth file pointing to source
2. pth file pointing to a directory containing a "farm of file links that mimic the project structure"
3. pth file that installs a custom finder and loader
4. not in the docs but I'd swear I've seen it in the past: file links that mimic the project structure directly in the site-packages dir.

I think all of these should work.

src/inmanta/env.py
sanderr, line 790: Wouldn't it be simpler to check `or not is_venv_version_up_to_date()` here, and to simply write the file in `write_inmanta_managed_files_in_venv()`? We wouldn't need `update_venv_version()`, it would be transparent for the compiler service and it sits right next to the other inmanta-specific venv files.

sanderr (approved):

PR #10669, opened 2026-08-10
Use a pickle dispatch table for the parser AST cache

bartv:
Isolates the parser-cache pickling change from the Lark port (#10053) so it can land, and be measured, on its own. Rebased on current master, so the numbers below are relative to the GC threshold change (#10667) already being in.

## The problem

The AST contains `Namespace` objects, which are tied to the live compile and must not be written to a cache file. `ASTPickler` stripped them out with `persistent_id`, which has two defects:

1. **It is a Python callback invoked for every object in the graph**, not just namespaces. The AST has ~82k reachable objects per module set.
2. **It is consulted before the memo lookup.** The namespace therefore never reaches pickle's back-reference table, so it is re-emitted in full at every occurrence instead of being referenced.

Opcode counts for one `.cf` file (`lsm/_init.cf`):

| | opcodes | `BINPERSID` | `persistent_load` calls on read |
| --- | ---: | ---: | ---: |
| before | 36,652 | 733 | 733 |
| after | 33,729 | 0 | 0 |

Defect 1 costs write time. Defect 2 costs write time, file size, and read time.

## The change

A `dispatch_table` entry replaces `persistent_id`. It is a C-level lookup keyed on exact type, consulted only for types in the table, and its result participates in the memo, so the namespace is written once and back-referenced after.

That leaves the stream referencing a module-level `_restore_namespace`, which has to be resolved against the namespace currently being loaded into. The obvious way to do that is to override `find_class`, but `find_class` resolves *every* global in the stream, so overriding it in Python routes every AST class lookup through a Python frame and off CPython's internal fast path. Measured, that costs about as much as the memoisation saves: the read path comes out flat.

Publishing the namespace on a `ContextVar` for the duration of `load()` avoids that. The C unpickler keeps its native lookup path, and the cost is two operations per file rather than one per global. A `ContextVar` rather than a thread local so that concurrent and re-entrant loading both stay correct.

## Measured

Cache time per benchmark compile, 3 runs per variant on an idle machine, against master:

| benchmark | write | read | cache total | timed region |
| --- | ---: | ---: | ---: | ---: |
| connect_demo | -59% | -58% | -58% | -40.5% |
| connect_infra | -31% | -63% | -52% | -32.2% |
| systemtenant | -55% | -40% | -46% | -24.6% |
| inmanta_infra | -59% | -9% | -34% | ~0 |
| connect_enet | -44% | -8% | -32% | ~0 |

The two benchmarks showing no end-to-end movement are the ones where the parser cache is a small share of the compile to begin with (0.6s and 0.17s respectively), so there is little to win. The gain scales with compiles-per-session times module-set size, which is the shape of a `pytest-inmanta` module test suite.

Isolated, both implementations over 112 real `.cf` files (~82k objects), interleaved with a `gc.collect()` before each timed section: write 262ms to 129ms, cache file 6.84MB to 6.49MB.

## Cache format

The format changes, in both directions. An entry written by an older version fails to load with `UnpicklingError`, and vice versa. `CacheManager` already treats any load failure as a miss, so the entry is discarded, the file is reparsed, and the cache rewritten. The `.cfc` filename is keyed on the inmanta version, so a release upgrade never sees a stale entry; within one version string (branch switching on a dev install) the first compile logs one warning per `.cf` file and then self-heals.

`_restore_namespace` raises when called outside an `ASTUnpickler`, so a cache file cannot be coerced into yielding a `Namespace` through a plain `pickle.load`.

## Testing

`tests/compiler/` + `tests/test_parser.py`: 847 passed, 1 skipped, 2 xfailed.

Five tests added to `tests/compiler/test_parser_cache.py`, covering round-trip, namespace mismatch, the plain-`Unpickler` path, nested loads, and context cleanup after `load()`. Verified by mutation: removing the `ContextVar` token reset fails three of them.

### DIFF
--- a/src/inmanta/parser/pickle.py
+++ b/src/inmanta/parser/pickle.py
@@ -1,5 +1,5 @@
 """
-Copyright 2020 Inmanta
+Copyright 2026 Inmanta
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -14,34 +14,83 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 Contact: code@inmanta.com
+
+Custom pickler/unpickler for AST Statement objects.
+
+Namespace objects cannot be pickled (they're tied to runtime compilation context),
+so they are replaced with their fully-qualified name string during pickling and
+restored from the unpickler instance during unpickling.
 """
 
-from io import BytesIO
+import contextvars
+import copyreg
+import types
 from pickle import Pickler, Unpickler, UnpicklingError
-from typing import Optional
+from typing import IO, Callable
 
 from inmanta.ast import Namespace
 
+# Namespace the active ASTUnpickler is restoring into. A ContextVar rather than a
+# thread local so that concurrent and re-entrant unpickling both stay correct.
+_current_namespace: contextvars.ContextVar[Namespace] = contextvars.ContextVar("ast_unpickler_namespace")
+
+
+def _reduce_namespace(
+    ns: object,
+) -> tuple[Callable[..., object], tuple[str]]:
+    """Reducer for Namespace objects, replaces with full name string."""
+    assert isinstance(ns, Namespace)
+    return (_restore_namespace, (ns.get_full_name(),))
+
+
+def _restore_namespace(full_name: str) -> Namespace:
+    """Restore a Namespace during unpickling.
+
+    This module-level function is referenced in the pickle stream via dispatch_table.
+    It resolves the namespace from the ASTUnpickler that is currently loading, so the
+    C unpickler keeps its native find_class fast path for every other global.
+
+    If called directly (outside ASTUnpickler), raises UnpicklingError.
+    """
+    namespace: Namespace | None = _current_namespace.get(None)
+    if namespace is None:
+        raise UnpicklingError(f"_restore_namespace({full_name!r}) called outside ASTUnpickler context")
+    if namespace.get_full_name() != full_name:
+        raise UnpicklingError(f"Namespace mismatch: expected {namespace.get_full_name()}, got {full_name}")
+    return namespace
+
 
 class ASTPickler(Pickler):
-    def persistent_id(self, obj: object) -> Optional[tuple[str, str]]:
-        if isinstance(obj, Namespace):
-            # Don't pickle namespaces
-            return ("Namespace", obj.get_full_name())
-        else:
-            return None
+    """Pickler that replaces Namespace objects with their fully-qualified name.
+
+    Uses dispatch_table instead of persistent_id. persistent_id is a Python callback
+    invoked for every object in the graph, and it runs before the memo lookup, so a
+    namespace is re-emitted at each occurrence rather than referenced. dispatch_table
+    is a C-level type lookup and its result is memoized: roughly 2x faster to write,
+    2x faster to read back, and a 5% smaller cache file.
+    """
+
+    dispatch_table = types.MappingProxyType({**copyreg.dispatch_table, Namespace: _reduce_namespace})
+
+    def __init__(self, file: IO[bytes], protocol: int = 4) -> None:
+        super().__init__(file, protocol=protocol)
 
 
 class ASTUnpickler(Unpickler):
-    def __init__(self, file: BytesIO, namespace: Namespace) -> None:
-        super().__init__(file)
-        self.namespace = namespace
-        self.namespace_name = namespace.get_full_name()
+    """Unpickler that restores Namespace objects from the namespace being loaded into.
 
-    def persistent_load(self, pid: tuple[str, str]) -> object:
-        type_tag, key_id = pid
-        if type_tag == "Namespace":
-            assert self.namespace_name == key_id
-            return self.namespace
-        else:
-            raise UnpicklingError("unsupported persistent object")
+    Publishes the namespace on a ContextVar for the duration of load() rather than
+    overriding find_class(). Overriding find_class() would route every global lookup
+    in the stream through a Python callback, which costs more than it saves.
+    """
+
+    def __init__(self, file: IO[bytes], namespace: Namespace) -> None:
+        super().__init__(file)
+        self._namespace = namespace
+
+    def load(self) -> object:
+        token = _current_namespace.set(self._namespace)
+        try:
+            return super().load()
+        finally:
+            _current_namespace.reset(token)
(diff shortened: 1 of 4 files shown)


src/inmanta/parser/pickle.py
```diff
@@ -14,34 +14,83 @@
 limitations under the License.
 
 Contact: code@inmanta.com
+
+Custom pickler/unpickler for AST Statement objects.
+
+Namespace objects cannot be pickled (they're tied to runtime compilation context),
+so they are replaced with their fully-qualified name string during pickling and
+restored from the unpickler instance during unpickling.
```
jptrindade, line 22: I think this is rather uncommon in our code base, but I like it

src/inmanta/parser/pickle.py
```diff
@@ -14,34 +14,83 @@
...
+import types
 from pickle import Pickler, Unpickler, UnpicklingError
-from typing import Optional
+from typing import IO, Callable
 
 from inmanta.ast import Namespace
 
+# Namespace the active ASTUnpickler is restoring into. A ContextVar rather than a
+# thread local so that concurrent and re-entrant unpickling both stay correct.
+_current_namespace: contextvars.ContextVar[Namespace] = contextvars.ContextVar("ast_unpickler_namespace")
```
jptrindade: We try to avoid methods/variables starting with an underscore if they are not explicitly private methods of a class

tests/compiler/test_parser_cache.py
```diff
@@ -61,3 +70,107 @@ def test_caching(snippetcompiler):
...
+    ASTPickler(buf, protocol=4).dump(stmts)
+
+    buf.seek(0)
+    restored = ASTUnpickler(buf, ns).load()
+    assert isinstance(restored, list)
+    assert len(restored) == len(stmts)
+    for orig, rest in zip(stmts, restored):
+        assert type(orig) is type(rest)
+        assert isinstance(rest, Statement)
+        assert rest.namespace is ns
```
jptrindade, line 106: can we also compare the stmts themselves?

tests/compiler/test_parser_cache.py
```diff
@@ -61,3 +70,107 @@ def test_caching(snippetcompiler):
...
+
+    buf = io.BytesIO()
+    ASTPickler(buf, protocol=4).dump(parser.base_parse(ns, "test", "x = 1"))
+    buf.seek(0)
+
+    with pytest.raises(UnpicklingError, match="outside ASTUnpickler"):
+        pickle.Unpickler(buf).load()
+
+
+def _make_namespace(name: str) -> Namespace:
```
jptrindade: I would remove the starting underscores here and in the methods/class below. (Or put them inside the test that uses them)

src/inmanta/parser/pickle.py
```diff
@@ -14,34 +14,83 @@
...
+    invoked for every object in the graph, and it runs before the memo lookup, so a
+    namespace is re-emitted at each occurrence rather than referenced. dispatch_table
+    is a C-level type lookup and its result is memoized: roughly 2x faster to write,
+    2x faster to read back, and a 5% smaller cache file.
+    """
+
+    dispatch_table = types.MappingProxyType({**copyreg.dispatch_table, Namespace: _reduce_namespace})
+
+    def __init__(self, file: IO[bytes], protocol: int = 4) -> None:
+        super().__init__(file, protocol=protocol)
```
jptrindade: Do we actually need this `__init__` ?

tests/compiler/test_parser_cache.py
jptrindade: and possibly reuse this method for the test cases above

jptrindade (changes requested):

src/inmanta/parser/pickle.py
bartv: *Claude here on behalf of @bartv.*

Thanks. Worth recording why it ended up this way: the obvious alternative was overriding `find_class`, which is what the Lark branch does. That turned out to cost roughly what the change saves, because `find_class` resolves *every* global in the stream, so it puts a Python callback on all of them. The `ContextVar` costs two operations per file instead. Measured, that's the difference between the read path being flat and being 40-63% faster.

src/inmanta/parser/pickle.py
bartv: *Claude here on behalf of @bartv.*

Renamed to `current_namespace`, `reduce_namespace` and `restore_namespace` in db8f83cfd. Left `ASTUnpickler._namespace` as-is since it is class-private.

One thing to be aware of: `restore_namespace`'s qualname is written into the pickle stream, so the rename shifts the cache format. The parent commit already changes that format, so there's no extra cost here — but it does mean this rename couldn't be deferred to a follow-up without a second format break.

src/inmanta/parser/pickle.py
bartv: *Claude here on behalf of @bartv.*

No. Removed in db8f83cfd. Both call sites (`cache.py:136` and the tests) pass `protocol=4` explicitly, so it only restated a default.

Checked it doesn't disturb the mypy baseline — the baselined `arg-type` entry this PR removes is on `ASTUnpickler`, whose `__init__` stays.

tests/compiler/test_parser_cache.py
bartv: *Claude here on behalf of @bartv.*

Added in db8f83cfd. `Statement` has no `__eq__` (it's identity), so the comparison is on the rendered form and the source location:

```python
assert str(rest) == str(orig)          # 'x = 1', "y = 'hello'"
assert rest.location.file == orig.location.file
assert rest.location.lnr == orig.location.lnr
```

`str()` renders the reconstructed node, so a mangled operand or literal shows up.

Being upfront about the limit of this: two structurally different ASTs could in principle render to the same string, so it is weaker than a full attribute-by-attribute walk. I stopped here because there is no generic AST-walk helper in the test suite to reuse. Happy to add one if you would rather have the stricter check.

tests/compiler/test_parser_cache.py
bartv: *Claude here on behalf of @bartv.*

Both done in db8f83cfd: renamed to `make_namespace`, `pickled`, `NestedLoad`, `load_inner`, hoisted the two functions to the top of the file, and used them in the three earlier tests. Those each built a namespace and pickled inline, so the reuse removes about 30 lines.

jptrindade (approved):

PR #10672, opened 2026-08-10
Only import logfire when tracing is enabled

bartv:

This PR takes the stubbing we did a few years ago a bit further so that no loading happens at all. Impact is mostly on faster tests.

Importing `logfire` pulls in `opentelemetry` and `rich`, costing roughly 200 ms of process startup. Without a `LOGFIRE_TOKEN` none of it does anything, so the imports in `tracing.py` and `logging.py` move under the existing token check, and `enable()` imports logfire itself.

Guarding those imports alone achieves nothing: logfire's `logfire-plugin` pydantic entry point is autoloaded on the first `BaseModel` subclass and pulls it straight back in. `tests/conftest.py` therefore sets `PYDANTIC_DISABLE_PLUGINS` when no token is set, where the cost was paid on every spawned subprocess.

On `tests/test_app_cli.py`, 38 of 40 interpreter starts loaded logfire before and 1 after, 52.9 s to 45.7 s. Dev/CI only: logfire ships in the optional `tracing` extra.

For review: a process calling `tracing.enable()` late, rather than setting `LOGFIRE_TOKEN` at startup, now loses pydantic instrumentation. Nothing does that today.

### DIFF
--- a/src/inmanta/tracing.py
+++ b/src/inmanta/tracing.py
@@ -25,20 +25,20 @@ LOGGER = logging.getLogger("inmanta")
 
 # We need this early to make @instrument work
 enabled = os.getenv("LOGFIRE_TOKEN", None) is not None
-try:
-    import logfire._internal.config
-    import logfire.integrations
-    import logfire.integrations.pydantic
-    import logfire.propagate
-    from opentelemetry.instrumentation.asyncpg import AsyncPGInstrumentor
+if enabled:
+    # Importing logfire costs ~200ms and pulls in opentelemetry and rich, so only do it when it is actually going to be used.
+    try:
+        import logfire._internal.config
+        import logfire.integrations
+        import logfire.integrations.pydantic
+        import logfire.propagate
+        from opentelemetry.instrumentation.asyncpg import AsyncPGInstrumentor
 
-    # Make sure we don't get warnings when it is off
-    logfire._internal.config.GLOBAL_CONFIG.ignore_no_config = True
-    enabled = os.getenv("LOGFIRE_TOKEN", None) is not None
+        # Make sure we don't get warnings when it is off
+        logfire._internal.config.GLOBAL_CONFIG.ignore_no_config = True
 
-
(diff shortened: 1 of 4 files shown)


src/inmanta/tracing.py
```diff
@@ -25,20 +25,20 @@
...
+    # Importing logfire costs ~200ms and pulls in opentelemetry and rich, so only do it when it is actually going to be used.
+    try:
+        import logfire._internal.config
+        import logfire.integrations
+        import logfire.integrations.pydantic
+        import logfire.propagate
+        from opentelemetry.instrumentation.asyncpg import AsyncPGInstrumentor
 
+        # Make sure we don't get warnings when it is off
+        logfire._internal.config.GLOBAL_CONFIG.ignore_no_config = True
```
sanderr, line 38: Previously, we set this warning-related setting whether or not logfire is enabled. Now we only set it when it is (because we can't otherwise). But from the PR description I understand that Pydantic might still pull logfire in even if it is disabled.

I don't have enough context on this warning to judge if this is an issue.

tests/conftest.py
```diff
@@ -16,6 +16,14 @@
 Contact: code@inmanta.com
 """
 
+import os
+
+if os.getenv("LOGFIRE_TOKEN") is None:
+    # Logfire registers a pydantic entry point plugin, which pydantic autoloads on the first BaseModel subclass. That
+    # pulls in logfire, opentelemetry and rich on every interpreter start, including every subprocess the tests spawn.
+    # Without a token none of it does anything, so suppress it. Must happen before the first pydantic model is created.
+    os.environ.setdefault("PYDANTIC_DISABLE_PLUGINS", "logfire-plugin")
```
sanderr, line 25: This seems a bit broad to me. If we're confident that we don't use any other plugins I'm fine with it. But some day we may need to find an alternative.

sanderr (changes requested):

PR #10673, opened 2026-08-10
Defer heavy imports out of the compiler startup path

bartv:

Every `inmanta` invocation, including a plain `inmanta compile`, imported strawberry, cookiecutter, the server bootloader and the resource scheduler. The compiler uses none of them. This imports them where they are actually needed.

## Effect

`import inmanta.app`, measured as the min of 5 `-X importtime` runs:

| | modules | import work |
| ------- | ------- | ----------- |
| master  | 1738    | 793.2 ms    |
| this PR | 1385    | 696.7 ms    |

**-96.5 ms and -353 modules**; wall-clock min drops by roughly 140 ms. Removed self-time by package: strawberry 23.8 ms, graphql 20.7 ms, inmanta 17.6 ms, jinja2 11.5 ms, pydantic 9.2 ms, cookiecutter 2.0 ms, plus assorted smaller ones.

## Changes

- **`graphql/result.py`** only needs strawberry for one annotation on a classmethod, so it moves under `TYPE_CHECKING`. On master this single import pulled strawberry and graphql into every compile, via `inmanta.protocol.methods_v2`.
- **`moduletool.py`** imported cookiecutter at module scope but only uses it in the two template commands.
- **`agent/__init__.py`** eagerly re-exported `Agent` from `agent_new`, which dragged in `deploy.scheduler` -> `server.services.resourceservice` -> `server.agentmanager` -> `server.server`. `Agent` and `collect_report` are now resolved lazily through a module `__getattr__`, with a `TYPE_CHECKING` re-export so type checking is unaffected. **Deliberately not deleted**: `from inmanta.agent import Agent` is used by inmanta-lsm and lsm test suites, so removing it would be a cross-repo break.
- **`app.py`** imported tornado and `InmantaBootloader` at module scope; both are only used inside function bodies. All three `InmantaBootloader` sites are deferred, not just one, otherwise the `env.py` change below has no effect for the CLI.
- **`env.py`** imported `InmantaBootloader` at module scope for a single use inside a method. This also drops a compiler -> server layering edge.
- **`util/__init__.py`** imported asyncpg only for two annotations.

**The tornado and asyncpg moves have no effect yet.** Both are still pulled in regardless, tornado by `inmanta/util/__init__.py` (`from tornado import gen`) and asyncpg by `inmanta.data`. They are included because they are correct on their own terms and become effective once `inmanta/data/__init__.py` is thinned out, which is the next and much larger step.

## The test change

`tests/compiler/test_plugin_types.py::test_conversion` asserted an exact member order for the unions returned by `to_dsl_type`. That order comes from `typing.get_args`, and `typing._tp_cache` is a **bounded LRU (maxsize 128) whose key treats `Union[int, str]` and `Union[str, int]` as equal** - they compare equal and hash equal. So the ordering you get back depends on what else the process has subscripted and on what has been evicted since. Changing the import graph shifted the eviction and flipped it, making the test fail deterministically at `--randomly-seed=12345`.

This is pre-existing fragility rather than a regression here: **the test fails on master in exactly the same way**, same line and same repr, once `Optional[Union[str, int]]` is resident in the cache at test time (reproducible with a plugin that subscripts it in `pytest_runtest_setup`). The assertions now compare union members without regard to order. `Union.__eq__` is left alone, since member order is semantically meaningful for its `to_python` dispatch.

## Verification

`inmanta compile` succeeds; `--version`, `module`, `project`, `export` and `list-commands` behave and exit as before; `GraphQLResult.from_execution_result` still works against a real strawberry `ExecutionResult`; `from inmanta.agent import Agent` and attribute access both still resolve, and an unknown attribute still raises `AttributeError`.

black, isort and flake8 are clean. Full mypy produces a new-violation set byte-identical to master (the repo's baseline has some pre-existing drift, unchanged by this PR). `tests/compiler` passes at seeds 12345, 777 and 20260810; `tests/test_app.py`, `tests/util` and `tests/moduletool/test_module_tool_basics.py` pass (223 tests). The DB-backed suites could not be run locally, so `tests/deploy` and `tests/server` coverage of `scheduler_mocks`, `ExhaustedPoolWatcher` and `get_protected_inmanta_packages` rests on CI.

Branched off `master`, so it does not include #10672.

### DIFF
--- a/changelogs/unreleased/lazy-compiler-startup-imports.yml
+++ b/changelogs/unreleased/lazy-compiler-startup-imports.yml
@@ -1,0 +1,7 @@
+description: Cut compiler startup time by importing strawberry, cookiecutter, tornado, asyncpg, the server bootloader and the resource scheduler only where they are actually used, instead of on every invocation.
+change-type: patch
+destination-branches:
+  - master
+  - iso9
+sections:
+  minor-improvement: "{{description}}"
--- a/src/inmanta/agent/__init__.py
+++ b/src/inmanta/agent/__init__.py
@@ -16,8 +16,24 @@ limitations under the License.
 Contact: code@inmanta.com
 """
 
-from inmanta.agent.agent_new import Agent
+from typing import TYPE_CHECKING
 
 # flake8: noqa: F401
-# Backward compatibility
-from inmanta.agent.reporting import collect_report
+# Agent and collect_report are kept importable from this package for backward compatibility, but are resolved lazily by
+# __getattr__ below. Importing agent_new eagerly pulls the scheduler and the whole server into the import graph of everything
+# that touches inmanta.agent, including the compiler.
+if TYPE_CHECKING:
+    from inmanta.agent.agent_new import Agent
+    from inmanta.agent.reporting import collect_report
+
+
+def __getattr__(name: str) -> object:
+    if name == "Agent":
+        from inmanta.agent import agent_new
+
+        return agent_new.Agent
+    if name == "collect_report":
+        from inmanta.agent import reporting
+
+        return reporting.collect_report
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/src/inmanta/app.py
+++ b/src/inmanta/app.py
@@ -50,8 +50,6 @@ from configparser import ConfigParser
 from typing import Optional
 
 import click
-from tornado.httpclient import AsyncHTTPClient
-from tornado.ioloop import IOLoop
 
 import inmanta.compiler as compiler
 from inmanta import const, module, moduletool, protocol, tracing, util
@@ -66,7 +64,6 @@ from inmanta.export import cfg_env
 from inmanta.logging import InmantaLoggerConfig, _is_on_tty
 from inmanta.protocol import common
 from inmanta.server import config as opt
-from inmanta.server.bootloader import InmantaBootloader
 from inmanta.server.services.databaseservice import initialize_database_connection_pool
 from inmanta.server.services.metricservice import MetricsService
 from inmanta.signals import safe_shutdown, setup_signal_handlers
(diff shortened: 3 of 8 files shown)


bartv:

Correction to the verification note in the PR description: it says the DB-backed suites could not be run locally. That was wrong — `pytest_postgresql` spawns its own instance from the system `initdb`/`pg_ctl` binaries, so no running server is needed. I have now run them:

- `tests/deploy` + `tests/test_data.py` — **214 passed** (covers `scheduler_mocks`' `from inmanta.agent import Agent` through the new module `__getattr__`)
- `tests/server/test_databaseservice.py` + `tests/server/test_compilerservice.py` — **56 passed** (covers `ExhaustedPoolWatcher`'s asyncpg annotations and `get_protected_inmanta_packages`' deferred bootloader import)

So all three code paths I had flagged as CI-only are now verified locally as well.

bartv:

Closing in favour of a structural fix rather than a set of deferred imports.

Measurements that prompted the rethink: `inmanta.compiler` alone accounts for 931 ms of `inmanta.app`'s 980 ms, so the CLI wiring is at most a 5% slice. More importantly, model code imports the plugin API directly (`std` pulls `agent.handler`, `export`, `protocol` and `data`; `yang` pulls the `inmanta.agent` package), so a real compile ends with 1406 modules loaded and everything back. Against a project with only `std`, this branch moved a real compile from 998 ms to 991 ms — the gain was almost entirely on the `--help` path.

The replacement design puts the boundary at the module-facing plugin API and removes two dependency inversions (the plugin API reaching up for the protocol client, and the DTOs reaching up to the DAO), with a test to enforce the direction.

The individual changes here are still sound and are expected to return as part of that work: cookiecutter/tornado/bootloader as function-local imports, strawberry under `TYPE_CHECKING`, and the order-insensitive fix to `test_plugin_types.test_conversion` (that assertion depends on `typing._tp_cache`, a bounded LRU whose key treats `Union[int, str]` and `Union[str, int]` as equal, and it fails on master too once the other ordering is resident). The one part not being carried forward is the module `__getattr__` in `inmanta/agent/__init__.py`.

Branch `compiler-startup-lazy-imports` is kept for reference.

PR #10674, opened 2026-08-11
Move the database access layer to inmanta.data.dao

bartv:

> **Stacked on #10673.** The base branch is `compiler-startup-lazy-imports`, so the diff here shows only the incremental change. Merge #10673 first; this one will retarget to `master` automatically.

`inmanta/module.py` needs `inmanta.data.model.PipConfig`, but `model.py` lives inside the `inmanta.data` package, whose `__init__.py` was the 7381-line asyncpg access layer. Python always executes a package `__init__` before any of its submodules, so **every compile paid for sqlalchemy, asyncpg and the whole database layer just to reach a handful of pydantic DTOs.**

## Approach

The access layer moves to `inmanta/data/dao.py`. `__init__.py` becomes thin and resolves the names lazily through a module `__getattr__`, with an explicit `TYPE_CHECKING` re-export list so type checkers still see the real types rather than `object`.

**No call sites change.** `inmanta.data.<name>` keeps working both as an attribute and via `from inmanta.data import <name>` — which matters, because there are 467 such references in `src/`, 1168 in `tests/`, plus extension code and seven `@stable_api` classes. Renaming them would have been a cross-repo break.

Three companion fixes were needed so the compiler stops reaching the layer at all:

- `data/model.py` imported `inmanta.protocol` at module scope for two `json_encode` calls inside methods, and `asyncpg` for two annotations.
- `agent/handler.py` had two `data.LogLine` annotations, which are evaluated at import time.
- `app.py` imported the database and metrics services at module scope; both are only used inside `start_scheduler`.

## Effect

`import inmanta.app`, measured back to back against this PR's base (min of 5 `-X importtime` runs, min of 12 wall runs):

| | modules | import work | wall |
| ------- | ------- | ----------- | ---- |
| base (#10673) | 1385 | 716.6 ms | 878 ms |
| this PR | 1204 | **505.8 ms** | **618 ms** |

sqlalchemy and the whole `inmanta.data.dao` subtree are gone from every `inmanta` invocation.

## Behaviour change

`from inmanta.data import *` no longer works. Nothing in the codebase used it, and it was never a sensible way to use this module, but it is a real difference for anyone doing it downstream.

## Notes for review

- The `__getattr__` must not resolve submodules through dao: `from inmanta.data import dao` looks up `dao` on the package first, which comes straight back into `__getattr__` and recurses forever. It uses `importlib.import_module` plus an explicit submodule guard.
- Only dunders are excluded from lazy resolution, not all `_`-prefixed names — `tests/conftest.py` reaches `data._classes`.
- Six names in the re-export list are imported by dao rather than defined there, but callers reach them as `inmanta.data.<name>`: `AuthMethod`, `PipConfig`, `AgentModules`, `InmantaModule`, `ModuleFiles`, `ResourceIdStr`. They are re-exported from their real homes.
- The 168 mypy baseline entries for `src/inmanta/data/__init__.py` are re-keyed to `src/inmanta/data/dao.py`. The errors themselves are unchanged.

## Not included

Deferring the `inmanta.protocol` chain (plugins.py, handler.py, app.py, logging.py, deploy/state.py, util) is worth roughly another 100 ms and removes protocol, jwt, cryptography and tornado from the compile path — locally that took it to 1104 modules. I have left it out because it surfaces a mypy error that master does not report:

```
handler.py:572: Incompatible return value type
  (got "ClientCall[BaseModel | UUID | bool | int | float | <6 more items> | None]",
   expected "Awaitable[Result[bool]]")  [return-value]
```

It looks like a true positive that was previously masked — `ClientCall(Awaitable[Result[R]])` and `Client.__getattr__` returns the wide union, so the declared narrowing to `Result[bool]` does not hold under invariance. That deserves a real fix from someone who owns that code rather than being buried in this refactor.

## Verification

**1254 tests pass locally**: 916 (`tests/test_data.py` + `tests/deploy` + `tests/compiler`) and 338 (`tests/db` migrations + server `databaseservice`/`compilerservice`/`user` + `tests/test_app.py` + `tests/util`).

black, isort and flake8 clean. Full mypy reports no new violations against the base. `inmanta compile` succeeds; `--version` and `list-commands` unchanged. The access layer is reachable via attribute access, from-import, `issubclass` and pickle, and identity is preserved (`data.Environment is dao.Environment`).


bartv:

Closing: the lazy `inmanta.data` package is the wrong mechanism, and a structural move replaces it.

Making a package resolve its contents through `__getattr__` means it no longer tells the truth about itself, and three separate things tripped over that during development:

- `tests/conftest.py` reaches `data._classes`, so the guard could not exclude single-underscore names
- `inmanta.graphql.schema` walks the module with `inspect.getmembers`, which a module `__getattr__` is invisible to — the docstring cache came out empty and every GraphQL field silently lost its description (this is what failed CI here)
- `inmanta_plugins/std/resources.py:86` carries `has_autostarted_agent_map: bool = hasattr(data, "AUTOSTART_AGENT_MAP")` at class-body level, so a compatibility probe imports 7381 lines of DAO code plus sqlalchemy and then returns `False`

Each needed its own patch, and each was silently wrong until something happened to trip over it.

The replacement reaches the same place by moving a module instead: the DTOs move out of `inmanta/data/model.py` into their own package, `inmanta/data/model.py` stays as a re-export shim for the `@stable_api` contract, and `inmanta/data/__init__.py` goes back to being the DAO layer verbatim with no lazy machinery. That also requires removing two genuine layering inversions: `DiscoveredResourceInput.to_dao()` (a DTO constructing a DAO object) and the DTOs calling `protocol.common.json_encode`, plus moving the environment-setting name constants down so handler code in modules does not need the DAO for a string.

Measured proxy for the effect: `inmanta.data.model` goes from 852 ms to 425 ms.

Branch `compiler-startup-thin-data` is kept for reference; the 168 mypy baseline re-keyings and the full public-name inventory of the DAO layer are on it and worth reusing.

PR #10685, opened 2026-08-12
Tracking: what remains of the import cleanup train

bartv:

**This is the tracking PR for what remains of the import cleanup train. Its branch predates the split, so the diff GitHub shows below is not meaningful yet** — `compiler-startup-dto-split` was cut before #10702 and #10703 existed, and it contains its own single-module version of the DTO move. Each item below gets rebuilt as a car on top of #10703 rather than replayed.

Base is `dto-package` (#10703), so once the items are rebuilt this diff becomes exactly the outstanding work.

## Landed, or up for review

| | what | state |
| - | ---- | ----- |
| ✅ | Break the `inmanta.data` ↔ `inmanta.data.model` cycle | #10702, open |
| ✅ | DTOs into the `inmanta.dto` package of thematic modules | #10703, open, stacked on #10702 |
| ✅ | Keep the database layer off the lsm compile path | inmanta/inmanta-lsm#2481 and inmanta/lsm#1004, open |
| ✅ | Drop the dead `std::AgentConfig` handler | inmanta/std#677, merged, awaiting a release |

## Still open

**1. Migrate call sites off the `inmanta.data.model` re-export.** 33 files under `src/inmanta` import it, 96 name imports in total, spread across all 17 themes: `resource` 18, `code` 14, `pip` 9, `scheduler` 8, `diff` 8, `status` 8, `desiredstate` 4, `auth` 3, `parameter` 3, and the rest in ones and twos. Seven of them import `BaseModel`, which should come from `inmanta.types` instead. This is mechanical and scriptable per theme. Nothing measurable changes until the last importer on a given path is gone, at which point that path stops loading the database layer.

**2. Break the second cycle.** `inmanta.resources.Resource.to_action()` returns a `data.ResourceAction`, the same DTO-constructs-DAO violation that #10702 fixes for `to_dao`. This cycle is why `ResourceLog` currently has to sit next to `LogLine` in `dto/log.py`: inside the strongly connected component it creates, the pydantic mypy plugin does not reliably see a base model from another module.

**3. Unify the handler `LogLine` with the pydantic `dto.log.LogLine`.** `inmanta/data/__init__.py` still carries a `DataDocument` based `LogLine`, which predates pydantic. Folding the two removes `DataDocument` entirely. `LogLine._data` stays readable so that handler tests reaching into it keep working.

**4. Empty `inmanta/agent/__init__.py`.** It re-exports `Agent` from `agent_new`, so importing `inmanta.agent.handler` — which every module's Python code does — pulls in the scheduler and the server.

**5. Defer the server side imports** that sit at module scope for the sake of a single function body: the bootloader in `inmanta/env.py`, and the database and metrics services in `inmanta/app.py`.

**6. Put strawberry under `TYPE_CHECKING`** in `inmanta/graphql/result.py`. One annotation currently puts strawberry and the whole graphql package on the import path of every compile. Independent of the rest and worth ~286 modules.

## Ordering constraint

Item 3 removes `DataDocument`, which **breaks the lsm extension** in two places. Both fixes are version agnostic and want to merge ahead of it:

- `inmanta_lsm/data/__init__.py:35` imports `DataDocument` from `inmanta.data`; it is only used in one isinstance tuple, where the branch is dead because `LogLine` was core's only `DataDocument` subclass and the extension never references it.
- `from inmanta.agent import Agent` in `inmanta-lsm/tests/test_failure_diagnoser.py` and two `modules/lsm/tests/integration/` test files, which item 4 breaks.

Found by running both lsm suites against this branch, not by inspection.

## Where the numbers come from

The measurements quoted around this effort were taken on this combined branch, which is why it stays open as the reference until the cars replace it. Real `inmanta compile` of a model importing lsm, production-like venv, minimum of five runs: 1453 loaded modules and 1069 ms today, against 954 and 672 ms with the whole train applied, with `inmanta.data` and `sqlalchemy` absent.


bartv:

Pushed `c272711da`, following the review point that `DataDocument` predates pydantic and the DTO approach.

`DataDocument` and `inmanta/data_document.py` are **removed**. Its only subclass, the `LogLine` handlers produce, duplicated `inmanta.dto.LogLine` field for field — `level`, `msg`, `args`, `kwargs`, `timestamp` — so the two really were the same concept stored two ways. The behaviour is folded into the pydantic `inmanta.dto.LogLine`, and `inmanta.data` re-exports it so `data.LogLine` keeps working for the server and deploy code.

**The serialized shape is unchanged.** That matters, because these land in the `resourceaction.messages` column and in API responses. Verified equal to the old class byte for byte through `json_encode`, through `to_dict`, and through a pickle round trip carrying `datetime` and `uuid` values in `kwargs`.

Four things a reviewer should look at specifically:

- `__getstate__` and `__setstate__` still route through json rather than pickle, so a log line crossing the executor IPC boundary is coerced exactly like one crossing the json based RPC. They now exchange a dict instead of a str, because that is the contract `pydantic.BaseModel` declares — returning a str was a Liskov violation mypy correctly flagged.
- `timestamp` gains a default of now. The old class tolerated its absence and filled it in while serializing, and `tests/forking_agent/test_pickle.py` relies on that. This loosens nothing on the API: `LogLine` and `ResourceLog` only ever appear in responses, never in requests.
- `handler.py` reached into `log._data["msg"]` directly. That was behind 57 test failures on the first attempt and now uses `log.msg`.
- `log_level` is kept as a property alias for `level`, since `data.LogLine.log_level` was reachable.

The four baselined typing errors for these classes are **deleted rather than re-keyed** — they were artifacts of the untyped `_data` dict and no longer occur.

Verification on this commit: **1087 tests pass** (315 across `test_data`/`deploy`/`test_server`/`forking_agent`/`test_data_model`/`test_graphql`, 772 across `compiler`/`db`/`databaseservice`). Lint clean, and full mypy is back to exactly master's baseline drift with no new violations. A compile now loads 1528 modules with `inmanta.data` and `sqlalchemy` absent.

bartv:

Pushed `886e56a18`. **`_data` is kept as a read-only alias, so no downstream repo has to be released in lockstep with this PR.**

I searched `~/workspace` for code reaching into the old `DataDocument` internals. Three live first-party users, all reads:

| where | lines | impact without the alias |
| ----- | ----- | ------------------------ |
| `pytest-inmanta/pytest_inmanta/plugin.py` | 1232, 1237, 1241, 1242, 1682, 1685, 1690, 1691 | every module's tests would report `AttributeError` instead of the real assertion whenever a deploy test failed |
| `projects/enet/ciena-saos/tests/test_handler.py` | 433, 452 | asserts on `_data["kwargs"]["response"]` in **passing** tests, so those break outright |
| `modules/checkpoint/tests/common.py` | 70, 75, 79, 80 | test helper printing `_data["msg"]` and `_data["kwargs"]` |

Upstream `modules/openstack` is clean; the other openstack hits are vendored old copies under `projects/inmanta-demo/libs` and `projects/training/*/libs`, which is evidently where the idiom originated before it was copied into pytest-inmanta and checkpoint.

`LogLine._data` is now a property returning `to_dict()`. All four idioms found in the wild work unchanged:

```
log._data["msg"]
log._data["kwargs"].items()
"traceback" in log._data["kwargs"]
result.logs[-1]._data["kwargs"]["response"]
```

One documented limitation: reads behave as before, and writes into nested values such as `_data["kwargs"][...]` still propagate because `to_dict` hands out the same kwargs object, but assigning a **top level** key no longer takes effect. Nothing outside the class ever did that. The only `_data` writes anywhere in the workspace were the old class's own timestamp handling, which is gone.

So downstream fixes become optional cleanup rather than a release-ordering constraint. The version-agnostic form, if anyone wants to drop the private access anyway, is `log.msg` and `log.to_dict()["kwargs"]` — note plain `log.kwargs` is *not* backwards compatible, since the old class had no `kwargs` property, which is exactly why everyone reached into `_data` for it.

Re-verified on this commit: lint clean, mypy at exactly master's baseline drift, **268 tests pass** across `test_data`, `deploy`, `forking_agent`, `test_data_model` and `test_graphql`.

bartv:

Pushed `7742ef164`, adding the strawberry fix from the closed #10673. It turned out to be the single largest remaining item, and much bigger than the ~53 ms I estimated.

`inmanta/graphql/result.py` imported `strawberry.types.execution` at module scope for one annotation on one classmethod. `inmanta.protocol.methods_v2` imports that module, and the compiler imports the protocol layer, so that one line put strawberry **and the whole graphql package** on every compile: **286 modules**, more than strawberry alone, because it drags in graphql's schema machinery.

This matters more than when it was first spotted, precisely because the protocol client is staying on the compile path by design. Nothing else was ever going to remove it.

### Compile time, measured back to back against master in one session, with std from master

| | master | this PR | change |
| ------- | ------ | ------- | ------ |
| dev venv, logfire installed | 1127 ms / 1739 modules | **733 ms / 1243 modules** | **-35%** |
| production-like, logfire absent | 984 ms / 1430 modules | **581 ms / 933 modules** | **-41%** |

The production row is the one that matters for shipped installs, since logfire is only in the optional `tracing` extra. A stock compile goes from just under a second to under 600 ms.

Both figures still need std 8.7.5 released. With std 8.7.1 the gain is far smaller, because its `from inmanta import data` pulls the database back in whatever core does.

### Verification

`GraphQLResult.from_execution_result` checked against a real strawberry `ExecutionResult`, both the success and the error path. **1046 tests pass**: `tests/test_graphql.py` (29), plus `protocol`, `compiler`, `test_data.py` and `deploy` (1017). The one failure in that run, `test_protocol.py::test_2151_method_header_parameter_in_body`, is `[Errno 98] Address already in use` from an orchestrator container holding 127.0.0.1:8888 on this machine; it fails identically on pristine master.

Lint clean, mypy at exactly master's baseline drift.

src/inmanta/types.py
```diff
@@ -100,6 +100,11 @@ def issubclass(sub: type, super: Union[type, tuple[type, ...]]) -> bool:
 
 
 PrimitiveTypes = Optional[uuid.UUID | bool | int | float | datetime.datetime | str]
+
+PRIMITIVE_SQL_TYPES = Union[str, int, bool, datetime.datetime, uuid.UUID]
+"""
+    The primitive types that can be stored in, and ordered by, a database column.
+"""
```
bartv, line 107:
```suggestion
""" The primitive types that can be stored in, and ordered by, a database column. """
```

bartv:
Closing: this branch predates the split into #10702 and #10703, so its diff is not meaningful. The remaining work is being rebuilt as cars on top of #10703, starting with the call site migration. The branch stays around locally as the reference for what still needs porting.

PR #10687, opened 2026-08-12
Make the maximum request body size accepted by the server configurable

jptrindade:
# Description

The server relied on Tornado's implicit default for `max_body_size`: `HTTPServer` was constructed without it, so `HTTP1Connection` fell back to `stream.max_buffer_size`, which is 100MiB. This surfaced when the addition of the `send_event`/`receive_event` flags on inmanta-performance pushed an export of 200k resources over that limit.

This exposes the limit as the `server.max-request-body-size` config option, keeping the default at 100MiB so behaviour is unchanged unless the option is set.

Discussed [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1786519779489309).

# Self Check:

- [x] Changelog entry
- [x] If this PR fixes a race condition in the test suite, also block the sha of the improvement on normal test suite runs

### DIFF
--- a/src/inmanta/server/config.py
+++ b/src/inmanta/server/config.py
@@ -178,6 +178,15 @@ server_bind_port = Option(
     is_int,
 )
 
+server_max_request_body_size = Option(
+    "server",
+    "max-request-body-size",
+    100 * 1024 * 1024,
+    "The maximum size in bytes of the body of a request accepted by the server. Requests with a larger body are "
+    "rejected. When a request is compressed, this limit applies to the compressed body.",
+    is_lower_bounded_int(1),
+)
+
 server_tz_aware_timestamps = Option(
     "server",
     "tz_aware_timestamps",
--- a/src/inmanta/server/config.py
+++ b/src/inmanta/server/config.py
@@ -183,7 +183,7 @@ server_max_request_body_size = Option(
     "max-request-body-size",
     100 * 1024 * 1024,
     "The maximum size in bytes of the body of a request accepted by the server. Requests with a larger body are "
-    "rejected. When a request is compressed, this limit applies to the compressed body.",
+    "rejected. This limit is applied to both compressed and decompressed bodies.",
     is_lower_bounded_int(1),
 )
 
--- a/tests/protocol/test_max_request_body_size.py
+++ b/tests/protocol/test_max_request_body_size.py
@@ -60,7 +60,7 @@ async def test_max_request_body_size_decompressed(server) -> None:
     """
     file_hash, _, file_content = make_random_file(size=MAX_REQUEST_BODY_SIZE)
     is_zipped, body = protocol.gzipped_json({"content": file_content})
-    assert zipped
+    assert is_zipped
     # The request only makes it past the limit on the compressed body because it is this much smaller than the
     # decompressed one it carries.
     assert len(body) < MAX_REQUEST_BODY_SIZE < len(json.dumps({"content": file_content}))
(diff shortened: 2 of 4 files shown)


src/inmanta/server/config.py
```diff
@@ -178,6 +178,15 @@ def default_db_pool_max_size() -> int:
     is_int,
 )
 
+server_max_request_body_size = Option(
+    "server",
+    "max-request-body-size",
+    100 * 1024 * 1024,
+    "The maximum size in bytes of the body of a request accepted by the server. Requests with a larger body are "
+    "rejected. When a request is compressed, this limit applies to the compressed body.",
```
arnaudsjs: It also applies to the decompressed body, right?

tests/protocol/test_max_request_body_size.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    file_hash, _, file_content = make_random_file(size=MAX_REQUEST_BODY_SIZE if above_limit else 1)
+    body = json.dumps({"content": file_content}).encode()
+    assert (len(body) > MAX_REQUEST_BODY_SIZE) == above_limit
+
+    port = opt.server_bind_port.get()
+    request = HTTPRequest(url=f"http://localhost:{port}/api/v1/file/{file_hash}", method="PUT", body=body)
+    response = await AsyncHTTPClient().fetch(request, raise_error=False)
+
+    assert response.code == (400 if above_limit else 200)
```
arnaudsjs, line 53: Do I understand correctly that the server will never return a 599 in this case, because the size of the request is close to the limit?

arnaudsjs (changes requested):

src/inmanta/server/config.py
jptrindade: yes

tests/protocol/test_max_request_body_size.py
jptrindade: In my testing, that was the case.

jptrindade:
@arnaudsjs I believe we also need this in iso8. But if I recall correctly we can't/shouldn't keep adding minors to iso8. Should I change this to a patch?

arnaudsjs:
> @arnaudsjs I believe we also need this in iso8. But if I recall correctly we can't/shouldn't keep adding minors to iso8. Should I change this to a patch?

Indeed. Mergetool doesn't allow minors in iso8. So we will have to change this to patch.

tests/protocol/test_max_request_body_size.py
jptrindade: 599 is raised only for slow uploads. So in theory it could happen for example if inmanta-performance takes too long to upload the body for the request, so it makes sense to test it there, but not here.

tests/protocol/test_max_request_body_size.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+async def test_max_request_body_size_decompressed(server) -> None:
+    """
+    Verify that the limit is applied to the decompressed body as well: a request that stays below the limit while
+    compressed is still rejected when it inflates beyond it.
+    """
+    file_hash, _, file_content = make_random_file(size=MAX_REQUEST_BODY_SIZE)
+    zipped, body = protocol.gzipped_json({"content": file_content})
```
arnaudsjs: I got confused by this. `zipped` sounds like it contains the zipped content, but in fact it's just a boolean.
```suggestion
    is_zipped, body = protocol.gzipped_json({"content": file_content})
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #10699, opened 2026-08-13
Fix compatibility with Python 3.14

jptrindade:
# Description

*Claude here, opening this on behalf of @jptrindade.*

`import inmanta` fails on Python 3.14, which blocks
[irt#update-to-python-3.14](https://jenkins.inmanta.com/job/integration-tests/job/irt-test/job/update-to-python-3.14/1/).
As of 3.14 the annotations of a class are evaluated on first read and cached outside of the class dict, in a scope that
does not see the class namespace (PEP 649). Three places depended on the old behaviour:

* `BaseDocument.load_fields` read `cls.__annotations__` from within a loop over `cls.__dict__`, so caching them resized
  the dict being iterated: `RuntimeError: dictionary changed size during iteration`.
* `ProjectMetadata.relation_precedence_policy` constrained itself with a pattern defined on the class body, which the
  lazily evaluated annotation can no longer resolve, so pydantic got a `ForwardRef`. The pattern moved to module level.
* `build_strawberry_output_type` and `build_composed_filter_input` read the annotations of their mixins out of
  `__dict__`, where they no longer are. That returned nothing, silently: the fields extensions contribute were dropped
  from the GraphQL types and the duplicate field guard on filters stopped guarding.

All three now read them through `inspect.get_annotations()`, which behaves the same on 3.13 and on 3.14.

Verified on both interpreters: every `inmanta.*` module imports, `tests/test_data.py` passes (68 tests), and the
`BaseDocument` field metadata and generated GraphQL SDL are identical. `make mypy` is clean on 3.13. No tests are added
because the existing suite already covers this on 3.14: with only the `schema.py` hunk reverted,
`test_custom_extension_contributions` and `test_resolved_model_version_available_to_contributions` fail with
`MissingFieldAnnotationError`, and the other two issues break `import inmanta` outright.

# Self Check:

- [x] Changelog entry
- [x] ~~End user documentation is included or an issue is created for end-user documentation~~

### DIFF
--- a/changelogs/unreleased/python-3.14-compatibility.yml
+++ b/changelogs/unreleased/python-3.14-compatibility.yml
@@ -1,4 +1,4 @@
 description: Fix compatibility with Python 3.14.
-change-type: patch
+change-type: minor
 destination-branches:
   - master
--- a/changelogs/unreleased/python-3.14-compatibility.yml
+++ b/changelogs/unreleased/python-3.14-compatibility.yml
@@ -1,0 +1,10 @@
+description: >-
+  Fix compatibility with Python 3.14, where the annotations of a class are evaluated on first access and cached outside
+  of the class dict (PEP 649). Importing inmanta failed with a `RuntimeError: dictionary changed size during iteration`
+  raised by the metaclass of `BaseDocument`, and the GraphQL output types and filters silently lost the fields their
+  mixins declare.
+change-type: patch
+destination-branches:
+  - master
+sections:
+  bugfix: "{{description}}"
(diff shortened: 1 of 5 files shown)


jptrindade:
@arnaudsjs do we want to update `python_requires >= 3.13` ? I believe so, I'm just wondering, what PR has to go in first (or if it matters) this one or irt's

arnaudsjs:
> do we want to update `python_requires >= 3.13` ?

Yes. I think so.

> what PR has to go in first (or if it matters) this one or irt's

The one of IRT has to go in first.

arnaudsjs (approved):

PR #10706, opened 2026-08-14
Move extension discovery from the server bootloader into inmanta.util

bartv:

**Stacked on #10705 → #10703 → #10702. Review those first.** Base branch is `migrate-dto-imports`, so the diff here is only this step.

`inmanta/env.py` manages virtual environments, which the compiler uses. It has to protect extension packages from being uninstalled, and asked the server for the list of them:

```python
from inmanta.server.bootloader import InmantaBootloader
...
for ext_name in InmantaBootloader.get_available_extensions().keys()
```

That one call put the server slice machinery and the database access layer on the compiler's import path.

`get_available_extensions` does not need the server. It walks the `inmanta_ext` namespace package with `pkgutil` and caches the result, so it moves to `inmanta.util` along with its `iter_namespace` helper and the cache. `InmantaBootloader.get_available_extensions` stays and proxies to it, so its callers and the `@stable_api` surface of the class are unaffected.

## Effect

| import | before | after |
| ------ | -----: | ----: |
| `inmanta.env` | 1580 modules, server + `inmanta.data` + `sqlalchemy` all loaded | **920 modules, none of them loaded** |
| `inmanta.module` | 1692 | 1692 |
| `inmanta.app` | 1757 | 1757 |

`inmanta.module` and `inmanta.app` do not move, because they still reach `inmanta.data` by a route this PR does not touch:

```
module.py:60 → plugins.py:39 → protocol/__init__.py:56 → methods.py:26 → from inmanta import const, data, resources
```

`plugins.py` importing the protocol layer is legitimate — that is `Context.get_client()`. What is not is `protocol/methods.py` importing the database at module scope for one server side getter, and `protocol/auth/policy_engine.py` doing the same for one line. Those two are the next step; deferring them takes `import inmanta.logging` from 1574 to 1395 modules with the database absent, which matters because nearly everything imports `inmanta.logging`.

## Also fixed

`iter_namespace` was annotated `-> Generator[ModuleInfo, None, None]` but returns `pkgutil.iter_modules(...)`, an `Iterator`. The mismatch was suppressed by a baseline entry rather than fixed. Since the function moves, the annotation is corrected to `Iterator[ModuleInfo]` and the baseline entry is deleted.

## Verification

- CI mypy and the fast test suite. Locally, `mypy-baseline filter` with a cold cache reports the same set of pre-existing violations master produces, one baseline entry lighter.
- black, isort and flake8 clean.
- `InmantaBootloader.get_available_extensions()` and `inmanta.util.get_available_extensions()` return equal results, and `import inmanta.env` no longer loads `inmanta.server.*` at all.

### DIFF
--- a/changelogs/unreleased/extension-discovery-in-util.yml
+++ b/changelogs/unreleased/extension-discovery-in-util.yml
@@ -1,0 +1,5 @@
+description: Moved extension discovery from InmantaBootloader to inmanta.util, so managing a virtual environment no longer imports the server.
+change-type: patch
+destination-branches:
+  - master
+  - iso9
--- a/mypy-baseline.txt
+++ b/mypy-baseline.txt
@@ -979,7 +979,6 @@ src/inmanta/resources.py:0: error: Returning Any from function declared to retur
 src/inmanta/server/agentmanager.py:0: error: Call to untyped function "get_status" in typed context  [no-untyped-call]
 src/inmanta/server/agentmanager.py:0: error: Item "None" of "Sequence[Agent] | None" has no attribute "__iter__" (not iterable)  [union-attr]
 src/inmanta/server/agentmanager.py:0: error: Value of type "dict[str, Any] | None" is not indexable  [index]
-src/inmanta/server/bootloader.py:0: error: Incompatible return value type (got "Iterator[ModuleInfo]", expected "Generator[ModuleInfo, None, None]")  [return-value]
 src/inmanta/server/config.py:0: error: Cannot infer value of type parameter "T" of "Option"  [misc]
 src/inmanta/server/config.py:0: error: Cannot infer value of type parameter "T" of "Option"  [misc]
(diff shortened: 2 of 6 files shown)


bartv:

Pushed a fix: removing `InmantaBootloader.AVAILABLE_EXTENSIONS` broke `tests/conftest.py`, which saved and restored that attribute around the `deactive_venv` fixture. Every test using that fixture errored, 201 of them in the subset I ran.

`inmanta.util` now exposes `reset_available_extensions_cache()` and the fixture calls it instead. Restoring the previous value is unnecessary: the fixture restores the venv snapshot first, so rediscovery yields the same set.

I found this while working on the PR stacked on top, having only run mypy on this branch and not the test suite. 128 tests now pass across `test_extension_loading.py`, `test_loader.py`, `test_env.py` and `test_data.py`; the two failures there (`test_git_based_install`, `test_register_setting`) reproduce on pristine master.

Issue #10590, opened 2026-07-15
migrate iso10 to Python 3.14


sanderr:
- [x] migrate final pipelines that don't use `uv` yet
- [x] Follow Python upgrade guide from internal docs. Take into account that a lot of the processes have changed in the past few monthts, so be critical of the steps.

PR #10717, opened 2026-08-18
Ensure compatibility with python3.14

jptrindade:
# Description

Drop support for python 3.13. 
3.14 is now the supported version for iso10

related to: https://github.com/inmanta/inmanta-core/issues/10590

# Self Check:

- [x] Changelog entry

### DIFF
--- a/.ci-performance.yml
+++ b/.ci-performance.yml
@@ -1,1 +1,0 @@
-pipeline_template: ci/inmanta-core/jenkinsfile-performance
--- a/src/inmanta/moduletool.py
+++ b/src/inmanta/moduletool.py
@@ -748,6 +748,7 @@ When a development release is done using the \--dev option, this command:
             raise ModuleVersionException(f"Expected a v1 module, but found v{mod.GENERATION.value} module")
         ModuleConverter(mod).convert_in_place()
 
+    # The return annotation is quoted because the list method below shadows the builtin in this class' annotation scope.
     def build(
         self,
         path: Optional[str] = None,
@@ -756,7 +757,7 @@ When a development release is done using the \--dev option, this command:
         byte_code: bool = False,
         wheel: bool = False,
         sdist: bool = False,
-    ) -> list[str]:
+    ) -> "list[str]":
         """
         Build a v2 module and return the path to the build artifact(s).
 
@@ -831,7 +832,8 @@ When a development release is done using the \--dev option, this command:
             project = self.get_project(load=True)
             return project.get_module(module, allow_v1=True)
 
-    def get_modules(self, module: Optional[str] = None) -> list[Module]:
+    # The return annotation is quoted because the list method below shadows the builtin in this class' annotation scope.
+    def get_modules(self, module: Optional[str] = None) -> "list[Module]":
         if module is not None:
             return [self.get_module(module)]
         else:
--- a/src/inmanta/moduletool.py
+++ b/src/inmanta/moduletool.py
@@ -748,7 +748,7 @@ When a development release is done using the \--dev option, this command:
             raise ModuleVersionException(f"Expected a v1 module, but found v{mod.GENERATION.value} module")
         ModuleConverter(mod).convert_in_place()
 
-    # The return annotation is quoted because the list method below shadows the builtin in this class' annotation scope.
+    # The return annotation is quoted because the list method below shadows builtins.list in this class' annotation scope.
     def build(
         self,
         path: Optional[str] = None,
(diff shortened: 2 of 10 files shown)


arnaudsjs (changes requested):
You are missing a changelog entry

jptrindade:
> You are missing a changelog entry

forgot to commit it my bad

arnaudsjs (approved):

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -298,7 +297,9 @@ def safe_decode(token: lex.LexToken, warning_message: str, start: int = 1, end:
     try:
         # This first block will try to decode the value and turn any deprecation warning into an actual Exception.
         with warnings.catch_warnings():
-            warnings.filterwarnings("error", message="invalid escape sequence", category=DeprecationWarning)
+            warnings.filterwarnings(
+                "error", message=re.escape(r'"\." is an invalid escape sequence.'), category=DeprecationWarning
```
sanderr: Why narrow the message?

src/inmanta/parser/plyInmantaLex.py
```diff
@@ -282,10 +283,8 @@ def safe_decode(token: lex.LexToken, warning_message: str, start: int = 1, end:
     Check for the presence of an invalid escape sequence (e.g. "\.") in the value attribute of a given token.
     This function assumes to be called from within a t_STRING or a t_MLS rule.
 
-    - Python < 3.12 raises a DeprecationWarning when encountering an invalid escape sequence
-    - Python 3.12 will raise a SyntaxWarning
-    - Future versions will eventually raise a SyntaxError
-    (see https://docs.python.org/3.12/whatsnew/3.12.html#other-language-changes )
+    Decoding an invalid escape sequence raises a DeprecationWarning. Future Python versions will eventually raise a
```
sanderr, line 285: This contradicts what is said above. Which is correct?

src/inmanta/moduletool.py
```diff
@@ -748,6 +748,7 @@ def v1tov2(self, module: str) -> None:
             raise ModuleVersionException(f"Expected a v1 module, but found v{mod.GENERATION.value} module")
         ModuleConverter(mod).convert_in_place()
 
+    # The return annotation is quoted because the list method below shadows the builtin in this class' annotation scope.
```
sanderr: `builtins.list` would be more explicit imo and not require a comment per se. But I'm fine either way.

.ci-performance.yml
sanderr, line 1: What's this?

sanderr (changes requested):

src/inmanta/parser/plyInmantaLex.py
arnaudsjs: This message changed so we are not really narrowing this. It's an actual fix. But I can fix this by prefixing the regex with `.*` so that it does a substring search instead. I'll do that.

.ci-performance.yml
arnaudsjs: I think Joao removed it because we have a new performance test and we no longer use this file.

src/inmanta/parser/plyInmantaLex.py
arnaudsjs: The new one is correct. The SyntaxWarning is a compile time warning while the Deprecationwarning is the runtime warning.

sanderr: My comment was also just wrong. Sorry about that.

src/inmanta/protocol/auth/auth.py
```diff
@@ -381,7 +381,7 @@ def _load_config_and_validate(cls) -> None:
         else:
             cls._config_successfully_loaded = True
 
-    # The return annotation is quoted because this method shadows the builtin in the class' annotation scope.
+    # The return annotation is quoted because this method shadows the builtins.list in the class' annotation scope.
     @classmethod
     def list(cls) -> "list[str]":
         """
```
sanderr, line 387: To be clear: I meant in the annotation. Still fine with this though. (Feel free to dismiss my review)

sanderr (changes requested):

src/inmanta/protocol/auth/auth.py
arnaudsjs: It seems that mypy doesn't like that.

sanderr: Interesting. I'll have to look into that some day, because I'm curious as to why.

sanderr (approved):

PR #10719, opened 2026-08-18
Ensure stability of resource hash

### LINKED ISSUE #10622: Unstable desired state: Resource.fields order is randomized, reordering serialized references/mutators between exports
*Filed by Claude on behalf of @edvgui.*

## Problem

Two consecutive exports of an **unchanged model** produce a **different desired state**: the `references` and `mutators` attributes of a resource contain the same entries, but in a different order. This shows up in the console's desired state compare view as a change on every resource that carries more than one reference, and it makes every export produce a new version even when nothing changed.

Observed on `inmanta-core` 18.1.2; the offending code is unchanged on master.

## Root cause

`ResourceMeta.__new__` in `src/inmanta/resources.py`:

```python
def __new__(cls, class_name, bases, dct):
    fields = cls._get_parent_fields(bases)
    if "fields" in dct:
        ...
        fields.extend(dct["fields"])

    dct["fields"] = tuple(set(fields))   # <-- here
```

`fields` is deduplicated through a `set` of strings. Python randomizes string hashing per process (`PYTHONHASHSEED`), so **`Resource.fields` has a different order in every compiler process**.

`Resource.create_from_model` then iterates that tuple with a single shared `ReferenceCollector`:

```python
fields: dict[str, object] = {
    field: resource_cls.map_field(exporter, entity_name, field, model_object, reference_collector)
    for field in resource_cls.fields
}

fields[const.RESOURCE_ATTRIBUTE_REFERENCES] = list(reference_collector.references.values())
fields[const.RESOURCE_ATTRIBUTE_MUTATORS] = reference_collector.mutators
```

The collector appends mutators to a list and inserts references into a dict **in field-visit order**, so both emitted lists inherit the randomized field order.

## Reproduction

```python
# f.py
from inmanta.resources import Resource

class A(Resource):
    fields = ("account_id", "tunnel_id", "uri", "config_items", "api_token")

print(A.fields)
```

```console
$ for s in 1 2 3; do PYTHONHASHSEED=$s python f.py; done
('references', 'api_token', 'account_id', 'config_items', 'mutators', 'uri', 'tunnel_id')
('account_id', 'references', 'mutators', 'config_items', 'api_token', 'uri', 'tunnel_id')
('references', 'config_items', 'tunnel_id', 'account_id', 'mutators', 'uri', 'api_token')
```

In the real environment where this was found, 48 resources of the same type differed between two versions, all of them only on `mutators`/`references`, with identical ids in swapped order: the resource has exactly two reference-carrying fields (a `std::FactReference` on one field and a `files::JinjaReference` on another), hence exactly two possible orderings.

Diff excerpt (same resource, two consecutive versions, `mutators`):

```
from_value: [ Replace(destination="tunnel_id", value=ref 0ed6baa6-...),
              Replace(destination="api_token", value=ref 8cf4e2dc-...) ]
to_value:   [ Replace(destination="api_token", value=ref 8cf4e2dc-...),
              Replace(destination="tunnel_id", value=ref 0ed6baa6-...) ]
```

## Suggested fix

Make the dedup order-preserving:

```python
dct["fields"] = tuple(dict.fromkeys(fields))
```

`dict.fromkeys` dedups exactly like `set` but keeps first-occurrence order, which is stable across processes.

Optionally, additionally sort the emitted `references` / `mutators` lists deterministically (e.g. by reference id) before they are attached to the resource, so their order is also stable if the set of fields itself ever changes.

arnaudsjs:
# Description

Fix bug where a resource can be considered as updated across model versions although its attribute set has not changed.

closes #10622

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/inmanta/references.py
+++ b/src/inmanta/references.py
@@ -262,11 +262,12 @@ ArgumentTypes = typing.Annotated[
 """ A list of all specific types of arguments. Pydantic uses this to instantiate the correct argument class
 """
 
+
 def get_id_reference_or_mutator(
     type: ReferenceType,
     args: list[ArgumentTypes],
 ) -> uuid.UUID:
-    data = json.dumps({"type": self.type, "args": args}, default=util.api_boundary_json_encoder, sort_keys=True)
+    data = json.dumps({"type": type, "args": args}, default=util.api_boundary_json_encoder, sort_keys=True)
     hasher = hashlib.md5()
     hasher.update(data.encode())
     return uuid.uuid3(uuid.NAMESPACE_OID, hasher.digest())
--- a/src/inmanta/references.py
+++ b/src/inmanta/references.py
@@ -291,8 +291,12 @@ class ReferenceModel(SerializedReferenceLike):
 class MutatorModel(SerializedReferenceLike):
     """A mutator"""
 
+    _id: uuid.UUID | None = pydantic.PrivateAttr(default=None)
+
     def get_id(self) -> uuid.UUID:
-        return compute_reference_or_mutator_id(reference_type=self.type, args=self.args)
+        if self._id is None:
+            self._id = compute_reference_or_mutator_id(reference_type=self.type, args=self.args)
+        return self._id
 
 
 C = typing.TypeVar("C", bound="ReferenceLike")
@@ -423,7 +427,9 @@ class Mutator(ReferenceLike):
         """Emit the correct pydantic objects to serialize the reference in the exporter."""
         if not self._model:
             arg_id, arguments = self.serialize_arguments()
-            self._model = MutatorModel(type=self.type, args=arguments)
+            model = MutatorModel(type=self.type, args=arguments)
+            model._id = arg_id
+            self._model = model
 
         assert isinstance(self._model, MutatorModel)
         return self._model
--- a/src/inmanta/references.py
+++ b/src/inmanta/references.py
@@ -263,11 +263,11 @@ ArgumentTypes = typing.Annotated[
 """
 
 
-def get_id_reference_or_mutator(
-    type: ReferenceType,
+def compute_reference_or_mutator_id(
+    reference_type: ReferenceType,
     args: list[ArgumentTypes],
 ) -> uuid.UUID:
-    data = json.dumps({"type": type, "args": args}, default=util.api_boundary_json_encoder, sort_keys=True)
+    data = json.dumps({"type": reference_type, "args": args}, default=util.api_boundary_json_encoder, sort_keys=True)
     hasher = hashlib.md5()
     hasher.update(data.encode())
     return uuid.uuid3(uuid.NAMESPACE_OID, hasher.digest())
@@ -292,7 +292,7 @@ class MutatorModel(SerializedReferenceLike):
     """A mutator"""
 
     def get_id(self) -> uuid.UUID:
-        return get_id_reference_or_mutator(type=self.type, args=self.args)
+        return compute_reference_or_mutator_id(reference_type=self.type, args=self.args)
 
 
 C = typing.TypeVar("C", bound="ReferenceLike")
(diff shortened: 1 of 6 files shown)


src/inmanta/references.py
```diff
@@ -412,7 +427,9 @@ def serialize(self) -> MutatorModel:
         """Emit the correct pydantic objects to serialize the reference in the exporter."""
         if not self._model:
             arg_id, arguments = self.serialize_arguments()
-            self._model = MutatorModel(type=self.type, args=arguments)
+            model = MutatorModel(type=self.type, args=arguments)
+            model._id = arg_id
```
jptrindade, line 431: This looks weird. Would `model.get_id()` provide a different result? If so, and it is expected, I would expect `id` to be an optional member of the constructor

jptrindade (changes requested):

src/inmanta/references.py
arnaudsjs: `MutatorModel` is a pydantic type, so I had to follow the pydantic way of setting private attributes to make sure that _id attribute is not serialized. Apparently this is the way it should be done. `get_id()` doesn't provide a different result. By setting `_id` here we prevent that the has has to be calculated more than once.

jptrindade (approved):

src/inmanta/resources.py
```diff
@@ -187,7 +187,10 @@ def __new__(cls, class_name, bases, dct):
 
             fields.extend(dct["fields"])
 
-        dct["fields"] = tuple(set(fields))
+        # Deduplicate while preserving first-occurrence order. Using a set here would make the order depend on the
+        # per-process string hash seed (PYTHONHASHSEED),  producing a new desired state version even when nothing
+        # changed (see issue #10622).
+        dct["fields"] = tuple(dict.fromkeys(fields))
```
sanderr, line 193: Cool, I didn't know about this method. Simpler than the list comprehension I typically use.

src/inmanta/resources.py
```diff
@@ -281,23 +291,30 @@ class ReferenceCollector(ReferenceSubCollector):
 
     def __init__(self, resource: "Resource") -> None:
         super().__init__()
-        self.mutators: list[references.MutatorModel] = []
+        self.mutators: dict[uuid.UUID, references.MutatorModel] = {}
         self.resource = resource
 
+    def get_mutators_sorted(self) -> list["references.MutatorModel"]:
```
sanderr, line 297: I'm not entirely clear on why this is required, i.e. what is the root cause that causes this to be nondeterministic if we don't sort. But I also think it can't hurt.

src/inmanta/references.py
sanderr: I'm not following with this. `get_id` caches it, so why do we need to set it explicitly as `_id` here? Also, we first call `serialize_arguments()` that calculates the id, and then we construct `MutatorModel` to calculate it again.

No objection, just confused.

src/inmanta/references.py
```diff
@@ -263,6 +263,16 @@ def get_arg_value(
 """
 
 
+def compute_reference_or_mutator_id(
+    reference_type: ReferenceType,
+    args: list[ArgumentTypes],
```
sanderr, line 268: It's been a long time since I last saw this code, so I'm not very confident. But I wonder if `args` could contain (nested) sets or other nondeterministic values. That would make our uuid nondeterministic as well, meaning that to rely on it for sorting doesn't hold up in those cases.

Claude confirms my hypothesis, but I don't have a lot of confidence in it either tbh. Wdyt?

arnaudsjs: I considered this, but I concluded that this would not be possible because the Inmanta DSL doesn't support sets and the keys of dict are sorted by the serializer. But now that I think about it again, I realize that the serializer of a resource can serialize to any python type so also sets. I am going to create a little POC to validate this.

arnaudsjs: Sets are not json serializable so that's not an issue. But it is of course possible that a user uses a set in the exporter of a resource to remove duplicates and then converts it back to a list. This would cause hash instability, but I don't think we have control over that.

src/inmanta/resources.py
arnaudsjs: The previous implementation used a list of MutatorModel objects. The order in which these objects were populated in the list was undeterministic. To make the order (as as such the attribute hash consistent) we now sort the objects based on their uuid.

src/inmanta/references.py
arnaudsjs: I did that to make sure we don't have to calculate the id again when `mode.get_id()` is called.

PR #10735, opened 2026-08-21
scheduler: only redeploy resources affected by new version

### LINKED ISSUE #10591: resource scheduler: on new version, schedule only affected resources
The current implementation schedules the entire dirty set. Instead we should schedule only affected resources.

Open question: does this go into iso9 as well? It's a behavioral change, but one with relevant performance improvements, so I'm inclined to include it. Definitely not right before a release.

Current goal (agreed with @edvgui):
- [x] add to iso9 and iso10 with a config option to enable it. Something like `redeploy_failed_on_export=true` could be to the point.
- [x] Ask solutions (Guillaume) to enable the new behavior it on a bunch of solutions pipelines
- [ ] If there is no fallout, consider dropping the config option altogether. Drop it from iso10 either way, but first give solutions a chance to fix fallout.

Current draft in #10735

sanderr:
# Description

On model export, only redeploy resources affected by the new version, i.e. resources with changes to their intent or to their blocked status.

This should be a huge improvement in resource starvation for environments with a lot of failed resources. See the ticket for more details.

Status:
- [x] Implement fix
- [x] Add config option
- [x] Fix broken tests
- [x] Add new tests
- [ ] Merge and ask solutions team to enable it on their pipelines to battle-test it.

closes #10591

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/changelogs/unreleased/10591-redeploy-failed-on-export.yml
+++ b/changelogs/unreleased/10591-redeploy-failed-on-export.yml
@@ -1,0 +1,8 @@
+description: >
+  Added the temporary `scheduler.redeploy-failed-on-export` config option. When set to false, the scheduler only deploys
+  resources that are new, that have an updated desired state or that became unblocked when a new model version is
+  exported, rather than everything that is not in a known good state. It defaults to true, which retains the existing
+  behavior. This option is meant to be dropped again once the new behavior has been validated.
+issue-nr: 10591
+change-type: patch
+destination-branches: [master, iso9]
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -34,7 +34,7 @@ from typing import ClassVar, Optional, Self
 import asyncpg
 
 from inmanta import const, data, types
-from inmanta.agent import executor
+from inmanta.agent import config, executor
 from inmanta.agent.code_manager import CodeManager
 from inmanta.const import HandlerResourceState
 from inmanta.data import Environment
@@ -1190,8 +1190,11 @@ class ResourceScheduler(TaskManager):
             # - new_intent may contain undefineds
             # - unskipped may be transitively_blocked
             # - transitive_unblocked may be up to date if it was only blocked for a short time
-            # TODO: make this configurable for backwards compatibility
-            deploy_triggers: Set[ResourceIdStr] = (new_intent | unskipped | transitive_unblocked) & self._state.dirty
+            deploy_triggers: Set[ResourceIdStr] = []
+            if config.scheduler_redeploy_failed_on_export:
+                deploy_triggers = self._state.dirty
+            else:
+                deploy_triggers = (new_intent | unskipped | transitive_unblocked) & self._state.dirty
 
             # Remove timers for resources that are:
             #    - about to be triggered for deploy
--- a/src/inmanta/deploy/scheduler.py
+++ b/src/inmanta/deploy/scheduler.py
@@ -1110,6 +1110,7 @@ class ResourceScheduler(TaskManager):
         # assert invariants of the constructed sets
         assert len(intent_changes) == len(deleted.keys() | new | updated) == len(deleted) + len(new) + len(updated)
         assert len(became_defined | became_undefined) == (len(became_defined) + len(became_undefined))
+        new_intent: Set[ResourceIdStr] = new | updated  # subset of intent_changes that reflect *new* intent (incl undefined)
 
         # pass control to IO loop once more before we acquire the lock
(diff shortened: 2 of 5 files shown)


src/inmanta/deploy/state.py
```diff
@@ -689,6 +695,8 @@ def update_blocked_status_to_blocked(resource: "ResourceIdStr") -> bool:
                 if is_hard_block:
                     is_blocked.add(to_be_blocked)
                 my_state = self.resource_state[to_be_blocked]
+                # TODO: Claude flagged a pre-existing bug here, stating that the "not" should be dropped.
+                #       Probably best deferred to another ticket. Priotity TBD depending on impact.
```
arnaudsjs: I am looking into this, but I leave the comment in the code for now.

changelogs/unreleased/10591-redeploy-failed-on-export.yml
```diff
@@ -0,0 +1,8 @@
+description: >
+  Added the temporary `scheduler.redeploy-failed-on-export` config option. When set to false, the scheduler only deploys
+  resources that are new, that have an updated desired state or that became unblocked when a new model version is
+  exported, rather than everything that is not in a known good state. It defaults to true, which retains the existing
+  behavior. This option is meant to be dropped again once the new behavior has been validated.
```
jptrindade, line 5: What does this mean?

> This option is meant to be dropped again once the new behavior has been validated.

Do we intend for it to always be `True`?

src/inmanta/deploy/scheduler.py
```diff
@@ -1172,27 +1169,46 @@ async def _new_version(
...
+                    unskipped.add(resource)
+
+            transitive_unblocked: Set[ResourceIdStr]  # BLOCKED -> NOT_BLOCKED
+            transitive_blocked: Set[ResourceIdStr]  # NOT_BLOCKED / TEMPORARILY_BLOCKED -> BLOCKED
+            # update transitive state
+            transitive_unblocked, transitive_blocked = self._state.update_transitive_state(
+                new_undefined=became_undefined,
+                verify_blocked=added_requires.keys(),
+                verify_unblocked=became_defined | dropped_requires.keys(),
+            )
```
jptrindade, line 1182: Is there a reason why this jumped down?

src/inmanta/deploy/scheduler.py
```diff
@@ -1172,27 +1169,46 @@ async def _new_version(
...
 
             # Update set of in-progress deploys that became unmanaged
             self._deploying_unmanaged.update(self._deploying_latest & (new | deleted.keys()))
             # Update set of in-progress non-stale deploys by trimming resources with new state
             self._deploying_latest.difference_update(intent_changes.keys(), transitive_blocked)
 
+            # intersection with self._state dirty because
+            # - new_intent may contain undefineds
+            # - unskipped may be transitively_blocked
```
jptrindade, line 1191: Why? 
This was the definition above:
            `unskipped: set[ResourceIdStr] = set()  # TEMPORARILY_BLOCKED -> NOT_BLOCKED`

is this not the case anymore?

changelogs/unreleased/10591-redeploy-failed-on-export.yml
arnaudsjs: No. The idea is that the new behavior will be as if `scheduler.redeploy-failed-on-export` is set to `false`. The expectation is that this new behavior will not cause any problems, but we want to play it safe. Hence, we now introduce a configuration option that has a default to keep the behavior unchanged (`scheduler.redeploy-failed-on-export=true`). Guillaume is then going to disable this option on some labs to validate this is not going to cause problems. If so, we drop the configuration options and we use the new behavior all the time, i.e. don't redeploy failed resources on export.

src/inmanta/deploy/scheduler.py
arnaudsjs: In the previous implementation the order between these two didn't matter, because only the content of the dirty set would determine what had to be deployed. The new implementation (with `scheduler.redeploy_failed_on_export=false`) requires us to know whether a resource became unblocked, because those resources have to be deployed. For that reason the transitive update has to become last, because a resource might at the same time become unblocked due to a reset requirements and become blocked due to a dependency that became undefined.

src/inmanta/deploy/scheduler.py
arnaudsjs: See previous comment. A resource might become unblocked due to a reset dependency and at the same time blocked due to a dependency that became undefined.

jptrindade (approved):

PR #10750, opened 2026-08-25
Let a GraphQL contribution resolve its page columns, and expose version selection

jptrindade:
Claude (Claude Code) opened this PR on behalf of @jptrindade.

## What

An extension contributing to the `resources` GraphQL query had one way to add data to each row: contribute SQLAlchemy columns, which the framework added to the paginated statement. The database therefore computed that data for **every row the query could return before the page limit applied**, and again for `totalCount`.

`GraphQLContribution.get_sqlalchemy_columns` / `populate_sqlalchemy_columns` are replaced by `resolve_page_columns(session, rows, requested_fields)`, an async hook that `get_connection` runs on the fetched page. The extension fills in its data with a query of its own, over at most a page worth of rows.

Two additions let a contribution resolve that data at the right model version:

| addition                                     | what it is                                                                              | why an extension needs it                                                                                                          |
| -------------------------------------------- | ---------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
| `RESOLVED_MODEL_VERSION_FIELD`               | the version each returned row was resolved at, via `with_expression`                     | so page-resolved data is resolved at the same version selection that decided which rows came back                                      |
| `ResourceFilterABC.latest_available_version` | `coalesce(rps.orphaned_after, <latest scheduled>)`, on `resource_persistent_state` alone | so a component whose filter depends on the version can express it in `apply_filter_fast_count` and keep the optimized `totalCount` |

`filter_latest_available_version` is now that same expression with the `Configurationmodel.version ==` around it, so the two cannot drift. `latest_scheduled_version` (previously `_latest_scheduled_version_subquery`) is public for a component that has to build the coalesce on a second, aliased reference to the table.

With no extension able to contribute columns, `build_composed_sqlalchemy_model` loses its `contributions` parameter and its duplicate-column check.

## Why

This exists to serve inmanta-lsm, whose `serviceInstances` field and service-instance filters are the reason the nightly performance job reports seconds per page of `resources` at 200k resources. What this hook buys there is a **change of shape**: that extension's per-row data stops being O(environment size) and becomes O(page). The measured figures are in inmanta/inmanta-lsm#2490, along with an honest account of how stable they are.

## Safety of the new hook

`resolve_page_columns` hands a contribution the loaded, session-tracked row objects and lets it `setattr` its data onto them. That has one sharp edge, and the loop runs inside `session.no_autoflush` because of it: a name colliding with a **mapped** attribute would otherwise be flushed as an `UPDATE` by the next query on that session — a read-only GraphQL query issuing a write. With autoflush suspended a collision is a wrong answer rather than a write, and the hook documents the constraint that avoids it.

`apply_filter_fast_count` also documented a contract it does not have (*"any filters added here must only access ResourcePersistentState"*), which the first implementation of it already contradicts. It now states the invariant that actually makes the count correct: one row per `ResourcePersistentState` row, reaching other tables through a subquery rather than a join, and selecting the same resources as `apply_filter` at the version each is taken at.

### DIFF
--- a/changelogs/unreleased/graphql-resolve-page-columns.yml
+++ b/changelogs/unreleased/graphql-resolve-page-columns.yml
@@ -2,7 +2,8 @@ description: >
   A GraphQL contribution now fills in the data it adds to each row with `resolve_page_columns`, over the fetched page,
   instead of contributing SQLAlchemy columns the database had to compute for every row the query could return. This
   removes `GraphQLContribution.get_sqlalchemy_columns` and `populate_sqlalchemy_columns`. A component whose filter
-  depends on the model version a resource is taken at can now express it with `ResourceFilterABC.latest_available_version`
-  and so keep the optimized `totalCount`.
-change-type: major
+  depends on the model version a resource is taken at can read core's `isOrphan` off `ResourceFilterABC` and express
+  that version with its `latest_scheduled_version` or `latest_available_version`, and so keep the optimized
+  `totalCount`.
+change-type: minor
 destination-branches: [master, iso9]
--- a/changelogs/unreleased/graphql-resolve-page-columns.yml
+++ b/changelogs/unreleased/graphql-resolve-page-columns.yml
@@ -1,9 +1,8 @@
 description: >
-  Replaced the `GraphQLContribution.get_sqlalchemy_columns` / `populate_sqlalchemy_columns` pair with
-  `resolve_page_columns`, which lets an extension fill in the extra data it adds to each row of a fetched page with a
-  query of its own. The pair it replaces added that data to the paginated statement, so the database computed it for
-  every row the query could return before the page limit applied, and again for `totalCount`; the new hook runs once,
-  over at most a page worth of rows. Every row the `resources` query returns now also carries the model version it was
-  resolved at, so a contribution can resolve version-dependent data at the version that decided the page.
+  A GraphQL contribution now fills in the data it adds to each row with `resolve_page_columns`, over the fetched page,
+  instead of contributing SQLAlchemy columns the database had to compute for every row the query could return. This
+  removes `GraphQLContribution.get_sqlalchemy_columns` and `populate_sqlalchemy_columns`. A component whose filter
+  depends on the model version a resource is taken at can now express it with `ResourceFilterABC.latest_available_version`
+  and so keep the optimized `totalCount`.
 change-type: major
 destination-branches: [master, iso9]
--- a/changelogs/unreleased/graphql-resolve-page-columns.yml
+++ b/changelogs/unreleased/graphql-resolve-page-columns.yml
@@ -1,0 +1,9 @@
+description: >
+  Replaced the `GraphQLContribution.get_sqlalchemy_columns` / `populate_sqlalchemy_columns` pair with
+  `resolve_page_columns`, which lets an extension fill in the extra data it adds to each row of a fetched page with a
+  query of its own. The pair it replaces added that data to the paginated statement, so the database computed it for
+  every row the query could return before the page limit applied, and again for `totalCount`; the new hook runs once,
+  over at most a page worth of rows. Every row the `resources` query returns now also carries the model version it was
+  resolved at, so a contribution can resolve version-dependent data at the version that decided the page.
+change-type: major
+destination-branches: [master, iso9]
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -1067,6 +1067,36 @@ def decode_cursor(cursor: str) -> str:
     return decoded_cursor.split(prefix)[1]
 
 
+def bookmark_cursor(bookmark: str) -> str:
+    """
+    The position a sqlakeyset bookmark addresses, without the direction it was produced for. A bookmark carries that
+    direction as a leading '<' or '>', and it is the server's to choose when it pages -- `get_connection` puts one
+    back depending on whether it was given `after` or `before` -- so it is not part of the position.
+
+    The result is not yet a cursor a client can be handed: `encode_cursor`, or `Edge.resolve_edge` for an edge,
+    base64-encodes it first.
+
+    :param bookmark: a bookmark as sqlakeyset serialises it, e.g. from `Page.bookmark_items` or `Page.bookmark_next`.
+    """
+    return bookmark[1:]
+
+
+@dataclasses.dataclass(frozen=True)
+class PageItem:
+    """
+    One item of a fetched page: the object the statement selected, and the cursor addressing it.
+
+    Every statement `get_connection` pages selects a single entity, so `entity` is that entity itself rather than the
+    result row wrapping it.
+    """
+
+    cursor: str
+    """This item's position, as `bookmark_cursor` returns it: not yet base64-encoded, which `Edge.resolve_edge` does."""
+
+    entity: models.Base
+    """The ORM object the statement selected for this item."""
+
+
 def describe_modified_attributes(session: AsyncSession) -> str:
     """
     The mapped attributes that were modified on the objects the session tracks, as `<model>.<attribute>` entries, for
@@ -1101,7 +1131,18 @@ async def get_connection[*Ts](
     We do not call `ListConnection.resolve_connection` because:
      1) We already got the PageInfo arguments from sqlakeyset
      2) It calls `Edge.resolve_edge` with a cursor that is not useful to us
+
+    :param stmt: the statement to page. It must select exactly one ORM entity, which is then what each `PageItem`
+        carries and what the connection's nodes are resolved from. Rejected below if it does not.
     """
+    # Checked on the statement rather than on the fetched rows: how wide a row is is a property of the query, so this
+    # is one check instead of one per row, it runs before anything is executed, and it names the statement that is
+    # wrong rather than a row that looks odd. Without it, a statement selecting two entities would quietly serve
+    # nodes built from whichever one comes first.
+    selected = [description["name"] for description in stmt.column_descriptions]
+    if len(selected) != 1:
+        raise Exception(f"A paged statement must select exactly one entity, got {len(selected)}: {selected}.")
+
     async with get_session() as session:
         per_page: int
         # Get results per page and sanitation of input arguments
@@ -1139,16 +1180,18 @@ async def get_connection[*Ts](
 
         # Fetch the page using sqlakeyset
         result = await select_page(session, stmt, per_page=per_page, page=page)
-        page_rows = [(str(cursor)[1:], next(iter(value._mapping.values()))) for cursor, value in result.paging.bookmark_items()]
-        if contributions and page_rows:
+        page_items = [
+            PageItem(cursor=bookmark_cursor(bookmark), entity=row[0]) for bookmark, row in result.paging.bookmark_items()
+        ]
+        if contributions and page_items:
             # Let the contributions fill in the columns they resolve per page, before the rows go to Strawberry. They
             # do that by setting attributes on the rows, which are loaded and therefore tracked, so autoflush is
             # suspended: a name colliding with a mapped attribute stays a wrong answer instead of becoming a write.
             requested_fields = {to_snake_case(name) for name in get_selected_field_names(info)}
-            rows = [row for _, row in page_rows]
+            entities = [item.entity for item in page_items]
             with session.no_autoflush:
                 for contribution in contributions:
-                    await contribution.resolve_page_columns(session, rows, requested_fields)
+                    await contribution.resolve_page_columns(session, entities, requested_fields)
                     # Checked per contribution rather than once at the end, so the error names the one that did it.
                     # Not an assertion: this is the only thing standing between a contribution naming a mapped
                     # attribute and a read-only query issuing an UPDATE, and it has to hold when run with -O.
@@ -1164,18 +1207,18 @@ async def get_connection[*Ts](
         # it is fine to call them repeatedly
         connection = cast(type[CustomListConnection[Node]], mapper._connection_type_for(model))
 
-        for formatted_cursor, sqla_obj in page_rows:
+        for item in page_items:
             edge = cast(relay.Edge, mapper._edge_type_for(model))
-            node = connection.resolve_node(sqla_obj, info=info)
-            edges.append(edge.resolve_edge(cursor=formatted_cursor, node=node))
+            node = connection.resolve_node(item.entity, info=info)
+            edges.append(edge.resolve_edge(cursor=item.cursor, node=node))
 
         return connection(
             edges=edges,
             page_info=strawberry.relay.PageInfo(
                 has_next_page=result.paging.has_next,
                 has_previous_page=result.paging.has_previous,
-                start_cursor=encode_cursor(result.paging.bookmark_previous[1:]),
-                end_cursor=encode_cursor(result.paging.bookmark_next[1:]),
+                start_cursor=encode_cursor(bookmark_cursor(result.paging.bookmark_previous)),
+                end_cursor=encode_cursor(bookmark_cursor(result.paging.bookmark_next)),
             ),
             total_count=total_count,
         )
--- a/src/inmanta/graphql/schema.py
+++ b/src/inmanta/graphql/schema.py
@@ -380,8 +380,9 @@ def walk_selected_fields(selection: Selection) -> typing.Iterator[SelectedField]
 
 def get_selected_field_names(info: Info) -> set[str]:
     """
-    Return the (camelCase) names of every field selected anywhere in the query, recursing through nested
-    selections and fragments. This includes the names of the top-level resolved fields themselves (e.g. `resources`);
+    Return the (camelCase) names of every field selected anywhere in the selection the current field is resolved for,
+    recursing through nested selections and fragments. This includes the names of the top-level resolved fields
+    themselves (e.g. `resources`);
 
     :param info: the Strawberry resolver info holding the selected fields of the current query.
     """
@@ -799,7 +800,7 @@ class ResourceFilterABC(StrawberryFilter):
         """
         The model version a resource is taken at when no component takes over version selection: the currently
         scheduled version if the resource is still managed, or otherwise the last version it appeared in. Expressed on
-        ResourcePersistentState alone, so a component whose filter depends on the version can also express it in
+        ResourcePersistentState's own columns, so a component whose filter depends on the version can also express it in
         `apply_filter_fast_count`. A component that needs it on a second, aliased reference to that table can build it
         from `latest_scheduled_version`.
         """
@@ -807,10 +808,16 @@ class ResourceFilterABC(StrawberryFilter):
 
     def apply_filter_fast_count[*Ts](self, stmt: Select[tuple[*Ts]]) -> Select[tuple[*Ts]] | None:
         """
-        Apply this component's filter to the optimized total count query. Concretely, any filters added here must only
-        access ResourcePersistentState and version pinning to any version other than the latest for each resource is not
-        allowed. Returns None if this component has one or more filters that are not compatible with the optimized
-        mode.
+        Apply this component's filter to the optimized total count query, a count over ResourcePersistentState, which
+        holds exactly one row per resource. Returns None if this component has one or more filters that are not
+        compatible with the optimized mode, which falls the caller back to counting the whole statement.
+
+        Two things make an implementation correct. The statement has to stay one row per ResourcePersistentState row,
+        so narrow it with predicates and reach any other table through a subquery: a join both fans the count out and
+        risks colliding with the joins another component adds to the same statement. And it has to select the same
+        resources as `apply_filter` does, at the version each resource is taken at, which here can only be
+        `latest_available_version` -- ResourcePersistentState is not versioned, so a component pinning any other
+        version has to return None.
 
         The default implementation should suffice for most components. It disables the optimized query when at least one
         filter is present.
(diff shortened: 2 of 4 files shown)


src/inmanta/graphql/schema.py
```diff
@@ -1421,14 +1474,19 @@ class ContributableGraphQLType:
...
 # The object types extensions can register GraphQL contributions for (see GraphQLContribution), mapping each SQLAlchemy
 # model to its core building blocks. `get_schema` composes each of these from the core building blocks and the
 # registered contributions; registrations for any other model are rejected.
 CONTRIBUTABLE_MODELS: "Mapping[type[models.Base], ContributableGraphQLType]" = {
     models.Resource: ContributableGraphQLType(
-        core_mixin=CoreResourceMixin, base_filter=ResourceFilterABC, core_filter=CoreResourceFilter
+        core_mixin=CoreResourceMixin,
+        base_filter=ResourceFilterABC,
+        core_filter=CoreResourceFilter,
+        core_columns={RESOLVED_MODEL_VERSION_FIELD: query_expression()},
```
bartv: This stores a single `query_expression()` instance at module level, but every `get_schema()` call reuses it for a brand-new `ComposedResource` subclass. SQLAlchemy re-parents the shared MapperProperty onto the newest class, which corrupts previously built composed models and emits a duplicate-declarative-class `SAWarning` per rebuild. With `get_schema()` running twice in one process (GraphQLSlice restart, or the `schema = None` + `start()` pattern this PR's own tests use), a query against the first `ComposedResource` raises `AttributeError: 'NoneType' object has no attribute 'supports_population'` (reproduced on SQLAlchemy 2.0.49). On master, `get_sqlalchemy_columns()` returned a fresh `query_expression()` per build. Construct the `query_expression()` inside `get_schema()`/`build_composed_sqlalchemy_model()` instead of storing one instance in the module-level dict.

changelogs/unreleased/graphql-resolve-page-columns.yml
```diff
@@ -0,0 +1,9 @@
+description: >
+  A GraphQL contribution now fills in the data it adds to each row with `resolve_page_columns`, over the fetched page,
+  instead of contributing SQLAlchemy columns the database had to compute for every row the query could return. This
+  removes `GraphQLContribution.get_sqlalchemy_columns` and `populate_sqlalchemy_columns`. A component whose filter
+  depends on the model version a resource is taken at can read core's `isOrphan` off `ResourceFilterABC` and express
+  that version with its `latest_scheduled_version` or `latest_available_version`, and so keep the optimized
+  `totalCount`.
+change-type: minor
```
bartv: This should be `major`: the PR deletes `GraphQLContribution.get_sqlalchemy_columns`/`populate_sqlalchemy_columns`, extension-facing API that shipped in v19.0.0 and that inmanta-lsm implements. The merge tool computes version bumps from this field on every destination branch. An lsm that is not upgraded in lockstep with inmanta/inmanta-lsm#2490 installs cleanly against a minor core release (its core dependency range is open) while its GraphQL contribution silently stops working, with no version-constraint or release-note signal of a breaking change.

src/inmanta/graphql/schema.py
```diff
@@ -1133,45 +1203,34 @@ def get_target_model(cls) -> type[models.Base]:
...
-        """
-        Return SQLAlchemy column descriptors (e.g. `query_expression()`) keyed by attribute name. These are
-        merged onto a dynamically built subclass of the target model, so the query can select a single ORM object
-        that carries extra, SQL-backed columns (joins/subqueries) that don't live on the target's table. The value
-        of each column is populated per query by `populate_sqlalchemy_columns`.
-
-        For each column declared here, the matching output field should also be declared with a concrete type on
-        the mixin returned by `get_graphql_output_type_mixin` (otherwise the mapper would try to
-        auto-map the untyped column).
+    async def resolve_page_columns(
```
bartv: With the old hooks removed, nothing fails fast for an extension still implementing them: `register_graphql_contribution_for_extension` accepts a stale contribution class and its overridden hooks are silently never called. Concretely, released inmanta-lsm's `LsmGraphQLContribution` overrides exactly the removed methods, and its mixin resolver does `getattr(self, SERVICE_INSTANCES_FIELD, None)` -> `None` -> `[]`, so under version skew (any lsm without inmanta/inmanta-lsm#2490, including the window between the two merges) every resource's `serviceInstances` returns `[]` with no log, error, or startup failure. Rejecting contributions that still define the removed method names at registration time would turn this into a clear startup error.

src/inmanta/graphql/schema.py
```diff
@@ -1041,13 +1071,23 @@ async def get_connection[*Ts](
...
      1) We already got the PageInfo arguments from sqlakeyset
      2) It calls `Edge.resolve_edge` with a cursor that is not useful to us
+
+    :param stmt: the statement to page. It must select exactly one ORM entity, which is then what each `PageItem`
+        carries and what the connection's nodes are resolved from. Rejected below if it does not.
     """
+    # Checked on the statement rather than on every single fetched row.
+    # Without this, a statement selecting two entities would quietly serve nodes built from whichever one comes first.
+    selected = [description["name"] for description in stmt.column_descriptions]
+    if len(selected) != 1:
```
bartv: This checks `len(stmt.column_descriptions) == 1` but never verifies the one selected element is an ORM entity, so `select(models.Resource.resource_id)` passes the exact check whose error message says the statement "must select exactly one entity". `PageItem.entity` (typed `models.Base`) then holds a plain `str`, contributions doing `row.attr = ...` raise `AttributeError` on `str`, and `resolve_node` fails far from the actual mistake - the quiet failure this guard was added to catch. Checking `description["entity"] is not None` (it is `None` for plain columns) makes the guard match its message.

src/inmanta/graphql/schema.py
```diff
@@ -1133,45 +1203,34 @@ def get_target_model(cls) -> type[models.Base]:
...
-            added (e.g. via `with_expression`).
-        :param model: the dynamically built subclass of the target model that carries the `query_expression()`
-            placeholders (so its core columns keep their type, while the extra columns are read with `getattr`).
-        :param requested_fields: the (snake_case) names of every field selected anywhere in the GraphQL query
-            (not only the target's output fields), so a column counts as requested when its name appears in the set.
+        Runs after the page is fetched, so the work is bounded by the page. Adding the same data to the paginated
+        statement instead would have the database compute it for every row the query could return before the page
+        limit applies, and again for `totalCount`.
+
+        The attribute an implementation sets must be one the target model does not map, and distinct from the names
```
bartv: Distinctness of these attribute names used to be a machine-checked build-time invariant (the duplicate-column rejection in `build_composed_sqlalchemy_model`, plus its test, both deleted in this PR); it is now only this docstring request. Two extensions that stash into the same scratch attribute (their mixin fields differing, so the surviving mixin-collision check never fires) build fine, and at request time the later contribution overwrites the earlier one's values per row in registration order - one extension's field serves the other's data. On master this failed at server start with "Column ... defined more than once in Resource contributions". Can an enforced equivalent come back, e.g. contributions declaring the attribute names they write and registration checking for overlap?

src/inmanta/graphql/schema.py
```diff
@@ -1519,11 +1544,17 @@ async def environments(
...
             return await get_connection(
-                stmt, info=info, model="Environment", first=first, after=after, last=last, before=before
+                stmt,
+                info=info,
+                model="Environment",
+                first=first,
+                after=after,
+                last=last,
+                before=before,
+                contributions=extension_contributions.get(graphql_type_name(models.Environment), []),
```
bartv: Each of the three resolvers re-derives contributions inline like this (also lines 1582 and 1670) although the build loop in `get_schema` already computes exactly that per type (line 1513) and throws it away. `get_connection` also takes `model: str` and `contributions` separately - a pair that can silently disagree: a future query type passing `model` but omitting `contributions` gets no error, extension fields just return `null` via the `getattr(default=None)` resolvers. Store the per-type contributions in a `get_schema` local next to the composed model and pass that name, or resolve contributions inside `get_connection` from `model`.

src/inmanta/graphql/schema.py
```diff
@@ -1081,33 +1126,58 @@ async def get_connection[*Ts](
...
             ),
             total_count=total_count,
         )
 
 
-# TODO: once everything is in (including the lsm side), do a critical final pass about how everything fits together,
-# and whether the are redundancies, verbosities, complexties, ... that we can clean up by tweaking the interface.
 class GraphQLContribution(ABC):
     """
     Extension hook that lets extensions (e.g. LSM) contribute extra information to a single GraphQL output type
```
bartv: The rest of this class docstring (just below this hunk) still says core "merges them onto the GraphQL output type and its backing SQLAlchemy model (see build_strawberry_output_type and build_composed_sqlalchemy_model)", but this PR removed exactly that: contributions no longer touch the SQLAlchemy model and `build_composed_sqlalchemy_model` no longer takes contributions. An extension author reading the hook's own docstring is pointed at the deleted column-contribution mechanism. Drop "and its backing SQLAlchemy model" and the `build_composed_sqlalchemy_model` reference.

changelogs/unreleased/graphql-resolve-page-columns.yml
```diff
@@ -0,0 +1,9 @@
+description: >
```
bartv: The description is three sentences of internal API rationale; the hook-method names mean nothing to release-note readers. One short user-facing sentence, e.g. "Improved the performance of the resources GraphQL query when extensions contribute data to it." - the rationale lives in the PR description.

src/inmanta/graphql/schema.py
```diff
@@ -1081,33 +1126,58 @@ async def get_connection[*Ts](
...
         # We use the private methods for the mapper because their respective public attributes like `mapper.connection_types`
         # Are only filled when the private methods are called first. The private methods use the public attributes as cache so
         # it is fine to call them repeatedly
         connection = cast(type[CustomListConnection[Node]], mapper._connection_type_for(model))
 
-        for cursor, value in result.paging.bookmark_items():
-            formatted_cursor = str(cursor)[1:]
-            sqla_obj = next(iter(value._mapping.values()))
+        for item in page_items:
             edge = cast(relay.Edge, mapper._edge_type_for(model))
```
bartv: `mapper._edge_type_for(model)` is loop-invariant but evaluated once per page item, while the equivalent `_connection_type_for` lookup was hoisted above the loop. Hoist this one next to it - the asymmetry also makes a reader hunt for a per-item dependency that does not exist.

src/inmanta/graphql/schema.py
bartv, line 1489: This stores a single `query_expression()` instance at module level, but every `get_schema()` call reuses it for a brand-new `ComposedResource` subclass. SQLAlchemy re-parents the shared MapperProperty onto the newest class, which corrupts previously built composed models and emits a duplicate-declarative-class `SAWarning` per rebuild. With `get_schema()` running twice in one process (GraphQLSlice restart, or the `schema = None` + `start()` pattern this PR's own tests use), a query against the first `ComposedResource` raises `AttributeError: 'NoneType' object has no attribute 'supports_population'` (reproduced on SQLAlchemy 2.0.49). On master, `get_sqlalchemy_columns()` returned a fresh `query_expression()` per build. Construct the descriptor per build instead of storing one instance in the module-level dict; #10776 does this by making `core_columns` a factory.

src/inmanta/graphql/schema.py
```diff
@@ -1081,33 +1126,58 @@ async def get_connection[*Ts](
...
+        if contributions and page_items:
+            # Let the contributions fill in the columns they resolve per page, before the rows go to Strawberry. They
+            # do that by setting attributes on the rows, which are loaded and therefore tracked, so autoflush is suspended.
+            requested_fields = {to_snake_case(name) for name in selected_fields}
+            entities = [item.entity for item in page_items]
+            with session.no_autoflush:
+                for contribution in contributions:
+                    await contribution.resolve_page_columns(session, entities, requested_fields)
+                    # Checked per contribution rather than once at the end, so the error names the one that did it.
+                    if session.dirty:
```
bartv, line 1142: This guard inspects only `session.dirty`, so the docstring's promise ("Setting a mapped attribute is rejected") only holds for un-flushed attribute assignment. It misses `session.add` (`session.new`), `session.delete` (`session.deleted`), direct `session.execute(update/delete)`, a contribution calling `flush()`/`commit()` itself (which clears `dirty`; `commit` even persists the write and expires the page entities), and in-place mutation of mapped JSONB dicts: `Resource.attributes` has no `MutableDict`, so `row.attributes['x'] = 1` fires no event yet is served in the response. Both the JSONB mutation and a raw `execute(update(...))` were verified to pass silently.

The underlying issue is that the hook's only return channel is mutating session-tracked rows. The resource-filters design proposed the other shape for this: per-request DataLoaders in the context, with the field's resolver loading through them, so the extension never sees the rows and there is nothing to guard. #10776 implements that on top of this branch; measured against this PR it has the same scaling and costs a constant 3 to 10 ms per page of 100. If the hook stays, please check `session.new` and `session.deleted` too and document the vectors the guard cannot see.

src/inmanta/graphql/schema.py
```diff
@@ -1062,9 +1102,14 @@ async def get_connection[*Ts](
         else:
             per_page = DEFAULT_PER_PAGE
 
+        if per_page > APILIMIT:
```
bartv, line 1105: This caps only the upper bound. `first: -2` passes through to sqlakeyset, which compiles `LIMIT -1` (`q.limit(per_page + 1)`, no validation of its own); PostgreSQL rejects that (`LIMIT must not be negative`) and the client gets an opaque internal error instead of the clean 400 this check produces. `first: 0` and `first: -1` silently return an empty page. Strawberry's own `first < 0` check lives in `resolve_connection`, which this path deliberately bypasses. Add a `per_page < 1` branch to this same check.

Separately: this is a sixth hand-rolled copy of the APILIMIT validation (`DataView.validate_limit` in `dataview.py:470`, plus inline copies in `resourceservice.py:250`, `agentmanager.py:593`, `orchestrationservice.py:445` and `479`, `compilerservice.py:1253`), with a third message wording where the existing sites all say `limit parameter can not exceed {APILIMIT}, got {limit}.` Consider hoisting the existing check into a shared helper next to `APILIMIT`, parameterised on the parameter name.

src/inmanta/graphql/schema.py
```diff
@@ -1062,9 +1102,14 @@ async def get_connection[*Ts](
...
+            raise BadRequest(f"`first`/`last` can not exceed {APILIMIT}, got {per_page}.")
```
bartv, line 1106: Raising `BadRequest` (a tornado `web.HTTPError`) inside the resolver leaks tornado's formatting into the client-visible GraphQL error: the message becomes `HTTP 400: Bad Request (Invalid request: 'first'/'last' can not exceed 1000, got 1001.)` (verified response content), unlike every other `get_connection` validation error, which is a plain `Exception` message. The exception's own 400 status is never consulted either: `GraphQLResult` derives 400 from `data` being `None`. The test at `tests/test_graphql.py:441` passes only because it asserts a substring. Raise a plain `Exception` here like the sibling checks do.

changelogs/unreleased/graphql-resolve-page-columns.yml
bartv, line 8: This should be `major`: the PR deletes `GraphQLContribution.get_sqlalchemy_columns`/`populate_sqlalchemy_columns`, extension-facing API that shipped in v19.0.0 and that inmanta-lsm implements. The merge tool computes version bumps from this field on every destination branch. An lsm that is not upgraded in lockstep with inmanta/inmanta-lsm#2490 installs cleanly against a minor core release (its core dependency range is open) while its GraphQL contribution silently stops working, with no version-constraint or release-note signal of a breaking change.

src/inmanta/graphql/schema.py
bartv, line 1213: With the old hooks removed, nothing fails fast for an extension still implementing them: `register_graphql_contribution_for_extension` accepts a stale contribution class and its overridden hooks are silently never called. Concretely, released inmanta-lsm's `LsmGraphQLContribution` overrides exactly the removed methods, and its mixin resolver does `getattr(self, SERVICE_INSTANCES_FIELD, None)` -> `None` -> `[]`, so under version skew (any lsm without inmanta/inmanta-lsm#2490, including the window between the two merges) every resource's `serviceInstances` returns `[]` with no log, error, or startup failure. Rejecting contributions that still define the removed method names turns this into a clear error; #10776 does it in `GraphQLContribution.__init_subclass__`, at class definition.

src/inmanta/graphql/schema.py
bartv, line 1088: This checks `len(stmt.column_descriptions) == 1` but never verifies the one selected element is an ORM entity, so `select(models.Resource.resource_id)` passes the exact check whose error message says the statement "must select exactly one entity" (`entity` is `Resource` for that scalar column too). `PageItem.entity` (typed `models.Base`) then holds a plain `str`, contributions doing `row.attr = ...` raise `AttributeError` on `str`, and `resolve_node` fails far from the actual mistake, the quiet failure this guard was added to catch. Compare `description["expr"]` with the model the query is meant to select (the class itself for an entity select, a column attribute for a scalar one); #10776 does that against `BuiltType.model`.

src/inmanta/graphql/schema.py
bartv, line 1224: Distinctness of these attribute names used to be a machine-checked build-time invariant (the duplicate-column rejection in `build_composed_sqlalchemy_model`, plus its test, both deleted in this PR); it is now only this docstring request. Two extensions that stash into the same scratch attribute (their mixin fields differing, so the surviving mixin-collision check never fires) build fine, and at request time the later contribution overwrites the earlier one's values per row in registration order: one extension's field serves the other's data. On master this failed at server start with "Column ... defined more than once in Resource contributions". With resolvers loading through their own DataLoaders (#10776) there is no shared namespace to police; if the hook stays, contributions should declare the names they write so registration can check for overlap.

src/inmanta/graphql/schema.py
```diff
@@ -1604,7 +1641,11 @@ async def resources(
             if version_handler is None:
                 stmt = CoreResourceFilter.filter_latest_available_version(stmt, environment=filter.environment)
 
-            stmt = populate_extension_columns(stmt, models.Resource, resource_model, info)
+            # Carry the version each row is taken at, so a contribution resolving version-dependent data for the fetched page
+            # resolves it at the same version the row was selected at.
+            stmt = stmt.options(
+                with_expression(getattr(resource_model, RESOLVED_MODEL_VERSION_FIELD), models.Configurationmodel.version)
```
bartv, line 1647: `RESOLVED_MODEL_VERSION_FIELD` is declared generically via `core_columns` (line 1489) but populated only by this hand-placed `with_expression`, 160 lines from the declaration, and `query_expression()` defaults to SQL `NULL`. If a refactor drops this on a resource-paging path, contributions silently read `resolved_model_version = None` and resolve version-dependent data at no version: wrong or empty extension data, no error. inmanta/inmanta-lsm#2490 does not trust the contract either (its `get_resolved_model_version` raises when the field is unpopulated), which is the right instinct but has to be repeated by every contribution. Also, the generic `resolve_page_columns` docstring names the field without scoping it to Resource; a contribution for Environment/Notification that follows it gets an `AttributeError`. #10776 declares and populates the column in one helper (`select_resources`, next to the version selection it belongs with) and gives contributions `resolved_model_version(row)`, which raises on `None`.

src/inmanta/graphql/schema.py
bartv, line 1557: Each of the three resolvers re-derives contributions inline like this (also lines 1582 and 1670) although the build loop in `get_schema` already computes exactly that per type (line 1513) and throws it away. `get_connection` also takes `model: str` and `contributions` separately, a pair that can silently disagree: a future query type passing `model` but omitting `contributions` gets no error, extension fields just return `null` via the `getattr(default=None)` resolvers. One value per built type carrying model, output type, filter input and components (`BuiltType` in #10776) removes the pairing.

src/inmanta/graphql/schema.py
bartv, line 1183: The rest of this class docstring (just below this hunk) still says core "merges them onto the GraphQL output type and its backing SQLAlchemy model (see build_strawberry_output_type and build_composed_sqlalchemy_model)", but this PR removed exactly that: contributions no longer touch the SQLAlchemy model and `build_composed_sqlalchemy_model` no longer takes contributions. An extension author reading the hook's own docstring is pointed at the deleted column-contribution mechanism. Drop "and its backing SQLAlchemy model" and the `build_composed_sqlalchemy_model` reference.

changelogs/unreleased/graphql-resolve-page-columns.yml
bartv, line 1: The description is three sentences of internal API rationale; the hook-method names mean nothing to release-note readers. One short user-facing sentence, e.g. "Improved the performance of the resources GraphQL query when extensions contribute data to it." The rationale lives in the PR description.

src/inmanta/graphql/schema.py
bartv, line 1165: `mapper._edge_type_for(model)` is loop-invariant but evaluated once per page item, while the equivalent `_connection_type_for` lookup was hoisted above the loop. Hoist this one next to it; the asymmetry also makes a reader hunt for a per-item dependency that does not exist.

src/inmanta/graphql/schema.py
```diff
@@ -774,12 +769,36 @@ def handles_version(self) -> bool:
...
+        a larger statement. It is NULL if the scheduler has not processed any version yet.
+        """
+        return (
+            select(models.Scheduler.last_processed_model_version)
+            .where(models.Scheduler.environment == environment)
+            .scalar_subquery()
+        )
+
+    @classmethod
+    def latest_available_version(cls, *, environment: uuid.UUID) -> SQLColumnExpression[int | None]:
```
bartv, line 785: `latest_available_version(cls, *, environment)` is keyword-only while its sibling `latest_scheduled_version(cls, environment)` just above takes the same single parameter positionally, and both are new public API for extensions. inmanta/inmanta-lsm#2490 already trips over the split: its `apply_filter_fast_count` calls one positionally and the other with the keyword in the same expression. Pick one convention (simplest: drop the `*`). More broadly, the reason lsm needs these two at all is to re-derive core's version selection in its fast count; #10776 hands components the selection as a value (`VersionSelection`) instead, so neither classmethod is needed.

sanderr (commented):
Deferred until we have more context, as discussed.

PR #10758, opened 2026-08-28
Replace redeploy-failed-on-export config option with environment setting

arnaudsjs:
# Description

Replaced the temporary `scheduler.redeploy-failed-on-export` config option with the `redeploy_failed_on_export` environment setting, so that the behavior can be configured per environment at runtime, without restarting the scheduler.

The setting keeps the meaning of the config option: it defaults to true, which retains the existing behavior, and when set to false the scheduler only deploys resources that are new, that have an updated desired state or that became unblocked when a new model version is exported.

part of #10591

# Self Check:

- [x] Changelog entry

This pull request was created by Claude on behalf of Arnaud.


src/inmanta/deploy/scheduler.py
```diff
@@ -1110,6 +1118,9 @@ async def _new_version(
         assert len(became_defined | became_undefined) == (len(became_defined) + len(became_undefined))
         new_intent: Set[ResourceIdStr] = new | updated  # subset of intent_changes that reflect *new* intent (incl undefined)
 
+        # Fetch this setting before acquiring the lock, because everything below the lock is synchronous.
+        redeploy_failed: bool = await self.get_redeploy_failed_on_export(connection=connection)
```
jptrindade: redeploy_failed made it seem like we tried a redeploy and it failed
```suggestion
        redeploy_failed_resources: bool = await self.get_redeploy_failed_on_export(connection=connection)
```

jptrindade (changes requested):
We now have nothing saying that this is a temporary change, is that still the case? 
Also, I have this a bit in cold memory, but would these also deploy non_compliant report only resources? If so, I would suggest changing the name/docstring to reflect those as well

sanderr (approved):

sanderr:
> would these also deploy non_compliant report only resources

No, those are currently only included in repairs, though they do get the deploy timer for faster retries. It might be something to revisit at some point: once this new option becomes the default, one of the arguments against including non-compliant resources in deploys disappears.

arnaudsjs:
> We now have nothing saying that this is a temporary change, is that still the case?

This is intentional. We are going to make this option permanent.

jptrindade (approved):

PR #10765, opened 2026-08-31
Disable redeploy_failed_on_export environment setting by default

### LINKED ISSUE #10591: resource scheduler: on new version, schedule only affected resources

arnaudsjs:
# Description

This PR makes sure that the redeploy_failed_on_export environment setting is now disabled by default.

Closes #10591

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/flip-default-redeploy-failed-on-export.yml
+++ b/changelogs/unreleased/flip-default-redeploy-failed-on-export.yml
@@ -1,0 +1,8 @@
+description: >-
+  Flipped the default value of the `redeploy_failed_on_export` environment setting from true to false.
+issue-nr: 10591
+change-type: major
+destination-branches: [master]
+sections:
+  upgrade-note: >-
+    The `redeploy_failed_on_export` environment setting now defaults to false.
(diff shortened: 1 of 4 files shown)


sanderr (approved):
I'm surprised that there aren't more test failures tbh. Might be worth running the full suite if you didn't already.

arnaudsjs:
> I'm surprised that there aren't more test failures tbh. Might be worth running the full suite if you didn't already.

Me too. But the slow tests succeed to. I didn't check inmanta-lsm or the lsm module yet.

PR #10768, opened 2026-09-01
Get all parent entities in consistent order

arnaudsjs:
# Description

Added a method to get all the parents of a certain entity in a sorted/stable order.

Required by: https://github.com/inmanta/lsm/pull/1011

# Self Check:

- [x] Changelog entry


Hugo-Inmanta (approved):

arnaudsjs:
We probably want to update the `get_default_values()` and the `get_all_attributes()` methods so that they use the new `get_all_parent_entities_sorted()` method instead of `self.parent_entities`. Like that all resolution logic uses the same entity traversal algorithm consistently.

src/inmanta/ast/entity.py
```diff
@@ -237,11 +238,37 @@ def get_all_parent_names(self) -> "List[str]":
 
         return parents
 
+    def _get_all_parent_entities_sorted(self) -> Mapping["Entity", None]:
+        """
+        Helper method to return the parent entities of this entity as the keys
+        of a dictionary. This method uses a dictionary to remove duplicates and
+        to keep track of order, since iterating over a dictionary respects
+        insertion order. Iterating over the returned dictionary sorts
+        the parent entities in parent-to-child and right-to-left order.
```
sanderr: Right-to-left so that left can override right?

arnaudsjs: Correct. I'll update the docstring.

sanderr (approved):

PR #10777, opened 2026-09-02
Fix resolution attributes and defaults

arnaudsjs:
# Description

Fixed the resolution of attributes and default values for an entity that inherits the same attribute via multiple parents. The definition made by the most derived entity in the inheritance hierarchy is now always used. Between unrelated parents, the attribute of the left-most parent is used.

# Self Check:

- [x] Changelog entry
- [x]  Code is clear and sufficiently documented

### DIFF
--- a/tests/compiler/test_define.py
+++ b/tests/compiler/test_define.py
@@ -254,3 +254,56 @@ end
         "__config__::Left",
     ]
     assert leaf.get_all_parent_entities() == set(leaf.get_all_parent_entities_sorted())
+
+
+def test_attribute_shadowing_in_diamond_hierarchy(snippetcompiler) -> None:
+    """
+    Verify that, when an attribute is defined more than once in the inheritance hierarchy,
+    Entity.get_all_attributes() reports the attribute of the most derived entity that defines it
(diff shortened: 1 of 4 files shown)


tests/compiler/test_define.py
```diff
@@ -254,3 +254,56 @@ def get_sorted_parent_names(entity_name: str) -> list[str]:
...
+
+    assert get_default("Base", "shadowed") == 1
+    assert get_default("Redefines", "shadowed") == 3
+    assert get_default("Inherits", "shadowed") == 1
+    assert get_default("Leaf", "shadowed") == 3
+
+    assert get_default("Base", "removed") == 1
+    assert get_default("Redefines", "removed") is None
+    assert get_default("Inherits", "removed") == 1
+    assert get_default("Leaf", "removed") is None
```
jptrindade, line 309: I'm not sure why this is not 1.

arnaudsjs: `undef` removes the default value of an attribute. The nearest parent from Leaf is Refines, which removes the default. Hance there is no default anymore and the result becomes None.

jptrindade (approved):

sanderr (approved):

PR #10781, opened 2026-09-03
Clean up module code when a model version is deleted

### LINKED ISSUE #10778: Module code is never cleaned up when a model version is deleted (iso9, iso10)
# Module code is never cleaned up when a model version is deleted (iso9, iso10)

Labels: `bug`, `server`

---

*Filed by Claude on behalf of @Hugo-Inmanta.*

## Problem

Deleting a model version leaves the code of the inmanta modules it used behind in the database, permanently: `inmanta_module` and `module_files` rows are never removed. This affects both the explicit `delete_version` endpoint and the automatic cleanup of old versions (`OrchestrationService._purge_versions`, which calls the same `ConfigurationModel.delete_cascade`), so it accumulates on its own on any running orchestrator.

## Cause

`ConfigurationModel.delete_cascade` (`src/inmanta/data/__init__.py`) cleans the module tables in this order:

```python
await AgentModules.delete_version(environment=self.environment, model_version=self.version, connection=con)
await InmantaModule.delete_version(environment=self.environment, model_version=self.version, connection=con)
await ModuleFiles.delete_version(environment=self.environment, model_version=self.version, connection=con)
```

Both `InmantaModule.delete_version` and `ModuleFiles.delete_version` (`src/inmanta/data/sqlalchemy.py`) pick their targets with a subquery over `agent_modules` **for the version being deleted**:

```sql
DELETE FROM inmanta_module
WHERE (environment, name, version) IN (
    SELECT environment, inmanta_module_name, inmanta_module_version
    FROM public.agent_modules
    WHERE environment=$1 AND cm_version=$2
)
```

`AgentModules.delete_version`, one line earlier, has already deleted exactly those rows. The subquery returns nothing, so both statements are unconditional no-ops.

Reordering is not the fix: run them first and the `RESTRICT` foreign key from the *other* versions still using the same module version blocks the delete. The premise is wrong — a module version is shared by every model version that uses it (a module rarely changes between compiles), so it can only be deleted once the last of them is gone.

## Impact

Row growth in `inmanta_module` and `module_files`, proportional to the number of distinct module versions ever exported, unbounded in time. No functional breakage: re-registering the same module version is an `ON CONFLICT DO NOTHING` insert, so stale rows are simply reused. This is *row* growth only — content in the `file` table is never reclaimed in any case (separate, older gap), so the leak adds nothing beyond the module metadata itself.

`Environment.clear` is not affected: it deletes `agent_modules` before `inmanta_module`, so the environment-wide path does free everything. Only the version-delete path leaks.

## Affected versions

Verified present, with identical code and call order, on `iso9` (core 18.3.0), `iso9-stable`/`iso9-next` (18.2.0), `iso10-stable`/`iso10-next` (19.0.0) and `master`. Introduced by d59916f2c ("Rework the code transport between the compiler and the agent", issue #8729, PR #8816), first released in core **v16.0.0**. `iso8` (15.5.7) is **not** affected: it predates the rework, and its `Code` table was cleaned up correctly with `Code.delete_all(environment=..., version=...)`.

## Why this is not a cherry-pick from master

Fixed on master on branch `issue/improve-agent-code-install-stage2-install-refinements` (changelog `cleanup-module-code-on-version-delete.yml`). That fix keys the "unused module" check off the `configurationmodel_modules` table, which exists only on master (migration `v202608070`, unreleased). On iso9/iso10 the only table referencing `inmanta_module` with `ON DELETE RESTRICT` is `agent_modules` — `module_files` references it with `CASCADE` — so the check reduces to a single `NOT EXISTS`.

### Suggested patch

In `src/inmanta/data/sqlalchemy.py`, replace both `delete_version` methods with an unused-module sweep sharing one subquery:

```python
# Subquery selecting the inmanta modules of one environment ($1) that no model version uses any more. A module version
# is shared by every model version that uses it, so it can only be deleted along with the last one. agent_modules is
# the only table that references inmanta_module with ON DELETE RESTRICT, which makes deleting what this returns safe.
_UNUSED_INMANTA_MODULES = """
    SELECT unused_module.environment, unused_module.name, unused_module.version
    FROM public.inmanta_module AS unused_module
    WHERE unused_module.environment=$1
    AND NOT EXISTS (
        SELECT 1
        FROM public.agent_modules AS agent_module
        WHERE agent_module.environment=unused_module.environment
        AND agent_module.inmanta_module_name=unused_module.name
        AND agent_module.inmanta_module_version=unused_module.version
    )
"""
```

```python
class InmantaModule(Base):
    @classmethod
    async def delete_unused(cls, environment: uuid.UUID, connection: asyncpg.connection.Connection) -> None:
        """
        Delete the inmanta modules of the given environment that no model version uses any more. Expected to be called
        once the registrations of a deleted model version are gone, as the last step of its cleanup: the files of these
        modules have to be deleted first (see ModuleFiles.delete_unused).
        """
        await connection.execute(
            f"""
            DELETE FROM {cls.__tablename__}
            WHERE (environment, name, version) IN ({_UNUSED_INMANTA_MODULES})
            """,
            environment,
        )


class ModuleFiles(Base):
    @classmethod
    async def delete_unused(cls, environment: uuid.UUID, connection: asyncpg.connection.Connection) -> None:
        """
        Delete the files of the inmanta modules of the given environment that no model version uses any more. Expected
        to be called before those modules themselves are deleted (see InmantaModule.delete_unused): a module still has
        to be there to be found unused.
        """
        await connection.execute(
            f"""
            DELETE FROM {cls.__tablename__}
            WHERE (environment, inmanta_module_name, inmanta_module_version) IN ({_UNUSED_INMANTA_MODULES})
            """,
            environment,
        )
```

and in `ConfigurationModel.delete_cascade`:

```python
# Drop the module registrations of this version, then the code of the modules that this leaves unused. A module
# version is shared by every model version that uses it, so it outlives this one unless it was the last to use it.
await AgentModules.delete_version(environment=self.environment, model_version=self.version, connection=con)
await ModuleFiles.delete_unused(environment=self.environment, connection=con)
await InmantaModule.delete_unused(environment=self.environment, connection=con)
```

The files are deleted explicitly, before the module they belong to, rather than through the `CASCADE` on `inmanta_module` — matching the "this method doesn't rely on DELETE CASCADE" convention of `delete_cascade`.

Because the sweep is environment-wide, it also reclaims the rows that leaked before the fix, the next time any version of that environment is deleted. No data migration needed, and no schema change at all.

Optional tidy-up while in there: in `Environment.clear`, `ModuleFiles.delete_all` runs *after* `InmantaModule.delete_all`, so the `CASCADE` has already removed its rows and the call is dead. Moving it before makes it do the work it looks like it does.

### Test

`tests/test_server.py`, next to `test_clear_environment`: two versions sharing one module version, plus a second module used only by version 2. Deleting version 2 must leave the shared module (version 1 still uses it) and remove the other one with its files; deleting version 1 must leave nothing. Assert row counts in `agent_modules`, `inmanta_module` and `module_files` for the environment at each step — it fails on all three steps before the fix.

Two gotchas: the version has to carry a resource for the agent the module is registered for, because `agent_modules` has a foreign key to `agent` and an agent row only exists once a resource is assigned to it; and set `AUTO_DEPLOY` to false so that no version is released and `delete_version` doesn't refuse the active version. The iso9 `module_version_info` shape is:

```python
InmantaModuleDTO(
    name="shared",
    version="abc",
    files_in_module=[ModuleSourceMetadata(name="inmanta_plugins.shared", is_byte_code=False, hash_value=uploaded_hash)],
    requirements=[],
    for_agents=["agent1"],
)
```

Changelog: `change-type: patch`, with the maintenance branches being fixed as `destination-branches` — master is covered separately by `cleanup-module-code-on-version-delete.yml`.

Hugo-Inmanta:
*Opened by Claude on behalf of @Hugo-Inmanta.*

Fixes #10778.

## Problem

Deleting a model version left the code of the inmanta modules it used behind in the database, permanently. This hit both the explicit `delete_version` endpoint and the automatic cleanup of old versions (`OrchestrationService._purge_versions`, which calls the same `ConfigurationModel.delete_cascade`), so it accumulated on its own on any running orchestrator.

`InmantaModule.delete_version` and `ModuleFiles.delete_version` both picked their targets with a subquery over `agent_modules` **for the version being deleted** — rows that `AgentModules.delete_version` had already deleted one line earlier in `delete_cascade`. The subquery returned nothing, so both statements were unconditional no-ops.

Impact is row growth in `inmanta_module` and `module_files`, proportional to the number of distinct module versions ever exported, unbounded in time. No functional breakage: re-registering the same module version is an `ON CONFLICT DO NOTHING` insert, so stale rows were simply reused.

## Fix

Reordering the existing calls does not work: a module version is shared by every model version that uses it (a module rarely changes between compiles), so it can only be deleted once the last of them is gone — and the `ON DELETE RESTRICT` foreign key from the *other* versions' `agent_modules` rows blocks the delete otherwise.

Instead, both methods become an unused-module sweep over one shared subquery (`_UNUSED_INMANTA_MODULES`): the modules of the environment that no `agent_modules` row references any more. `agent_modules` is the only table that references `inmanta_module` with `ON DELETE RESTRICT` (`module_files` references it with `CASCADE`), so deleting what the subquery returns is safe. `delete_cascade` drops the registrations of the deleted version first, then the files and then the modules that this leaves unused — the files explicitly rather than through the cascade, matching the "this method doesn't rely on DELETE CASCADE" convention of that method.

Because the sweep is environment wide, it also reclaims the rows that leaked before this fix, the next time any version of that environment is deleted. **No schema change and no data migration.** The anti-join is served by the existing `agent_modules_environment_module_name_module_version_index`.

`Environment.clear` was never affected — it deletes `agent_modules` before `inmanta_module`, so the environment-wide path did free everything. Its `ModuleFiles.delete_all` call ran *after* `InmantaModule.delete_all` though, so the cascade had already removed its rows and the call was dead; it now runs before, in the same order as `delete_cascade`.

## Test

`tests/test_server.py::test_delete_version_cleans_up_module_code`, next to `test_clear_environment`: two versions sharing one module version, a second module used only by version 2, and a third registered for no version at all (the shape of a pre-fix leaked row). Deleting version 2 must leave the shared module and reclaim the other two; deleting version 1 must leave nothing. It asserts row counts in `agent_modules`, `inmanta_module` and `module_files` and fails on all three steps before the fix. `test_clear_environment` now registers a module too, so it locks the environment-wide path.

Verified locally: `tests/test_server.py`, `tests/test_data.py`, `tests/test_data_concurrency.py`, `tests/deploy/e2e/test_code_loading.py` and `tests/deploy/e2e/test_facts.py` all pass; `black`/`isort`/`flake8` clean and `make mypy` exits 0.

## Notes for the reviewer

- `destination-branches: [master, iso9]`. The issue suggested leaving master out because a fix for it exists on the `issue/improve-agent-code-install-stage2-install-refinements` branch, but master's module tables and their `delete_version` code are byte-identical to iso9's today, so master is affected and unfixed until that branch lands. Once this is merged, the equivalent commit on that branch should be dropped, and its `configurationmodel_modules` table added as a second `NOT EXISTS` clause in `_UNUSED_INMANTA_MODULES` instead (that table is a second `ON DELETE RESTRICT` referrer of `inmanta_module`, so the sweep has to check it too). Its `cleanup-module-code-on-version-delete.yml` changelog entry should go as well, in favour of this one.
- iso8 predates the code-transport rework (`d59916f2c`, first released in core v16.0.0) and cleaned its `Code` table up correctly, so it is not affected.
- The two sweep statements run in separate transactions, so an export that re-registers the very module version being reclaimed, concurrently with a version delete on the same environment, can make the `inmanta_module` delete hit the `agent_modules` foreign key. That surfaces as an error on `delete_version` (or a logged error in `_purge_versions`, self-healing on the next run), not as data loss, and the pre-existing code had the same exposure — flagging it rather than growing the scope of a patch fix.

### DIFF
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -2939,9 +2939,12 @@ class Environment(BaseDocument):
             await Parameter.delete_all(environment=self.id, connection=con)
             await Notification.delete_all(environment=self.id, connection=con)
 
+            # The registrations of a module first, then its files and only then the module itself: agent_modules
+            # references a module with ON DELETE RESTRICT, and its files are deleted here rather than through the
+            # cascade on the module.
             await AgentModules.delete_all(environment=self.id, connection=con)
-            await InmantaModule.delete_all(environment=self.id, connection=con)
             await ModuleFiles.delete_all(environment=self.id, connection=con)
+            await InmantaModule.delete_all(environment=self.id, connection=con)
 
             await DiscoveredResource.delete_all(environment=self.id, connection=con)
             await EnvironmentMetricsGauge.delete_all(environment=self.id, connection=con)
@@ -6762,9 +6765,12 @@ class ConfigurationModel(BaseDocument):
             await Compile.delete_all(environment=self.environment, version=self.version, connection=con)
             await DryRun.delete_all(environment=self.environment, model=self.version, connection=con)
 
+            # Drop the module registrations of this version, then the code of the modules that this leaves unused. A
+            # module version is shared by every model version that uses it, so it outlives this one unless it was the
+            # last to use it.
             await AgentModules.delete_version(environment=self.environment, model_version=self.version, connection=con)
-            await InmantaModule.delete_version(environment=self.environment, model_version=self.version, connection=con)
-            await ModuleFiles.delete_version(environment=self.environment, model_version=self.version, connection=con)
+            await ModuleFiles.delete_unused(environment=self.environment, connection=con)
+            await InmantaModule.delete_unused(environment=self.environment, connection=con)
 
             await UnknownParameter.delete_all(environment=self.environment, version=self.version, connection=con)
             await self._execute_query(
--- a/src/inmanta/data/sqlalchemy.py
+++ b/src/inmanta/data/sqlalchemy.py
@@ -118,6 +118,23 @@ class SetValidatedMixin:
             )
 
 
+# Subquery selecting the inmanta modules of one environment ($1) that no model version uses any more. A module version
+# is shared by every model version that uses it, so it can only be deleted along with the last one. agent_modules is the
+# only table that references inmanta_module with ON DELETE RESTRICT, which makes deleting what this returns safe.
+_UNUSED_INMANTA_MODULES = """
+    SELECT unused_module.environment, unused_module.name, unused_module.version
+    FROM public.inmanta_module AS unused_module
+    WHERE unused_module.environment=$1
+    AND NOT EXISTS (
+        SELECT 1
+        FROM public.agent_modules AS agent_module
+        WHERE agent_module.environment=unused_module.environment
+        AND agent_module.inmanta_module_name=unused_module.name
+        AND agent_module.inmanta_module_version=unused_module.version
+    )
+"""
+
+
 class InmantaModule(Base):
     __tablename__ = "inmanta_module"
 
@@ -225,21 +242,18 @@ class InmantaModule(Base):
             )
 
     @classmethod
-    async def delete_version(
-        cls, environment: uuid.UUID, model_version: int, connection: asyncpg.connection.Connection
-    ) -> None:
+    async def delete_unused(cls, environment: uuid.UUID, connection: asyncpg.connection.Connection) -> None:
+        """
+        Delete the inmanta modules of the given environment that no model version uses any more. Expected to be called
+        once the registrations of a deleted model version are gone, as the last step of its cleanup: the files of these
+        modules have to be deleted first (see ModuleFiles.delete_unused).
+        """
         await connection.execute(
             f"""
-            DELETE FROM {InmantaModule.__tablename__}
-            WHERE (environment, name, version) IN (
-                SELECT environment, inmanta_module_name, inmanta_module_version
-                FROM public.agent_modules
-                WHERE environment=$1
-                AND cm_version=$2
-            )
+            DELETE FROM {cls.__tablename__}
+            WHERE (environment, name, version) IN ({_UNUSED_INMANTA_MODULES})
             """,
             environment,
-            model_version,
         )
 
 
@@ -275,21 +289,18 @@ class ModuleFiles(Base):
     file: Mapped["File"] = relationship("File", back_populates="module_files")
 
     @classmethod
-    async def delete_version(
-        cls, environment: uuid.UUID, model_version: int, connection: asyncpg.connection.Connection
-    ) -> None:
+    async def delete_unused(cls, environment: uuid.UUID, connection: asyncpg.connection.Connection) -> None:
+        """
+        Delete the files of the inmanta modules of the given environment that no model version uses any more. Expected
+        to be called before those modules themselves are deleted (see InmantaModule.delete_unused): a module still has
+        to be there to be found unused.
+        """
         await connection.execute(
             f"""
-            DELETE FROM {ModuleFiles.__tablename__}
-            WHERE (environment, inmanta_module_name, inmanta_module_version) IN (
-                SELECT environment, inmanta_module_name, inmanta_module_version
-                FROM {AgentModules.__tablename__}
-                WHERE environment=$1
-                AND cm_version=$2
-            )
+            DELETE FROM {cls.__tablename__}
+            WHERE (environment, inmanta_module_name, inmanta_module_version) IN ({_UNUSED_INMANTA_MODULES})
             """,
             environment,
-            model_version,
         )
 
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -17,13 +17,14 @@ Contact: code@inmanta.com
 """
 
 import asyncio
+import base64
 import functools
 import json
 import logging
 import os
 import sys
 import uuid
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Mapping, Sequence
 from datetime import UTC, datetime, timedelta, timezone
 from functools import partial
 
@@ -36,6 +37,8 @@ from inmanta.agent import executor
 from inmanta.const import ParameterSource
 from inmanta.data import AUTO_DEPLOY, ResourcePersistentState
 from inmanta.data.model import AttributeStateChange
+from inmanta.data.model import InmantaModule as InmantaModuleDTO
+from inmanta.data.model import ModuleSourceMetadata
 from inmanta.deploy import persistence, state
 from inmanta.protocol import Client
 from inmanta.resources import Id
@@ -45,6 +48,7 @@ from inmanta.server.bootloader import InmantaBootloader
 from inmanta.server.protocol import ServerStartFailure
 from inmanta.server.services.databaseservice import PostgreSQLVersion
 from inmanta.types import ResourceIdStr, ResourceVersionIdStr
+from sqlalchemy import func, select
 from utils import insert_with_link_to_configuration_model, log_contains, log_doesnt_contain, retry_limited
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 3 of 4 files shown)


src/inmanta/data/__init__.py
```diff
@@ -2939,9 +2939,12 @@ async def clear(self, connection: Optional[asyncpg.connection.Connection] = None
             await Parameter.delete_all(environment=self.id, connection=con)
             await Notification.delete_all(environment=self.id, connection=con)
 
+            # The registrations of a module first, then its files and only then the module itself: agent_modules
+            # references a module with ON DELETE RESTRICT, and its files are deleted here rather than through the
+            # cascade on the module.
             await AgentModules.delete_all(environment=self.id, connection=con)
-            await InmantaModule.delete_all(environment=self.id, connection=con)
             await ModuleFiles.delete_all(environment=self.id, connection=con)
```
Hugo-Inmanta, line 2946: The old order was making `ModuleFiles.delete_all` a noop since we already cascade when deleting InmantaModule.
This new order is explicitely not relying on the cascade, which is in line with the docstring a couple of line up.

src/inmanta/data/__init__.py
```diff
@@ -6762,9 +6765,12 @@ async def delete_cascade(self, connection: Optional[asyncpg.connection.Connectio
             await Compile.delete_all(environment=self.environment, version=self.version, connection=con)
             await DryRun.delete_all(environment=self.environment, model=self.version, connection=con)
 
+            # Drop the module registrations of this version, then the code of the modules that this leaves unused. A
+            # module version is shared by every model version that uses it, so it outlives this one unless it was the
+            # last to use it.
             await AgentModules.delete_version(environment=self.environment, model_version=self.version, connection=con)
-            await InmantaModule.delete_version(environment=self.environment, model_version=self.version, connection=con)
-            await ModuleFiles.delete_version(environment=self.environment, model_version=self.version, connection=con)
```
Hugo-Inmanta, line 6767: This is where the big bug is: if we first delete a given model version from `AgentModules,` then both `InmantaModule.delete_version` and `ModuleFiles.delete_version` calls were actually noops since they were looking for the version that was just deleted in `AgentModules`.

src/inmanta/data/__init__.py
```diff
@@ -2939,9 +2939,12 @@ async def clear(self, connection: Optional[asyncpg.connection.Connection] = None
             await Parameter.delete_all(environment=self.id, connection=con)
             await Notification.delete_all(environment=self.id, connection=con)
 
+            # The registrations of a module first, then its files and only then the module itself: agent_modules
+            # references a module with ON DELETE RESTRICT, and its files are deleted here rather than through the
+            # cascade on the module.
```
arnaudsjs, line 2944: I don't understand this comment.

src/inmanta/data/__init__.py
```diff
@@ -6762,9 +6765,12 @@ async def delete_cascade(self, connection: Optional[asyncpg.connection.Connectio
             await Compile.delete_all(environment=self.environment, version=self.version, connection=con)
             await DryRun.delete_all(environment=self.environment, model=self.version, connection=con)
 
+            # Drop the module registrations of this version, then the code of the modules that this leaves unused. A
+            # module version is shared by every model version that uses it, so it outlives this one unless it was the
+            # last to use it.
```
arnaudsjs, line 6770: Same here. I don't understand this comment.

src/inmanta/data/sqlalchemy.py
```diff
@@ -225,21 +242,18 @@ async def register_modules(
             )
 
     @classmethod
-    async def delete_version(
-        cls, environment: uuid.UUID, model_version: int, connection: asyncpg.connection.Connection
-    ) -> None:
+    async def delete_unused(cls, environment: uuid.UUID, connection: asyncpg.connection.Connection) -> None:
+        """
+        Delete the inmanta modules of the given environment that no model version uses any more. Expected to be called
+        once the registrations of a deleted model version are gone, as the last step of its cleanup: the files of these
```
arnaudsjs, line 248: This PR often talks about registrations. What does a registration mean?

tests/test_server.py
```diff
@@ -440,6 +502,101 @@ async def compile_done():
...
+        )
+
+    assert await get_module_code_row_counts(environment) == {
+        "agent_modules": 3,
+        "inmanta_module": 3,
+        "module_files": 3,
+    }
+
+    # Version 1 still uses the shared module, so only the code of the dropped module is cleaned up. The cleanup is
+    # environment wide, so it reclaims the already unused module as well.
```
arnaudsjs, line 582:
```suggestion
    # Version 1 still uses the shared module, so only the code of the dropped module and the unused module are cleaned up.
```

tests/test_server.py
```diff
@@ -440,6 +502,101 @@ async def compile_done():
...
 
+    assert await get_module_code_row_counts(environment) == {
+        "agent_modules": 0,
+        "inmanta_module": 0,
+        "module_files": 0,
+    }
+
+
+@pytest.mark.parametrize("no_agent", [True])
+async def test_delete_version_cleans_up_module_code(client, server, clienthelper, environment) -> None:
```
arnaudsjs, line 513: Let's extend the test so that it also creates modules in another environment, to make sure we respect environment scoping.

arnaudsjs (changes requested):

src/inmanta/data/sqlalchemy.py
Hugo-Inmanta: That's the semantics for the AgentModules table that is used to register which agents use which Inmanta modules in which cm version. But this is really not obvious with only the context of this current PR, I'll improve the comments. 
I tend to forget that some of these Claude comments make sense to me in a broader context, but they are actually only useful if they are standalone.

Hugo-Inmanta:
Superseded by https://github.com/inmanta/inmanta-core/pull/10783

PR #10786, opened 2026-09-04
Add a ResourceSetListener extension point on the resource write path

jptrindade:
_Opened by Claude on behalf of @jptrindade._

An extension that maintains data derived from the resources of a model version has no way of being told they were
written. The two existing listeners core offers are notifications after the (i.e. a compile completing (`CompileStateListener`)an environment changing (`EnvironmentListener`) ).

With the `ResourceSetListener` we wanted to guarantee that the information derived from the resource sets was written in the same transaction, and if the listener somehow failed, that we are able to rollback the export.

To let a listener scope its work to what actually changed, `ResourceSet.insert_sets_and_resources` now returns the ids of the updated resource sets.

Nothing registers a listener in core, so there is no behaviour change without an extension.
The extension side is inmanta/inmanta-lsm#2499.

### DIFF
--- a/changelogs/unreleased/10780-resource-set-listener.yml
+++ b/changelogs/unreleased/10780-resource-set-listener.yml
@@ -1,0 +1,7 @@
+description: >
+  Added the `ResourceSetListener` extension point. A listener registered on the orchestration service is notified, in
+  the transaction that writes them, of the resource sets a model version was written with. Unlike the compile and
+  environment listeners it takes part in that transaction, so an extension can maintain data derived from the
+  resources without it ever being committed out of step with them.
+change-type: minor
+destination-branches: [master]
--- a/src/inmanta/data/__init__.py
+++ b/src/inmanta/data/__init__.py
@@ -5105,7 +5105,7 @@ class ResourceSet(BaseDocument):
         deleted_resource_sets: Optional[abc.Set[str]] = None,
         *,
         connection: asyncpg.connection.Connection,
-    ) -> None:
+    ) -> abc.Set[uuid.UUID]:
         """
         Inserts resources and resource sets and links resource sets to the target version.
 
@@ -5127,6 +5127,8 @@ class ResourceSet(BaseDocument):
         :param deleted_resource_sets: These are the resource set names from the base version which were removed
             in this partial compile. Not applicable for a full compile.
         :param connection: The connection to use. Must be in a transaction context.
+        :return: The ids the updated resource sets were inserted under. Resource sets that were linked to the target
+            version unchanged are not included.
         """
 
         is_partial_update = base_version is not None
@@ -5192,7 +5194,7 @@ class ResourceSet(BaseDocument):
         # insert the updated resource sets and resources into the database and link everything together
         # (resource -> set and set -> model)
         with pyformance.timer("sql.insert_sets_and_resources.insert").time():
-            await cls._execute_query(
+            inserted_resource_sets = await cls._fetch_query(
                 """\
                 -- insert resource sets and keep track of name-id mapping
                 WITH inserted_resource_sets AS (
@@ -5224,36 +5226,42 @@ class ResourceSet(BaseDocument):
                         UNNEST($9::text[]) AS attribute_hash,
                         UNNEST($10::boolean[]) AS is_undefined,
                         UNNEST($11::text[]) AS resource_set
+                ), inserted_resources AS (
+                    -- insert resources
+                    INSERT INTO public.resource(
+                        environment,
+                        resource_id,
+                        resource_type,
+                        resource_id_value,
+                        agent,
+                        attributes,
+                        attribute_hash,
+                        is_undefined,
+                        resource_set
+                    )
+                    SELECT
+                        $1,
+                        r.resource_id,
+                        r.resource_type,
+                        r.resource_id_value,
+                        r.agent,
+                        r.attributes,
+                        r.attribute_hash,
+                        r.is_undefined,
+                        rs.id
+                    FROM resource_data AS r
+                    -- this join has been tested to be up to four times faster than joining with
+                    -- resource_configuration_model, even if the latter would have the name column directly
+                    -- (for 5k models, 5k sets, updating 1-1000 sets, with 100-10k resources per set).
+                    -- Order of magnitude for reference: 0.5s when updating 10 sets with 1k resources per set.
+                    INNER JOIN inserted_resource_sets AS rs
+                        ON r.resource_set IS NOT DISTINCT FROM rs.name
                 )
-                -- insert resources
-                INSERT INTO public.resource(
-                    environment,
-                    resource_id,
-                    resource_type,
-                    resource_id_value,
-                    agent,
-                    attributes,
-                    attribute_hash,
-                    is_undefined,
-                    resource_set
-                )
-                SELECT
-                    $1,
-                    r.resource_id,
-                    r.resource_type,
-                    r.resource_id_value,
-                    r.agent,
-                    r.attributes,
-                    r.attribute_hash,
-                    r.is_undefined,
-                    rs.id
-                FROM resource_data AS r
-                -- this join has been tested to be up to four times faster than joining with
-                -- resource_configuration_model, even if the latter would have the name column directly
-                -- (for 5k models, 5k sets, updating 1-1000 sets, with 100-10k resources per set).
-                -- Order of magnitude for reference: 0.5s when updating 10 sets with 1k resources per set.
-                INNER JOIN inserted_resource_sets AS rs
-                    ON r.resource_set IS NOT DISTINCT FROM rs.name
+                -- The ids the sets were inserted under. gen_random_uuid() produces them inside this statement, so
+                -- returning them here is what spares the caller a second query to find them. A data-modifying CTE
+                -- runs to completion whether or not the primary query reads it, so wrapping the resource insert in
+                -- one does not make it conditional.
+                SELECT irs.id FROM inserted_resource_sets AS irs
                 """,
                 *common_values,
                 cls._get_value(updated_resource_sets),
@@ -5277,6 +5285,8 @@ class ResourceSet(BaseDocument):
                     updated_resource_sets=updated_resource_sets,
                     connection=connection,
                 )
+
+        return {cast(uuid.UUID, record["id"]) for record in inserted_resource_sets}
 
     @classmethod
     async def clear_resource_sets_in_version(
--- a/src/inmanta/server/services/orchestrationservice.py
+++ b/src/inmanta/server/services/orchestrationservice.py
@@ -56,6 +56,7 @@ from inmanta.server import (
 from inmanta.server import config as opt
 from inmanta.server import diff, protocol
 from inmanta.server.services import resourceservice
+from inmanta.server.services.resourcesetlistener import ResourceSetListener
 from inmanta.server.validate_filter import InvalidFilter
 from inmanta.types import Apireturn, JsonType, PrimitiveTypes, ResourceIdStr, ResourceVersionIdStr, ReturnTupple
 
@@ -381,6 +382,14 @@ class OrchestrationService(protocol.ServerSlice):
 
     def __init__(self) -> None:
         super().__init__(SLICE_ORCHESTRATION)
+        self.resource_set_listeners: list[ResourceSetListener] = []
+
+    def add_resource_set_listener(self, listener: ResourceSetListener) -> None:
+        """
+        Register a listener to be notified, in the transaction that writes them, of the resource sets a model version
+        was written with. Listeners are registered while the server starts, before the API becomes available.
+        """
+        self.resource_set_listeners.append(listener)
 
     def get_dependencies(self) -> list[str]:
         return [SLICE_RESOURCE, SLICE_AGENT_MANAGER, SLICE_DATABASE]
@@ -892,7 +901,7 @@ class OrchestrationService(protocol.ServerSlice):
 
             all_ids: set[Id] = {Id.parse_id(rid, version) for rid in rid_to_resource.keys()}
             try:
-                await data.ResourceSet.insert_sets_and_resources(
+                written_resource_sets = await data.ResourceSet.insert_sets_and_resources(
                     environment=env.id,
                     updated_resources=list(rid_to_resource.values()),
                     target_version=version,
@@ -902,6 +911,10 @@ class OrchestrationService(protocol.ServerSlice):
                 )
             except data.InvalidResourceSetMigration as e:
                 raise BadRequest(e.message)
+            # Deliberately not guarded against failure: a listener maintains data derived from these resources, so it
+            # has to be committed with them or not at all.
+            for listener in self.resource_set_listeners:
+                await listener.resource_sets_written(env.id, version, written_resource_sets, connection=connection)
             await cm.recalculate_total(connection=connection)
             await data.UnknownParameter.insert_many(unknowns, connection=connection)
 
--- a/src/inmanta/server/services/resourcesetlistener.py
+++ b/src/inmanta/server/services/resourcesetlistener.py
@@ -1,0 +1,54 @@
+"""
+Copyright 2026 Inmanta
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+Contact: code@inmanta.com
+"""
+
+import abc
+import uuid
+from collections import abc as collections_abc
+
+import asyncpg
+
+
+class ResourceSetListener(abc.ABC):
+    """
+    Base class for listeners on the resources of a model version being written.
+
+    Unlike the compile and environment listeners, this one takes part in the transaction that writes the resources: it
+    is handed that transaction's connection, and an exception it raises aborts the export. A listener that cannot
+    offer that guarantee does not belong here, because whatever it maintains would then be committed out of step with
+    the resources it is derived from.
+    """
+
+    @abc.abstractmethod
+    async def resource_sets_written(
+        self,
+        environment: uuid.UUID,
+        model_version: int,
+        resource_sets: collections_abc.Set[uuid.UUID],
+        *,
+        connection: asyncpg.connection.Connection,
+    ) -> None:
+        """
+        Called when the resources of a model version have been written, before the transaction commits.
+
+        :param environment: The environment the resources were written for.
+        :param model_version: The model version the resource sets were linked to.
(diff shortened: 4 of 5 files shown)


jptrindade (commented):
_Review posted by Claude on behalf of @jptrindade._

## Summary

Self-review of the `ResourceSetListener` extension point. The shape is right: in-transaction, connection handed in, failure deliberately not swallowed, and returning the generated ids from `insert_sets_and_resources` is the correct call since `gen_random_uuid()` makes them unobservable from outside the statement. Wrapping the resource insert in a data-modifying CTE is sound — such a CTE runs to completion whether or not the primary query reads it, and this statement already relied on that for `linked_resource_sets`.

Five comments inline. None of them block; the two that matter are the missing failure context (an extension taking exports down as a bare 500 is hard to diagnose) and the under-specified contract (the call site sits between `insert_sets_and_resources` and `cm.recalculate_total()`, and that ordering is load-bearing for anyone writing derived rows).

One thing not attached to a line: the class docstring should state that a listener runs on the critical path of **every** export, inside the transaction, so its cost is added to every export and to the time `resource`/`resource_set` locks are held. That is the guidance an extension author needs before deciding to register one, and it is the counterpart to the guarantee the class already documents.

See inmanta/inmanta-lsm#2499 for the review of the consumer side; two of the findings there argue for dropping the `model_version` argument's only use, but the parameter is cheap and plausibly useful, so no change requested here.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -902,6 +911,10 @@ async def _put_version(
                 )
             except data.InvalidResourceSetMigration as e:
                 raise BadRequest(e.message)
+            # Deliberately not guarded against failure: a listener maintains data derived from these resources, so it
+            # has to be committed with them or not at all.
+            for listener in self.resource_set_listeners:
+                await listener.resource_sets_written(env.id, version, written_resource_sets, connection=connection)
```
jptrindade, line 917: The whole premise of this extension point is that a listener failure aborts the export. Right now that surfaces as a bare 500 with nothing saying an extension caused it, or which one. Given how much this costs the caller, it is worth the four lines:

```suggestion
            # Deliberately not guarded against failure: a listener maintains data derived from these resources, so it
            # has to be committed with them or not at all.
            for listener in self.resource_set_listeners:
                try:
                    await listener.resource_sets_written(env.id, version, written_resource_sets, connection=connection)
                except Exception:
                    LOGGER.error(
                        "Resource set listener %s failed for version %d of environment %s; the export is aborted.",
                        type(listener).__name__,
                        version,
                        env.id,
                    )
                    raise
```

src/inmanta/server/services/resourcesetlistener.py
```diff
@@ -0,0 +1,54 @@
...
+    async def resource_sets_written(
+        self,
+        environment: uuid.UUID,
+        model_version: int,
+        resource_sets: collections_abc.Set[uuid.UUID],
+        *,
+        connection: asyncpg.connection.Connection,
+    ) -> None:
+        """
+        Called when the resources of a model version have been written, before the transaction commits.
```
jptrindade, line 46: The contract is missing what a listener may assume about the state of the transaction it is handed. The call sits between `insert_sets_and_resources` and `cm.recalculate_total()`, and that position is load-bearing for anyone writing rows derived from these resources. Worth spelling out here:

- the `configurationmodel` row for `model_version` exists, and so do the `resource`, `resource_set` and `resource_set_configuration_model` rows for `resource_sets` — a listener may reference all four
- `cm.recalculate_total()` has **not** run yet and the unknowns are **not** yet inserted, so `configurationmodel.total` is not final
- `resource_sets` includes the shared set (the one with `name IS NULL`) whenever it was written. A listener that filters for named sets will silently miss it, which is exactly the kind of thing that is invisible until a partial export in production.

src/inmanta/server/services/resourcesetlistener.py
```diff
@@ -0,0 +1,54 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+Contact: code@inmanta.com
+"""
+
+import abc
+import uuid
+from collections import abc as collections_abc
```
jptrindade, line 21: `collections_abc` as an alias appears nowhere else in core. The established pairing when both `abc`s are needed is `import abc` plus a direct name import — see `src/inmanta/data/dataview.py`:

```python
import abc
import uuid
from collections.abc import Set
```

with `resource_sets: Set[uuid.UUID]` at the signature below.

changelogs/unreleased/10780-resource-set-listener.yml
```diff
@@ -0,0 +1,7 @@
+description: >
```
jptrindade, line 1: The filename points at #10780, which is _"Explain the List modified after freeze error and suggest how to fix the model"_ — unrelated to this change. The file also carries no `issue-nr`, and by the convention in this directory the numeric prefix _is_ the issue number.

Either attach the real issue, or drop the prefix and name it `resource-set-listener.yml`; there are already entries here without a number (`graphql-version-selection.yml`, `add-method-get-all-parent-entities.yml`).

tests/server/test_resource_set_listener.py
```diff
@@ -0,0 +1,177 @@
...
+        raise Exception("this listener cannot do its work")
+
+
+@contextlib.contextmanager
+def registered(orchestration_service: OrchestrationService, listener: ResourceSetListener) -> abc.Iterator[None]:
+    """
+    Register a listener for the duration of the test. Production code registers during server startup and never
+    unregisters, so there is no API for this.
+    """
+    orchestration_service.resource_set_listeners.append(listener)
```
jptrindade, line 68: Appending to the list directly means the only public entry point, `add_resource_set_listener`, is never exercised by a test.

```suggestion
    orchestration_service.add_resource_set_listener(listener)
```

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -381,6 +382,14 @@ class OrchestrationService(protocol.ServerSlice):
 
     def __init__(self) -> None:
         super().__init__(SLICE_ORCHESTRATION)
+        self.resource_set_listeners: list[ResourceSetListener] = []
+
+    def add_resource_set_listener(self, listener: ResourceSetListener) -> None:
```
sanderr: It might receive resource sets as argument, but isn't it more a `model_version_listener` / `put_version_listener`?

src/inmanta/server/services/orchestrationservice.py
sanderr, line 909: Don't we need to communicate these as well? I believe lsm might need it

src/inmanta/data/__init__.py
```diff
@@ -5224,36 +5225,38 @@ async def insert_sets_and_resources(
...
-                    r.agent,
-                    r.attributes,
-                    r.attribute_hash,
-                    r.is_undefined,
-                    rs.id
-                FROM resource_data AS r
-                -- this join has been tested to be up to four times faster than joining with
-                -- resource_configuration_model, even if the latter would have the name column directly
-                -- (for 5k models, 5k sets, updating 1-1000 sets, with 100-10k resources per set).
-                -- Order of magnitude for reference: 0.5s when updating 10 sets with 1k resources per set.
```
sanderr, line 5254: Why drop this part of the comment?

src/inmanta/data/__init__.py
```diff
@@ -5224,36 +5225,38 @@ async def insert_sets_and_resources(
...
-                INNER JOIN inserted_resource_sets AS rs
-                    ON r.resource_set IS NOT DISTINCT FROM rs.name
+                -- The ids the sets were inserted under. gen_random_uuid() produces them inside this statement, so
+                -- returning them here is what spares the caller a second query to find them.
+                SELECT irs.id FROM inserted_resource_sets AS irs
```
sanderr, line 5261: Doesn't `INSERT` have `RETURNING`, which would require less of a change?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -902,6 +911,20 @@ async def _put_version(
                 )
             except data.InvalidResourceSetMigration as e:
                 raise BadRequest(e.message)
+            # A listener failure aborts the export. A listener maintains data derived from these
+            # resources, so it has to be committed with them or not at all. The handler below only names the
+            # listener that failed, it does not swallow.
```
sanderr, line 916: We may need to refine this if we ever add more listeners (perhaps some are less strict on the all-or-nothing part. But for now this makes sense.

src/inmanta/server/services/model_version_listener.py
sanderr, line 1: I'd like Arnaud's review on this module when he's back. Not merge-blocking imo, but could you make sure to ask him?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -381,6 +382,14 @@ class OrchestrationService(protocol.ServerSlice):
 
     def __init__(self) -> None:
         super().__init__(SLICE_ORCHESTRATION)
+        self.resource_set_listeners: list[ResourceSetListener] = []
+
+    def add_resource_set_listener(self, listener: ResourceSetListener) -> None:
+        """
+        Register a listener to be notified of the resource sets a model version was written with,
+        in the transaction that writes them. Listeners are registered while the server starts, before the API becomes available.
```
sanderr, line 390: So during `prestart`? Could be useful to be explicit in the docstring

src/inmanta/server/services/model_version_listener.py
sanderr, line 1: This isn't really `snake_case`. I'm assuming Claude based this on the `environmentlistener`, which has the same issue. I'm still inclined to do it right for this one.

src/inmanta/server/services/resourcesetlistener.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+import abc
+import uuid
+from collections.abc import Set
+
+import asyncpg
+
+
+class ResourceSetListener(abc.ABC):
```
sanderr: I'd prefer a slightly more generic name (and accompanying method + field in orchestrationservice), like `ModelVersionListener` or `ExportListener`. If we then ever want to evolve it by adding more notification methods, the class name remains appropriate.

sanderr (changes requested):

src/inmanta/data/__init__.py
jptrindade: I thought it was added in my session. Claude likes to post statistics in comments but I find them to not be relevant, and I believe @arnaudsjs shares that opinion as well. I left part of the comment that still justifies the join.

src/inmanta/server/services/orchestrationservice.py
jptrindade: The lsm part as it stands doesn't need this. If the desired state of an instance changed to delete a resource set, then only the new sets will be present on the new desired state version (and conversely, only those sets will appear on the new table)

src/inmanta/server/services/orchestrationservice.py
jptrindade: Not exactly. The lsm part needs the database slice to be started, so it actually happens in `start` (to fetch the `InstanceResourceRepository`).

src/inmanta/data/__init__.py
sanderr: I understand. And I feel the same in many cases. In this case I added the comment (iirc) in order to support the decision for this approach (which took a quite some time to arrive at; to be fair, with Claude it might be easier nowadays). I wanted tobe clear on the circumstances in which it was chosen, so that we have context if we ever consider changing the query in the future.

src/inmanta/server/services/orchestrationservice.py
sanderr: Right, of course. And if the transfer's own set is deleted, the latest model version wouldn't see the service at all, which is appropriate.

src/inmanta/data/__init__.py
jptrindade: It does, but the last `INSERT` is on the resource table and not on the resource_set table

src/inmanta/data/__init__.py
jptrindade: Ok, I'll revert this change.

src/inmanta/server/services/orchestrationservice.py
jptrindade: I agree that we can leave this for when we need to add more listeners

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -382,14 +382,14 @@ class OrchestrationService(protocol.ServerSlice):
 
     def __init__(self) -> None:
         super().__init__(SLICE_ORCHESTRATION)
-        self.resource_set_listeners: list[ResourceSetListener] = []
+        self.model_version_listeners: list[ModelVersionListener] = []
 
-    def add_resource_set_listener(self, listener: ResourceSetListener) -> None:
+    def add_model_version_listener(self, listener: ModelVersionListener) -> None:
```
sanderr: I think we often use `register_...`

sanderr (approved):

changelogs/unreleased/model-version-listener.yml
```diff
@@ -0,0 +1,7 @@
+description: >
+  Added the `ModelVersionListener` extension point. A listener registered on the orchestration service is notified, in
+  the transaction that writes them, of the resource sets a model version was written with. Unlike the compile and
+  environment listeners it takes part in that transaction, so an extension can maintain data derived from the
+  resources without it ever being committed out of step with them.
```
arnaudsjs: What does "being committed out of step with them" mean?

changelogs/unreleased/model-version-listener.yml
```diff
@@ -0,0 +1,7 @@
+description: >
+  Added the `ModelVersionListener` extension point. A listener registered on the orchestration service is notified, in
+  the transaction that writes them, of the resource sets a model version was written with. Unlike the compile and
```
arnaudsjs: What do you mean with "the transaction that writes them"? Don't you mean "in the transaction started by the caller"?

src/inmanta/server/services/model_version_listener.py
```diff
@@ -0,0 +1,56 @@
...
+
+import asyncpg
+
+
+class ModelVersionListener(abc.ABC):
+    """
+    Base class for listeners on the model versions written to an environment.
+
+    Unlike the compile and environment listeners, this one takes part in the transaction that writes the resources: it
+    is handed that transaction's connection, and aborts the export if an exception is raised.
```
arnaudsjs:
```suggestion
    is handed that transaction's connection, and aborts the export (and the transaction) if an exception is raised.
```

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: @jptrindade Could you elaborate? I don't see why the database slice would have to be started. The registration doesn't perform any database calls, does it?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -902,6 +911,20 @@ async def _put_version(
...
+                try:
+                    await listener.resource_sets_written(env.id, version, written_resource_sets, connection=connection)
+                except Exception:
+                    LOGGER.error(
+                        "Model version listener %s failed for version %d of environment %s. The export is aborted.",
+                        type(listener).__name__,
+                        version,
+                        env.id,
+                    )
+                    raise
```
arnaudsjs: This feels a bit out of its place to me. I would expect that the `insert_sets_and_resources` method calls into the listener. Like that we can't forget to call into the listener if we ever call the `insert_sets_and_resources` method from a different place. So I would move all the listener stuff onto the database slice. @jptrindade @sanderr Any opinions on this?

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -902,6 +911,20 @@ async def _put_version(
...
+            for listener in self.model_version_listeners:
+                try:
+                    await listener.resource_sets_written(env.id, version, written_resource_sets, connection=connection)
+                except Exception:
+                    LOGGER.error(
+                        "Model version listener %s failed for version %d of environment %s. The export is aborted.",
+                        type(listener).__name__,
+                        version,
+                        env.id,
+                    )
```
arnaudsjs: Wouldn't it be better to raise an exception here with the above-mentioned message or not catch the exception at all? The exception will be logged by the protocol layer. Like this will will have two log messages for the same issue.

src/inmanta/server/services/orchestrationservice.py
```diff
@@ -902,6 +911,20 @@ async def _put_version(
                 )
             except data.InvalidResourceSetMigration as e:
                 raise BadRequest(e.message)
+            # A listener failure aborts the export. A listener maintains data derived from these
+            # resources, so it has to be committed with them or not at all. The handler below only names the
```
arnaudsjs:
```suggestion
            # resources, so it has to be committed with them or not at all. The exception handler below only names the
```

arnaudsjs (changes requested):

src/inmanta/server/services/orchestrationservice.py
jptrindade: I don't agree. Our model version listener needs a guarantee that the model version was indeed persisted which happens here a bit above. And like the name implies, it is listening for the model version to be inserted and not the resource sets (although in the case that we have now we only care about the resource sets).

changelogs/unreleased/model-version-listener.yml
jptrindade: I meant in the transaction that writes the resource_sets and model version. But "in the transaction started by the caller" could be applied as well

changelogs/unreleased/model-version-listener.yml
jptrindade: It means that the derived data and the resource data itself are never out of sync. One is never written without the other being written as well.

src/inmanta/server/services/orchestrationservice.py
jptrindade: I've loosened the description a bit. We can register it in `prestart` or `start` as we wish. [It is just the case that the current implementation of the lsm part requires it to be in `start` ](https://github.com/inmanta/inmanta-lsm/pull/2499/changes#diff-0b1d8a1e4546f7fc39568c72aadf66e36837a6c062428176c328f19635f33200R261-R265)

arnaudsjs: Sorry. I still don't get it. Can you describe what would go wrong if we would move the call to `self.orchestration_service.register_model_version_listener()` to the `prestart()` for the specific situation in the lsm extension?

arnaudsjs: I understand now. You can ignore the comment above.

src/inmanta/server/services/orchestrationservice.py
arnaudsjs: I think I see what your intention was. Isn't it rather weird to call the listener method `resource_sets_written()` in that case? Wouldn't it be better to call it `notify_new_model_version(environment, version, new_or_updated_resource_sets, connection)`?

arnaudsjs (changes requested):
See comment

arnaudsjs (approved):

Issue #10480, opened 2026-06-16
Modify the scheduler and the appropriate endpoints to accept specific resource_ids on dryrun


jptrindade:
Same issue as https://github.com/inmanta/inmanta-core/issues/10479 but for dryrun.

PR #10801, opened 2026-09-09
Added support for scoped dryrun on agent-side

arnaudsjs:
# Description

Add an optional `resources` argument to the internal `do_dryrun` server-to-agent endpoint, which restricts the dryrun to the given resources.

Part of https://github.com/inmanta/inmanta-core/issues/10480

# Self Check:

- [x] Changelog entry

### DIFF
--- a/changelogs/unreleased/do-dryrun-resources-filter.yml
+++ b/changelogs/unreleased/do-dryrun-resources-filter.yml
@@ -1,0 +1,5 @@
+description: >-
+  Add an optional `resources` argument to the internal `do_dryrun` server-to-agent endpoint, which restricts the
+  dryrun to the given resources.
+change-type: patch
+destination-branches: [master, iso9]
--- a/src/inmanta/agent/agent_new.py
+++ b/src/inmanta/agent/agent_new.py
@@ -307,17 +307,36 @@ class Agent(SessionEndpoint):
         return 200
 
     @protocol.handle(methods.do_dryrun, env="tid", dry_run_id="id")
-    async def run_dryrun(self, env: uuid.UUID, dry_run_id: uuid.UUID, agent: str, version: int) -> Apireturn:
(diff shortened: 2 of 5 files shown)


jptrindade (changes requested):
What we have now is fine, but we still want to store the information regarding which resources we actually triggered in the `DryrunReport`. I realize that this was missing from the ticket description but I believe that this is something that we should also do in this PR.
Also keep in mind that there is also `DryRunReport` (with a different capitalization). I'm a bit unfamiliar with the dryrun mechanism so I'm not sure if we still use both and what endpoints we actually use

arnaudsjs:
> What we have now is fine, but we still want to store the information regarding which resources we actually triggered in the DryrunReport. I realize that this was missing from the ticket description but I believe that this is something that we should also do in this PR.

Every resource that was checked by the dryrun is stored in the db. The API call to fetch the result of a dryrun includes all resources and their dryrun status. So I believe all information is there to know which resources were triggered by the dryrun. Or do you mean we should be able to know the resources that were triggered before we run the dryrun itself? Because now we only know when the full dryrun has finished as the changes dictionary is only populated when the dryrun for a resource has been executed.

jptrindade:
> Because now we only know when the full dryrun has finished as the changes dictionary is only populated when the dryrun for a resource has been executed.

I believe that the changes dictionary is only populated when something actually changes (correct me if I'm wrong).
So we don't know if we did a dryrun of everything and everything succeeded or if we did a dryrun of a resource only and that resource succeeded (while others might have failed)

arnaudsjs:
That's indeed not correct. We also populate the changes dictionary when nothing has changed, but the web-console doesn't show these resources. That's probably where the confusion comes from.

jptrindade (approved):

sanderr (approved):

PR #10808, opened 2026-09-11
docs(lsm): document web_* annotations on lifecycle states and transfers

### LINKED ISSUE #10564: Docs: annotations on lifecycle states and transfers (attributes_metadata.md)
_(Filed by Claude on behalf of Bart.)_ Part of the web-console annotations epic: inmanta/web-console#7091.

## Goal

Document the `web_*` vocabulary for lifecycle **states** and **transfers** as the single source of truth, alongside the existing `web_presentation` / `web_suggested_values` entries model authors already consult.

## Work

Extend `docs/lsm/attributes_metadata/attributes_metadata.md` with an "Annotations on lifecycle states and transfers" section covering:
- The full `web_*` vocabulary: on a state `web_label`, `web_icon`, `web_description`; on a transfer `web_button_label`, `web_icon`, `web_button_type` (primary/secondary/tertiary/link), `web_button_variant` (danger/warning), `web_confirm`, `web_advanced_state`.
- That state/transfer annotations are read off the constructed `lsm::State` / `lsm::StateTransfer` instance (so they can come from plugins/variables), unlike entity/attribute annotations which are type-level defaults.
- The reserved-prefix convention (`web_*`, `inmanta_*`) and that annotations are opaque pass-through (no upstream validation).
- The documentation-tab `setState` button precedence (codeblock override > annotation default > hard-coded default).

Design (merged): [inmanta/designs#160](https://github.com/inmanta/designs/pull/160). Backend: [inmanta/lsm#990](https://github.com/inmanta/lsm/pull/990), [inmanta/inmanta-lsm#2428](https://github.com/inmanta/inmanta-lsm/pull/2428).

bartv:
Documents the `__annotations` attribute on `lsm::State` / `lsm::StateTransfer` and the full `web_*` vocabulary the web console reads from it, as the single source of truth for model authors. Also documents the `setState` code block of the documentation tab, which was undocumented so far, because the annotations supply its defaults.

Goes to `master` and `iso9`, which is where the feature ships: all six web-console PRs and the extension field ([inmanta-lsm#2428](https://github.com/inmanta/inmanta-lsm/pull/2428)) went to both. The new section sits before *Documentation tabs*, where the page is identical on both branches, so the cherry-pick applies cleanly (verified).

What a reviewer cannot see in the diff: the text describes what the web console actually does today, which deviates from the design on two points. `web_button_type` only styles the documentation-tab button, not the Actions-menu entry (no PatternFly equivalent for a menu item), and `web_confirm` on an `on_update` transfer has no effect because Edit navigates to the form instead of opening a dialog. Both are stated explicitly.

Screenshots taken from `basic-service` on the local-setup.

closes #10564


docs/lsm/attributes_metadata/attributes_metadata.md
```diff
@@ -126,6 +129,126 @@ __annotations__ = lsm::RelationAnnotations(
...
+| Annotation           | Effect                                                                                         |
+| -------------------- | ---------------------------------------------------------------------------------------------- |
+| `web_button_label`   | Text of the menu entry.                                                                        |
+| `web_icon`           | [Font awesome icon](https://react-icons.github.io/react-icons/icons/fa/) on that entry.        |
+| `web_button_variant` | `danger` or `warning`: colours the entry to signal how disruptive the transfer is.             |
+| `web_advanced_state` | When `true`, moves the entry into an *Advanced* section that is collapsed by default.          |
+| `web_confirm`        | Text of the confirmation prompt shown before the transfer is executed.                         |
+| `web_button_type`    | `primary`, `secondary`, `tertiary` or `link`: emphasis of the button in the documentation tab. |
+
+The example above results in a menu entry that is tucked away behind *Advanced*:
```
Hugo-Inmanta: What example ? I think it is missing and it would be nice to have a concrete example of "these values produce this visual output" we're missing the "these values" part.

Hugo-Inmanta (changes requested):

docs/lsm/attributes_metadata/attributes_metadata.md
bartv: (Claude replying on behalf of Bart.)

You're right - the example existed, but it sat 45 lines up with the key table and the whole *State presentation* subsection in between, so the pairing was invisible. Fixed in 929fbe8af by splitting it:

- the state example now sits in *State presentation*, right after the table, with a sentence saying which part of the badge comes from the annotations and which from `label`;
- the transfer example sits directly above the two screenshots, and the lines between them name the annotation responsible for each thing you see ("`web_advanced_state` keeps the entry out of the main list, `web_button_label` names it, `web_icon` gives it the sliders icon and `web_button_variant` colours it", then "`web_confirm` supplies the prompt");
- the `setState` section points back at the same transfer for `web_button_type`, which is the one key with no effect on the menu entry, so it now has a concrete example too.

Both screenshots are of exactly that transfer, captured from `basic-service` on the local-setup, so the values and the output really are the same thing.

Hugo-Inmanta (approved):

AronH99:
I am bit confused why the form annotations are mixed in with the lifecycle state annotations?
I thought those 2 implementation were kind of separate things? 
That's how we worked on it at least but correct me if I am wrong.

Also if these docs are meant to represent the full potential/changes for the form annotations it's lacking some things as well:
- Graphql flavour examples (https://github.com/inmanta/web-console/issues/7010)
- No cascading field examples (https://github.com/inmanta/web-console/issues/7011)

bartv:
(Claude on behalf of Bart.) Thanks Aron - both points are fair, and the second one turned out to be bigger than the two features you named.

**On mixing them:** they are not mixed, they are sibling `###` sections on one page - *Annotations on lifecycle states and transfers* next to *Documentation tabs*, *Suggested values* and *Form tabs*. You are right that the two implementations are separate things; the page is deliberately not. That was a decision in the design ([designs#160](https://github.com/inmanta/designs/pull/160)): *"one page documents every `web_*` key, with states/transfers and entities/attributes together"*, chosen over a separate web-console page for discoverability, so a model author looks up `web_*` in one place. Happy to split it into two pages if you prefer, but that is a change to the decision, not to this PR.

**On the missing features:** [#10813](https://github.com/inmanta/inmanta-core/pull/10813) documents the graphql flavor and cascading fields. While writing it I checked the rest of the epic and the gap is wider: three more features ship in ISO9 with no docs there.

| Feature | web-console | documented on master | documented on iso9 |
| --------------------------------------------------------- | ---------------- | -------------------- | ------------------ |
| label/value suggestions ([#7008](https://github.com/inmanta/web-console/issues/7008)) | master + iso9 (3.2.0) | yes | no |
| parameter-name variables ([#7009](https://github.com/inmanta/web-console/issues/7009)) | master + iso9 (3.2.0) | yes | no |
| form tabs ([#7012](https://github.com/inmanta/web-console/issues/7012)) | master + iso9 (3.2.0) | yes | no |
| graphql flavor ([#7010](https://github.com/inmanta/web-console/issues/7010)) | master + iso9 | [#10813](https://github.com/inmanta/inmanta-core/pull/10813) | [#10814](https://github.com/inmanta/inmanta-core/pull/10814) |
| cascading fields ([#7011](https://github.com/inmanta/web-console/issues/7011)) | master + iso9 | [#10813](https://github.com/inmanta/inmanta-core/pull/10813) | [#10814](https://github.com/inmanta/inmanta-core/pull/10814) |

The three older ones were documented in inmanta-core#10596, which went to master only even though the features went to both branches. [#10814](https://github.com/inmanta/inmanta-core/pull/10814) fixes that by taking the page from master verbatim onto iso9, which also makes the two pages identical again so the next change to it is a single PR with both destination branches.

One more thing worth knowing, unrelated to the PRs: the `suggestions-showcase-service` registered in the shared local-dev orchestrator still carries the pre-release annotation syntax (`"${form:$.site}"` with a colon, and dotted filter keys like `candidate_attributes.site`). Both are model errors against what shipped (`${form.site}`, nested filter objects). The model in `front-end-lsm-test-model` master is correct - the environment's catalog is stale because the v1 binding stays at version 0 - so anyone developing against that environment will see error states rather than working suggestions.

PR #10809, opened 2026-09-11
docs(lsm): document web_* annotations on lifecycle states and transfers (iso9)

bartv:
iso9 counterpart of [#10808](https://github.com/inmanta/inmanta-core/pull/10808). The documented section is byte-identical; only the surrounding page differs, because iso9 does not carry the form-tabs and label/value-suggestion additions that went to master only. That makes the tail of the file diverge, so the merge tool cannot cherry-pick the master commit, and the two branches each get their own change entry.

This belongs on iso9 because that is where the feature ships: all six web-console PRs went to `master` and `iso9`, and so did the extension field ([inmanta-lsm#2428](https://github.com/inmanta/inmanta-lsm/pull/2428)). iso8 has neither.

part of #10564


bartv:
Closing: no longer needed. Moving the new section up to sit before *Documentation tabs*, where the page is identical on both branches, makes the cherry-pick apply cleanly, so [#10808](https://github.com/inmanta/inmanta-core/pull/10808) now carries `destination-branches: [master, iso9]` and the merge tool handles iso9 by itself. The split only existed because the section was appended at the end of the file, which is exactly where the two branches diverge.

PR #10813, opened 2026-09-14
docs(lsm): document the graphql suggestion flavor and cascading fields

bartv:
Follows up on Aron's review of [#10808](https://github.com/inmanta/inmanta-core/pull/10808): two shipped form-annotation features had no documentation at all, the `graphql` flavor of `web_suggested_values` ([web-console#7010](https://github.com/inmanta/web-console/issues/7010)) and the `${form.*}` / `${self.*}` references that make one field's suggestions depend on another ([web-console#7011](https://github.com/inmanta/web-console/issues/7011)).

Adds a subsection for each, alongside the literal and parameters flavors: the fields of the `query` dict, the navigational-only jsonpath restriction on `label`/`value`, the camelCase filter keys (a dotted key is a model error), and for cascading fields the `form` vs `self` scoping, the blocking and re-query behaviour, and the two form-level model errors. `${environment}` is documented too - the console has always substituted it, it was just never written down.

Examples are taken from the showcase service in `front-end-lsm-test-model` and checked against the implementation, not against the issue text, which differs in places (jsonpath rather than DictPath, `${form.x}` rather than the older `${form:$.x}`).

Both features ship on `iso9` too, so the change entry is `[master, iso9]`. Its precondition is met: [#10814](https://github.com/inmanta/inmanta-core/pull/10814) is merged, the two pages are byte-identical again, and this commit cherry-picks onto the current iso9 tip without conflict (verified).


AronH99 (approved):
Looks quite good/correct to me

bartv:
(Claude on behalf of Bart.) Restructured both PRs so the same content is not up for review twice.

Before, [#10814](https://github.com/inmanta/inmanta-core/pull/10814) carried this PR's 151 lines as well, because I built its branch from master *plus this commit* instead of from master. Now:

- [#10814](https://github.com/inmanta/inmanta-core/pull/10814) is a pure backport of what is already on master (148 lines: label/value, parameter variables, form tabs), destination `[iso9]`. Merge it first.
- This PR keeps only the new material and its change entry is now `[master, iso9]`, so the merge tool ships it to both branches by itself.

Verified locally in that order: #10814 onto iso9, then this commit cherry-picked on top applies without conflict and leaves the iso9 page byte-identical to master's.

PR #10831, opened 2026-09-18
Allow `object` as a type annotation wherever `Any` is allowed in a method definition

sanderr:
# Description

`strict_typing=False` allows less strict annotations on REST methods, but it only accepted `Any`, not `object`. This PR expands it to allow both. Using `object` allows us to be less strict with type annotations rather than to ignore all type checking entirely for those fields.

# Self Check:

- [x] Attached issue to pull request — ~~no issue~~, spun out of the `deploy_filtered` cleanup, where `Mapping[str, object]` is the annotation we want for a GraphQL filter body
- [x] Changelog entry
- [x] ~~End user documentation is included or an issue is created for end-user documentation~~

# Reviewer Checklist:


changelogs/unreleased/object-annotation-non-strict.yml
```diff
@@ -0,0 +1,9 @@
+description: >-
+  A method definition can now use `object` as a type annotation wherever `Any` was already allowed, i.e. on a
+  method that sets `strict_typing=False`. Both describe an unconstrained JSON value, but `object` still requires
+  a handler to narrow the value before using it, so opting out of strict typing no longer forces the loosest of
+  the two annotations. `object` remains rejected on a method that uses strict typing.
```
sanderr:
```suggestion
  method that sets `strict_typing=False`.
```

tests/protocol/test_protocol.py
```diff
@@ -1000,13 +1000,13 @@ def test_method2(name: Iterator[str]) -> None:
 
         @auth(auth_label=const.CoreAuthorizationLabel.TEST, read_only=False)
         @protocol.typedmethod(path="/test", operation="PUT", client_types=[const.ClientType.api])
-        def test_method3(name: list[object]) -> None:
+        def test_method3(name: list[tuple]) -> None:
```
sanderr, line 1003: This is a test that verifies the error message when an invalid type is used. `object` is no longer invalid in the same way.

arnaudsjs (approved):

